PR #3947, opened 2021-02-16
module: Heater, extended ControlPid to support different Kd for risin…

xbladen:
module: Heater, extending ControlPID by two new variables allowing different Kd for rising and falling temp_diff

Introducing two new variables self.Kd_pos and self.Kd_neg filled by config file with default 0.
If not contained in configfile, they are overwritten by self.Kd, that still needs to be in configfile.
If both are submitted, self.Kd is unused, probably someone will work with Kd and Kd_neg, for example.
This gives the possibility to use two different Kd values for rising and falling temp_fiff of the heater.
Typically the heater has a much longer decay time than heat time. This can lead to a situation using a high Kd,
that a falling temperature above target temperatur leads to heating up, much more than needed together with
slow decay. Temperature is oscillating above and never reaches target temperature.
One possibility is to lower Kd, but this can lead to low performance of PID because of a large overshoot on heatup.
Using different Kd for rising/falling can minimize overshoot on heatup (large Kd)
and oscillating (small Kd) on falling temperature.
This is not covered by PID-tuning. A good starting point is the Kd from PID-tune and provide additional
Kd_neg at maybe 1/10 of Kd, depending on the fractional of heat-time to cool-time of some degree.

Line 268 does not need to be changed ...

Signed-off-by: Stefan Verse <verse@amotronics.de>

### DIFF
--- a/klippy/extras/heaters.py
+++ b/klippy/extras/heaters.py
@@ -178,6 +178,12 @@ class ControlPID:
         self.Kp = config.getfloat('pid_Kp') / PID_PARAM_BASE
         self.Ki = config.getfloat('pid_Ki') / PID_PARAM_BASE
         self.Kd = config.getfloat('pid_Kd') / PID_PARAM_BASE
+        self.Kd_pos = config.getfloat('pid_Kd_pos', 0.) / PID_PARAM_BASE
+        self.Kd_neg = config.getfloat('pid_Kd_neg', 0.) / PID_PARAM_BASE
+        if not self.Kd_pos:
+            self.Kd_pos = self.Kd
+        if not self.Kd_neg:
+            self.Kd_neg = self.Kd
         self.min_deriv_time = heater.get_smooth_time()
         imax = config.getfloat('pid_integral_max', self.heater_max_power,
                                minval=0.)
@@ -202,7 +208,10 @@ class ControlPID:
         temp_integ = self.prev_temp_integ + temp_err * time_diff
         temp_integ = max(0., min(self.temp_integ_max, temp_integ))
         # Calculate output
-        co = self.Kp*temp_err + self.Ki*temp_integ - self.Kd*temp_deriv
+        if temp_deriv>=0:
+            co = self.Kp*temp_err + self.Ki*temp_integ - self.Kd_pos*temp_deriv
+        else:
+            co = self.Kp*temp_err + self.Ki*temp_integ - self.Kd_neg*temp_deriv
         #logging.debug("pid: %f@%.3f -> diff=%f deriv=%f err=%f integ=%f co=%d",
         #    temp, read_time, temp_diff, temp_deriv, temp_err, temp_integ, co)
         bounded_co = max(0., min(self.heater_max_power, co))
@@ -265,8 +274,7 @@ class PrinterHeaters:
     def setup_sensor(self, config):
         modules = ["thermistor", "adc_temperature", "spi_temperature",
                    "bme280", "htu21d", "lm75", "rpi_temperature",
-                   "temperature_mcu", "ds18b20"]
-
+                   "temperature_mcu"]
         for module_name in modules:
             self.printer.load_object(config, module_name)
         sensor_type = config.get('sensor_type')


KevinOConnor:
Interesting, thanks.  Do you have any links describing research on this approach, or testing results showing the improvement on real-world heaters?

Do you have information or insight on how the given values could be auto-tuned?

As high-level feedback, I prefer to "go slow" with PID changes.  I would generally prefer to keep the PID algorithm simple, and avoid options that aren't critical.

-Kevin

xbladen:
Hi Kevin,
i just tested it on my heater, auto-tune gave a high Kd of 150.
It is a artillary sidewinder with 63W heater-cartridge – means fast heatup. Target at 255 degree ended with waiting forever at 256 to 257 degree.
I could see, that when the temperature decreased, the heater switched on with pwm value of 87% or something similar. This was followed by a temperature-ris by 2 degree.
Every time the temperature was settling down it was heated up inbetween by 1 to 2 degree.
This is also dependent on the accuracy of the thermistor and adc.
In my case the original 100k thermistor was broken, so i replaced by a 10k thermistor i already had. Maybe i have more noise than others on temperature-measurement. Noise with high Kd gives high pwm-values.
So lowering Kd could be the solution, but degrades PID performance on heating up.
I think for autotune the qutioet of time_to_heatup to time_to_cooldown when cycling around target could give a meaningful value, but i don’t know how pid-tuning is calculating ist values.
The cooldown in my system is so slow, that Kd could be set to zero for this direction, because there is no noticeable undershoot by cooldown, so Kd-values for cooldown are only a guess.
The Ki by autotune for heatup is faster than my cooldown-time, so this is stable with no undershoot. You could think about diiferent sets of Kp,Ki,Kd for heatup and cooldown, but i guess cooldown wont be noticeable faster by that.
On heatup  the overshoot is clearly visible.

Found some links :
https://www.splatco.com/skb/2657.htm
https://support.industry.siemens.com/cs/document/66995021/how-are-the-automatic-pid-controller-settings-(optimization)-made-for-asymmetrical-temperature-control-paths-in-step-7-(tia-portal)-v11-and-higher-?dti=0&lc=en-DK

I could later make a log-file of this by deleting my new Kd_neg from printer.cfg.
You should look at issue https://github.com/KevinOConnor/klipper/issues/3948
And tell me what you think of this – code seems to be working.

Stefan

Von: KevinOConnor [mailto:notifications@github.com]
Gesendet: Freitag, 19. Februar 2021 02:38
An: KevinOConnor/klipper
Cc: Verse, Stefan; Author
Betreff: Re: [KevinOConnor/klipper] module: Heater, extended ControlPid to support different Kd for risin… (#3947)


Interesting, thanks. Do you have any links describing research on this approach, or testing results showing the improvement on real-world heaters?

Do you have information or insight on how the given values could be auto-tuned?

As high-level feedback, I prefer to "go slow" with PID changes. I would generally prefer to keep the PID algorithm simple, and avoid options that aren't critical.

-Kevin

xbladen:
One additional note :
The way i added it to PID, it doesn’t need to be used and maybe others have no need for it.
Standard behaviour would not change, but you have the ability to change, if there is a need for.
If Kd_pos or Kd_neg are not given, it is the old PID algorythm.


Von: KevinOConnor [mailto:notifications@github.com]
Gesendet: Freitag, 19. Februar 2021 02:38
An: KevinOConnor/klipper
Cc: Verse, Stefan; Author
Betreff: Re: [KevinOConnor/klipper] module: Heater, extended ControlPid to support different Kd for risin… (#3947)


Interesting, thanks. Do you have any links describing research on this approach, or testing results showing the improvement on real-world heaters?

Do you have information or insight on how the given values could be auto-tuned?

As high-level feedback, I prefer to "go slow" with PID changes. I would generally prefer to keep the PID algorithm simple, and avoid options that aren't critical.

-Kevin

xbladen:
Hi Kevin.
i have added 3 logs with high Kd, Kd is choosen very high, to show the effect, that happened before to me - normally i go with lower Kd araound 100 and Kd_neg at 15.
In Kd450_emergency_stop you can see heatup M109 to 255 with overshoot and rippeling about 258, never reaching target.
At this point i could not control the printer further, because it still waited to reach target. Even firmware restart did not work. Emergency stop worked. I think M112 would have done it,too. But the point is, in a print job it would have waited forever, when not watching the printer.
Lowering Kd_neg results in working M109, two samples with different Kd_neg and same Kd are provided.

The Problem at this is, that you can not estimate your time constants. There is one for heating, one for cooling, one for heating while extruding and one for cooling while extruding and extruding at different speeds gives a huge bunch of time constants.

So maybe the only possible way would be to calculate Kd by extrusion speed for heat and cool, but who should calibrate theese curves.

So it is a tradeoff between working wait_for target_temperature and good performance while extruding.
On the other hand, the wait_for_target_temperature routine could tolerate bigger differences over target_temperature. When you start extruding, the temperature will come done further.

Stefan


Von: KevinOConnor [notifications@github.com]
Gesendet: Freitag, 19. Februar 2021 02:38
An: KevinOConnor/klipper
Cc: Verse, Stefan; Author
Betreff: Re: [KevinOConnor/klipper] module: Heater, extended ControlPid to support different Kd for risin… (#3947)


Interesting, thanks. Do you have any links describing research on this approach, or testing results showing the improvement on real-world heaters?

Do you have information or insight on how the given values could be auto-tuned?

As high-level feedback, I prefer to "go slow" with PID changes. I would generally prefer to keep the PID algorithm simple, and avoid options that aren't critical.

-Kevin

KevinOConnor:
Interesting.  I'd say the next steps would be to try and build an auto-calibration routine, or write documentation explaining how one can set the desired Kd_neg/pos.  Then try to get a dozen or so volunteers to perform that calibration on their printers followed by various real-world prints.  If we can observe a general improvement and no significant regressions then it would be a good candidate for merging.

As before, this is something that I would "go slow" with.  I'd guess the best case scenario is a merge in 3-4 months.

Separately, the links above are interesting, but neither mentions using different Kd values (the first actually recommends not using different PID values for heaters).

Cheers,
-Kevin

Sineos:
Maybe an interesting read: https://portal.research.lu.se/portal/files/3633151/5463952.pdf

xbladen:
On the other hand it is a real, world problem – explicitely at my printer. That is where i came from.
Made an autotune PID, inserted the values into printer.cfg and started some prints – lets say 1 of 4 print ended sitting on 2-3 degrees above target waiting to settle down for minutes.
So i started to look into the reason and found, that the high Kd is heating up on falling temperatures to much.

This could be solved by, lower Kd ( which is against autotune Kd) or changing the wait_for _temprature routine(wider tolerance), or changing hardware or …

So i build it this way, that it can be used but doesn’t need to, at least it was not intended to replace existing autotune, just to fix my problem.
At least a complete Kp,Ki,Kp set for rising and falling temp would be probably a good idea, but for me a second Kd was ok, because the regulation is so slow.
Lets express it this way:
For my problem (no extrusion, fan off, initial heating) i need a special set of parameterst o reach target temperature .
If it comes to extrusion, maybe this are the wrong parameters, because falling temperatures need a higher Kd to keep temperature in range.
If the fan is on i maybe would take the next set of Parameters.
Does it improve temperature stability – maybe.
Will it improve prints – maybe.

This uses asymetrical pid :
https://ieeexplore.ieee.org/document/4153300

Found this, could be promising, too:
https://stackoverflow.com/questions/48190821/asymmetric-pid-control-loop
perhaps a model for forward correction depending on extrusion and fan-cooling and outer temperature is what really helps, keep pid itself as it is.

Interesting the added pdf, too.

This is about enhancing pid, even feed_forward :
https://www.controleng.com/articles/when-is-pid-not-the-answer/

Perhaps i take a closer look on temperatures itself –
The performance during heatup is much less critical than during print.
Maybe i watch the temperature during print, if they fluctuate over 5 degrees, then it is unneccessary to keep 1 degree tolerance for heatup, maybe get the tolerance parameter from configfile.


Von: KevinOConnor [mailto:notifications@github.com]
Gesendet: Dienstag, 23. Februar 2021 23:25
An: KevinOConnor/klipper
Cc: Verse, Stefan; Author
Betreff: Re: [KevinOConnor/klipper] module: Heater, extended ControlPid to support different Kd for risin… (#3947)


Interesting. I'd say the next steps would be to try and build an auto-calibration routine, or write documentation explaining how one can set the desired Kd_neg/pos. Then try to get a dozen or so volunteers to perform that calibration on their printers followed by various real-world prints. If we can observe a general improvement and no significant regressions then it would be a good candidate for merging.

As before, this is something that I would "go slow" with. I'd guess the best case scenario is a merge in 3-4 months.

Separately, the links above are interesting, but neither mentions using different Kd values (the first actually recommends not using different PID values for heaters).

Cheers,
-Kevin

KevinOConnor:
Any further updates on this?

-Kevin

KevinOConnor:

-Kevin

tanaes:
Apologies for the necromancy, but I have been seeing quite a number of folks running Klipper on the Artillery Discord experiencing similar problems with PID tuning. In particular, a number of users are reporting bed temperatures that overshoot the target and then never come back down, eventually leading to a timeout error. In most cases, replacing the Klipper-tuned PID values with their previous Marlin-tuned values gets rid of the problem. 

At a bit of a loss as to why this might be. The only meaningful difference I can find in the code between the two tuning loops is that the [Marlin code](https://github.com/MarlinFirmware/Marlin/blob/4f8191b4818b97bd20eb9db2042dc07c97cce6cc/Marlin/src/module/temperature.cpp#L620) has a Kp adjustment factor that changes between hotend (0.6) and bed (0.2) tuning, where Klipper seems to use 0.6 for all. (Caveat: I am easily lost in C code!) 

If the problem really is an overestimated Kp due to asymmetric temperature response, is it possible that this (presumably empirically-derived) correction is worth implementing?

RafaelSchridi:
I'm one of the users that @tanaes is talking about, I've tested both Marlin and Klipper PID tuning, and with Klipper sometimes it just overshoots and and never goes down as seen here:
image
While tuning with marlin also overshoots but it doesn't stay there as seen here:
image
The values produced by each PID tune are also vastly different.

image

I'm not sure if it is the right place to put this, but the original request seems to experience the same issues with potentially a solution.

RafaelSchridi:
This wasn't the right place, continuing in #4427

PR #3952, opened 2021-02-17
module: host_temperature

shavrin777:
Replacing rpi_temperature by host_temperature with expanded functionality.
This makes possible to change the path to the temperature sensor.

Signed-off-by: Evgenii Shavrin shavrin0591@gmail.com

### DIFF
--- a/docs/Command_Templates.md
+++ b/docs/Command_Templates.md
@@ -265,7 +265,7 @@ The following are common printer attributes:
   from the sensor.
 - `printer["lm75 <sensor_name>"].temperature`: The last read
   temperature from the sensor.
-- `printer["rpi_temperature <sensor_name>"].temperature`: The last read
+- `printer["host_temperature <sensor_name>"].temperature`: The last read
   temperature from the sensor.
 - `printer["temperature_sensor <config_name>"].temperature`: The last read
   temperature from the sensor.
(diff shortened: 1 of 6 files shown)


KevinOConnor:
Thanks.  Can you provide some additional information on how you plan to use this extended functionality?  What machine does this work on, and what are the paths a user may configure?

Separately, a non-backwards compatible change like this would require an entry in docs/Config_Changes.md .  If we were to change the name of the module, I would prefer temperature_host.py .

-Kevin

shavrin777:
> Thanks. Can you provide some additional information on how you plan to use this extended functionality? What machine does this work on, and what are the paths a user may configure?
I use it to display CPU temperature. Klipper runs on old laptop based on Intel Atom N450. And default path for RPi doesn't work.
User can specify path to every file that contains temperature in format like "49000" for 49°C.

KevinOConnor:
Thanks.  I made a couple of minor changes and committed this (3d35c003).

-Kevin

theopensourcerer:
@shavrin777 See the discussion here: 

https://github.com/KevinOConnor/klipper/commit/3d35c003db3d4fd25a3cadde1e70373cf450e2f7

It looks like there might be an error in the pull request somewhere.

PR #3956, opened 2021-02-18
Multi mcu homing

KevinOConnor:
This PR adds initial (currently incomplete) support for homing and probing when steppers and their endstops are on different micro-controllers.  As an example, this may be useful with "toolhead boards" that have a Z probe attached, but don't directly control the Z stepper motors.

The main challenge with multi-mcu homing is having a reliable way to stop the steppers, even if there is a communication fault during homing.  (Otherwise, if the mcu with the endstop is unable to communicate with the mcu controlling the steppers, the steppers may continue to move the carriage past the endstop, potentially causing significant damage.)  To account for this issue, the code in this PR uses a system where each micro-controller announces that it is still active every 10ms and requires that it receive a message indicating all other mcus are still active at least every 25ms.  Should a mcu not receive that response, it will stop the homing sequence (stepper movement will cease).  Thus, even in the result of a communication failure, the steppers should not overshoot by more than 25ms of movement (eg, 0.500mm if homing at 20mm/s).

This PR is a work-in-progress.  The current code does "work" in that it will home, but it is currently too inacurate to be used for printing.  This is an early release that other developers can look at should they be interested.

Some known issues:
1) To improve accuracy the code will also need to change how it calculates the final homing position.  The current code uses the final stepper position, but that will need to change to the position that the stepper was at when the endstop signal occurred.  (With the code on this PR, there could be some small movement after the endstop signal is received, but prior to the propagation of that signal.)
2) If a multi-stepper axis (eg, stepper_z1) is used and those steppers are on different mcus, then this code does not currently handle that accurately.  (The propagation time of the endstop signal may be different for each mcu, and thus the overshoot on each mcu would need to be accounted for.)
3) The code on this branch may not behave correctly if one were to use multi-mcu homing on the endstops of a delta printer.  (Should a communication timeout occur, it may only stop one of the towers instead of stopping all three towers.)
4) The overshoot during multi-mcu homing is subtle.  It may make sense to add a "max overshoot" config parameter to ensure users are aware of the implications of enabling multi-mcu homing.

Despite the above limitations, this is a fairly significant feature addition to Klipper.  I think it may enable some very interesting hardware capabilities in the future.

@mattshepcar - fyi.

-Kevin

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -119,7 +119,9 @@ different names for the stepper (eg, `stepper_x` vs `stepper_a`).
 Below are common stepper definitions.
 
 See the [rotation distance document](Rotation_Distance.md) for
-information on calculating the `rotation_distance` parameter.
+information on calculating the `rotation_distance` parameter. See the
+[Multi-MCU homing](Multi_MCU_Homing.md) document for information on
+homing using multiple micro-controllers.
 
 ```
 [stepper_x]
@@ -152,8 +154,10 @@ microsteps:
 #   distance the axis travels for one full rotation of the final gear.
 #   The default is to not use a gear ratio.
 endstop_pin:
-#   Endstop switch detection pin. This parameter must be provided for
-#   the X, Y, and Z steppers on cartesian style printers.
+#   Endstop switch detection pin. If this endstop pin is on a
+#   different mcu than the stepper motor then it enables "multi-mcu
+#   homing". This parameter must be provided for the X, Y, and Z
+#   steppers on cartesian style printers.
 #position_min: 0
 #   Minimum valid distance (in mm) the user may command the stepper to
 #   move to.  The default is 0mm.
@@ -1611,7 +1615,9 @@ stepper_z config section.
 ```
 [probe]
 pin:
-#   Probe detection pin. This parameter must be provided.
+#   Probe detection pin. If the pin is on a different microcontroller
+#   than the Z steppers then it enables "multi-mcu homing". This
+#   parameter must be provided.
 #deactivate_on_each_sample: True
 #   This determines if Klipper should execute deactivation gcode
 #   between each probe attempt when performing a multiple probe
--- a/docs/Multi_MCU_Homing.md
+++ b/docs/Multi_MCU_Homing.md
@@ -1,0 +1,42 @@
+# Multiple Micro-controller Homing and Probing
+
+Klipper supports a mechanism for homing with an endstop attached to
+one micro-controller while its stepper motors are on a different
+micro-controller. This support is referred to as "multi-mcu
+homing". This feature is also used when a Z probe is on a different
+micro-controller than the Z stepper motors.
+
+This feature can be useful to simplify wiring, as it may be more
+convenient to attach an endstop or probe to a closer micro-controller.
+However, using this feature may result in "overshoot" of the stepper
+motors during homing and probing operations.
+
+The overshoot occurs due to possible message transmission delays
+between the micro-controller monitoring the endstop and the
+micro-controllers moving the stepper motors. The Klipper code is
+designed to limit this delay to no more than 25ms. (When multi-mcu
+homing is activated, the micro-controllers send periodic status
+messages and check that corresponding status messages are received
+within 25ms.)
+
+So, for example, if homing at 10mm/s then it is possible for an
+overshoot of up to 0.250mm (10mm/s * .025s == 0.250mm). Care should be
+taken when configuring multi-mcu homing to account for this type of
+overshoot. Using slower homing or probing speeds can reduce the
+overshoot.
+
+Stepper motor overshoot should not adversely impact the precision of
+the homing and probing procedure. The Klipper code will detect the
+overshoot and account for it in its calculations. However, it is
+important that the hardware design is capable of handling overshoot
+without causing damage to the machine.
+
+Should Klipper detect a communication issue between micro-controllers
+during multi-mcu homing then it will raise a "Communication timeout
+during homing" error.
+
+Note that an axis with multiple steppers (eg, `stepper_z` and
+`stepper_z1`) need to be on the same micro-controller in order to use
+multi-mcu homing. For example, if an endstop is on a separate
+micro-controller from `stepper_z` then `stepper_z1` must be on the
+same micro-controller as `stepper_z`.
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -41,18 +41,19 @@ communication with the Klipper developers.
     using adxl345 accelerometer hardware to measure resonance.
 - [Pressure advance](Pressure_Advance.md): Calibrate extruder
   pressure.
-- [Slicers](Slicers.md): Configure "slicer" software for Klipper.
+- [G-Codes](G-Codes.md): Information on commands supported by Klipper.
 - [Command Templates](Command_Templates.md): G-Code macros and
   conditional evaluation.
   - [Status Reference](Status_Reference.md): Information available to
     macros (and similar).
 - [TMC Drivers](TMC_Drivers.md): Using Trinamic stepper motor drivers
   with Klipper.
+- [Multi-MCU Homing](Multi_MCU_Homing.md): Homing and probing using multiple micro-controllers.
+- [Slicers](Slicers.md): Configure "slicer" software for Klipper.
 - [Skew correction](skew_correction.md): Adjustments for axes not
   perfectly square.
 - [PWM tools](Using_PWM_Tools.md): Guide on how to use PWM controlled
   tools such as lasers or spindles.
-- [G-Codes](G-Codes.md): Information on commands supported by Klipper.
 
 ## Developer Documentation
 
--- a/docs/_klipper3d/mkdocs.yml
+++ b/docs/_klipper3d/mkdocs.yml
@@ -94,14 +94,15 @@ nav:
       - Resonance_Compensation.md
       - Measuring_Resonances.md
     - Pressure_Advance.md
-    - Slicers.md
+    - G-Codes.md
     - Command templates:
       - Command_Templates.md
       - Status_Reference.md
     - TMC_Drivers.md
+    - Multi_MCU_Homing.md
+    - Slicers.md
     - skew_correction.md
     - Using_PWM_Tools.md
-    - G-Codes.md
   - Developer Documentation:
     - Code_Overview.md
     - Kinematics.md
--- a/klippy/mcu.py
+++ b/klippy/mcu.py
@@ -125,6 +125,9 @@ class MCU_trsync:
             s.note_homing_end()
         return params['trigger_reason']
 
+TRSYNC_TIMEOUT = 0.025
+TRSYNC_SINGLE_MCU_TIMEOUT = 0.250
+
 class MCU_endstop:
     RETRY_QUERY = 1.000
     def __init__(self, mcu, pin_params):
@@ -139,15 +142,27 @@ class MCU_endstop:
         self._rest_ticks = 0
         ffi_main, ffi_lib = chelper.get_ffi()
         self._trdispatch = ffi_main.gc(ffi_lib.trdispatch_alloc(), ffi_lib.free)
-        self._trsync = MCU_trsync(mcu, self._trdispatch)
+        self._trsyncs = [MCU_trsync(mcu, self._trdispatch)]
     def get_mcu(self):
         return self._mcu
     def add_stepper(self, stepper):
-        if stepper.get_mcu() is not self._mcu:
-            raise pins.error("Endstop and stepper must be on the same mcu")
-        self._trsync.add_stepper(stepper)
+        trsyncs = {trsync.get_mcu(): trsync for trsync in self._trsyncs}
+        trsync = trsyncs.get(stepper.get_mcu())
+        if trsync is None:
+            trsync = MCU_trsync(stepper.get_mcu(), self._trdispatch)
+            self._trsyncs.append(trsync)
+        trsync.add_stepper(stepper)
+        # Check for unsupported multi-mcu shared stepper rails
+        sname = stepper.get_name()
+        if sname.startswith('stepper_'):
+            for ot in self._trsyncs:
+                for s in ot.get_steppers():
+                    if ot is not trsync and s.get_name().startswith(sname[:9]):
+                        cerror = self._mcu.get_printer().config_error
+                        raise cerror("Multi-mcu homing not supported on"
+                                     " multi-mcu shared axis")
     def get_steppers(self):
-        return self._trsync.get_steppers()
+        return [s for trsync in self._trsyncs for s in trsync.get_steppers()]
     def _build_config(self):
         # Setup config
         self._mcu.add_config_cmd("config_endstop oid=%d pin=%s pull_up=%d"
@@ -157,7 +172,7 @@ class MCU_endstop:
             " rest_ticks=0 pin_value=0 trsync_oid=0 trigger_reason=0"
             % (self._oid,), on_restart=True)
         # Lookup commands
-        cmd_queue = self._trsync.get_command_queue()
+        cmd_queue = self._trsyncs[0].get_command_queue()
         self._home_cmd = self._mcu.lookup_command(
             "endstop_home oid=%c clock=%u sample_ticks=%u sample_count=%c"
             " rest_ticks=%u pin_value=%c trsync_oid=%c trigger_reason=%c",
@@ -173,8 +188,12 @@ class MCU_endstop:
         self._rest_ticks = rest_ticks
         reactor = self._mcu.get_printer().get_reactor()
         self._trigger_completion = reactor.completion()
-        etrsync = self._trsync
-        etrsync.start(print_time, self._trigger_completion, .250)
+        expire_timeout = TRSYNC_TIMEOUT
+        if len(self._trsyncs) == 1:
+            expire_timeout = TRSYNC_SINGLE_MCU_TIMEOUT
+        for trsync in self._trsyncs:
+            trsync.start(print_time, self._trigger_completion, expire_timeout)
+        etrsync = self._trsyncs[0]
         ffi_main, ffi_lib = chelper.get_ffi()
         ffi_lib.trdispatch_start(self._trdispatch, etrsync.REASON_HOST_REQUEST)
         self._home_cmd.send(
@@ -183,7 +202,7 @@ class MCU_endstop:
              etrsync.get_oid(), etrsync.REASON_ENDSTOP_HIT], reqclock=clock)
         return self._trigger_completion
     def home_wait(self, home_end_time):
-        etrsync = self._trsync
+        etrsync = self._trsyncs[0]
         etrsync.set_home_end_time(home_end_time)
         if self._mcu.is_fileoutput():
             self._trigger_completion.complete(True)
@@ -191,10 +210,10 @@ class MCU_endstop:
         self._home_cmd.send([self._oid, 0, 0, 0, 0, 0, 0, 0])
         ffi_main, ffi_lib = chelper.get_ffi()
         ffi_lib.trdispatch_stop(self._trdispatch)
-        res = etrsync.stop()
-        if res == etrsync.REASON_COMMS_TIMEOUT:
+        res = [trsync.stop() for trsync in self._trsyncs]
+        if any([r == etrsync.REASON_COMMS_TIMEOUT for r in res]):
             return -1.
-        if res != etrsync.REASON_ENDSTOP_HIT:
+        if res[0] != etrsync.REASON_ENDSTOP_HIT:
             return 0.
         if self._mcu.is_fileoutput():
             return home_end_time


ArkadiuszRaj:
Hi @KevinOConnor 

I have modified Pontus great toolboard tiny design, changing to use STM32G04 mpcu & easy high voltage inductive probe connection. G04 for FD-CAN mainly  (expect another PR on stm32g4 support ;) )

During rebuild of one of my Voron I plan to use set of 7 huvud boards: 
- 4 of them installed on Z motors (tmc2209, soft endostop & motor temperature)
- 2 for A-B motors, same config as Z ones
- 1 as toolboard (extruder motor, filament stop, z inductive probe, part cooling fan, hotend heater, temp sensor and fan
Everything connected  using something similar to ring network with CAN / FD-CAN and talking to RPi Zero via USBCAN dongle as Pontus suggested (CAN2) for the beginning.

WIll that be a good test bed for this PR?

KevinOConnor:
> During rebuild of one of my Voron I plan to use set of 7 huvud boards:

FYI, that's similar to my current setup.  I have a voron 2.4 with 3 huvud boards - one on the toolhead and two controlling the corexy motors.  I did purchase additional huvud boards with the intent of converting the Z motors to them, but ultimately decided to use a low-cost skr board for the Z instead (I'm using the skr-mini-mz).

I'm not sure which USB2CAN board you are planning on using, but I would not recommend purchasing the innomaker board.  It is reordering packets and it appears the stm32 chip on it has been been locked down so that no firmware fixes are possible.

-Kevin

ArkadiuszRaj:
Well, I have discussed that with Pontus and he gave me some insight on that toopis. Citing him:
`
https://github.com/candle-usb/candleLight_fw is a brilliant firmware for an USB-CAN dongle. Linux have native support for that CAN USB adapter standard. Boards running that can be a bit tricky to find though, there are some on alixpress, cantact is one of them.  I modified an old Huvud board and put an STM32F042 and use it as a CAN-USB adapter, it works really well :) 
`

Ultimatelly I can design and order from JLCPCB such a dongle simple dongle having F042 & FDCAN tranceiver.

brandonheller:
Looks like really useful code with some interesting problems to solve, and I'd love to try it out. 

One Q: Do you have any sense for the amount of inaccuracy to potentially expect, related to known issue 1 above?  Presumably, as long as the time between triggering an endstop and detecting the change is small... but what would that amount potentially be?

Thanks!

pecirep:
> amount of inaccuracy to potentially expect

it's timing based, so it depends on your homing speed. assuming up to 25ms, this means 0.25mm at 10mm/s, 0.5mm at 20mm/s...

WardBenjamin:
> 
> 
> Well, I have discussed that with Pontus and he gave me some insight on that toopis. Citing him:
> `https://github.com/candle-usb/candleLight_fw is a brilliant firmware for an USB-CAN dongle. Linux have native support for that CAN USB adapter standard. Boards running that can be a bit tricky to find though, there are some on alixpress, cantact is one of them. I modified an old Huvud board and put an STM32F042 and use it as a CAN-USB adapter, it works really well :) `
> 
> Ultimatelly I can design and order from JLCPCB such a simple dongle having only F042 & FDCAN tranceiver.

Unfortunately, the STM32F042 is out of stock virtually everywhere right now. You will need to use a G0/G4 with CAN and write some custom firmware. Fortunately a decent amount of candleLight is portable.

brandonheller:
I've got a Huvud doing all it needs to print, and this would remove 2 wires used for the X endstop.  I've added those notes here:
https://www.notion.so/Instructions-Getting-a-Huvud-Toolhead-Working-with-Klipper-611ebb87efba4e22ae3b0641e5267ab3

... but I'm hesitant to try out these branch.
(1) The original description says this is an incomplete PR - I'm not sure if it's from a performance/safety perspective, or from a 'basic functionality one.
(2) Current has a merge conflict.

@KevinOConnor -  would you be up to rebase this on master now that the CAN improvements are merged?  Thanks for those improvements, BTW.  Worked great for me on the first try today.  This would be icing on that cake.

KevinOConnor:
FYI, I rebased this branch.  Note though, that accuracy problems are still present - I would not recommend using multi-mcu homing/probing  on Z.

-Kevin

ETE-Design:
@KevinOConnor Will it be a general problem using multi-mcu for z probing, or is it just for now, and will it be bossible to fix in near feature? Really looking forward for this feature :-)

KevinOConnor:
This branch is still very much a "work in progress" - its intended audience is other developers.  I hope to fix all of its deficiencies.  I'll only merge it into the master branch when it is ready.

-Kevin

brandonheller:
tl;dr: Works For Me.

I rebased atop today's master w/no issues, recompiled for a Huvud and an SKR to get past an endstop format error, restarted to avoid a `No buffer space available` error for the CAN interface, updated my Klipper config:
```
[stepper_x]
...
# With no toolhead:
#endstop_pin: PC0
# With toolhead:
endstop_pin: head0:PA1 
````

... and homing seems to work fine.  2 more wires down on the toolhead.   Thanks!

**Comment by KevinOConnor** [comment]
I have rebased this branch and made several improvements:
1. Multi-mcu homing and probing should now be accurate.
2. Multi-mcu homing behaviour should be okay even for homing on a delta printer.  (In the event an endstop has a comms timeout, the homing sequence for the other towers should also now abort within ~250ms.)

This new code should, hopefully, be testable on a number of different printers.  However, there are still some caveats with this code:
1. If one does multi-mcu homing on a multi-stepper axis, and the steppers on that axis are also on separate mcus, then this code will not handle it correctly.  (After each home/probe attempt, the steppers may become out of sync.)
2. The endstop_phases code does not work correctly.
3. The mcu code will not compile on the Beaglebone PRU due to code size restrictions on that mcu.  Future infrastructure work will be needed on the PRU to free up space on that platform.

This code continues to use a 25ms max timeout when implementing multi-mcu homing.  That is, when using multi-mcu homing/probing, it is possible for the carriage to continue moving even after the endstop/probe has signaled.  For example, if homing at 10mm/s then the carriage may overshoot by as much as .250mm.  This timeout is not configurable, and I'm leaning towards not making it configurable (it's unclear that a user can reasonably choose an alternate setting).  The new code will determine the overshoot and (assuming it does not cause a mechanical failure) it will be accounted for - so the overshoot should not negatively impact accuracy.

If you run tests with this new code, please let me know your results (success or failure).

-Kevin

**Comment by zruncho3d** [comment]
Thanks Kevin.  I'll see if I can print up a small dial test indicator mount to verify repeatability.

**Comment by brandonheller** [comment]
@KevinOConnor I tried out the updated branch, and the code no longer works for me. 

With the printer in the middle, X and Y homing work fine.  But then at the extent of motion (back right), with endstops already triggered, when Y or Z homing is triggered, I get a 'move out of range' error and motors shut off and make a horrible skipping noise:

`Move out of range: 118.037 112.500 0.000 [0.000]`

This is a CoreXY w/slight-less-than-120mm square bed (Voron 0).  Probably relevant:

`
[stepper_x]
...
position_endstop: 118
position_max: 118

[stepper_y]
...
position_endstop: 117.5
position_max: 117.5
`

It does look like the code is compensating for the variability, as I can see the X value varying between 118.03-118.05 after X-only homing.  Increasing position_max a bit doesn't seem to help. 

To revert, it looks like I'll have reload the firmware on the SKR and the Huvud... argh:

`Command format mismatch: endstop_state ..` 

At least I'm now saving the SKR firmwares, so the build/menuconfig cycle doesn't add delay.

**Comment by revilo196** [comment]
@KevinOConnor I tried out the mult mcu homing. 
Using Probe(BL-Touch) and X-Axis-Endstop on a diffrent MCU.
YXZ homing sequence.

While homing Y everything works as normal
But as soon as the X homing starts klipper goes into shutdown. 

Im getting:

`Unable to obtain 'trsync_state' response`
`MCU 'tool' shutdown: Timer too close`

Im Using a 500Mbit CANbus for both MCU's

[klippy.log](https://github.com/KevinOConnor/klipper/files/6298280/klippy.log)

**Comment by KevinOConnor** [comment]
> I get a 'move out of range' error

Okay - that will need to be fixed.  In the interim, you should be able to update the homing sequence so that each axis is individually homed and then backed away from the endstop prior to homing any other axis.

> motors shut off and make a horrible skipping noise

Not sure why you'd get skipped steps in that case though.

-Kevin

**Comment by KevinOConnor** [comment]
> Im getting:
Unable to obtain 'trsync_state' response
MCU 'tool' shutdown: Timer too close

Okay, thanks.  The code definitely did something wrong.  I'll try to track it down.

-Kevin

**Comment by brandonheller** [comment]
@KevinOConnor  For whatever reason, the code drives the motors when already homed, in either +Y or +X, which would both trigger skips.  I can test this without the motors actually connected if that would help, to see if this is CoreXY-specific, without the horrible noises.  Let me know if that would useful; should be easy enough after reflashing the code. 

Has this code been tested on CoreXY yet?  Thanks.

**Comment by KevinOConnor** [comment]
> Has this code been tested on CoreXY yet?

Yes - I tested the code on my voron 2.4 printer.  Multi-mcu homing with each xy stepper on its own huvud board.

-Kevin

**Comment by KevinOConnor** [comment]
I've rebased this branch and fixed some bugs.  I think the "timer too close" issue should now be fixed.

There is still the issue of an overshoot during homing causing "invalid move" errors.  To workaround this issue for now, either 1) always back off from the axis after homing it, or 2) increase the position_min/position_max of the axis so that the valid range is a little past the position_endstop.

-Kevin

**Comment by revilo196** [comment]
Can confim the "timer too close" issue is fix works now fine for me. 
Tested with the same setup from above. (BL-Touch and X-Axis-Endstop) 
Will keep testing now.
Good work!

**Comment by revilo196** [comment]
Im Testing this branch now for a while. 
Im getting: 
`Communication timeout during homing` 
Sometimes when homing or Probing. In the worst case I need upto 4-attempt to Home successfully
Using 500Mbit CAN-Bus.

How can I test if my bus looses packets, or anything to investigate this issue ? (bus-usage...)

**Comment by KevinOConnor** [comment]
> Im getting:
Communication timeout during homing

If something wrong is occurring, then please run `M112` after the event and attach the full klipper log here.

-Kevin

**Comment by KevinOConnor** [comment]
> How can I test if my bus looses packets, or anything to investigate this issue ? (bus-usage...)

The Unix `ifconfig` tool will report bytes read/written and errors at the interface level.  It's also possible to look in the Klipper log file for `bytes_retransmit` and `bytes_invalid`, and other errors.

-Kevin

**Comment by henrikssn** [comment]
I am running this branch and is generally working fine for me.

One hurdle I stumped over is that I could not get `safe_z_home` to work (got the `Communication timeout during homing` error and klipper goes into shutdown), however doing the equivalent with `homing_override` works perfectly fine. Unfortunately I forgot to save the log, if this doesn't ring a bell I can give it another shot tonight and give you the logs.

I'm also running a 500Kbit CAN bus (similar to @revilo196) using a CandleLight firmware USB interface.

**Comment by KevinOConnor** [comment]
Alas, there really isn't much I can do without the full log.  I haven't seen a comms timeout on my setup.

-Kevin

**Comment by henrikssn** [comment]
> Alas, there really isn't much I can do without the full log. I haven't seen a comms timeout on my setup.
> 
> -Kevin

To reproduce this, I tried changing back to using `safe_z_home` and it has been working flawlessly for the last days. Only thing I changed is to allow the axes to move to -0.5mm during homing to allow for some delay in the homing routine.
 
 I do have occasional re-attempts of the probing during QGL but that could just as well be my printer or the probe and doesn't really disturb me much.

**Comment by KevinOConnor** [comment]
FYI, I've rebased this branch on top of the latest master branch.  At this point, I'm aware of two issues with this code:
1. If one does multi-mcu homing on a multi-stepper axis, and the steppers on that axis are also on separate mcus, then this code will not handle it correctly. (After each home/probe attempt, the steppers may become out of sync.)
2. The endstop_phases code does not work correctly when using multi-mcu homing.

To test this code:
`cd ~/klipper ; git fetch ; git checkout origin/work-homing-20210217 ; sudo service klipper stop ; make flash ; sudo service klipper start`

My plan at this point is to extract subsets of the change (where possible) and commit them to the master branch.  I'm hoping to complete the merge in the next few weeks.

-Kevin

**Comment by zruncho3d** [comment]
Thanks for the update.

Question for @KevinOConnor - to be really specific, if I have a QGL-ing, 4-Z-motor printer that has a separate CAN toolboard (Huvud) with a bed probe... will this code work?  

I'm asking the question to understand whether I should add a wire to the toolhead loom to connect back to the SKR that connects to all 4 Z motors.

**Comment by henrikssn** [comment]
@zruncho3d I am running precisely that setup on my Voron 2.4 (inductive probe om CAN toolhead board, 4 Z motors on mainboard, QGL and separate Z endstop) and it had been working flawlessly for the last 2 weeks.

**Comment by ameyp** [comment]
@zruncho3d I'm running the same setup too. A Huvud with X endstop wired to it, 4 Z motors and inductive probe wired to SKR.

**Comment by KevinOConnor** [comment]
FYI, I've pushed a notable portion of this branch into the master branch.  The low-level mcu code to perform multi-mcu homing is now in the master branch.  However, the actual multi-mcu homing commits have not been pushed yet.  There are still some outstanding issues and more testing needed on that part of the change.

This branch should also now have endstop_phases working with multi-mcu homing.  I am aware of one remaining issue with this code:
1. If one does multi-mcu homing on a multi-stepper axis, and the steppers on that axis are also on separate mcus, then this code will not handle it correctly. (After each home/probe attempt, the steppers may become out of sync.)

-Kevin

**Comment by bartlammers** [comment]
Small 'issue' I encountered on this branch, not sure if it's not on main but never noticed before this.

On my Voron V1 (corexy, a and b steppers on same board) when homing XY I get a tiny bit of overshoot it seems now and then. I had my x endstop position set to 310 and my maximum travel also to 310, then before homing z I get an out of bounds error because the x was at 310.1 (or something similar, can't remember exactly).

Remedied by setting my max travel a tiny bit higher

**Comment by henrikssn** [comment]
> Small 'issue' I encountered on this branch, not sure if it's not on main but never noticed before this.
> 
> On my Voron V1 (corexy, a and b steppers on same board) when homing XY I get a tiny bit of overshoot it seems now and then. I had my x endstop position set to 310 and my maximum travel also to 310, then before homing z I get an out of bounds error because the x was at 310.1 (or something similar, can't remember exactly).
> 
> Remedied by setting my max travel a tiny bit higher

I can confirm this, I added 0.5mm travel past the endstops which completely resolved the issue for me.

**Comment by mtw3d** [comment]
Hi Kevin, I am also getting the trsync_state error:

```
Recv: !! Unable to obtain 'trsync_state' response
Recv: !! Unable to obtain 'trsync_state' response
```

I got this once early on, but now I am getting it consistently.

If you need to see my config, it is available [here.](https://www.dropbox.com/s/nfukko3u2o9ot6t/printer.cfg?dl=0) The last thing I changed was the `stepper_y` and `steppery1` blocks, so if there is a config issue, that is likely where it is.

KevinOConnor:
Yes - overshoot is a known change introduced by this code.  My current thinking is to allow the overshoot as it is now, but audit the homing code to ensure it wont prevent homing of unrelated axes.

-Kevin

robthide37:
so if wanting to run a bl touch of the huvud boad the master can now do this or still need to switch branches ? i would like to start testign as i have two troodon/voron/railcore hybrids that are running a duet with a huvud on the extruder on one and the on a ,b, and extruder on the other . Plus i have a few people running my cnc gantry setup on their vorons or troodons and i had them convert to a toolboard already as well

zruncho3d:
Sharing my experiences... 

I have a currently-printing new printer, running 2x SKR Minis, running the latest code from this PR, with a Huvud toolhead.  More here (https://github.com/zruncho3d/f-zero) but from a control perspective, it might as well be a Voron 2.

I'm able to get 100% reliable printing with this setup, but only once the print has started.  I've been seeing two issues:

(1) When doing XY homing with the gantry at the far Y position, after the X home, the gantry moves diagonally in the +X,-Y direction.  Either most times or every time, this happens.  As I type this, I realize I should link to a short video, which I can do later if it would help.

(2) Occasionally I get timeout errors during X or Y homing.  That's a problem, because when it occurs, the motors shut off and the gantry potentially drops on the print with a hot hotend, plus, I have to re-run QGL.  I don't have the detailed logs from this, as I had thought these issues were due to a failing SKR (USB died, and then I switched to UART, but the problems remained).    

I guess I could revert to a known-good earlier version of this branch that I'm using on a different Voron 0 (with no issues for months), or I could accept it, and try to get some logs, and possibly do some quick prints to try to trigger.  Going to do the latter, but @KevinOConnor  - any thoughts/suggestions to track this down?

Again, ZERO issues during printing, but intermittent issues during homing only.

KevinOConnor:
Okay.  However, without a full log from the incident there's really not much I can do to help.

-Kevin

GerogeFu:
@KevinOConnor Hi, Kevin. I run into homing issue too. First i move the head to the middle of machine, then do QUAD_GANTRY_LEVEL. When finished, i do G28, then it tell me "!! Move out of range: 350.038 345.000 10.000 [0.000]". After that i try more commands like "G28 X" "G28 Y" "FIRMWARE_RESTART" "RESTART", finally it can home now. I give you my log, hope it will help you debug.
[klippy.log](https://github.com/KevinOConnor/klipper/files/6899334/klippy.log)

eddietheengineer:
@Tircown @KevinOConnor I'm running this branch on a printer with hybrid-corexy kinematics (IDEX) and two CAN Huvud toolhead boards. The X enstops are mounted to the toolheads and routed through the Huvud. 

When I try to home X, the first toolhead moves toward the left (X0) as expected, triggers the endstop, retracts, and triggers it again. However, when it triggers the second time, I get the following error:

> Internal error on command:"G28"
> Once the underlying issue is corrected, use the
> "FIRMWARE_RESTART" command to reset the firmware, reload the
> config, and restart the host software.
> Printer is shutdown

If I switch the first toolhead to a sensorless X home and leave the second toolhead with the toolhead MCU endstop, the first toolhead homes properly, but the second toolhead errors out in the same way as the first case (triggers the endstop once, retracts, and faults out the second time it homes). I tried slowing down the triggering and slowing down the retract speed (to add a bit more delay before it homes the second time), but it didn't change the case. 

I assume something is interacting between the IDEX X axis homing routine and the multi-mcu homing. For now I'll just run extra wires for the X endstops, but I wanted to document it for reference.

[klippy (40).log](https://github.com/KevinOConnor/klipper/files/6912202/klippy.40.log)

eddietheengineer:
@Tircown's pull request fixed the issue with idex_modes.py!

KevinOConnor:
@GerogeFu - sorry I missed your message earlier.  I'm not sure why that would be.  It is possible on this branch for the toolhead to nominally end outside the build area with this change.  But, I'm not sure what would lead to getting the "move out of range" error.  If you can recreate, can you issue an `M112` command immediately after the problem occurs, and then attach that log here?

Separately, it should be possible to avoid the issue by changing your start scripts to move to a valid position after a `G28`.  However, I'd still be interested in finding out what sequence of commands triggers that error.

-Kevin

KevinOConnor:
FYI, I plan to merge the next set of changes on this branch.  I plan to merge the "homing based on timing" part of this change in the next few days.  The current Klipper code calculates the home position based on where the steppers are located after homing - with this next merge Klipper will calculate the home position based on the time that the endstop switch triggers.

The above merge will not include multi-mcu support, but the remaining changes after that are small.

-Kevin

KevinOConnor:
FYI, I just merged the last of the changes on this branch in to the master branch (commit 95047784).

Of the original issues identified, the `If a multi-stepper axis (eg, stepper_z1) is used and those steppers are on different mcus then this code does not currently handle that accurately` was not addressed in code on this merge.  Instead, the code will raise an error if one attempts to configure multi-mcu-homing mixed with multi-mcu-stepper-axis.  Support for multi-mcu-stepper-axis and multi-mcu-homing can be added at a later date if there is interest in that.

-Kevin

brandonheller:
Thanks Kevin!  This is really exciting, as it means I can put both a Z probe (Klicky) and the X endstop togther on a toolhead board (Huvud), and not be managing off-master code.  4 toolhead wires go away, bringing the total from 9 down to 5 for me (CANH, CANL, 24V, 5V, GND).  

Nice.

And a Q re: the mutli-stepper axis limitation... I've never seen a >1-stepper printer board that doesn't have at least 4 drivers, and even with QGL or TBL you're covered.  

Is the current limitation you describe only going to affect those with individual-motor boards on each Z axis then?  Unless we live in a future with single-motor-integrated toolhead boards, I can't think of a use case.

robthide37:
The huvud .6 has 5v on it and support for up to 3 probes it can make you
drop to just 4

KevinOConnor:
Mixing multi-mcu-homing and multi-mcu-axis would be a rare case.  I don't know of anyone doing that today.

-Kevin

robthide37:
Question @KevinOConnor/klipper
***@***.***>  the
huvud v.61 has 3 endstop pins as pb10 pb11 pb12 and a 5v and grnd power but
the servo will not respond I assumed it was cause the no multi mcu but
updated and still nothing. So I tried defining as a servo pin and manually
commanding but got nothing as well. Is this cause of something with klipper
still or maybe the bl touch needs the pwm signal to be 5v for the servo pin
. I ordered a few 4 channel line level converters to try just encase

TP-SpeeDj:
[klippy.log](https://github.com/KevinOConnor/klipper/files/7087561/klippy.log)
Here's my log files, I hope they help.

PR #3957, opened 2021-02-18
Updated measure resonances doc on how to connect two ADXL345 to one RPi

gjnijenhuis:
Updated the docs on how to connect two ADXL345 to the same PI.

I have this setup on my printer with an RPi4 and is working flawlessly (no lost packets or other connection issues).

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -215,7 +215,7 @@ resonances of X axis with the accelerometer attached to the toolhead and the
 resonances of Y axis - to the bed (the usual bed slinger setup).
 
 However, you can also connect two accelerometers simultaneously, they can both
-be connected to the RPi as the RPi has support for 2 SPI slaves. In this 
+be connected to the RPi as the RPi has support for 2 SPI slaves. In this
 configuration they can't be used at the same time as they share the same SPI
 interface but as we are measuring either the hotend or the bed they will never
 be used at the same time.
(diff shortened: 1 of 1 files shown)


KevinOConnor:
The problem with the adxl345 is that if the CS pin is low, the chip itself treats the MISO/MOSI/CLK lines as an I2C bus.  In some situations, a command to another spi device could be seen as an i2c request and the adxl345 would respond in i2c mode - causing corruption of messages.  It's difficult to predict the messages that would cause this to occur, and difficult to predict the effects should it occur, so to keep things simple, we simply state that the adxl345 should not be on a shared SPI bus.

-Kevin

dmbutyugin:
It seems that many RPi with 40 pin header have 2 SPI buses, right?
https://www.raspberrypi.org/documentation/usage/gpio/

SPI

- SPI0: MOSI (GPIO10); MISO (GPIO9); SCLK (GPIO11); CE0 (GPIO8), CE1 (GPIO7)
- SPI1: MOSI (GPIO20); MISO (GPIO19); SCLK (GPIO21); CE0 (GPIO18); CE1 (GPIO17); CE2 (GPIO16)


Then it might be feasible to suggest to connect to SPI0 and SPI1 bus instead (e.g. spidev0.0 and spidev1.0). I would also still mention this possibility of connecting 2 adxl345 to MCU and RPi in the docs.

KevinOConnor:
I'm going to close this as it looks like the conversation has concluded.

-Kevin

PR #3940, opened 2021-02-15
Fixes clk Init for STM32F401 boards w odd crystals, eg. 25Mhz

pragun:
Previous clock initialization used 2Mhz as the PLL base freq.
This would mean that for an ODD crystal such as 25MHZ the
PLL_M divisor would be calculated as 12 (from 12.5).
This sets the PLL base freq to 1M allowing proper divison.
In addition fixes the following for STM32F401--
- Changes pll_freq to CONFIG_CLOCK_FREQ * 4
- Adds a 25Mhz crystal option to the KConfig
- Changes FREQ_PREIPH to CONFIG_CLOCK / 2

Why?
- STM32F401 is capped at 84Mhz and peripheral clock
can go upto 42Mhz
- There are many board designs out there that
(for some reason) use 25Mhz, this would avoid
undue frustration for folks who design their boards
based of those designs. 

Clock initialization for stm32f405 and stm32f407 (and other mcus) 
is not affected.  I don't want to touch something I have no way of testing.

PR #3961, opened 2021-02-19
Add stm32f429 clock initialization

ArkadiuszRaj:
Add missing clock initailziation for STM32F429.
Correct PLL setup for all clocks including 25MHz one.
Also F401 changes: 
Added P=4 PLL variable, 
Added APB1 divisor (for 84MHz it must be at least 2),
Improved  peripherial clock for F401 to 42Mhz.

Signed off by Arkadiusz Raj <arek.raj@gmail.com>


KevinOConnor:
Thanks.  At a high-level it looks fine.  However, it would help if this could be rebased to stand-alone commits (Kconfig fix, stm32f429 clock config, 25Mhz fixes, stm32f401 pll_freq updates, stm32f401 PERIPH_FREQ optimization).  This is a tricky part of the code and having the commits separate helps with "git bisect" regression finding.

Also, I'd prefer if we could add the checks using the C runtime style (eg, `if (CONFIG_MACH_STM32F401) ...`) instead of the C ifdef style.

This commit seems to have a lot of overlap with #3940 - @pragun - fyi.

-Kevin

ArkadiuszRaj:
Yes I know it overlaps.
I am not able to squish the code now. Alternatively can prepare new commit.

But now I think it would be best to split f401 from f4xx and have clean
code without runtime checks.

What now? Shall I rework this PR to have clean commits and you close 3940
or vice versa?


W dniu pt., 19.02.2021 o 16:27 KevinOConnor <notifications@github.com>
napisał(a):

ArkadiuszRaj:
Corrected code in https://github.com/KevinOConnor/klipper/pull/3963
This PR can be closed then.

PR #3964, opened 2021-02-19
module: Function_Macro, new function adds ability to execute macro fr…

xbladen:
…om functions by name

function macro is a printer-object that builds lists of macros together with their names from config.
theese macros can be called out of functions just by their names in just one line.
if no macro with this name is in the list, nothing is executed.
typically the functions defines the name of the macro.
for example the heaters.py and mcu.py are modified with this function call.

if you add a heater named extruder to configfile, the coeesponding function_macro gets the name [hester_name]_temp
in this case 'extruder_temp', for a heater named heater_bed the name of the macro would be 'heater_bed_temp'.
You can now define a section
[functionmacro extruder_temp]
macro:
...

the provided macro is executed everytime the extruder temperature is updated.
The macro can consist of a list of gcode commands. If you want to use scripting functionallity of gcode_macro,
just define a gcode_macro and call it in function_macro - means add it to the list of commands

mcu.py is modified to provide function_macro call with the extensions
'_home_start', '_home_finish', '_change', '_close', '_open' to endstops.

if you define a endstop for a stepper in stepper-section with the name P1.2 you can add now function_macro like
'P1.2_change', 'P1.2_close', 'P1.2_open' .....

function_macro is name based and benefits from other names defined in configfile. so the function_macro can match
to other elements used in configfile.
this is the reason, why the class Heater in heaters.py now gets the heater_name, could maybe solved in another way.

the use of macro in well known functions can have unwanted side-effects:
the macro execution consumes computing time and changes runtime behaviour

I even don't know if the function_maco itself could be written smarter - the jinja_stuff means nothing to me.
Please review.

Signed-off-by: Stefan Verse <verse@amotronics.de>

### DIFF
--- a/klippy/extras/function_macro.py
+++ b/klippy/extras/function_macro.py
@@ -3,50 +3,21 @@
 # Copyright (C) 2021  Stefan Verse <verse@amotronics.de>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import traceback, logging, ast, copy
-import jinja2
+import logging
 
-# Wrapper around a Jinja2 template
-class TemplateWrapper:
-    def __init__(self, printer, env, name, script):
-        self.printer = printer
-        self.name = name
-#        function_macro = self.printer.lookup_object('function_macro')
-        try:
-            self.template = env.from_string(script)
-        except Exception as e:
-            msg = "Error loading template '%s': %s" % (
-                 name, traceback.format_exception_only(type(e), e)[-1])
-            logging.exception(msg)
-            raise printer.config_error(msg)
-
-# Main function macro template tracking
 class PrinterFunctionMacro:
     def __init__(self, config):
         self.printer = config.get_printer()
-        self.gcode = self.printer.lookup_object('gcode')
-        self.env = jinja2.Environment('{%', '%}', '{', '}')
-        self.scripts = []
-        self.names = []
-    def load_template(self, config, option, default=None, function_name=None):
-        name = "%s:%s" % (config.get_name(), option)
+        self.macro = self.printer.load_object(config, 'macro_list')
+    def load_config(self, config, option, default=None, function_name=None):
         if default is None:
             script = config.get(option)
         else:
             script = config.get(option, default)
-        self.scripts.append(script)
-        self.names.append(function_name)
-        return TemplateWrapper(self.printer, self.env, name, script)
-    def run_macro_from_name(self, function_name):
-        try:
-            index=self.names.index(function_name.upper())
-            self.gcode.run_script_from_command(self.scripts[index].strip())
-        except:
-            return
+        self.macro.add_script(script, function_name)
 
 def load_config(config):
     return PrinterFunctionMacro(config)
-
 
 ######################################################################
 # FUNCTION_MACRO printer object
@@ -58,7 +29,8 @@ class FUNCTION_MACRO:
         self.alias = name.upper()
         self.printer = printer = config.get_printer()
         function_macro = printer.load_object(config, 'function_macro')
-        self.template = function_macro.load_template(config, 'macro',
-                                                     function_name=name.upper())
+        function_macro.load_config(config, 'macro', function_name=name.upper())
+
 def load_config_prefix(config):
     return FUNCTION_MACRO(config)
+
--- a/klippy/extras/function_macro.py
+++ b/klippy/extras/function_macro.py
@@ -1,0 +1,74 @@
+# Add ability to define custom macros mapped to functions
+#
+# Copyright (C) 2021  Stefan Verse <verse@amotronics.de>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import traceback, logging, ast, copy
+import jinja2
+
+# Wrapper around a Jinja2 template
+class TemplateWrapper:
+    def __init__(self, printer, env, name, script):
+        self.printer = printer
+        self.name = name
+#        function_macro = self.printer.lookup_object('function_macro')
+        try:
+            self.template = env.from_string(script)
+        except Exception as e:
+            msg = "Error loading template '%s': %s" % (
+                 name, traceback.format_exception_only(type(e), e)[-1])
+            logging.exception(msg)
+            raise printer.config_error(msg)
+
+# Main function macro template tracking
+class PrinterFunctionMacro:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.gcode = self.printer.lookup_object('gcode')
+        self.env = jinja2.Environment('{%', '%}', '{', '}')
+	self.scripts = []
+	self.names = []
+    def load_template(self, config, option, default=None, function_name=None):
+        name = "%s:%s" % (config.get_name(), option)
+        if default is None:
+            script = config.get(option)
+        else:
+            script = config.get(option, default)
+        self.scripts.append(script)
+        self.names.append(function_name)
+#        for name in self.names:
+#            msg = "Error PrinterFunctionMacro: loading template '%s':'%s'" %(name,script,)
+#            logging.exception(msg)
+        return TemplateWrapper(self.printer, self.env, name, script)
+    def run_macro_from_name(self, function_name):
+        try:
+	    index=self.names.index(function_name.upper())
+#            msg = "Error PrinterFunctionMacro: enter run_macro_from_name %d '%s'" %(index, self.scripts[index],)
+#            logging.exception(msg)
+#            commands = self.scripts[index].split('\n')
+#            for line in commands:
+#                msg = "Error PrinterFunctionMacro: enter run_script_split '%s'" %(line,)
+#                logging.exception(msg)
+#            self.gcode.run_script_from_command(str(self.scripts[index]))
+            self.gcode.run_script_from_command(self.scripts[index].strip())
+        except:
+	    return
+
+def load_config(config):
+    return PrinterFunctionMacro(config)
+
+
+######################################################################
+# FUNCTION_MACRO printer object
+######################################################################
+
+class FUNCTION_MACRO:
+    def __init__(self, config):
+        name = config.get_name().split()[1]
+        self.alias = name.upper()
+        self.printer = printer = config.get_printer()
+        function_macro = printer.load_object(config, 'function_macro')
+        self.template = function_macro.load_template(config, 'macro', function_name=name.upper())
+
+def load_config_prefix(config):
+    return FUNCTION_MACRO(config)
--- a/klippy/extras/function_macro.py
+++ b/klippy/extras/function_macro.py
@@ -33,4 +33,3 @@ class FUNCTION_MACRO:
 
 def load_config_prefix(config):
     return FUNCTION_MACRO(config)
-
--- a/klippy/extras/function_macro.py
+++ b/klippy/extras/function_macro.py
@@ -58,7 +58,7 @@ class FUNCTION_MACRO:
         self.alias = name.upper()
         self.printer = printer = config.get_printer()
         function_macro = printer.load_object(config, 'function_macro')
-        self.template = function_macro.load_template(config, 'macro', 
+        self.template = function_macro.load_template(config, 'macro',
                                                      function_name=name.upper())
 def load_config_prefix(config):
     return FUNCTION_MACRO(config)
--- a/klippy/extras/function_macro.py
+++ b/klippy/extras/function_macro.py
@@ -58,7 +58,7 @@ class FUNCTION_MACRO:
         self.alias = name.upper()
         self.printer = printer = config.get_printer()
         function_macro = printer.load_object(config, 'function_macro')
-        self.template = function_macro.load_template(config, 'macro', function_name=name.upper())
-
+        self.template = function_macro.load_template(config, 'macro', 
+                                                     function_name=name.upper())
 def load_config_prefix(config):
     return FUNCTION_MACRO(config)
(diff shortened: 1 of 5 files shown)


xbladen:
Sorry, the PID_stuff is included, too...

xbladen:
Some content of printer.cfg as example for tests:
```
[function_macro stepper_x_endstop_close]
macro:
    SET_LED LED=my_neopixel RED=1.0 GREEN=0.0 BLUE=0.0 TRANSMIT=1

[function_macro stepper_x_endstop_open]
macro:
    M105

[function_macro P1.29_change]
macro:
    SET_LED LED=my_neopixel RED=1.0 GREEN=0.0 BLUE=0.0 TRANSMIT=1

[function_macro P1.29_home_start]
macro:
    SET_LED LED=my_neopixel RED=0.0 GREEN=1.0 BLUE=0.0 TRANSMIT=11

[function_macro extruder_temp]
macro:
    M998

[gcode_macro M998]
gcode:  
   #if the extruder is off
   {% if printer.extruder.target == 0 %}
      #Set the LED to red if the extruder is off but is still hot, otherwise 
      # set the color to green
      {% if printer.extruder.temperature > 60.0 %}
         SET_LED LED=my_neopixel RED=1 GREEN=0 BLUE=0 TRANSMIT=1
      {% else %}
         SET_LED LED=my_neopixel RED=0 GREEN=1 BLUE=1 TRANSMIT=1
      {% endif %}     
   {% else %}
         #if the extruder temp is at target temperature 
         {% if printer.extruder.temperature >= printer.extruder.target - 4.0 %}
            SET_LED LED=my_neopixel RED=1 GREEN=0 BLUE=1 TRANSMIT=1
         #if the extruder is still heating
         {% else %}
            {% set scaler = printer.extruder.temperature|float / printer.extruder.target|float %}
            SET_LED LED=my_neopixel RED={ scaler|float * 1 } GREEN=0 BLUE=0 TRANSMIT=1
         {% endif %}  
   {% endif %}
```

xbladen:
macro_list.py could have been integrated in gcode.py, but the functionality gets more clear in separate files and printer_objects

Turge08:
Works great so far.

Btw, where can one get a list of the various events available for each object?

For example, what events would be available for bed_mesh?

xbladen:
Ok, i see what you mean.
Till now it was most important to me, that it works, but i have inserted the functions and know their names.
You are right for others, they need the list. So i will extend it by registering the possible functions to the component and try to make 2 commands (i have no idea, how such gcode_functions are registerd):
function_macro_active
function_macro_possible
you could call by terminal like gcode and get back the lists.
I think i change the naming in 'component'_'instatiotion'_'function' like :
heater_extruder_temp
endstop_p1.2_open
if you put a enstop in the printer.cfg (like P1.2), all possible function_macros are added to the function_macro_possible-list.
the name itself is sometimes not meaningful, because you can call a heater whatever you like in configfile.

xbladen:
Concerning bed_mesh - someone needs to identify interesting functions and put the calls to function_macro into the python-scripts. if you check the changes in heaters.py (first commit) the call is (actually it is 2 lines, because this one is too long) :
self.printer.lookup_object('function_macro').run_macro_from_name(self.heater_name+'_temp')

which just get the object 'function_macro' from pinter and calls it run_macro_from_name, if it was previously registered from  configfile. if not in the configfile, nothing is executed. The name of macro is a combination of instantiation_name and a choosen function_name
For heater.py the temperature-changes to react on were of interest, so i added the function_call to the callback-function, when a new temperature is send back from klipper. On every temperature message coming in, the macro is run.
Someone else would like to have a call when wait_for_temperature finishes, so there need the line to be inserted.
But i have no idea what the people want to do, and putting the call everywhere maybe is not a good idea.
So what would you like to do, when bed_mesh is going on ? Where do you want to execute an additional macro ?
And most important, is there an action to be identified to hook the call on.
For endstop, if there is a query for endstop, the value is compared to the last value to detect a change. This wasn't in before, so it was added to generate a trigger for the function_call. it only works, if klipper sends endstop_info and that is actually queried by hand and not polled like temperature.
You would need a function that periodically polls the endstop. I guess a macro could do that - i'm not shure.
For bed_mesh, i imagine there would be a lot of triggers like
- start
- finish
- probe down
- probe up
- level measured
- ....
I have not read the code for bed mesh. maybe you like to ?

xbladen:
Want i really mean - inserting this code in every file and every function means a lot of testing and i think Kevin needs to check for it or give his go for this. At least i'm new to klipper and don't know exactly, how the components act together and where it is maybe critical to change. I'm also new to python, the jinja-stuff means nothing to me and finding the meaning behind the scipts is not the easiest task.

xbladen:
Just a story to that, i had committed the code, that worked for me, and the automated checks threw errors.
What happened :
i had a function_macro defined in configfile, so a function_macro-object was added to printer, everything worked
the automated checks had no function_macro defined, so no function_macro-object was generated and the first call to it blew all up.
I needed to add early function_macro to klippy like gcode (how to do that ???)
I needed to import that file (why needs this to be in the klippy folder ???)
So i put the file in the klippy folder (why are no function_macro extracted from configfile)
Ah, need to split the file into two parts, macro_list to klippy and function_macro in extras folder

Turge08:
So I'm not really committed with triggering a macro from a bed mesh event. I was just doing some testing and assumed these events already existed.

For fun, I triggered the leds to change green whenever the BL Touch probe is extended but that was easy based on your examples above (PX.XX_change).

This could open up a wealth of opportunities and I guess it just needs to be clear to everyone what the limits are.

Anyways, great work! I just installed the neopixels this week and to already be able to control them based on heater temp is amazing.

xbladen:
That was the status, when i started this, have an artillary sidewinder pimped up and installed neopixels, but had limited use of them, because they were not updated when heating up.
What i'm missing now is an printer-idle_state, so that the neopixels are switched off when in idle - maybe set an idle-timeout in configfile ...

KevinOConnor:
Interesting.  Unfortunately, the implementation here is not going to work reliably.  (It will, unfortunately, lead to occasional deadlocks and mysterious crashes.)

At a high-level, a "gcode command" is designed to be an atomic action.  Run one command and then the next.  No two commands ever run at the same time.  This is an important abstraction, because if the user runs `G28` and then runs `G1 X100`, it's critical that the printer first home and then move to X100 - moving to X100 prior to or during the homing operation would be a spectacular failure.

If I understand correctly, you'd like to change the neopixel output during an `M109` command.  That's understandable, but it can not be done by executing "gcode" commands.  (No `SET_LED` commands; no gcode macros.)  As you've seen, Klipper can alter the neopixel output at any point, but it isn't valid to run the `SET_LED` command at any point.  Specifically, `M109` is defined to be a command that waits for a heater to reach a temperate and *does not allow any other command to run* while it waits.  So, for example, an `M109` followed by `SET_LED` must change the led after the `M109` completes.

The code on this branch runs the `gcode.run_script_from_command()` method, but runs it at points that are not "from a command".  That is incorrect and can cause internal corruption.

Of course, this raises the question - how can one change the neopixel output during an `M109`?  To accomplish that, it would be necessary to define an interface to the neopixel.py code that is not a gcode command (and thus can run in parallel with gcode commands).  What I'd look at is the https://www.klipper3d.org/API_Server.html and the associated code in klippy/webhooks.py .  We designed that interface to support asynchronous operations.  So, at a high-level it should be possible to use the API to "subscribe to heater temperature updates" and then execute "neopixel endpoints" as needed.  (There aren't any current "neopixel endpoints", but those could be added.)

I'm sure there are also other possible solutions outside of the API server.

Cheers,
-Kevin

xbladen:
Damn,
it works so well for neopixels...
That M109 waits and moves have to be executed after is meaningful, that neopixel cannot be updated is not.
As i could see wait is done in heater in high level, not in firmwware. So neopixel does not disturb M109, but as gcode commands can be anything, you are right.
So sad, but i will keep it for my printer setup for a while.

To hook up on endpoint actions was also really nice.
Can i use the existing neopixel send_data stuff for a new interface ? Just create a new entry point ?

Stefan

Von: KevinOConnor [notifications@github.com]
Gesendet: Sonntag, 21. Februar 2021 18:33
An: KevinOConnor/klipper
Cc: Verse, Stefan; Author
Betreff: Re: [KevinOConnor/klipper] module: Function_Macro, new function adds ability to execute macro fr… (#3964)


Interesting. Unfortunately, the implementation here is not going to work reliably. (It will, unfortunately, lead to occasional deadlocks and mysterious crashes.)

At a high-level, a "gcode command" is designed to be an atomic action. Run one command and then the next. No two commands ever run at the same time. This is an important abstraction, because if the user runs G28 and then runs G1 X100, it's critical that the printer first home and then move to X100 - moving to X100 prior to or during the homing operation would be a spectacular failure.

If I understand correctly, you'd like to change the neopixel output during an M109 command. That's understandable, but it can not be done by executing "gcode" commands. (No SET_LED commands; no gcode macros.) As you've seen, Klipper can alter the neopixel output at any point, but it isn't valid to run the SET_LED command at any point. Specifically, M109 is defined to be a command that waits for a heater to reach a temperate and does not allow any other command to run while it waits. So, for example, an M109 followed by SET_LED must change the led after the M109 completes.

The code on this branch runs the gcode.run_script_from_command() method, but runs it at points that are not "from a command". That is incorrect and can cause internal corruption.

Of course, this raises the question - how can one change the neopixel output during an M109? To accomplish that, it would be necessary to define an interface to the neopixel.py code that is not a gcode command (and thus can run in parallel with gcode commands). What I'd look at is the https://www.klipper3d.org/API_Server.html and the associated code in klippy/webhooks.py . We designed that interface to support asynchronous operations. So, at a high-level it should be possible to use the API to "subscribe to heater temperature updates" and then execute "neopixel endpoints" as needed. (There aren't any current "neopixel endpoints", but those could be added.)

I'm sure there are also other possible solutions outside of the API server.

Cheers,
-Kevin

KevinOConnor:
> Can i use the existing neopixel send_data stuff for a new interface ? Just create a new entry point ?

Yes.  Just be careful how you obtain the `print_time` (you can't call `toolhead.get_last_move_time()` without the gcode mutex).

-Kevin

xbladen:
Need some more help,
your suggestion of using API interface seems not to be the solution, that i’m looking for.
It would need to "subscribe to heater temperature updates", but also grap status of heater setting, name of extruder, name of heater, maybe printer status etc.
Executing a gcode_macro was the idea of using existing scripting possibilities and not to reinvent the wheel.
Using API means to embed the functionallity to Fluidd, octoklipper, Mainsail or write a sepparate application just for changing neopixel. (Handling connection problems, reconnect when klipper is restarted, find connction parameters, gui to set connection information, extend all the interfaces for getting status etc.)

It would be fine, if such neopixel or extra scripting support is included in klipper. Because everybody wants to express something different by neopixel, keeping the script-definition of neopixel-behaviour.
Defining the script in printer.cfg, the name of heater is given in the configfile, so defining script right is an easy task. Just add 10 Lines to configfile, and neopixel does what you want …

So what about this :
Copy the scripting-functionality from gcode_macro to function_macro, so it is independend of gcode_macro (that uses gcode.run_script_from_command()).
Resticted commandset ‚SET_LED‘ and ‚SET_PIN_LEVEL‘ (<- don’t know the function name) and use separat entry_points in neopixel and mcu.

Stefan

Von: KevinOConnor [mailto:notifications@github.com]
Gesendet: Sonntag, 21. Februar 2021 18:33
An: KevinOConnor/klipper
Cc: Verse, Stefan; Author
Betreff: Re: [KevinOConnor/klipper] module: Function_Macro, new function adds ability to execute macro fr… (#3964)


Interesting. Unfortunately, the implementation here is not going to work reliably. (It will, unfortunately, lead to occasional deadlocks and mysterious crashes.)

At a high-level, a "gcode command" is designed to be an atomic action. Run one command and then the next. No two commands ever run at the same time. This is an important abstraction, because if the user runs G28 and then runs G1 X100, it's critical that the printer first home and then move to X100 - moving to X100 prior to or during the homing operation would be a spectacular failure.

If I understand correctly, you'd like to change the neopixel output during an M109 command. That's understandable, but it can not be done by executing "gcode" commands. (No SET_LED commands; no gcode macros.) As you've seen, Klipper can alter the neopixel output at any point, but it isn't valid to run the SET_LED command at any point. Specifically, M109 is defined to be a command that waits for a heater to reach a temperate and does not allow any other command to run while it waits. So, for example, an M109 followed by SET_LED must change the led after the M109 completes.

The code on this branch runs the gcode.run_script_from_command() method, but runs it at points that are not "from a command". That is incorrect and can cause internal corruption.

Of course, this raises the question - how can one change the neopixel output during an M109? To accomplish that, it would be necessary to define an interface to the neopixel.py code that is not a gcode command (and thus can run in parallel with gcode commands). What I'd look at is the https://www.klipper3d.org/API_Server.html and the associated code in klippy/webhooks.py . We designed that interface to support asynchronous operations. So, at a high-level it should be possible to use the API to "subscribe to heater temperature updates" and then execute "neopixel endpoints" as needed. (There aren't any current "neopixel endpoints", but those could be added.)

I'm sure there are also other possible solutions outside of the API server.

Cheers,
-Kevin

KevinOConnor:
Understood.  Alas, I don't have much advice to offer.  Although the api_server is not a convenient option today, I think over the long term it's probably the best option for user-interface related functionality.  YMMV.

-Kevin

mental405:
Based on my experience with driving LED's via gcode macro, it really is not feasible to expect them to update fast enough to be pleasing to the eye. In some cases I have straight up killed Klipper with gcode macros. I have been working on getting some LED effects going in klipper that people can spin up without a lot of effort, but finding the time an willing ~~guinea pigs~~ volunteers has been difficult.

KevinOConnor:
Any further comments on this?

-Kevin

KevinOConnor:

-Kevin

PR #3979, opened 2021-02-22
Better support for emulated ST7920 displays

teeminus:
This PR addes a new ST7920E display type which is better suited for displays that emulate the ST7920 display driver. Those displays rely on using the CS (or more precisely the ST7920 enable pin).

The main problem (which leads to implementing the ST7920E driver) is, that for some mainboard/display combinations, the mainboard is not rebooted when klipper is restarted. Therefore the display is not restarted as well which leads to bit shifts in the SPI input buffer which then renders them inoperable until the SPI bus is reset. Usually, these displays use interrupts on the CS line to enable/disable the SPI peripherals when it's not needed which makes the display work again by cyling the CS pin.

To implement the improved CS handling for emulated ST7920s without changing the MCU firmware I created a new ST7920E display type which uses a software SPI and drives the CS pin directly. As the ST7920 has no MISO line, this signal has to be assigned to an unused pin on the mainboard.
The original ST7920 driver has been split into the DisplayBase which contains all frame buffers etc. and the "new" ST7920 driver which is a child class for the DisplayBase and contains the hardware driver. The new ST7920E device is based on the DisplayBase as well but uses a software spi as described above.
Therefore. no config changes are needed.

The required MISO pin for the software spi has been named "dummy_pin" in the config file. I don't know if this is a good choice, so this needs to be discussed.


teeminus:
I just thought about using the software-spi pin names for the ST7920E class.

KevinOConnor:
Interesting, thanks.  As a high-level comment, the functionality seems useful to me.  Some suggestions:
1) Name the feature `emulated_st7920` - I fear st7920e seems like a chip revision.
2) The docs/Config_Reference.md should be updated with this change.
3) I think it would be preferable for the new code to instantiate `bus. MCU_SPI` and `bus.MCU_bus_digital_out()` instead of manually looking up the low-level commands.
4) Just out of curiosity, is it necessary to toggle CS on every message instead of just during init?

-Kevin

Sineos:
I tried to test this PR but apparently, I'm setting the wrong configuration. This is what I have right now:

```
[board_pins]
aliases:
    # EXP1 header
    EXP1_1=P1.30, EXP1_3=P1.18, EXP1_5=P1.20, EXP1_7=P1.22, EXP1_9=<GND>,
    EXP1_2=P0.28, EXP1_4=P1.19, EXP1_6=P1.21, EXP1_8=P1.23, EXP1_10=<5V>,
    # EXP2 header
    EXP2_1=P0.17, EXP2_3=P3.26, EXP2_5=P3.25, EXP2_7=P1.31, EXP2_9=<GND>,
    EXP2_2=P0.15, EXP2_4=P0.16, EXP2_6=P0.18, EXP2_8=<RST>, EXP2_10=<NC>
    # Pins EXP2_1, EXP2_6, EXP2_2 are also MISO, MOSI, SCK of bus "ssp0"

[display]
lcd_type: emulated_st7920
spi_software_miso_pin: EXP2_1
spi_software_mosi_pin: EXP2_6
spi_software_sclk_pin: EXP1_5
en_pin: EXP1_3
#cs_pin: EXP1_4
#sclk_pin: EXP1_5
#sid_pin: EXP1_3
encoder_pins: ^EXP2_3, ^EXP2_5
click_pin: ^!EXP1_2
```

The commented pins in the display-section are the working values I'm using so far (prior to this PR). Would you please point into the right direction. Thanks.

teeminus:
You mixed up some pins. Check out this
st7920 pin <=> emulated_st7920 pin
cs_pin <=> en_pin
sclk_pin <=> spi_software_sclk_pin
sid_pin <=> spi_software_mosi_pin

So your config should look like this:
```
[display]
lcd_type: emulated_st7920
spi_software_miso_pin: ?????????
spi_software_mosi_pin: EXP1_3
spi_software_sclk_pin: EXP1_5
en_pin: EXP1_4
```

You need to pick the miso pin yourself as I don't know your setup.

Sineos:
Hardware: BT SKR 1.4T / BIGTREE_TFT35_E3_V3_0
Firmware TFT: https://github.com/teeminus/NoTouchScreenFirmware/commit/3297ceb947acd17b7a6f4bc8b553bf8865192411

```
[display]
lcd_type: emulated_st7920
spi_software_miso_pin: EXP2_1
spi_software_mosi_pin: EXP1_3
spi_software_sclk_pin: EXP1_5
en_pin: EXP1_4
encoder_pins: ^EXP2_3, ^EXP2_5
click_pin: ^!EXP1_2
```

I can confirm it works:
- `service klipper restart` works
- `RESTART` works
- `FIRMWARE_RESTART` works
- Power-cycling the board followed by `FIRMWARE_RESTART` works

Great job. Many thanks

teeminus:
@Sineos Thanks (again) for testing the PR 😄 I made a small update to the code so that the enable pin is only toggled once upon restart of the firmware.

@KevinOConnor Thanks for your very useful suggestions. I implemented all changes. Regarding your question: It's not required to toggle the CS pin on each message but I thought it would make the implementation a bit more robust. But after testing the changes of the latest commit (a653fa1) I came to the conclusion that toggling the enable pin once when the driver is loaded is sufficient.
Btw.: I have not added myself to the copyright header as I consider this a pretty small change which is only useful for a pretty small group of people.

Sineos:
> I made a small update to the code so that the enable pin is only toggled once upon restart of the firmware.

Verified working 👍

KevinOConnor:
Thanks.  Looks fine to me.  I'll give it a few more days to see if anyone else has comments.

Separately, in the future, we could possibly extend the micro-controller SPI code to support an inverted CS pin.  We could also possibly introduce a special "NOT_A_PIN" pin to the mcu that could be used in places where we don't actually want to drive a pin (eg, for the miso pin here).

-Kevin

teeminus:
Having that NOT_A_PIN feature would be great. The inverted CA feature would be nice as well as it would reduce the size of the python code at the cost of increasing the complexity of the MCU code.

An other interesting thought that came into my mind: By either reducing the SPI speed or introducing appropriate delays between each pair of two bytes this code should work for real ST7920 devices as well (thus making the MCU ST7920 driver obsolete).

KevinOConnor:
Thanks.

-Kevin

rs-development:
Should this also work for TFT24 V1.1?

[
photo_2021-03-07_00-25-23
](url)

teeminus:
These screen artifacts are caused by the incomplete ST7920 implementation of the BTT firmware. If you just use the ST7920 mode on your display (and none of the touch features) you can try this firmware: https://github.com/teeminus/NoTouchScreenFirmware

KevinOConnor:
FWIW, if someone wanted to implement it, the `emulated_st7920` Klipper mode could use only graphics updates (not use the st7920 character generator), and thus improve compatibility with the base firmware.

-Kevin

JoeK1973:
Hi - I've got the TFT35 E3 3.0 - how do I get this working? I've got the SKR Mini E3 v1.2 by the way. I've added the text above to my printer.cfg but am getting an error that 
`Reported error: pin EXP1_4 is reserved for <RST>`

Any hints warmly appreciated!

Sineos:
See https://github.com/teeminus/NoTouchScreenFirmware/wiki/Klipper-display-configuration or open a discussion there for additional support

PR #3857, opened 2021-01-29
Add filament_motion_sensor

TheJoshW:
Add functionality to support a Filament Motion Sensor for detecting extruder jams as well as runouts.  Works by an encoder toggling the switch_pin 0/1 as the filament is pulled through the sensor.

Everything seems to work ok, except I have seen some weirdness in pause_resume.  When pausing, the printer runs the runout_gcode however it sometimes moves back over the print job (does not affect Z, only X (and maybe Y)).  Increasing the pause_delay may resolve this.

Also pause_resume seemed a little inconsistent in how it was handled so I tried to address, not sure if this is ideal but it should resolve most issues and get people going using this plugin.  I have documented this in detail in FilamentMotionSensor.md.

Resolves #1327
Resolves #886
Resolves #249
Resolves #3715

Signed-off-by: Joshua Wherrett <thejoshw.code@gmail.com>

PR #3982, opened 2021-02-23
Add support for host tracking of the history of stepper motion

KevinOConnor:
This extends the low-level step compression code to keep a history of `queue_step` commands sent to the micro-controller.  The host can use this information to calculate the position of the stepper motor given a recent time.

This will be useful for filament sensors, as it should allow the host code to obtain the extruder position at the time of a signal from the filament sensor.

This support will also be needed for multi-mcu homing.

@Arksine , @TheJoshW, @mattshepcar - FYI.

-Kevin

### DIFF
--- a/klippy/stepper.py
+++ b/klippy/stepper.py
@@ -124,6 +124,11 @@ class MCU_stepper:
         return self._tag_position
     def set_tag_position(self, position):
         self._tag_position = position
+    def get_past_commanded_position(self, clock):
+        ffi_main, ffi_lib = chelper.get_ffi()
+        sq = self._stepqueue
+        mcu_pos = ffi_lib.stepcompress_find_past_position(sq, clock)
+        return mcu_pos * self._step_dist - self._mcu_position_offset
     def set_stepper_kinematics(self, sk):
         old_sk = self._stepper_kinematics
         self._stepper_kinematics = sk
@@ -145,7 +150,11 @@ class MCU_stepper:
         if not did_trigger or self._mcu.is_fileoutput():
             return
         params = self._get_position_cmd.send([self._oid])
-        mcu_pos_dist = params['pos'] * self._step_dist
+        last_pos = params['pos']
+        ret = ffi_lib.stepcompress_set_last_position(self._stepqueue, last_pos)
+        if ret:
+            raise error("Internal error in stepcompress")
+        mcu_pos_dist = last_pos * self._step_dist
         if self._invert_dir:
             mcu_pos_dist = -mcu_pos_dist
         self._mcu_position_offset = mcu_pos_dist - self.get_commanded_position()
--- a/klippy/chelper/__init__.py
+++ b/klippy/chelper/__init__.py
@@ -35,6 +35,10 @@ defs_stepcompress = """
         , int32_t set_next_step_dir_msgtag);
     void stepcompress_free(struct stepcompress *sc);
     int stepcompress_reset(struct stepcompress *sc, uint64_t last_step_clock);
+    int stepcompress_set_last_position(struct stepcompress *sc
+        , int64_t last_position);
+    int64_t stepcompress_find_past_position(struct stepcompress *sc
+        , uint64_t clock);
     int stepcompress_queue_msg(struct stepcompress *sc
         , uint32_t *data, int len);
 
--- a/klippy/chelper/stepcompress.c
+++ b/klippy/chelper/stepcompress.c
@@ -14,6 +14,7 @@
 // This code is written in C (instead of python) for processing
 // efficiency - the repetitive integer math is vastly faster in C.
 
+#include <math.h> // sqrt
 #include <stddef.h> // offsetof
 #include <stdint.h> // uint32_t
 #include <stdio.h> // fprintf
@@ -42,6 +43,25 @@ struct stepcompress {
     // Step+dir+step filter
     uint64_t next_step_clock;
     int next_step_dir;
+    // History tracking
+    int64_t last_position;
+    struct list_head history_list;
+};
+
+struct step_move {
+    uint32_t interval;
+    uint16_t count;
+    int16_t add;
+};
+
+#define HISTORY_EXPIRE (30.0)
+
+struct history_move {
+    struct list_node node;
+    uint64_t first_clock, last_clock;
+    int64_t start_position;
+    int step_count;
+    struct step_move sm;
 };
 
 
@@ -83,12 +103,6 @@ minmax_point(struct stepcompress *sc, uint32_t *pos)
 // maxerror / (count*count)".  The "6 + 4*sqrt(2)" is 11.65685, but
 // using 11 works well in practice.
 #define QUADRATIC_DEV 11
-
-struct step_move {
-    uint32_t interval;
-    uint16_t count;
-    int16_t add;
-};
 
 // Find a 'step_move' that covers a series of step times
 static struct step_move
@@ -237,6 +251,7 @@ stepcompress_alloc(uint32_t oid)
     struct stepcompress *sc = malloc(sizeof(*sc));
     memset(sc, 0, sizeof(*sc));
     list_init(&sc->msg_queue);
+    list_init(&sc->history_list);
     sc->oid = oid;
     sc->sdir = -1;
     return sc;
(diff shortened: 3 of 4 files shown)


TheJoshW:
Thanks @KevinOConnor I'll refactor my code to use this. My code also uses a look ahead to get the time when the extruder will extrude a particular length of filament in order to trigger a runout if the runout sensor does not trigger before this time. How can I do this with your changes? Thanks.

KevinOConnor:
> My code also uses a look ahead to get the time when the extruder will extrude a particular length of filament in order to trigger a runout if the runout sensor does not trigger before this time. How can I do this with your changes?

I'd probably use the reactor to schedule a recurring timer every 250ms or so - it can check if everything looks okay or if the filament appears stuck.

-Kevin

Arksine:
Nice, I'll see what Josh comes up with then dust off the pat9125 code.  Since they work under similar principle of "motion detection" I may be able to reuse some code.  

It seems that this could also be useful for the display and API server clients.  Displaying the location at the current event time vs the gcode or toolhead position would probably be more informative to users.  I'm not certain which "get_status()" method this would be appropriate in.  Perhaps the toolhead since it would be ideal to display X,Y,Z, and E?

KevinOConnor:
> I'm not certain which "get_status()" method this would be appropriate in. Perhaps the toolhead since it would be ideal to display X,Y,Z, and E?

So, at this time, the `find_past_position()` isn't particularly fast.  It needs to walk the list of all the past queue_step commands on the stepper.  In addition, calculating the XYZ position would require performing the "forward kinematics" math which can be expensive on some kinematics. It's certainly fine to do all that 10 or so times a second, but I'm not sure I'd want to do that work in a `get_status()` method.  The fear would be that a user looking for some other info would incur all that cpu processing just to discard the results of the calculation.

Not sure what the best solution is though.  It does seem like a useful feature.  Just "thinking out loud", I guess we could build a module that caches recent positions.  Not sure.

-Kevin

KevinOConnor:
Again, "thinking out loud", we could also keep a history of recent "trapq" items.  That might be a bit easier for reporting real-time location, because it would not require calculating the "forward kinematics".  I'd still be a little leery of walking the trapq list from `get_status()` though.

-Kevin

TheJoshW:
My current code basically keeps a history of recent trapq items, but does it in filament_motion_sensor.py, in order to minimise changes to extruder.py.  This was not acceptable due to the high frequency use of printer.send_event().  However I can move this processing to extruder.py and use methods such as position_at_print_time(print_time), print_time_at_position(pos) and callback_at_position(position, callback).  Would that be quicker/more efficient?  callback_at_position() would be handy for me so I can trigger a runout when I reach that position.

KevinOConnor:
> However I can move this processing to extruder.py and use methods such as position_at_print_time(print_time), print_time_at_position(pos) and callback_at_position(position, callback).

We can add additional helper functions to extruder.py, but I'd be reluctant to add more logic to `check_move()`, `move()`, or `calc_junction()`.  Those functions are the "fast path" - they could potentially be called a 1000+ times per second.  In contrast, I'd think we really wouldn't want to check for filament issues more than 5-10 times per second.

Perhaps it would help if you could explain the high-level logic of the filament sensor checking?  I'd of thought one could run a timer every 250ms, have it check if the "current extruder position" minus "position of last sensor signal" is greater than "distance of between sensor signals" and run the error handling in that case.  Maybe I've missed something though.

-Kevin

TheJoshW:
Thanks @KevinOConnor that makes sense, see high level runout logic below:

Initially a filament_runout_pos is defined as the extruder position to trigger a runout at.
On extruder move (in extruder.py), trapq data is appended to the _extruder_move_list (in filament_motion_sensor.py).
This includes; print time at start and end of move, extruder pos at start and end of move and move cruise velocity (accel/decel on extruder is assumed to be insignificant and ignored for this purpose).
Initially _extruder_move_list range may not include filament_runout_pos, in this case, the list is checked and if it covers filament_runout_pos, the time that filament_runout_pos will be reached is approximated using t=(((filament_runout_pos - start_pos) / cruise_vel) + start_time) and a timer is set to trigger at this time.
If the filament_motion_sensor encoder wheel doesn't trigger on filament movement within that time, then the runout timer triggers and the print is paused, etc.

The filament_motion_sensor code is entirely event driven with no timers checking the extruder position continuously to see if it exceeds the filament_runout_pos.
However this may be innefficient if the extruder is processing more than say 4 moves per second, and as you mentioned this could be 1000+ times a second.

I could use a 250ms timer, that is disabled when the printer is idle or paused.

WindoC:
how to use the function find_past_position()? what is the "print_time" have to input? is it the same thing as "eventtime" in the timer?

So I can get the real E position in the below timer handler function. Right?

```
def _timer_handler(self, eventtime):
  theposition = self.extruder.find_past_position(eventtime)
```

KevinOConnor:
> how to use the function find_past_position()? what is the "print_time" have to input? is it the same thing as "eventtime" in the timer?

No, an eventtime is not a printtime - there is a brief description of times at: https://www.klipper3d.org/Code_Overview.html#time

If you need to go from the current system time to a print time, then it would look something like: `printtime = mcu.estimated_print_time(eventtime)`

-Kevin

klippy/stepper.py
```diff
@@ -124,6 +124,14 @@ def get_tag_position(self):
         return self._tag_position
     def set_tag_position(self, position):
         self._tag_position = position
+    def get_past_commanded_position(self, clock):
+        ffi_main, ffi_lib = chelper.get_ffi()
+        sq = self._stepqueue
+        mcu_pos = ffi_lib.stepcompress_find_past_position(sq, clock)
+        mcu_pos_dist = mcu_pos * self._step_dist
+        if self._invert_dir:
```
TheJoshW, line 132: Hi @KevinOConnor, I am testing this on my fork now and am getting an increase in the negative direction with positive extrusion.  I have self._invert_dir = True in my printer.cfg, so maybe inversion is already taken care of in stepcompress_find_past_position. Please confirm, thanks.

TheJoshW:
I made a comment on stepper.py and removed the dir inversion logic in my fork, see PR #3857.
However further testing revealed that the extrude.find_past_position() returns increasing, then decreasing, then increasing values, when it should only return increasing values, so something else is not working.  See log attached.
[klippy_inconsistent_extruder_pos.log](https://github.com/KevinOConnor/klipper/files/6096466/klippy_inconsistent_extruder_pos.log)

TheJoshW:
Hi @KevinOConnor,
I made a comment earlier but can't see where it went.  I merged this into my fork (see PR #3857), and deleted the invert_dir checks in stepper.py first.  I also instrumented `filament_motion_sensor.py` (attached) to get the log output (also attached).  Looks like extruder.find_past_position() is reporting increasing values as expected, then starts to report decreasing values as time continuously increments, I used your recommended technique to get extruder position:
```
print_time = self.mcu.estimated_print_time(eventtime)
extruder_pos = self.extruder.find_past_position(print_time)
```
So maybe the invert_dir logic in stepper.py is fine and the issue is possibly in `stepcompress_find_past_position()`

[klippy_inconsistent_extruder_pos.log](https://github.com/KevinOConnor/klipper/files/6096704/klippy_inconsistent_extruder_pos.log)

[filament_motion_sensor.py.txt](https://github.com/KevinOConnor/klipper/files/6096705/filament_motion_sensor.py.txt)

KevinOConnor:
Thanks.  There were indeed a couple of errors in the handling of stepper direction.  I've made some fixes and rebased this branch.  Hopefully the errors are fixed now.

-Kevin

TheJoshW:
Thanks, that fixed it.  This change appears to work for me now and I have incorporated it into my PR #3857.

WindoC:
I can call the function with print_time = self.mcu.estimated_print_time(eventtime) from the eventtime from the timer but I would like to call it in (register_event_handler) "klippy:ready" or (register_mux_command) GCODE command. I try to get the printtime by "self.mcu.estimated_print_time(self.mcu.estimated_print_time(self.reactor.monotonic()))" but it seem always get error "can't convert negative number to unsigned".

example:
```
theout = self.extruder.find_past_position(self.mcu.estimated_print_time(self.mcu.estimated_print_time(self.reactor.monotonic())))
OverflowError: can't convert negative number to unsigned
```
my code: https://github.com/WindoC/klipper/blob/WindoC-patch-2/klippy/extras/filament_jam_sensor.py

```
  File "/home/pi/klipper/klippy/gcode.py", line 182, in _process_commands
    handler(gcmd)
  File "/home/pi/klipper/klippy/gcode.py", line 120, in <lambda>
  File "/home/pi/klipper/klippy/gcode.py", line 286, in _cmd_mux
    values[key_param](gcmd)
  File "/home/pi/klipper/klippy/extras/filament_jam_sensor.py", line 114, in cmd_SET_FILAMENT_JAM_SENSOR
    self.timer_usage_last = self.filament_usage_last = self.get_filament_usage(self.mcu.estimated_print_time(self.reactor.monotonic()))
  File "/home/pi/klipper/klippy/extras/filament_jam_sensor.py", line 96, in get_filament_usage
    theout = self.extruder.find_past_position(self.mcu.estimated_print_time(eventtime))
  File "/home/pi/klipper/klippy/kinematics/extruder.py", line 158, in find_past_position
    return self.stepper.get_past_commanded_position(clock)
  File "/home/pi/klipper/klippy/stepper.py", line 130, in get_past_commanded_position
    mcu_pos = ffi_lib.stepcompress_find_past_position(sq, clock)
OverflowError: can't convert negative number to unsigned
```

TheJoshW:
This works:
```
print_time = self.mcu.estimated_print_time(eventtime)
theout = self.extruder.find_past_position(print_time)
```

PR #3994, opened 2021-02-26
virtual_sdcard: exact filename match before case insensitive one

andrepl:
currently, if there are 2 files on the virtual sd card whose names differ only in
case (eg. MyFile.gcode vs myfile.gcode) the actual file that gets loaded is at
best unpredictable.  this patch checks for an exact match before attempting a
case-insensitive one.

Signed-off-by: Andre LeBlanc <andrepleblanc@gmail.com>


KevinOConnor:
Thanks.

-Kevin

Issue #133, opened 2018-01-21
SET_PIN not usable for laser engraving


GitHub1712:
Using SET_PIN for laser raster engraving is very slow.
Same code runs much faster on marlin using m42 instead of SET_PIN.
Attached gcode to test.
Removing m400 will make execution asynchronous.

[klippy.log](https://github.com/KevinOConnor/klipper/files/1649539/klippy.log)

[KlipperRasterTest.zip](https://github.com/KevinOConnor/klipper/files/1649517/KlipperRasterTest.zip)

KevinOConnor:
On Sun, Jan 21, 2018 at 04:03:58AM +0000, GitHub1712 wrote:
> Using SET_PIN for laser raster engraving is very slow.
> Same code runs much faster on marlin using m42 instead of SET_PIN.
> Attached gcode to test.
> Removing m400 will make execution asynchronous.

It will be very slow if there are a lot of M400 commands in the file.
The M400 command instructs Klipper to completely wait for all activity
on the printer to stop.

If you're seeing an issue without the M400 commands, can you post the
log of a session without the M400 commands (but with the M112)?

-Kevin

KevinOConnor:
It's worth noting that the SET_PIN command was not intended as a mechanism for high speed updates.  A SET_PIN command has a minimum time of 100ms - if multiple commands are sent within a smaller window, then the subsequent commands will be delayed so that each pin setting is active for at least 100ms.

This behavior is due to how commands are buffered between host and micro-controller.  Things like step times are buffered on the micro-controller, and hundreds of thousands of them can be queued.  However, pin updates have a very small queue on the micro-controller (no more than 1 update) and instead most of the queuing is done on the host.  So, there would need to be development work necessary to make high speed arbitrary pin updates with precise timing work.

-Kevin

GitHub1712:
thank you for checking,
so thats a point better working with marlin. Many are attaching laser on 3d
printer. I think set_pin is something that should work fast.
I will try to get it working with extruder commands to controll the laser.

2018-01-21 6:57 GMT+01:00 KevinOConnor <notifications@github.com>:

KevinOConnor:
In issue #152 Lucas wrote:
> As the title says, the printer stops with a 'timer too close' error.
> 
> Now, before anything else, I should mention 2 things:
> 1. It's not 3d printing. It's laser engraving. I've attached a laser driver to the fan output and control the laser with M106 and M107 (only on and off, no power control)
[...]

PR #3995, opened 2021-02-26
Add 'tool' module 

Cirromulus:
This PR adds a extras/tool module to control PWM-tools like lasers or spindles via `M3-M5`.
Also adds a feature of a safety timer and a default shutdown value to increase safety when handling with lasers.

As long as longer commands are not split into smaller segments, the "keep going"-message is not sent in time.
This is because `register_lookahead_callback` only appends the function to the last move.
I did not find a more suitable way to insert this _resend callback_.

The timeout is still usable, but needs to be configured to always be longer than the longest running command.
This shows especially with arc moves.

With adding `M3-M5`, this also increases the compatibility with existing CAM-tools.

This adds a step towards #133. The only missing link left would be a fast path for PWM-Messages, e.g. over the [stepcompress system](https://github.com/Cirromulus/klipper/tree/myfork/queue_pwm).


KevinOConnor:
Interesting.  As high-level feedback, I'm not really sure what this module does that's different from the output_pin.py module.  Is there something that is done here, that couldn't be done with macros and SET_PIN?

If it's the max_duration stuff, could we add that to output_pin instead?  Also, we should be able to implement the max_duration checks using reactor callbacks so that the user does not need to worry about limiting command duration.

Also, I think the module name "tool" is too generic.  If we do need a new module I think we should come up with a more specific name.

Thanks,
-Kevin

Cirromulus:
>If it's the max_duration stuff, could we add that to output_pin instead? 

I did not want to touch existing codebase, but this is probably the most "klipper way", so I agree.

>Also, we should be able to implement the max_duration checks using reactor callbacks

Ok, then I need to dig deeper into the reactor callbacks. Until now, I only found a way to add a command to the queue, effectively waiting for longer commands.

>If we do need a new module I think we should come up with a more specific name.

As you said, it is probably best SET_PIN gets the max_duration and shutdown_value, while implementing G3-5 simply in Macros.
Are one-liners anyway.

Cirromulus:
> we should be able to implement the max_duration checks using reactor callbacks so that the user does not need to worry about limiting command duration.

This is now implemented. The callback now calculates the exact needed time derived from `last_print_time`.
Updates are not sent during homing (drip mode) however. This should be OK, because there is no sense in having a laser/drill/... running during homing.

Cirromulus:
Squashed commits.

lenne0815:
Cirro you are an absolute beast, thank you so much for your work ! Could you tell me how it roughly performs ? I can do 200mm/s at 300 DPI with batched gcode via lightburn / cohesion, i guess this outperforms it by a rather large margin ?

Cirromulus:
I did not test it to the limit.
I am currently lasering at .1mm resolution (~250DPI) and 30-40 mm/s.
This is mainly because I have a weak laser, but also the missing _proportionality_ is starting to show up on high speeds. Makes up for nice effects sometimes, though.
The top switching speed will be board-dependent. I have only an ATMega2560  and my gantry shakes at 100mm/s already*, so I wont go higher. Laser switching was accurate, though.

*: I use my printer with a custom head-swapping mechanism. So it will be primarily a 3D-Printer that also can laser engrave, but not the other way around.

lenne0815:
Thanks for the heads up ! My pure co2 laser can easily sustain 2k mms so im in for some speed testing :D Im gonna test it as soon as everything is merged and my pi arrived ! I believe this is a rather big deal for co2 machines because until now everyone was capped around 200 mms with the fastest non dsp controllers, dsp controllers only start from ~500 euros onward.

lenne0815:
Just one more thought; for slower lasers it would be important to scale pwm according to the actual speed of the machine, could you implement that as well ?

Cirromulus:
> Just one more thought; for slower lasers it would be important to scale pwm according to the actual speed of the machine, could you implement that as well ?

We'll see that when this PR is merged.
The Roadmap for that is documented in https://github.com/KevinOConnor/klipper/pull/3338#issuecomment-719760533
So right now, only my fork (not in this PR) is capable of raster engraving. 
This needs a revamp which @KevinOConnor might do, but probably only after this PR .

klippy/extras/output_pin.py
```diff
@@ -66,5 +81,15 @@ def cmd_SET_PIN(self, gcmd):
         toolhead.register_lookahead_callback(
             lambda print_time: self._set_pin(print_time, value, cycle_time))
 
+    def _resend_current_val(self, eventtime):
+        toolhead = self.printer.lookup_object('toolhead')
+        toolhead.register_lookahead_callback(lambda print_time:
+            self._set_pin(self.last_print_time + 0.8 * self.safety_timeout,
+                           self.last_value, self.last_cycle_time, True)
+            )
```
KevinOConnor, line 89: Thanks.  Unfortunately, it's not valid to call `toolhead.register_lookahead_callback()` without holding the gcode mutex.  That is, it can't be called directly from a reactor callback.  You probably want to do something like `print_time = self.mcu_pwm.get_mcu().estimated_print_time(eventtime)` instead - though the time will need to be padded by 100ms (to allow for transmission time to the mcu).

-Kevin

Cirromulus: Is it allowed to just set the pin directly, as it is now?
100ms time padding is now explicitly in the "relax_margin" which is supposed to give the scheduler some time.

KevinOConnor: The `toolhead.get_last_move_time()` and related `toolhead.register_lookahead_callback()` always return a time in the future, so that it is possible to schedule events at that time.  A call to `mcu.estimated_print_time(systime)` returns the time of "now" on the micro-controller - if one were to schedule an event at that time, it would always result in a "timer too close" as it takes time to transmit that schedule to the micro-controller.

I'm not sure what the goal of the "relax_margin" is.  You only need the PIN_MIN_TIME - any additional time is probably not a good thing as it limits how response future updates are.

-Kevin

Cirromulus: the "relax margin" was supposed to reduce stress between "now" and the scheduled time the pin has to update (and to reduce the possibility of too close timers).
But if you say that this just removes responsiveness, it can be replaced for just the PIN_MIN_TIME, instead of 10% timeout + PIN_MIN_TIME.

KevinOConnor:
As an additional minor comment, could we change the "safety_timeout" parameter - maybe something like "host_verification_time"?  I'm a little leery of implying that the setting will somehow make something "safe".  Also, the parameter should have a reasonable minval - maybe .500 .

-Kevin

Cirromulus:
That is no problem.
In the newest commit, it may be either zero or at least `1.25*PIN_MIN_TIME`, so that the scheduling later (`0.8 * safety_timeout - PIN_MIN_TIME`) does not evaluate to the past.
I agree that reasonably values under 0.5s create unnecessary effort on the machine, so maybe the lower limit is just 0.5s .

Cirromulus:
I suggest `host_acknowledge_timeout`, because verification sounds a bit ambiguous (to me, at least).

klippy/extras/output_pin.py
```diff
@@ -72,7 +72,8 @@ def _set_pin(self, print_time, value, cycle_time, is_resend=False):
                     self.resend_timer, self.reactor.NEVER)
             else:
                 self.reactor.update_timer(
-                    self.resend_timer, self.reactor.monotonic())
+                    self.resend_timer, self.reactor.monotonic()
```
Cirromulus, line 75: I am not sure about this one.
self.reactor.monotonic() returns the "current" systime.
But the `print_time` may be somewhere in the future (around the lookahead time).
Is it here better to instead get the estimated systime at the print_time @KevinOConnor?
e.g.:
```python
eventtime = self.mcu_pin.get_mcu()._clocksync.estimate_clock_systime(print_time)
self.reactor.update_timer(
    self.resend_timer, eventtime + (0.8 * self.host_ack_timeout) - PIN_MIN_TIME
)
```

KevinOConnor: Might be simpler to do all the scheduling work in the timer.  For example, just have something like the following in `_set_pin()`:
```python
if self.host_ack_timeout and self.resend_timer is not None:
    self.resend_timer = self.reactor.add_timer(self._resend, self.reactor.NOW)
```

And then do something like the following in `_resend()`:
```python
def _resend(self, eventtime):
    if self.last_value == self.shutdown_value:
        self.reactor.unregister_timer(self.resend_timer)
        self.resend_timer = None
        return self.reactor.NEVER
    systime = self.reactor.monotonic()
    print_time = self.mcu_pin.get_mcu().estimated_print_time(systime)
    time_diff = print_time - (self.last_print_time + .8 * self.host_ack_timeout - PIN_MIN_TIME)
    if time_diff > 0.:
        # Reschedule for resend time
        return systime + time_diff
    self._set_pin(print_time + PIN_MIN_TIME, self.last_value, self.last_cycle_time, True)
    return systime + (0.8 * self.host_ack_timeout) - PIN_MIN_TIME
```

Cirromulus: I can try this code tomorrow.
My first thought is, that when a lot of pin updates come (With my "slow" laser setup at ~500 updates/s), will there be 500 spawned timers concurrently rescheduling themselves?

Currently, the timer keeps on being rescheduled upon these updates, without being fired in a dense update phase.

KevinOConnor: This demo code only adds the timer if no timer is already running.  It is optimized for just the scenario you mentioned - not only will it not add a timer - it doesn't even need to reschedule the timer on each update.  (Instead it "lazily reschedules" when the timer triggers.)

FYI, it's just demo code - I haven't tested it at all.

-Kevin

Cirromulus:
@KevinOConnor 
Do you have any more comments or requests?

docs/Config_Reference.md
```diff
@@ -2484,6 +2484,13 @@ pin:
 #shutdown_value:
 #   The value to set the pin to on an MCU shutdown event. The default
 #   is 0 (for low voltage).
+#host_acknowledge_timeout:
```
KevinOConnor, line 2487: Would something like `maximum_mcu_duration` be better?  It's not the host that has the timeout, but the mcu.

docs/Using_PWM_Tools.md
```diff
@@ -0,0 +1,73 @@
+This document describes how to setup a PWM-Controlled laser or spindle
+using `output_pin` and some macros.
```
KevinOConnor, line 2: Generally new documents should also be added to docs/Overview.md (otherwise, no one is likely to find it).  Though I wonder if adding this info to something like config/sample-laser.cfg might be better?

KevinOConnor (commented):
Thanks.  In general it looks fine to me, but I did notice some minor things.

-Kevin

klippy/extras/output_pin.py
```diff
@@ -22,30 +23,41 @@ def __init__(self, config):
...
             self.last_value = static_value / self.scale
             self.mcu_pin.setup_start_value(
                 self.last_value, self.last_value, True)
         else:
+            self.reactor = self.printer.get_reactor()
+            self.host_ack_timeout = config.getfloat('host_acknowledge_timeout',
+                                                    0, minval=0.)
+            #ensure that safety timeout is big enough for comm. latency
+            if self.host_ack_timeout > 0:
+                self.host_ack_timeout = max(self.host_ack_timeout, 0.500)
```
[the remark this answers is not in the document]
Cirromulus: > `minval=max(self.host_ack_timeout, 0.500)`
Wouldn't this disallow zero? Currently, zero indicates "no timeout".

> it would give the host 100ms to respond to the event 
For this, the `relax_time` was supposed. But no problem there.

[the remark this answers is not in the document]
KevinOConnor: The minmax checks do not apply to the default.  So, it wont stop the default of zero - though it would prevent a zero in the config - but that seems okay to me.

-Kevin

docs/Using_PWM_Tools.md
```diff
@@ -0,0 +1,67 @@
...
+so that when your host or MCU encounters an error, the tool will stop.
+
+For an example configuration, see `config/sample-pwm-tool-cfg`.
+
+## Current Limitations
+
+There is a limitation of how frequent PWM updates may occur.
+While being very precise, a PWM update may only occur every 0.1 seconds,
+rendering it almost useless for raster engraving.
+However, there exists an [experimental branch](https://github.com/Cirromulus/klipper/tree/laser_tool) with its own tradeoffs.
```
Cirromulus, line 32: Is this mention ok for you, or shall this be kept out until the PR is actually started?

KevinOConnor:
Thanks.  I committed this along with a couple of minor changes on top (commit 861144d8 and d02c80ef).

-Kevin

Issue #3981, opened 2021-02-23
Klipper process doesn't survive FIRMWARE_RESTART with configured raspberry MCU


ScotchTape:
When editing printer,cfg in one of the WebUIs and issue a Save&Restart or FIRMWARE_RESTART klippy exits and needs to be restarted manually when a Raspberry Pi is configured as additional MCU. Once it's removed a FIRMWARE_RESTART comes back as expected.
I'm using a Duet2 Wifi and, to be sure, setup a new image yesterday and using the most recent pull from today.
The klippy.log and the journal is attached with both cases.

[journal.zip](https://github.com/KevinOConnor/klipper/files/6026246/journal.zip)
[klippy.log](https://github.com/KevinOConnor/klipper/files/6026247/klippy.log)

dalegaard:
I believe this happens because Klippers gets killed by a `SIGHUP`, presumably because `klipper_mcu` took its pseudo-terminal with it when it restarted. I've been launching `klippy` with `nohup` to avoid this issue, as I wasn't sure if this was a problem on my setup or a general problem and didn't have time to dig into this deeper. This "fixed" the problem by forcing Klippy to restart when it encounters an error when trying to talk to `klipper_mcu`.

ScotchTape:
Ok, thanks for the tip, I’ll give it a try. 
But regardless if the workaround fixes it, it should be fixed, since I’m not the only one with this problem, according to the official discord. :)

dalegaard:
Since this is affecting me, I'll take another crack at it!

Sineos:
Can confirm this behavior. I created a `service klipper restart` link in octoprint to easily work around this

dalegaard:
This should be fixed with the merge of #3996

ScotchTape:
I can confirm it's fixed for me, reverted the nohup fix in the unit file, rebuild the linux process, installed it and a FIRMWARE_RESTART works as expected.
I leave it open for now.

Thanks

PR #3996, opened 2021-02-27
Fix klipper_mcu causing klippy to die

dalegaard:
When `klippy` requests a `FIRMWARE_RESTART` from `klipper_mcu` the open pseudo-TTY between them is closed and a `SIGHUP` gets sent to `klippy` which subsequently kills it. This is the cause of #3981.

Here I've added a `SIGHUP` handler that simply ignores the signal, but that seems like throwing the baby out with the bathwater, as it will stop `klippy` from dying if run from e.g. a terminal which is then closed(instead of properly killed e.g. with `Ctrl-C`/`SIGINT`). I'd love suggestions on how to better handle this, if relevant, before merging this.

I've been scouring over the kernel code for `devpts` and from I've been able to tell it's not possible to disable this behavior of sending a `SIGHUP` so we need to somehow handle this situation ourselves.

I also did a quick change in `serialqueue.c` to stop it from outputting errors like `Got error 0 in read: (0)Success`.

### DIFF
--- a/klippy/chelper/serialqueue.c
+++ b/klippy/chelper/serialqueue.c
@@ -573,7 +573,10 @@ input_event(struct serialqueue *sq, double eventtime)
     int ret = read(sq->serial_fd, &sq->input_buf[sq->input_pos]
                    , sizeof(sq->input_buf) - sq->input_pos);
     if (ret <= 0) {
-        report_errno("read", ret);
+        if(ret < 0)
+            report_errno("read", ret);
+        else
+            errorf("Got EOF when reading from device");
         pollreactor_do_exit(&sq->pr);
         return;
     }


KevinOConnor:
I can't reproduce this issue on my boxes (neither a Fedora 33 Linux desktop nor on Octopi).  I can issue both RESTART and FIRMWARE_RESTART without any errors or problems.

So, this will need to start with the troubleshooting information showing the problem and, ideally, an investigation into why it differs between machines.

-Kevin

dalegaard:
@KevinOConnor very interesting. Are you using systemd or old style init scripts in your setups? In my setup `klippy` is run as a systemd service and when I issue a `FIRMWARE_RESTART` the daemon dies and `systemctl status klipper` shows the daemon as having exited with a `HUP`.

dalegaard:
@KevinOConnor I played around a little bit more with this, I expect you are either not using the same systemd unit as given in the install scripts(I'm using the one from `install-debian.sh`), or you are using classic init. 

The systemd units specify `RemainAfterExit=yes` which is why Klipper doesn't automatically restart. Because of this setting, when the daemon dies with a `HUP`, systemd doesn't concern it to have died and thus doesn't restart it.

If I remove this directive, a `FIRMWARE_RESTART` kills klipper and then restarts it 10 seconds later(the set `RestartSec` directive). If using e.g. `start-stop-daemon` I expect a similar thing would happen, though I don't have any experience with this.

Firstly, I think `RemainAfterExit=yes` should definitely be removed from these scripts, I can't see a specify reason for having it. What is your take on this?

Secondly, should this be fixed in Klipper itself(as I've tried to do in this commit) or should it be the job of a process supervisor(e.g. systemd, `start-stop-daemon`) to restart Klipper after it dies from the `klipper_mcu`-induced `SIGHUP`?

dalegaard:
Upon further inspection it looks like `start-stop-daemon`, which is used by Octopi as far as I can tell, doesn't implement anything for restarting the daemon. So in this case, `klippy` would simply die and not get restarted.

KevinOConnor:
I'm using the standard OctoPi install - as described in docs/Installation.md .  The klippy process is not being restarted - I see no adverse effects to a FIRMWARE_RESTART.

-Kevin

dalegaard:
@KevinOConnor okay, I'll try a clean install using OctoPi on my Raspberry Pi and verify. For reference, here's my entire config to replicate this locally:
```
[printer]
kinematics: none
max_velocity: 1
max_accel: 1

[mcu host]
serial: /tmp/klipper_host_mcu
```
This is obviously cut down from my full normal config. Using this config and issuing a `FIRMWARE_RESTART` kills klipper with:
```
klipper.service: Main process exited, code=killed, status=1/HUP
```

dalegaard:
@KevinOConnor can I get you to try something like `ps aux | grep klippy` on your OctoPi installation? Specifically I'd like to see if the process is launched as a session leader under `start-stop-daemon`. On my installation, with `systemd` as the supervisor, I get:
```
pi        2065  3.0  1.0  52928 17700 pts/6    Ssl+ 21:03   0:00 /home/pi/klippy-env/bin/python /home/pi/klipper/klippy/klippy.py /home/pi/klipper_config/printer.cfg -l /tmp/klippy.log -a /tmp/klippy_uds
```
The lower-case s in the `Ssl+` flags is indicating that `klippy` is a session leader. The `SIGHUP` sent is sent to the session leader of the pseudo-TTY and thus if `klippy` is not launched as a sesson leader it obviously won't get that signal. E.g. if I launch `klippy` straight from a terminal I get:
```
pi        2101 51.0  1.0  52928 17636 pts/3    Sl+  21:05   0:00 /home/pi/klippy-env/bin/python /home/pi/klipper/klippy/klippy.py /home/pi/klipper_config/printer.cfg -l /tmp/klippy.log -a /tmp/klippy_uds
```
Which makes sense, as the terminal is the session leader.

When running like this, I see `Got error 0 in read: (0)Success` in my `klippy.log` though - as the pseudo-TTY was closed, we reached EOF.

KevinOConnor:
```
pi@octopi:~/klipper $ ps aux | grep klippy
pi       14695  8.5  2.1  69872 19176 ?        Sl   21:12   0:01 /home/pi/klippy-env/bin/python /home/pi/klipper/klippy/klippy.py /home/pi/printer.cfg -l /tmp/klippy.log
```

dalegaard:
@KevinOConnor thanks, that explains the difference we are seeing then!

I see two ways to handle this:
- `klippy` gets modified to support being a session leader(what I've tried in this commit)
- Stop `klippy` from becoming a session leader. A simple way to do this for `systemd` is to wrap the `ExecStart=` line in `/bin/sh -c "..."` line.

While I don't really like the "hackish" way of the second choice, it is very simple to implement and avoids having to change Klippy and add special-case code for this problem.

jakep82:
I checked out this branch, and it finally solves the longstanding issue with Klipper dying on my Replicape. I assumed it was specific to the PRU, but I'm now able to e-stop and firmware_restart successfully.

dalegaard:
@KevinOConnor I've been thinking a bit about this, specifically ignoring all SIGHUP signals and thus breaking foreground usage of `klippy`.

Instead of ignoring SIGHUP in all cases, we could ignore the signal only if `klippy` is running as a session leader (`getsid(0) == getpid()`).

This means that in existing usage where `klippy` runs as a non-leader, nothing changes.

In cases where `klippy` is the session leader (generally, that would be when run by a process supervisor) the only SIGHUP that we would be receiving is from the `klipper_mcu` device pseudo-TTY going away and thus it's safe to ignore.

What do you think of such a solution?

dalegaard:
@KevinOConnor I fixed this the proper way now.

TIL that under systemd a process must take care to avoid grabbing a controlling terminal, to avoid the issues caused by that device going away. This is done by setting `O_NOCTTY` on the file when opening. PySerial already did this under the covers which is why this was never an issue with e.g. a serial device getting unplugged.

I've force pushed a proper fix.

Also, thanks for pushing back on this, I learned something new about Unix in the process :-)

KevinOConnor:
Thanks.  I pushed the noctty change (with a minor formatting change) to the master branch (commit e0db9f3a).

The other change doesn't look correct to me - we really want a debugging message if the file gets closed on us.  (Separately, I'm not sure the current code is valid because it calls a python function via cffi using a thread not created by python - I'm unsure if you're reporting that here, or reporting a confusing debug message.)

-Kevin

dalegaard:
@KevinOConnor super!

The other commit was because of this confusing error message:
```
Got error 0 in read: (0)Success
```
I can definitely see your logic of wanting so show that the file closed instead of simply squashing the error message!

I've pushed f99496c just now to change this to show a message `Got EOF when reading from device` instead of simply ignoring the error.

KevinOConnor:
Thanks.

-Kevin

PR #4000, opened 2021-02-28
Update printer-flsun-q5-2020.cfg

FrYakaTKoP:
we just found a small problem: Some people have slightly higher endstop positions, which caused them to have a move out of range during probe calibrate. 

-FrY

Signed-off-by:  Christoph Frei <fryakatkop@gmail.com>


KevinOConnor:
Thanks.

-Kevin

PR #4005, opened 2021-02-28
config: fan pin number correction for Robin Nano 3

alexeyzel:
There are two fan pins on Robin Nano 3 and the fan1 pin is PC14, not PB1.
PB1 pin that was in config example is used for the fan2 output.

Signed-off-by: Oleksii Zelivianskyi alexeyzel@gmail.com


KevinOConnor:
Thanks.

-Kevin

PR #4011, opened 2021-03-02
Create printer-hiprecy-leo-2019.cfg

morganfw:
A working configuration example for Hiprecy Leo printer (https://www.hiprecy.com/product/402/) with no Touchscreen funcionalities because separated LCD firmware needed.


KevinOConnor:
Thanks, but commits need to have a signed-off-by line for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

Also, all printer-XXX.cfg files need to end in a date (a rough year the printer is from).  Also, it would be good to identify the gearing ratio on the extruder - as we typically prefer to set the rotation_distance based on the effective circumference of the final hobbed bolt.

-Kevin

PR #4012, opened 2021-03-02
Add support for STM32F1 I2C1 alternate pin mapping

1am3r:
This adds support for the alternate pin mapping of I2C1 on PB8/9 in STM32F1 devices


KevinOConnor:
Thanks.

-Kevin

PR #4017, opened 2021-03-04
Add CNC type commands

nytowl:
CNC machines use G0[0-9] for rep-rap type commands G[0-9]. Parse the CNC
style as well.

Is this all thats need or are there more spots to change this?

### DIFF
--- a/klippy/extras/gcode_arcs.py
+++ b/klippy/extras/gcode_arcs.py
@@ -20,7 +20,9 @@ class ArcSupport:
         self.gcode_move = self.printer.load_object(config, 'gcode_move')
         self.gcode = self.printer.lookup_object('gcode')
         self.gcode.register_command("G2", self.cmd_G2)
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  However, my high-level feedback is that this can already be done without code changes by adding `[gcode_macro G01]` type config sections to the printer.cfg file.  So, I'd be inclined not to make this code change.

-Kevin

grigi:
Is the point not also to have good defaults? Could this then not be included as a standard macro, just so that people don't end up with lots of boilerplate in their config file?

KevinOConnor:
> Is the point not also to have good defaults?

FWIW, I prefer to target "slicer support" and not "gcode support".  If we can identify a commonly used program that produces a particular gcode dialect (in its default configuration), then it might be worth looking at adding support for that particular program.

-Kevin

KevinOConnor:

-Kevin

PR #4018, opened 2021-03-04
changed standard timout gcode

fleinze:
When using klipper with a machine without heaters (e.g. a plotter) there is an error when the idle timeout is applied: (Unknown command:"TURN_OFF_HEATERS")

This PR fixes this by checking if there are any heaters to turn off.

I tested this with a normal printer and with a machine without heaters.

--fleinze

### DIFF
--- a/klippy/extras/idle_timeout.py
+++ b/klippy/extras/idle_timeout.py
@@ -6,7 +6,9 @@
 import logging
 
 DEFAULT_IDLE_GCODE = """
-TURN_OFF_HEATERS
+{% if 'heaters' in printer %}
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  The change looks fine to me, but commits need to have a signed-off-by line for me to directly commit - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

fleinze:
Sorry for that. I updated the commit message and it should now conform to the contribution guidelines.

KevinOConnor:
Thanks.

-Kevin

PR #4021, opened 2021-03-05
New configuration printer-hiprecy-leo-2019.cfg

morganfw:
A working configuration example for Hiprecy Leo printer (https://www.hiprecy.com/product/402/) with no Touchscreen funcionalities because separated LCD firmware needed.
TMC2130 SPI drivers.
The Self-levelling plate sensor is PINDA probe type
Direct Drive is a BMG "50:17" (3:1) gear ratio

Signed-off-by: Morgan Barresi <morganfw@gmail.com>

### DIFF
--- a/config/printer-hiprecy-leo-2019.cfg
+++ b/config/printer-hiprecy-leo-2019.cfg
@@ -135,6 +135,10 @@ kick_start_time: 0.100
 heater_temp: 50.0
 fan_speed: 1.0
 
+#fan for control board FAN2
+[heater_fan board_fan]
+pin: PL3
+
 #Prevents communication issues with SPI drivers
 [static_digital_output disable_sdcard]
 pins: PB0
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Commits need to contain a signed-off-by line containing your real full name - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

I can usually commit a config file using my signed-off-by line if you prefer.  However, I would only do that for a base config (no comments or added features).

-Kevin

morganfw:
> Commits need to contain a signed-off-by line containing your real full name - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md
> 
> I can usually commit a config file using my signed-off-by line if you prefer. However, I would only do that for a base config (no comments or added features).
> 
> -Kevin

Thanks Kevin, I added more fine tuning and uncommented the bed_mesh and homing_override parts.
I would ask if it is possible to commit as is, or you need something else?

KevinOConnor:
Thanks.  I committed this change.  I had to add in a probe z_offset for the test cases to pass (commit b0f94e50).  If there is a better default z_offset, let me know.

-Kevin

PR #4024, opened 2021-03-08
Improve ARM Cortex-M control transfer from bootloaders

mattthebaker:
Resets stack value in a way that doesn't add hazards in vendor specific main functions.

Resets all user interrupt enables, pending interrupts and interrupt priorities.

First instruction after the reset vector compiles to `CPSID i` to disable interrupts.

May want to consider only compiling in the manual interrupt resets when a bootloader is specified in the `menuconfig`. On M3 and M4 there are 240 interrupt priority registers.

If interrupts are not enabled before jumping to the F0 system bootloader it doesn't work correctly, which is why they are turned back on before calling `armcm_main` instead of letting the klipper system code eventually handle that. I assume other bootloaders may be similar. klipper appears to work fine without it.

### DIFF
--- a/src/generic/armcm_boot.c
+++ b/src/generic/armcm_boot.c
@@ -18,6 +18,7 @@ extern uint32_t _data_start, _data_end, _data_flash;
 extern uint32_t _bss_start, _bss_end, _stack_start;
 extern uint32_t _stack_end;
 
+void ResetHandlerC(void);
 
 /****************************************************************
  * Basic interrupt handlers
@@ -27,11 +28,20 @@ extern uint32_t _stack_end;
 void
 ResetHandler(void)
 {
+    __disable_irq();
+
+    // Explicitly load the stack pointer (for some bootloaders that don't)
+    asm volatile("mov sp, %0\n bx %1"
+                 : : "r"(&_stack_end), "r"(ResetHandlerC));
+}
+
+void
+__noreturn
+ResetHandlerC(void)
+{
     // Disable SysTick irq (for some bootloaders that don't)
     SysTick->CTRL = SysTick_CTRL_CLKSOURCE_Msk;
 
-    // Explicitly load the stack pointer (for some bootloaders that don't)
-    asm volatile("mov sp, %0" : : "r"(&_stack_end));
     barrier();
 
     // Copy global variables from flash to ram
@@ -47,6 +57,7 @@ ResetHandler(void)
     //__libc_init_array();
 
     // Run the main board specific code
+    __enable_irq();
     armcm_main();
 
     // The armcm_main() call should not return
--- a/src/generic/armcm_boot.c
+++ b/src/generic/armcm_boot.c
@@ -39,10 +39,28 @@ void
 __noreturn
 ResetHandlerC(void)
 {
+    // Clear all enabled user interrupts and user pending interrupts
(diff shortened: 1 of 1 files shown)


mattthebaker:
I should note I've only verified this on an M0, STM32F042. It builds correctly for M3 and M4.

The SHP registers are something that need a test run on M3/M4. The POR value for all 12 bytes is 0, but technically some of the bytes are reserved.

src/generic/armcm_boot.c
```diff
@@ -27,12 +28,39 @@ extern uint32_t _stack_end;
 void
 ResetHandler(void)
 {
+    __disable_irq();
+
+    // Explicitly load the stack pointer (for some bootloaders that don't)
+    asm volatile("mov sp, %0\n bx %1"
+                 : : "r"(&_stack_end), "r"(ResetHandlerC));
+}
```
KevinOConnor, line 36: It would be preferable to move this to be next to the `DECL_ARMCM_IRQ(ResetHandler, -15);`, and avoid the forward declaration for `ResetHandlerC()`.

src/generic/armcm_boot.c
```diff
@@ -27,12 +28,39 @@ extern uint32_t _stack_end;
...
+
+void
+__noreturn
+ResetHandlerC(void)
```
KevinOConnor, line 40: Function should be declared `static`.  I think we can improve the name - maybe something like `reset_handler_stage_two()`?

src/generic/armcm_boot.c
```diff
@@ -27,12 +28,39 @@ extern uint32_t _stack_end;
...
+{
+    // Clear all enabled user interrupts and user pending interrupts
+    for (uint8_t i=0; i < sizeof(NVIC->ICER)/sizeof(NVIC->ICER[0]); i++) {
```
KevinOConnor, line 43: I'd avoid uint8_t on arm processors (they cause gcc to emit poor code).  I'd just use `int`.  I'd also put the `int i` declaration outside the for loop as some gcc variants complain about that C99 variant in some cases.  Use `ARRAY_SIZE()` helper instead of manually calling `sizeof()`.

src/generic/armcm_boot.c
```diff
@@ -47,6 +75,7 @@ ResetHandler(void)
     //__libc_init_array();
 
     // Run the main board specific code
+    __enable_irq();
```
KevinOConnor, line 78: FWIW, might be more clear to move the irq enable up to just after the irq setup code completes.

KevinOConnor (commented):
Thanks.  It looks good to me.  I have a few very minor comments.  Nothing that would hold up merging.

-Kevin

mattthebaker:
Thanks for the comments and insights as always. It should be all cleaned up now.

I reordered operations slightly, after realizing the interrupt enables must be cleared and flushed to memory before the pending bits can be cleared. Otherwise there is a possible race condition. There are a couple additional DSBs for this.

In look closer at the barriers I needed to consult the manual briefly: https://static.docs.arm.com/dai0321/a/DAI0321A_programming_guide_memory_barriers_for_m_profile.pdf . I believe the DSB/ISB combination before `__enable_irq()` is necessary per section 4.6, to ensure all pending bits are clear and no vector is pipelined.

PR #4028, opened 2021-03-08
config: Add BTT-SKR board for the Creality CR-6SE

Doridian:
Adds support for the BTT CR-6 after market upgrade board.
This board allows using the USB communication feature (and requires !PA14 in initial pins).
This board also runs the TMC2209 in UART mode and has basically a completely different pinout.
I also added safe_z_home to prevent the printer from trying to home Z at the far edge and crashing (as the nozzle was not over the bed at all)

Note: The PID tuning might not be correct for everyone, as I do run a different nozzle and a flexible bed, not the default glass one.

### DIFF
--- a/config/printer-creality-cr6se-btt-skr-2021.cfg
+++ b/config/printer-creality-cr6se-btt-skr-2021.cfg
@@ -1,0 +1,160 @@
+# This file contains pin mappings for the BTT-SKR Mainboard on the 2021 Creality CR6-SE.
+# To use this config, during "make menuconfig" select the
+# STM32F103 with a "28KiB bootloader" and with "Use USB for
+# communication" enabled. Also ensure "!PA14" is in the initial pins.
+
+# Flash this firmware by copying "out/klipper.bin" to a SD card and
+# turning on the printer with the card inserted. The firmware
+# filename must be named "firmware.bin".
+
+# See docs/Config_Reference.md for a description of parameters.
+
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks, but we only include example config files for stock printers.  There are just too many custom printers to try and track them.

-Kevin

Doridian:
No problem. I only PR'd this because the BTT board is a very popular addon for the CR6-SE (as the 4.5.2 vanilla mainboard is known to have quite a few issues that can lead to damaging the printer's daughterboard)

eduncan911:
> Thanks, but we only include example config files for stock printers. There are just too many custom printers to try and track them.
> 
> -Kevin

@KevinOConnor where do you recommend us curating custom board configs?  A wiki?

eduncan911:
@KevinOConnor so, this one was not merged... but this one was merged?

https://github.com/Klipper3d/klipper/pull/4378

There are some required information for the stock CR6SE printer in this PR, that is not in the PR listed above that was merged.  Such as the Probe requirement.

KevinOConnor:
I don't own this printer, so I can't comment on it.  If something is missing, someone wants to create a PR for it, and a regular contributor reviews it, then I'll merge it.

There is some info on the example configs at https://www.klipper3d.org/Example_Configs.html and we're planning to change the PR process as described at #4939 .

Cheers,
-Kevin

PR #4032, opened 2021-03-09
shaper_calibrate: Estimate max_accel and ignore resonance 'tails' during calibration

dmbutyugin:


dmbutyugin:
BTW, it would be great to get some feedback about max_accel estimation from the users: how much predicted max_accel values match the expectations and the actual values used?

dmbutyugin:
I expanded the scope of this PR to include the following features:

- Estimate max_accel that prevents excessive smoothing
- Ignore small vibrations levels when scoring input shaper configuration; this should help with long 'tails' of resonances
- Adjusted/reduced minimum allowed shaper frequencies to match projected max_accel ~= 1500

plus a few other small improvements.

frap129:
I just used your branch to calibrate input shaper on my Ender 2, and it seems to work well. I haven't had a chance to compare results against a manual test print yet, but the recommendations look reasonable. Only issue I ran into was 
`SyntaxError: Non-ASCII character '\xe2' in file /home/pi/klipper/scripts/calibrate_shaper.py on line 101, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details`
Changing the approx. equal sign back to ~= and the less than or equal to <= fixed it, so I would recommend using those for compatibility with all charsets

Sineos:
I gave this a run as well.
My current accel and shaping config for printing is:

```
[printer]
kinematics: cartesian
max_velocity: 300
max_accel: 3000
max_accel_to_decel: 2000
max_z_velocity: 6
max_z_accel: 80

[input_shaper]
shaper_freq_x: 73.6
shaper_type_x: 2hump_ei
shaper_freq_y: 56.8
shaper_type_y: ei
```

I would call these setting conservative and my printing results are really good (= basically no visible ringing at all, sharp edges). I guess I could go quite a tad higher with the accels but I'm not that speed junkie.

Generally I use following command to determine my results:
`python3 ~/klipper/scripts/calibrate_shaper.py /tmp/raw_data_....csv -o /tmp/....png --max_smoothing=0.1`

Following results are from my last run (without this branch)
shaper_calibrate_x
shaper_calibrate_y

These are the results from this PR:
20210314_shaper_calibrate_x
20210314_shaper_calibrate_y

Although the graphs of the two runs look very much similar, the "remaining vibration calculation" and the recommendation are very different.
Not being the expert but the pre-PR values made more sense in my eyes (especially for the Y-axis).

Raw values attached:
[shaper.zip](https://github.com/KevinOConnor/klipper/files/6136282/shaper.zip)

dmbutyugin:
> Changing the approx. equal sign back to ~= and the less than or equal to <= fixed it, so I would recommend using those for compatibility with all charsets

Thanks, I've rolled back that breaking change.

@Sineos Thanks for sharing the results!

> I would call these setting conservative and my printing results are really good (= basically no visible ringing at all, sharp edges). I guess I could go quite a tad higher with the accels but I'm not that speed junkie.

Fair enough, though then I guess we won't know how good or bad the max acceleration prediction is for your printer. Though high values for max_accel should not be seen as an instruction to go that high, but more like, you should not really go higher than that, but going slower is totally fine :)

> Although the graphs of the two runs look very much similar, the "remaining vibration calculation" and the recommendation are very different.

Yes, the scope of this PR was changed to include improvements in the calculation of residual vibrations, which also affects the shaper recommendations.

> Not being the expert but the pre-PR values made more sense in my eyes (especially for the Y-axis).

 Hmm, looking at your data, I wouldn't just brush away the new code recommendations. FWIW, it seems that max_smoothing=0.1 may be a bit too aggressive for your printer: it seems that the resonance at 44-45 Hz on X axis is a bit undercompensated by 2hump_ei at 73.6 Hz - you can see a bit of a spike in cyan color there. However, with MZV at 45.2 Hz it is actually not there, and the remaining vibrations remain lower across the spectrum. More importantly, if you run the script without `max_smoothing` parameter, you'd get the following results:
shaper_x
shaper_y
which I think make a lot of sense. I'm not sure how it was previously without `max_smoothing` parameter, but now it works pretty well, in my opinion. You could definitely give new parameters a try and see how they perform.

For example, I gave a try the new algorithm on the data from my printer (Ender 3 Pro). Here are the tuning results (left is old, right is new):
| old_algo_x | new_algo_x |
|:--:|:--:|
| old_algo_y | new_algo_y |

I gave the new parameters a try, and did not observe any differences in print quality: I did not get more ringing. However, with the new suggestions I can set the max acceleration a bit higher than previously.

theopensourcerer:
I ran this yesterday on my CoreXY. Just FYI, the ADXL345 was mounted vertically on the toolhead.

shaper_calibrate_x  

shaper_calibrate_y

dmbutyugin:
@theopensourcerer Hmm, but you don't have the accel estimation on your charts, so it seems that you were running the mainline code instead.

theopensourcerer:
> @theopensourcerer Hmm, but you don't have the accel estimation on your charts, so it seems that you were running the mainline code instead.

Sorry. My bad.

Sineos:
@dmbutyugin Many thanks for the detailed reply. Will definitively try it.

> Though high values for max_accel should not be seen as an instruction to go that high

Luckily I know what "less than" and "greater than" means. Although the above sentence should go into the instructions in bold, red and blinking.

From the `Measuring_Resonances.md`

> A calibration script provides an estimate for `max_accel` parameter that should not create too much smoothing. When tuning the input shaper, a minimum value of the two `max_accel` suggested values for X and Y axes should be used. **[here people stop reading]**

Let's read this as "@dmbutyugin script told me to us 6k accel" 😏

dmbutyugin:
@Sineos 

> > A calibration script provides an estimate for `max_accel` parameter that should not create too much smoothing. When tuning the input shaper, a minimum value of the two `max_accel` suggested values for X and Y axes should be used. **[here people stop reading]**
> 
> Let's read this as "@dmbutyugin script told me to us 6k accel" smirk

Well, you can totally try that and see how it works :) Or 5000 as per script run without max_smoothing parameter. 

Snarking aside, that's my worry that people will interpret it like that. Probably worth rewriting that part a bit.

Sineos:
Maybe something like:

> The "max acceleration" as displayed by the calibration script is only the theoretical maximum at which this shaper type still is able to work without producing too much smoothing. It is by no means a recommendation to set this acceleration for printing.
> The maximum acceleration your printer is able to sustain depends solely on its mechanical properties and the maximum torque of the used stepper motors.

This kind of brings me to my comment here: https://github.com/KevinOConnor/klipper/issues/3027#issuecomment-795446727
Not having to manually modify the cfgs accel values for the test would be incredibly useful. An own "measurement acceleration" could be also beneficial for using the adxl345 as endstop / probe.

KevinOConnor:
Looks good to me!  Let me know when you would like me to commit.  It does look like this will do a nice job of improving the recommendations.

BTW, are the new calibrate-*.png images from the latest code?  The examples still seem to show 3hump-ei.  Just curious.

-Kevin

dmbutyugin:
> Looks good to me! Let me know when you would like me to commit. It does look like this will do a nice job of improving the recommendations.
> 
> BTW, are the new calibrate-*.png images from the latest code? The examples still seem to show 3hump-ei. Just curious.
> 
> -Kevin

I'll make another pass over the docs and update this thread. The images (and sample logs) indeed only include the `max_accel` estimation, but not the updated algorithm to choose the input shaper. Maybe I'll leave one of the images at their old state as an example for `max_smoothing` adjustment, cause I think I don't have good examples with the new algorithm right now.


@Sineos 
> > The "max acceleration" as displayed by the calibration script is only the theoretical maximum at which this shaper type still is able to work without producing too much smoothing. It is by no means a recommendation to set this acceleration for printing.
> > The maximum acceleration your printer is able to sustain depends solely on its mechanical properties and the maximum torque of the used stepper motors.

FWIW, in my experience the stepper motors can sustain quite a bit. Like, even on Ender 3 I printed the 3D benchy at 7K accel, and the print finished successfully without motors skipping steps. It did not look particularly good because of smoothing, but that's exactly the point.

dmbutyugin:
@KevinOConnor Actually, I've updated the docs. Also included the notice about the different PCB designs from the Discord server.

dmbutyugin:
@KevinOConnor I guess if there is no further feedback, this PR can be committed.

Sineos:
Short feedback:

After some mechanical changes and maintenance, I repeated the measurement with both master and this PR

Master
```
ubuntu@ubuntu ~/klipper $ python3 ~/klipper/scripts/calibrate_shaper.py /tmp/raw_data_x_20210318_175900.csv -o /tmp/20210318_shaper_calibrate_x_pre.png --max_smoothing=.1
Fitted shaper 'zv' frequency = 56.6 Hz (vibrations = 33.7%, smoothing ~= 0.055)
Fitted shaper 'mzv' frequency = 45.2 Hz (vibrations = 16.0%, smoothing ~= 0.100)
Fitted shaper 'ei' frequency = 56.8 Hz (vibrations = 14.8%, smoothing ~= 0.100)
Fitted shaper '2hump_ei' frequency = 73.6 Hz (vibrations = 13.1%, smoothing ~= 0.100)
Fitted shaper '3hump_ei' frequency = 90.6 Hz (vibrations = 13.3%, smoothing ~= 0.100)
Recommended shaper is 2hump_ei @ 73.6 Hz
ubuntu@ubuntu ~/klipper $ python3 ~/klipper/scripts/calibrate_shaper.py /tmp/raw_data_y_20210318_180637.csv -o /tmp/20210318_shaper_calibrate_y_pre.png --max_smoothing=.1
Fitted shaper 'zv' frequency = 46.2 Hz (vibrations = 26.0%, smoothing ~= 0.078)
Fitted shaper 'mzv' frequency = 45.2 Hz (vibrations = 17.0%, smoothing ~= 0.100)
Fitted shaper 'ei' frequency = 56.8 Hz (vibrations = 16.5%, smoothing ~= 0.100)
Fitted shaper '2hump_ei' frequency = 73.6 Hz (vibrations = 16.9%, smoothing ~= 0.100)
Fitted shaper '3hump_ei' frequency = 90.6 Hz (vibrations = 17.1%, smoothing ~= 0.100)
Recommended shaper is mzv @ 45.2 Hz
```

PR
```
ubuntu@ubuntu ~/klipper $ python3 ~/klipper/scripts/calibrate_shaper.py /tmp/raw_data_x_20210318_194150.csv -o /tmp/20210318_shaper_calibrate_x_post.png
Fitted shaper 'zv' frequency = 53.6 Hz (vibrations = 16.7%, smoothing ~= 0.060)
To avoid too much smoothing with 'zv', suggested max_accel <= 11200 mm/sec^2
Fitted shaper 'mzv' frequency = 44.6 Hz (vibrations = 0.0%, smoothing ~= 0.102)
To avoid too much smoothing with 'mzv', suggested max_accel <= 5900 mm/sec^2
Fitted shaper 'ei' frequency = 53.4 Hz (vibrations = 0.0%, smoothing ~= 0.113)
To avoid too much smoothing with 'ei', suggested max_accel <= 5300 mm/sec^2
Fitted shaper '2hump_ei' frequency = 66.4 Hz (vibrations = 0.0%, smoothing ~= 0.122)
To avoid too much smoothing with '2hump_ei', suggested max_accel <= 4900 mm/sec^2
Fitted shaper '3hump_ei' frequency = 79.4 Hz (vibrations = 0.0%, smoothing ~= 0.130)
To avoid too much smoothing with '3hump_ei', suggested max_accel <= 4600 mm/sec^2
Recommended shaper is mzv @ 44.6 Hz
ubuntu@ubuntu ~/klipper $ python3 ~/klipper/scripts/calibrate_shaper.py /tmp/raw_data_y_20210318_195922.csv -o /tmp/20210318_shaper_calibrate_y_post.png
Fitted shaper 'zv' frequency = 52.8 Hz (vibrations = 3.3%, smoothing ~= 0.062)
To avoid too much smoothing with 'zv', suggested max_accel <= 10900 mm/sec^2
Fitted shaper 'mzv' frequency = 52.0 Hz (vibrations = 0.1%, smoothing ~= 0.075)
To avoid too much smoothing with 'mzv', suggested max_accel <= 8000 mm/sec^2
Fitted shaper 'ei' frequency = 46.4 Hz (vibrations = 0.0%, smoothing ~= 0.150)
To avoid too much smoothing with 'ei', suggested max_accel <= 4000 mm/sec^2
Fitted shaper '2hump_ei' frequency = 53.4 Hz (vibrations = 0.0%, smoothing ~= 0.189)
To avoid too much smoothing with '2hump_ei', suggested max_accel <= 3200 mm/sec^2
Fitted shaper '3hump_ei' frequency = 67.6 Hz (vibrations = 0.0%, smoothing ~= 0.179)
To avoid too much smoothing with '3hump_ei', suggested max_accel <= 3300 mm/sec^2
Recommended shaper is mzv @ 52.0 Hz
ubuntu@ubuntu ~/klipper $
```

With both recommendation I printed the "ringing_tower":
* 80 mm/s without any slowdowns in vase mode
* accel: 3500
* accel_decel: 2500

Result:
* Both X and Y side of the ringing tower still had some ringing left.
* X Master: 1 barely visible resonances / X PR: 1 visible + 1 barely visible
* Y Master: 2 visible resonances / Y PR: 4

[20210318.zip](https://github.com/KevinOConnor/klipper/files/6167306/20210318.zip)

dmbutyugin:
@Sineos Thanks for testing!

I have some clarifying questions, could you help me understand? You ran the shaper calibration script on master branch and PR on 2 different data sets, and not on the same? Why is that? It seems that X-axis is not too different (but still a bit different), but Y axis is notoriously different:
| x_axis | y_axis |
|:--:|:--:|

Which of the results corresponded more to the your current printer setup?

Then on the results. You printed the test not using TESTING_TOWER, but with fixed settings instead, right?


> Result:
> 
>     * Both X and Y side of the ringing tower still had some ringing left.
> 
>     * X Master: 1 barely visible resonances / X PR: 1 visible + 1 barely visible
> 
>     * Y Master: 2 visible resonances / Y PR: 4

What do you mean here? Is "resonance" == "1 wave of ringing"? Meaning that with PR recommendation the X axis had 1 more wave, and the first - a stronger one, right? Similarly for Y axis, PR recommendation resulted in 2 more waves?

Sineos:
> You ran the shaper calibration script on master branch and PR on 2 different data sets, and not on the same? Why is that? It seems that X-axis is not too different (but still a bit different), but Y axis is notoriously different:

Yes, true, I repeated the measurement with the PR branch because I wasn't 100% sure that your changes only affected the "evaluation" script. I noticed the differences as well.

> Then on the results. You printed the test not using TESTING_TOWER, but with fixed settings instead, right?

Correct. I use the ringing tower only to assess a given setting. That is, I printed the ringing tower with no shaper to compare to a baseline and to compare different shaper settings against each other (anyone interested in brand new ringing towers for sale?)

> What do you mean here? Is "resonance" == "1 wave of ringing"? Meaning that with PR recommendation the X axis had 1 more wave, and the first - a stronger one, right? Similarly for Y axis, PR recommendation resulted in 2 more waves?

Correct: "One resonance" ==  "1 wave of ringing"

I did some more testing and ran the resonance testing again 3 times. The results on the Y axis are more stable than on the X-axis.
How did you do these comparison graphs? Looks useful

Results from the visual inspection:
* Baseline X: 7 clearly visible waves / resonances / maxima
* Baseline Y: 9 clearly visible waves
* X: 2hump_ei@73,6: 1 maybe two barely visible waves / bit of smoothing
* X: 2hump_ei@66,4: no waves really visible / bit of smoothing same as above
* X: mzv@41.4: no waves really visible / bit of smoothing same as above
* Y: 2hump_ei@60.0: 1 maybe two barely visible waves bit stronger than on X / bit of smoothing
* Y: mzv@48.4: 1 and maybe a half wave visible / bit of smoothing same as above
* Y: mzv@46.8: 1 and maybe a half wave visible, a tad better than 48.4 / bit of smoothing same as above

So overall I would call the recommendations excellent. 👍 

Link to the raws:
https://drive.google.com/file/d/1jcJN8jNHHXyt8J-4P_Y1KJ4PQMQetjK0/view?usp=sharing

KevinOConnor:
Thanks!

-Kevin

Sineos:
> How did you do these comparison graphs? Looks useful

Done. RTFM helps

PR #1957, opened 2019-09-11
TMC2130, TMC2209, TMC5160 homing current

blondak:
Reducing TMC current while homing #1850

PR #4039, opened 2021-03-11
Revised sensorless homing configuration guide

KevinOConnor:
I've been tuning sensorless homing on my Voron 2.4 machine.  As part of that process, I've come up with a revised guide to tuning sensorless homing.  The main changes are: there are now tmc2209 and tmc2660 examples, the guide now recommends a procedure for finding a "minimum homing sensitivity" / "maximum homing sensitivity", and there are additional notes on selecting homing speed, currents, macros, etc.

I've only tested this procedure on tmc2209 drivers.  Feedback on users of tmc2660, tmc2130, and tmc5160 drivers would be great.

The new guide can be found in: https://github.com/KevinOConnor/klipper/blob/work-tmcdoc-20210311/docs/TMC_Drivers.md

@yschroeder @Schnilz @trevjonez - FYI.

-Kevin

### DIFF
--- a/docs/TMC_Drivers.md
+++ b/docs/TMC_Drivers.md
@@ -35,28 +35,20 @@ your (cartesian) printer. However, it works the same with all other
 axes (that require an end stop). You should configure and tune it for
 one axis at a time.
 
-## Prerequisites
-
-A few prerequisites are needed to use sensorless homing:
-
-1. StallGuard capable TMCxxxx stepper driver
-2. SPI / UART interface of the TMCxxxx wired to MCU (stand-alone mode
-   does not work)
-3. DIAG1/DIAG pin of TMCxxxx connected to the MCU
-
 ## Limitations
 
 Be sure that your mechanical components are able to handle the load of
 the carriage bumping into the limit of the axis repeatedly. Especially
-spindles (on the Z axis) might generate a lot of force. Homing a Z
-axis by bumping the nozzle into the printing surface might not be a
-good idea.
+leadscrews might generate a lot of force. Homing a Z axis by bumping
+the nozzle into the printing surface might not be a good idea. For
+best results, verify that the axis carriage will make a firm contact
+with the axis limit.
 
 Further, sensorless homing might not be accurate enough for your
 printer. While homing X and Y axes on a cartesian machine can work
-well, homing the Z axis is generally not accurate enough and results
-in inconsistent first layer height. Homing a delta printer sensorless
-is not advisable due to missing accuracy.
+well, homing the Z axis is generally not accurate enough and may
+result in an inconsistent first layer height. Homing a delta printer
+sensorless is not advisable due to missing accuracy.
 
 Further, the stall detection of the stepper driver is dependent on the
 mechanical load on the motor, the motor current and the motor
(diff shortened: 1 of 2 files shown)


trevjonez:
I should be able to walk myself through this over the weekend. Thanks for the heads up!

jperrin72:
Hi @KevinOConnor 

My printer uses TMC2130 drivers with sensorless homing - it was working great since I upgraded to v0.9.1-306. I think former version was around v0.9.1-2xx. Homing is broken and I can see now the following message in klippy.log:

...
Configured MCU 'mcu' (592 moves)
Starting heater checks for extruder
TMC stepper_x failed to init: Unable to write tmc spi 'stepper_x' register CHOPCONF
TMC stepper_y failed to init: Unable to write tmc spi 'stepper_y' register CHOPCONF
TMC stepper_z failed to init: Unable to write tmc spi 'stepper_z' register CHOPCONF
TMC stepper_z1 failed to init: Unable to write tmc spi 'stepper_z1' register CHOPCONF
TMC extruder failed to init: Unable to write tmc spi 'extruder' register CHOPCONF
Starting heater checks for heater_bed
Starting heater checks for chamber_heater
...

I'm going to open an issue for this.

KevinOConnor:
FYI, I updated the docs with some additional notes on homing CoreXY printers.

-Kevin

jperrin72:
That will be helpful – I understood the root cause for the SPI issue didn’t manage to connect it to the SPI SDCARD reader issue on FYSECT F6.

Thanks

Jérôme 


De : KevinOConnor
Envoyé le :vendredi 12 mars 2021 18:25
À : KevinOConnor/klipper
Cc : jperrin72; Comment
Objet :Re: [KevinOConnor/klipper] Revised sensorless homing configurationguide (#4039)

FYI, I updated the docs with some additional notes on homing CoreXY printers.
-Kevin

docs/TMC_Drivers.md
```diff
@@ -70,124 +62,259 @@ of the motor, so the TMC cannot detect stalls anymore. It is advised
...
+and maximum, but slightly closer to the minimum. Round the final value
+to the nearest integer value.
+
+For tmc2209 set this in the config as `driver_SGTHRS`, for other TMC
+drivers set this in the config as `driver_SGT`.
+
+If the range between `maximum_sensitivity` and `minimum_sensitivity`
+is small (eg, less than 10) then it may result in unstable homing. A
+faster homing speed may increase the range and make the operation more
+stable.
```
trevjonez, line 249: I don't know if less than 10 range being equal to unstable homing is a safe statement. Having gone through a few iterations of tests I am seeing my range be from 1-4 on both x and y with good results as far as I can tell without having run prints.

This sort of competes with the guidance above about wanting a slower speed to reduce load on the frame, but if we are to try getting a wider range, I would guess we would need higher speeds. Would it be worth making an acknowledgement to that?

I ended up settling with 40 for the speed and keeping my max as the threshold, since for the most part my motor and driver pairing seems to work great or not at all.

docs/TMC_Drivers.md
```diff
@@ -70,124 +62,259 @@ of the motor, so the TMC cannot detect stalls anymore. It is advised
...
+homing and set a new current after the carriage has moved away. This
+also allows a hold_current to be set during prints (a hold_current
+is not recommended during sensorless homing).
+
+An example macro might look like:
+```
+[gcode_macro SENSORLESS_HOME_X]
+default_parameter_HOMECUR: 0.700
+default_parameter_RUNCUR:  0.800
+default_parameter_HOLDCUR: 0.500
```
trevjonez, line 276: This seems dangerous to me. duplicating something as critical as current values in a macro vs the driver config seems like a really good place to end up with bug reports reading "I changed my currents but it didn't seem to do anything"

I know its been brought up exhaustively, but since the concept of homing current seems specific to sensorless homing why not fold this behavior into the existing driver configs instead so we might avoid duplication, macro complexity and risk? To me at least the macro feels like a clever workaround to a missing feature, which is pretty cool you can do it, its just complicated and I fear makes the approach for new comers even more daunting. 

Though I am still not convinced homing specific current is needed broadly enough to justify it as a feature complicating homing logic further.

Is there a way to create a top level constant in the cfg file that could be referenced here and in the driver config? might be a simpler approach to tame the risk, though wouldn't eliminate it.

trevjonez (commented):
Just sort of thinking aloud as I went through it.

A few ideas that came to mind while going through all this. though I don't expect they would go anywhere with any urgency, more for fun than anything.

1. It would be pretty slick if you had analytics feeding back to you so you could aggregate configuration details to help guide your efforts in a data driven way. Might be able to see some interesting data point correlation and better guide what a good range for sensitivity is given a driver+kinematic configuration. Or you might not see anything, just a thought.

2. when the stall guard threshold is not sensitive enough, we end up with the axi just vibrating against the end of the path, could the recent accelerometer support be tied into this eventually to automate the threshold search? it might provide a resonable detection that you hit max/min and need to back off and try a new value. would be probably a lot of work but a super slick feature. Could even add in a max/min current and max/min speed and let the system try to do all the hard work. take the human out of the loop as much as possible sort of idea.

3. or maybe even just a temporary mechanical switch configured as a button rather than the accelerometer could be used for tuning? then the host code might know that the axi reached the end and kept running longer than it should have. then you wouldn't necessarily need to precisely align the switch but know to keep searching the parameter matrix.

4. support for a digital + serial capable dial indicator to measure repeatability in a smart way. Could make tuning more objective should someone want to do so.

KevinOConnor:
Thanks for reviewing!

> this gcode command was not quite correct. seems the field is case sensitive

Good catch - should be fixed now.

> I find that having maximum_sensitivity in a code block here was initially confusing to me

Makes sense - I updated that.

> or maybe even just a temporary mechanical switch configured as a button rather than the accelerometer could be used for tuning

FYI, I believe it is possible to do a fully automatic search by just tracking the step position.  That is, no extra hardware is needed.

Specifically, if the user starts the process with the carriage next to the endstop, the code could track the stepper position and thus know if a homing move ends early or goes too far (ie, it clicks).

Alas, implementing that is a huge amount of work and the process is different for cartesian, delta, corexy, multi-stepper axes, etc..  So, not something I'm planning to do.

> I am still using step distance not rotation distance

step_distance is deprecated (see docs/Config_Changes.md) - it will be completely removed in a couple of months.

> SGT=1 wouldn't move, SGT=2 wouldn't trip the stall guard.

Interesting - can you describe the axis (kinematics, .9 or 1.8 degree stepper, teeth on pulley, etc.)?

> I already knew that the homing speed impacts the SGT range so I immediately doubled it 
...
I think it might be useful to describe a binary search process in lay terms 
...
Would it be worth adding some further info or decision tree about how to decide if one should adjust speed vs current or how to finally decide sensorless is not the right choice?

FWIW, I'm expecting the reader to read the document in entirety, try to understand it, be able to "read between the lines", and then put it into practice.  I think the concept as a whole is not too difficult (find the min and max sensitivity and the ideal value is between the two), but commanding a carriage into the frame is not something I'd advise for anyone "skimming the docs".

If you have suggestions on how to word that, let me know.

> I don't know if less than 10 range being equal to unstable homing is a safe statement.

Yeah - I don't know either.  I get  a range of about 30 on my tmc2209s.  The tmc2209s are notably different than the other drivers.  (Though, I expect the tmc2130, tmc2660, and tmc5160 to be very similar.)  I'm surprised the range would be so small on the tmc2130.

> why not fold this behavior into the existing driver configs instead so we might avoid duplication, macro complexity and risk?

Someone would have to implement it - and it isn't easy.  In particular, I'd only feel comfortable if the current was increased after moving away from the frame - but that changes the homing behavior - which needs rollouts and documentation.  Also, see #1957 .

> Is there a way to create a top level constant in the cfg file that could be referenced here and in the driver config? 

It's not difficult to read the config file values from a macro.  I considered doing that in the example, but I was afraid it would complicate the example to the point that it may no longer convey the "high level idea".

Thanks again,
-Kevin

KevinOConnor:
On a closer look, grabbing run_current and hold_current from the driver config isn't too bad - so I updated the example macro to do that.

-Kevin

DigDagGoo:
I did some tests with a Duet2 board (clipper new, out of the box). The board uses the TMC2660 driver. I only tested with one stepper, the stepper_x. The line from SG_TST goes as already described over the 74HC32 to PB5, I can confirm that. According to the documentation I made the printer.cfg. After a "Power ON" I can query the ENDSTOP, it is "open". Now I set SGT to 4 (not yet optimal, but it will work). Now I am homing X (G28 X0). The engine starts properly and stops properly at higher loads.

So far everything is great - now the problem ...

A new homing is no longer possible. When I query the ENDSTOP I get the wrong status "TRIGGERED" - it should be set to "open".

I can do what I want, the status remains "TRIGGERED". ONLY when I turn off the power to the TMC2660, the ENDSTOP status is reset to "open".

It seems as if a value has to be reset in the TMC2660 after a bump.

trevjonez (approved):
> Interesting - can you describe the axis (kinematics, .9 or 1.8 degree stepper, teeth on pulley, etc.)?

this was on the X axis of a MK3 Bear 2.1 frame, tmc5160 drivers, LDO 0.9 motor (LDO-42STH40-1684MAC), 16 tooth 6mm GT3 (Misumi GPA16MR2060-K-P5), healthy but not brand new super lube on misumi rods and bearings.

Out of all of the hardware config, the only thing I think would be the issue is perhaps the PowerGrip GT3 on such a short belt path. I find that the tension range is far smaller than the 2GT type belts that are OEM for a prusa machine. To set the tension I tighten until I feel the axi have a hard/easy stepping as I move it, at that point I back it off until it moves smoothly again. I assume from flex on the motor shaft or tiny inconsistencies in the belt or pully manufacturing, but it is consistent across all x/y on all three machines. I think I will order some PowerGrip GT2 which should be compatible with my existing pully/idler config and play with it again to see if I can isolate the cause of what I am feeling.

> FWIW, I'm expecting the reader to read the document in entirety, try to understand it, be able to "read between the lines", and then put it into practice. I think the concept as a whole is not too difficult (find the min and max sensitivity and the ideal value is between the two), but commanding a carriage into the frame is not something I'd advise for anyone "skimming the docs".
> 
> If you have suggestions on how to word that, let me know.

I don't think i have deep enough knowledge in the driver tech to formulate that in a way I would feel comfortable publishing. More so I was just thinking out loud in terms of "it would be nice". Obviously there is tons of machines working fine without it, so probably not worth your effort/time to do it.

> step_distance is deprecated (see docs/Config_Changes.md) - it will be completely removed in a couple of months.

Oops guess i better do that before I replicate this new config to my other machines. Thanks for the heads up, I need to watch the change logs closer.



Anyway, glad to help out when I can, writing documentation is rarely the fun part and this seems to be a really good addition for what can be a hard process to understand.

FWIW I have a handful of unused 2130 drivers now that I'd be happy to send your way if you are interested in having them on hand for testing or anything else. Personal email is in the git history here some place if you want to send me a postal address. No cost to you of course.

DigDagGoo:
I have read the latest documents (TMC_Drivers.md) and also created a macro for sensorless homing.

At first I thought it was working fine. But the first homing does not work after a power on. 

The sensorless homing does not work on my DUET2 with the TMC2660 the first time after a PowerOn. Instead of making a reference run, the stepper twitches briefly (G28 X) and then moves away (G90 and G0 ...) . All of the following homings work perfectly.

Initially believed it had something to do with the power supply, but that's not the reason for it. It seems as if the driver immediately triggers a sg-tst on its first run (after power on).

I am using the Duet2 Ethernet board, it is operated at 24V. The behavior as described above is identical with and without 5V supply via USB.

Can someone help me? Or maybe i's a Bug?

KevinOConnor:
The Duet2 Ethernet boards implement a goofy "logical or" on all the SG_TST lines.  I suspect someone would have to implement code to actually do sensorless homing on that board.  At the very least, one would have to configure every other stepper motor driver with the least sensitive `sgt` setting (63) so that they don't interfere with the homing of the desired stepper motor.

-Kevin

DigDagGoo:
I know that this is goofy... but my problem is (even) easier ...

I'm not there yet, unfortunately I already have the situation with the first driver. I will have to solve the problem you describe later - if necessary, i have to do manual changes on the board.

The output of the 74hc32 is currently only assigned to one limit switch (Line PB5)

My first problem is currently the first reference run after PowerOn with the first stepper.

And I see no reason why this shouldn't work.

Ideas?

stableconnection:
it may interest you that i have the same "only homes on 2nd try after restart" issue on the tmc2209 (worked fine previously on the tmc2130) i tried different speed and sensitivity settings but i always come back to that.

stableconnection:
> 
> 
> I know that this is goofy... but my problem is (even) easier ...
> 
> I'm not there yet, unfortunately I already have the situation with the first driver. I will have to solve the problem you describe later - if necessary, i have to do manual changes on the board.
> 
> The output of the 74hc32 is currently only assigned to one limit switch (Line PB5)
> 
> My first problem is currently the first reference run after PowerOn with the first stepper.
> 
> And I see no reason why this shouldn't work.
> 
> Ideas?

also, could you provide your stepper config ? i am currently building a delta printer with a duet2 so that would be most helpful

kolergy:
Hi, I'm starting with Klipper but I've been playing with your SENSORLESS_HOME macro & adapted it to delta printer in case it is of use for any one and added a macro STR to test different field strength (ST) and homing current (HC)
I seem to get the best results with very low currents 0.2 it is surprising because I see people putting rather high currents for the homing.

`[gcode_macro SENSORLESS_HOME]
default_parameter_HC: 0.2
gcode:
    {% set HOME_CUR        = HC %}
    {% set driver_config_A = printer.configfile.settings['tmc2209 stepper_a'] %}
    {% set RUN_CUR_A       = driver_config_A.run_current                      %}
    {% set HOLD_CUR_A      = driver_config_A.hold_current                     %}
    {% set driver_config_B = printer.configfile.settings['tmc2209 stepper_b'] %}
    {% set RUN_CUR_B       = driver_config_B.run_current                      %}
    {% set HOLD_CUR_B      = driver_config_B.hold_current                     %}
    {% set driver_config_C = printer.configfile.settings['tmc2209 stepper_c'] %}
    {% set RUN_CUR_C       = driver_config_C.run_current                      %}
    {% set HOLD_CUR_C      = driver_config_C.hold_current                     %}
    {% set step_config_a   = printer.configfile.settings['stepper_a'        ] %}
    {% set TARGET_Z        = step_config_a.position_endstop - 40              %}
    # Set current for sensorless homing
    SET_TMC_CURRENT STEPPER=stepper_a CURRENT={HOME_CUR} HOLDCURRENT={HOME_CUR}
    SET_TMC_CURRENT STEPPER=stepper_b CURRENT={HOME_CUR} HOLDCURRENT={HOME_CUR}
    SET_TMC_CURRENT STEPPER=stepper_c CURRENT={HOME_CUR} HOLDCURRENT={HOME_CUR}
    # Home
    G28
    # Move away
    G90
    G1 Z{TARGET_Z} F1200
    # Set current during print
    SET_TMC_CURRENT STEPPER=stepper_a CURRENT={RUN_CUR_A} HOLDCURRENT={HOLD_CUR_A}
    SET_TMC_CURRENT STEPPER=stepper_b CURRENT={RUN_CUR_B} HOLDCURRENT={HOLD_CUR_B}
    SET_TMC_CURRENT STEPPER=stepper_c CURRENT={RUN_CUR_A} HOLDCURRENT={HOLD_CUR_A}

[gcode_macro STR]
default_parameter_ST: 255
default_parameter_HC: 0.2
gcode:
    SET_TMC_FIELD STEPPER=stepper_a FIELD=SGTHRS VALUE={ ST }
    SET_TMC_FIELD STEPPER=stepper_b FIELD=SGTHRS VALUE={ ST }
    SET_TMC_FIELD STEPPER=stepper_c FIELD=SGTHRS VALUE={ ST }
    SENSORLESS_HOME HC={ HC }
`

PR #4040, opened 2021-03-11
Add support for BMP280 and BME680 sensors

1am3r:
This adds support for BMP280 and BME680 sensor ICs,
along with fixing calibration data readout for BME280.

Gas sensor readout for the BME680 is just the raw compensated value.
To get actual meaningful values, more research is needed.

Signed-off-by: Michael Kurz <michi.kurz@gmail.com>

### DIFF
--- a/config/sample-macros.cfg
+++ b/config/sample-macros.cfg
@@ -117,7 +117,7 @@ gcode:
     RESTORE_GCODE_STATE NAME=M600_state
 
 ######################################################################
-# BME280 Environmental Sensor
+# BMP280/BME280/BME680 Environmental Sensor
 ######################################################################
 
 # The macro below assumes you have a BME280 sensor_type defined in one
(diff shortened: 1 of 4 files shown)


Arksine:
After looking close, it does look like I had an error in H5 where I'm taking the wrong bits from address 0xE5 that you corrected.  However I think your change to H1 is incorrect.  H1 should be assigned to address 0xA1 and you have it assigned to 0x88.

1am3r:
You're totally right. You have no idea how often I looked over these values and still got it wrong. Thanks, I'll correct it.

I should have done the readout fix in a different commit to make it easier to review, totally forgot it. Sorry.

KevinOConnor:
Thanks.  It looks good to me.  One thing I noticed is that docs/Command_Templates.md should be updated.

-Kevin

KevinOConnor:
Thanks.

-Kevin

Larbaco:
so...
Can you add support for the BMP180 too?

here is the datasheet: https://cdn-shop.adafruit.com/datasheets/BST-BMP180-DS000-09.pdf
here is some python implementation: https://github.com/m-rtijn/bmp180/blob/master/bmp180/bmp180.py
Here is some C implementation: https://github.com/SMFSW/arm_i2c_drivers/blob/master/BMP180.h

PR #4045, opened 2021-03-12
Additional Notes for Creality Board v4.2.2

jmaisonguillard:
This adds further notes to the Ender 3 and Ender 3 Pro configurations that come with the stock v4.2.2 Creality STM Board as of July 2020. 

With this also, the Ender 3 profile now should resemble the Ender 3 V2 Profile because of the board change.

Signed-off-by: Julian Maison-Guillard <jmaisonguillard@gmail.com>

### DIFF
--- a/config/printer-creality-ender3-2020.cfg
+++ b/config/printer-creality-ender3-2020.cfg
@@ -1,5 +1,5 @@
 # This file contains pin mappings for the stock 2020 Creality Ender 3
-# starting from July 2020. To use this config, during "make menuconfig" 
+# starting from July 2020. To use this config, during "make menuconfig"
 # select the STM32F103 with a "28KiB bootloader" and with "Use USB for
 # communication" disabled. Also, select "Enable extra low-level
 # configuration options" and configure "GPIO pins to set at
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .

I'm a little confused by the change to printer-creality-ender3pro-2020.cfg - if the board doesn't use USB, why does PA14 need to be set?

-Kevin

jmaisonguillard:
> Thanks. However, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .
> 
> I'm a little confused by the change to printer-creality-ender3pro-2020.cfg - if the board doesn't use USB, why does PA14 need to be set?
> 
> -Kevin
Hello Kevin,

This is just a common issue on stock Ender 3 & Ender 3 Pro in where this has to be enabled or you will not be able to get Klipper to flash properly and running with the Raspberry Pi. I had tried a few other things and this had happened to be the only one that fixed that issue.

KevinOConnor:
The example config files should be based on the hardware schematic - if there is something in the schematic that indicates PA14 needs to be set then it should be identified and we can then discuss the best way to configure it.

-Kevin

KevinOConnor:

-Kevin

PR #3906, opened 2021-02-08
config: Implement aliases for advanced config editing

Annex-Engineering:
Aliases are an optional configuration file feature that resolve keys in
a configuration file, replacing them with a value defined in another 
configuration file. They behave the same way as `#define` in C/C++.
Aliases can be used for commonly changed values, or values that occur 
in multiple places within a configuration. They offer simplification  
for advanced users when frequently editing configurations.

This feature does not affect existing configurations.

Aliases are referenced within configuration files by defining keys, 
denoted by a string with a leading `$`, like `$KINEMATICS_TYPE` or 
`$MAX_VELOCITY`. Keys must be the last thing on a line (discarding 
comments). 

Aliases are defined in an `aliases.cfg`, which must reside in the same path 
as `printer.cfg`. Key-value pairs are defined as `KEYNAME = VALUE`. 
Values can consist of strings or numbers, and are replaced literally.
If `aliases.cfg` is not present, no aliases are loaded. Aliases defined 
here do not need to be used. Comments are allowed.

Example config usage:
```
#[printer]
#kinematics: $KINEMATICS_TYPE
#max_velocity: $MAX_VELOCITY
```

Example `aliases.cfg`:
```
# Aliases File
KINEMATICS_TYPE = cartesian
MAX_VELOCITY = 100
```

Aliases are implemented similarly to the "include" functionality, at the
initial line-by-line parse, in `configfile.py`. An internal method 
`_read_aliases()` is added to `PrinterConfig` to handle file loading and 
alias parsing.

Signed-off-by: Brandon Warzecho ez3dpstore@gmail.com

PR #4047, opened 2021-03-12
Add support for config file enhanced interpolation

ShawnBuckley:
Adds a backported version of `ConfigParser` to enable support for enhanced interpolation of config values. This could be useful to simplify configuration machines that use multiple Z steppers. One value becomes the single source of truth.

An example of how this is used would be (a sample from the Voron 1.8 SKR 1.4 config):
```
[stepper_z]
step_pin: P0.22
dir_pin: P2.11
enable_pin: !P0.21

rotation_distance: 8			
microsteps: 16
full_steps_per_rotation: 200

[stepper_z1]
step_pin: P2.13
dir_pin: P0.11
enable_pin: !P2.12

rotation_distance: ${stepper_z:rotation_distance}
microsteps: ${stepper_z:microsteps}
full_steps_per_rotation: ${stepper_z:full_steps_per_rotation}

```

Concerns:
- This requires a change to `requirements.txt` for klippy. Does that pose a problem for users upgrading with a previously installed pythonenv?
-  `$` becomes a keyword. It isn't in any config examples in the repo and doesn't seem like users would have any reason for it to be within their configs.

### DIFF
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -3,7 +3,8 @@
 # Copyright (C) 2016-2021  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import os, glob, re, time, logging, ConfigParser as configparser, StringIO
+import os, glob, re, time, logging, StringIO
+from backports import configparser
 
 error = configparser.Error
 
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Interesting, thanks.  I need to think on this a little.  I don't have any concern with updating the klippy-requirements.txt file.  I would like to think further about the `$` and about the implication of complicating (and thus having to document) changes to the config file format.

FYI, there was also discussion on this at #3906.

-Kevin

KevinOConnor:
I have given this further thought.  I don't think this is a good candidate for the Klipper master branch at this time.  I'm leery about adding additional syntax to the config file.  At this time, I'm not sure the gain (not duplicating some fields) is worth the complexity (documenting a new syntax, understanding its implications).  In particular, the syntax has some similarity, yet has many differences to the existing Jinja2 syntax.

It is interesting, though. Thanks.
-Kevin

PR #4057, opened 2021-03-16
module: Palette2, Add support for P2 devices

cmroche:
Module which bring support for P2 devices running in connected mode. This enables using a Palette 2 directly with Klippy without Octoprint, or without Palette2 plugins for Octoprint which are not functional out of the box with Klipper.

Module currently supports most of the expected workflows for Palette

- Pause, resume and cancel print directly from device
- Initial print setup, filament load work as expected
- Smart load from device is also working
- Heartbeat check and testing before print start to verify proper connection to the device

Module also introduces the following commands

- PALETTE_CONNECT / PALETTE_DISCONNECT
- PALETTE_CUT
- PALETTE_CLEAR
- PALETTE_SMART_LOAD

Depends on **[virtual_sdcard]** and **[pause_resume]** to provide all the needed functionality. If either of these are not present on load of **[palette2]** it will emit a configuration error information the user to add the needed configurations.

Example print (both printed at the same time using 4 input colors).

IMG_0703


klippy/extras/palette2.py
```diff
@@ -0,0 +1,610 @@
...
+        # Omega code matchers
+        self.omega_header = [None] * 9
+        omega_handlers = ["O" + str(i) for i in range(33)]
+        for cmd in omega_handlers:
+            func = getattr(self, 'cmd_' + cmd, None)
+            desc = getattr(self, 'cmd_' + cmd + '_help', None)
+            if func:
+                self.gcode.register_command(cmd, func, desc=desc)
+            else:
+                self.gcode.register_command(cmd, self.cmd_OmegaDefault)
```
KevinOConnor, line 85: What do all these commands do?  Is there a reason they are registered with the main gcode command handler?

Also, for something like this, it is a good idea to update docs/G-Codes.md .

klippy/extras/palette2.py
```diff
@@ -0,0 +1,610 @@
...
+                if res:
+                    matcher[0](params)
+                    return True
+
+        return False
+
+    def _run_Read(self, eventtime):
+        while self.serial.in_waiting > 0:
+            try:
+                raw_line = self.serial.readline()
```
KevinOConnor, line 525: Unfortunately, this is problematic, as the `readline()` call could block indefinitely - which would result in a hard to debug failure in Klipper.  Generally speaking, the Klipper host code (klippy.py) is intended to only read/write to the mirco-controller code so as to avoid issues like this.

klippy/extras/palette2.py
```diff
@@ -0,0 +1,610 @@
...
+                break
+
+            if text_line:
+                self.omega_last_command = text_line
+                l = text_line.strip()
+                logging.debug("%s -> P2 : %s" % (time.time(), l))
+                terminated_line = "%s\n" % (l)
+
+                try:
+                    self.serial.write(terminated_line.encode())
```
KevinOConnor, line 570: Same issue here - this call could block indefinitely.

klippy/extras/palette2.py
```diff
@@ -0,0 +1,610 @@
+# Palette 2 MMU support, Firmware 9.0.9 and newer supported only!
+#
+# Copyright (C) 2021 Clifford Roche <clifford.roche@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+import logging
+import os
+import threading
+import time
```
KevinOConnor, line 10: FYI, it's not a good idea to use `time.time()` in klippy - it's preferable to use `reactor.monotonic()` instead.

klippy/extras/palette2.py
```diff
@@ -0,0 +1,610 @@
...
+            raise self.printer.command_error(
+                    "Cannot initialize print, palette 2 is not connected")
+
+        startTs = time.time()
+        while self.heartbeat is None and self.heartbeat < (
+                time.time() -
+                HEARTBEAT_TIMEOUT) and startTs > (
+                time.time() -
+                HEARTBEAT_TIMEOUT):
+            time.sleep(1)
```
KevinOConnor, line 237: It's not valid to call `time.sleep()` in klippy - it'll hang the entire print process.  Use `reactor.pause()` instead.

klippy/extras/palette2.py
```diff
@@ -0,0 +1,610 @@
...
+            raise self.printer.config_error(
+                "Palette 2 requires [virtual_sdcard] to work,"
+                " please add it to your config!")
+
+        try:
+            self.pause_resume = self.printer.lookup_object("pause_resume")
+        except BaseException:
+            raise self.printer.config_error(
+                "Palette 2 requires [pause_resume] to work,"
+                " please add it to your config!")
```
KevinOConnor, line 139: It's not valid to raise an error from the `handle_ready` event - at that point we're noting the printer is ready and we can't "not ready" it.  You could use `handle_connect` instead, but it's actually probably simpler to call `printer.load_object()` in the constructor instead - that will cause Klipper to automatically load the given modules.

KevinOConnor (commented):
Thanks.  I have some comments below.  I am confused on what the gcode command registration is doing, so I could not complete a full review.

-Kevin

klippy/extras/palette2.py
cmroche: The Palette 2 works by embedded a services of device code into the gcode file, and reading them in the GCode stream. These are referred to as "O" or "Omega" codes.

As an example, this is what the intro sequence to my test print would look like (there are codes throughout the file though).

```
O21 D0014
O22 D373fd85d66ac76f6
O23 D0001
O24 D0000
O25 D0 D11565c0Dark_Blue_Filament D1ffd700Yellow_Filament D0
O26 D0005
O27 D0007
O28 D0001
O29 D0000
O30 D1 D42cc5f08
O30 D2 D438b30c2
O30 D1 D450a70f6
O30 D2 D45518aab
O30 D1 D458af459
O32 D11 D0000 D0000 D0000
O1 DMosaic-Keychain D0000115f
M0
; Sliced by CANVAS
;
; *** Actual Slicing Settings As Used ***
;

...

```

All OCodes read from the file, need to be passed to the device. Some OCodes require additional processing.

As for documentation though, as you can see, these codes are really not meant to be input on a terminal. They are highly statefull, calling in the wrong order, with incomplete data, etc, they would be incredibly difficult to document and of no practical value.

Instead special commands like PALETTE_CUT, CLEAR are exposed as separate commands to navigate this. I did, however, not add them to documentation yet.

klippy/extras/palette2.py
cmroche: Serial is initialized with a timeout, readline is guaranteed to return within about 500ms and it will insert EOL at the end. The device only returns a few bytes at a time so this is more than sufficient headroom, actually could be safely brought down to 250ms.

I could move to threads and use consumer queues on the timers instead, but didn't see other examples of this in the code so was unsure if it was simply too complex for the use case.

klippy/extras/palette2.py
cmroche: I will need to check, but the timeout should apply here too. Gauranteed to return in 500ms. Same comment as above though, would you prefer threads?

klippy/extras/palette2.py
cmroche: Noted, I'll make the change.

klippy/extras/palette2.py
cmroche: Noted, I'll make the change

klippy/extras/palette2.py
cmroche: Ok, my understanding from the documentation was that we couldn't rely on load order of modules... though I guess I misinterpreted that to mean something other than what it did. I'll make this change and test shortly.

klippy/extras/palette2.py
KevinOConnor: FWIW, there's a lot of overhead to threads.  Simpler would probably be to set the serial fd to non-blocking mode.

-Kevin

cmroche: Moved to single-threaded fully non-blocking.

klippy/extras/palette2.py
cmroche: Moves to single threaded fully non-blocking. Serial sets **timeout=0**, **write_timeout=0** in **cmd_Connect**

klippy/extras/palette2.py
cmroche: I added documentation for the user custom GCodes.

KevinOConnor:
Thanks.  In general it looks fine to me.  I do have a couple of minor comments.

I think it would be worthwhile to document the O1..O31 commands - I don't think it is necessary to document what they do, but I think it would be helpful to document there existence.  Perhaps something like:

`O1` ... `O31`: These "omega" commands are generated by slicer FooMatic and are passed to the Palette2 device.

Out of curiosity, what slicer/tool does generate these "omega" codes?

The other minor comment I have is that I found the `cmd_P2_` prefix to be confusing.  If using `cmd_` to denote gcode commands, I'd use a different prefix to denote the command handlers for the Palette2 (eg, maybe something like `p2cmd_`).

-Kevin

cmroche:
> Out of curiosity, what slicer/tool does generate these "omega" codes?

There is an online slicing tool called Canvas, https://canvas3d.io, which is commonly used. Probably the most common, but there is another OSS project called P2PP which makes slicing possible with Slic3r (and variants) by post processing the MMU toolchanger instructions in the GCode and adding the computed corresponding OCodes.

> O1 ... O31: These "omega" commands are generated by slicer FooMatic and are passed to the Palette2 device.

Very reasonable suggestion, I'll work on it this weekend.

> I'd use a different prefix to denote the command handlers for the Palette2 (eg, maybe something like p2cmd_).

Not an issue, I'll make this change on (or by) the weekend too.

cmroche:
I think this module is ready, ran a long test last night (ignore the bad retraction settings) and validated features to cancel, auto load, swap filament, etc from the device. Everything looks good, color swaps are happening at the 30%-40% mark (which is the target with Canvas purge towers) and in this print there were 40 pings, each at 100.2% - 100.6%

88BE7AEB-C2FD-4F24-907A-10241059224A_1_105_c

KevinOConnor:
Thanks.

-Kevin

devjj:
On a Palette 2S Pro I had to lower the baud to 115200 to connect.

cfelicio:
Any special considerations to use it with p2pp? I haven't received my p2 yet, but looking forward to test this with fluidd / moonraker / p2pp! :-)

cmroche:
> Any special considerations to use it with p2pp? I haven't received my p2 yet, but looking forward to test this with fluidd / moonraker / p2pp! :-)

Nope, while I have not tested it with P2PP it should work fine.

cfelicio:
Umm, I just tried adding it to my config, but I get an error. Let me know if I should open an issue for this:

Unhandled exception during run
  File "/home/pi/klipper/klippy/klippy.py", line 188, in run
    self.reactor.run()
  File "/home/pi/klipper/klippy/reactor.py", line 269, in run
    g_next.switch()
  File "/home/pi/klipper/klippy/reactor.py", line 310, in _dispatch_loop
    timeout = self._check_timers(eventtime, busy)
  File "/home/pi/klipper/klippy/reactor.py", line 156, in _check_timers
    t.waketime = waketime = t.callback(eventtime)
  File "/home/pi/klipper/klippy/webhooks.py", line 426, in _do_query
    res = query[obj_name] = po.get_status(eventtime)
  File "/home/pi/klipper/klippy/extras/palette2.py", line 641, in get_status
    "ping": self.omega_pings[-1],
IndexError: list index out of range
Transition to shutdown state: Unhandled exception during run


Here is what I added to my config file:

[palette2]
serial: /dev/serial/by-id/usb-FTDI_FT230X_Basic_UART_DM00NU90-if00-port0
#   The serial port to connect to the Palette 2.
#baud: 115200
#   The baud rate to use. The default is 115200.
#feedrate_splice: 0.8
#   The feedrate to use when splicing, default is 0.8
#feedrate_normal: 1.0
#   The feedrate to use after splicing, default is 1.0
#auto_load_speed: 2
#   Extrude feedrate when autoloading, default is 2 (mm/s)
#auto_cancel_variation: 0.1
#   Auto cancel print when ping varation is above this threshold

cmroche:
@cfelicio Thanks for the heads up. I'll look into it this evening.

cfelicio:
Just tested, seems to be working well! QQ, the P2PP adds the following at gcode generated:

O30 D2 D472df0bd
O30 D0 D472e5574
O30 D2 D472ee744
O30 D0 D4731086c
O32 D11 D0000 D0000 D0000
O1 D1_Black.p2pp D0000b108
M0
T0
;---------------------
; - SPLICE INFORMATION-
;---------------------

And klipper returns on the terminal:

21:39:09
// Unknown command:"M0"
21:39:09
// Unknown command:"T0"

Digging around the p2pp issues, I found this:

https://github.com/tomvandeneede/p2pp/pull/85

So looks like to work well with p2pp, an extra line is needed:

;P2PP KLIPPER_TOOLCHANGE

I haven't tested yet, but leaving here in case anyone else runs into the same. Thanks so much for the integration btw, really appreciated!

cmroche:
Those warnings can just be ignore

yhaiovyi-tripactions:
I'm wondering if this works with p3

cmroche:
As far as I know, it does not work with p3. I'm not sure that P3 even supports a USB based connected mode yet (at least it didn't initially though I thought it was planned). Either way, I don't have a P3, and won't buy one so unless someone is willing to donate I won't be able to test or work out how to build a compatible module.

Accessory mode should work though without this module.

myronschram:
When I try to connect to my P2S, I get this error.
19:07:05 $ PALETTE_CONNECT
19:07:15 !! No response from Palette 2
19:07:15 // Disconnecting from Palette 2

This is what I added to my printer.cfg file:
[palette2]
serial: /dev/serial/by-id/usb-FTDI_FT230X_Basic_UART_DM02GW51-if00-port0

chickenwoman:
I'm trying to figure out how move the head away from the print when the runout detector on the PP2 is triggered. Right now it just stops and sits on the print.

[gcode_macro PAUSE]
description: Pause the actual running print
rename_existing: PAUSE_BASE
# change this if you need more or less extrusion
variable_extrude: 1.0
gcode:
  ##### read E from pause macro #####
  {% set E = printer["gcode_macro PAUSE"].extrude|float %}
  ##### set park positon for x and y #####
  # default is your max posion from your printer.cfg
  {% set x_park = printer.toolhead.axis_maximum.x|float - 20.0 %}
  {% set y_park = printer.toolhead.axis_maximum.y|float - 20.0 %}
  ##### calculate save lift position #####
  {% set max_z = printer.toolhead.axis_maximum.z|float %}
  {% set act_z = printer.toolhead.position.z|float %}
  {% if act_z < (max_z - 2.0) %}
      {% set z_safe = 2.0 %}
  {% else %}
      {% set z_safe = max_z - act_z %}
  {% endif %}
  ##### end of definitions #####
  PAUSE_BASE
  G91
  {% if printer.extruder.can_extrude|lower == 'true' %}
    G1 E-{E} F2100
  {% else %}
    {action_respond_info("Extruder not hot enough")}
  {% endif %}
  {% if "xyz" in printer.toolhead.homed_axes %}
    G1 Z{z_safe} F900
    G90
    G1 X{x_park} Y{y_park} F6000
  {% else %}
    {action_respond_info("Printer not homed")}
  {% endif %} 

[gcode_macro RESUME]
description: Resume the actual running print
rename_existing: RESUME_BASE
gcode:
  ##### read E from pause macro #####
  {% set E = printer["gcode_macro PAUSE"].extrude|float %}
  #### get VELOCITY parameter if specified ####
  {% if 'VELOCITY' in params|upper %}
    {% set get_params = ('VELOCITY=' + params.VELOCITY)  %}
  {%else %}
    {% set get_params = "" %}
  {% endif %}
  ##### end of definitions #####
  {% if printer.extruder.can_extrude|lower == 'true' %}
    G91
    G1 E{E} F2100
  {% else %}
    {action_respond_info("Extruder not hot enough")}
  {% endif %}  
  RESUME_BASE {get_params}  


[gcode_macro M0]
gcode:
   Pause

chickenwoman:
[mainsail.txt](https://github.com/Klipper3d/klipper/files/8500538/mainsail.txt)
[printer.txt](https://github.com/Klipper3d/klipper/files/8500539/printer.txt)
The Palette 2 has a runout sensor that pauses the print. I cant get it to move and park the head after it pauses. It just stops and sits on the print. I've configured my pause macros in the Mainsail config. Every other pause situation will call my pause macro and move the head away from the print. I changed the M25 command to also move the head on pause. I created a M0 command just incase the Palette uses it. I'm uploading my config files. This is driving me crazy, any help would be great. Thanks

Blade765:
Am I correct in that you just plug the palette2 into my raspi running klipper, do a ls/dev/serial/by-id/* , enter that into my printer.cfg, along with the other options required by the palette2 module, then enter the PALETTE_CONNECT command in mainsail, and go from there as normal?

EVERYTIME I enter that command in mainsail, klipper crashes.

Diddle24:
Hello, I too have the problem with the Palette 2s pro. I can't get a connection. When I click Palette_Connect, klipper immediately shuts down. I'm just a bit desperate and hope for help. here is my klippy.log

[klippy.log](https://github.com/Klipper3d/klipper/files/9997872/klippy.23.log)
Unbenannt

Gutbuket:
I have exactly the same issue. Any ideas on a solution?

cfelicio:
I have not used mine for a while, I'll get it running this weekend and report back. I'll update to the latest Klipper. It was working fine back in July.

Diddle24:
I also tried to make the pallet work by resetting Klipper. But it didn't help. I then wrote to the seller, he would like to clone his SD card for me. Said and done. This is how the pallet runs. I then exchanged all cfg's, switched from Mainsail to Fluidd and made all updates. Running.

Gutbuket:
Hi @Diddle24, thanks for updating that it can work, but in the information you provided it isn't clear to me precisely what you needed/did to get it working. Are you able to elaborate on this or share files?

Diddle24:
gladly. i only need the micro sd card from the raspberry where the palette worked with klipper.

Diddle24:
I cloned the micro sd card.

Diddle24:
updated!

we found out the following (thanks to meteyou). currently the palette 2s pro only runs with python2. I have played mainsail v0.6.1 on an sd card and entered the following command via putty:

sudo apt install virtualenv python-dev
sudo systemctl stop klipper
rm -rf ~/klippy-env
CD ~
virtualenv -p python2 ./klippy-env
./klippy-env/bin/pip install -r ./klipper/scripts/klippy-requirements.txt
sudo systemctl start klipper

this will import python2 and palette2 should work. An update is to be refrained from. This bug is already in progress:

https://klipper.discourse.group/t/palette-2-connection-giving-error/4923

cfelicio:
That makes a lot of sense, I'm still running python2, that's prob why everything works :-)

cmroche:
@Diddle24 Thanks for the information, I'm setting up to today so I can sort out the cause of the issue.

Hundsbuah:
Does this also work with the Palette 3 pro?

Diddle24:
Das kann ich dir nicht sagen🤷‍♂️Am 06.05.2023 um 15:55 schrieb Christian L. ***@***.***>:﻿
Does this also work with the Palette 3 pro?

Sineos:
> Does this also work with the Palette 3 pro?

a) This is neither a help or discussion thread
b) That it likely does not work with P3 has been answered a few posts above

PR #4059, opened 2021-03-16
Added config for creality cr5pro with skr mini

Fank:
This is the config from my upgrade because i was trying to flash the existing controller with klipper and it broke during that.

https://www.thingiverse.com/thing:4783089


KevinOConnor:
Thanks, but we only add example configs for stock printers in Klipper - there are just too many customized printers for us to add examples for them.

Separately, commits need to have a signed-off-by line in order for me to commit - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

Fank:
The reason why i submitted it is:
- I was not able to make klipper work on the stock controller (Creality 3D V2.5.1)
- The CR-5 Pro is delievered fully assembled
- To share the possibility to upgrade a CR-5 Pro e.g. with a BLTouch.

If you are fine with reason above i will sign off the commits, by rebasing it.
Otherwise close the PR

KevinOConnor:
I'm going to close this as I don't think it is a good candidate for the master Klipper branch.

We did recently add a section to the Klipper Discourse server for those interested in sharing custom configs: https://community.klipper3d.org/ .  Maybe that would be a better location for this.

Thanks,
-Kevin

PR #4072, opened 2021-03-18
gcode_macro - expose math functions to gcode macros

mental405:
Added python math lib to gcode_macro and exposed it to the templates so functions like sin and cos can be caculated from gcode macros.

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -4,8 +4,7 @@
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import traceback, logging, ast, copy
-import jinja2
-
+import jinja2, math
 
 ######################################################################
 # Template handling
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks, but I'm not sure this is a good fit for the master Klipper branch.  There's a lot of stuff in the math module and I'd be concerned calling some of those functions could have adverse side-effects if called from a macro (particularly if it takes a lot of processing time).  From my perspective, the intent of the macro system is really for "quick variable substitution" and not a fully programmable interface - "extras" and the "api server" are better matches for doing stuff like that.

Separately, commits need to have a signed-off-by line for me to commit - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

mental405:
what about just
 
ceil, floor, sin, cos, tan, asin, acos, atan, fabs, fmod, trunc, pi, and  e?

Issue #4074, opened 2021-03-18
HTU21D sensor missing method, can't be used for watermark control 


tanaes:
The HTU21D class lacks the method `get_report_time_delta` required by watermark control, per issue #403. 

This method is defined in other sensors, such as the BME280: https://github.com/KevinOConnor/klipper/blob/b4437f8eeeaddf60f893ceaeaf4d9ed06d57eeae/klippy/extras/bme280.py#L126

Trying to use an HTU21D sensor to control a `heater_generic` throws an issue:

```
[heater_generic chamber_heater]
sensor_type: HTU21D
heater_pin: host.P17
i2c_mcu: host
i2c_bus: i2c.1
i2c_address: 64
gcode_id: chamber
control: watermark
max_delta: 2.0
min_temp: 0
max_temp: 70
```

```
22:45:28 Internal error during connect: HTU21D instance has no attribute 'get_report_time_delta'

Once the underlying issue is corrected, use the "RESTART"
command to reload the config and restart the host software.
Printer is halted
Klipper state: Not ready
22:45:28 Internal error during connect: HTU21D instance has no attribute 'get_report_time_delta'
```

[klippy.log](https://github.com/KevinOConnor/klipper/files/6167623/klippy.log)

PR #4075, opened 2021-03-18
module: HTU21D sensor definition fix

tanaes:
HTU21D sensor class lacked the method `get_report_time_delta` invoked
during watermark control loops. This short fix simply adds that method
in, calling the `self.report_time` attribute created during class
construction.

Fixes issue #4074

Signed-off-by: Jon Sanders <jonsan@gmail.com>


KevinOConnor:
Thanks.

-Kevin

PR #4076, opened 2021-03-19
Allow sharing endstop pins with buttons

kageurufu:
The intent is to allow an MMU unit to share the manual_stepper endstop
 with a gcode_macro (or filament_switch_sensor) to determine filament
 runout and trigger an automated filament swap, i.e for continual printing
 just swapping spools on runout.

I'm not aware if this is going to cause issues with endstops, merely that
 it works simultaneously, where a stepper can home on the endstop and a
 filament_switch_sensor activates correctly.

There may also be a better way of doing this, but I wasn't able to find
 anyone else talking about the idea

### DIFF
--- a/klippy/extras/buttons.py
+++ b/klippy/extras/buttons.py
@@ -227,7 +227,8 @@ class PrinterButtons:
         mcu = mcu_name = None
         pin_params_list = []
         for pin in pins:
-            pin_params = ppins.lookup_pin(pin, can_invert=True, can_pullup=True, share_type='endstop')
+            pin_params = ppins.lookup_pin(
+                pin, can_invert=True, can_pullup=True, share_type='endstop')
             if mcu is not None and pin_params['chip'] != mcu:
                 raise ppins.error("button pins must be on same mcu")
             mcu = pin_params['chip']
--- a/klippy/extras/buttons.py
+++ b/klippy/extras/buttons.py
@@ -227,7 +227,7 @@ class PrinterButtons:
         mcu = mcu_name = None
         pin_params_list = []
         for pin in pins:
-            pin_params = ppins.lookup_pin(pin, can_invert=True, can_pullup=True)
+            pin_params = ppins.lookup_pin(pin, can_invert=True, can_pullup=True, share_type='endstop')
             if mcu is not None and pin_params['chip'] != mcu:
                 raise ppins.error("button pins must be on same mcu")
             mcu = pin_params['chip']
--- a/klippy/pins.py
+++ b/klippy/pins.py
@@ -253,7 +253,8 @@ class PrinterPins:
     def setup_pin(self, pin_type, pin_desc, share_type=None):
         can_invert = pin_type in ['endstop', 'digital_out', 'pwm']
         can_pullup = pin_type in ['endstop']
-        pin_params = self.lookup_pin(pin_desc, can_invert, can_pullup, share_type)
+        pin_params = self.lookup_pin(
+            pin_desc, can_invert, can_pullup, share_type)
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Interesting, but I'm not sure this is a good way to do that.  I'd have two concerns with this - if a user uses the same pin for two different devices it's most likely an error that the user should be warned about; and if the user did want to share the two devices it is unlikely they'd want to run the switch action during an actual homing request.

I think if some functionality is desired that uses a single pin for both endstop and something else then we'd really need Python code to make that work.  It seems unlikely to work well by just allowing the pin to be used multiple times.  Maybe you could describe in more detail how you've used this change?

Separately, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

kageurufu:
Signed off on those commits, I knew I was forgetting something.

So the basic use case right now is for the [Enraged Rabbit Project](https://github.com/EtteGit/EnragedRabbitProject/), a DIY multi-material filament changer. It has essentially a filament sensor immediately past the changer, that is used as an endstop for the filament drive gears. Having a separate pin for essentially the same switch seemed silly, so this was our brainstormed "fix" to get the multiple uses there. The second issue, I thought about and we're testing with `SET_FILAMENT_SENSOR` to toggle things before/after homing, but thats definitely a special case too.

For the first issue, the better idea I had was a way to define a reuseable pin alias which would make users be explicit about sharing. Either by augmenting the pin definition parsing to allow adding a "shared" flag, or with explicit aliases.

```
[pin_alias ercf_filament_sensor_pin]
pin:  !P1.0
shared: true

[manual_stepper gear_stepper]
...
endstop_pin: pin:ercf_filament_sensor_pin

[filament_switch_sensor ercf_filament_sensor]
pin: pin:ercf_filament_sensor_pin
```

If you want to chat there, i'm on the klipper discord as well, kageurufu#9003

EtteGit:
Hi! Just to add,  as the ERCF dev, this hack works nicely, since we're never using the 2 fonctions (endstop on a manual stepper and filament sensor) at the same time
 But I guess a cleaner implementation (user side I mean,  like declaring explicitly that you want the same pins for the two objects) would be better.

KevinOConnor:
FWIW, I agree this functionality is useful.  However, my current feedback is that I don't think it would be a good idea to use the existing "shared pins" system for this.  My fear is that it would be too easy for a user to inadvertently use the same pin twice due to a typo.  

Unfortunately, I don't have good advice on a better implementation either.

-Kevin

GregoryHlavac:
Maybe a configuration switch on that pin somewhere in config that makes it explicitly allowed? Or else an error can be pitched?

Tircown:
maybe a virtual_endstop for the filament_switch_sensor module like the ones used for the tmc drivers, i.e: `filament_switch_sensor_<name>:virtual_endstop` as value for the manual_stepper endstop_pin?

KevinOConnor:
FYI, I have a separate PR up for disabling the "pin used multiple times in the config" error check - PR #4388.  I'm not sure that's a good solution for your use case, but it may be of interest.

Separately, I have a longer term goal of reworking the endstop code so that it internally registers its pins as "buttons", so that the frontends can get immediate feedback on endstop switch changes.  Again, not sure if that helps with your use case.

-Kevin

th3fallen:
@KevinOConnor i can confirm that your branch works perfectly for this use case!

KevinOConnor:
Is this PR still needed now that #4388 is merged?

-Kevin

th3fallen:
> Is this PR still needed now that #4388 is merged?
> 
> -Kevin

No #4388 will do this job perfectly

PR #4080, opened 2021-03-20
Create printer-prusa-mk3s-2021.cfg

tillin9:
Example printer.cfg for a Prusa MK3S and MK3.  

It includes macros (I noticed most other examples don't) but some of these are needed 
for the filament sensor as well as to restore the Tram Z/ Calibrate Z functionality. 
I also really wanted something that would work fully out-of-the-box with all Prusa 
default gcode and even with a UI like Mainsail.   

Explicitly put the Prusa USB-to-serial issue note in there as well as workarounds based 
on discord feedback by Arksine. 

My normal config is just this with the appropriate options uncommented for my BMG, 
0.9-degree stepper Prusa, and then my PID, mesh, and IS along with an extra optional 
macro or two I find helpful for UI. 

Explicitly confirmed this file loads without error in latest Klipper.

Based largely off the old example in the PrusaOwners fork. 

Signed-off-by: Scott Knauert <sknauert@apsec.com>

### DIFF
--- a/config/printer-prusa-mk3s-2021.cfg
+++ b/config/printer-prusa-mk3s-2021.cfg
@@ -1,21 +1,21 @@
-# This file contains pin mappings and reasonable defaults for Prusa i3 MK3S 
-# printers. It will work with MK3 (non-S) by selecting the alternative filament 
-# sensor. References to common community mods are included as well. 
+# This file contains pin mappings and reasonable defaults for Prusa i3 MK3S
+# printers. It will work with MK3 (non-S) by selecting the alternative filament
+# sensor. References to common community mods are included as well.
 #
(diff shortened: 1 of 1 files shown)


tillin9 (commented):
Fixed the whitespace so build completes.

KevinOConnor:
Thanks.  I'm happy to add a prusa example config file, but the example would need to be updated to match the style and formatting of the existing config files.  We want the configs to focus on describing the hardware.

I have added a document for some guidelines on the configs: https://www.klipper3d.org/Example_Configs.html

-Kevin

KevinOConnor:
I'm going to close this PR as it looks like it has gone stale.

-Kevin

Issue #3038, opened 2020-07-02
Pressure advance not applied to extruder_stepper


theophile:
So I noticed that pressure advance settings get applied to the target extruder but not to a defined extruder_stepper that is synced with the active extruder. In my case, where the extruder_stepper is between the active extruder and the hotend, this essentially means no pressure advance.

I know you can't pass the extruder_stepper config name to SET_PRESSURE_ADVANCE with the EXTRUDER parameter. Is there some other way to do it?

klipper-gitissuebot:
Hi @theophile,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

Unfortunately, too many people have opened tickets without providing the log. That consumes developer time; time that would be better spent enhancing the software. If this ticket references an event that has occurred while running the software then the Klipper log must be attached to this ticket. Otherwise, this ticket will be automatically closed in a few days.

For information on obtaining the Klipper log file see: https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md

The log can still be attached to this ticket - just add a comment and attach the log to that comment.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

theophile:
[klippy.log](https://github.com/KevinOConnor/klipper/files/4873763/klippy.log)

I poked around in extruder.py enough to figure out (I think) that pressure advance is something that can only be applied to an "extruder" object, not a "stepper" object, and that since an extruder_stepper is not an "extruder," it doesn't seems like it's currently possible for pressure advance to be applied to an extruder_stepper. I made an initial attempt to apply the pressure advance setting of the current extruder to the extruder_stepper, similar to how the last commanded position of the current extruder is applied to the extruder_stepper, but it became clear that it's not that simple, because the extruder_stepper object doesn't know what to do with pressure advance settings. Conceptually, I would think that the extruder_stepper should always have the same pressure advance setting as the extruder it's synced to, as I can't imagine a situation where a user would want two steppers in the same filament path to have different pressure advance settings.

dmbutyugin:
[...]

PR #3920, opened 2021-02-12
extruder: Add mixing extruder (with example usage in a printer.cfg)

konsumverweigerer:
Supported G-codes are:

M163 Sx Pa.a:b.b:..:h.h, M164 Sx
     - set ratios for virtual mixing extruders
M567 Px Ea.a:b.b:..:h.h
     - set mixing for virtual mixing extruders
G1 Ea.a:b.b:..:h.h
     - extrude with mixing (allowed when mixingextruder is active)

ACTIVATE_EXTRUDER EXTRUDER=mixingextruder<x>
MIXING_STATUS EXTRUDER=mixingextruder<x>

Signed-off-by: Peter Gruber <gruberp@googlemail.com>

PR #4082, opened 2021-03-20
Add support for pressure advance on extruder steppers

theophile:
This PR is intended to address the issue discussed here: #3038
It should cause an extruder's pressure advance settings to be applied to all extruder steppers synced with that extruder. It works as expected in my setup that has a single extruder stepper that is synced to only one of six extruders at a given time. I haven't been able to test with multiple extruder steppers synced to a single extruder, but it should work the same even in that case.

This also enables the use of the "SET_STEPPER_ENABLE" extended gcode command on extruder steppers. For my application, this is useful for disabling the extruder stepper to avoid needless wear on the BMG gears during parts of filament load/unload operations when the extruder stepper is "empty."


KevinOConnor:
Thanks and sorry for the delay in responding.

My main high-level feedback is that I'm not sure if we should set pressure advance for an "extruder" (ie, hotend) or for each stepper motor.  I'd be inclined to say that it should apply to the stepper motor and not the hotend though.  There was some related discussion on this at #3920.

Separately, the stepper_enable change does not look correct to me.  Unless I'm missing something, the `stepper.PrinterStepper()` call should have already added the given stepper motor to stepper_enable.

-Kevin

theophile:
Thanks @KevinOConnor. I'll double check the stepper_enable code. Regarding the main issue, as I appreciate the discussion at #3920, I think a mixing hotend needs multiple stepper motors that can move independently from one another, so as to be able to control the rate at which multiple filaments are melted and then pushed out of the nozzle. I don't think the extruder_stepper module currently allows for that.

If the extruder_stepper module were to be reworked so as to allow for an extruder_stepper to be controlled independently and not simply "mirror" the moves of another extruder, then I think that should include the ability to set PA independently on the extruder_stepper as well. But the current behavior of the extruder_stepper module is that the extruder_stepper essentially just mirrors the moves of the synced extruder, with two exceptions (that I've observed anyway):

1. The step_distance (and I assume the new rotation_distance) of each stepper motor is applied independently; and
2. The extruder_stepper ignores PA moves regardless of how the synced extruder is configured.

The first distinction is good and helpful because it allows multiple stepper motors to be dialed in independently to ensure that they each push out _precisely_ the same amount of filament on the same extrusion move.

The second distinction is, in my view, not expected or intuitive. I can't currently think of a use case where a user would want to have an extruder_stepper synced with an extruder, but not perform any of the PA moves that the main extruder performs. It seems most intuitive to me that when an extruder_stepper is synced with an extruder, it should mirror _all_ the moves of the synced extruder, including PA moves, and not just the main extrusion moves. That's what this PR is trying to do.

KevinOConnor:
Thanks.  My high-level feedback is that I think we need to implement some architectural changes to Klipper's treatment of extruder stepper motors before we can add enhancements in this area. Currently the klipper extruder config section defines the parameters to a hotend, a heater, a temperature sensor, and an extruder stepper motor. I think we need a more clear abstraction so that it is possible to better control "extruder stepper motors".  If we can better control extruder stepper motors, it will make it easier to set the pressure advance parameters for each extruder stepper - which I think may be necessary in some cases.

FYI, this also impacts #4566 and #4489 .

-Kevin

KevinOConnor:

-Kevin

KevinOConnor:
FYI, PR #5143 adds support for calling SET_PRESSURE_ADVANCE on extruder_stepper objects.

-Kevin

PR #4083, opened 2021-03-20
Z tilt autodetect

sensille:
Hi Kevin,

This branch adds two commands, Z_TILT_CALIBRATE and Z_TILT_AUTODETECT. The goal is to improve the accuracy of Z_TILT_ADJUST, please see the documentation commit.
This is only tested on my machine, so before merging it would be good to find more testers, I'll ask around.
If you have any comments regarding coding style / documentation/ integration, I could work on them in the meantime.

-Arne

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -953,7 +953,15 @@ extended [G-Code command](G-Codes.md#z-tilt) becomes available.
 #   stepper. It is described using nozzle coordinates (the XY position
 #   of the nozzle if it could move directly above the point). The
 #   first entry corresponds to stepper_z, the second to stepper_z1,
-#   the third to stepper_z2, etc. This parameter must be provided.
+#   the third to stepper_z2, etc. This parameter must be provided,
+#   unless the parameter "extra_points" is provided. In that case only
+#   the command Z_TILT_AUTODETECT can be run to automatically determine
+#   the z_positions. See 'extra_points' below.
+#z_offsets:
+#   A list of Z offsets for each z_position. The z_offset is added to each
+#   probed value during Z_TILT_ADJUST to offset for unevenness of the bed.
+#   This values can also be automatically detected by running
+#   Z_TILT_CALIBRATE. See "extra_points" below.
 #points:
 #   A list of X,Y coordinates (one per line; subsequent lines
 #   indented) that should be probed during a Z_TILT_ADJUST command.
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Interesting, thanks.  My high-level feedback is that I fear the complexity of this new code and configuration limits its usefulness.  I suspect there will not be many users willing to navigate the additional configuration parameters to optimize an already working z tilt system.

Similarly, on the implementation side, it seems there is a lot of new math added and I'm struggling to fully understand it.

Is it possible to simplify this functionality?  For example, could the code record the results of a multi-round Z_TILT_ADJUST session and then use the coordinate_descent() code to recommend refinements to the existing pivot points?

-Kevin

sensille:
Thanks for the feedback. I share your concerns regarding the configuration. I was thinking of using an existing mesh configuration to define the extra points. That would bring the necessary extra config down to zero. Regarding the math it looks worse than it is, it's just basic linear algebra. I probably just did a bad job of describing what I did there. It's unfortunate I have to pull in numpy for this, though.
For now I haven't been able to come up with a simpler algorithm, but I will give it another try. The problem is that we have 6-9 variables here. You need a lot of data to get good results for that many variables.
I think we really have to distinguish between the 2 features I added, they're independent of each other. Imho Z_TILT_CALIBRATE is simple and can be generally useful, especially when we  leverage an existing mesh config.

-Arne

KevinOConnor:
> Regarding the math it looks worse than it is, it's just basic linear algebra. I probably just did a bad job of describing what I did there. It's unfortunate I have to pull in numpy for this, though.
For now I haven't been able to come up with a simpler algorithm, but I will give it another try. 

FWIW, I've found that `coordinate_descent()` was able to simplify the code of similar tasks.

> The problem is that we have 6-9 variables here. 

Could you list what those variables are?  That would help me to review and I think may help in discussing alternatives.

-Kevin

sensille:
> Could you list what those variables are? That would help me to review and I think may help in discussing alternatives.

Simply the x/y coordinates of the pivot points (6), plus the distance of them to the bed (3). The distance plays a role, too, because the bed does not turn around the configured pivot point's x/y, but around a point below the bed. That makes a difference.

I'm glad for any ideas how to approach this differently. Although the algorithm works extremely well on my machine, testing has shown that on machines quite different from mine the results are not that convincing. I have been quite busy the last few weeks, but I hope I can get back to this soon.

tntclaus:
This looks very interesting. Will give it a try tomorrow.

tntclaus:
@sensille I tried it on my custom built triple z axis printer and I can say that this is definitely useful functionality. Requires some polishing and optimisations, but it is already better than manually finding x/y offsets. 

What I didn't understand is `z_offset`. FWIW including z coordinate into `z_positions` list to indicate how far below table mount point is might be better.

sensille:
> @sensille I tried it on my custom built triple z axis printer and I can say that this is definitely useful functionality. Requires some polishing and optimisations, but it is already better than manually finding x/y offsets.

@tntclaus Thanks for testing and the feedback, greatly appreciated.

> 
> What I didn't understand is `z_offset`. FWIW including z coordinate into `z_positions` list to indicate how far below table mount point is might be better.

z_offset belongs to the Z_TILT_CALIBRATE feature, which is independent of Z_TILT_AUTODETECT. It gives the offsets (or, the bed height errors) of the probing points and has nothing to do with the mount points. Currently the way to compensate for an uneven bed is to increase the number of points to probe. My idea here is to instead measure the bed errors in the 3 probing points and take them into account in the calculation. So you measure with many points, save the z_offsets, and from then on you get the same leveling as you would get from many points, but with only 3 points. This speeds up the Z_TILT_ADJUST process.

I could make those offsets part of the probe point definition, but as the points are configured manually, while the z_offsets are determined automatically I kept them separate.

tntclaus:
@sensille now I see. Thanks for explanation and for your work :) 

I can say that this works as intended on my printer.

KevinOConnor:

-Kevin

sensille:
According to the feedback I need to rework the configuration and improve the algorithm. As I currently don't have much time for this, if you prefer I can retract the PR until I got around to it.

arne182:
Hi @sensille, 

I should be soon building a printer that can use this functionality and could give you a hand with the code.

-Arne

KevinOConnor:
Okay, I will leave the "pending feedback" label on this then.

-Kevin

arne182:
ETA of my kit is next week. So it might take me a few days/weeks to build till I can test this. But will get to this.

PR #4089, opened 2021-03-21
Add hardware PWM support for STM32F1 and LPC1769x

1am3r:
This adds hardware PWM support for STM32F1 and LPC176x processors.

This should also work for STM32 F0/F2/F4 processors, but I don't have one of those to test it.

Signed-off-by: Michael Kurz <michi.kurz@gmail.com>

### DIFF
--- a/src/lpc176x/Kconfig
+++ b/src/lpc176x/Kconfig
@@ -12,6 +12,7 @@ config LPC_SELECT
     select HAVE_GPIO_BITBANGING
     select HAVE_STRICT_TIMING
     select HAVE_CHIPID
+    select HAVE_GPIO_HARD_PWM
 
 config BOARD_DIRECTORY
     string
--- a/src/lpc176x/Makefile
+++ b/src/lpc176x/Makefile
@@ -22,6 +22,7 @@ src-$(CONFIG_HAVE_GPIO_SPI) += lpc176x/spi.c
 src-$(CONFIG_USBSERIAL) += lpc176x/usbserial.c lpc176x/chipid.c
 src-$(CONFIG_USBSERIAL) += generic/usb_cdc.c
 src-$(CONFIG_SERIAL) += lpc176x/serial.c generic/serial_irq.c
+src-$(CONFIG_HAVE_GPIO_HARD_PWM) += lpc176x/hard_pwm.c
 
 # Build the additional bin output file
 target-y += $(OUT)klipper.bin
--- a/src/lpc176x/gpio.h
+++ b/src/lpc176x/gpio.h
@@ -21,6 +21,13 @@ struct gpio_in gpio_in_setup(uint8_t pin, int8_t pull_up);
 void gpio_in_reset(struct gpio_in g, int8_t pull_up);
 uint8_t gpio_in_read(struct gpio_in g);
 
+struct gpio_pwm {
+    void *reg;
+    uint8_t channel;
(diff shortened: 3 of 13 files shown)


KevinOConnor:
Thanks.  In general, it looks good to me.  However, the change to  lib/lpc176x/device/system_LPC17xx.c does not look correct to me - changes to the clock should be made via `enable_pclock()` - and the new code does that, so I don't think the change to the lib code is needed.

Also, if you could add the module name to each commit that will help me merge these changes.  The first line of the commit should start with "stm32: " (on the first commit) and "lpc176x: " (on the second commit).  See https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

1am3r:
The errata referenced in [system_LPC17xx.c#L523](https://github.com/KevinOConnor/klipper/blob/master/lib/lpc176x/device/system_LPC17xx.c#L523) mentions that changing the  PCLKSELx registers may not take effect if PLL0 is already enabled and connected. 
At least for the timer clock on my LPC1769 this was the case. I think the only other workaround for this would be to switch clocks, disable the PLL, set PCLKSELx registers, re-enable and reconnect the PLL and switch clocks back in enable_pclock(). I can add this, but I'm not sure if this is a better way.

Sorry for the missing 'module'. I totally overlooked it in CONTRIBUTING.md.

KevinOConnor:
Ah, interesting.  I'm not sure why we haven't run into that errata before.

However, I think we should handle that problem separately from this PR.  I think we should be able to fix that problem by using the default CLOCK_FREQ/4 for the peripherals - then we don't have to change PCLKSELx at runtime.  I put up a test branch for that on a new work-lpc176x-20210324 branch. `cd ~/klipper ; git fetch ; git checkout origin/work-lpc176x-20210324 ; sudo service klipper stop ; make flash ; sudo service klipper start`

-Kevin

1am3r:
Ok, so I'll change it here to work with CLOCK_FREQ/4 for now, and we'll handle the clock changing in the work-lpc176x-20210324 branch, or should I rebase on the work-lpc176x-20210324 and we'll merge this after the clock change is solved?

KevinOConnor:
I went ahead and merged the work-lpc176x-20210324 into the master branch.  So, you can just rebase this branch to master.

Note that both stm32 and lpc176x likely need to determine the ratio between peripheral clock frequency and system clock frequency when programming the cycle time.  Maybe something like `clock_div = CONFIG_CLOCK_FREQ / get_pclock_frequency(...)`.

-Kevin

1am3r:
I added the clock divider like you suggested. For STM32 i had to add the `get_timer_frequency()` as the timer frequency is double the peripheral frequency for F1, F2 and F4.

KevinOConnor:
Thanks.  I committed this, but with slightly different pclock calculation code.  (Commits 39188e1c and 59c4c498.)

-Kevin

Ramalama2:
@1am3r 
Could you add support for stm32 f4?
I can donate you the board if you want 🙈

1am3r:
@Ramalama2 I got a F4 based board, just need to find some time to implement it...

Ramalama2:
> @Ramalama2 I got a F4 based board, just need to find some time to implement it...

Wow amazing 🙈
Take your time and thanks very much!

PR #4091, opened 2021-03-22
Added BLTouch and Z Probe section for this board (both tested and working)

Nume1977:
Added BLTouch and Z Probe section for this board (both tested and working)


KevinOConnor:
Thanks, but I don't think it makes sense to add this to the example "board" config files.  There are too many possible customizations to list in the example configs - the main docs/Config_Reference.md file is intended to be the reference in this case.

-Kevin

KevinOConnor:

-Kevin

PR #4094, opened 2021-03-22
New Discourse server as Klipper community forum

KevinOConnor:
This is a proposal to introduce a Discourse server dedicated to Klipper.  The goal would be to use it as a "Community forum".  This is a follow up to PR #3762 - in that PR the Contact.md document was updated to point users from github to a Discord chat server.  This PR would update the Contact.md file to point users at either the new "community forum" or the "discord chat" server.

The new Discourse server is at: https://community.klipper3d.org/ .  Feel free to use that server now.  This is an experiment though - it may not work out over the long term.  The proposed Contact.md file is at: https://github.com/KevinOConnor/klipper/blob/work-contact-20210320/docs/Contact.md

My impression on the results of PR #3762 is generally positive.  I feel it has helped notably with reducing the amount of "please help me" requests on github, and my impression is that the number of contributors on github has increased (both relatively speaking and in absolute terms).

However, I think there are some areas where Discord chat does not do well.  Specifically:
1. I fear Discord may not be a good option for non-native english speakers.
2. I fear Discord can be a poor option for some international users - they may be active when other knowledgeable users are not active.  Given the volume of messages on the server, if a conversation isn't achieved in about a half hour, I fear it may never get resolved.
3. I fear Discord can be overwhelmed by people asking the same question.
4. A forum may be helpful to share macros/configs/experiences.

None of the above will be a surprise to many.  Still, I think it was worth exploring github+Discord, and my thinking right now is that github+Discourse+Discord is also worth exploring.  Thus this PR.

To be clear, I do think the Discord chat server is still a very effective tool.  I've seen many useful conversations, and I do think the real-time nature of communication on Discord is an asset.  This experiment with Discourse is intended to complement Discord - not replace it.

Comments?
-Kevin

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -3,6 +3,9 @@ contact_links:
   - name: Directions
     url: https://www.klipper3d.org/Contact.html
     about: Have a question? Need help? Start here.
+  - name: Klipper Forum
+    url: https://community.klipper3d.org/
+    about: Klipper community Discourse server.
   - name: Klipper Chat
     url: https://discord.klipper3d.org/
     about: Klipper community Discord chat server.
--- a/docs/Contact.md
+++ b/docs/Contact.md
@@ -1,21 +1,41 @@
 This document provides contact information for Klipper.
 
-1. [I have a question about Klipper](#i-have-a-question-about-klipper)
-2. [I have a feature request](#i-have-a-feature-request)
-3. [Help! It doesn't work!](#help-it-doesnt-work)
-4. [I would like to share the findings of work that I have done](#i-would-like-to-share-the-findings-of-work-that-i-have-done)
-5. [I found a bug in Klipper](#i-found-a-bug-in-klipper)
-6. [I am making changes that I'd like to include in Klipper](#i-am-making-changes-that-id-like-to-include-in-klipper)
-7. [Klipper Community Discord](#discord)
-8. [Klipper mailing list](#mailing-list)
+1. [Community Forum](#community-forum)
+2. [Discord Chat](#discord-chat)
+3. [I have a question about Klipper](#i-have-a-question-about-klipper)
+4. [I have a feature request](#i-have-a-feature-request)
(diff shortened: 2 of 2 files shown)


Sineos:
I think this is and will be a great asset. 👍 
What about a dedicated "Support" Category?

KevinOConnor:
> What about a dedicated "Support" Category?

I'm leery of making a "Support" or "Help" category, as I'd like to emphasize that it is a community server where everyone is expected to contribute.  Everyone gets stuck sometimes and it makes total sense to be able to reach out to get assistance.  But, I think there should be an expectation that those asking for help will have done their own research first and be willing to share that research - both during their query and with others facing similar issues.

The "general discussion" category is the intended place to start in that case.  Let me know if there is a better way to word that.

-Kevin

theophile:
I'm in favor of the forum, as I find it to be a much more helpful format for finding and sharing information than live chat. But what about using the GitHub Discussions feature? It appears to exist for the very purpose described here. Also, "issues" filed in the wrong place can be moved into Discussions (https://docs.github.com/en/discussions/managing-discussions-for-your-community/managing-discussions-in-your-repository#transferring-a-discussion), apparently automatically based on tags. So if users do mispost in "issues," then it could be tagged as "invalid" and then automatically moved into Discussions where the discussion could continue, rather than the poster being redirected somewhere else.

KevinOConnor:
> But what about using the GitHub Discussions feature?

I have reservations about using "github discussions".  I gave some background for this at https://github.com/KevinOConnor/klipper/pull/3762#issuecomment-759586579 .

Github discussions is not something I will pursue at this time.

-Kevin

theophile:
> I have reservations about using "github discussions". I gave some background for this at [#3762 (comment)](https://github.com/KevinOConnor/klipper/pull/3762#issuecomment-759586579) .
> 

I'm sorry, I missed that (obviously). But I think you're right. I hadn't considered that there might be Klipper users uncomfortable with github.

Sineos:
The announcement of the new Discourse and the general policy might be worth a "Pinned Issue"

PR #4096, opened 2021-03-23
Added config for Eryone Thinker Series

dapostol73:
Would like to add this config for others to use for an Eryone Thinker S.  If this needs to be integrated to another branch, please let me know...thanks.


KevinOConnor:
Thanks, but please see the guidelines (just recently added) for example config files: https://www.klipper3d.org/Example_Configs.html

Also, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

dapostol73:
Hey Kevin,

Thanks for the documentation and feedback, I believe I have resolved all issues, please let me know if you have any feedback.

Also, the automation tests are a great help!

-Demetrius

config/printer-eryone-thinker-series-v2.cfg
```diff
@@ -0,0 +1,225 @@
...
+max_accel: 1500
+max_z_velocity: 5
+max_z_accel: 150
+
+[pause_resume]
+
+[virtual_sdcard]
+path: ~/.octoprint/uploads/
+
+[respond]
```
KevinOConnor, line 104: Thanks.  However, these changes aren't specific to the printer - they're specific to your host setup.  I don't think they should be added to the example config files (see docs/Example_Configs.md).

config/printer-eryone-thinker-series-v2.cfg
```diff
@@ -0,0 +1,225 @@
...
+    G90
+    G1 X50 Y0 F3000
+    G91
+    G1 E-5 F300
+
+[gcode_macro CHANGE_FILAMENT_END]
+gcode:
+    G91
+    RESTORE_GCODE_STATE NAME=M600_state
+    RESUME
```
KevinOConnor, line 135: Similarly, unless I'm missing something, these macros don't look like they are specific to the printer.

config/printer-eryone-thinker-series-v2.cfg
```diff
@@ -0,0 +1,225 @@
...
+# Use a default 1kHz tone if S is omitted.
+default_parameter_P: 100
+# Use a 10ms duration is P is omitted.
+gcode:
+    SET_PIN PIN=beeper VALUE={S}
+    G4 P{P}
+    SET_PIN PIN=beeper VALUE=0
+
+[gcode_macro M205]
+gcode:
```
KevinOConnor, line 190: Similar comment.

config/printer-eryone-thinker-series-v2.cfg
dapostol73: Sure I can remove these.

config/printer-eryone-thinker-series-v2.cfg
dapostol73: No they are not, but their is feedback from the community that they would like these to be easier for none technical people to not have figure this out on their own....is it cool to have these commented out, similar to the BL Touch settings?

config/printer-eryone-thinker-series-v2.cfg
dapostol73: Can I have them commented out for user?  Again the community feedback is when they try Klipper their is a lot of functionality that doesn't work out of the box for them.  Ideally these would be part of the core.

config/printer-eryone-thinker-series-v2.cfg
dapostol73: I'll remove them and add example to their Facebook Group

KevinOConnor:
Thanks.  The config looks fine, but the filename should end in a year in it (as described at https://www.klipper3d.org/Example_Configs.html ).

> No they are not, but their is feedback from the community that they would like these to be easier for none technical people to not have figure this out on their own....is it cool to have these commented out, similar to the BL Touch settings?

FWIW, we recently added a section to the Klipper Discourse server for sharing custom configs ( https://community.klipper3d.org/ ).  So, if you want to add the basic config here and the advanced config on Discourse that would be fine.

Cheers,
-Kevin

dapostol73:
Thank you, for all your hard work...really enjoying Klipper

PR #4098, opened 2021-03-23
Fix extruder1 step_pin in Robin Nano V2 example

hakimio:
extruder1 `step_pin` was set incorrectly in MKS Robin Nano V2 example config.
Complete pin map can be found [here](https://github.com/makerbase-mks/MKS-Robin-Nano-V2.X/wiki/Wiring_Pinout_and_Size#pin-map).


KevinOConnor:
Thanks.

-Kevin

PR #4114, opened 2021-03-28
Palette2, add ping variation + ping display

cmroche:
Couple of minor additions to make the to the **Palette2** module:

- After a bit of testing with other users, 115200 is a more reliable port speed so changed the baud default from 250000 to 115200.
- Improvements to the PALETTE_CONNECT command means it will now check and verify a successful connection and report an error on failure. Previously it wouldn't say anything until the heartbeat timeout was reached.
- Improvements to the heartbeat timeout process, the Palette 2 doesn't always respond to heartbeats during the setup process, so while a timeout is still recorded it will not abort the the print while Palette setup is happening.
- Add support for auto cancel on ping variation. If a ping is over the configured threshold the print will be cancelled as this typically indicates a jam or broken splice.
- Add support for ping display. When enabled the ping percentage will be sent to the printer display using M117.


KevinOConnor:
Thanks.  I'll commit if you wish.  WRT, to M117 change though, I think it would be more flexible if the palette2.py module introduced a `get_status()` method that exported its statistics.  Then users could modify the display to their liking (see klippy/display/display.cfg for examples).

-Kevin

cmroche:
@KevinOConnor Thanks for the info. I agree, the `get_status()` approach sounds like the best way to go. I'll make the changes this weekend and ping you to let you know when this PR is ready again, but I think we should hold off a merge until I can make that change.

cmroche:
@KevinOConnor Should be good to go now.

klippy/extras/palette2.py
```diff
@@ -215,6 +218,20 @@ def cmd_OmegaDefault(self, gcmd):
...
             self.write_queue.put(gcmd.get_commandline())
 
+    def _wait_for_heartbeat(self):
+        startTs = self.reactor.monotonic()
+        while self.heartbeat is None and self.heartbeat < (
+                self.reactor.monotonic() -
+                SETUP_TIMEOUT) and startTs > (
+                self.reactor.monotonic() -
+                SETUP_TIMEOUT):
+            self.reactor.pause(1)
```
KevinOConnor, line 228: I know it's not related to this patch, but in review I noticed at error.  The `pause()` method takes an absolute time not a duration.  So, it should be something like `self.reactor.pause(curtime + 1.)`.

Also, as a minor note, for efficiency reasons it's possible to obtain a new time directly from `pause()` - as in `curtime = self.reactor.pause(curtime + 1.)`

klippy/extras/palette2.py
```diff
@@ -618,6 +637,13 @@ def _run_Smart_Load(self, eventtime):
             return self.reactor.NOW
         return eventtime + AUTOLOAD_TIMER
 
+    def get_status(self, eventtime=None):
+        status = {
+            "ping": self.omega_pings[-1],
+            "remaining_load_length": self.remaining_load_length,
+            "is_splicing": self.is_splicing
+        }
+        return status
```
KevinOConnor, line 646: When adding a `get_status()` method, please also update the docs in docs/Command_Template.md .

KevinOConnor (commented):
Thanks.  At a high-level the changes look okay to me.  I did notice a couple of minor things when reviewing though - see below.

-Kevin

klippy/extras/palette2.py
```diff
@@ -563,12 +579,14 @@ def _run_Read(self, eventtime):
 
     def _run_Heartbeat(self, eventtime):
         self.write_queue.put(COMMAND_HEARTBEAT)
+        self.reactor.pause(5)
```
KevinOConnor, line 582: Incorrect usage of `reactor.pause()` here.

-Kevin

cmroche: Should be the last one.

KevinOConnor:
Thanks.

-Kevin

cfelicio:
I see the get_status supports latest ping, are there any plans to support the other info currently displayed by the octoprint plugins? Example:

Splice Info (e.g. Current Splice)
Filament Used
Pong

Thanks!

PR #2369, opened 2020-01-11
FFCP LCD support

fragginright:

PR #4121, opened 2021-03-28
added lcd support to the mightyboard

le-Bark:
The mightyboard use a shift register between the mcu and lcd to save pins.

I added an option to choose between the standard pins configuration and the shift_register.
If no option is specified, the 4-bit interfaced is used.

I also tested to make sure I didn't break the default option.


KevinOConnor:
FYI, there was a similar PR at #2369 .  My comments here would be similar to my comments on that PR.  Could this be done using the software spi system instead of introducing another "bit banging" interface in the micro-controller code?

-Kevin

le-Bark:
Hi Kevin, thanks for reviewing my PR. 

I looked a little and found that the miso pin is not optional on the software spi extra.
Is there a way to not use that pin? If not, I would need to specify a dummy pin witch there is not many. 
Also I am not sure what the mode parameter does.

-Marc-André

KevinOConnor:
Yes - using software spi would currently require specifying a dummy miso pin.  We could change that though (this was also discussed a bit at  #3979).  The mode pin determines when to read wrt writes - I don't think it would matter if you don't read.

-Kevin

le-Bark:
OK, I will specify a dummy pin for now and maybe modify the software spi in the future.

-Marc-André

le-Bark:
To initialize the lcd correctly, I need to generate dead time between commands. I don't see any way of doing this in python. Also, the spi software on the avr does not seem to use the rate parameter. 
How should I generate theses delays?

-Marc-André

KevinOConnor:
Thanks.  At a high-level the changes looks okay to me.  I did notice a couple of things though:
1. It would be preferable if you could separate out the example config file to its own PR.  Also, there are some new guidelines for the example configs at: https://www.klipper3d.org/Example_Configs.html
2. Unfortunately, it's not valid to call `sleep()` in Klipper.  Instead, to obtain a delay for display init, one would use the `minclock` parameter to ensure the transmitted messages are held until the appropriate time.  Note that you can't reliably time 5ms with minclock - so we use much larger delays of 100ms even when only a small delay is needed.  It looks like the code already has support for calculating minclock, but the `minclock` parameter needs to be propagated to the ultimate `spi_send()` command.

-Kevin

le-Bark:
I removed the flashforge-creator-pro config file and added the minclock delay.
This should be good now.

-Marc-André

config/generic-mightyboard.cfg
```diff
@@ -117,3 +117,16 @@ scl_pin: PJ5
...
 wiper: 0.50
 scale: 0.773
+
+[display]
+lcd_type: hd44780_spi
+spi_software_mosi_pin: PC3
+spi_software_sclk_pin: PC2
+#miso not used, dummy pin.
+spi_software_miso_pin: PJ1
+latch_pin: PC4
```
KevinOConnor, line 127: In general, it looks fine, but it does look like a `latch_pin` config option was added and not documented.

-Kevin

le-Bark: I added the documentation.

KevinOConnor:
Thanks.

-Kevin

Issue #797, opened 2018-10-19
[FR] SET_NOZZLE_SIZE 


Hywelmartin:
I haven't found any way(other than edit printer.cfg) to alter nozzle size

So my suggestion is to have a extended gcode like:
 SET_NOZZLE_SIZE 0.4
that could be a part of start gcode or just punched in in the octopi terminal

And by the way GREAT FW...

/Martin

FHeilmann:
Nozzle size is not something the printer is aware of. This is entirely handled by the slicer. Which functionality do you expect to achieve with this command?

FHeilmann:
It's used for a security feature where the firmware rejects moves that extrude more material than the hotend can handle. Unless you are running into issues where your printer rejects gcode due to too much filament extruded you can leave this value at 0.4 and simply ignore it.

Hywelmartin:
confusing..
The volumetric flow is not dependent on the nozzle size.... 

To be able to know the volumetric flow you need the steps/mm from the extruder and the filament diameter... not the nozzle size... what you put in shall get out... if the fw depend on the nozzle size to determent the volumetric flow there is at least one calculation to many...

I switch nozzles rather frequently  so I taught that it would be beneficial to be able to adjusting the parameters in another way... So I can put at a arbitrary value and forget it.... good to know..

/Martin

Hywelmartin:
well I'm not a coder so ... i don't get anything out of this... well something

but my point stands if it's a useful parameter that we would be able to adjust it by extended gcode..

feel free to evaluate and take the next step..

jakep82:
As @FHeilmann noted this is only a safety feature.  If you're going to change nozzles, it's not unreasonable to also update your config, so this doesn't need to be run-time configurable.  I just leave mine at 0.5 which is the largest nozzle I ever use, and don't bother changing it.

faboaic:
I often print with 0.4mm nozzle diameter but extrusion width 0.8 - 1.2 mm. 
So this safety feature just ruined a print of mine. 

Now I set nozzle diameter in printer.cfg to 0.8mm and see what happens...

FHeilmann:
your use case is what the `max_extrude_cross_section` configuration parameter is for. By default it is `4*nozzle diameter^2` which allows the equivalent of .8 extrusion width for a .4mm nozzle.

[...]

PR #4123, opened 2021-03-29
extruder: Add SET_EXTRUDER g-code

clinta:
As discussed in #797, being able to set the nozzle_diameter via gcode is useful. Because of how inter-related nozzle_diameter, filament_diameter, max_cross_section, max_e_velocity and max_e_accel are, this change adds a SET_EXTRUDER command that can set any of these parameters.

The last configured value for max_cross_section, max_e_velocity and max_e_accel are persisted so that invocations of the gcode can know if these parameters were explicitly configured, or automatically calculated. If they were explicitly configured, and not specified in the g-code invocation, they will not be changed. If they were not configured, or are set to 0 by g-code, they will be automatically calculated.


KevinOConnor:
Thanks.  However, I don't think this is a good candidate for the main Klipper repository.  I don't think we want to add commands that change the configured safety checks.  If a user needs to change them, I think it's reasonable for them to update the config file and issue a RESTART command.

-Kevin

ianespana:
This would be extremely useful, especially considering different filaments have different requirements and would make sense to allow the slicer to change some settings based on your filament selection

KevinOConnor:

-Kevin

miklschmidt:
@KevinOConnor I actually think this is a significant usability issue currently. Users are used to dealing with nozzle_diameter in their slicers, and they don't understand why klipper throws a fit when they change nozzles. What they've started doing is leaving their nozzle_diameter at an arbitrary high number so they don't need to worry about it, making this setting unreliable in macros. Adjusting nozzle_diameter during runtime would make nozzle_diameter actually useful. For example to automatically adjust prime line macros to use the correct flow. Nozzle_diameter is available as a parameter in every major slicer, so it would be easy to pass this to a START_PRINT macro. Are you still against adjusting nozzle_diameter during runtime, or was it more because this pull request tried to do too much?

sprior:
As a relatively new Klipper user I also agree that this feature would be useful.  With modern hot ends nozzle swaps are quite trivial and don't even require the machine to be electrically restarted.  I was doing a lot of nozzle swaps recently and then realized that I had done several successful prints with a 0.4 nozzle and nozzle_diameter still set to 0.8 and wondered why I hadn't had an issue.  With the Voron 2.4 every time I do a firmware restart I have to go back and do a G28, QUAD_GANTRY_LEVEL, CALIBRATE_Z just to keep going, so if this could have been handled with an extended gcode and not required a firmware restart that would have been a nice win.

matti125:
I would also vote for adding the possibility to dynamically set at least the nozzle size, possibly with a non-dynamically-settable max_nozzle_size, which would work as the real safety limit. Right now I set the nozzle size to the max I use, which means that I  loose some of the klipper safety checks that are based on nozzle size.

kubark42:
I also vote for this feature. Setting the nozzle diameter in multiple places violates the principle of least surprise, and as such we are seeing uses circumvent the safety features because for them the safety features cause more harm than good. This could be addressed by allowing the slicer to inform Klipper about the nozzle.

The user could be expected to properly configure Klipper with the list of nozzles which are potentially installed. If a new nozzle is requested, Klipper could throw an error which explains what the problem is and how to fix it.

Fundamentally, I think we're moving to a world where nozzles are going to be swapped with the same frequency and ease as materials. Lumping the nozzle in with more permanent, inviolate machine configurations-- such as gantry size, extruder gear ratio, endstop locations, etc...-- has become a bit of a misfit.

sprior:
If you look at https://github.com/Klipper3d/klipper/issues/797 mentioned above it seems that the nozzle size is just used for one safety check feature and is otherwise unimportant.  So @kubark42 I completely agree with you that nozzles can be swapped frequently, but it's also ok in this case to just leave this setting at the max nozzle size you'll ever use and ignore it.

kubark42:
@clinta, that's exactly right. If the safety check is valuable, then it needs to work. If it's not valuable, then it should be removed. I'm not sure if it's well appreciated just how much negative value it makes to teach people to ignore and circumvent Klipper's safety algorithms.

@sprior maybe it's workable to simply rename the variable `max_nozzle_dia` and include a helpful comment that this should be set to the max nozzle size which will be used on the machine.

sprior:
@kubark42 I'm not a Klipper dev so I wouldn't be involved in the variable rename.  If you look up in the comments I'm just a user who made the same points as you back in March and am now just suggesting a way forward for you.

TheBurnDoc:
I understand the rationale behind not making a safety feature configurable @KevinOConnor but the `nozzle_diameter` is not a safety feature, `max_extrude_cross_section` is the safety feature. `nozzle_diameter` is a feature of the haredware being used - and as several folks have observed, in the age of Revo, nozzle swaps are a trivial now. 

Given the newer workflow the Revo facilitates, updating the configuration file each time the nozzle changes now seems a tad archaic - granted you can just set the \nozzle_diameter\ to the largest one you use (as I have done), but this now technically invalidates the `max_extrude_cross_section` safety feature for when swapping back to smaller nozzles.

For me I would much rather have the ability to updates the nozzle size in my PRINT_START, it doesn't need to be an extended G-Code, but perhaps simple a read/write status variable.  Eg in PrusaSlicer I might add the following line:

```
{% set printer.nozzle_diameter = [nozzle_diameter[0]] %}
```

This would implicitly set the safety feature to an appropriate value for each print.

stefanfoulis:
With Revo I'm switching nozzles so often that a new problem arose: forgetting to swap the nozzle to the correct size when starting a new print.

Wild idea: have the nozzle cam detect the currently installed nozzle size based on the color of the nozzles silicon sleeve at the beginning of the print and set the nozzle size with `SET_NOZZLE_DIAMETER` - automatically getting the correct safety features.
Then have a macro complain at the beginning of the print if the nozzle size the slicer is expecting doesn't match the actual one installed in the printer.

PR #4124, opened 2021-03-30
Faster "pinup" on BL-Touch

KevinOConnor:
When probing, the current bltouch code will detect a trigger signal, acknowledge the mcu homing event, and then send a pin_up command.  The "mcu acknowledge" stage may require a few messages to be sent to the micro-controller which normally takes a few milliseconds.  However, in the event of communication errors or delays, it's possible for that process to delay the pin_up command.  As we work on multi-mcu homing (PR #3956) the "mcu acknowledgement" becomes more complex and it increases the risk of a delay.

This PR alters the bltouch code so that it can send the pin_up immediately after a trigger signal.  This should reduce the risk of a delayed pin_up command.

I don't actually own a bltouch, so I can't directly test this code.  Testers would be welcome.  Basic verification that probing continues to work (in both stow_on_retract=False and stow_on_rectract=True modes) should be sufficient.

`cd ~/klipper ; git fetch ; git checkout origin/work-bltouch-20210329 ; sudo service klipper restart`

@FanDjango - FYI.

-Kevin

### DIFF
--- a/klippy/extras/bltouch.py
+++ b/klippy/extras/bltouch.py
@@ -38,7 +38,9 @@ class BLTouchEndstopWrapper:
         self.mcu_pwm = ppins.setup_pin('pwm', config.get('control_pin'))
         self.mcu_pwm.setup_max_duration(0.)
         self.mcu_pwm.setup_cycle_time(SIGNAL_PERIOD)
+        # Command timing
         self.next_cmd_time = self.action_end_time = 0.
+        self.finish_home_complete = self.wait_trigger_complete = None
         # Create an "endstop" object to handle the sensor pin
         pin = config.get('sensor_pin')
         pin_params = ppins.lookup_pin(pin, can_invert=True, can_pullup=True)
@@ -82,6 +84,7 @@ class BLTouchEndstopWrapper:
         self.set_output_mode(self.output_mode)
         try:
             self.raise_probe()
+            self.verify_raise_probe()
         except self.printer.command_error as e:
             logging.warning("BLTouch raise probe error: %s", str(e))
     def sync_mcu_print_time(self):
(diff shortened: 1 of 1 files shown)


FanDjango:
Thanks for addressing this, Kevin. 

It closes the aperture for error state on BLTouch probes and the main observed difference to Marlin, the 2nd being the non-continuous PWM (which has been discussed and is not an issue).

After introduction of SW-Mode in klipper, there was a way to avoid the error state when unable to **either** raise the probe **or** stow the pin within about 700ms due to inner delays. SW-Mode has no max-time when triggered.

But this change is very positive nonetheless as you intend to add multi-mcu homing. Do not want to have only the choice of using SW-Mode, it is nice to be able to use the native mode and not have the delay glitches.

Unfortunately I won't be able to test this for you for another few weeks as the (klipper) printers are in the office and I won't be going in for some time.

I dare say that any one of the current BLTouch users who test this should be able to see an immediate difference in homing/probing behaviour.

jnimety:
I was able to confirm that my bltouch 3.1 operates as expected for both stow_on_each_sample True and False. Probing during G28, SCREWS_TILT_CALCULATE, and PROBE_ACCURACY all work as expected.

KevinOConnor:
Okay, thanks.  I went ahead and committed this change.

FYI, I don't think this will be noticeable to most users.  I expect the response time to now be around ~102ms instead of ~110ms.  The main goal is to prevent that time from further increasing with future changes and to make it less likely comms jitter impacts the timing.

-Kevin

jnimety:
I'll be honest, I didn't notice any speed changes.

FanDjango:
> I dare say that any one of the current BLTouch users who test this should be able to see an immediate difference in homing/probing behaviour.

So disregard my statement above. I stand corrected. 

I **was** actually hoping for a discernable improvement.

I  understand now that we are getting an "upper limit" on possible delays due to the factors Kevin mentioned.

PR #4128, opened 2021-03-31
sync channel for fast move_queue population

Cirromulus:
This PR adds the possibility to update SW/HW PWM faster than the original implementation (0.100s).
This would close https://github.com/KevinOConnor/klipper/issues/133, enabling laser raster engraving.
While it will probably not reach the stepper step-times (which are compressed), I suppose it may come close to some multiples of `cycle_time` in bursts, depending on the serial bandwidth to the MCU.
Early testers report around 400-800 updates per second, peaks higher (as the queue is buffered).

### Known Issues
  - Code seems to work on some different hardware setups. *More testers appreciated!*

If you want to try this out, you may follow this [small HowTo](https://github.com/Cirromulus/klipper/blob/pwm_sync_channel/docs/Using_PWM_Tools.md).

### How it works
This PR adds a "sync_channel" (name may be subject to change) to the steppersync/stepcompress system.
Basically, it allows a command to be inserted into the mcu's `move_queue` while respecting the watermark and timing.
This enables much tighter packing of pin-update commands.
To ensure the steppersync queues are flushed regularly even when the toolhead is not generating any movement,
a combination of `note_kinematic_activity()`, `update_move_time()` and `check_stall()` is used.

### Other Notes
PWM Frequency: I suggest to not mean too well with the tool frequency. Choose the lowest frequency that still has a pulse time of a small multiple of your highest resolution. E.g. Resolution of 0.1mm at a speed of 50mm/s I recommend 1kHz (cycle_time: 0.001)

Hardware-PWM may be less calculation intense on the MCU, but the MCU will choose in "best effort" a frequency that is near the desired frequency.
The maximum pin update-rate is this chosen frequency.
Software-PWM will be less precise, but may offer more flexibility at odd frequencies.
So if you plan to have an update rate of 1000 updates / s, choose at least a frequency of 1kHz, thus a cycle_time of 0.001.

### Unaddressed features, Roadmap
1. This PR wraps the "ht-mode" (high throughput) at the command wrapper, reducing the impact on `MCU_pwm` code.
In the state of this PR, one pin can be put in ht-mode. In long term (another PR), I plan to rework the stepcompress system, dividing the two functionalities (1. compressing steps, 2. making sure the move queue is populated and ordered), so that the PWM updates don't need to "piggy back" on kinematics-stepper update flushes. This would enable even faster pinupdates and, most importantly, any number of ht-mode pins.

2. This PR only adds the possibility to update a pin much more frequently.
For laser applications, a _proportionality feature_ is useful. This would dim the laser during acceleration and deceleration phase proportional to its actual speed (see also *S*-Parameter in [G0 in RepRap documentation](https://reprap.org/wiki/G-code#G-commands)), just like it does with the extruder *E* command.

### DIFF
--- a/klippy/extras/output_pin.py
+++ b/klippy/extras/output_pin.py
@@ -63,7 +63,7 @@ class PrinterOutputPin:
         return {'value': self.last_value}
     def _set_pin(self, print_time, value, cycle_time, is_resend=False):
         if value == self.last_value and cycle_time == self.last_cycle_time:
-            if not is_resend:
+            if not is_resend or value == self.shutdown_value:
                 return
         print_time = max(print_time, self.last_print_time + self._pin_min_time)
         if self.is_pwm:
--- a/klippy/extras/output_pin.py
+++ b/klippy/extras/output_pin.py
@@ -4,6 +4,7 @@
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 
+PIN_MIN_TIME = 0.100
 RESEND_HOST_TIME = 0.300
 MAX_SCHEDULE_TIME = 5.0
 
@@ -12,6 +13,7 @@ class PrinterOutputPin:
         self.printer = config.get_printer()
         ppins = self.printer.lookup_object('pins')
         self.is_pwm = config.getboolean('pwm', False)
+        self._pin_min_time = PIN_MIN_TIME
         if self.is_pwm:
             self.mcu_pin = ppins.setup_pin('pwm', config.get('pin'))
             cycle_time = config.getfloat('cycle_time', 0.100, above=0.,
@@ -38,6 +40,7 @@ class PrinterOutputPin:
         else:
             if config.getboolean('high_throughput', False):
                 self.mcu_pin.setup_high_throughput_mode()
+                self._pin_min_time = self.default_cycle_time
 
             max_mcu_duration = config.getfloat('maximum_mcu_duration', 0.,
                                                minval=0.500,
@@ -62,7 +65,7 @@ class PrinterOutputPin:
         if value == self.last_value and cycle_time == self.last_cycle_time:
             if not is_resend:
                 return
-        print_time = max(print_time, self.last_print_time + cycle_time)
+        print_time = max(print_time, self.last_print_time + self._pin_min_time)
         if self.is_pwm:
             self.mcu_pin.set_pwm(print_time, value, cycle_time)
         else:
--- a/klippy/extras/output_pin.py
+++ b/klippy/extras/output_pin.py
@@ -4,8 +4,7 @@
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 
-PIN_MIN_TIME = 0.100
-RESEND_HOST_TIME = 0.300 + PIN_MIN_TIME
+RESEND_HOST_TIME = 0.300
 MAX_SCHEDULE_TIME = 5.0
 
 class PrinterOutputPin:
@@ -37,6 +36,9 @@ class PrinterOutputPin:
             self.mcu_pin.setup_start_value(
                 self.last_value, self.last_value, True)
         else:
+            if config.getboolean('high_throughput', False):
+                self.mcu_pin.setup_high_throughput_mode()
+
             max_mcu_duration = config.getfloat('maximum_mcu_duration', 0.,
                                                minval=0.500,
                                                maxval=MAX_SCHEDULE_TIME)
@@ -60,7 +62,7 @@ class PrinterOutputPin:
         if value == self.last_value and cycle_time == self.last_cycle_time:
             if not is_resend:
                 return
-        print_time = max(print_time, self.last_print_time + PIN_MIN_TIME)
+        print_time = max(print_time, self.last_print_time + cycle_time)
         if self.is_pwm:
             self.mcu_pin.set_pwm(print_time, value, cycle_time)
         else:
@@ -92,11 +94,12 @@ class PrinterOutputPin:
         systime = self.reactor.monotonic()
         print_time = self.mcu_pin.get_mcu().estimated_print_time(systime)
         time_diff = (self.last_print_time + self.resend_interval) - print_time
+
         if time_diff > 0.:
             # Reschedule for resend time
             return systime + time_diff
-        self._set_pin(print_time + PIN_MIN_TIME,
-                      self.last_value, self.last_cycle_time, True)
+        self._set_pin(print_time + RESEND_HOST_TIME/2, self.last_value,
+                      self.last_cycle_time, True)
         return systime + self.resend_interval
 
 def load_config_prefix(config):
--- a/klippy/mcu.py
+++ b/klippy/mcu.py
@@ -378,11 +378,13 @@ class MCU_pwm:
         self._mcu.register_config_callback(self._build_config)
         self._pin = pin_params['pin']
         self._invert = pin_params['invert']
+        self._is_ht = False
         self._start_value = self._shutdown_value = float(self._invert)
         self._is_static = False
         self._last_clock = self._last_cycle_ticks = 0
         self._pwm_max = 0.
         self._set_cmd = self._set_cycle_ticks = None
+
     def get_mcu(self):
         return self._mcu
     def setup_max_duration(self, max_duration):
@@ -399,11 +401,12 @@ class MCU_pwm:
         self._start_value = max(0., min(1., start_value))
         self._shutdown_value = max(0., min(1., shutdown_value))
         self._is_static = is_static
+    def setup_high_throughput_mode(self):
+        self._is_ht = True
     def _build_config(self):
         if self._max_duration and self._start_value != self._shutdown_value:
             raise pins.error("Pin with max duration must have start"
                              " value equal to shutdown value")
-        cmd_queue = self._mcu.alloc_command_queue()
         curtime = self._mcu.get_printer().get_reactor().monotonic()
         printtime = self._mcu.estimated_print_time(curtime)
         self._last_clock = self._mcu.print_time_to_clock(printtime + 0.200)
@@ -411,6 +414,7 @@ class MCU_pwm:
         mdur_ticks = self._mcu.seconds_to_clock(self._max_duration)
         if mdur_ticks >= 1<<31:
             raise pins.error("PWM pin max duration too large")
+        self._min_clock_diff = cycle_ticks
         if self._hardware_pwm:
             self._pwm_max = self._mcu.get_constant_float("PWM_MAX")
             if self._is_static:
@@ -419,8 +423,10 @@ class MCU_pwm:
                     % (self._pin, cycle_ticks,
                        self._start_value * self._pwm_max))
                 return
-            self._mcu.request_move_queue_slot()
             self._oid = self._mcu.create_oid()
+            if not self._is_ht:
+                self._mcu.request_move_queue_slot()
+
             self._mcu.add_config_cmd(
                 "config_pwm_out oid=%d pin=%s cycle_ticks=%d value=%d"
                 " default_value=%d max_duration=%d"
@@ -432,7 +438,8 @@ class MCU_pwm:
                                      % (self._oid, self._last_clock, svalue),
                                      on_restart=True)
             self._set_cmd = self._mcu.lookup_command(
-                "queue_pwm_out oid=%c clock=%u value=%hu", cq=cmd_queue)
+                "queue_pwm_out oid=%c clock=%u value=%hu",
+                fast_track=self._is_ht) # might be ugly without cq?
             return
         # Software PWM
         if self._shutdown_value not in [0., 1.]:
@@ -443,8 +450,11 @@ class MCU_pwm:
             return
         if cycle_ticks >= 1<<31:
             raise pins.error("PWM pin cycle time too large")
-        self._mcu.request_move_queue_slot()
         self._oid = self._mcu.create_oid()
+        cmd_queue = None
+        if not self._is_ht:
+            self._mcu.request_move_queue_slot()
+            cmd_queue = self._mcu.alloc_command_queue()
         self._mcu.add_config_cmd(
             "config_digital_out oid=%d pin=%s value=%d"
             " default_value=%d max_duration=%d"
@@ -459,34 +469,48 @@ class MCU_pwm:
             "queue_digital_out oid=%d clock=%d on_ticks=%d"
             % (self._oid, self._last_clock, svalue), is_init=True)
         self._set_cmd = self._mcu.lookup_command(
-            "queue_digital_out oid=%c clock=%u on_ticks=%u", cq=cmd_queue)
+            "queue_digital_out oid=%c clock=%u on_ticks=%u",
+            fast_track=self._is_ht)
         self._set_cycle_ticks = self._mcu.lookup_command(
-            "set_digital_out_pwm_cycle oid=%c cycle_ticks=%u", cq=cmd_queue)
+            "set_digital_out_pwm_cycle oid=%c cycle_ticks=%u", cq=None)
     def set_pwm(self, print_time, value, cycle_time=None):
-        clock = self._mcu.print_time_to_clock(print_time)
+        req_clock = self._mcu.print_time_to_clock(print_time)
+        # FIXME: let sync_channel replace uncommitted values
+        #        so that this is not necessary
         minclock = self._last_clock
-        self._last_clock = clock
+        clock = max(req_clock, self._last_clock + self._min_clock_diff)
         if self._invert:
             value = 1. - value
         if self._hardware_pwm:
             v = int(max(0., min(1., value)) * self._pwm_max + 0.5)
-            self._set_cmd.send([self._oid, clock, v],
+            # queue_pwm_out oid=%c clock=%u value=%hu
+            #data = (self._set_cmd, self._oid, clock & 0xFFFFFFFF, v)
+            self._set_cmd.send([self._oid, clock & 0xFFFFFFFF, v],
+                                minclock=minclock, reqclock=clock)
+        else:
+            # Soft pwm update
+            if cycle_time is None:
+                cycle_time = self._cycle_time
+            cycle_ticks = self._mcu.seconds_to_clock(cycle_time)
+            if cycle_ticks != self._last_cycle_ticks:
+                if cycle_ticks >= 1<<31:
+                    raise self._mcu.get_printer().command_error(
+                        "PWM cycle time too large")
+                self._set_cycle_ticks.send([self._oid, cycle_ticks],
+                                           minclock=minclock, reqclock=clock)
+                # FIXME: this makes sure new cycle_ticks is applied
+                # _before_ new on_ticks arrive
+                clock = self._mcu.print_time_to_clock(print_time +
+                                                      self._last_cycle_ticks)
+                self._last_cycle_ticks = cycle_ticks
+                self._min_clock_diff = cycle_ticks
+            on_ticks = int(max(0., min(1., value)) * float(cycle_ticks) + 0.5)
+            # queue_digital_out oid=%c clock=%u on_ticks=%u
+            #data = (self._set_cmd, self._oid, clock & 0xFFFFFFFF, on_ticks)
+            self._set_cmd.send([self._oid, clock & 0xFFFFFFFF, on_ticks],
                                minclock=minclock, reqclock=clock)
-            return
-        # Soft pwm update
-        if cycle_time is None:
-            cycle_time = self._cycle_time
-        cycle_ticks = self._mcu.seconds_to_clock(cycle_time)
-        if cycle_ticks != self._last_cycle_ticks:
-            if cycle_ticks >= 1<<31:
-                raise self._mcu.get_printer().command_error(
-                    "PWM cycle time too large")
-            self._set_cycle_ticks.send([self._oid, cycle_ticks],
-                                       minclock=minclock, reqclock=clock)
-            self._last_cycle_ticks = cycle_ticks
-        on_ticks = int(max(0., min(1., value)) * float(cycle_ticks) + 0.5)
-        self._set_cmd.send([self._oid, clock, on_ticks],
-                           minclock=minclock, reqclock=clock)
+
+        self._last_clock = clock
 
 class MCU_adc:
     def __init__(self, mcu, pin_params):
@@ -551,6 +575,22 @@ class MCU_adc:
 # Main MCU class
 ######################################################################
 
+class FastCommandWrapper:
+    def __init__(self, mcu, cmd_id, sync_channel, queue_msg_fn):
+        self._mcu = mcu
+        self._cmd_id = cmd_id
+        self._sync_channel = sync_channel
+        self._queue_msg_fn = queue_msg_fn
+        self._reactor = mcu.get_printer().get_reactor()
+        self._toolhead = mcu.get_printer().lookup_object('toolhead')
+    def send(self, data, minclock, reqclock):
+        data.insert(0, self._cmd_id)
+        self._queue_msg_fn(self._sync_channel, data, len(data), reqclock)
+        print_time = self._mcu.clock_to_print_time(reqclock)
+        # TODO: Is here actually `_async_` needed?
+        self._reactor.register_async_callback(
+            lambda ev: self._toolhead.note_synchronous_command(print_time))
+
 class MCU:
     error = error
     def __init__(self, config, clocksync):
@@ -598,12 +638,18 @@ class MCU:
         self._init_cmds = []
         self._mcu_freq = 0.
         # Move command queuing
-        ffi_main, self._ffi_lib = chelper.get_ffi()
+        self._ffi_main, self._ffi_lib = chelper.get_ffi()
         self._max_stepper_error = config.getfloat('max_stepper_error', 0.000025,
                                                   minval=0.)
         self._reserved_move_slots = 0
         self._stepqueues = []
         self._steppersync = None
+
+        #
+        self._sync_channels = []
+        self._active_fasttrack_queues = 0
+        #
+
         # Stats
         self._get_status_info = {}
         self._stats_sumsq_base = 0.
@@ -768,9 +814,12 @@ class MCU:
         ffi_main, ffi_lib = chelper.get_ffi()
         self._steppersync = ffi_main.gc(
             ffi_lib.steppersync_alloc(self._serial.get_serialqueue(),
-                                      self._stepqueues, len(self._stepqueues),
-                                      move_count-self._reserved_move_slots),
-            ffi_lib.steppersync_free)
+                              self._stepqueues, len(self._stepqueues),
+                              self._sync_channels, len(self._sync_channels),
+                              move_count-self._reserved_move_slots),
+            ffi_lib.steppersync_free
+        )
+
         ffi_lib.steppersync_set_time(self._steppersync, 0., self._mcu_freq)
         # Log config information
         move_msg = "Configured MCU '%s' (%d moves)" % (self._name, move_count)
@@ -854,6 +903,8 @@ class MCU:
         return self.print_time_to_clock(t) + slot
     def register_stepqueue(self, stepqueue):
         self._stepqueues.append(stepqueue)
+    def register_sync_channel(self, sync_channel):
+        self._sync_channels.append(sync_channel)
     def request_move_queue_slot(self):
         self._reserved_move_slots += 1
     def seconds_to_clock(self, time):
@@ -869,8 +920,22 @@ class MCU:
         self._serial.register_response(cb, msg, oid)
     def alloc_command_queue(self):
         return self._serial.alloc_command_queue()
-    def lookup_command(self, msgformat, cq=None):
-        return CommandWrapper(self._serial, msgformat, cq)
+    def lookup_command(self, msgformat, cq=None, fast_track=False):
+        if not fast_track:
+            return CommandWrapper(self._serial, msgformat, cq)
+
+        if self._active_fasttrack_queues > 0:
+            raise error ("Currently, only one high throughput pin is allowed!")
+        self._active_fasttrack_queues = self._active_fasttrack_queues + 1
+        sync_channel = self._ffi_main.gc(
+                self._ffi_lib.sync_channel_alloc(),
+                self._ffi_lib.sync_channel_free)
+        self.register_sync_channel(sync_channel)
+        cmd_tag = self.lookup_command(msgformat).get_command_tag()
+        return FastCommandWrapper(self, cmd_tag,
+                                  sync_channel,
+                                  self._ffi_lib.sync_channel_queue_msg)
+
     def lookup_query_command(self, msgformat, respformat, oid=None,
                              cq=None, is_async=False):
         return CommandQueryWrapper(self._serial, msgformat, respformat, oid,
--- a/klippy/mcu.py
+++ b/klippy/mcu.py
@@ -483,8 +483,6 @@ class MCU_pwm:
             value = 1. - value
         if self._hardware_pwm:
             v = int(max(0., min(1., value)) * self._pwm_max + 0.5)
-            # queue_pwm_out oid=%c clock=%u value=%hu
-            #data = (self._set_cmd, self._oid, clock & 0xFFFFFFFF, v)
             self._set_cmd.send([self._oid, clock & 0xFFFFFFFF, v],
                                 minclock=minclock, reqclock=clock)
         else:
@@ -505,8 +503,6 @@ class MCU_pwm:
                 self._last_cycle_ticks = cycle_ticks
                 self._min_clock_diff = cycle_ticks
             on_ticks = int(max(0., min(1., value)) * float(cycle_ticks) + 0.5)
-            # queue_digital_out oid=%c clock=%u on_ticks=%u
-            #data = (self._set_cmd, self._oid, clock & 0xFFFFFFFF, on_ticks)
             self._set_cmd.send([self._oid, clock & 0xFFFFFFFF, on_ticks],
                                minclock=minclock, reqclock=clock)
 
--- a/klippy/mcu.py
+++ b/klippy/mcu.py
@@ -424,8 +424,10 @@ class MCU_pwm:
                        self._start_value * self._pwm_max))
                 return
             self._oid = self._mcu.create_oid()
+            cmd_queue = None
             if not self._is_ht:
                 self._mcu.request_move_queue_slot()
+                cmd_queue = self._mcu.alloc_command_queue()
 
             self._mcu.add_config_cmd(
                 "config_pwm_out oid=%d pin=%s cycle_ticks=%d value=%d"
@@ -438,8 +440,8 @@ class MCU_pwm:
                                      % (self._oid, self._last_clock, svalue),
                                      on_restart=True)
             self._set_cmd = self._mcu.lookup_command(
-                "queue_pwm_out oid=%c clock=%u value=%hu",
-                fast_track=self._is_ht) # might be ugly without cq?
+                "queue_pwm_out oid=%c clock=%u value=%hu", cq=cmd_queue,
+                fast_track=self._is_ht)
             return
         # Software PWM
         if self._shutdown_value not in [0., 1.]:
@@ -469,10 +471,10 @@ class MCU_pwm:
             "queue_digital_out oid=%d clock=%d on_ticks=%d"
             % (self._oid, self._last_clock, svalue), is_init=True)
         self._set_cmd = self._mcu.lookup_command(
-            "queue_digital_out oid=%c clock=%u on_ticks=%u",
+            "queue_digital_out oid=%c clock=%u on_ticks=%u", cq=cmd_queue,
             fast_track=self._is_ht)
         self._set_cycle_ticks = self._mcu.lookup_command(
-            "set_digital_out_pwm_cycle oid=%c cycle_ticks=%u", cq=None)
+            "set_digital_out_pwm_cycle oid=%c cycle_ticks=%u", cq=cmd_queue)
     def set_pwm(self, print_time, value, cycle_time=None):
         req_clock = self._mcu.print_time_to_clock(print_time)
         # FIXME: let sync_channel replace uncommitted values
--- a/klippy/mcu.py
+++ b/klippy/mcu.py
@@ -424,10 +424,9 @@ class MCU_pwm:
                        self._start_value * self._pwm_max))
                 return
             self._oid = self._mcu.create_oid()
-            cmd_queue = None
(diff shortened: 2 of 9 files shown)


Arti4ever:
Hi @Cirromulus, 
I am very interested in using laser engraver with Klipper, so I gave a try to your PR. 
Unfortunately, your code do not work for me, Klipper crash when executing homing with error : " Exception in flush_handler".
Logs for this error : [klippy_G28.log](https://github.com/KevinOConnor/klipper/files/6289527/klippy_G28.log)

I also tested just using M3 S5 command, but there is also an error : "MCU 'mcu' shutdown: Timer too close"
Logs for this error : [klippy_M3.log](https://github.com/KevinOConnor/klipper/files/6289531/klippy_M3.log)

I hope this may help you to fix theses errors. Or maybe advise me if I am doing something wrong ?

klippy/mcu.py
```diff
@@ -246,32 +251,43 @@ def _build_config(self):
...
-        cycle_ticks = self._mcu.seconds_to_clock(cycle_time)
-        if cycle_ticks != self._last_cycle_ticks:
-            self._set_cycle_ticks.send([self._oid, cycle_ticks],
-                                       minclock=minclock, reqclock=clock)
-            self._last_cycle_ticks = cycle_ticks
-        on_ticks = int(max(0., min(1., value)) * float(cycle_ticks) + 0.5)
-        self._set_cmd.send([self._oid, clock, on_ticks],
-                           minclock=minclock, reqclock=clock)
+            # queue_pwm_out oid=%c clock=%u value=%hu
+            data = (self._set_cmd, self._oid, clock & 0xFFFFFFFF, v)
```
Cirromulus, line 269: This may be wrong on non-8-bit boards

Cirromulus:
Thanks for testing!
I will look into that soon. 
You are using a 32bit board, do you?
I never tested it with one, so maybe the manual cap on the scheduled time is wrong (mcu.py)
Just for kicks, does it behave the same when using hardware_pwm: true  ?

Arti4ever:
Yes I am using a 32 board.
Using hardware pwm do not work for me. It raise a config error.

Arti4ever:
I finally tried with hardware PWM (support for LPC176x was added 2 weeks ago, so my mcu was not up to date, that's why it did not work at first...), but I have the same behavior.

KevinOConnor:
Interesting.

At a high-level, the code doesn't look correct to me - at a minimum, the `set_pwm()` command has to be able to work from a background thread, which means it isn't valid for that function to call the toolhead code.  (The heater code will change PWM settings directly from the background thread used to process temperature reports.)  Alas, I haven't had time to look into this, so I don't have any implementation advice.

-Kevin

Cirromulus:
Hi Kevin, thanks for your first feedback.
I will look into that (I believe I initially produced the same condition in the pin safety timeout PR).
So basically, it is not safe to call `....send([self._oid, clock, v], minclock=minclock, reqclock=clock)` inside the GCode-code?
I suspect implementing a similar style as the `command_queue` would be more elegant and probably allow for concurrency.

KevinOConnor:
> So basically, it is not safe to call ....send([self._oid, clock, v], minclock=minclock, reqclock=clock)

I'm not sure about the above.  My comments were referring to the `self._th.note_synchronous_command(print_time)` call from `set_pwm()` - it's not valid to call toolhead methods from a background thread.

-Kevin

klippy/toolhead.py
```diff
@@ -536,6 +536,32 @@ def register_lookahead_callback(self, callback):
...
+
+        # Also: Is it allowed for a second kin_time
+        # to be smaller than a previous one?
+
+        self.note_kinematic_activity(kin_time)
+        if(kin_time >= self.last_kin_move_time):
+            if self.special_queuing_state == "Drip":
+                self._update_drip_move_time(kin_time)
+            else:
+                self._update_move_time(kin_time)
```
Cirromulus, line 560: Is this always the correct way to flush steppersync queues?

klippy/mcu.py
```diff
@@ -246,32 +256,47 @@ def _build_config(self):
...
+            on_ticks = int(max(0., min(1., value)) * float(cycle_ticks) + 0.5)
+            # queue_digital_out oid=%c clock=%u on_ticks=%u
+            data = (self._set_cmd, self._oid, clock & 0xFFFFFFFF, on_ticks)
+
+        self._ffi_lib.sync_channel_queue_msg(self._sync_channel,
+                                             data, len(data), clock)
+
+        # TODO: Is here actually `register_async_callback` needed?
+        self.reactor.register_callback(
+            lambda ev: self._th.note_synchronous_command(print_time))
```
Cirromulus, line 297: @KevinOConnor If heater-code (background task) calls this, is actually `register_async_callback` needed?
It worked both ways for me.

Cirromulus:
@Arti4ever If you want to try it again, it may be better now.
Toolhead code is now called by the reactor. 

@TazerReloaded
What PWM frequency do you use? The Code limits a maximum update frequency to the pwm frequency.

TazerReloaded:
I wrote this config:
```
[output_pin intensity]
pin: P2.0
pwm: true
cycle_time: 0.001
scale: 255
value: 0
shutdown_value: 0
[gcode_macro M3]
default_parameter_S: 255
gcode:
  SET_PIN PIN=intensity VALUE={S}
[gcode_macro M5]
gcode:
  SET_PIN PIN=intensity VALUE=0
```
So currently I'm using 1kHz software PWM i guess, which is a bit slow considering 50mm/s at a resolution of 0.1 already means 500 pin updates per second. How fast can/should I go with software PWM?
I also tried `hardware_pwm: True`, but that did not change the frequency (checked with scope), despite having the newest (as of yesterday) firmware built from your branch on the SKR. Do I have to specify a frequency for hardware PWM to work? Pin 2.0 seems to be PWM capable according to the chip manual.
I also noticed that my laser seems to respond to low analog voltages. I have a 470 Ohm resistor on the PWM line to protect the MCU from over current (already killed one with a short to VCC due to bad wiring), which means the pull-up on the laser board always pulls the PWM line to around 0.2V when the duty cycle is 0, and that makes the laser shimmer a bit. I use that like a feature because that low intensity can be used to align the laser with the zero point, and when not in use the 12V power to the laser is cut via a relay. I don't think it affects the PWM signal.

I also pulled your latest changes, will continue testing next weekend. Awesome work so far, already works great, thank you!

Arti4ever:
It but it still doesn't work (for me at least) when I try do to a homing. the error is not the same though : " Failed to home x: Timeout during homing"
Same config on the master branch works great.
Here my log : [klippy.log](https://github.com/KevinOConnor/klipper/files/6360478/klippy.log)

TazerReloaded:
Also noticed an issue while trying to print today: Whenever I enable a heater, Klipper shuts down with a "Timer too close" error after a few seconds. Same config on master works fine. Klippy log attached.
[klippy.log](https://github.com/KevinOConnor/klipper/files/6360505/klippy.log)

Cirromulus:
>How fast can/should I go with software PWM?

That depends on the computing power of your board. I suggest testing higher freqencies and measuring the jitter with a oscilloscope.

> I also tried hardware_pwm: True, but that did not change the frequency 

Then, the hardware-pwm was able to be configured to your desired frequency.
With `cycle_time: 0.001` you specified a frequency of 1kHz (1/cycle_time). My board can't go faster, but yours might.

> Whenever I enable a heater, Klipper shuts down with a "Timer too close" error after a few seconds. 

I can reproduce that. I suspect that my way of `toolhead.note_synchronous_command()` is susceptible to concurrently flushing the move queues (in the past).
This is because I somehow rely on the move commands in the `steppersync` system to flush regurlarly.
This way, pwm-commands do not starve the stepper-queues. This introduces a lot of flushing and concurrency issues, however.
I will need some time to analyze the previously-used `command_queue`s handled by the serial (slow) path.

Cirromulus:
@Arti4ever I see that you use a BL-Touch as well. Does it even stow?
I had this problem during development as well, but it was solved within the last commits.

I like the fact that you also can cut the power supply to the Laser.
But it is possible, that this breaks the setup (two different pin-updates immediately after each other... did not test it yet!)
Klipper will try to update these pins exactly at the same time, I suppose.

TazerReloaded:
The relay for laser power is attached to the Pi and controlled via Moonraker power commands, so I don't think that affects Klipper in any way. I also control the printers 12/24V PSU and the enclosure lights that way. I usually leave the room before powering the laser, don't want to take any risks, and if anything goes wrong, I can cut the power to everything remotely.
I can't think of a use case where pin changes have to happen exactly simultaneously, maybe if someone builds a needle printer running Klipper, but from an user perspective I'd expect it to be possible.

TazerReloaded:
I tested the output frequency today, and on my hardware (SKR 1.4), I get a nice clean 10.8kHz square wave with the following settings:
```
hardware_pwm: true
cycle_time: 0.0001
```
It is a little bit faster than the expected frequency, but that could be caused by the hardware PWM clock scaling. Edges look very clean and even values like 0.005 (0.5%) are represented accurately.
I also installed a new laser, which responds very precisely to those PWM values. I'm cutting 4mm plywood, so no high speeds, but stock Klipper is still way too slow, leaving ugly marks when moving the laser before shutting it off. Your PR on the other hand works perfectly, no timing issues so far. If it wouldn't render heaters unusable at the moment I would already use it in my everyday firmware.

Cirromulus:
Nice hearing that!
Unfortunately, getting the heaters to work requires partly a new concept.
Currently no free time for that milestone 🐱‍🐉

At least, one can check out the different branch, restart klipper and go on lasing.
Before pwm_queue was not implemented in the mcu firmware, one would have to re-flash the printer with every switch 🤕

TazerReloaded:
Unfortunately, the time of re-flashing is coming back sooner or later. The master branch has already altered the communication protocol, and introduced a few fixes for my setup. I flashed a DFU loader onto my board, so at least I don't have to fiddle the memory card out of the electronics enclosure every time. According to GitHub here are no conflicts so far, so a patch would be possible, but I expect it to be only a matter of a few weeks until conflicts arise.
I'd like to help, but I'm also short on free time, and I doubt that I can come up with a better solution without investing at least a few days into understanding the inner mechanisms of Klipper.
Anyways, thank you for all the effort so far, it has some edges, but it's a working solution!

Cirromulus:
>According to GitHub here are no conflicts so far, so a patch would be possible, but I expect it to be only a matter of a few weeks until conflicts arise.

You are correct. 

>I doubt that I can come up with a better solution without investing at least a few days into understanding the inner mechanisms of Klipper.

That is my problem right now. I need a free weekend to uncouple the sync_channel from the move queues.
Right now, if any pin-update flushes the move queue (or the mq itself), a different pin (e.g. from a background task) will queue its update, but will not be flushed till the next round, which may or may not be too late for the update to happen.

Alexdad76:
"Timer too close" error after a few seconds after the command M3 (1-255) . Same config on master works fine. Klippy log attached.
SKR 1.1
[klippy.log](https://github.com/KevinOConnor/klipper/files/6467869/klippy.log)

Alexdad76:
> 
> 
> > According to GitHub here are no conflicts so far, so a patch would be possible, but I expect it to be only a matter of a few weeks until conflicts arise.
> 
> You are correct.
> 
> > I doubt that I can come up with a better solution without investing at least a few days into understanding the inner mechanisms of Klipper.
> 
> That is my problem right now. I need a free weekend to uncouple the sync_channel from the move queues.
> Right now, if any pin-update flushes the move queue (or the mq itself), a different pin (e.g. from a background task) will queue its update, but will not be flushed till the next round, which may or may not be too late for the update to happen.

I want to clarify. Does you need more information? https://github.com/KevinOConnor/klipper/pull/4128#issuecomment-839947083

Cirromulus:
Your log points to an error in the message flush system. This relates to the strong coupling between stepper moves and the pin output.

> Line 870: Sent 99 5414.972272 5414.972272 14: seq: 10, queue_pwm_out oid=4 clock=1735086958 value=100
> Line 971: Receive: 99 5414.973191 5414.972272 12: seq: 11, shutdown clock=1745260113 static_string_id=Timer too close

This indicates that the second `queue_pwm_out` was sent out far too late (10.173.155 clock cycles, or around 0,1 seconds).
Do you have `maximum_mcu_duration` turned on? If configured, I suggest either turning it off or increasing the value for testing purposes.

Alexdad76:
maximum_mcu_duration in config turned on:
`maximum_mcu_duration = 5`

When trying to turn off an error occurs:
`maximum_mcu_duration = 0`
Error: Option 'maximum_mcu_duration' in section 'output_pin TOOL' must have minimum of 0.5

When trying to increase more than 5, an error occurs
`maximum_mcu_duration = 15`
Error: Option 'maximum_mcu_duration' in section 'output_pin TOOL' must have maximum of 5.0

Cirromulus:
Yeah, this one is a bit tricky. To turn it off, you have to delete the line. Then it will default to zero.

Alexdad76:
> 
> 
> Yeah, this one is a bit tricky. To turn it off, you have to delete the line. Then it will default to zero.

HOORAY! Klipper stopped shutdown! The idle passage without a laser was successful. The LED on the output of the fan worked correctly.I will try real burning. Thank you so much for the hint!

Cirromulus:
No Problem. Once this system gets more stable, you should activate it again, though. For safety ;)

Alexdad76:
> 
> 
> No Problem. Once this system gets more stable, you should activate it again, though. For safety ;)

Everything works fine:
Test80-80
Laser: NEJE Module 7W
Software: https://laserweb.yurl.ch/
Mode: "Laser Raster" 
Laser Diameter: 0,1
Speed: 80 mm/s, Power: 0-80%

1. Heard strange knocks over each square.
2. There are some artifacts - lines about 6 and 7 - the laser was not turned off (maybe the G-code bug).

By the way, LaserWeb does not create a G-code if in the section, use M3 $ Intensity. It works only simply "M3"

Just in case I attach klippy.log
[klippy80_80.log](https://github.com/KevinOConnor/klipper/files/6487635/klippy80_80.log)

Arti4ever:
I finally got it "work" by : 
- Disabling BLTouch (if not, I was not able to do homing. always got "timer too close" issue)
- Commenting `maximum_mcu_duration` (thanks fo the hint, because like @Alexdad76 I always got "timer too close" issue when using M3 command)

Besides that, it works pretty good. 
I got some freezing and other "timer too close" issues when navigating trough octoprint while the laser is burning. So now when I launch an engravure, I touch nothing and it goes pretty well.

Alexdad76:
Maybe add section in config:
```
[mode]
Printer | Laser
```

In the laser mode, turn off control and enable heaters. I doubt that someone will simultaneously use the heaters during laser engraving / cutting.

TazerReloaded:
Klipper already supports modular config, you an just do:
```
[include laser.cfg]
#[include extruder.cfg]
```
Then define the heater and extruder stuff in extruder.cfg, and the laser pins in laser.cfg. You can also have different probes and travel limits for each tool this way. Then just comment the includes according to the current setup.

Alexdad76:
> 
> 
> Klipper already supports modular config, you can just do:
> 
> ```
> [include laser.cfg]
> #[include extruder.cfg]
> ```
> 
> Then define the heater and extruder stuff in extruder.cfg, and the laser pins in laser.cfg. You can also have different probes and travel limits for each tool this way. Then just comment the includes according to the current setup.

In the context of the current PR - send resources to generate PWM for Laser, and not just change the configuration.

TazerReloaded:
> send resources to generate PWM for Laser, and not just change the configuration.

I don't think Klipper uses any resources for components that are not defined in the config. I have just the motion system configured in `printer.cfg`.
The extruder stepper, heaters and thermistors are only defined when I include `extruder.cfg`. The laser output pin and end stop is only defined when I include `laser.cfg`. That way I also can't accidentally enable a heater when the laser is in use, heaters are just not available in that mode.
Klipper works fine without an extruder definition, and switching output pins with GCode macros makes it possible to attach almost anything to the carriage, and extra config files for every tool make perfect sense, because most attachments will have different dimensions, probe offsets, home positions and so on.

Sinclair5:
> > send resources to generate PWM for Laser, and not just change the configuration.
> 
> I don't think Klipper uses any resources for components that are not defined in the config. I have just the motion system configured in `printer.cfg`.
> The extruder stepper, heaters and thermistors are only defined when I include `extruder.cfg`. The laser output pin and end stop is only defined when I include `laser.cfg`. That way I also can't accidentally enable a heater when the laser is in use, heaters are just not available in that mode.
> Klipper works fine without an extruder definition, and switching output pins with GCode macros makes it possible to attach almost anything to the carriage, and extra config files for every tool make perfect sense, because most attachments will have different dimensions, probe offsets, home positions and so on.





is there an example of the laser.cfg?

TazerReloaded:
Sure, here you go:
```
[probe]
pin: ^P0.10
x_offset: -30
y_offset: 15
z_offset: 0
speed: 5

[safe_z_home]
home_xy_position: 60,125
speed: 80
z_hop: 7
z_hop_speed: 5

[output_pin laser_intensity]
pin: P2.0
pwm: true
hardware_pwm: true
cycle_time: 0.0001
scale: 255
value: 0
shutdown_value: 0

[gcode_macro M3]
default_parameter_S: 255
gcode:
  SET_PIN PIN=laser_intensity VALUE={S}

[gcode_macro M5]
gcode:
  SET_PIN PIN=laser_intensity VALUE=0
```
I attached a switch to the laser, so I can home it on the stock and automatically have the laser in focus. I printed a little holder for everything, and on the left is a fan to blow the smoke away from the lens:
IMG_20210607_192430821

Sinclair5:
thank you so  so much

Sinclair5:
do major slicer changes needed right?

TazerReloaded:
You need a slicer that can generate Marlin like GCode for a laser, like LaserWeb or similar. Grbl GCode does not work with Klipper at the moment. I wrote my own little tool [here](http://laser.pc-id.de/).

Sinclair5:
does laser web also controll the z movement?

TazerReloaded:
Usually not, but you can do multiple passes and lower the height a bit on each pass to move the focal point through the stock. You could also theoretically do a height map of the surface, but that's a more advanced topic.

Sinclair5:
yea ur right,   i feel like i get close to getting the software part of a diy sls printer done,  the i get far again.      i will just wait until an easier solution is out for klipper

obelisk79:
@Cirromulus 

Seems like interest/work/testing on this has stalled a bit. What do you consider the current status to be. Also, what sort of testing is still needed with this? I am in the camp that I would very much like to see this type of pwm laser support added to klipper.

Cirromulus:
Hi,
the current status is "works under special conditions", I would say.
This is a proof-of-concept that shows that (technically) klipper is indeed able to drive PWM fast enough for raster engraving.
Currently I can't find time to integrate it in such a way that does not break a lot of other stuff.

As I wrote earlier:
>That is my problem right now. I need a free weekend to uncouple the sync_channel from the move queues.
Right now, if any pin-update flushes the move queue (or the mq itself), a different pin (e.g. from a background task) will queue its update, but will not be flushed till the next round, which may or may not be too late for the update to happen.

In other words, I would have to generalize "fast channel" serial queues in which the move queue would just be one of.
This is a pretty deep dive into klippers internals (on the borders of the c_helpers).
I do not plan to abandon this project, as klipper is a great thing to be contributing to. But as it is currently working for me and some others, it is not my highest priority. (especially in the thin time of being able to leave the house between now and the coming lockdowns 🥶)

TazerReloaded:
"works under special conditions" is the correct description, I have a separate Klipper folder with this branch, and every time I use my lasers, I rename the folder and flash the firmware to the MCU and the Pi. This is quite a bit of effort, especially considering that I build a quick swap mechanism which allows me to switch tools within seconds.
The other unresolved Problem is the missing influence of acceleration/deceleration. I move my laser at a constant speed for all operations to minimize inconsistencies, but for proper operation, linear scaling of the PWM according to current actual movement speed would be required.
I would like to help, but unfortunately I have quite a few other projects to complete. So I'll just hope that @Cirromulus or someone else is willing to invest some time into this. I'll happily help with further testing, I have to engrave things for other projects from time to time.

klippy/mcu.py
```diff
@@ -314,8 +307,9 @@ def _build_config(self):
             self._mcu.add_config_cmd("queue_pwm_out oid=%d clock=%d value=%d"
                                      % (self._oid, self._last_clock, svalue),
                                      on_restart=True)
-            self._set_cmd = self._mcu.lookup_command_tag(
-                "queue_pwm_out oid=%c clock=%u value=%hu")
+            self._set_cmd = self._mcu.lookup_command(
+                "queue_pwm_out oid=%c clock=%u value=%hu",
+                fast_track=self._is_ht) #might be ugly without cq?
```
Cirromulus, line 312: TODO: actually reserve move queue slot when not in HT-Mode

klippy/extras/output_pin.py
```diff
@@ -60,7 +62,7 @@ def _set_pin(self, print_time, value, cycle_time, is_resend=False):
         if value == self.last_value and cycle_time == self.last_cycle_time:
             if not is_resend:
                 return
-        print_time = max(print_time, self.last_print_time + PIN_MIN_TIME)
```
Cirromulus, line 63: make this dependent on ht_mode

Cirromulus:
Squashed current work status. Seems to work simultaneously with printing and lasercutting, but still needs long-term testing.
After a small conversation with @KevinOConnor  on discord, I added a _command wrapper_ that takes care about how `output_pin`  updates are sent: normally or via the "fast track" in the stepcompress system.
The current way via the stepcompress system is not optimal, but it works for (at least) one pin.
The main problem with it is, that updates in the "past" (relative to `kin_time`) are not sent. This is under normal operation only a problem when using the `max_mcu_duration`, as long (linear) moves are compressed and thus not split.
With this wrapper, though, one could change this backend without breaking user's setups.

My questions here are:
 - Is the `CommandWrapper`-Interface used ok, or shall it be done in another style?
 - What parts of this are sensible to merge?

PS.: My working setup is having all common settings to my hardware in one file, and then including it in the main files for printer- and lasercutter-config. This way, I can change the printheads and load the new config pretty fast, with all common changes updated.

TazerReloaded:
Could you update your branch with the current changes from master? There are a few conflicts in "klippy/mcu.py", and I'm not sure how to merge them. I need some recent updates from master for my setup, so I can't just use the old version.
Thank you!

Cirromulus:
I can see when I find the time, this should be a small fix.

Regarding the state of this PR, this has to get some attention from @KevinOConnor .
My changes are not as beautiful as I would hope, but it works in my setup.
Some user-experiences from people using this fork would be nice, so that both him and me can identify todos.
Patching behind master-commits is not that fun after a year.

TazerReloaded:
Just used it today, for engraving the handle of a hatchet for a friend. Worked fine, although my laser would be capable of much faster speeds. But above around 30 mm/s the areas where the tool head has to accelerate or break are visibly darker than the rest of the engraving. So for laser applications, proportional control of the intensity is required.
It would theoretically be possible to solve this issue in the software that generates the GCode (accelerating in advance), but that seems way more complicated than doing it in firmware.
Unfortunately I can't test printing at the moment, because my printer requires some recent changes to sensor curves from master, and I don't want to merge pieces of both together. If the sync channel is completely inactive when the pin is not configured, are there any drawbacks when integrating it into master? What is the reason for the one pin limitation? On most boards, only one register has to be written to change the PWM duty cycle. I guess it's because of limited mcu communication bandwidth?
I engraved a raster test piece to illustrate the acceleration/deceleration problem (extreme example):
IMG_20211219_200400879
Engraved at 100 mm/s, I think my hardware (BTT SKR 1.4, RPI 3) could go even faster. Resolution 0.1mm. GCode processor written by me, it just visits the area it absolutely has to, so no overshoot on either side.
Thank you again for all your work!

Cirromulus:
> But above around 30 mm/s the areas where the tool head has to accelerate or brake are visibly darker than the rest of the engraving. So for laser applications, proportional control of the intensity is required.

This will be a separate PR. The complexity in that is why klipper is so good: Input Shaping and Stepcompress system.
Proportionality (in `G0/1 S<0-255>` command) is not so simple as it requires a command from the updated queue system, and I need a lot of time for that to wrap my head around.

In the meantime, for this I would either recommend using margin (overshoot) to let your laser-head accelerate to the required speed for now.

> If the sync channel is completely inactive when the pin is not configured, are there any drawbacks when integrating it into master? 

This is more of a question for Kevin. But my changes are pretty deep, so I think he will want to really be sure nothing breaks on other ends. Do you use the PR also for printing, and if so, did you notice anything?

>What is the reason for the one pin limitation? On most boards, only one register has to be written to change the PWM duty cycle. I guess it's because of limited mcu communication bandwidth?

Funny thing, but it's the way I insert pin-updates into the command queue on the host side.
The tasking system is complex and I most certainly break stuff when two competing pin queues flush the move queue.
I don't know in which usecase more than one "realtime" pin is really needed, though.
To make this work "beautifully", I would have to re-design the current `stepcompress` system, as it is currently more than just compressing steps. It also makes sure that the MCUs command queue never over- or underflows.
My current system uses that fact to "piggy back" the flushing mechanism, but on the downside relies on the stepper flushing system in the scheduler. I don't know, but this is probably why Kevin is so reluctant merging this code into master.
When I have more time, I will try and contact core dev in discord.

>Thank you again for all your work!

Thats nice to hear!
And hearing that people actually use this will generate more momentum keeping development.

Cirromulus:
@TazerReloaded you could try my quick testing branch at https://github.com/Cirromulus/klipper/tree/pwm_sync_channel_rebased.
Basic sanity checks are OK, but I could not yet test it with a real laser.

Cirromulus:
Just tested it, it works both for my printer and my laser toolhead.
Burn mark on the Piano is caused by not printing via the virtual SD card.
brettchen

Arti4ever:
I just tested this branch, and I can confirm that it works great for me too, for both laser and printer mode. 
This is the first time I have the Laser head that works seamlessly ! previously I always had some "timer to close" error.
It works with no issue a 150mm/s in raster mode. (did not try higher speed).
Thank you for the work you've done !

FYI, my hardware setup is SKR 1.4 + RP3 + BLTouch.

Arti4ever:
After few more tests, laser rastering an image works fine only if the image is in black and white with very few details. 
I tried to engrave an image in gray scale mode (like a photo), but this is very "laggy" and, every time, I got a "timer too close" error. 
But it works pretty good though.

TazerReloaded:
I tried some cutting and engraving over the last days, worked perfectly as soon as I figured out the bugs in LaserWeb. Made this:
image
Birch plywood with elm veneer inlays. It's going to be a key rack, the little dots mark the hook locations. The black lines from the laser make it look like it's 3d, but the surface is perfectly smooth.
Thank you again for all the hard work you've put into this @Cirromulus, I'll try some printing next and let you know how it goes.

TazerReloaded:
I completed some prints over the last days, longest was around 3 hours. No issues with the sync channel branch, load seems normal, heaters work, quality as usual. Will continue testing and report!

KevinOConnor:
Thanks.

> Regarding the state of this PR, this has to get some attention from @KevinOConnor .

Unfortunately, due to time constraints on my side, I think it is unlikely I'll be able to review this PR in the near future.

The support does sound interesting and useful.  However, as I understand the current state of this PR, it seems to be in an "experimental" state.  That is, it seems to have some artificial limitations and there isn't confidence that it will interact well with other common Klipper components.  While in this state, it isn't a good candidate for merging into the Klipper master branch ( https://www.klipper3d.org/CONTRIBUTING.html ).

FYI, a place to discuss this PR, get testers, and engage a reviewer is on the Klipper Discourse server ( https://www.klipper3d.org/Contact.html ).

-Kevin

Cirromulus:
Ok, I reached out in the Discord channel, but there was no reaction. But indeed, this is not a trivial change. I will try creating an explanation-post on discourse.
>That is,That is, it seems to have some artificial limitations and there isn't confidence that it will interact well with other common Klipper components.

At least this one is now better, as the command wrapper only uses my changes if explicitly configured in the `[output_pin]` section.
Let's see if more people can test this, and a certain confidence builds up.
In the meantime, it makes no sense for me implementing the next steps in the roadmap, but that's ok.

tregnier:
Hi @Cirromulus 
I tried this PR version of the code, it works beautifully for laser!
RPi4 + BTT Octopus and 15W laser, tested up to 150mm/s but i have to slow down as the marking on hard wood cannot actually follow

Below example done at 40mm/s, i can finally use my laser!!
IMG_20220117_002615

Let me know if you need some help for testing. I'll see soon if all works with printing as well and we'll share feedback here if any issue

Thanks for your work, hope to see it one day in Klipper!

bk227865:
I cannot change the CYCLE_TIME via the SET_PIN command, even for non high throughput pins.
Trying on a Ender 3 S1.
I recompiled the firmware to be on the same level as the PR.
[klippy.log](https://github.com/Klipper3d/klipper/files/8749424/klippy.log)
[moonraker.log](https://github.com/Klipper3d/klipper/files/8749425/moonraker.log)

Cirromulus:
Hi,
thanks for testing. In your logs, I dont find the error you describe (klippy.log, line numbers):

1. 1-394: MCU is already shutdown as you restart klipper, you do a firmware_restart
2.  395-1949: You reset it with an emergency stop? MCU shutdown request was initated from host.
3. 1950-end: It runs idle

I can't test it myself right now, but if you record a log where you only play your Macro `END_TUNE` (or another minimal example where the error occurs) I could take another look at it.

bk227865:
yeah i did an emergency stop to end the beeping.
i redid the test but first i issued a non existing gcode.
You can see that appearing in the logfile as : Unknown command:"TEST_WILL_FOLLOW_AFTER_THIS_FAKE_COMMAND"
right after that i did a SET_PIN with a CYCLE_TIME , but it  shows nothing in the logfile about that , execept the Stats.
 
[klippy_out.log](https://github.com/Klipper3d/klipper/files/8754635/klippy_out.log)

Cirromulus:
Just letting you know that I did not forget you, I just can't find the time right now.

thiagolocatelli:
@Cirromulus can you please share your full config, mcu and laser module you are using? I am in the process of start assembling my custom laser engraver/cutter and would like to try your pull request.

Cirromulus:
Hi, in https://github.com/Cirromulus/klipper/blob/pwm_sync_channel/config/sample-pwm-tool.cfg you can find an example configuration.
My configuration is very specific to my Geeetech A10, and also setup modularly so that I can quickly change between printing and lasing. 

I attached the config anyways, for the curious: [klipper-config.zip](https://github.com/Klipper3d/klipper/files/9196995/klipper-config.zip). But please keep that in mind.

Cirromulus:
[bk227865](https://github.com/bk227865)
> All i can say is that it seems to be related somehow with changing the CYCLE_TIME. As long as the SET_PIN just turns on or off through the VALUE setting or the default CYCLE_TIME there is no problem.

I can reproduce your error after changing the CYCLE_TIME a second time, resulting in a lock-up of klipper.
I will investigate that once I find time in the next weeks. Thanks for noticing!

thiagolocatelli:
@Cirromulus Thanks for answering and sharing your config. I will be giving it a try using the SKR Pico since it does have a specific port for laser, with 3 pins and V+ can be 12v or 24v. I am also adding a latching switch button on the V+ so when I turn the machine on, there won't be any power to the laser module unless I switch it on.

This is still a WIP: https://github.com/3DPrintingMods/LaserEngraverCoreXY

thiagolocatelli:
@tregnier where can I find that test file you have?

JeremyKennedy:
I just got this working very nicely with my Sovol laser engraver using a SKR E3 Mini v3.0 board (along with [this](https://github.com/Klipper3d/klipper/pull/5657) patch). I'll be testing this same firmware for regular 3D prints shortly!

JeremyKennedy:
I've been lasering things and printing things on this PR for almost two weeks, no issues!

StoneDiggerX:
I would like to test this PR with my Ender-3 S1 pro.
@bk227865 which configuration do you use with E3 S1 for  [output_pin TOOL] and [output_pin TOOL_ENABLE]?

RuniNonfjall:
Has anyone figured out how to use this with Ender 3 S1 Pro?

Cirromulus:
Your question is welcome in the discourse server, where I am happy to answer.
This PR conversation shall be limited to problems or successes regarding these code changes to keep this topic concise.
Thanks.

holgarsson:
@Cirromulus I have the same question as @RuniNonfjall . How can I find you on discord? What is your username?

Cirromulus:
> @Cirromulus I have the same question as @RuniNonfjall . How can I find you on discord? What is your username?

My username is the same in discord and discourse.
Especially because it interests multiple people, it would be nice discussing it in a new forum post over on [discourse](https://klipper.discourse.group/).

Cirromulus:
@bk227865 
> I cannot change the CYCLE_TIME via the SET_PIN command, even for non high throughput pins. Trying on a Ender 3 S1. If i select a different CYCLE_TIME then the command seems to lock up and i have to reset to stop the PWM. Further SET_PIN commands do nothing anymore, even those without the CYCLE_TIME component. I recompiled the firmware to be on the same level as the PR. [klippy.log](https://github.com/Klipper3d/klipper/files/8749424/klippy.log) [moonraker.log](https://github.com/Klipper3d/klipper/files/8749425/moonraker.log)

Hi, I could reproduce your problem; it had something to do with the wrong calculation of the guard time between beep and cycle_time change.
Could you test it again?
By the way: `cycle_time` changes are not queued, this means that you can't change the frequency as fast as the actual value (even in high throughput mode).
So in your beep-macros it will default to the "normal" operation.

KevinOConnor:
FYI, I suspect the fixes in #5975 will improve the results of this code.

-Kevin

Cirromulus:
Thanks, I will rebase and test it on my system. Perhaps it has something to do with [this issue](https://klipper.discourse.group/t/testers-wanted-laser-support-fast-pwm-updates/1771/44?u=cirromulus) where a user gets not-flushed elements after a `CANCEL_PRINT` and a timer-too-close when it moves again.

Cirromulus:
Rebased. Users report that this indeed fixes the `CANCEL_PRINT` issue.

Rneil-ca:
Hi Cirromulus,

Just wanted to say thank you for creating this fork.  I have it successfully working on a skr v1.4 board with a Chinese laser with a 5v pwm input. Just got it working tonight and so far it is working great.  Found your fork when I wasn't able to get my laser to burn properly with the standard version of Klipper.  

Thanks again for all the work you have put into this.

Ryan

Rneil-ca:
Any updates on this being merged into the main Klipper branch?  I have been using this for months and it works great.

Thanks,
ryan

naikymen:
Same here! It does require a bit of merging work though.

I merged the branch yesterday, but some functions have changed names in upstream Klipper since this PR.

I  don't understand that part of the code, so It's a bit beyond my capability to update it for now.

Cirromulus:
@Rneil-ca nice to hear. Do you use it for both printing and lasing purposes?

@naikymen Do you need a specific feature of the klipper master branch? The fork works well for me, and I grow tired of the continuous rebasing.

naikymen:
Hi @Cirromulus, thanks for the reply.

Yes but sort of... I need to use [this](https://github.com/naikymen/klipper-for-cnc/) fork, aimed at multipurpose CNC, which I keep in sync with upstream klipper. 

I keep it synced because I like the fixes and new features across the complete stack (i.e. MR and Mainsail). I do not plan to open more PRs here, as they are a lot of work and hard to get merged.

> I grow tired of the continuous rebasing.

Makes sense. I was still expecting that your work be merged to master, but if this is not the case then I'd humbly request one more rebase.

Some folks and I have been discussing klipper for more general CNC over [here](https://klipper.discourse.group/t/klipper-for-cnc-initiatives-and-projects-list/5698/), and yours is the only branch implementing laser support that I'm aware of.

I'd gladly support the rebase with a small donation if any of this makes sense to you.

Thanks for your work!

Rneil-ca:
> @Rneil-ca nice to hear. Do you use it for both printing and lasing purposes?
> 
This machine does both CNC and Laser, I use normal Klipper on my 3d printers.  So far this has been working great, so no need anything now.  Would be nice if this was permanently merged into the main branch so we can stay up to date, but not sure if they are willing to do that.

Thanks again for all your work on this!

Ryan

Cirromulus:
Rebased against current `master`.

> Some folks and I have been discussing klipper for more general CNC over [here](https://klipper.discourse.group/t/klipper-for-cnc-initiatives-and-projects-list/5698/), and yours is the only branch implementing laser support that I'm aware of.

Also cool to hear. I followed the thread a bit. Hopefully someone contributes a 'XY-only' kinematic for the folks without Z axis.

> I'd gladly support the rebase with a small donation if any of this makes sense to you.

Thank you very much, but I want to give back to the FOSS-community when I find the time :)

Cirromulus:
@Rneil-ca 
> This machine does both CNC and Laser, I use normal Klipper on my 3d printers.

The thing @KevinOConnor is most concerned about is that nothing breaks normal 3d printing operations.
So you could help us with using the fork on your actual printers as well, and then report back whether you encounter any problems or not (my preferred feedback :smile:).

ssendev:
It's been some time but when I tested it printing worked fine but it crashed KlipperScreen which was the reason i didn't leave it installed permanently.

c---:
I haven't been following the discussion on this. Why can't this be driven like an additional stepper instead of this complex method? That would seem more "correct" to me, and faster.

naikymen:
> Also cool to hear. I followed the thread a bit. Hopefully someone contributes a 'XY-only' kinematic for the folks without Z axis.

I think this is working already, there is a config draft over here: https://gitlab.com/pipettin-bot/forks/firmware/klipper-stack/-/blob/pipetting/printer_data/config/cnc_shield_xy/printer.cfg#L19

We're making a new machine so I haven't had much time to test these features.

Thanks **a lot** for rebasing, I've merged the branch without conflicts. :partying_face: 

Over here: https://github.com/naikymen/klipper-for-cnc/tree/develop-pwm_sync_channel-merge

Rneil-ca:
> The thing @KevinOConnor is most concerned about is that nothing breaks normal 3d printing operations. So you could help us with using the fork on your actual printers as well, and then report back whether you encounter any problems or not (my preferred feedback 😄).

Thanks for the rebase, will install the latest on both of my printers and my CNC and will let you know if there are any issues.

Thanks again for all your work on this.

Ryan

HWiese1980:
I've been running this change for about a year now on a Raspi Zero 2 and BTT Octopus without any issues. Printing works like a charm, as usual, just like using the laser engraving feature with a laser attached to a fan port. Never encountered any problems. 

Please merge!

KevinOConnor:
The functionality in this PR should now be on the main branch (with the merge of PRs #6369, #6410, #6420).

-Kevin

PR #4131, opened 2021-04-01
bed_mesh:  lookup the toolhead object in "handle_connect"

### LINKED ISSUE #4129: Klipper 0.9.1.-393 broke printer connection ("none type" has no attribute "get_position")
I think Klipper got accidentally broken during last nights update to 0.9.1-393.  I was on a previous version, just finished printing, then updated and got "Internal error during ready callback: 'NoneType' object has no attribute 'get_position'" and cant connect to the printer anymore.  Everything else seems up-to-date (see KIAUH screenshot).  Please find Klippy.log attached.  Let me know if I did something wrong or how I could revert back (sorry not very familiar with git commands).
Screenshot 2021-04-01 073105
Screenshot 2021-04-01 073205
[klippy.log](https://github.com/KevinOConnor/klipper/files/6241243/klippy.log)
[moonraker.log](https://github.com/KevinOConnor/klipper/files/6241245/moonraker.log)

Arksine:
This is a bugfix that resolves #4129.

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -59,6 +59,8 @@ class BedMesh:
         self.printer = config.get_printer()
         self.printer.register_event_handler("klippy:ready",
                                             self.handle_ready)
+        self.printer.register_event_handler("klippy:connect",
+                                            self.handle_connect)
         self.last_position = [0., 0., 0., 0.]
         self.bmc = BedMeshCalibrate(config, self)
         self.z_mesh = None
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I committed this change.  Is it necessary to use the "klippy:ready" event at all - could we do everything from the "klippy:connect" event?

-Kevin

Arksine:
I think would could do it all from the connect event.  I know we can log the generated points from there, and now that gcode_move calls `reset_last_position` it shouldn't be a problem to load the profile there.

KevinOConnor:
Okay, thanks.  I went ahead and made that change (commit b5b2fe2d).

-Kevin

PR #4132, opened 2021-04-01
make the miso pin optional

le-Bark:
Make the miso pin optional for software spi.

### DIFF
--- a/klippy/extras/bus.py
+++ b/klippy/extras/bus.py
@@ -118,8 +118,17 @@ def MCU_SPI_from_config(config, mode, pin_option="cs_pin",
     if config.get('spi_software_sclk_pin', None) is not None:
         sw_pin_names = ['spi_software_%s_pin' % (name,)
                         for name in ['miso', 'mosi', 'sclk']]
-        sw_pin_params = [ppins.lookup_pin(config.get(name), share_type=name)
-                         for name in sw_pin_names]
+        sw_pin_params = []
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks for looking at this.  Some comments:
1. I think this will require micro-controller support.  I don't think it's valid to configure a pin as both input and output on all the different micro-controllers.  I think it would be preferable if the src/spi_software.c module detected this case and handled it explicitly.  (For example, it could detect that MISO pin is equal to the SCLK pin and set appropriate flags so that the MISO pin was never actually used).
2. The emulated st7920 and hd44780_spi don't need MISO, and the max6675 doesn't need MOSI (klippy/extras/spi_temperature.py).  So, if doing this, it would be nice to also support read-only software spi channels.
3. I think in the general case the code should continue to raise an error if the omits a MISO pin.  Only those devices that don't use a MISO pin should have that ability, and on those devices I think it should not be possible to specify a software MISO pin.

-Kevin

le-Bark:
I am trying to evaluate the scope of this.
Should I limit this modification to the software spi implementation or also include the hardware?
Including the hardware would also require to redo how the pins are defined.

KevinOConnor:
> Should I limit this modification to the software spi implementation or also include the hardware?

If it was me, I'd focus on the software spi implementation.

-Kevin

KevinOConnor:
Any further updates on this?

-Kevin

PR #4133, opened 2021-04-02
Config: Add Generic-Duet3-Mini.cfg Board Config

lukeslaboratory:
- Adds Duet3 Mini 5+ Config with support for expansion header
- Includes instructions on how to flash MCU with stock bootloader

Signed-off-by: Luke Ashley Luke@Lukeslab.online


KevinOConnor:
Thanks!  Nice work!

Some comments:
1. Can you test with the appropriate clock crystal defined?  It shouldn't be necessary to use "internal clock" and using a crystal does provide better clock predictions.  (It's desirable when using multi-mcu with Klipper.)
2. I think this board should be using the adc_scaled system ( https://www.klipper3d.org/Config_Reference.html#adc_scaled ).  See https://github.com/KevinOConnor/klipper/blob/master/config/generic-duet2-maestro.cfg for an example configuration using it.
3. I noticed a few things in the config (eg, defining pressure_advance) that don't make sense for an example config.  There's some new guidelines on example configs at: https://www.klipper3d.org/Example_Configs.html
4. Do you know what bootloader is on the board?  I'm surprised that we can't use one of the existing USB flash scripts to make it work.

Thanks again,
-Kevin

lukeslaboratory:
Thanks for the feedback! 
1) I'll give it a shot and look for a crystal. 
2) Ok. Will try.
3) Pressure advance was an oopise on my part, I thought i stripped most of the 
4) https://github.com/Duet3D/uf2-samdx1/tree/master/boards <- is my bet. [Duet's documentation](https://duet3d.dozuki.com/Wiki/Duet_3_Mini_5plus_Firmware#Section_Programming_the_bootloader) says that the bootloader is locked - I tried flashing a few other things related to the bootloader, but nothing changed related to flashing new firmware, so I interpreted it as actually locked.

ReXT3D:
Curious if there has been any progress?  I'm watching this pretty much daily since I have a Duet 3 Mini in the process of being installed in my CR-10S Pro.

ReXT3D:
With respect to item 4, I can report that using `make flash FLASH_DEVICE=/dev/ttyACM0` worked perfectly on my board via USB, after quickly double-tapping the reset button.  There was no need for a .UF2 conversion or manual copying of the file to the boot loader "drive".  The flash process was very fast compared to my old Creality board.

lukeslaboratory:
All updates requested have been made. Thanks to Desuuuuuuu There is external clock information. Thanks to REXT3D for testing driver currents to prove that the documentation for the TMC2209 is very unclear.

1) 25Mhz crystal option added, with overclocks up to 200mhz. Duet overclocks to 180Mhz in RRF.
2) Includes ADC Scaled and implements it in the default heater and extruder
3) I have done my best to strip out all unnessessary items. Deviations nessessary: ADC Scaled, Sense_resistor, TX_Pin, Select_pin, and I left in two fans. 
4) Yep, `make flash FLASH_DEVICE=/dev/ttyACM0` works. Erased all the long writeup I did and kept the instructions.

Big thanks to the whole community for the assist!

KevinOConnor:
Thanks.  The 25Mhz clock changes look good and can be committed separately from the config file.  Can you confirm you are signing off on the clock changes (or @Desuuuu is) - as described in https://github.com/KevinOConnor/klipper/blob/master/docs/developer-certificate-of-origin ?

-Kevin

lukeslaboratory:
> Thanks. The 25Mhz clock changes look good and can be committed separately from the config file. Can you confirm you are signing off on the clock changes (or @Desuuuu is) - as described in https://github.com/KevinOConnor/klipper/blob/master/docs/developer-certificate-of-origin ?
> 
> -Kevin

Yes. it will be option C for the clock code, given by @Desuuuu "under" option A (og code).

Do you want a seperate PR for the clock changes? or are you just saying that they can be seperately merged?

Thanks!

KevinOConnor:
Thanks.  If it's Option C and Option A, then I would like to also record Desuuuu's signed-off-by line in the commit.  If there is a concern with this, and you'd like to discuss it off-line, feel free to email me at kevin@koconnor.net .

-Kevin

KevinOConnor:
I added the 25Mhz crystal option with my signed-off (commit bd65c37e) and also added the example config with some minor changes (commit 080b7f68).

Thanks,
-Kevin

arrowcircle:
It looks like this pull request does not add can support. Any change of having support of can bust with toolboard 1lc support?

PR #4134, opened 2021-04-03
stm32f1: reset peripherals in enable_pcclock()

Arksine:
Hi Kevin, as previously discussed this patch resets the peripheral in `enable_pcclock`.  This fixes hardware SPI on the BTT SKR Mini (and its derivatives), where the stock bootloader does not properly disable SPI after use.

SIgned-off-by:  Eric Callahan <arksine.code@gmail.com>


KevinOConnor:
Thanks.

-Kevin

PR #1356, opened 2019-03-04
probe: Add support for probe slow down

nythil:
Added an option to increase probing accuracy by first doing a fast
probe, backing off, and then performing a slow probe. This allows
to set a reasonable probe speed for bed leveling while retaining
the accuracy of a slow probe speed.

This a solution for #1297 and a partial solution for #1249.

PR #4135, opened 2021-04-03
probe: Add pre-probing move

dalegaard:
Currently when probing, all probing moves happen at the same speed, including the initial probe which may be a much longer move than the subsequent move on multi-sample probing. Even for single-sample probing, picking a slow speed to ensure accuracy can end up resulting in very long probing moves.

This adds "preprobing" moves to the probing code. If enabled, a single probing is performed at a different(usually higher) speed, a small retraction is made, and then the standard probing proceeds as usual. The result of the pre-probe is not part of the determined Z height.

This was implemented already in #1356 (in a slightly different way) but sadly wasn't finished and merged.

Using my standard bed mesh setup:
```
[probe]
...
z_offset: 7.5
speed: 2
samples: 3
samples_tolerance: 0.050
samples_tolerance_retries: 3
sample_retract_dist: 2
preprobe_speed: 20
lift_speed: 20

[bed_mesh]
speed: 400
horizontal_move_z: 12.5 # probe z_offset + 5mm
mesh_min: 15,15
mesh_max: 220,215
probe_count: 5,5
```
I went from 214 to 193 seconds(-9.8%) with no loss in precision for a full bed mesh, and can now reduce my probing speed a lot without suffering from very long bed mesh runs.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1551,6 +1551,18 @@ z_offset:
 #   not obtained in the given number of retries then an error is
 #   reported. The default is zero which causes an error to be reported
 #   on the first sample that exceeds samples_tolerance.
+#preprobe_speed: 0
+#   If set (in mm/s), enables pre-probe moves. Every probing sequence
+#   will start out with a single probing with this speed, after which
+#   the probe will retract a small distance and then the standard
+#   probing sequence (usually done at a slower speed for more accuracy)
+#   will run. Pre-probing does not measure Z height.
+#preprobe_retract_dist:
+#   The distance (in mm) to retract after the pre-probing move, if
+#   enabled. Defaults to the value of sample_retract_dist.
+#preprobe_lift_speed:
+#   The speed (in mm/s) to retract after the pre-probing move, if
+#   enabled. Defaults to the value of lift_speed.
 #activate_gcode:
 #   A list of G-Code commands to execute prior to each probe attempt.
 #   See docs/Command_Templates.md for G-Code format. This may be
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  However, unless I'm missing something, users can already obtain similar behaviour by defining their own probing macro that first calls something like `PROBE PROBE_SPEED=20` followed by the real `PROBE` call.

As such, I'm not sure this PR is a good match for the master Klipper branch.

-Kevin

dalegaard:
Hi Kevin,

Is it possible to use such a macro with the bed mesh calibration system? From my reading of the code, it directly uses the probing module interface, and does not go through the gcode execution path. As such, I can't see how such a macro would help?

For a single use of `PROBE` I completely agree, but the big one for me has definitely been bed mesh and everything else that uses the `ProbePointsHelper` class.

Best regards

KevinOConnor:
Ah, I thought you were saying you wanted to improve the time it takes for the initial dive.  No, it's not possible to override each bed mesh probe using a macro.

In that case, though, it is already possible to reduce the probing time by using a lower horizontal_move_z.

-Kevin

dalegaard:
Hi Kevin

Yes, lowering `horizontal_move_z` is an option as well for bed mesh, but for e.g. the gantry leveling modules, where the bed may be off by several millimeters this doesn't seem viable.

As you can see in the test config, I'm using a 5mm `horizontal_move_z`, which I believe is the default.

Even in the case of a low `horizontal_move_z` though, some probes are best at very low speed(the linked stale PR is suggesting 1 mm/s or lower for the IR probe) and this change could save significant time for these without having to resort to very low horizontal_move_z values.

Best regards

KevinOConnor:
I understand.  However, I don't currently think this is a good candidate for the Klipper master branch.

-Kevin

dalegaard:
:+1:

PR #4137, opened 2021-04-03
resonance_tester: Fixed multi-point resonance testing

dmbutyugin:

### DIFF
--- a/klippy/extras/resonance_tester.py
+++ b/klippy/extras/resonance_tester.py
@@ -178,7 +178,10 @@ class ResonanceTester:
                             "%s-axis accelerometer measured no data" % (
                                 chip_axis,))
                 new_data = helper.process_accelerometer_data(chip_values)
-                data = data.join(new_data) if data else new_data
+                if data is None:
+                    data = new_data
+                else:
+                    data.add_data(new_data)
         if csv_output:
             csv_name = self.save_calibration_data('resonances', name_suffix,
                                                   helper, axis, data)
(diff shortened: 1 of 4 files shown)


Sineos:
Is this fix of any concern WRT previous measurements?

dmbutyugin:
No, this is a fix for the cases when multiple `probe_points` were configured - currently `TEST_RESONANCES` command simply fails due to a bug.

klippy/extras/resonance_tester.py
```diff
@@ -178,7 +178,10 @@ def cmd_TEST_RESONANCES(self, gcmd):
                             "%s-axis accelerometer measured no data" % (
                                 chip_axis,))
                 new_data = helper.process_accelerometer_data(chip_values)
-                data = data.join(new_data) if data else new_data
+                if data is None:
+                    data = new_data
+                else:
+                    data.join(new_data)
```
KevinOConnor, line 184: Should this be `data = data.join(new_data)`?

-Kevin

dmbutyugin: No, this is exactly the bug I am trying to fix. join method joins in place and does not return anything.

KevinOConnor: Ah, okay.  The standard python string has a `join()` method which returns the new value.  Perhaps we should rename `CalibrationData.join()` to `CalibrationData.extend()` so that it better matches the Python method naming scheme?

-Kevin

dmbutyugin: I renamed the method to `add_data`.

KevinOConnor:
Thanks.

-Kevin

PR #4142, opened 2021-04-05
bed_mesh: introduce "faulty_region" options

Arksine:
This pull request adds the ability for users to specify "faulty regions" on the bed that cannot be accurately probed.  The specific use case for this is to improve the results for beds with integrated neodymium magnets, as magnets tend to change the trigger distance of an inductive probe.  Users can define "faulty regions" where the magnets are located.  If a generated point lies within one of these regions, bed_mesh will replace that point with up to 4 points at the boundaries of the region.  When probing is complete the Z values of the replacement points are averaged, and the result will be inserted into the mesh at the X,Y location of the point they replaced.  The included illustration shows how bed mesh would replace points at various locations in the mesh.


KevinOConnor:
Thanks.  Let me know when you are ready to commit and I will do so.

-Kevin

Arksine:
Hi Kevin, I believe this is good to go.  Thanks!

KevinOConnor:
Thanks.

-Kevin

PR #4145, opened 2021-04-06
add china mirrors for tsinghua university

mirokymac:
pypi, debian and raspbian will use tsinghua mirror while klipper is installing.
speed up the install process in china mainland.


raymondh2:
I'm of the opinion that this isn't needed. You should be setting up your apt repos on your own. Not relying on Klipper's install scripts to do so.

KevinOConnor:
As mentioned above, I don't think this is a good candidate for the master Klipper branch.  The scripts here are intended as examples for installing Klipper - not for configuring the operating system.

-Kevin

PR #4147, opened 2021-04-07
Create printer-creality-cr6se-4-5-3.cfg

cbluoss:
New config for more fun with CR-6 SE and klipper


KevinOConnor:
Example configs in the Klipper repo should follow the guidelines described at: https://www.klipper3d.org/Example_Configs.html

Also, if you'd like to share a more custom config, we recently introduced a section to the Klipper Discourse server for that ( https://community.klipper3d.org/ ).

-Kevin

cbluoss:
This is very much the 2020 config with updated heater pins for the rev 4.5.3 board (produced since late 2020)

Named the cfg 2021 as most units with this board should be sold since 2021.

KevinOConnor:
Okay - the config is fine, but commits need to have a signed-off-by line for me to commit directly - see https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .

-Kevin

cbluoss:
Any further changes needed?

KevinOConnor:
Thanks.

-Kevin

PR #4150, opened 2021-04-08
RGB support for the mightyboard

le-Bark:
Add support for the pca9533 led driver ic.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2471,6 +2471,24 @@ clock_pin:
 #initial_BLUE: 0.0
 #   See the "neopixel" section for information on these parameters.
 ```
+## [PCA9533]
+PCA9533 LED support. The PCA9533 is used on the mightyboard.
+
+```
+[pca9533 my_pca9533]
+#i2c_bus:
+#    The i2c peripheral, "twi" on the atmega
+#i2c_address: 98
+#    The i2c address, 98 or 99
+#initial_RED: 1
+#initial_GREEN: 1
+#initial_BLUE: 1
+#initial_WHITE: 1
+#    The PCA9533 only supports 1 or 0, any other values will be considered as 1.
+#    On the mightyboard, the white led is not populated.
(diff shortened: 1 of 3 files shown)


docs/Config_Reference.md
```diff
@@ -2471,6 +2471,24 @@ clock_pin:
 #initial_BLUE: 0.0
 #   See the "neopixel" section for information on these parameters.
 ```
+## [PCA9533]
+PCA9533 LED support. The PCA9533 is used on the mightyboard.
+
+```
+[pca9533 my_pca9533]
+#i2c_bus:
+#    The i2c peripheral, "twi" on the atmega
```
KevinOConnor, line 2480: Please use the boiler-plate text that the other i2c devices use:
```
#i2c_mcu:
#i2c_bus:
#i2c_speed:
#   See the "common I2C settings" section for a description of the
#   above parameters.
```

docs/Config_Reference.md
```diff
@@ -2471,6 +2471,24 @@ clock_pin:
...
+#i2c_address: 98
+#    The i2c address, 98 or 99
```
KevinOConnor, line 2482: If there is a default, please state what the default is.  Otherwise state that "this parameter is required".  See the other config examples (eg, see https://www.klipper3d.org/Config_Reference.html#mcp4728 ).

KevinOConnor (commented):
Thanks.  In general it looks fine to me, but I have a few minor comments.  See below.

-Kevin

KevinOConnor:
Thanks.  I committed this change with a minor commit on top (commit cacf1197).

Note, though, that the timing of updates on this implementation may be confusing.  The SET_LED command will apply the change immediately, instead of in order with other toolhead updates.  The code should probably obtain a print_time, convert it to a clock, and pass it as a minclock to `self.i2c.send()` for proper timing.  See klippy/extras/dotstart.py for example code that does that.

-Kevin

PR #4156, opened 2021-04-09
Config for the flashforge-creator-pro

le-Bark:

### DIFF
--- a/config/printer-flashforge-creator-pro-2018.cfg
+++ b/config/printer-flashforge-creator-pro-2018.cfg
@@ -1,0 +1,175 @@
+# This file contains common pin mappings for the FlashForge-Creator-Pro
+# To use this config, the firmware should be compiled for
+# the Atmel atmega2560.
+# Use 57600 baud for the bootloader
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[stepper_x]
+step_pin: PF1
+dir_pin: !PF0
+enable_pin: !PF2
+microsteps: 16
+rotation_distance: 33.99
+endstop_pin: ^!PL1
+position_endstop: 116
+position_max: 116
+position_min: -116
+homing_speed: 50
+
+[stepper_y]
+step_pin: PF5
+dir_pin: !PF4
+enable_pin: !PF6
+microsteps: 16
+rotation_distance: 33.99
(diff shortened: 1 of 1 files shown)


config/printer-flashforge-creator-pro-2018.cfg
```diff
@@ -0,0 +1,175 @@
+# This file contains common pin mappings for the FlashForge-Creator-Pro
+# To use this config, the firmware should be compiled for
+# the Atmel atmega2560.
+# Use 57600 baud for the bootloader
```
KevinOConnor, line 4: Thanks.  In general this looks good to me.  One comment though - can you elaborate on what `Use 57600 baud for the bootloader`.  I'm leery of mentioning "baud" without clarification, because for some strange reason people tend to set the Klipper baud rates to incorrect values.

-Kevin

le-Bark: Yes, I think that the best way would be to an an optional BAUDRATE parameter in the makefile and specify the command in the config file.
I will edit the baud comment and create a PR for the makefile.

-Marc-André

KevinOConnor: As a general rule, we don't add flash rules to the build code (there are just too many different ways to flash various boards).  If you have a specific flash command, I recommend you add the full command to the comments of the config file - for example, see: https://github.com/KevinOConnor/klipper/blob/master/config/generic-melzi.cfg

-Kevin

etet100:
What is the difference between your config and generic mightyboard? https://github.com/KevinOConnor/klipper/blob/master/config/generic-mightyboard.cfg

le-Bark:
The rotation distance, the stepper current settings and an additional extruder.

etet100:
I understand. However, I think both configurations were created for the same printer

le-Bark:
It is possible, what do you suggest?

KevinOConnor:
FWIW, I created the generic-mightyboard.cfg file for the controller board - not for any particular printer.

If this file supercedes that file, I'm fine with removing generic-mightyboard.cfg.

Separately, I'm fine with committing this PR.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #4157, opened 2021-04-10
z_thermal_adjust: Add Z thermal adjuster module

alchemyEngine:
## Context/Rationale
For some enclosed 3D printers, particularly larger-format machines, thermal expansion of the vertical frame members can continue for hours after the build plate and air temperature have stabilized. If the machine has not reached equilibrium when a print starts, the toolhead can slowly drift upwards while the frame continues to expand. This is of particular concern with long first layer times, and can even result in weak adhesion with the second layer.

Using MCU position and a nozzle-triggered endstop on my Voron V2 printer: 
image

Enclosed delta and moving gantry printers likely suffer most from this phenomenon.

## Approach
Couple a temperature probe to a vertical member of the frame and use the coefficient of linear expansion for the frame material to calculate the expansion relative to last homing measurement, and transform Z accordingly.

I've personally had success with this approach, and several members of the Voron community have tested it on their machines with positive results.

All comments, criticism, and feedback are very welcome! Thank you.

Signed-off by: Robert Pazdzior <robertp@norbital.com>


th3fallen:
Can confirm this works perfectly

KevinOConnor:
Thanks.  In general it looks fine to me.  I'd like to give a few more days for others to comment though.

I also saw some minor things:
1. Please format the changes to Config_Reference.md like the other existing sections (eg, do not add blank lines between fields).
2. The choice of coeff, frame_z_length, max_comp_z, etc, seem odd to me.  I wonder if it be simpler to accept config options like `temp1`, `z_offset_at_temp1`, `temp2`, `z_offset_at_temp2` and perform the coefficient calculations internally.
3. It would be preferable to keep the copyright at the top of new code.py type files the same as all the other files - 1 line summary, 1 blank line, 1 line copyright, 1 blank line, 1 line license statement.  Feel free to add any number of additional comment lines after the initial header block.
4. I'm not sure the value in the QUERY_FRAME_COMP command - wouldn't it be easier to export the info via `get_status()`?
5. Why track measured/maximum temperatures?

-Kevin

alchemyEngine:
Thanks for having a look. 

1. Will do.
2. This is generally not determined experimentally. The coefficient can be easily looked up, is a physical constant for a given material, and can be difficult to accurately measure without proper equipment.
3. Will do.
4. Good point, I should expose that information in `get_status()` as well. I think it also useful, however, to be able to monitor the current compensation (last homing reference temp, current offset, and state) while printing via terminal.
5. I believe this is necessary for [Moonraker compatibility](https://github.com/Arksine/moonraker/blob/4d1a2668f5c195e54229266ce8cebe053a5357b3/docs/printer_objects.md#temperature_sensor-sensor_name), but don't have it running personally to test.

-Rob

KoiosLabs:
I've been running this branch for a a couple months now and its fantastic. It seems to greatly improve my first layer's which have been a problem for me since voron 2.1...

th3fallen:
I've also had zero issues. I'd love to see it merged

KevinOConnor:
Thanks.  I agree it looks like a useful tool.  It would be great to get feedback on the type of printer and the specific configuration for each deployment that is currently using this PR.

-Kevin

KoiosLabs:
> Thanks. I agree it looks like a useful tool. It would be great to get feedback on the type of printer and the specific configuration for each deployment that is currently using this PR.
> 
> -Kevin

350^3 voron 2.4

th3fallen:
Voron 2.4 300^3

alchemyEngine:
> 
> 
> Thanks. I agree it looks like a useful tool. It would be great to get feedback on the type of printer and the specific configuration for each deployment that is currently using this PR.
> 
> -Kevin

Everybody I know currently using it is running a Voron V2. Since I only shared/discussed it on the Voron Discord, that's also probably the only source of testers unless somebody is motivated by this PR.

In theory, any enclosed printer in which the XY gantry is connected to the build surface through an appreciable length of aluminum frame would also be susceptible. The common fixed top gantry and descending lead-screw driven bed layout should be manifest the same expansion symptoms. 

Though here it gets a bit more complicated: 

- the lead-screw steel expansion coefficient is much smaller than that of aluminum
- the expansion is Z height dependant, as the length of screw between the bottom of the frame and bed varies
- the absolute compensation required would therefore be the different between the two

Since this is a more common arrangement (see Ender 5, TronXY X5, Hypercubes, Voron V1, etc). Perhaps implementation of configuration options for this two-factor expansion might broaden the applicability of the module?

glowtape:
That's an interesting idea. Does it run permanently, or just adjust actual print commands? I'm just asking, because I could swear I see some minor frame expansion when doing long bed probing (e.g. 5x5 multisampled).

robschwieb:
As a newer Voron V2.4 owner, this feature in invaluable. Not necessary for people only printing small footprint items but once you start printing full plates, absolutely necessary. I was seeing upwards of 0.04mm of drift due to frame expansion that was unable to be solved by heatsoaking alone. Something about printing that first layer heats up my 300mm V2.4 more than just hours of idle heatsoaking  is capable of.

This feature is a must have and made me go from hating my expensive and time consuming Voron to loving it again when printing full plates of parts.

Only thing it needs is a smoothing function to cut down on the unnecessary rapid fire z-stepping at times. Should be easy to implement.

klippy/extras/frame_expansion_compensation.py
```diff
@@ -0,0 +1,195 @@
...
+        'Called when Klipper successfully connected, ready.'
+        self.toolhead = self.printer.lookup_object('toolhead')
+        gcode_move = self.printer.lookup_object('gcode_move')
+
+        # Z stepper config check
+        kin = self.printer.lookup_object('toolhead').get_kinematics()
+        steppers = [s.get_name() for s in kin.get_steppers()]
+        if not self.z_stepper_name in steppers:
+            raise self.printer.config_error(
+                "'%s' is not a valid stepper" % self.z_stepper_name)
```
KevinOConnor, line 70: It's not valid to raise an error in a "klippy:ready" event handler (see docs/Code_Overview.md).  You probably want the "klippy:connect" event handler.

klippy/extras/frame_expansion_compensation.py
```diff
@@ -0,0 +1,195 @@
...
+    def check_eligible(self, endstops):
+        'Check if homed axis is `stepper_z`.'
+        steppers_being_homed = [s.get_name()
+                                for es in endstops
+                                for s in es.get_steppers()]
+        return self.z_stepper_name in steppers_being_homed
+
+    def handle_homing_move_end(self, hmove):
+        'Triggered when Z axis is homed.'
+        if self.check_eligible(hmove.get_mcu_endstops()):
```
KevinOConnor, line 101: This should catch the "homing:home_rails_end" event and look for `2 in homing_state.get_axes()`.

klippy/extras/frame_expansion_compensation.py
```diff
@@ -0,0 +1,195 @@
...
+            self.next_transform.move(corrected_pos, speed)
+            self.last_position[:] = newpos
+        self.last_position[:] = newpos
+
+    def temperature_callback(self, read_time, temp):
+        'Called everytime the thermistor is read'
+        self.last_temp = temp
+        if temp:
+            self.measured_min = min(self.measured_min, temp)
+            self.measured_max = max(self.measured_max, temp)
```
KevinOConnor, line 151: The `if temp` check doesn't make sense - it's always valid (or there is a severe error occurring).  Unless there is a need for it, I don't think this code should be tracking measure_min/measured_max.

KevinOConnor (commented):
Thanks.

I guess my main comment is that the configuration of this module seems more complicated than I think it could be.  Is there any reason we can't simplify this to:
```
[frame_expansion_compensation]
#adjustment:
#   The amount of Z adjustment (in millimeters per degree Celsius) to apply.
#sensor_type:
#sensor_pin:
#min_temp:
#max_temp:
#gcode_id:
#   See the "heater_generic" and "extruder" sections for the definition of these
#   parameters.
```

If I understand the code correctly, it seems to apply a Z offset based on the temperature difference since the last Z home.  If so, I think both the config and code could be simplified.

See some additional comments below.

-Kevin

oab1:
I have a question, as I have often wondered if this was some of the reason my second layers were failing on larger prints, also on a 350mm^3 voron v2. 

Why do you think the frame expansion effects the Z height of the print head? On the Voron, wouldn't it only increase the tension on the z belt loops? I dont see how that directly translates into the gantry rising. 

Your measurements speak for themselves, as do the positive results others are finding. I look forward to trying it myself. Just curious as to the reason why you think this is happening.

alchemyEngine:
> 
> 
> I have a question, as I have often wondered if this was some of the reason my second layers were failing on larger prints, also on a 350mm^3 voron v2.
> 
> Why do you think the frame expansion effects the Z height of the print head? On the Voron, wouldn't it only increase the tension on the z belt loops? I dont see how that directly translates into the gantry rising.
> 
> Your measurements speak for themselves, as do the positive results others are finding. I look forward to trying it myself. Just curious as to the reason why you think this is happening.

Hello,

The reason the tension increases is because the belts are stretching: the Z idlers are mounted to the top of the frame while the drive pulleys are mounted to bottom, therefore the expansion increases the belt length and consequently the tension. Because of the top mounted idler acting as a pulley, the movement gets reduced to 0.5x the increase in frame length (`gantry_factor: 0.5` for Voron V2s).

klippy/extras/frame_expansion_compensation.py
alchemyEngine: I'm not running Mainsail or other Moonraker-interfaced frontends, so I can't test this at the moment without some effort, but in so far as I understood min/max tracking was necessary for compatibility. It was a kind contribution from @fsironman necessary to get it running on his system.

klippy/extras/frame_expansion_compensation.py
alchemyEngine: > 
> 
> Thanks.
> 
> I guess my main comment is that the configuration of this module seems more complicated than I think it could be. Is there any reason we can't simplify this to:
> 
> ```
> [frame_expansion_compensation]
> #adjustment:
> #   The amount of Z adjustment (in millimeters per degree Celsius) to apply.
> #sensor_type:
> #sensor_pin:
> #min_temp:
> #max_temp:
> #gcode_id:
> #   See the "heater_generic" and "extruder" sections for the definition of these
> #   parameters.
> ```
> 
> If I understand the code correctly, it seems to apply a Z offset based on the temperature difference since the last Z home. If so, I think both the config and code could be simplified.
> 
> See some additional comments below.
> 
> -Kevin

Thanks for the review, Kevin.

In principal I agree, and would like to simplify both before finalizing.

I would like to keep the `coefficient` config parameter name, as this is consistent with the technical term, "coefficient of linear expansion". As well, the length of the frame member is a necessary input as the change is length is proportional to the total length: `l_total = l_ref + l_ref * (coeff * (t_current - t_ref)`

`gantry_factor` is a kind of stand-in for broader compatibility I'd like to introduce for top mounted fixed gantry/moving bed machine down the line. Hard-coding the 0.5 factor necessary for Voron V2 would make it specific to that machine, so until I can think of a nice way to take care of as many use-cases as possible I would like to keep this factor accessible in the config.

Is there any reason to omit the optional limiting safety parameters? Perhaps I could hard-code some safety checks in this regard. If the `min_temp` parameter is defined, will the ADC range error out in time to avoid crashing the head into the bed while trying to e.g. correct for thermal contraction of suddenly dropping to -80C?

alchemyEngine:
@KevinOConnor I'd like to do a `docs` write-up on the module's intended use-case, how it works, and how to set it up properly. Would this be welcome on your end?

klippy/extras/frame_expansion_compensation.py
fsironman: The frame_expansion_compensation class is more or less a **temperature sensor class** that can apply a Z-offset.
The feature itself will work without tracking measure_min/measured_max yes,
but wont work with any third party (in this case moonrake/mainsail)
for which it makes sense to display the temperature from frame_expansion_compensation in its UI.

Hence the same functions as the generic temperature sensor are included (https://github.com/KevinOConnor/klipper/blob/master/klippy/extras/temperature_sensor.py)

This will allow 3rd party pluging to display the temperature in its UI as if its a generic temperature sensor:
unknown
(below without it, on top proper display with including min/max change)

KevinOConnor:
> the length of the frame member is a necessary input as the change is length is proportional to the total length: l_total = l_ref + l_ref * (coeff * (t_current - t_ref)

If I understand this code correctly, if I define a printer with `coeff = 23.4`, `frame_z_length = 300`, `gantry_factor = .5`, and if I home at exactly 25 degrees - then the adjustment formula falls out to:
`offset = coeff * .000001 * frame_z_length * gantry_factor * (last_temp - last_home_temp)`

In this case, I think it would be preferable to define a single `adjustment = .003510` in the config and use an internal formula of `offset = adjustment * (last_temp - last_home_temp) - last_home_offset`.  I think both the code and config are easier to understand with that simpler math.

FWIW, with the current code, if I home at 25 degrees, and print a layer at 30 degrees it will apply an offset of .017550mm, and if I print another layer at 35 degrees it will apply an offset of .035100mm.  However, if I home at 30 degrees and print a layer at 35 degrees it will apply an offset of 0.0175705335mm.  It seems odd to me that going from 30-35 degrees when homing at 25 degrees would have a different offset than going from 30-35 degrees when homing at 30 degrees (.017550mm vs 0.0175705335mm).  It looks like a code error to me, but perhaps I'm missing something.

-Kevin

alchemyEngine:
> 
> 
> > the length of the frame member is a necessary input as the change is length is proportional to the total length: l_total = l_ref + l_ref * (coeff * (t_current - t_ref)
> 
> If I understand this code correctly, if I define a printer with `coeff = 23.4`, `frame_z_length = 300`, `gantry_factor = .5`, and if I home at exactly 25 degrees - then the adjustment formula falls out to:
> `offset = coeff * .000001 * frame_z_length * gantry_factor * (last_temp - last_home_temp)`
> 
> In this case, I think it would be preferable to define a single `adjustment = .003510` in the config and use an internal formula of `offset = adjustment * (last_temp - last_home_temp) - last_home_offset`. I think both the code and config are easier to understand with that simpler math.
> 
> FWIW, with the current code, if I home at 25 degrees, and print a layer at 30 degrees it will apply an offset of .017550mm, and if I print another layer at 35 degrees it will apply an offset of .035100mm. However, if I home at 30 degrees and print a layer at 35 degrees it will apply an offset of 0.0175705335mm. It seems odd to me that going from 30-35 degrees when homing at 25 degrees would have a different offset than going from 30-35 degrees when homing at 30 degrees (.017550mm vs 0.0175705335mm). It looks like a code error to me, but perhaps I'm missing something.
> 
> -Kevin

Sorry, I'm really not explaining this clearly and forgot to mention that the length of the frame at the time of homing is also calculated assuming the length provided in the config was measured at room temperature (25C)

KevinOConnor:
> the length of the frame at the time of homing is also calculated assuming the length provided in the config was measured at room temperature (25C).

Okay, but that calculation does not look correct to me.  In the example above, if I home at 25 degrees, and later put down a layer at 40 degrees followed by a layer at 41 degrees then that layer itself has a Z offset of .003510mm subtracted from it.  That makes sense as the Z stepper should move less because of heat expansion.  However, if I home at 30 degrees and later put down a layer at 40 degrees followed by a layer at 41 degrees then that layer has .0035141067mm subtracted from it.  It does not look correct to me that homing temperature is altering the impact of temperature later in the print.

Cheers,
-Kevin

lvitol:
A more accurate relationship between current_temperature and frame length may like this:
length = exp(coeff * 1e-6 * (current_temperature - ref_temperature)) * ref_length
ref_length is length of the frame measured at ref_temperature (for example: ref_length=300mm at ref_temperature=25C)

lvitol:
The offset computation does not need  homing temperature.

alchemyEngine:
> 
> It does not look correct to me that homing temperature is altering the impact of temperature later in the print.
> 

Actually, this is correct, and the reason for sampling the temperature at the time of Z-homing: longer objects expand to a greater (absolute) extent than shorter objects. If you home at 30c then your frame is already longer than it was at 25c, therefore grows more per degC/K than at 25c - hence the larger offset you mentioned. And since homing to Z-min essentially 'resets' the expansion by driving down past this expansion and re-referencing Z, we don't really care how much a 25c frame would have expanded from 40-41c but rather how much the 30c frame we homed to and are referencing is expanding. 

I hope that clarifies my considerations in laying out the calculation.

alchemyEngine:
> 
> 
> A more accurate relationship between current_temperature and frame length may like this:
> length = exp(coeff * 1e-6 * (current_temperature - ref_temperature)) * ref_length
> ref_length is length of the frame measured at ref_temperature (for example: ref_length=300mm at ref_temperature=25C)

I'm sorry but I don't see why this would be any *more* accurate. Could you please explain?

oab1:
> > I have a question, as I have often wondered if this was some of the reason my second layers were failing on larger prints, also on a 350mm^3 voron v2.
> > Why do you think the frame expansion effects the Z height of the print head? On the Voron, wouldn't it only increase the tension on the z belt loops? I dont see how that directly translates into the gantry rising.
> > Your measurements speak for themselves, as do the positive results others are finding. I look forward to trying it myself. Just curious as to the reason why you think this is happening.
> 
> Hello or z
> 
> The reason the tension increases is because the belts are stretching: the Z idlers are mounted to the top of the frame while the drive pulleys are mounted to bottom, therefore the expansion increases the belt length and consequently the tension. Because of the top mounted idler acting as a pulley, the movement gets reduced to 0.5x the increase in frame length (`gantry_factor: 0.5` for Voron V2s).

That doesn't answer my question. I agree the belt tension increases, and stated as much. The question is how much does the increase in tension raise the gantry? I would propose it would depend on the height of the gantry, ie if you are in the middle of your Z travel, your factor of 0.5x the rate of frame length extension makes good sense. However, if you are at z = 0 or z = max, that surely wouldn't still be the case, would it? I would state the most accurate factor should be a a function of % of z travel currently being used. ( total frame expansion in z  * %z height in use).

alchemyEngine:
> Why do you think the frame expansion effects the Z height of the print head? I dont see how that directly translates into the gantry rising.

You asked *why*, not how. The belts stretching with the frame and dragging the gantry with it is why. *How* is, however, a very good question and one that I'm honestly not entirely sure of the answer to. 

>  if you are in the middle of your Z travel, your factor of 0.5x the rate of frame length extension makes good sense. However, if you are at z = 0 or z = max, that surely wouldn't still be the case, would it?

I would expect the belt to stretch more or less uniformly, but I genuinely do not know. I'd like to look into this some more, and if an easy-to-use configuration for such compensation could be implemented (`belted: true|false`?) that would be great. However, it might be beyond the original intended use-case for this module: first/second layer consistency. 

If you're willing to look into the physics and provide some reading, that would be greatly appreciated, however. My main priority right now is cleaning up the code to Kevin's standards and implementing smoothing for the machine-gun microstepping sound.

lvitol:
> I'm sorry but I don't see why this would be any _more_ accurate. Could you please explain?

The difference is very small, ignore my comment.
I was trying to calculate frame length at temperate x, i found this:
frame_length_at_x_a = frame_length_at_ref_temperature * (1 + coeff * 1e-6 * (x - ref_temperature))
in your calculation:
frame_length_at_x_b = (frame_length_at_ref_temperature * (1 + coeff * 1e-6 * (homing_temperature - ref_temperature))) * (1 + coeff * 1e-6 * (x - homing_temperature))

**frame_length_at_x_a != frame_length_at_x_b**

then i found the difference is so small, we can just ignore.

oab1:
> > Why do you think the frame expansion effects the Z height of the print head? I dont see how that directly translates into the gantry rising.
> 
> You originally asked _why_, not how. The belts stretching with the frame and dragging the gantry with it is why. _How_ is, however, a very good question and one that I'm honestly not entirely sure of the answer to.
> 
> > if you are in the middle of your Z travel, your factor of 0.5x the rate of frame length extension makes good sense. However, if you are at z = 0 or z = max, that surely wouldn't still be the case, would it?
> 
> I would expect the belt to stretch more or less uniformly, but I genuinely do not know. I'd like to look into this some more down the road, and if an easy-to-use configuration for such compensation could be implemented (`belted: true|false`?) that would be great. However, it might be beyond the original intended use-case for this module: first/second layer consistency.
> 
> If you're willing to look into the physics and provide some reading, that would be greatly appreciated, however. My main priority right now is cleaning up the code to Kevin's standards and implementing smoothing for the machine-gun microstepping sound.

Yes, I am genuinely curious and think you are on to something important here, particularly for larger builds. I definitely appreciate your effort already on the subject. It has bothered me for awhile. 

Fair enough, I just need to do a bit of work to make sure the physics of what I'm suggesting pencil out on paper. My thinking is the relevant deflection / thermal expansion is the same regardless of belt or lead screw types of machines. Just that the thermal drift will be of the largest magnitude at the upper extreme of the travel. Which is to say the correction has to include current gantry height factored into the equation. I'll see what I can do.

viljamip:
While I agree we should take the pen and paper approach to work this out, based on my intuition and empirical tests, I would argue the displacement is evenly distributed along the belt. 

The marks are 10mm apart before stretching the rubber band. The tension should even out in the rubber band as the only forces are coming from the fixed ends. As the tension force is the same everywhere, so is the displacement. 

I did ignore gravity here. It actually increases the tension towards the bottom of the belt which also increases the displacement. However, the weight of the belt itself is so small that we can safely ignore effects of gravity altogether. 

Also the gantry adds tension to the part of the belt above it. I would think this is a separate thing from the stretch caused by thermal expansion of the frame. 

I might try to work this out on pen and paper and share it here for you to review. 

7E1BD380-A10D-455D-BB20-2CFED8DC7F41

alchemyEngine:
> The tension should even out in the rubber band as the only forces are coming from the fixed ends. As the tension force is the same everywhere, so is the displacement.

Thanks a lot for this nice rubber band experiment! I think based on this and some intuition, we can probably assume the `gantry_factor` is constant throughout the range of Z motion, or varies so little as to be negligible. But if anybody can present a compelling case to the contrary I'd be happy to hear it out.

oab1:
> > The tension should even out in the rubber band as the only forces are coming from the fixed ends. As the tension force is the same everywhere, so is the displacement.
> 
> Thanks a lot for this nice rubber band experiment! I think based on this and some intuition, we can probably assume the `gantry_factor` is constant throughout the range of Z motion, or varies so little as to be negligible. But if anybody can present a compelling case to the contrary I'd be happy to hear it out.

No! This is wrong. 

The tension and deflection throughout the rubber band (or belt or rod) is uniform. I cant think of any reason why it wouldn't be unless the distribution of material throughout it was non-uniform, which I don't think we need to worry about. That isn't the argument. The correction needing to factor in your travel along Z is true precisely because the thermal expansion is uniform through the material.

In the rubber band example, if you treat the bottom of the picture at your thumb as z=0, your positional error is increasing as you go up... Lets say you have 10 hash marks on your rubber band, 10 mm apart. Then you stretched the rubber band 10mm total. At your first hash mark, you would have 1 mm total deflection and therefore positional error = 1mm. At your second hash mark, you would have an additional 1mm of deflection, now your total position error = 2mm...etc. By the time you reach the top your positional error would be 10mm! While that deflection is a constant 10% through the rubber band, the position error is a function of how far away you are measuring from your 0 point. 

That is why it matters where the gantry is along its z travel when determining the correction factor based on thermal expansion. The thermal expansion will be uniform through the material, but the more material between where you are and where your zero is means more total displacement to account for. It actually is pretty simple mathematically to account for.

alchemyEngine:
> While the deflection is a consistent 10% through each section of the rubber band (1mm for each section), the positional error is increasing the further away you are measuring from your zero point.

Ah, yes, I think I see what you mean now. So the `gantry_factor` should scale linearly with Z height in the case of a V2 then?

oab1:
You could rewrite this:
``
offset = coeff * .000001 * frame_z_length * gantry_factor * (last_temp - last_home_temp)
``

To this:
``
offset = coeff * .000001 * gantry_height * (last_temp - last_home_temp)
``

alchemyEngine:
> 
> 
> Yes! Although it wouldn't be a factor as much as a ratio between current gantry height and total frame height, something like this:
> 
> `new_factor = gantry_height / frame_z_length`
> 
> Inserting that into your formula, you would find that the frame_z_length actually cancels out.
> 
> Old Formula:
> `offset = coeff * .000001 * frame_z_length * gantry_factor * (last_temp - last_home_temp)`
> 
> New Formula:
> `offset = coeff * .000001 * frame_z_length * [gantry_height / frame_z_length] * (last_temp - last_home_temp)`
> 
> Simplified:
> `offset = coeff * .000001 * gantry_height * (last_temp - last_home_temp)`
> 
> Actually a pretty nice simplification. That is because you don't actually need to calculate the expansion of the full height of the frame, then ratio it down to how much height the gantry is currently using. You can just calculate directly how much expansion is occurring between the print surface and the gantry. Which is just the height of the gantry itself.

While this is intuitive, and the logic seems sound, it does suggest that we need extremely little correction near Z-min. This runs counter to what I've seen and what I've heard from testers. One person even uses a gantry factor of 1.0 on his V2, and says it works much better than at 0.5.

oab1:
> > Yes! Although it wouldn't be a factor as much as a ratio between current gantry height and total frame height, something like this:
> > `new_factor = gantry_height / frame_z_length`
> > Inserting that into your formula, you would find that the frame_z_length actually cancels out.
> > Old Formula:
> > `offset = coeff * .000001 * frame_z_length * gantry_factor * (last_temp - last_home_temp)`
> > New Formula:
> > `offset = coeff * .000001 * frame_z_length * [gantry_height / frame_z_length] * (last_temp - last_home_temp)`
> > Simplified:
> > `offset = coeff * .000001 * gantry_height * (last_temp - last_home_temp)`
> > Actually a pretty nice simplification. That is because you don't actually need to calculate the expansion of the full height of the frame, then ratio it down to how much height the gantry is currently using. You can just calculate directly how much expansion is occurring between the print surface and the gantry. Which is just the height of the gantry itself.
> 
> While this is intuitive, and the logic seems sound, it does suggest that we need extremely little correction near Z-min. This runs counter to what I've seen and what I've heard from testers. One person even uses a gantry factor of 1.0 on his V2, and says it works much better than at 0.5.

Yes, I was just thinking that as well. I think that means we might be misdiagnosing the issue then. The cause might be something else, but still a function of the temperature you are measuring. Maybe better to continue with it as a 'factor' then, if that is the case since we don't know. Damn. Going to have to keep thinking about what the mechanism might be, then.

alchemyEngine:
I'm playing with the idea of an auto-calibration routine using the MCU homing position as a readout. If the relationship between frame temperature and delta_MCU position can be fit nicely and is reproducible, then in theory this might be a more universal, alternative solution.

KoiosLabs:
wouldnt it instead of not caring about frame height, need 2x frame height (or wherever height of pully is?) minus gantry height, since its suspended, and the belt stretch up and back down to motor is what we're compensating?

glowtape:
> While this is intuitive, and the logic seems sound, it does suggest that we need extremely little correction near Z-min. This runs counter to what I've seen and what I've heard from testers. One person even uses a gantry factor of 1.0 on his V2, and says it works much better than at 0.5.

Does the gantry height in all this consider the distance to the pulley of the Z-drive?

KevinOConnor:
> One person even uses a gantry factor of 1.0 on his V2, and says it works much better than at 0.5.

FWIW, if one is using a Voron2, one were to assume the frame expands, assume the belts do not expand, and assume that the gantry location is primarily determined by the belt running from the top of the gantry through the pulley and then back down to the stepper - then I think a gantry factor of 1.0 would make more sense.  Said another way, if the belt doesn't expand and expansion is taken up by slack in the belt not resisting gravity, then every mm of frame expansion requires a mm of stepper movement.

> I'm playing with the idea of an auto-calibration routine using the MCU homing position as a readout. If the relationship between frame temperature and delta_MCU position can be fit nicely and is reproducible, then in theory this might be a more universal, alternative solution.

I think that is a good idea.

On a Voron2, I wonder if one could home the Z (via the Z switch) at room temperature, then heat soak the frame and home again (to the Z switch).  Ideally, the absolute difference due to heat expansion could be measured.  Might be tricky to get sufficient accuracy from the switch though.

-Kevin

oab1:
> wouldnt it instead of not caring about frame height, need 2x frame height (or wherever height of pully is?) minus gantry height, since its suspended, and the belt stretch up and back down to motor is what we're compensating?

It shouldn't, because the tension should be equal throughout the whole belt loop (so it is being pulled down some amount by the increase in belt tension as well as up). But I think that is an interesting comment since it seems like it behaves the way you describe. I haven't checked in awhile, but maybe the tension is not equal throughout. When I check z-belt tension I always do it above the gantry, since it is usually down low at the homing position.

I'm thinking now that maybe the tension on the short section of belt that goes down from the clamp on the gantry to the drive pulley is "loose", since the preload from the weight of the gantry will keep that side tense. the other side could be slack as a result, especially on machines where the z-belt tension is not set sufficiently high enough. Then it would be exactly as you are saying, as if the belt loop was a loop at all, but as though the gantry was just hanging from one open ended belt...

oab1:
> > One person even uses a gantry factor of 1.0 on his V2, and says it works much better than at 0.5.
> 
> FWIW, if one is using a Voron2, one were to assume the frame expands, assume the belts do not expand, and assume that the gantry location is primarily determined by the belt running from the top of the gantry through the pulley and then back down to the stepper - then I think a gantry factor of 1.0 would make more sense. Said another way, if the belt doesn't expand and expansion is taken up by slack in the belt not resisting gravity, then every mm of frame expansion requires a mm of stepper movement.

Yes, that is putting it better than I did. Maybe this is the real mechanics of the z-drive, though not the intended function of the design.

oab1:
> > > One person even uses a gantry factor of 1.0 on his V2, and says it works much better than at 0.5.
> > 
> > 
> > FWIW, if one is using a Voron2, one were to assume the frame expands, assume the belts do not expand, and assume that the gantry location is primarily determined by the belt running from the top of the gantry through the pulley and then back down to the stepper - then I think a gantry factor of 1.0 would make more sense. Said another way, if the belt doesn't expand and expansion is taken up by slack in the belt not resisting gravity, then every mm of frame expansion requires a mm of stepper movement.
> 
> Yes! That is putting it better than I did. The more I think about it the more I realize this must be the way the z-drive is operating. It cannot equalize tension across the bottom turn of the belt because the stepper motor is holding the pulley and preventing it from rotating enough to do so. So the z-drive loops function more as open ended belts hanging the gantry, rather than a closed loop. The lower non-tension section just feeds and retracts the excess.
> 
> gantry_factor 1.0 it is! Excited because this was the most nagging issue I had with my v2, and will reduce the substantial time needed to spend heat-soaking, which would be great. Thanks again!

Correction: I thought about Kevin's comment some more, and I think the factor should actually be 2.0 to keep gantry position the same. For 1mm of expansion, the stepper motor will have to extend the belt length by 2mm (1mm of extra belt for each side of the pulley). Then, take into account the bed plate will rise by some amount as well with temperature relative to the z-drive, since there is a couple of inches separating them. You end up with something like KoiosLabs proposed:

gantry_factor = (frame height*2 - print_surface_height)*coeff

where print surface height is distance from print surface (z=0) to lower z-drive, and is a constant.

Only other pitfall I see is I'm sure this won't be perfect because I bet the spring rate of the belt changes with temperature, so the gantry sags down as the belts get hotter/less stiff. Could be why a factor of 1.0 is working well already. Also could be machine size dependent as well, as the frame expansion and belt stiffness might balance out to some degree on a 250mm^3 v2 or a particular frame size, for example. Maybe that is why maybe current 250mm printers don't notice this issue,

alchemyEngine:
Just wanted to give a quick update since I've been rather quiet the last few weeks.

I've come to the conclusion that the most direct and kinematic/printer agnostic approach to compensation will be some form of one time 'auto-calibration' routine. This would involve measuring the expansion of the frame empirically while heating using e.g. a Zmin endstop or a tool-mounted microswitch probe, fitting the data and using that to apply the compensation based on the frame temperature live.

I'm currently gathering data to test the feasibility of this approach and try to better understand the phenomenon. As an example, here's an overview plot from a data collection run using my 350^3 Voron V2:
image
Since the V2 uses a nozzle-triggered endstop pin right behind the bed, the initial upward movement at the beginning of the run is probably caused by the bed heating the extrusions on which it is mounted as well as the the endstop itself, expanding and/or bowing upwards. Such initial anomalies in the data will need to be considered.

The predicted (currently implemented) thermal expansion correlate semi-linearly with measured values, however, even on machines of very similar construction, they differ quite wildly in the slope of this fit. This slope is mathematically equivalent to the `gantry_factor` parameter in config, and could probably be used to adjust correction in lieu of the auto-calibration routine for now.
fit_comparison_github_08-06-2021

I'll keep collecting data trying to work out the best way to collect and analyze data for frame expansion correction in this module. Any and all feedback is very welcome and if anybody is really keen (and running Moonraker) I'll consider uploading the measurement script I'm currently using for data collection to a repo.

alchemyEngine:
@KevinOConnor I will need to use numpy or scipy for curve fitting in the auto calibration routine as the frame response is often not perfectly linear with the temperature sensor (the beta 3950 does get better if you define it with 3 points rather than a beta value, it seems though).

Would it be best to follow dmbutyugin's input shaper approach and break it out into a script?

tanaes:
In case it's useful to the discussion, I did some empirical testing of this feature on my Voron after calibrating based on @alchemyEngine's script. 

It performed really well, I think. The below test samples represent the 1st, 5th, and 17th of a grid of 25 printed under four conditions: from cold (top row) or 2-hour preheated (bottom row), and without (left column) or with (right column) the frame compensation enabled. 
 
comparison_per-run

You can see the full documentation of my tests [here](https://github.com/tanaes/whopping_Voron_mods/blob/main/docs/frame_expansion/frame_thermal_compensation_howto.md#print-tests).

alchemyEngine:
> 
> 
> It's hard for me to say what the best option is. If it's possible to use the existing coordinate_descent() and not introduce external dependencies then that would be preferable. Can you provide more details on what you are trying to fit the data to?
> 
> -Kevin

Thanks Kevin, I'll have a look at `coordinate_descent()`.

I'd like to fit frame temperature readings and endstop/probe trigger positions using linear regression. Possibly using a sliding window to find the optimal measurement range for fitting that avoids initial bed/endstop expansion (first 15-20min on Voron V2s, generally). Not to mention the statistical tools that `numpy` provides for assessing the quality of the data.

KevinOConnor:
@tanaes - nice write-up.  Thanks for sharing.

-Kevin

KevinOConnor:

Updating the Klipper dependencies is a possibility, but given the installation time increase for that and due to time constraints on my side, I don't think that is something that would be committed in 2021.

-Kevin

alchemyEngine:
> 
> 
> I'm not sure what the state of this PR is.
> 
> Updating the Klipper dependencies is a possibility, but given the installation time increase for that and due to time constraints on my side, I don't think that is something that would be committed in 2021.
> 
> -Kevin

Hi Kevin,
Measurements of thermal expansion have led down a rabbit hole of sorts involving bimetallic expansion and quite some theory crafting. These initial finding are (once again) well documented by @tanaes here: https://github.com/tanaes/whopping_Voron_mods/tree/main/extrusion_backers

Unfortunately,my own time constraints and uncertainty with how best to proceed with a widely applicable means of expansion measurement for calibration have temporarily stalled development of this PR though.

I am, however, unclear as to the issue with regards to numpy dependency for analysis of expansion calibration data; this is already a requirement for accelerometer measurements and subsequent input shaping calibration.

KevinOConnor:
Thanks.  I'm not sure what the state of this PR is.  If there is still interest in this topic then the PR will need to be updated and a reviewer will need to volunteer to review it ( https://www.klipper3d.org/CONTRIBUTING.html ).

-Kevin

arne182:
This PR would still be very useful. My 500mm ratrig v-core 3 does do at least 0.1mm of Z travel while heating from room temp to operating temp. This would help to avoid having to constantly adjust the Z height or have to wait till the extrusions are at temp.

jalfeld:
This PR is still useful and should be merged into the main line. I have been running the code as is for months with absolutely no issues, it has absolutely improved print quality on my 300mm^3 Voron 2.4.

alchemyEngine:
Hey Kevin. Apologies for the dead air. Life...
> I'm not sure what the state of this PR is

I believe the core functionality of the module is solid.

I hit some decision paralysis with respect to the best way to bundle a frame expansion measurement/analysis procedure, however. I now see measurements being covered by a separate configurable module and/or script accompanied by proper documentation. The key configuration parameter, `temp_coeff`, is very intuitive, measured in millimetres of apparent toolhead Z movement per Kelvin/degC of the provided sensor. As such, it could be measured a number of printer-specific ways or even estimated based on intuition then tuned to satisfaction.

I'd appreciate your take on this, and will engage with the Klipper community for their opinions outside of my usual Voron-centric hangout.

KevinOConnor:
So, you do wish to see this merged?

FWIW, the current code seems to contain a lot of "debugging code", deprecation checks, and "experimental" code.  If you want to implement a simple "z transformation based on temperature" module, then I think I should be able to review it.  The current code is more complex than I'm comfortable reviewing.

Separately, I'm not sure that frame_compensation.py is a good name for the module - as it seems a bit ambiguous.  Maybe something like z_temperature.py would be more clear?

-Kevin

dewi-ny-je:
> Separately, I'm not sure that frame_compensation.py is a good name for the module - as it seems a bit ambiguous. Maybe something like z_temperature.py would be more clear?

In the "files changed" I see "frame_expansion_compensation.py" which is not so ambiguous, but indeed it could be better defined as "compensation_frame_z_expansion.py"

alchemyEngine:
> So, you do wish to see this merged?

Yes. I'll tidy up the code as you mentioned and implement some additional g-code commands for better runtime control and tuning. I will update when it is ready for another look.
> Separately, I'm not sure that frame_compensation.py is a good name for the module - as it seems a bit ambiguous. Maybe something like z_temperature.py would be more clear?

Agreed. Perhaps "z_temp_comp.py"?

KevinOConnor:
FWIW, both "temp" and "comp" seem ambiguous to me.

-Kevin

alchemyEngine:
I find z_temperature to be non-specific, but z_temperature_compensation to be too long unless the commit message length only starts counting after the module name... Can we go halfway on z_temperature_comp?

Arksine:
Just spitballing, maybe `z_thermal_adjust.py`?

alchemyEngine:
I could live with that.

KevinOConnor:
It's fine with me.

-Kevin

docs/Config_Reference.md
```diff
@@ -1128,7 +1128,43 @@ the nature of skew correction these lengths are set via gcode. See
...
+#   Temperature sensor to use for frame temp measurement. Use full config
+#   section name without quoutes. E.g. temperature_sensor frame
+#smooth_time:
+#   Smoothing window applied to the temp_sensor, in seconds. Can reduce motor
+#   noise from excessive small corrections in response to sensor noise. The
+#   default is 2.0 seconds.
+#max_comp_z:
+#   Disables compensation above this Z height [mm]. The last computed correction
+#   will remain applied until the toolhead moves below the specified Z position
+#   again. The default is 0.0mm (always on).
```
dewi-ny-je, line 1157: If it's disabled above the specified Z, with "0.0" it should be always "off", not always "on".

docs/Config_Reference.md
```diff
@@ -1128,7 +1128,43 @@ the nature of skew correction these lengths are set via gcode. See
...
+#max_z_offset:
+#   Maximum absolute compensation that can be applied to the Z axis [mm]. The
+#   default is 99999999.0mm (unlimited).
+z_stepper:
```
dewi-ny-je, line 1161: What if there are multiple Z steppers? from this description I would not understand how to interpret this setting

docs/Config_Reference.md
```diff
@@ -1128,7 +1128,43 @@ the nature of skew correction these lengths are set via gcode. See
...
+2020 aluminum extrusion.
+
+See also: [extended g-code commands](G-Codes.md#Frame-Expansion-Compensation).
+
+```
+[frame_expansion_compensation]
+#temp_coeff:
+#   The temperature coefficient of expansion, in mm/K. For example, a
+#   temp_coeff of 0.01 mm/K will move the Z axis downwards by 0.01 mm for every
+#   Kelvin/degree celcius that the frame temperature increases. Defaults to 0.0,
```
dewi-ny-je, line 1145: It should be "Celsius".

dewi-ny-je (commented):
Minor things in the non-code parts

docs/Config_Reference.md
alchemyEngine: Good point, will switch to the 99999 default "always on" for clarity.

docs/Config_Reference.md
alchemyEngine: Indeed confusing, that's why this config option has been removed and is no longer required for the module internally (I'm just holding off on doc rework until after I'm happy with the code).

docs/Config_Reference.md
alchemyEngine: Thanks

docs/Config_Reference.md
dewi-ny-je: Then better write "The default is 999999 mm (when enabled, compensation active at every Z value).".
To avoid confusion between "temp_coeff" which states "enabled when >0", and this one which states "always enabled".

docs/Config_Reference.md
```diff
@@ -1131,7 +1131,40 @@ the nature of skew correction these lengths are set via gcode. See
 [skew_correction]
 ```
 
-## Customized homing
```
KevinOConnor, line 1134: Looks like this makes an unintended change to this section header.

docs/Config_Reference.md
```diff
@@ -1131,7 +1131,40 @@ the nature of skew correction these lengths are set via gcode. See
...
+
+See also: [extended g-code commands](G-Codes.md#z_thermal_adjust).
+
+```
+[z_thermal_adjust]
+#temp_coeff:
+#   The temperature coefficient of expansion, in mm/K. For example, a
+#   temp_coeff of 0.01 mm/K will move the Z axis downwards by 0.01 mm for every
+#   Kelvin/degree Celsius that the temperature sensor increases. Defaults to
+#   0.0, which applies no adjustment.
```
KevinOConnor, line 1149: Please use only Celsius in the description.  I find it very confusing to use different units for different settings, so the convention in Klipper is for distances to always be in mm, temperatures in Celsius, time in seconds, etc.  I understand Kelvin and Celsius are interchangeable in this context, but for consistency I think the documentation should only use Celsius.

docs/G-Codes.md
```diff
@@ -1202,6 +1202,30 @@ print.
...
+already applied, but will freeze the current adjustment value - this prevents
+potentially unsafe downward Z movement. Re-enabling can potentially cause upward
+tool movement as the adjustment is updated and applied. `COEFF` allows run-time
+tuning of the adjustment temperature coefficient (i.e. the `TEMP_COEFF` config
+parameter). `COEFF` values are not saved to the config.
+
+#### SET_Z_THERMAL_ADJUST_REF
+`SET_Z_THERMAL_ADJUST_REF [TEMPERATURE=<value>]`: Set the thermal adjustment
+reference temperature to the current reading, or to the value supplied with
+`TEMPERATURE`.
```
KevinOConnor, line 1227: I think it would be preferable to use a single command for all three of the above - SET_Z_THERMAL_ADJUST with no parameters just reports the current state, SET_Z_THERMAL_ADJUST with a REFERENCE_TEMPERATURE=123.4 sets the reference temperature, SET_Z_THERMAL_ADJUST with COEFF=567.8 sets the coefficient, etc.

klippy/extras/z_thermal_adjust.py
```diff
@@ -0,0 +1,195 @@
...
+        self.last_z_adjust_mm = self.z_adjust_mm
+        return [pos[0], pos[1], new_z, pos[3]]
+
+    def calc_unadjust(self, pos):
+        'Remove Z adjustment'
+        unadjusted_z = pos[2] - self.z_adjust_mm
+        return [pos[0], pos[1], unadjusted_z, pos[3]]
+
+    def get_position(self):
+        return self.calc_unadjust(self.next_transform.get_position())
```
KevinOConnor, line 125: I'm confused why this doesn't set `self.last_position`?

klippy/extras/z_thermal_adjust.py
```diff
@@ -0,0 +1,195 @@
...
+        try:
+            self.sensor = self.printer.lookup_object(self.temp_sensor_name)
+        except Exception as e:
+            msg = '''
+            %s\nUse the full config section name in the "temp_sensor" parameter,
+            e.g. "temp_sensor: temperature_sensor frame".
+            ''' % e
+            raise self.printer.config_error(msg)
+        else:
+            self.sensor.sensor.setup_callback(self.z_adj_temperature_callback)
```
KevinOConnor, line 73: It's not valid to "peak" into another class like this (see docs/Code_Overview.md ).  I don't undersand why this is being done.  Shouldn't the code instantiate a new temperature sensor like other temperature modules?

KevinOConnor (commented):
Thanks.  In general it looks fine to me.  I have a few comments below.

-Kevin

klippy/extras/z_thermal_adjust.py
alchemyEngine: Forgive my ignorance, is that just an expected property for a module like this?

alchemyEngine: Forgive my ignorance, is that just an expected property for a module like this?

klippy/extras/z_thermal_adjust.py
alchemyEngine: I legit forgot that I did this. It's a workaround; I originally implemented it as you described, but people complained they couldn't see it in their front end, use it with a temperature_fan, etc. I think `duplicate_pin_override` might be a viable workaround for those people now, no need to do it in here.

klippy/extras/z_thermal_adjust.py
KevinOConnor: A gcode transform must be able to both translate to a new coordinate system (`move()`) and report the current position in the old coordinate system (`get_position()`).  If some external process changes the position outside of a normal gcode `move()` call, then the system arranges for `get_postiion()` to be called so that all components in the gcode stream know the current position.

It seems the code is tracking a `last_position`, but it isn't resetting it on a `get_position()` call.  That seems suspect to me.

-Kevin

alchemyEngine: Thank you for the explanation, Kevin!

KevinOConnor:
What is the current status of this PR?

-Kevin

alchemyEngine:
Hi Kevin,

I've implemented most of your suggestions and corrections:

- [x] docs unit consistency & unintended header changes
- [x] Use SET_Z_THERMAL_ADJUST command for all gcode functionality
- [x] Instantiate own temp. sensor, don't peak into other class

Just a couple things to wrap up before another review, etc:

- [ ] Finalize all docs changes
- [ ] [Reset last_position on get_position call](https://github.com/Klipper3d/klipper/pull/4157#discussion_r868763346) 

I've been travelling for the past month, so I haven't been able to work on or test things. But within the next two weeks I should have those last things shored up.

alchemyEngine:
@KevinOConnor Sorry for the inactivity, got stuck on the code and then summer carried me away. I think I've addressed all of your previous points, and the PR should be ready for another look.

Thanks,
-alch3my

KevinOConnor:
Thanks.  Looks fine to me.  Can you confirm this is still ready to merge?  If so, I'll go ahead and merge.

-Kevin

alchemyEngine:
I'd say it's ready to go out the door, merge away! Thanks for all your help and the discussions along the way, Kevin.

KevinOConnor:
Thanks.

-Kevin

danorder:
Internal error during connect: 'stepper_z'   I assume the current implementation doesn't support delta kinematics?

PR #4160, opened 2021-04-10
config: Add example generic-bigtreetech-skr-2.cfg config

popshansen:
module: Generic config for the BigTreeTech SKR 2 config file

Contains sample of TMC2209 configuration.

Signed-off-by: Lars R. Hansen popshansen@hotmail.com


KevinOConnor:
Thanks.

-Kevin

PR #4161, opened 2021-04-10
Add stow_on_each_sample param for Probe

nlef:
I'm used to using this param with BlTouch clone, and i implemented it for generic probe ( BfpTouch micro)

Signed-off-by: Nick Lefskiy n.lefskiy@yandex.ru

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1515,11 +1515,10 @@ stepper_z config section.
 [probe]
 pin:
 #   Probe detection pin. This parameter must be provided.
-#stow_on_each_sample: True
-#   This determines if Klipper should command the pin to move up
+#deactivate_on_each_sample: True
+#   This determines if Klipper should execute deactivation gcode
 #   between each probe attempt when performing a multiple probe
-#   sequence. Read the directions in docs/BLTouch.md before setting
-#   this to False. The default is True.
+#   sequence. The default is True.
 #x_offset: 0.0
 #   The distance (in mm) between the probe and the nozzle along the
 #   x-axis. The default is 0.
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks, but I'm not sure this is a good candidate for the Klipper master branch.  The bltouch devices can use the `[bltouch]` config section.  Other devices can implement what they need using macros.  I'd prefer not to add new user facing options to save a small amount of time on rarely used devices.

-Kevin

nlef:
But BfpTouch ( also known as "poor man bltouch") is quite popular with users of inexpensive (like ender-3) or self-assembled printers. and it'saccuracy is higher. it is more reliable. also this option makes macros easier.

nlef:
maybe it would be more correct to make a separate BfpTouch module by analogy with Bltouch? in this case standart Probe module would have no changes.

KevinOConnor:
Okay, I'll give it some further thought.

Thanks,
-Kevin

docs/Config_Reference.md
```diff
@@ -1584,6 +1588,10 @@ control_pin:
 #pin_move_time: 0.680
 #   The amount of time (in seconds) to wait for the BLTouch pin to
 #   move up or down. The default is 0.680 seconds.
+#   This determines if Klipper should command the pin to move up
+#   between each probe attempt when performing a multiple probe
+#   sequence. Read the directions in docs/BLTouch.md before setting
+#   this to False. The default is True.
```
KevinOConnor, line 1594: This text change looks like a copy-and-paste error.

KevinOConnor (commented):
Okay, thanks.  On further thought I think we can add this change.  I did notice one error in the documentation though - see below.

-Kevin

docs/Config_Reference.md
nlef: I have kept the existing description of the parameter in the documentation. and renamed the parameter in the "probe" section to be logically separate.
I can rewert this back to same naming in both "probe" and "bltouch" sections

nlef: Sorry, my mistake. Made a commit with fixes.

KevinOConnor:
Thanks.

-Kevin

PR #4164, opened 2021-04-12
thermocouple: max31865: attempt to recover from faults

Xorlev:
I'm running a PT1000 attached to an Adafruit Max31865 board, and
struggled with noise from rapid extruder movements pushing the
voltage just outside of what the max31685 board expects. Even running
my RTD's wires outside the printer (a Voron 2.4) only helped somewhat.
Given that rapid movements (like (de)?retractions) are momentary, I
figured that ignoring the error and trying again would be sufficient to
continue on successfully.

When `rtd_consecutive_fault_limit` (default: 0, as it is implicitly today)
is set to a value greater than 0, all faults raised by the max31685 will be
ignored, logged, and the max31865 re-initialized using the original
configuration. Once the max consecutive failures are reached, both the 
firmware and host shut down as is the current behavior. Failures return
the last known good temperature value. The firmware is responsible for
tracking the overall failure count (and resetting said count when a
successful measurement is performed), with the failure count being sent
back to the host along with each response. The host performs the reset
of the max31865 board by re-sending the initial configuration command
(which includes the fault reset bit). Simply resetting the fault bit isn't
enough, as it doesn't automatically sample and you end up with the same
temperature value with Klipper none the wiser.

This is currently only implemented for the Max31865, as that's the only
board I own currently. The code in this PR has ~50 hours of print time, and
recovered from dozens of events. The log looks a little like this:

```
Max31865: recovered from fault (consecutive_faults=1, consecutive_fault_limit=3): Overvoltage or undervoltage fault
#output: thermocouple fault detected, consecutive faults: 1, allowed faults: 3
#output: thermocouple fault recovered, resetting fault count
```

when operating.

Signed-off-by: Michael Rose <elementation@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2034,6 +2034,17 @@ sensor_pin:
 #   The above parameters control the sensor parameters of MAX31865
 #   chips. The defaults for each parameter are next to the parameter
 #   name in the above list.
+#rtd_consecutive_fault_limit: 0
+#   The above paramter controls Klipper's retry behavior when
+#   handling faults raised by the MAX31865. When set to a value
+#   greater than 0, Klipper will ignore consecutive failures up
+#   to the limit, each time resetting the MAX31865 with initial
+#   configuration parameters. A successful measurement resets
+#   this counter. This can help with sensors in noisy electrical
+#   environments. Only set this if you're having problems with
+#   sporadic firmware shutdowns due to thermocouple failures, and
+#   do not set this too high (e.x. <= 5-10). Setting this too high
+#   could lead to thermal runaway.
 ```
 
 ## BMP280/BME280/BME680 temperature sensor
--- a/klippy/extras/spi_temperature.py
+++ b/klippy/extras/spi_temperature.py
@@ -4,7 +4,7 @@
 # Copyright (C) 2018  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import math
+import logging, math
 from . import bus
 
 
@@ -15,14 +15,17 @@ from . import bus
 REPORT_TIME = 0.300
 
 class SensorBase:
-    def __init__(self, config, chip_type, config_cmd=None, spi_mode=1):
+    def __init__(self, config, chip_type, config_cmd=None,
+                 spi_mode=1, consecutive_fault_limit=0):
         self.printer = config.get_printer()
         self.chip_type = chip_type
         self._callback = None
         self.min_sample_value = self.max_sample_value = 0
         self._report_clock = 0
+        self.consecutive_fault_limit = consecutive_fault_limit
         self.spi = bus.MCU_SPI_from_config(
             config, spi_mode, pin_option="sensor_pin", default_speed=4000000)
+        self.config_cmd = config_cmd
         if config_cmd is not None:
             self.spi.spi_send(config_cmd)
         self.mcu = mcu = self.spi.get_mcu()
@@ -41,8 +44,10 @@ class SensorBase:
         return REPORT_TIME
     def _build_config(self):
         self.mcu.add_config_cmd(
-            "config_thermocouple oid=%u spi_oid=%u thermocouple_type=%s" % (
-                self.oid, self.spi.get_oid(), self.chip_type))
+            "config_thermocouple oid=%u spi_oid=%u thermocouple_type=%s "
+            "consecutive_fault_limit=%u" % (
+                self.oid, self.spi.get_oid(), self.chip_type,
+                self.consecutive_fault_limit))
         clock = self.mcu.get_query_slot(self.oid)
         self._report_clock = self.mcu.seconds_to_clock(REPORT_TIME)
         self.mcu.add_config_cmd(
@@ -51,7 +56,12 @@ class SensorBase:
                 self.oid, clock, self._report_clock,
                 self.min_sample_value, self.max_sample_value), is_init=True)
     def _handle_spi_response(self, params):
-        temp = self.calc_temp(params['value'], params['fault'])
+        consecutive_faults = 0
+        if 'consecutive_faults' in params:
+            consecutive_faults = params['consecutive_faults']
+
+        temp = self.calc_temp(params['value'], consecutive_faults,
+                              params['fault'])
         next_clock      = self.mcu.clock32_to_clock64(params['next_clock'])
         last_read_clock = next_clock - self._report_clock
         last_read_time  = self.mcu.clock_to_print_time(last_read_clock)
@@ -120,7 +130,7 @@ class MAX31856(SensorBase):
     def __init__(self, config):
         SensorBase.__init__(self, config, "MAX31856",
                             self.build_spi_init(config))
-    def calc_temp(self, adc, fault):
+    def calc_temp(self, adc, consecutive_faults, fault):
         if fault & MAX31856_FAULT_CJRANGE:
             self.fault("Max31856: Cold Junction Range Fault")
         if fault & MAX31856_FAULT_TCRANGE:
@@ -192,7 +202,7 @@ MAX31855_MULT = 0.25
 class MAX31855(SensorBase):
     def __init__(self, config):
         SensorBase.__init__(self, config, "MAX31855", spi_mode=0)
-    def calc_temp(self, adc, fault):
+    def calc_temp(self, adc, consecutive_faults, fault):
         if adc & 0x1:
             self.fault("MAX31855 : Open Circuit")
         if adc & 0x2:
@@ -221,7 +231,7 @@ MAX6675_MULT = 0.25
 class MAX6675(SensorBase):
     def __init__(self, config):
         SensorBase.__init__(self, config, "MAX6675", spi_mode=0)
-    def calc_temp(self, adc, fault):
+    def calc_temp(self, adc, consecutive_faults, fault):
         if adc & 0x02:
             self.fault("Max6675 : Device ID error")
         if adc & 0x04:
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Interesting.  Can you elaborate on why the board is getting so many faults?  Is this an issue between the mcu and the max31685, or an issue between the max31685 and the temperature sensor?

At a high-level, I think it makes sense to have the code be more tolerant to failures.  However, I'd prefer less logic in the mcu code.  Perhaps a general "invalid count" could be implemented for both temperature out of range and faults (similar to how the src/adccmds.c has a general fault count - `range_check_count`).

-Kevin

Xorlev:
The problem exists somewhere between the temperature sensor and the max31685 as far as I can tell, and appears to be interference from the extruder running on my direct drive toolhead. Another fix would be shielded cable, but I didn't want to resort to that.

> However, I'd prefer less logic in the mcu code. Perhaps a general "invalid count" could be implemented for both temperature out of range and faults (similar to how the src/adccmds.c has a general fault count - range_check_count).

There's some necessary evil in that a max31865 has to be reinitialized after a fault, just clearing the fault bit isn't enough and someone needs to do that. An earlier revision of this code tried that, and ended up with a stuck (and incorrect!) temperature value. In this PR, the mcu propagates the failure state since the host is the only one who knows how to do it. The other option would be to teach the mcu to do the initialization, but that pushes more logic into the mcu.

I do think the suggestion to handle temp out of range _and_ faults together is good though, I can work on that, but I'd like to know how you'd prefer I proceed re: the reinitialization of the max31865 before I get too far down that road.

Thanks for the look, Kevin.

Michael

KevinOConnor:
> There's some necessary evil in that a max31865 has to be reinitialized after a fault, just clearing the fault bit isn't enough and someone needs to do that. An earlier revision of this code tried that, and ended up with a stuck (and incorrect!) temperature value. In this PR, the mcu propagates the failure state since the host is the only one who knows how to do it. 

Make sense.  I guess I was thinking that if the mcu reports a fault, the host code could react accordingly and effectively clear the fault before the mcu gets to X successive faults.  Thus, no particular mcu code is needed (other than checking for no more than X successive faults).  It's possible I'm missing something though.

-Kevin

Xorlev:
> Make sense. I guess I was thinking that if the mcu reports a fault, the host code could react accordingly and effectively clear the fault before the mcu gets to X successive faults.

Other than the diagnostic logging (which I can remove), this does exactly that: the mcu counts faults, resets faults if it receives a faultless response from the max31865, and forwards along the current fault count to the host. I could avoid forwarding that along, if you'd rather keep it less chatty.

KevinOConnor:
> I could avoid forwarding that along, if you'd rather keep it less chatty.

Yes, FWIW, I think it would be better to make the mcu code less chatty.

Cheers,
-Kevin

gjsmo:
I'd like to add that I've had similar issues with a Pt100 attached to a MAX31865, and I haven't been able to replicate the failure after the printer shuts down. In every case, manually measuring the RTD resistance gives a perfectly reasonable value. I had assumed it was a loose connection but wasn't able to find anything or reproduce it by vigorously wiggling (bordering on abusing) each connection. Noise would certainly make sense, and maybe this is an argument for using shielded wire. However, it does make sense that it should be possible to recover from a temporary fault.

Would it make sense to implement this as a more general check for all temperature sensors? Surely it's possible for a K type attached to a MAX31856, or even a standard thermistor to encounter a similar loose connection or temporary fault? Obviously if feedback is lost entirely for more than a second or so then it's important to shutdown the print as a matter of safety.

KevinOConnor:
Thanks.  Unfortunately, I haven't gotten a chance to review this fully, and it's likely going to be a week or so before I can.

One question I did have - is there a reason to pass `consecutive_faults` from the mcu to the host?  Wouldn't the host be able to effectively count the consecutive faults itself (if it cares)?

Also, I'm not sure it makes sense to have a configurable rtd_consecutive_fault_limit - as I can't see a user actually configuring this.  Could we just set it in the code to 3 and not burden the user with an unnecessary choice?

-Kevin

ReXT3D:
FWIW, my $0.02 here based on experience with complex critical systems integration and their fault tolerant design.

The approach proposed by @Xorlev is conceptually aligned with fault tolerant design principles of complex systems.  In general, one would establish a tolerance "bucket" that is either time or sample based, over which the system can tolerate lack of valid data.  Each sample of data would be then be flagged as either valid or not valid, based on some established criteria (in this case a max31865 fault).  The bucket would then be incremented (filled up) with each invalid sample and decremented with each valid sample.  The increment value and decrement value (or rates) would generally be different, again based on the specific parameter being monitored and the amount of tolerance the system has to the lack of this parameter.  An empty bucket is therefore a 100% healthy system, while a full bucket represents a failed interface or a failed parameter, leading to printer shutdown in this case.  In a situation like this, one could choose to increment the bucket by 20% per faulty sample (33% suggested by the author), let's say, and decrement the bucket by 10% with each valid  (non-faulty) sample (100% suggested by the author).  This affords a large degree of fault tolerance, while ensuring that progressively deteriorating malfunctions get identified and handled.  Of course the control algorithm needs to be aware of the data validity and either use an alternate source or last valid value for each invalid sample received, etc.

Distribution of this logic between system components, like the MCU & the Raspberry Pi in this case, would then depend on the system requirements & selected architecture, including computational margins, bus data rates & latencies, etc.

I apologize if all of this is obvious, but I thought that it could be helpful in this case or in general.

ReXT3D:
I think in the case of the hot end temperature it would be easy to tolerate up to 1 second of continuous temperature data loss before halting the print.  Klipper could use the last valid temperature reading for seeding the PID control loop and freeze the hot end PWM power at a value based on rolling average of a number of prior valid samples.  Full clearing of the fault condition could use half of the slope, equivalent to 2 seconds of fault free temperature samples...

Of course if the originally proposed 3 consecutive bad samples to halt and one good sample to reset works reliably, it is much simpler to implement...

henkyprayoga:
In my case, by touching one of PT1000 lead using bare hand could trigger undervoltage/overvoltage fault, or even sensor disconnected fault.

akshaal:
Xorlev, thanks for the patch. I hope it will be merged soon. I experience a similar problem with my Voron 2.4 and max31865. In my case it is just 1 error per more than 100 hours of printing, but if it aborts a 6 hours print in a middle, then I still feel bad about it. Marlin takes a similar approach (error budget) in the src/module/temperature.cpp and I considered to make a patch for Klipper myself, but it would be hard for me to test it with my 1 error per 100 hours..

KevinOConnor:

-Kevin

ReXT3D:
I very much hope that this PR or equivalent is implemented.  Some level of temperature fault tolerance is very much required.

Xorlev:
Life became pretty busy and haven't been able to devote any time to OSS efforts. However, I would like to get this PR merged, since it's clear there's interest.

> One question I did have - is there a reason to pass consecutive_faults from the mcu to the host? Wouldn't the host be able to effectively count the consecutive faults itself (if it cares)?

I saw this as a way to ensure the host didn't fall out of sync with the mcu, and could provide helpful logging. There's some paranoia on my part as well, ensuring that the host stops trusting the MCU's values and raises a sensible fault message.

> Also, I'm not sure it makes sense to have a configurable rtd_consecutive_fault_limit - as I can't see a user actually configuring this. Could we just set it in the code to 3 and not burden the user with an unnecessary choice?

Sure, that seems reasonable. I'll rebase the code on the master branch and remove the configuration. My thinking was that some folks would prefer to fail fast. WDYT about just defaulting it to 3, and allowing the choice to be made to increase/decrease it?

KevinOConnor:
> I saw this as a way to ensure the host didn't fall out of sync with the mcu, and could provide helpful logging. There's some paranoia on my part as well, ensuring that the host stops trusting the MCU's values and raises a sensible fault message.

Okay, but I think the code could be simpler.  I would think the mcu code could be changed to count excessive faults (or out of range) before issuing a shutdown.  I would think the host code could be changed to clear errors on a fault indicator.  If the fault can't be cleared, the mcu will shutdown after the configured limit; that shutdown will propagate normally.

> WDYT about just defaulting it to 3, and allowing the choice to be made to increase/decrease it?

FWIW, I regularly receive feedback that the config complexity is a stumbling block for many new users.  I would prefer to limit the config (and config documentation) to "high impact settings".  It doesn't seem to me that a user can reasonably tune this value, so I would suggest not complicating the documentation with it.  (To be clear, I would still make the mcu configurable from the host.)

-Kevin

BlackStump:
I have also run into random shutdowns with the Max31865 Overvoltage or Undervoltage fault
I am using the Fysetc PT100Stick
I used this PR and just finished a 9 hour print, looking through the log there was 3 instances of Max31865 faults no consecutive faults, so thankyou @Xorlev this certainly allowed me to finish a print.
@Xorlev  any updates on this PR?

ReXT3D:
Trying to bump this out of inactive as I think this is an important improvement.

Were you able to do any work on this @Xorlev ?  Thank you.

samtny:
I have this error when I scuff my feet and touch the frame of the printer (whether a print is in progress or not). Using a MAX31865 stepstick, 2-wire PT100, PTFE wire through chains to the heater block on a Voron V2.4, Octopus board. Would sure be nice to... ignore my little static electric mistakes in this case.

Taking a look at the code now, not sure if I can be helpful so far as "simplifying" it though.

Gilabite:
lay down an ESD floor mat in front of your printer, lay one down under your printer, wear a wrist strap, LIFT your feet when you walk LMAO

gjsmo:
Is it possible to reopen this? It sounds like the bulk of the work was completed and what's needed now is merely "cleanup".

ReXT3D:
FWIW, I made an attempt at addressing Kevin's earlier input in a commit referenced above that I made to "my branch" of Klipper.  But a word of warning - I don't really know anything about Python.  Having said that the changes seem to work well and as intended in my test environment.

PR #4167, opened 2021-04-13
Add example configs for fysetc spider board

GerogeFu:
Add example configs for fysetc spider board

### DIFF
--- a/config/generic-fysetc-spider-voron2.cfg
+++ b/config/generic-fysetc-spider-voron2.cfg
@@ -1,0 +1,542 @@
+# This file contains common pin mappings for the Fysetc Spider board.
+# To use this config, the firmware should be compiled for the STM32F446.
+# When calling "menuconfig", enable "extra low-level configuration setup"
+# and select the "12MHz crystal" as clock reference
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  In order for me to commit there would need to be a signed-off-by line in the commit message - see: https://www.klipper3d.org/Overview.html .  The change would also need to pass the build tests.

The generic-fysetc-spider-voron2.cfg wouldn't be a good candidate for the master Klipper branch (as we only commit "stock" printer configs to the github repository).  The generic-fysetc-spider.cfg would be a useful addition though.  The non-stock configs can be shared on the discourse forum - there's some info on this at: https://www.klipper3d.org/Example_Configs.html

-Kevin

GerogeFu:
Ok , i see , what do i need to do ? Make another PR or rebase with a new PR and force push ?

GerogeFu:
I made another PR : https://github.com/KevinOConnor/klipper/pull/4192

KevinOConnor:
Closing this as the spider config was merged as part of #4192 .

-Kevin

PR #4169, opened 2021-04-15
TMC2209: config internal_Rsense

revilo196:
added the field config for configurating  a TMC2208/2209 for internalRDS

Used to configure the driver for RDSon mode.
To use internal current mesurement instead of external resistors

Signed of by Oliver Walter <oli1111@web.de>

### DIFF
--- a/klippy/extras/tmc2208.py
+++ b/klippy/extras/tmc2208.py
@@ -214,6 +214,7 @@ class TMC2208:
         set_config_field(config, "pwm_autograd", True)
         set_config_field(config, "PWM_REG", 8)
         set_config_field(config, "PWM_LIM", 12)
+        set_config_field(config, "internal_Rsense", 0)
     def read_translate(self, reg_name, val):
         if reg_name == "IOIN":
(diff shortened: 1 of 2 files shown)


revilo196:
I developed my own tool board with tmc2209 in RDSon mode and klipper did not support this config. 

Also klipper did overwrite the otp default setting for this. 
May be worth investigaing how to mask bits that have otp defaults set inside the driver.
the otp_internalRsense value is loaded into GCONF:internalRsense register at startup by the driver

KevinOConnor:
Interesting, but this change doesn't look correct to me.  If internal_Rsense were to be set, the run_current, hold_current, and sense_resistor settings would no longer be accurate.

I'm also not sure this would be a good candidate for the master Klipper branch, as I don't know of any mainstream boards that use the Rdson method for current measurement.

-Kevin

revilo196:
Yes, your are probably right. This was just a solution for my specific case. 
But it could increase the flexablilty of klipper.

PR #4171, opened 2021-04-15
Palette2, Fix get_status

cmroche:
Quick fix for a missing null check in `get_status` should address the issue reported here: https://github.com/KevinOConnor/klipper/pull/4057#issuecomment-820093695

### DIFF
--- a/klippy/extras/palette2.py
+++ b/klippy/extras/palette2.py
@@ -638,10 +638,12 @@ class Palette2:
 
     def get_status(self, eventtime=None):
         status = {
-            "ping": self.omega_pings[-1],
+            "ping": None,
             "remaining_load_length": self.remaining_load_length,
             "is_splicing": self.is_splicing
         }
+        if self.omega_pings:
+            status["ping"] = self.omega_pings[-1]
         return status
 
 def load_config(config):


cmroche:
@cfelicio  FYI

klippy/extras/palette2.py
```diff
@@ -638,10 +638,12 @@ def _run_Smart_Load(self, eventtime):
 
     def get_status(self, eventtime=None):
         status = {
-            "ping": self.omega_pings[-1],
+            "ping": None,
             "remaining_load_length": self.remaining_load_length,
             "is_splicing": self.is_splicing
         }
+        if self.omega_pings:
+            status.update({ "ping": self.omega_pings[-1] })
```
KevinOConnor, line 646: Why use that form as opposed to the more common `status["ping"] = self.omega_pings[-1]`?

-Kevin

cmroche: Yeah that would make a lot of sense. I blame a lack of coffee when I wrote this.

cmroche: FYI, trivial change so I rebased.

KevinOConnor:
Thanks.

-Kevin

PR #4174, opened 2021-04-16
resonance_tester: Set max_accel ignoring the limit from the config

dmbutyugin:

### DIFF
--- a/klippy/extras/resonance_tester.py
+++ b/klippy/extras/resonance_tester.py
@@ -32,25 +32,30 @@ class VibrationPulseTest:
         return ['x', 'y']
     def get_start_test_points(self):
         return self.probe_points
-    def prepare_test(self, toolhead, gcmd):
+    def prepare_test(self, gcmd):
         self.freq_start = gcmd.get_float("FREQ_START", self.min_freq, minval=1.)
         self.freq_end = gcmd.get_float("FREQ_END", self.max_freq,
                                        minval=self.freq_start, maxval=200.)
         self.hz_per_sec = gcmd.get_float("HZ_PER_SEC", self.hz_per_sec,
                                          above=0., maxval=2.)
-        # Attempt to adjust maximum acceleration and acceleration to
-        # deceleration based on the maximum test frequency.
-        max_accel = self.freq_end * self.accel_per_hz
-        toolhead.cmd_SET_VELOCITY_LIMIT(self.gcode.create_gcode_command(
-            "SET_VELOCITY_LIMIT", "SET_VELOCITY_LIMIT",
-            {"ACCEL": max_accel, "ACCEL_TO_DECEL": max_accel}))
-    def run_test(self, toolhead, axis, gcmd):
+    def run_test(self, axis, gcmd):
+        toolhead = self.printer.lookup_object('toolhead')
         X, Y, Z, E = toolhead.get_position()
         if axis not in self.get_supported_axes():
             raise gcmd.error("Test axis '%s' is not supported", axis)
         vib_dir = (1, 0) if axis == 'x' else (0., 1.)
         sign = 1.
         freq = self.freq_start
+        # Override maximum acceleration and acceleration to
+        # deceleration based on the maximum test frequency
+        systime = self.printer.get_reactor().monotonic()
+        toolhead_info = toolhead.get_status(systime)
+        old_max_accel = toolhead_info['max_accel']
+        old_max_accel_to_decel = toolhead_info['max_accel_to_decel']
+        max_accel = self.freq_end * self.accel_per_hz
+        self.gcode.run_script_from_command(
+                "SET_VELOCITY_LIMIT ACCEL=%.3f ACCEL_TO_DECEL=%.3f" % (
+                    max_accel, max_accel))
         input_shaper = self.printer.lookup_object('input_shaper', None)
         if input_shaper is not None and not gcmd.get_int('INPUT_SHAPING', 0):
             input_shaper.disable_shaping()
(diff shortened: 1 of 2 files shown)


dmbutyugin:
@Sineos FYI, this is following the discussion in #3027.

@KevinOConnor I think I was able to overcome the problem with potential MCU 'No next step' errors. I have added the additional oscillation that is performed using the configured max_accel. I have tested this as follows:

- set `max_accel = 100` and `square_corner_velocity = 1` in the `printer.cfg`;
- checked that without that extra oscillation I get MCU 'No next step' at the end of the normal resonances test;
- with that added oscillation the problem disappears and the test finishes successfully.

Why it should work: MCU No next step error is generated to prevent cases of sudden loss of communication between the host and MCU, so it tests that the last 2 step commands for the stepper are separated by not less than a certain amount of time. So this error cannot be generated during the test on any kinematics that I am aware of, because if the stepper is moving for a certain axis, it will keep moving for the whole duration of the test. However, the default speed of the toolhead during the test is 75 / 4 = 18.75 mm/sec which is more than max_axis_halt returned by toolhead.get_max_axis_halt ~= 9 mm/sec. So the error is possible a the end of the test only when the speed of the toolhead can drop from 18.75 mm/sec to 0 too quickly. The additional moves take care of a proper deceleration of the toolhead.

KevinOConnor:
Interesting, thanks.  I'm a little leery about this approach though.  I agree that it should work in practice.  However, I'm concerned that the approach may be "fragile" in the sense that minor changes to other parts of the code may invalidate the assumptions here.  Also, I'm not sure it's a good idea to have a max_accel config parameter that mostly (but not always) limits the acceleration.

Perhaps an alternative solution would be to add a new `limit_max_accel` parameter to the toolhead that sets the upper limit that one may set the max_accel to at runtime.  (Thus making the current max_accel parameter a startup value instead of a runtime maximum.)  There are probably other alternatives as well.

-Kevin

dmbutyugin:
@KevinOConnor OK, thanks for the feedback. A few thoughts about this from my side:

I see the `max_accel` configuration for the resonance_tester module as an implementation detail. Most of the users should not care about it, and I'm afraid that currently most of them do not realize what that does exactly. At least, I saw messages and reports that actually indicate user confusion. Note that `max_accel` does not work as a safety or preventive measure in this case. If the printer will skip steps, or the resonances become too strong, that happens not at the maximum frequency, when the acceleration is really `max_accel`, but rather at the resonance frequency, which is typically smaller than `max_freq` parameter of the test. So the right course of action for the user is to reduce `accel_per_hz` parameter in the config, as the docs suggest, but not `max_accel`. What also worries me right now is that increased `max_accel` applies to every move, including homing moves and moves towards the testing point, and *those* moves may exceed the capabilities of the printer at `max_accel = 10000`. So it would indeed be better to just leave `max_accel` to its supposed value *for printing*, and let the test to adjust that constraint if necessary.

That being said, in light of your new developments in #4214, I'll just wait for the completion of that work, and only then see how the test should be updated to take advantage of that. In the meanwhile, I'll separate this PR into 2 (refactoring and the actual change to how max_accel is handled) and send the refactoring one out now. I think this removes some code duplication and makes it more straightforward.

dmbutyugin:
@KevinOConnor Now that #4214 is merged, I have updated this PR to only do the required stuff and removed the unnecessary hacks.

klippy/extras/resonance_tester.py
```diff
@@ -32,49 +32,54 @@ def get_supported_axes(self):
...
+        toolhead = self.printer.lookup_object('toolhead')
         X, Y, Z, E = toolhead.get_position()
         if axis not in self.get_supported_axes():
             raise gcmd.error("Test axis '%s' is not supported", axis)
         vib_dir = (1, 0) if axis == 'x' else (0., 1.)
         sign = 1.
         freq = self.freq_start
+        # Override maximum acceleration and acceleration to
+        # deceleration based on the maximum test frequency
+        _, old_max_accel = toolhead.get_max_velocity()
```
KevinOConnor, line 51: I think it would be worthwhile to save/restore max_accel_to_decel as well.  So, I think this could be done with something like:
```
systime = self.printer.get_reactor().monotonic()
toolhead_info = toolhead.get_status(systime)
...
old_max_accel = toolhead_info['max_accel']
old_max_accel_to_decel = toolhead_info['max_accel_to_decel']
```

klippy/extras/resonance_tester.py
```diff
@@ -32,49 +32,54 @@ def get_supported_axes(self):
...
+        max_accel = self.freq_end * self.accel_per_hz
+        toolhead.cmd_SET_VELOCITY_LIMIT(self.gcode.create_gcode_command(
+            "SET_VELOCITY_LIMIT", "SET_VELOCITY_LIMIT",
+            {"ACCEL": max_accel, "ACCEL_TO_DECEL": max_accel}))
```
KevinOConnor, line 55: FWIW, I think it might be a little simpler to call `self.gcode.run_script_from_command('SET_VELOCITY_LIMIT ACCEL=%.3f ACCEL_TO_DECEL=%.3f' % (max_accel, max_accel))`

klippy/extras/resonance_tester.py
```diff
@@ -32,49 +32,54 @@ def get_supported_axes(self):
...
+            toolhead.move([X, Y, Z, E], max_v)
             sign = -sign
             old_freq = freq
             freq += 2. * t_seg * self.hz_per_sec
             if math.floor(freq) > math.floor(old_freq):
                 gcmd.respond_info("Testing frequency %.0f Hz" % (freq,))
+        # Restore the original max acceleration value
+        toolhead.cmd_SET_VELOCITY_LIMIT(self.gcode.create_gcode_command(
+            "SET_VELOCITY_LIMIT", "SET_VELOCITY_LIMIT",
+            {"ACCEL": old_max_accel, "ACCEL_TO_DECEL": old_max_accel * .5}))
```
KevinOConnor, line 82: As mentioned above, I think it would be preferable to restore max_accel_to_decel.

docs/Config_Changes.md
```diff
@@ -6,6 +6,9 @@ All dates in this document are approximate.
 
 # Changes
 
+20210501: The `resonance_tester` module no longer requires modifying
+`max_accel` parameter for resonance testing.
+
```
KevinOConnor, line 11: I'm not sure we need to document this change, because the change is backwards compatible.  (That is, users could still use the old method.)  I guess if you wanted to we could document that the TEST_RESONANCES command will now increase max_accel beyond the configured limit.  I don't think it's needed though.

KevinOConnor (commented):
Thanks.  I have a few minor comments below.

-Kevin

klippy/extras/resonance_tester.py
dmbutyugin: OK, I've made it this way.

klippy/extras/resonance_tester.py
dmbutyugin: Indeed, thanks!

klippy/extras/resonance_tester.py
dmbutyugin: Done.

docs/Config_Changes.md
dmbutyugin: OK, I've removed this notice.

dmbutyugin:
Thanks Kevin, I've addressed your comments.

KevinOConnor:
Thanks!

-Kevin

Lingomat:
Agree, there is nothing in the description of max_accel that suggests it's really just a default acceleration now and M204 can do whatever it wants.

Sineos:
I guess there is a misunderstanding on your side:
- In previous versions you had to tune `max_accel` manually **for** resonance testing **only**
- After resonance testing you had to tune  `max_accel` back to its nominal value
- With this change the resonance test automatically tunes  `max_accel` to an appropriate value and you do no longer need to do this manually

PR #4176, opened 2021-04-16
config: Add Creality CR-30 PrintMill support

ezrec:
Add support for CoreXY belt printers, specifically the
Creality CR-30. The corexy_belt kinematic simply overrides
the corexy's home() function such that Z axis homing resets
the toolhead Z position, instead of performing a home on the
Z rail.

Additionally, an example configuration for the Creality CR-30
is added as a working demonstration of the feature.

Signed-off-by: Jason S. McMullan <jason.mcmullan@gmail.com>

### DIFF
--- a/config/generic-creality-v4.2.10.cfg
+++ b/config/generic-creality-v4.2.10.cfg
@@ -1,0 +1,144 @@
+# This file contains pin mappings for the Creality "v4.2.10" board. To
+# use this config, during "make menuconfig" select the STM32F103 with
+# a "28KiB bootloader" and serial (on USART1 PA10/PA9) communication.
+
+# Flash this firmware by copying "out/klipper.bin" to a SD card and
+# If you prefer a direct serial connection, in "make menuconfig"
+# select "Enable extra low-level configuration options" and select
+# serial (on USART3 PB11/PB10), which is broken out on the 10 pin IDC
+# cable used for the LCD module as follows:
+# 3: Tx, 4: Rx, 9: GND, 10: VCC
+
+# turning on the printer with the card inserted. The firmware
+# filename must end in ".bin" and must not match the last filename
+# that was flashed.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[stepper_x]
+step_pin: PC2
+dir_pin: !PB9
+enable_pin: !PC3
+microsteps: 16
+rotation_distance: 40
+endstop_pin: ^!PA3
+position_endstop: 0
+position_max: 220
+homing_speed: 50
+
+[stepper_y]
+step_pin: PB8
+dir_pin: !PB7
+enable_pin: !PC3
+microsteps: 16
+rotation_distance: 40
+endstop_pin: ^!PA7
+position_endstop: 0
+position_max: 240
+
+[stepper_z]
+step_pin: PB6
+dir_pin: !PB5
+enable_pin: !PC3
+microsteps: 16
+rotation_distance: 2.75
+endstop_pin: ^!PA5
+position_endstop: 0.0
+position_max: 200
+
+[extruder]
+step_pin: PB4
+dir_pin: !PB3
+enable_pin: !PC3
+microsteps: 16
+rotation_distance: 23.0
+nozzle_diameter: 0.400
+filament_diameter: 1.750
+heater_pin: PA0
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PC5
+control: pid
+pid_Kp: 21.527
+pid_Ki: 1.063
+pid_Kd: 108.982
+min_temp: 0
+max_temp: 250
+
+# K-FAN1
+[fan]
+pin: PA2
+
+[heater_bed]
+heater_pin: PA1
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PC4
+control: pid
+pid_Kp: 54.027
+pid_Ki: 0.770
+pid_Kd: 948.182
+min_temp: 0
+max_temp: 130
+
+[mcu]
+serial: /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
+baud: 250000
+restart_method: command
+
+[printer]
+kinematics: cartesian
+max_velocity: 200
+max_accel: 1500
+max_z_velocity: 20
+max_z_accel: 500
+
+# [filament_switch_sensor spool]
+# pause_on_runout: True
+# switch_pin: ^!PA6
+
+# K-FAN2
+# [fan_generic k_fan2]
+# pin: PC0
+
+# K-FAN3
+# [fan_generic k_fan3]
+# pin: PC1
+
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  I have some comments/questions on this change.

1. I thought the "belt" printers had a tilted xy gantry?  How does this work with just the Z homing disabled?
2. It looks like this change includes two sample config files.  It would be preferable for those configs to be in a separate PR.  They should also follow the guidelines described at: https://www.klipper3d.org/Example_Configs.html
3. The convention in Klipper is to not modify any member variables in classes declared in other files (as mentioned briefly in docs/Code_Overview.md).  The new corexy_belt.py code does not follow this convention (eg, it assumes there is a `self.limits` member variable created in corexy.py).  FWIW, this can be easily fixed by declaring `class CoreXYBeltKinematics` in corexy.py and just have a simple `load_kinematics()` method defined in corexy_belt.py .
4. I think if we're going to add a new kinematics that doesn't have a Z endstop, that we should change the code so that declaring a Z endstop is not possible.  (That is, use a `PrinterStepper()` instead of `LookupMultiRail()` for the axis.)

Cheers,
-Kevin

ezrec:
1. The 45 degree tilt has (it seems by some unknown consensus) been handled in the slicing layer, not in the printer kinematics itself. Therefore, the G-Code from N-degree-belt printer supporting slicers assumes that X-Y is the coordinate system of the gantry, and Z is strictly belt movement; **not** the X-Y-Z physical coordinate space.
2. I'd be more than happy to split that off into a separate PR. I'll amend this PR with those removed on my next update.
3. I like that, I wasn't sure if kinematics filename.py names needed to match `kinematics: ` names or not. Thanks for the clarification.
4. I started to try going down that rabbit hole, but ended up confusing myself. I'll need to study the classes better.

Is there a per-kinematics unit test I can run/add to? Just to make sure I've got all my edge cases covered.

KevinOConnor:
> The 45 degree tilt has (it seems by some unknown consensus) been handled in the slicing layer, not in the printer kinematics itself. 

Thanks.  I think I understand.

However, I'm not sure it's a good idea to add this as a new kinematics.  It seems we aren't really describing the "belt kinematics", but instead just disabling the z stepper of the existing corexy kinematics.

I think that should already be possible without a code change.  For example, something like:
```
[homing_override]
set_position_z: 0
gcode:
    G28 X0 Y0
```

-Kevin

ezrec:
> I think that should already be possible without a code change. For example, something like:
> 
> ```
> [homing_override]
> set_position_z: 0
> gcode:
>     G28 X0 Y0
> ```

Ah! Yes! After a bit of iteration, the best functionality is:

```
[homing_override]
axes: z
set_position_z: 0
gcode:
  G92 Z0
```
 
 I'll convert this PR to a pure configuration change!

ezrec:
Ready for review

KevinOConnor:
Thanks, at a high-level, it looks fine to me.  Some notes though:

1. If you want to add a "generic" config for just the board then it should follow the guidelines described at https://www.klipper3d.org/Example_Configs.html - that is, that config should use a more common kinematics and just describe the devices on the board (eg, no display).
2. Please don't specify a custom baud parameter in the printer config.  Please use the default 250000 setting.

Cheers,
-Kevin

ezrec:
> 1. If you want to add a "generic" config for just the board then it should follow the guidelines described at https://www.klipper3d.org/Example_Configs.html - that is, that config should use a more common kinematics and just describe the devices on the board (eg, no display).

Done, although since the Creality v4.2.10 doesn't have a pinout available online, I'll keep the `[display]` and other
sections describing the board connectors present, as comments.

> 2. Please don't specify a custom baud parameter in the printer config.  Please use the default 250000 setting.

The USB-to-Serial chip on this board (WHY?! It's a STM32F103 _and the USB pins aren't even used_! )  doesn't reliably support communication at 250000 baud - 230400 is the highest reliable baud rate it supports. (seems to only support 'natural' baud rates that are multiples of 9600 baud)

KevinOConnor:
Thanks.

The baud setting is a concern of mine - there's a lot of bad information on baud floating around the internet.  On all current Klipper boards, a baud of 250000 should be used.  It is possible, of course, that this is the first board where that is not true.  However, I'd like to see the debugging output confirming that.

It would help if you could recompile and flash the board with 250000 and then attach the resulting full klipper log file here.

-Kevin

ezrec:
Attaching both a log when firmware was flashed from a menuconfig where the baud was 230400, and a second log where the board was flashed to a baud of 250000.

In the first instance, the klipper comes up immediately at 230400.

In the second instance (at 250000) the klipper seems to hang waiting for some communication check after the initial version handshake.

After sending a 'G28' (which says 'you should probably FIRMWARE_RESTART') then sending FIRMWARE_RESTART, that seems to get klipper into a working state.

Subsequent power cycles at 250000 baud seems to work fine. I almost wonder if the USB to serial chip isn't fully powering down, and saving the prior baud rate for some time.

Regardless, I retract my objection to 250000 baud, and will update my PR.

[klipper-cr30-230400.log](https://github.com/KevinOConnor/klipper/files/6418845/klipper-cr30-230400.log)
[klipper-cr30-250000.log](https://github.com/KevinOConnor/klipper/files/6418846/klipper-cr30-250000.log)

ezrec:
Ah. Spoke too soon. Those logs are from my development machine (Ubuntu 21.04). 

On the MediaTek device I'm using to run Klipper on (OpenWRT Linux 5.4.104) I get the following for 250000 baud:

```
Starting serial connect
  File "/root/klipper/klippy/klippy.py", line 147, in _connect
    self.send_event("klippy:mcu_identify")
  File "/root/klipper/klippy/klippy.py", line 235, in send_event
    return [cb(*params) for cb in self.event_handlers.get(event, [])]
  File "/root/klipper/klippy/mcu.py", line 637, in _mcu_identify
    self._serial.connect_uart(self._serialport, self._baud, rts)
  File "/root/klipper/klippy/serialhdl.py", line 179, in connect_uart
    serial_dev.open()
  File "/usr/lib/python2.7/site-packages/serial/serialposix.py", line 272, in open
  File "/usr/lib/python2.7/site-packages/serial/serialposix.py", line 442, in _reconfigure_port
  File "/usr/lib/python2.7/site-packages/serial/serialposix.py", line 130, in _set_special_baudrate
ValueError: Failed to set custom baud rate (250000): [Errno 25] Not a tty
```

So it's a 'host side' issue - not something that should be in the .cfg - but this explains the prior failures I had.

ezrec:
- Update configs to specify 250000 baud.

KevinOConnor:
Thanks.  I committed this and made a couple of minor changes on top (commit dfcdd51e).

-Kevin

drphil3d:
I can't figure out a way to do live offset adjustment on the Y axis instead of the Z

PR #4178, opened 2021-04-17
virtual_sdcard: Add M808 repeat section code

ezrec:
To support continuous belt printing, add M808 repeat block support.
At this time, only the 'L' parameter is supported.

See [https://reprap.org/wiki/G-code#M808:_Set_or_Goto_Repeat_Marker](https://reprap.org/wiki/G-code#M808:_Set_or_Goto_Repeat_Marker)

Signed-off-by: Jason S. McMullan <jason.mcmullan@gmail.com>


kageurufu:
Marlin additionally supports `M808 K` from the host to cancel all current loops, essential for terminating infinite loops.

I updated the reprap.org wiki with some documentation for that

ezrec:
> Marlin additionally supports `M808 K` from the host to cancel all current loops, essential for terminating infinite loops.
> 
> I updated the reprap.org wiki with some documentation for that

Thanks! `M808 K` support added.

KevinOConnor:
Thanks.  My high-level feedback, however, is that I'm not really interested in adding any more "w31rd g-c0DE" commands to Klipper.

It's possible we could introduce new high-level commands for similar functionality (eg, SET_FILE_MARKER, REPEAT_FROM_FILE_MARKER), however I think we would want to discuss what format and syntax makes sense for that.  I think we may also want to discuss where that code should live.  (For example, maybe this could be done from the api-server?)

Also, I'm aware some people are working on adding more "CNC type" commands for Klipper.  I'm not sure if the M808 is commonly used by CNC machines (or if it's just another weird reprap thing).  I do know there are similar "goto" functionality needed by some common "CNC slicers".  So, we'll need to keep that in mind when looking at similar functionality for belt printers.

Cheers,
-Kevin

ezrec:
M808 is currently in use by Marlin, and is used by belt printers (specifically, the CR-30) to generate things like chains with distinct preamble-(repeat link x N)-postamble patterns, or repeat a model for some number of iterations (or, like I like to do, print models until I send `M808 K` to stop after the last one).

I have no objection for high level name instead of weird G-code, I was simply modelling from an existing implementation (especially as the Kiri::moto slicer already has UI and slicing support to generate M808 commands).

Suggestions:

- SECTION_START COUNT=n  ; Begin a new section (aka `M808 Ln`)
- SECTION_END ; End the current section (aka `M808`)
- SECTION_CANCEL ; Cancel current sections - any succeeding SECTION_END will be assumed to be the last iteration. (aka `M808 K`)

This is implemented at the `virtual_sdcard` layer, as only that layer knows how to 'go back in history' from a SECTION_END to the matching SECTION_START.

KevinOConnor:
Okay, thanks.  I must admit, I don't know enough about this to provide much advice.  I'll have to do some research on the common CNC "goto" commands to see if there is overlap.

> This is implemented at the virtual_sdcard layer, as only that layer knows how to 'go back in history' from a SECTION_END to the matching SECTION_START.

FWIW, we can add new methods to the virtual_sdcard code so that it can export current position and be commanded to change its current file position.

Cheers,
-Kevin

ezrec:
I did look into some '[classic CNC](https://www.cnccookbook.com/m98-m99-g-code-cnc-subprograms/)' options, but they don't handle the simple loop case elegantly in a way the Marlin or other firmware families also be able to handle well.

```
M98 Pnnnn Lcc ; Execute Subprogram nnnn  cc times
M30 ; Stop program
...
Onnnn ; Declare Subprogram section
M99 ; Return from Subprogram
```
This is interesting, but requires scanning the whole GCode in first to find the subprograms.

(Of note, `G65` is an alternative to `M98`, which makes parameter passing easier. The cool things you learn!)

Another variation is the IF/WHILE/DO/END supported by some CNC systems, ie:

```
#100 = 1  
WHILE [#100 LE 5] DO1
    ; G-Code goes here to be executred
    #100 = #100 + 1 ; Increment loop variable #100
END1
```
Again, interesting, and may be the direction you want Klipper to go in, but other 3D printer firmware families may be hard to extend to conditionals like this.

ezrec:
I added long GCode commands, and now both the long and M808 commands call proper `virtual_sdcard` subfunctions, and a unit test of the looping features.

Ready for review.

New long GCode commands:
  - SDCARD_LOOP_BEGIN COUNT=n  ; Loop for N times, or infinitely if N is 0
  - SDCARD_LOOP_END            ; End of loop
  - SDCARD_LOOP_DESIST         ; Complete all loops without iterating

ezrec:
Feedback gratefully appreciated!

Changes integrated into this diff:
- Refactored G-Code support for the new feature into `[loop_sdcard]`
- Added _only_ needed subfunctions into `[virtual_sdcard]`
- Updated documentation changes
- Updated 'unit tests' for the `[loop_sdcard]` feature
- Added `M808` as an example `[gcode_macro]` section in `config/sample-macros.cfg`

KevinOConnor:
Thanks.  At a high-level, it looks good to me.  I do have a few minor comments (should the commands be called LOOP_SDCARD_XXX if the module is named loop_sdcard?, next_file_position should be declared in virtual_sdcard's constructor, external code shouldn't peak at virtual_sdcard's cmd_from_sd member variable - instead a new method reporting that should be added and used, the top comment in loop_sdcard.py is a copy-and-paste error, the G-Codes.md document still lists M808, is it worthwhile to add a COUNT=0 hack as opposed to using COUNT=99999999?).

I'll look to merge this next week.

-Kevin

ezrec:
> (should the commands be called LOOP_SDCARD_XXX if the module is named loop_sdcard?

- Actually, I'll rename the module to be 'sdcard_loop', as I prefer the SDCARD_* specific gcodes to be in the same prefix.

> next_file_position should be declared in virtual_sdcard's constructor

Good catch! Fixing.

> external code shouldn't peak at virtual_sdcard's cmd_from_sd member variable - instead a new method reporting that should be added and used

- :facepalm: - fixing.

> the top comment in loop_sdcard.py is a copy-and-paste error

- Fixing

- the G-Codes.md document still lists M808

- Fixed

> is it worthwhile to add a COUNT=0 hack as opposed to using COUNT=99999999?).

I'd prefer to keep the COUNT=0 semantics, so that it's easier to craft macros for Marlin M808 gcode compatability

> I'll look to merge this next week.

Much appreciated, and than you for your patience.

Issue #3387, opened 2020-10-01
RPi 3 multiple mcu randomly throws "MCU 'host' shutdown: Stepper too far in past"


tntclaus:
Hi, I have multiple MCU corexy assembly where RAMPS is controlling z-steppers and extruder (mcu `mcu`) and RPi3 controlling x/y steppers (mcu 'host') and x/y steppers use x16 micro stepping. 

That is RPi3 mcu that fails, not RAMPS/Arduino. 

When I slice gcode in e.g. Cura (or Slic3r) using speeds above 60 mm/s I start getting random "Stepper in past" errors. It may happen in the middle of the print, in the beginning, or in the end. 

Attached is the `klippy.log` of such a print (and `printer.cfg), anywhere in gcode the speed does not exceed `F12000` (200 mm/s). This print fails in the first 1/3 of the model somewhere after the speed was set to `F9000`. 

I wonder why it was able to go `F12000` in this gcode many times before it finally failed? And why that happens in random places of the same gcode?

My setup of RPi3 can definitely keep speeds up to `F36000` (tested with custom gcode that moves toolhead along perimeters and diagonals, btw this gcode is also sometimes randomly fails, sometimes it doesn't). 

Right now the problem seems for me like a software bug that occurs on RPi. If it is I can try to fix it myself, but I will need help bcz the codebase is somewhat big an unfamiliar to me. 

[printer.cfg.1.txt](https://github.com/KevinOConnor/klipper/files/5311232/printer.cfg.1.txt)
[klippy.log.zip](https://github.com/KevinOConnor/klipper/files/5311163/klippy.log.zip)

KevinOConnor:
The "linux mcu" code is not intended to be a high-performance system.  (It uses a slow linux kernel interface to toggle gpio pins.)  In addition to not being able to reach very high speeds, the individual step pulses are likely to not have precise timing either.

The "linux mcu" code is more intended for minor background type tasks (turning on leds, fans, power relays, etc.).  

-Kevin

tntclaus:
@KevinOConnor that's what I thought firs, but in my case it doesn't seem like `klipper_mcu` process has no spare cpu cycles. I already made an experimental patch for RPi3 which utilises direct memory access to handle gpio control (x70 times performance gain from somewhat 0.6 to 36 MHz). 

[...]

PR #4190, opened 2021-04-19
Convert linux mcu code to use signal based "irq" handling

KevinOConnor:
This converts the linux mcu code to use Unix signals to implement the software timer dispatch.  As discussed in #3387 this can reduce scheduling jitter.  It should also be a little more CPU efficient in general.

@tntclaus - FYI the latest implementation is slightly different than previous iterations of this code.  The latest version should be a bit more CPU efficient as the irq_disable()/irq_enable() code now just sets internal flags.

-Kevin

### DIFF
--- a/src/linux/Makefile
+++ b/src/linux/Makefile
@@ -7,7 +7,7 @@ src-y += linux/pca9685.c linux/spidev.c linux/analog.c linux/hard_pwm.c
 src-y += linux/i2c.c linux/gpio.c generic/crc16_ccitt.c generic/alloc.c
 src-y += linux/sensor_ds18b20.c
 
-CFLAGS_klipper.elf += -lutil -lpthread
+CFLAGS_klipper.elf += -lutil -lrt -lpthread
 
 flash: $(OUT)klipper.elf
 	@echo "  Flashing"
--- a/src/linux/console.c
+++ b/src/linux/console.c
@@ -1,28 +1,27 @@
 // TTY based IO
 //
-// Copyright (C) 2017  Kevin O'Connor <kevin@koconnor.net>
+// Copyright (C) 2017-2021  Kevin O'Connor <kevin@koconnor.net>
 //
 // This file may be distributed under the terms of the GNU GPLv3 license.
 
+#define _GNU_SOURCE
 #include <errno.h> // errno
 #include <fcntl.h> // fcntl
-#include <poll.h> // poll
+#include <poll.h> // ppoll
 #include <pty.h> // openpty
 #include <stdio.h> // fprintf
 #include <string.h> // memmove
 #include <sys/stat.h> // chmod
-#include <sys/timerfd.h> // timerfd_create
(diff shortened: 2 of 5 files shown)


KevinOConnor:
After thinking about this further, this branch has a significant flaw and it's not currently suitable for merging into the master branch.  

Specifically, a signal could occur while the code is in a library function (eg, fprintf) and the signal handler could call shutdown() which calls longjmp().  That would put the code in an invalid state and likely crash the host process.  (Also, the call to pthread_create() in sensor_ds18b20.c would allow the signal handler to run in a separate thread, but that's easily fixed.)

There are workarounds for the above, but I'm not sure it's a great approach.  One possibility that I may try is a hybrid signal/polling approach - have the signal handler just set a flag that irq_poll() can check for.

-Kevin

tntclaus:
Maybe we can stay with optimizations we already have in master? Tomorrow I am very busy but on Thursday I can do benchmarks for master branch + native gpio and also run real life gcode to know performance and be sure stepping tolerance remains in 1ms window.

KevinOConnor:
> Maybe we can stay with optimizations we already have in master? 

That's probably a good idea for the short term.  If polling doesn't work, I think we may want to look at running the timers in a separate thread and avoid using signals.

FYI, I made a few additional minor optimizations on the master branch.  If you run further tests it's probably a good idea to grab the latest code.

-Kevin

tntclaus:
>  If you run further tests it's probably a good idea to grab the latest code.

Got it. Right now working with testing.

tntclaus:
From current master + native gpio:

| Steppers  | Ticks | Rate |
| -----------: | ------------: | ------: |
| 2 stepper | 237 | 211K |
| 2 stepper | 247 | 405K |
| 3 stepper | 257 | 584K |
| 4 stepper | 257 | 778K |
| 5 stepper | 272 | 919K  |
| 6 stepper | 295 | 1017K |

A bit worse than same benchmark done with irq branch but the difference is unnoticeable.

KevinOConnor:
Those results certainly look good.  I guess the big question is - are you able to perform regular printing without getting any "stepper in the past" errors?

-Kevin

tntclaus:
Yes, I can. Just finished 30 minute print with current master + native gpio at 150 mm/s and 20k accelerations and the part quality is perfect. Right now going to do a 4:30 long print of wing section at 200 mm/s speed, if something fails I'll let you know.

tntclaus:
Here are two parts printed from same g-code (highest speed 150 mm/s). 

On the left — current master + native gpio.
On the right — my old native gpio + shutdown for stepper to far in past disabled.

IMG_20210423_202418

Turns out `stepper too far in past` shutdown makes great sense. You can show this photo to those who don't understand and complain about this error :)

KevinOConnor:
Thanks.  I'm going to close this PR as I'm no longer confident in the signal based approach, and it appears it is no longer needed.

-Kevin

PR #4192, opened 2021-04-20
Add example configs for fysetc spider board

GerogeFu:
Add example configs for fysetc spider board


KevinOConnor:
Thanks.  I fixed the whitespace errors and committed this config (commit fee386b8).

-Kevin

PR #4195, opened 2021-04-21
tronxy x3s example

sbechet:


KevinOConnor:
Thanks, but example configs should follow the guidelines at: https://www.klipper3d.org/Example_Configs.html

-Kevin

sbechet:
OK. Sorry to be so long to answer.

You right, *x3s printer* is not a good candidate for commit after reading `Example_Configs.html`.

PR #4196, opened 2021-04-21
adxl345: Verify the register writes, added debug adxl345 commands

dmbutyugin:

### DIFF
--- a/klippy/extras/adxl345.py
+++ b/klippy/extras/adxl345.py
@@ -20,7 +20,10 @@ QUERY_RATES = {
     800: 0xd, 1600: 0xe, 3200: 0xf,
 }
 
-SCALE = 0.004 * 9.80665 * 1000. # 4mg/LSB * Earth gravity in mm/s**2
+ADXL345_DEV_ID = 0xe5
+
+FREEFALL_ACCEL = 9.80665 * 1000.
+SCALE = 0.0039 * FREEFALL_ACCEL # 3.9mg/LSB * Earth gravity in mm/s**2
 
 Accel_Measurement = collections.namedtuple(
     'Accel_Measurement', ('time', 'accel_x', 'accel_y', 'accel_z'))
@@ -93,7 +96,7 @@ class ADXL345Results:
         write_proc.daemon = True
         write_proc.start()
 
-# Printer class that controls measurments
+# Printer class that controls ADXL345 chip
 class ADXL345:
     def __init__(self, config):
         self.printer = config.get_printer()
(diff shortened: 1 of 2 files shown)


dmbutyugin:
This is a part of my adxl345-probe branch. But I think these changes are useful by themselves, because they add validation of the ADXL345 operations and some ADXL345 debugging tools.

klippy/extras/adxl345.py
```diff
@@ -135,11 +136,33 @@ def __init__(self, config):
         gcode.register_mux_command("ACCELEROMETER_QUERY", "CHIP", self.name,
                                    self.cmd_ACCELEROMETER_QUERY,
                                    desc=self.cmd_ACCELEROMETER_QUERY_help)
+        gcode.register_mux_command("READ_ADXL345", "CHIP", self.name,
+                                   self.cmd_ADXL345_DEBUG_READ,
+                                   desc=self.cmd_ADXL345_DEBUG_READ_help)
+        gcode.register_mux_command("SET_ADXL345", "CHIP", self.name,
+                                   self.cmd_ADXL345_DEBUG_WRITE,
+                                   desc=self.cmd_ADXL345_DEBUG_WRITE_help)
```
KevinOConnor, line 144: Typo?  Registers `READ_ADXL345` when documentation uses `ADXL345_DEBUG_READ`?

klippy/extras/adxl345.py
```diff
@@ -170,23 +193,35 @@ def _convert_sequence(self, sequence):
...
+            return
         rate = rate or self.data_rate
-        # Verify chip connectivity
-        params = self.spi.spi_transfer([REG_DEVID | REG_MOD_READ, 0x00])
-        response = bytearray(params['response'])
-        if response[1] != 0xe5:
-            raise self.printer.command_error("Invalid adxl345 id (got %x vs %x)"
-                                             % (response[1], 0xe5))
+        if not self.is_initialized():
+            self.initialize()
```
KevinOConnor, line 216: If the spi connection isn't wired correctly and all results are returning 0xff, would this pass without error?  That is, would `is_initialized()` return True because the read to REG_DATA_FORMAT returns non-zero, and thus skip the 0xe5 check in `initalize()`.

KevinOConnor (commented):
Thanks.  In general it looks good to me.  I have a couple of minor comments below.

-Kevin

klippy/extras/adxl345.py
dmbutyugin: Ah, that's a good catch, I was renaming the command to a more appropriate one, and I guess I forgot to rename these instances. Renamed.

klippy/extras/adxl345.py
dmbutyugin: That's fair. I added the check for 0xe5 device id to is_initialized too. I kept raising the exception in a single place though.

dmbutyugin:
Thanks for review, Kevin!

KevinOConnor:
Thanks.

-Kevin

PR #4200, opened 2021-04-22
neopixel: Add periodic refresh timer

mental405:
module: Neopixel.py

Sometimes signal noise can affect transmissions to neopixel strips. This
can cause them to blank out or display color erratically. I have added
an additional configuration option for neopixels that schedules a
periodic timer to re-send the last colors set to the LED strips.

If a SET_LED command is sent with TRANSMIT=0, the refresh timer
is paused until a SET_LED command with TRANSMIT=1 is sent.


mental405:
This started out by adding one line to `neopixel.cmd_SET_LED()` but that ended up opening up a can of worms where if someone put it in a delayed gcode timer on repeat it would reset the idle timeout.

KevinOConnor:
Interesting, thanks.  I guess my main high-level feedback is - how can we be sure that the refresh itself wont have signal noise?  That is, the lights may actually be correct looking, but we later corrupt it during a refresh..

Also, not super important, but a refresh may come at a time when the host and mcu are busy performing movement, which may be unexpected for a user.  (That is, the user may only have programmed the LED prior to starting a print, but the refreshes would occur during a print.)

-Kevin

mental405:
> how can we be sure that the refresh itself wont have signal noise?

You can't in practice.  It has been my experience with various projects using addressable LEDs... mostly in environments that are less than ideal.... it is better to just bombard them with frequent updates. The updates themselves may end up having a noise affect them. If there is a _lot_ of noise on the data line, this really won't help things. If there is a small amount of noise or occasional burst of noise this is intended to _course correct_ the LEDs.

> Also, not super important, but a refresh may come at a time when the host and mcu are busy performing movement, which may be unexpected for a user. (That is, the user may only have programmed the LED prior to starting a print, but the refreshes would occur during a print.)

The intent here is to maintain permanence in the colors that have been set. The ultimate goal is to keep the colors on the strip the same as they were without resetting the idle timeout. If someone were to program the colors prior to the start of a print, they would be refreshed with the same color values until the user changed them or the firmware was reset.

I personally have had issues with the 6 LEDs I have mounted underneath my toolhead that I use to illuminate the nozzle. The wire isn't particularly shielded and it runs through the cable chains next to all of the other toolhead wires. I can set the LED's to be a color during a "start print" macro but at some point they sometimes become corrupted and either go blank or change colors. I previously had a delayed gcode macro that would periodically reset them but that was less than ideal since it would also result in resetting the idle timeout after the print had finished.

A lot of people are using LEDs to represent print progress, temperature to target, etc... I have seen many macros that are wanting to sync print time to a "bar" of LEDs or to color them based on "temperature to target". These events would be occurring during a print which would be an expected change. If an update occurred for print time, for instance, and the color data did not correctly apply to the strip, this would correct that error as soon as the timer runs.


Additionally, I don't think this is really needed for the Dotstar/APA102s because they have a discrete clock signal

KevinOConnor:
Okay.  As a high-level comment, I'm okay with merging this.  However, given the nature of line noise, I wonder if it might be more flexible to add a new `SYNC=0` type of option to the SET_LED command so that it doesn't wake the toolhead time.  That might be simpler for some use cases.

-Kevin

mental405:
a `sync=0` in addition to the periodic refresh timer or instead of? 

Are you envisioning the functionality to be something along the lines of running the callback as just a regular reactor callback instead of the toolhead.register_lookahead_callback()?

mental405:
So I added an optional SYNC parameter.

Sending the command with `SYNC=1` or omitting the parameter will cause `SET_LED` to grab the print time and sync with the toolhead. The MCU command will still be sync'd with the print time.

Sending the command with `SYNC=0` will register  it as a regular reactor callback at the current reactor time.

The transmit parameter works the same and I have left the periodic_refresh timer and it still does the same thing.


so sending the command 

`SET_LED LED=display_led RED=1 GREEN=0 BLUE=0 INDEX=1 TRANSMIT=1 SYNC=0`

would result in the LED changing color, but not syncing with print time.

mental405:
I am really bad at branches and git.

KevinOConnor:
Thanks.  Looks like this got mangled in the git merge.

Could we just do something like the following to implement `SYNC=0`?
```diff
--- a/klippy/extras/neopixel.py
+++ b/klippy/extras/neopixel.py
@@ -119,6 +119,7 @@ class PrinterNeoPixel:
         white = gcmd.get_float('WHITE', 0., minval=0., maxval=1.)
         index = gcmd.get_int('INDEX', None, minval=1, maxval=self.chain_count)
         transmit = gcmd.get_int('TRANSMIT', 1)
+        sync = gcmd.get_int('SYNC', 1)
         # Update and transmit data
         def reactor_bgfunc(print_time):
             with self.mutex:
@@ -128,8 +129,13 @@ class PrinterNeoPixel:
         def lookahead_bgfunc(print_time):
             reactor = self.printer.get_reactor()
             reactor.register_callback(lambda et: reactor_bgfunc(print_time))
-        toolhead = self.printer.lookup_object('toolhead')
-        toolhead.register_lookahead_callback(lookahead_bgfunc)
+        if sync:
+            # LED update synced with timing of other g-code commands
+            toolhead = self.printer.lookup_object('toolhead')
+            toolhead.register_lookahead_callback(lookahead_bgfunc)
+        else:
+            # Send update immediately; don't reset idle_timeout
+            lookahead_bgfunc(None)
 
 def load_config_prefix(config):
     return PrinterNeoPixel(config)
```

-Kevin

mental405:
That is what I was thinking if the periodic refresh is not to be included.

PR #730, opened 2018-10-03
Feature: support half-stepping rotary encoders

nickbrennan01:
The existing implementation of the RotaryEncoder class only emits movement events after the encoder has cycled through 4 states. Some encoders only cycle through 2 states per physical detent, and so require two physical "clicks" to generate one menu movement.

This feature allows the user to select a step setting appropriate for their printer's encoder.

PR #4202, opened 2021-04-23
Fix rotary encoder handling - resurrection of #730

susisstrolch:
Original fix from @nickbrennan01

This feature supports half-stepping rotary encoders and encoders with inverted
inputs (like Velleman K8800).

Requires an additional config entry in [display] section:
  encoder_steps_per_detent: 2|4

encoder_steps_per_detent = 2: "regular" encoder
encoder_steps_per_detent = 4: half-stepping encoder

Signed-off-by: Juergen Orschiedt juorschiedt@gmail.com

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3183,6 +3183,12 @@ lcd_type:
 #encoder_pins:
 #   The pins connected to encoder. 2 pins must be provided when using
 #   encoder. This parameter must be provided when using menu.
+#steps_per_detent:
+#   The steps per detent defines if your are dealing with a half-stepping 
+#   encoder or a full stepping encoder. Allowed values are 4 (half-stepping) or
+#   2 (full stepping)  
+#   The default is 4
+
 #click_pin:
 #   The pin connected to 'enter' button or encoder 'click'. This
 #   parameter must be provided when using menu. The presence of an
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3187,8 +3187,7 @@ lcd_type:
 #   The steps per detent defines if your are dealing with a half-stepping 
 #   encoder or a full stepping encoder. Allowed values are 4 (half-stepping) or
 #   2 (full stepping)  
-#   The default is 4
-
+#   The default is 2.
 #click_pin:
 #   The pin connected to 'enter' button or encoder 'click'. This
 #   parameter must be provided when using menu. The presence of an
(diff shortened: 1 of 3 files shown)


jellevictoor:
updated with documentation

KevinOConnor:
Thanks - some high-level comments:
1. Please don't mix white space changes in with functional changes.  It makes it harder to review the code and makes it harder to find regressions (should a regression be introduced).  Also, FWIW, the white space changes wont be accepted.
2. FWIW, I find mixing the different detent config in the the state machine to be confusing.  Maybe using two different state machines would be simpler.

@mcmatrix - do you have any comments on this?

-Kevin

susisstrolch:
The white spacee changes were accidental - I've removed them.
So shall I push it again, after whitespace and line length are fixed?
Or shall I simply close the request, until anybody else comes up with a splitted state machine and simply let those with a half step encoder wait for the next volunteer who tries to solve the problem instead having them hanging around without working solution?

KevinOConnor:
I'm not sure what the state of this PR is.  Is there still interest in it?

-Kevin

girrrrrrr2:
> I'm not sure what the state of this PR is. Is there still interest in it?

I am interested in it, and I feel like it could be a great addition to klipper/3d printing as a whole, but I dont know if there are others that are also interested or if its just me, or if that comment was referenced to just the author...

KevinOConnor:
I suspect this feature will need a developer to lead it to completion.  Otherwise, I don't see it getting merged without a conclusion on the review comments and fixing the build failures.

-Kevin

jellevictoor:
Hi @KevinOConnor 
I took in account the fixes you suggested and fixed the build. is this ok to merge?

KevinOConnor:
Sorry, but when I look at the diff I still see the items I mentioned at https://github.com/KevinOConnor/klipper/pull/4202#issuecomment-828596744 .  There are still unrelated white space changes.  Also, I find the new state machine to be quite confusing (the new code seems to be mixing two different devices into one state machine).

-Kevin

jellevictoor:
Based on the comments of susisstrolch i thought the whitespaces were ok, I'll revisit
can you elaborate more on how you would split up the state machine? i have limited knowledge of python programming, i just applied the composition over inheritance, since this is not exposed to the outside, but i'm happy to take your suggestions so this could be merged

KevinOConnor:
> can you elaborate more on how you would split up the state machine?

I find it confusing to modify ENCODER_STATES; I would make an entirely new ENCODER_STATES_FOR_1_DETENT (or similar) and not modify the existing state machine.

Alas, due to time constraints on my time, this isn't something I can spend further time on.

-Kevin

PR #4212, opened 2021-04-25
Feature/add k8800 config

jellevictoor:
Added config for the velleman K8800 printer. validated and checked by myself en @susisstrolch
printing a while with Klipper and want to make others able to use it by sharing this


KevinOConnor:
Thanks.  In general it looks good to me.  Can you give some background on the source of rotation_distance for this printer though?  I don't know of any printer that has a 55mm rotation distance, nor an extruder with a 42mm circumference hobbed bolt (13.5mm diameter).  What type of belts are on the printer and what type of extruder does it use?

-Kevin

susisstrolch:
From Marlin Configuration.h:

  #define XYZ_FULL_STEPS_PER_ROTATION 200
   #define XYZ_MICROSTEPS 32
   #define XYZ_BELT_PITCH 2.5
   #define XYZ_PULLEY_TEETH 22

Extruder is a BMG...

Von meinem iPad gesendet

jellevictoor:
Hi Kevin
I kept only the config for the default Velleman setup, the extruder was really a weird number, but I followed the steps described in the documentation (https://github.com/KevinOConnor/klipper/blob/master/docs/Rotation_Distance.md#obtaining-rotation_distance-by-inspecting-the-hardware documentation.) to calculate it.
so the belt is this one https://www.velleman.eu/products/view/?id=441296, tooth profile: T2,5
combined with what @susisstrolch presented from the marlin firmware config for the k8800, you get 55

If you need more info, please let me know

KevinOConnor:
What's the status of this PR?  I think it used to pass the test cases but is no longer?

-Kevin

jellevictoor:
Apparently no newline EOF, added it

KevinOConnor:
Thanks.  I committed this change and made a couple of minor changes on top (commit c9e8298c).

BTW, is it possible the extruder is using 16 microsteps and actually has a bolt circumference of 21.2333mm?  The 42.4666mm seems very odd.

-Kevin

PR #4218, opened 2021-04-26
extruder: add E-axis velocity and acceleration limits

richfelker:
In theory these limits can be achieved by converting to spatial units and imposing the corresponding limit there; however, doing so depends on layer thickness and line width, and is not practical if those vary. Since this needs to be configurable at runtime to be useful (the limits may vary by material, etc.) I've added a new command `SET_EXTRUDER_VELOCITY_LIMIT`, for changing their values. It could also be extended to let the existing (extrude-only move) settings be changed at runtime.

The defaults are made effectively-infinite to avoid any change for users who do not set limits, and I would expect users aiming for maximal print speed to leave at least the acceleration unlimited. However, limiting E-axis acceleration is useful for optimizing extrusion accuracy with pressure advance.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -592,6 +592,11 @@ filament_diameter:
 #   specified then they are calculated to match the limit an XY
 #   printing move with a cross section of 4.0*nozzle_diameter^2 would
 #   have.
+#max_extrude_print_velocity:
+#max_extrude_print_accel:
+#   Maximum velocity (in mm/s) and acceleration (in mm/s^2) of the
+#   extruder motor for print moves. If not specified the limits are
+#   effectively infinite.
 #pressure_advance: 0.0
 #   The amount of raw filament to push into the extruder during
 #   extruder acceleration. An equal amount of filament is retracted
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -181,6 +181,9 @@ The following standard commands are supported:
   result in excessive pressure between extruder and hot end. Do proper
   calibration steps with filament before use. If 'DISTANCE' value is
   not included command will return current step distance.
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Thanks.  As high-level feedback, though, I'm not sure this is a good candidate for the Klipper master branch.  My main concern is, as you indicate, that one can not easily determine the impact this imposes on toolhead speeds.  I fear this type of option leads to "seemingly random" speed changes during prints.

I suspect this type of option is better implemented in the slicer - that has the advantage of being able to easily inspect what speed changes it imposes.  Indeed, as I understand it, several slicers already implement this flow-rate limiting capability.

Cheers,
-Kevin

richfelker:
Doing it in the slicer makes it impossible to test dynamically at print-time to tune extrusion, either with tuning tower or manual knobs, and it's not a feature existing slicers have. Also, if you do it in the slicer, you're stuck with the issue I described in my comment, where converting the E-axis acceleration limit to a spatial acceleration limit for the toolhead limits the centripetal velocity at junctions. This is not only unnecessary, but harms print speed and exrusion accuracy by producing unneeded decel and accel cycles for paths that should be able to maintain a constant extrusion speed.

KevinOConnor:
I understand.  However, I'm reluctant to add a config option which subtly alters the intent of other prominent config options.

FWIW, other firmware have these types of limits and in my experinece they are a nightmare to work with.  If I instruct my printer to move with an acceleration of 3000 then I want my printer to do just that - I do not want it to "out think me" and silently use a different acceleration for non-obvious reasons.

-Kevin

KevinOConnor:
Thanks, but I'm going to close this as I don't currently think it is a good match for the master Klipper branch.

-Kevin

PR #4220, opened 2021-04-26
Deprecate gcode_macro default_parameter_<name> config

KevinOConnor:
The `default_parameter_NAME` system has always been a bit goofy.  It's a holdover from gcode macros from before the introduction of Jinja2.  The Jinja2 "set" directive is more flexible and easier to understand.

This PR deprecates those config options and encourages using "set" as a replacement.  The change now is just to the documentation - actual support would be removed in around 6 months.

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -6,6 +6,12 @@ All dates in this document are approximate.
 
 # Changes
 
+20210503: The gcode_macro `default_parameter_<name>` config option is
+deprecated.  Use the `params` pseudo-variable to access macro
+parameters.  Other methods for accessing macro parameters will be
+removed in the near future.  See the [Command Templates
+document](Command_Templates.md#macro-parameters) for examples.
+
 20210430: The SET_VELOCITY_LIMIT (and M204) command may now set a
 velocity, acceleration, and square_corner_velocity larger than the
 specified values in the config file.
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1155,16 +1155,6 @@ G-Code macros (one may define any number of sections with a
 #   A list of G-Code commands to execute in place of "my_cmd". See
 #   docs/Command_Templates.md for G-Code format. This parameter must
 #   be provided.
-#default_parameter_<parameter>:
-#   One may define any number of options with a "default_parameter_"
-#   prefix. Use this to define default values for g-code parameters.
-#   For example, if one were to define the macro MY_DELAY with gcode
-#   "G4 P{DELAY}" along with "default_parameter_DELAY = 50" then the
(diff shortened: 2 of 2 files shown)


mental405:
I think this is probably an overall good thing as the `default_parameter_NAME` ends up creating confusion when you _also_ have parameters accessed via the `params` collection as well as being accessible directly by name. It seems like encourageing `set` would potentially condense defaulted values to a single concise `set` statement vs having to worry about if the specifying `default_parameter_x` is being used over the `params.x` defaulted value vs the `X|default()` .

zellneralex:
Yes on the one hand it makes sense to remove it but it will most likely break nearly any printer.cfg out there… That means it will be a supporting nightmare. We will see how it works when the support of step_distance ends.

KevinOConnor:
Looking closer at the code, it looks like it is currently also possible to directly access parameters via their name in a macro.  For example, with `MYMACRO SOMEPARAM=14` one could currently use `{SOMEPARAM}` in the gcode.  I'm going to deprecate that as well - so going forward, only `{params.SOMEPARAM}` will be valid.

-Kevin

KevinOConnor:
> but it will most likely break nearly any printer.cfg out there

I don't think custom macros that took parameters was that popular.  But, yes, it will break configs.  I think we can start with documentation and awareness.  Updating the syntax should not be a difficult change for anyone that is writing custom macros.

-Kevin

zellneralex:
Nearly any PRINT_START that I know use bed and extruder temp as parameter and most of them as default_parameter. The example macros for PAUSE/RESUME/CANCLE_PRINT in the mainsail docs also use it (ALREADY PLANNED TO CHANGE). We changed it for MainsailOS and also FLUIDDPI uses the changed version so we are at least save there. We will manage it but do not underestimate the impact of it.
Removing also the direct access via {Name} is a good decision as it forces a strict rule.

mental405:
It was brought up on discord yesterday the concern for front-ends and being able to adequately sniff parameters out of the gcode for UI buttons. I think making them accessible _only_ via the `params` collection is a good move as well.

zellneralex:
But {% if 'X' in params %} should be allowed also

mental405:
> But {% if 'X' in params %} should be allowed also

It should be. I am pretty sure that `params` is a dict so anything inside of it would be able to be checked with that syntax.

KevinOConnor:
Thanks for the feedback.  I went ahead and merged this documentation change.

I agree this may be harder to remove than I originally thought, however I still think it would be worthwhile to make it clear that the `params` method is the preferred method to use going forward.  We can evaluate in 6 months if more time is needed before removing the old support.

-Kevin

PR #4227, opened 2021-04-28
generic klipper configuration for BTT E3 RRF V1.1

Xploder:


KevinOConnor:
Thanks, but example configs should follow the guidelines described at: https://www.klipper3d.org/Example_Configs.html

-Kevin

KevinOConnor:
Closing as it looks like this has become stale.

-Kevin

PR #4234, opened 2021-04-29
config: Add motor power output pin to SKR2 config

FHeilmann:
The SKR2 shuts off power to the TMC drivers and most FETs unless the motor power pin is pulled high. This is supposed to act as a safety feature. In the context of klipper the absence of this output pin will cause motor movement to fail, since the MCU can't communicate with the drivers if motor power is not applied. This PR fixes that issue.

Signed-off-by: Florian Heilmann <Florian.Heilmann@gmx.net>

### DIFF
--- a/config/generic-bigtreetech-skr-2.cfg
+++ b/config/generic-bigtreetech-skr-2.cfg
@@ -84,6 +84,10 @@ pin: PB7
 #[heater_fan fan2]
 #pin: PB5
 
+[output_pin motor_power]
+pin: PC13
+value: 1
+
 [mcu]
 serial: /dev/serial/by-id/usb-Klipper_Klipper_firmware_12345-if00
 


Sineos:
https://www.youtube.com/watch?v=SjVkqXaQKtc

Apparently the board has a bug in conjunction with this stepper safety feature that can blow TMC2209 drivers.

Jasonpenman:
BTT have an announcement covering some physical changes that need to be made to use 2209's, it requires a new mosfet or some cables adding, 
https://docs.google.com/document/d/1IeKgfE2WIDjqH1fx5Yg7n1FOHVwhDFmDlZ-7QMlOEV0/edit?fbclid=IwAR08vmPPPkwAznbGQEEHlEAIRvemnTjF1-m8VDHkUo6HRyQPCZywdfP4vOs

KevinOConnor:
Thanks.  I think we have a few options here:
1. Merge this.
2. Remove the existing skr-2 config file.
3. Merge this and add a warning about the early skr-2 revs being broken.

Thoughts?
-Kevin

Sineos:
To quote from the doc linked by @Jasonpenman:

> Fortunately, we were monitoring the forums and groups and were able to catch this very fast after the board was released. This means that only a few hundred boards were ever sent out to customers or into distribution channels. We have called back all boards that have not been sold to customers and we are happy to say that from this point on you can buy an SKR 2 with the confidence of knowing that you will be getting a Rev B board.

and https://docs.google.com/document/d/1swmc4HvP9vxrxV2b9LVGa_I7GLTQGogQns7CfMYCckA/edit

> Note: The information in this guide only applies to the SKR 2 Rev A. The SKR 2 Rev B can be used with TMC drivers without these modifications. To identify whether you have a Rev A or a Rev B board look for a sticker on your board that shows "QC pass 02 " . If there is no sticker then you have a Rev A board.

This means:
- Just a few problematic boards out there
- Good boards are easy to identify by the second label "QC pass 02"

Maybe add a note to config like:

> If you bought the board from a retailer between April and May 2021 AND the board does not have a sticker "QC pass 02" on the PCB bottom then the board might have a bug leading to damaged TMC drivers. Revert back to Bigtreetech for further instructions.

Jasonpenman:
Does klipper have the driver protection code enabled? If not then one of the btt hardware fixes would be all it needed?

FHeilmann:
> Thanks. I think we have a few options here:
> 
> 1. Merge this.
> 2. Remove the existing skr-2 config file.
> 3. Merge this and add a warning about the early skr-2 revs being broken.
> 
> Thoughts?
> -Kevin

I'm torn between keeping the config and putting in the fix as well as the warning, and removing it altogether. On the one hand, all SKR2s in the wild should not be used with klipper (so therefore we shouldn't enable their use), but on the other hand, this would help to spread the warning to more users.

aabeuge:
I ran across this while trying to find answers to running TMC2209 on this board and need to chime in.  It is better to identify the rev A. boards based on the mosfet installed on them.  Rev A boards will have a G090N06 fet as explained here:
https://github.com/bigtreetech/SKR-2/issues/17

Wilk86:
Hi,

have the Rev A. 

The UART Mod is not working if the motor power pin is not pulled high.

I had this problem and have annoyed people all day :D thx

Funny is that Marlin makes the pin automatically or something pulled high. I'm not an expert. But with Marlin the TMCs connection works.  Witchcraft :)

```
########################################
# Comment out if UART not Working and your Board is REV A with MOSFET G090N06 
########################################

# [output_pin motor_power]
# pin: PC13
# value: 1

########################################
```
Just an idea

KevinOConnor:
I'm not sure what the state of this PR is.  I'm fine with any of the proposals.  But, if we can't decide, I'll remove the config until such time that we can decide (option 2 mentioned above).

-Kevin

timmit99 (commented):
I'd add a note to this and have the pin INITIALLY set to 0 and have the USER change the `value` to 1 once they read the note saying something along the lines of " only change this value if you are using an rev2 board or have made the necessary changes"

KevinOConnor:
Thanks.  I went ahead and made the change to the SKR2 example config (commit 87e12c04).

-Kevin

DeVilliers-Allegiance:
> > Thanks. I think we have a few options here:
> > 
> > 1. Merge this.
> > 2. Remove the existing skr-2 config file.
> > 3. Merge this and add a warning about the early skr-2 revs being broken.
> > 
> > Thoughts?
> > -Kevin
> 
> I'm torn between keeping the config and putting in the fix as well as the warning, and removing it altogether. On the one hand, all SKR2s in the wild should not be used with klipper (so therefore we shouldn't enable their use), but on the other hand, this would help to spread the warning to more users.

Why should they not be used with klipper?

PR #4235, opened 2021-04-29
query_adc: Add get_status wrapper

mental405:
module: query_adc.py

Added a get_status wrapper to report last read adc values from the QUERY_ADC command similar to how QUERY_ENDSTOPS works. 

`QUERY_ADC NAME=EXTRUDER` would add an entry to the dictionary containing the value and the timestamp the reading was taken. Additionally, the pullup value and resistance are added as separate entries if configured on the ADC pin.

Signed-off-by: Paul McGowan mental405@gmail.com

### DIFF
--- a/klippy/extras/query_adc.py
+++ b/klippy/extras/query_adc.py
@@ -8,6 +8,7 @@ class QueryADC:
     def __init__(self, config):
         self.printer = config.get_printer()
         self.adc = {}
+        self.last_values = {}
         gcode = self.printer.lookup_object('gcode')
         gcode.register_command("QUERY_ADC", self.cmd_QUERY_ADC,
                                desc=self.cmd_QUERY_ADC_help)
@@ -22,6 +23,8 @@ class QueryADC:
             gcmd.respond_info(msg)
             return
         value, timestamp = self.adc[name].get_last_value()
+        self.last_values.update({name: {'value': value,
+                                        'timestamp': timestamp}})
         msg = 'ADC object "%s" has value %.6f (timestamp %.3f)' % (
             name, value, timestamp)
         pullup = gcmd.get_float('PULLUP', None, above=0.)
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Can you give some background on how you are using this information?  I thought most of the information would be available already in other areas of the get_status() hierarchy.

As a high-level comment, going forward I think we want to avoid the "query command followed by {printer.xxx.last_yyy}" implementation.  Instead, I think it would be preferable (where possible) to have the get_status() always automatically pull the latest values.  That way, the api_server based frontends can get asynchronous updates to values as they change.

Cheers,
-Kevin

mental405:
I personally am not using the information returned from this at the present time though I probably will in the future for something.  The high level use case would be to report the raw value of an analog pin for something. It could be combined with `gcode_button` to perform specific actions based on the voltage from the ADC at the time the button was pressed.

The specific case was that someone was asking about using an FSR on a nozzle as a probe but the FSR is not _always_ attached to the nozzle. The though being to us QUERY_ADC in the same way one would use QUERY_ENDSTOPS or QUERY_PROBE to get a value from the FSR board. The analog value being reported could then be used to determine if it is safe to attempt a probing move.

I was planning to make an additional generic pin class that does nothing at all but I wanted to get some feedback on that before I started working on it. This module would literally just be [input_pin], it would basically be the opposite of [ouput_pin]; a pin that does nothing but serve a value. 

I have no objections to changing it from a last-value to a live reading. I was just trying to mimic the behavior of the QUERY_ENDSTOP and QUERY_PROBE wrappers since they require the command to be called prior to accessing the value in a macro. I actually feel that would be better overall since command templates get rather  clunky when you start having to run the queries and then evaluate them.

mental405:
I am closing this for now as I have other ideas

PR #4237, opened 2021-04-30
config: added a configuration file for FLSUN QQ-S Pro delta printer

CobraPi:
I noticed that the FLSUN Q5 was in the config section, but not the QQ-S Pro. I've taken the time to test this configuration and it provides a great starting point for people who own the QQ-S Pro.

Signed-off-by: Joseph Orduna < joey.orduna@gmail.com>


KevinOConnor:
Thanks.  The core config looks fine to me.  However, many of the comments and macros are specific to the frontend (eg, mainsail/fluidd) and non-stock components, instead of on the stock printer hardware.  I think it would be preferable for the example configs to focus on the printer hardware.  There is some information on this at: https://www.klipper3d.org/Example_Configs.html

-Kevin

CobraPi:
Thanks for the feedback. I will update the files according to the guidelines in the link.

CobraPi (commented):
I've changed the config file according to the guidelines as well as added it's path to the regression test case.

CobraPi:
I apologize for the minor typo.

KevinOConnor:
Thanks, and sorry for the delay in responding.  In general it looks fine to me.  I'll look to commit next week.  I have a couple of minor questions / comments:
1. Is the change to printer-flsun-q5-2020.cfg intentional?
2. Both flsun configs now use Robin_mini.bin, but the following comment says Robin_nano.bin
3. The extruder rotation_distance=9 is very unusual - it would be great if you could find the actual gear_ratio on the extruder.

-Kevin

KevinOConnor:
Any further updates on this?

-Kevin

CobraPi:
I changed the board from nano to mini. The reason the rotation_distance for the extruder is 9 is because that's what works best for the stock titan extruder, which was re-calculated from the deprecated step_distance parameter I've been using. It has a gear ratio of 3:1, so I can re-calculate the value if you'd like. I've shared this config with the printer's facebook group and people have been getting good results with this value.

KevinOConnor:
Okay - there's a merge conflict preventing me from merging this branch, but otherwise it looks fine to me.  If you can fixup the merge conflict, I'll merge it.

Separately, it's generally preferable to list the rotation_distance of the extruder as the effective circumference of the bolt that engages the filament, and to specify a gear_ratio (and count the actual teeth on the gears instead of using the "marketing" values).

-Kevin

CobraPi:
Okay, I'll try to get those measurements and re-commit soon. In the meantime, I'm not seeing a merge conflict on my end. Is this still an issue?

KevinOConnor:
Okay, thanks.  I committed this change.  I'm not sure why I originally saw a merge conflict, but I don't see it now.

Probably would be worthwhile to fixup the extruder rotation_distance if you get a chance.

-Kevin

Issue #4125, opened 2021-03-30
LPC1769 Serial connection does not work.


pmlody:
There is some problem with UART Serial communication.
I have valid configuration RPI <-> UART. It does work at marlin - device /dev/ttyAMA0 at 115200 and 250000 baud.
When I flash board with Klipper USB firmware, there is valid connection and everything is working.
But if I flash board with Klipper Serial firmware I only got:
```
Timeout on connect
Wait for identify_response
Traceback (most recent call last):
  File "/home/pi/klipper/klippy/serialhdl.py", line 65, in _get_identify_data
    params = self.send_with_response(msg, 'identify_response')
  File "/home/pi/klipper/klippy/serialhdl.py", line 249, in send_with_response
    return src.get_response([cmd], self.default_cmd_queue)
  File "/home/pi/klipper/klippy/serialhdl.py", line 306, in get_response
    cmd_queue)
  File "/home/pi/klipper/klippy/serialhdl.py", line 241, in raw_send_wait_ack
    raise error("Serial connection closed")
error: Serial connection closed
```
in klippy.log
I am certain there is no problem with connection between RPI and SKR board - same UART config works with RPI Zero and RPI 2B+ when using Marlin firmware. There must be some issue in firmware with serial connection itself. I have checked several historical versions, from commit c83688b7bc7fcbc1a413d5707d11a2d2b674d48d till master and every firmware returns same error. I also checked 115200 baud and 250000 baud.
I have proper printer.cfg:
```
[mcu]
serial: /dev/ttyAMA0
restart_method: command
baud: 250000
```
Same error with/without restart_method and baud.

Gilabite:
https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md#help-it-doesnt-work

pmlody:
[klippy.log](https://github.com/KevinOConnor/klipper/files/6249614/klippy.log)
I have added klipper.log file.

IncredibleHohol:
> There is some problem with UART Serial communication.
> 
> I have valid configuration RPI <-> UART. It does work at marlin - device /dev/ttyAMA0 at 115200 and 250000 baud.
> 
> When I flash board with Klipper USB firmware, there is valid connection and everything is working.
> 
> But if I flash board with Klipper Serial firmware I only got:
> 
> ```
> 
> Timeout on connect
> 
> Wait for identify_response
> 
> Traceback (most recent call last):
> 
[...]

PR #4248, opened 2021-05-02
lpc176x: Hotfix for problem serial communication.

totolook:
I revert some code from v0.9.1 in particolarity the pclk
because klipper had a bad sync when trying the connection.
Resolve this issue #4125
 
Signed-off-by: Antonio di Giovanni <crashvampiro@msn.com>

### DIFF
--- a/src/lpc176x/main.c
+++ b/src/lpc176x/main.c
@@ -49,6 +49,13 @@ void
 enable_pclock(uint32_t pclk)
 {
     LPC_SC->PCONP |= 1<<pclk;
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Unfortunately, we can not just revert commit ae89a659, because that would reintroduce the errata problem fixed in that commit.

-Kevin

KevinOConnor:
I'm going to close this as the underlying issue should be fixed with commit 45cd3543.

Thanks,
-Kevin

PR #4252, opened 2021-05-03
added ender3 with robin e3d 1.1

piontec:
This sample config file adds a combination of ender3 and Robin E3D 1.1 board, which is an in-place replacement for the stock creality board. This also solves a few problems I've found no solution in klipper configs so far:

- Robin E3 1.1 uses different Z axis pins than 1.0 - this is corrected and commented in the config
- the default PIN mapping, although correct for physical pin layout, doesn't work with stock ender3 display and connector; this is now fixed and working as well


KevinOConnor:
Please see the guide for example config files: https://www.klipper3d.org/Example_Configs.html

-Kevin

piontec:
Sorry, missed the whole naming convention. Should be corrected now, I think :)

KevinOConnor:
Ah, thanks.  However, we don't include configs for custom printers in the Klipper repo (only stock printer configs).  If this is for a new "board" (and not a particular printer) then we can add it with a "generic" prefix.  The guide above has some details on this.

Cheers,
-Kevin

piontec:
This is for a new board that is dedicated as a replacement for stock ender3/5. Figuring out the display connections took some time, and nobody on discord new the answer. So I think it's really valuable to include a working config to stop people from wasting time. But tell me what you think and what's the best way to include it. Still, I think it makes sense to keep it together as board+printer config, as the board is designed to be 1:1 replacement for ender3 board

KevinOConnor:
I'd guess it would be good to add a file like `generic-mks-robin-e3d.cfg` with the definitions for the board.  As for the screen, perhaps we can add a pin aliases section for the pins on the particular header - similar to the EXP1 and EXP2 aliases found on many other boards.

-Kevin

piontec:
The screen is a problem: the board has physically EXT1, EXT2, and EXT3 connectors. Pins are mapped correctly for EXT1 and EXT2. Now, EXT3 on the board conatins physically the same signal pins as EXT1 and EXT2, just a selection of pins from both, to make it easy to attach stock Ender screen. And as such I can't make another set of aliases, as they would duplicate signal names :( Is there any solution for that?

KevinOConnor:
It is possible to have multiple aliases to the same pin.  (It's just not possible to use the same pin for two different functions.)  FWIW, I'd avoid the name `EXT3` - maybe go with something like `CR_EXP3`.

-Kevin

KevinOConnor:
Any further updates on this?

-Kevin

piontec:
OK, I added my findings as comments in the `generic-mks-robin-e3.cfg`. That way anybody replacing stock board with this dedicated one will find all the pin mappings in a single place. Also, added missing 'material detect' pin and remapped EXT3 connector as you suggested.

KevinOConnor:
Okay, we can add additional PIN definitions and comments to the config/generic-mks-robin-e3.cfg file.  However, the kit-creality-ender3-mks-robin-e3-2021.cfg file does not fit the current config file scheme, as we don't include config files for after market or customized printers (there are just too many of them).

-Kevin

piontec:
Forgot to remove it, should be good now.

KevinOConnor:
Okay, thanks.  However, I don't see a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

piontec:
Commits edited.

KevinOConnor:
Thanks.  However, I just noticed that the new CR_CR3_x definition has the pins in the reverse order that we normally list them.  That is, pin 10 should be 5V and pin 9 should be GND.  (As described in config/sample-lcd.cfg .)  I know some boards have labels with reverse ordering, but following that order is very confusing.

-Kevin

piontec:
OK, named PINs so that pin 10 is 5V

KevinOConnor:
Thanks.

-Kevin

PR #4253, opened 2021-05-03
Add default display configuration for st7920

Reffter:
Default display configuration for st7920 based displays, connected on Ender3-EXP1 port


KevinOConnor:
We don't define displays in "generic" config files (as those files are describing the board and we have no way of knowing what display is attached to it).  Please see the example config file guide at: https://www.klipper3d.org/Example_Configs.html

-Kevin

PR #4255, opened 2021-05-03
SDCard update and compilation support for Fysetc S6 and Spider

wlhlm:
The Fysetc S6 and Spider come with an SDCard bootloader. This PR updates Kconfig to allow configuring the correct offset on an STM32F446 to prevent Klipper from overwriting the bootloader. Also new board definitions for `flash-sdcard.sh` are introduced.

This has been tested with an S6 v2 and a Spider v1.


KevinOConnor:
Thanks.  Looks good to me.  @Arksine, do you have any comments on this?

-Kevin

Arksine:
Looks good to me.  I would think that these boards would support DFU as well, but certainly no harm in adding `flash-sdcard.py` support.

KevinOConnor:
Thanks.

-Kevin

PR #4256, opened 2021-05-03
Printer.cfg for an Anycubic Chiron.

garbled1:
Tested with a stock chrion unmodified.
Print quality with this is far better than stock firmware.


KevinOConnor:
Please see the example config file guide: https://www.klipper3d.org/Example_Configs.html

-Kevin

KevinOConnor:
Closing as it looks like this PR has gone stale.

-Kevin

PR #4258, opened 2021-05-03
Anycubic Mega Zero 1 and 2 printer files

MrNoob0:
printer.cfg files for the anycubic mega zero 1 and 2


KevinOConnor:
Thanks, but commits need to have a signed-off-by line for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .  Also, the files seem to lack a ".cfg" extension - see the guide for example config files at: https://www.klipper3d.org/Example_Configs.html

The extruder rotation distance doesn't look correct (it's very small).  There's probably a gearing mechanism on the printer.  It would be worthwhile to document what that is.

Finally, I don't understand what the comment "If you are running the mega zero with the bigtreetech upgrade board "Bigtreetech SKR MINI MZ you will need to use the existing config in klipper/config" means.

Cheers,
-Kevin

KevinOConnor:
Thanks.  Both configs look good to me, but I'd need a signed-off-by line to commit them - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

Also, if the printers are using bondtech type extruders then it is preferable to use `gear_ratio: 50:17` and specify the rotation_distance as a circumference (see https://github.com/KevinOConnor/klipper/blob/master/docs/Rotation_Distance.md#using-a-gear_ratio ).

-Kevin

KevinOConnor:
Closing as it looks like this PR has gone stale.

-Kevin

PR #4261, opened 2021-05-04
Fixed typo in when comparing `print_stats.state` causing menus to not show up

DanyYanev:
As you can reference in https://github.com/KevinOConnor/klipper/blob/master/klippy/extras/print_stats.py

The state that is set is "paused" not "pause" causing some of the submenus to not show up.


KevinOConnor:
Thanks.

-Kevin

PR #4263, opened 2021-05-04
Ignore files created by PyCharm IDE

cbc02009:
module: .gitignore

Ignore any files created by the PyCharm IDE when used to work on Klipper codebase. This is simply a QOL change so that one doesn't need to wade through tons of virtual environment and settings files to find the code changes to be committed.

Signed-off-by: Christopher Conroy <cbc02009@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ out
 .config
 .config.old
 klippy/.version
+
+# Ignore files created by PyCharm
+venv/
+.idea/


KevinOConnor:
Sorry, but we don't only use gitignore to ignore temporary files that the Klipper build directly produces - we don't add editor files to gitignore.  There is a global gitignore file you can use on your machine to ignore editor files for all git repos on your machine.  See the git documentation for the details.

-Kevin

cbc02009:
Sorry, must have missed that in the documentation. I'll close this pull request.

PR #4266, opened 2021-05-05
fix STM32F446 ADC3 bug

bigtreetech:
PF3-PF10 on STM32F446ZE can only use ADC3,  So F446 also needs to include ADC3


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit them - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .  However, I made this change under my name - commit bde6215c.

-Kevin

PR #4271, opened 2021-05-07
safe_z_home: Always perform z limit checks (fixes #4249)

### LINKED ISSUE #4249: G28 fails when when z_pos is near to z_max + z_hop is set & M84 is issued (or "idle timeout" has passed)
Hi,
I think there's incorrect behaviour of G28( Z) when last known_position and "_[safe_z_home] z_hop_" is higher than _position_max_ of Z axis.

Lets say I have z_position_max = 215 and z_hop = 10:
- execute homing G28
- move Z axis to the z_position_max (either by printing or just by G0 Z215)
- execute G28 again.

In this scenario, G28 correctly says "_// No zhop performed, target Z out of bounds: 225.0_", it _does_ the homing of axis and everything is fine.

But if:
- execute homing G28
- move Z axis to the z_position_max
- wait until idle timeout or issue M18/M84 without waiting
- execute G28

In this scenario homing fails with "_!! Move out of range: 137.500 141.400 225.000 [0.000]_" and no homing is performed. I have to execute G28 one more time to make real homing (with message "_// No zhop performed..._").

[klippy.log](https://github.com/KevinOConnor/klipper/files/6411783/klippy.log)

FHeilmann:
This change causes the z boundary check to always be performed, regardless of the homed status of the Z axis. 

As described in #4249, not doing this can lead to errors if the motors were disabled near z-max (in this case, the zhop is currently performed, which causes out-of-bounds errors and (possibly worse) leaves the z axis in a homed state).

Signed-off-by: Florian Heilmann <Florian.Heilmann@gmx.net>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1128,11 +1128,8 @@ home_xy_position:
 #   applied to any homing command, even if it doesn't home the Z axis.
 #   If the Z axis is already homed and the current Z position is less
 #   than z_hop, then this will lift the head to a height of z_hop. If
-#   the Z axis is not already homed, then prior to any XY homing
-#   movement the Z axis boundary checks are disabled and the head is
-#   lifted by z_hop. If z_hop is specified, be sure to home the Z
-#   immediately after any XY home requests so that the Z boundary
-#   checks are accurate. The default is to not implement Z hop.
+#   the Z axis is not already homed the head is lifted by z_hop.
+#   The default is to not implement Z hop.
 #z_hop_speed: 20.0
 #   Speed (in mm/s) at which the Z axis is lifted prior to homing. The
 #   default is 20mm/s.
--- a/klippy/extras/safe_z_home.py
+++ b/klippy/extras/safe_z_home.py
@@ -34,22 +34,23 @@ class SafeZHoming:
 
         # Perform Z Hop if necessary
         if self.z_hop != 0.0:
-            # Check if Z axis is homed or has a known position
+            # Check if Z axis is homed and its last known position
             curtime = self.printer.get_reactor().monotonic()
             kin_status = toolhead.get_kinematics().get_status(curtime)
-            if 'z' in kin_status['homed_axes']:
-                # Check if the zhop would exceed the printer limits
-                pos = toolhead.get_position()
-                if pos[2] + self.z_hop > self.max_z:
-                    gcmd.respond_info(
-                        "No zhop performed, target Z out of bounds: " +
-                        str(pos[2] + self.z_hop))
-                elif pos[2] < self.z_hop:
-                    self._perform_z_hop()
-            else:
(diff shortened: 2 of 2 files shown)


FHeilmann:
I think the fix you suggest will cause problems for scenarios where the printer will lose power at a z position where a zhop will result in a crash into zmax and the gantry doesn't drop.

I don't see how we can support both use-cases. Either we assume printers that always require the zhop, no matter where the z-axis lost power, or we decide it's reasonable to assume that if the z-axis lost power near zmax that the rail is still near that position and a zhop should not be performed (which is what this changeset accomplishes).

KevinOConnor:
> I think the fix you suggest will cause problems for scenarios where the printer will lose power at a z position where a zhop will result in a crash into zmax and the gantry doesn't drop.

FWIW, I would consider the carriage contacting the top of the gantry a much smaller concern than having the nozzle contact the bed.

-Kevin

FHeilmann:
> > I think the fix you suggest will cause problems for scenarios where the printer will lose power at a z position where a zhop will result in a crash into zmax and the gantry doesn't drop.
> 
> FWIW, I would consider the carriage contacting the top of the gantry a much smaller concern than having the nozzle contact the bed.
> 
> -Kevin

Sure, I'll see if I can implement it in a nice-ish way :)

FHeilmann:
I chose to take a detailed look at the logic and simplify it somewhat. I also made the following changes:

- Remove the "out of bounds message". It was more of a debug statement and didn't really serve any purpose, other than maybe confuse the user
- Bring the extra in line with the documentation. Previously, if the z-axis was homed, but below z_hop, it would be lifted **by** z_hop and not **to** z_hop as suggested by the documentation. This also makes the behavior a bit more predictable, because previously (with, for example, `z_hop = 10`) a homed z_axis sitting at `9.9` would be lifted to `19.9`, whereas a homed z_axis sitting at `10` would not be lifted at all.
- Remove `_perform_z_hop()` as it is not used by later parts of the code anyway, and its usage doesn't really improve code readability/deduplication anymore.
- Remove the parts of the documentation regarding the z boundary checks. Since the z-axis is now returned to its unhomed state properly after the z-hop, this part of the documentation is obsolete.

I tried to test all cases, but just to be sure I'll list them here so I can be made aware if I missed any:

- Z axis homed:
  - `position_z < z_hop` -> z_hop performed, axis is lifted to z_hop
  - `z_hop >= z_hop` -> no z_hop performed
- Z axis not homed:
  - `z_hop` is always performed, regardless of what position the toolhead is physically or where its last known position was.

Let me know if there is any issues you have with my changes or other modifications I should make.

-Florian

KevinOConnor:
Thanks.  It looks good to me.  If there are no further comments, I'll look to merge early next week.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #4273, opened 2021-05-07
Add TEMPERATURE_WAIT_EQUILIBRIUM command

garethky:
module: temperature_wait_equilibrium

This adds a new `TEMPERATURE_WAIT_EQUILIBRIUM` command that waits for a temperature sensor's rate of change to fall to a desired level. This is useful in printers that have large thermal systems (think aluminum build plates or heated chambers) where you want to consistently bring those systems into thermal stability before starting a print. Warmup can now be part of your print start routine without any downsides. "Hot Starts" for successive prints don't have to be excessively long. Prints always start under nearly identical thermal conditions and environmental variations don't throw the process off.


KevinOConnor:
Ah, good points.  I didn't think enough about that when I made my original comments.

I'd guess the code could use some refactoring, so that all the temperature sensors have a single place that can do smoothing and slope calculation.  Perhaps something like a `TemperatureSensor()` helper class.  I'll try to do some further research.

-Kevin

garethky:
Thanks, I'll take a pass at the refactor this weekend. Once it's done I'll have to do some testing of higher `smooth_time` to see how that behaves for small slope values.

garethky:
This is an update to show where I have gotten to. I'm not satisfied with the current state. I think its too risky to not have `Heater` inherit from `TemperatureSensor`. The current code deletes some properties related to temperature from Heater that code that I cannot see (such as front ends or private extras) might rely on. I'll be making another update to fix this by having `Heater` be a special kind of `TemperatureSensor`. That sounds weird but it really is the way the code works today. Any feedback is appreciated!

What is working:
`TemperatureSensor` is a wrapper for a hardware temp sensor. It encapsulates reading of the config and maintains temp, last sample time, min, max, smoothed temp and temp slope. All code that lived elsewhere to do the above has been consolidated. It also allows the `temperature_callback` mechanism to be chained so derived classes don't have to re-implement any of that functionality.

`TEMPERATURE_WAIT` for `MAX_SLOPE` does work but I haven't tested it extensively. The units for `MAX_SLOPE` are degrees per second, which I don't think is intuitive. I want to change it to degrees per minute.

This builds and works on my printer. That's probably not saying much. It actually built and passed tests and still failed to start. Re-factoring dynamic languages is always treacherous.

garethky:
Updated to make `Heater` extend `TemperatureSensor`. I am much happier with that change in.

I have run it on my printer and everything seems to be operating as expected. `MAX_SLOPE` in degrees per minute is much easier to understand. I am waiting on my top bed sensor to hit less than 0.5 degrees per minute before mesh leveling. Having smoothed temps reported for all sensors in the front end is also really nice. 😁

I wish the TEMPERATURE_WAIT's output to the console included temp_slope and the MAX_SLOPE target if that was being waited on. Maybe I can get that in by delegating the individual outputs to each sensor.

garethky:
How about this, for TEMPERATURE_WAIT print only the sensor being waited on in this format:

`<gcode_id|SENSOR>:<smoothed_temp>@<slope> /<MINIMUM>/<MAXIMUM>/<MAX_SLOPE>`

e.g. if I want to wait for the bed temp to be over 60C and stabilized at less than 1/2 degree per minute of slope I would issue:

`TEMPERATURE_WAIT SENSOR=heater_bed MINIMUM=60 MAX_SLOPE=0.5`

And the temp wait lines would look like:
```
B:58.5@1.0 /60/inf/0.5
B:58.5@1.1 /60/inf/0.5
```

The same thing for my top bed sensor without a gcode Id:
```
TEMPERATURE_WAIT SENSOR=temperature_sensor top_bed MINIMUM=60 MAX_SLOPE=0.5
temperature_sensor top_bed:58.5@1.0 /60/inf/0.5
temperature_sensor top_bed:58.6@1.1 /60/inf/0.5
```

What do you think?

I have this working and its great to see both the temp and slope and to have the 3 parameters as a reference. The only down side is it breaks Fluidd's "hide temp waits" feature in the console because that expects them to conform to a [specific regex](https://github.com/cadriel/fluidd/blob/a81c1b76a6cb7951f642f7690a01ee932a8ce33e/src/store/console/getters.ts#L14). That actually block M105 output as well which seems broken! So maybe we could help the front ends do better by starting the lines with a fixed string like:

```
TEMPERATURE_WAIT B:58.5@1.0 /60/inf/0.5
TEMPERATURE_WAIT B:58.5@1.1 /60/inf/0.5
```

mjoconr:
I would like to say that I've been asking around and trying to work out how to do this for the last month or so. I would really like to see a feature like this in the code as soon as possible.

garethky:
`temp_slope` is too noisy to use with low values of `MAX_SLOPE`. At 0.5 degrees/minute sensor noise drowns out the real temp slope. I didn't want to bring back my least squares code so I experimented with the existing code and settled on this: `smoothed_temp_slope` is the derivative of `smoothed_temp` that is then smoothed with the existing time based smoothing algorithm. This produces output that's comparable to least squares in its consistency and low noise levels. e.g.:
Screen Shot 2021-05-26 at 2 25 17 PM

I'm happy running this on my printer 😁. Alright, @KevinOConnor its ready for you to look at.

garethky:
Sorry, white space, fixed; https://github.com/garethky/klipper/actions

KevinOConnor:
Thanks for looking at this!

I have a few comments:
1. I think it would be preferable for the new TemperatureSensor class to be a "helper class" and not a "base class".  That is, I think it would be preferable for each user to explicitly instantiate a `self.temp_sensor = heaters.TemperatureSensor(config)` and obtain information from that class via method calls (eg, `self.temp_sensor.get_temperature()`) and not via member accesses.  This goes to the Klipper coding style of not accessing member variables of classes defined in other files (mentioned briefly in docs/Code_Overview.md).
2. Because this is a low-level change I'd like to be careful with the change.  That will mean breaking out just the TemperatureSensor refactoring to a commit by itself.  I would also ask that the change not have any unrelated whitespace changes in it.  This is important because if a regression is introduced we'll need to be able to use `git bisect` and (if necessary) `git revert`.  Mixing core changes with other changes makes that process much harder.
3. This change moves `measured_min_temp` and `measured_max_temp` from PrinterSensorGeneric to TemperatureSensor.  These fields are quirks of PrinterSensorGeneric that I'm looking at possibly deprecating, so they shouldn't be moved to other sensors.
4. Did you have any high-level thoughts on how this would be extended to the bme280, temperature_host, temperature_mcu, etc. objects in the future?

Thoughts?
-Kevin

garethky:
> Thanks for looking at this!
> 
> I have a few comments:
> 
>     1. I think it would be preferable for the new TemperatureSensor class to be a "helper class" and not a "base class".  That is, I think it would be preferable for each user to explicitly instantiate a `self.temp_sensor = heaters.TemperatureSensor(config)` and obtain information from that class via method calls (eg, `self.temp_sensor.get_temperature()`) and not via member accesses.  This goes to the Klipper coding style of not accessing member variables of classes defined in other files (mentioned briefly in docs/Code_Overview.md).

Happy to code it up that way! Using a helper class means loosing all members on Heater and PrinterSensorGeneric that anyone may have been accessing. I like that better, I just figured it was higher risk.

> 
>     2. Because this is a low-level change I'd like to be careful with the change.  That will mean breaking out just the TemperatureSensor refactoring to a commit by itself.  I would also ask that the change not have any unrelated whitespace changes in it.  This is important because if a regression is introduced we'll need to be able to use `git bisect` and (if necessary) `git revert`.  Mixing core changes with other changes makes that process much harder.

I'm new to this GitHub Pull Request workflow. At work we use local branches and squash merge/rebase until the diff with mainline looks exactly the way we want it. Annoyingly on GitHub that workflow requires force push and most people frown on that. I don't mind force-pushing to my own repo, would you be ok with that? I could then get the patch branch looking exactly the way we want.

Should we keep this current branch for reference and start a new branch/pull-request just for the the refactor?

> 
>     3. This change moves `measured_min_temp` and `measured_max_temp` from PrinterSensorGeneric to TemperatureSensor.  These fields are quirks of PrinterSensorGeneric that I'm looking at possibly deprecating, so they shouldn't be moved to other sensors.

So given #1, I'm assuming that Heater and PrinterSensorGeneric will share some `getTemperatureSensor()` method that unites them as "things that contain a TemperatureSensor". PrinterSensorGeneric will loose most of its members but keep `measured_min_temp` and `measured_max_temp`. Sound right to you?

> 
>     4. Did you have any high-level thoughts on how this would be extended to the bme280, temperature_host, temperature_mcu, etc. objects in the future?

This is my first submission here and I don't know the code base that well...so bring the 🧂

There are a lot of kinds of sensors besides temperature (pressure, force, humidity, strain, acceleration etc.). And then there are sensor modules that measure several of those things. E.g. BME280 measures Pressure, Humidity and Temperature.

If this was Java I would make a set of interfaces for 'physical' sensors to implement for each measurement type that the physical device will expose (IHumiditySuppplier, ITemperatureSupplier, IPressureSupplier). That keeps the required implementation small but consistent for contributors. Then I would use the wrapper pattern to layer on common additional functionality, like what TempertureSensor is doing, to create a "logical sensor". Then register all of the logical sensors individually. Individual logical sensors support composition. Example:

```
[heater_bed]
#...
[BME280 chamber_conditions]
#...
[controller_fan]
temp_sensors: temperature:chamber_conditions, temperature:heater_bed
#...
```

[controller_fan] doesn't need access to Heaters, it can just use `printer.lookup_object("temperature:heater_bed")`. Everything it gets is type TemperatureSensor.


...anyway, this is _wildly_ off topic at this point! 😆

> 
> 
> Thoughts?
> -Kevin

KevinOConnor:
> I'm new to this GitHub Pull Request workflow. At work we use local branches and squash merge/rebase until the diff with mainline looks exactly the way we want it. Annoyingly on GitHub that workflow requires force push and most people frown on that. I don't mind force-pushing to my own repo, would you be ok with that? I could then get the patch branch looking exactly the way we want.

FWIW, I frequently rebase/force-push to my dev branches (work-xxx-ddddyymm branches).  I don't force-push to the Klipper master branch.  All patches to the master branch are rebase'd to it (never merge).

> Should we keep this current branch for reference and start a new branch/pull-request just for the the refactor?

That's up to you.

> I'm assuming that Heater and PrinterSensorGeneric will share some getTemperatureSensor() method that unites them as "things that contain a TemperatureSensor". PrinterSensorGeneric will loose most of its members but keep measured_min_temp and measured_max_temp. Sound right to you?

Yes.  FWIW, it may make sense for the callers to use `self.temp_sensor.get_status()` instead of adding a `getTemperatureSensor()` method - I'm not sure.

Thanks,
-Kevin

garethky:
I'm going to withdraw this. I've been working on a macro to solve the problem that others can use:
https://github.com/garethky/klipper-voron2.4-config/blob/mainline/klipper_config/heatsoak.cfg

I've learned a few things in writing the macro:
* Being able to stop the heat soaking process is a nice usability win. This PR wouldn't have supported that.
* Detecting small rates of temperature change (such as when heating a chamber) below the noise threshold of the sensor isn't possible with simple smoothing. It required collecting 20 seconds worth of temperature data and computing the regression on that. This PR would not have solved that either.

I think the one valuable thing in this PR is smoothing of temperatures on sensors that are not heaters. But that could be done in a much simpler PR.

KevinOConnor:
Okay, thanks for the update.  I apologize that I wasn't able to get to your other PR in a timely manner.

-Kevin

PR #4275, opened 2021-05-08
virtual_sdcard: fix case sensitive matching

Arksine:
This is minor fix to allow the `virtual_sdcard` to load case sensitive files.  The as the value returned by `get_file_list()` is a tuple, the original check would always evaluate to False.


KevinOConnor:
Thanks.

-Kevin

Issue #4126, opened 2021-03-30
Missing printer. endpoints for Filament motion sensor


zellneralex:
I am not sure if this is a bug report or a feature request, so I let Kevin decide ...

Installing a Filament motion sensor I found out that the following printer. endpoints are missing. As it uses the same GCodes I was expecting that it would use similar printer. variables like:

printer["filament_motion_sensor <config_name>"].enabled
printer["filament_motion_sensor <config_name>"].filament_detected 

in the same way as its already defined for:

printer["filament_switch_sensor <config_name>"].enabled: Returns True if the switch sensor is currently enabled.
printer["filament_switch_sensor <config_name>"].filament_detected: Returns True if the sensor is in a triggered state.

Arksine:
The module needs a `get_status()` method.  The contributor said he was going to look into it.

PR #4281, opened 2021-05-10
filament_motion_sensor: add get_status

zellneralex:
this change adds the get_status() to the filament_motion_sensor. See #4126


KevinOConnor:
Thanks.

-Kevin

PR #4286, opened 2021-05-12
config: Add printer-fokoos-odin5-f3.cfg

matthewlloyd:
This is a new Cartesian direct drive printer from FOKOOS (https://www.fokoos.com/) built around the Robin Nano v1.2, with several hundred owners so far. It's available direct and from Amazon. I am an owner and not affiliated with FOKOOS in any way. Klipper takes it from good to great.

I have run the relevant test, here's an excerpt (I commented out most of the other configs to save time):

```
    Starting test/klippy/printers.test (generic-creality-v4.2.7.cfg)
    Starting test/klippy/printers.test (generic-creality-v4.2.10.cfg)
    Starting test/klippy/printers.test (printer-fokoos-odin5-f3-2021.cfg)
    Starting test/klippy/printers.test (printer-tronxy-x5sa-v6-2019.cfg)
    Starting test/klippy/printers.test (printer-tronxy-x5sa-pro-2020.cfg)
```


KevinOConnor:
Thanks.  I committed this along with some minor changes on top (commit b4a22297).

Separately, if possible, it's best to use PID for the heated bed - otherwise, "Z banding" defects are common.

-Kevin

matthewlloyd:
Awesome, thanks for the quick merge.

Thanks also for the tip about PID for the heated bed! I copied the Marlin config there, but looking at the electronics I don't see any relays, so I will test PID and open a new PR if the heater does in fact tolerate PWM.

PR #4288, opened 2021-05-12
config: Enable PID for bed heater in printer-fokoos-odin5-f3-2021.cfg

matthewlloyd:

I neither hear nor see mechanical relays and PWM control seems to be working well.

### DIFF
--- a/config/printer-fokoos-odin5-f3-2021.cfg
+++ b/config/printer-fokoos-odin5-f3-2021.cfg
@@ -120,7 +120,12 @@ max_temp: 275
 heater_pin: PA0
 sensor_type: EPCOS 100K B57560G104F
 sensor_pin: PC0
-control: watermark
+control: pid
+# Tuned at 60C.
+pid_kp = 58.214
+pid_ki = 1.075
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks, however it would be preferable to use `option: value` syntax (instead of `=`).  This is mentioned briefly at https://www.klipper3d.org/Example_Configs.html

-Kevin

matthewlloyd:
Fixed. The equals came from the SAVE_CONFIG section... feature request to have this use colons instead, if you agree, I will implement it and open a PR.

#*# <---------------------- SAVE_CONFIG ---------------------->
#*# DO NOT EDIT THIS BLOCK OR BELOW. The contents are auto-generated.
#*#
#*# [heater_bed]
#*# control = pid
#*# pid_kp = 58.214
#*# pid_ki = 1.075
#*# pid_kd = 788.077

KevinOConnor:
Thanks.

Separately, I don't think it matters what the layout of the autogenerated save config section is - it's just nice to be consistent in the config/*.cfg example files.

Cheers,
-Kevin

PR #4289, opened 2021-05-13
add STM32F446 32K bootloader offset for Octopus board

bigtreetech:


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line with a full name for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

However, I went ahead and made this change with my sign off - commit 338bc82d.

-Kevin

PR #4293, opened 2021-05-15
Aarch64 Arch Linux install script

cbc02009:
module: Install Scripts

Klipper install script for Arch Linux Aarch64 (raspberry pi 3b+/4 or equivalent).

I've written a klipper install script for aarch64 Arch linux that I'd like to share in case someone else finds it useful. I've tested this and verified that it installs klipper successfully. The only issue is that I only have a BTT SKR 1.4 available to test firmware flashing. I've verified that it builds the binary correctly and that flashing the firmware via `./scripts/flash-sdcard.sh` works correctly, but I am unable to test the functionality for any ST or AVR boards.

I am unsure if this is useful, but figured I'd offer it and let the maintainers decide.

Signed-off-by: Christopher Conroy <battloid.kouji@gmail.com>

### DIFF
--- a/scripts/install-arch-aarch64.sh
+++ b/scripts/install-arch-aarch64.sh
@@ -121,6 +121,12 @@ verify_ready()
         echo "This script is only compatabile with aarch64-based installations"
         exit -1
     fi
+
+    if ! which sudo > /dev/null 2>&1; then 
+        echo "Sudo is not installed, please install Sudo and add user to sudoers list."
+        exit -1
+    fi
+
 }
 
 # Force script to exit if an error occurs
--- a/scripts/install-arch-aarch64.sh
+++ b/scripts/install-arch-aarch64.sh
@@ -89,7 +89,7 @@ WantedBy=multi-user.target
 Type=simple
 User=$KLIPPER_USER
 RemainAfterExit=yes
-ExecStart=${PYTHONDIR}/bin/python ${SRCDIR}/klippy/klippy.py ${CONFIGDIR}/printer.cfg -l ${KLIPPER_LOG}
+ExecStart=${PYTHONDIR}/bin/python ${SRCDIR}/klippy/klippy.py ${CONFIGDIR}/printer.cfg -l ${KLIPPER_LOG} -a /tmp/klippy_uds
 EOF
 # Use systemctl to enable the klipper systemd service script
     sudo systemctl enable klipper.service
--- a/scripts/install-arch-aarch64.sh
+++ b/scripts/install-arch-aarch64.sh
@@ -46,7 +46,7 @@ install_packages()
     # Install desired packages
     report_status "Installing packages..."
     paru -S --needed ${PKGLIST}
-     
+
     # Install ARM build package. PKGBUILD recompresses it, so this takes a *LONG* time
     # even on a Pi 4
     report_status "Installing arm-none-eabi-gcc. This will take some time..."
@@ -122,7 +122,7 @@ verify_ready()
         exit -1
     fi
 
-    if ! which sudo > /dev/null 2>&1; then 
+    if ! which sudo > /dev/null 2>&1; then
         echo "Sudo is not installed, please install Sudo and add user to sudoers list."
         exit -1
     fi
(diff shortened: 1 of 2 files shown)


scripts/install-arch-aarch64.sh
```diff
@@ -0,0 +1,138 @@
...
+    PKGLIST="${PKGLIST} avrdude avr-gcc avr-binutils avr-libc"
+
+    # Install desired packages
+    report_status "Installing packages..."
+    paru -S --needed ${PKGLIST}
+     
+    # Install ARM build package. PKGBUILD recompresses it, so this takes a *LONG* time
+    # even on a Pi 4
+    report_status "Installing arm-none-eabi-gcc. This will take some time..."
+    paru -S --skip-review gcc-arm-none-eabi-bin
```
544449, line 53: for me it doesn't work. I think --skip-review should be --skipreview.

cbc02009: Yup, that's what I get for changing something last minute. Should be fixed now.

wlhlm:
I'm also running Klipper on Archlinux ARM and I'm not sure this is well suited for the upstream Klipper repository (IMO, but I'm not the Klipper maintainer of course). As you've noticed, Archlinux ARM doesn't packages the embedded ARM gcc required to build Klipper for various ARM MCUs. Your solution is to install a binary package of said gcc from an unofficial/user-maintained package in the AUR which is risky. Also the install script only supports the `paru` AUR helper, which is problematic since there are no officially endorsed AUR clients and users may choose to use all kinds of different helpers. Supporting all is impossible but choosing just one also isn't great.

I think the best solution for this is to turn this into a PKGBUILD and submit this as `klipper-virtualenv-git` to the AUR.

tadejberlic:
I can confirm that it works on my AVR board.
My Aarch64 Arch Linux didn't come with sudo installed; 
so maybe a comment should be added that user should install sudo and add user to a sudoers file?

cbc02009:
> 
> 
> I'm also running Klipper on Archlinux ARM and I'm not sure this is well suited for the upstream Klipper repository (IMO, but I'm not the Klipper maintainer of course). As you've noticed, Archlinux ARM doesn't packages the embedded ARM gcc required to build Klipper for various ARM MCUs. Your solution is to install a binary package of said gcc from an unofficial/user-maintained package in the AUR which is risky. Also the install script only supports the `paru` AUR helper, which is problematic since there are no officially endorsed AUR clients and users may choose to use all kinds of different helpers. Supporting all is impossible but choosing just one also isn't great.
> 
> I think the best solution for this is to turn this into a PKGBUILD and submit this as `klipper-virtualenv-git` to the AUR.

Since the AUR package literally just contains a link to the official ARM toolchain builds and was included only to make updating it easier do you think it would make more sense for the script to just download and install the file directly from source instead of going through AUR? That would likely speed up the install process as well.

I don't have a good answer for the AUR issue in general. You're correct that it shouldn't require a specific AUR handler, but to be fair the current `install_arch.sh` script uses pacmatic (which is for Manjaro IIRC) and the command calls in the script are written in a way that assumes pacmatic will be used as far as I can tell. 

Maybe an ideal alternative would be to include the switches in an `AURPARAMS` variable and just have the final call be:
`$AURCLIENT ${AURPARAMS} ${AURLIST}`

That way the user could still choose to use a different AUR client by editing `AURCLIENT` and `AURPARAMS` and commenting out the `install_paru` line. Or maybe the inverse would be better, requiring the user to uncomment `install_paru` if they know they want it?

My goal here was to update the current `install_arch.sh` script to be easier to use with the default set up (headless on a raspberry pi) as none of the GCC packages in the current script are available for ARM, only for x86_64.


> 
> 
> I can confirm that it works on my AVR board.
> My Aarch64 Arch Linux didn't come with sudo installed;
> so maybe a comment should be added that user should install sudo and add user to a sudoers file?

Thank you for testing that for me, much appreciated. I thought the default Arch images always came with sudo installed, but I guess there are cases where that could not be true. I'll add checks for sudo to `verify_ready`.

KevinOConnor:
Okay thanks, however I'd tend to agree with the previous reviewer in that I'm leery of having scripts in the master Klipper repo that pull in 3rd party packages and run them as root.  In general, I view the install scripts has "guides" more than an "end user tool".  So, I don't think we necessarily have to have install scripts for the every distribution.

-Kevin

KevinOConnor:
Thanks, however, I'm closing this ticket as I'm not sure it is a good candidate for the Klipper master repo at this time.

-Kevin

PR #4301, opened 2021-05-18
Error in Ultimaker Config File

TehOcean:
Hi,

This is my first PR so please let me know if I am doing something wrong.

I believe the stepper current scale is incorrect, according to the [Ultimaker 2 Marlin](https://github.com/Ultimaker/Ultimaker2Marlin/blob/8698fb6ad43490ce452d044330f9ca3a17027dfc/Marlin/pins.h#L1221), PWM fully enabled should refer to 2A not 1.5A. 

On an Ultimaker 2, I had issues with steppers skipping with scale: 1.5, which were fixed when I changed to scale: 2.0.

Cheers,

### DIFF
--- a/config/generic-ultimaker-ultimainboard-v2.cfg
+++ b/config/generic-ultimaker-ultimainboard-v2.cfg
@@ -103,7 +103,7 @@ static_value: 1.0
 [output_pin stepper_xy_current]
 pin: PL5
 pwm: True
-scale: 1.5
+scale: 2.0
 # Max power setting.
 cycle_time: .000030
 hardware_pwm: True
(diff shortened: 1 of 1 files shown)


KevinOConnor:
@averyct - do you have any comments on this change?

Separately, commits should have a signed-off-by line for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .  However, I can make this change under my name if that's a concern.

-Kevin

KevinOConnor:
Okay - I changed the config file (commit 4f939c18).

-Kevin

PR #4305, opened 2021-05-19
Makefile: include debug symbols in the .elf

l29ah:


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .

Separately, if we're changing the debug setting, shouldn't we use `-ggdb3` in CFLAGS so that it applies to all files generated?

-Kevin

l29ah:
On Thu, May 20, 2021 at 04:21:17PM -0700, KevinOConnor wrote:
> Thanks.  Commits need to have a signed-off-by line for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .

Fixed.

> Separately, if we're changing the debug setting, shouldn't we use `-ggdb3` in CFLAGS so that it applies to all files generated?

Yes. Strangely, if i change -g (that is already present and should have produced the symbols) to -ggdb3 there, nothing changes, the symbols aren't produced, but if i append it at the end of CFLAGS, it works. I couldn't make sense of it from the gcc man page.

-- 
()  ascii ribbon campaign - against html mail
/\  http://arc.pasp.de/   - against proprietary attachments

KevinOConnor:
> Yes. Strangely, if i change -g (that is already present and should have produced the symbols) to -ggdb3 there, nothing changes, the symbols aren't produced, but if i append it at the end of CFLAGS, it works.

Okay, that's pretty strange.  We can't remove the `-g` from the basic CFLAGS as that variable is used in other builds.  What tool are you using to detect that debugging is in the final/intermediate files?

-Kevin

l29ah:
> We can't remove the -g from the basic CFLAGS as that variable is used in other builds.

This PR doesn't remove -g from CFLAGS.

> What tool are you using to detect that debugging is in the final/intermediate files?

gdb.

KevinOConnor:
Ah - I misunderstood your most recent changes.  Looks fine to me.  However, it looks like src/pru/pru.lds needs an update (probably by adding something like `*( .gnu.* )` to it).

-Kevin

PR #4307, opened 2021-05-20
Automated mcu flashing

KevinOConnor:
This PR is mainly for discussion around automated micro-controller flashing.  The goal is to see if we can arrive at a system where flashing is done from a web interface for most users (that is, they do not need to ssh into the host computer to flash their micro-controller).

The code on this branch adds some prototype code that can enable Klipper to compile and deploy the mcu code from commands sent via the API server.  Much of the code is just "stubbed out" - it's primarily intended for discussion.  (However, I was able to compile and flash my btt-skr-mini-e3-v2 via a manually crafted api server command.)

At a high-level the idea is that each example config file could contain information on how to configure and compile the micro-controller code.  For example, the `config/generic-bigtreetech-skr-mini-mz.cfg` file might have something like:
```
[flash_mcu btt-skr-mini-mz]
flash_method: sdcard,btt-skr-mini-mz
kconfig:
  CONFIG_MACH_STM32=y
  CONFIG_MACH_STM32F103=y
  CONFIG_STM32_FLASH_START_7000=y
  CONFIG_LOW_LEVEL_OPTIONS=y
  CONFIG_INITIAL_PINS="!PA14"
```

The information in the new `[flash_mcu]` config section would be similar to the information already in the comments for each example config file.

In order for this support to be useful it would also need support from moonraker and the web frontends.  The new calls are mentioned briefly in the updated API_Server.md document on this branch.

The current stubbed out code doesn't provide any live feedback to the flashing process - which is not ideal.  It is possible to extend this code to provide feedback via the api server, but that is difficult to do from klippy when flashing (because the klippy mcu connection must not run while flashing).  It may be worth discussing if some of this logic should be in moonraker (or somewhere else).

The current stubbed out code also doesn't support downloading of the final binary for manual flashing (via sdcard or programmer).  That feature should also be discussed.

@Arksine , @meteyou , @cadriel - FYI.

Thoughts?
-Kevin

### DIFF
--- a/config/generic-bigtreetech-skr-mini-mz.cfg
+++ b/config/generic-bigtreetech-skr-mini-mz.cfg
@@ -116,6 +116,15 @@ pin: PC6
 [mcu]
 serial: /dev/serial/by-id/usb-Klipper_Klipper_firmware_12345-if00
 
+[flash_mcu btt-skr-mini-mz]
+flash_method: sdcard,btt-skr-mini-mz
+kconfig:
+  CONFIG_MACH_STM32=y
+  CONFIG_MACH_STM32F103=y
+  CONFIG_STM32_FLASH_START_7000=y
+  CONFIG_LOW_LEVEL_OPTIONS=y
+  CONFIG_INITIAL_PINS="!PA14"
+
 [printer]
 kinematics: cartesian
 max_velocity: 300
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -378,3 +378,25 @@ might return:
 
 As with the "gcode/script" endpoint, this endpoint only completes
 after any pending G-Code commands complete.
+
+### flash_mcu/get_devices
+
+This endpoint provides a list of possible devices when attempting to
+flash a micro-controller. For example:
+`{"id": 123, "method": "flash_mcu/get_devices", "params": {"name":
+"my_board_type"}}`
+might return:
+`{"id": 123, "result": ["/dev/ttyUSB0", "/dev/ttyAMA0", "/dev/ttyACM0"]}`
+
+The "name" parameter must correspond with a [flash_mcu config
+section](Config_Reference.md#flash_mcu).
+
+### flash_mcu/flash
+
+This endpoint requests that the micro-controller code be compiled and
+flashed to a micro-controller. For example:
+`{"id": 123, "method": "flash_mcu/flash", "params": {"name":
+"my_board_type", "device": "/dev/ttyS0"}}`
+
+The "name" parameter must correspond with a [flash_mcu config
+section](Config_Reference.md#flash_mcu).
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -74,6 +74,24 @@ pins such as "extra_mcu:ar9" may then be used elsewhere in the config
 # See the "mcu" section for configuration parameters.
 ```
 
+### [flash_mcu]
+
+Hints for compiling and flashing Klipper's micro-controller code.
+
+```
+[flash_mcu my_board_type]
+kconfig:
+#   List of "kconfig declarations" (one per line) to use when
+#   compiling the micro-controller code. This parameter must be
+#   provided.
+flash_method:
+#   The type of bootloader on the target board. This parameter must be
+#   provided.
+#mcu: mcu
+#   A comma separated list of Klipper micro-controllers names that use
+#   this type of micro-controller. The default is "mcu".
+```
+
 ## Common kinematic settings
 
 ### [printer]
--- a/klippy/extras/flash_mcu.py
+++ b/klippy/extras/flash_mcu.py
@@ -1,0 +1,143 @@
+# Tool for compiling and updating micro-controller code
+#
+# Copyright (C) 2021  Kevin O'Connor <kevin@koconnor.net>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import sys, os, subprocess, logging, errno, tempfile, shutil
+import util, mcu
+
+class PrinterFlashMCU:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.name = config.get_name().split()[-1]
+        mcu_names = [n.strip() for n in config.get('mcu', 'mcu').split(',')]
+        self.mcus = [mcu.get_printer_mcu(self.printer, m) for m in mcu_names]
+        self.kconfig = config.get('kconfig')
+        fc = {c: c for c in self._get_flash_choices(["-T"])}
+        self.flash_method = config.getchoice('flash_method', fc)
+        pf = self.printer.load_object(config, "flash_mcu")
+        pf.register_mcu(self.name, self)
+    def _get_flash_choices(self, cmdargs):
+        srcdir = os.path.dirname(os.path.realpath(__file__))
+        flash_helper = os.path.join(srcdir, '..', '..', 'scripts',
+                                    'flash_helper.py')
+        args = [sys.executable, flash_helper] + cmdargs
+        try:
+            data = subprocess.check_output(args)
+        except:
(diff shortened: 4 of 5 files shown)


meteyou:
woow. looks very good for a start. is this api also available when klipper does not start, because of a too old klipper version on the mcu?

zellneralex:
Be aware many nowadays have more than one mcu. They also can but do not need to differ  differ best example is an mcu board and an linux mcu for the adxl.

-> there is a need to link different mcu to the flash-mcu devices and have multiple of them

altvnk:
Would it make sense to provide flashing method in the mcu definition section?
```
[mcu]
serial: /dev/serial/by-id/usb-Klipper_xxxxxxxxxx
restart_method: command
flash_method: sdcard,btt-skr-mini-mz
kconfig:
  CONFIG_MACH_STM32=y
  CONFIG_MACH_STM32F103=y
  CONFIG_STM32_FLASH_START_7000=y
  CONFIG_LOW_LEVEL_OPTIONS=y
  CONFIG_INITIAL_PINS="!PA14"

[mcu rpi]
serial: /tmp/klipper_host_mcu
flash_method: linux
```

Arksine:
Looks like a great start!   I do have a few thoughts based on some of the things you brought up.

> The current stubbed out code doesn't provide any live feedback to the flashing process - which is not ideal. It is possible to extend this code to provide feedback via the api server, but that is difficult to do from klippy when flashing (because the klippy mcu connection must not run while flashing). It may be worth discussing if some of this logic should be in moonraker (or somewhere else).

We could register a notification similar to the "gcode response" that pushes live feedback over the unix socket.  Moonraker can then forward to all connected websockets.  I believe that this should still be possible from the `disconnect` event handler, as long as we make sure that it fires before the one in the `ServerSocket` class of webhooks.py.  We might need to create a new event that closes the client sockets after `klippy:disconnect` to make this guarantee.

> The current stubbed out code also doesn't support downloading of the final binary for manual flashing (via sdcard or programmer). That feature should also be discussed.

I agree, I think we could move binaries to another folder and rename them to match the name of the [flash_mcu] section.  Moonraker could serve that folder.

One other thing that we should probably consider adding is the ability to do "batch" MCU updates.  We don't want a scenario where there is a protocol change, we flash one MCU then get stuck when Klipper restarts.  

As you mentioned, an alternative would be to have Moonraker run the flash_helper.  Klipper could still handle the build in this scenario.  This would also resolve the feedback issue, as Moonraker has the ability to capture feedback from stdout/stderr and push it to connected clients.

cadriel:
> We could register a notification similar to the "gcode response" that pushes live feedback over the unix socket. Moonraker can then forward to all connected websockets. I believe that this should still be possible from the `disconnect` event handler, as long as we make sure that it fires before the one in the `ServerSocket` class of webhooks.py. We might need to create a new event that closes the client sockets after `klippy:disconnect` to make this guarantee.

The update manager has similar notifications, maybe tacking on to that in some way - since they seem related. Ideally we'd need this kind of feedback.

> I agree, I think we could move binaries to another folder and rename them to match the name of the [flash_mcu] section. Moonraker could serve that folder.

Yup - as a new root, or to an existing one. Be nice for users to have visibility of these in cases where we can't automate.


How are we going to handle notifying the user that an MCU update is required as part of the klipper update? Do we even need to? Do we intend to start the process automatically as a part of the klipper upgrade?

In scenarios where users don't have the correct configuration for an upgrade, can we still notify them of this requirement prior to updating?

KevinOConnor:
> woow. looks very good for a start. is this api also available when klipper does not start, because of a too old klipper version on the mcu?

It should run regardless of connection status.  In particular, I think we want it to work during the initial install and the initial mcu flash (if possible).

> could you also include the /dev/serial/by-id/ devices in the "get_devices" response?

That's the rough idea - the current code is just "stubs".

> Would it make sense to provide flashing method in the mcu definition section?

It's possible to do that, but many users deploy the same board multiple times. In that case, it's easier to use the flash_mcu's `mcu` setting to apply the same settings to multiple boards (see the docs/Config_Reference.md on this branch for details).  It's also a little easier to manage the code with a separate module definition.

-Kevin

KevinOConnor:
> I believe that this should still be possible from the disconnect event handler, as long as we make sure that it fires before the one in the ServerSocket class of webhooks.py.

Yeah - I was thinking something similar.  However, the main challenge is that the reactor is not running during the "klippy:disconnect" handlers.  (This is intentional because we don't want unrelated timers kicking in during the disconnect phase as it could lead to messages being sent while the connection is in a "half way" state.)  Unfortunately, I'm not sure the webhooks code will work reliably with the reactor not running.

> We don't want a scenario where there is a protocol change, we flash one MCU then get stuck when Klipper restarts.

If we just use a basic "webhooks endpoint" then I don't think we should have a problem if klippy can't connect.  I think the only problem would be if we can't read the config file at all, but I don't think that's a huge concern, as the user would need to fix that anyway.

> As you mentioned, an alternative would be to have Moonraker run the flash_helper. Klipper could still handle the build in this scenario.

Yeah - that might be a good split.

-Kevin

KevinOConnor:
> How are we going to handle notifying the user that an MCU update is required as part of the klipper update? Do we even need to? Do we intend to start the process automatically as a part of the klipper upgrade?

It's very hard to predict if a particular code change is going to impact a user or not.  FWIW, my thinking is to try to "keep it simple" - provide a simple web based way for users to update their mcus, and also notify users when a host/mcu mismatch occurs.  Hopefully the majority of users will be able to "connect the dots".

-Kevin

mental405:
I am still on the fence about adding it to the MCU section or having a separate section for compilation settings.

Looking at the stub provided `[flash_mcu btt-skr-mini-mz]` would it make sense to do something like
```
[flash_mcu] <= Default single MCU use cases

[flash_mcu z_board] <= Multi-mcu cases
```

that way the serial ID can be used from the `[mcu]` definition for the correct board type.
I personally have 4 different mcu definitions with 3 different board types on one my printers and this would be highly advantageous for automated updating.

KevinOConnor:
For multi-mcu, the proposal here is to follow the same format as the `board_pins` config section - for example, something like:
```
[flash_mcu some_name]
mcu: mcu, my_z_board
...
```

-Kevin

mental405:
> For multi-mcu, the proposal here is to follow the same format as the `board_pins` config section - for example, something like:
> 
> ```
> [flash_mcu some_name]
> mcu: mcu, my_z_board
> ...
> ```
> 
> -Kevin

So in the case of
`() GPIO pins to set at micro-controller startup`

one could define

` ^mcu:P1.24, !my_z_boardu:P0.25 `

to get the expected result?

KevinOConnor:
The kconfig settings mirror the choices in "make menuconfig".  If a pin needs to be set at startup then one would need to add something like:
```
  CONFIG_LOW_LEVEL_OPTIONS=y
  CONFIG_INITIAL_PINS="!PA14"
```
to the `kconfig` option in the config.

If two micro-controllers have different kconfig settings then two different `[flash_mcu]` config sections would be needed.  If multiple micro-controllers have identical kconfig settings, then only one `[flash_mcu]` config section is needed, and the `mcu` option can list the micro-controllers it applies to.

-Kevin

randellhodges:
Is this something that is still being pursued or has it be tabled? I would love to see this implemented.

DroneMang:
Would it be possible to make this work for CAN bus enabled devices such as HermitCrab?  For the HermitCrab you have to connect regular usb separate 24V power and put it into DFU mode, I wish it were easier to update it.  Thanks!

KevinOConnor:
As per the guidelines in the updated CONTRIBUTING document, I'm no longer going to track this experiment here on the main Klipper3d github repo. I'm moving discussion to https://klipper.discourse.group/t/experiment-with-automated-mcu-flashing/1691

-Kevin

PR #4308, opened 2021-05-20
config: Add pin aliases to generic-creality-v4.2.7.cfg

pelrun:
Signed-off-by: James Churchill <pelrun@gmail.com>

### DIFF
--- a/config/generic-creality-v4.2.7.cfg
+++ b/config/generic-creality-v4.2.7.cfg
@@ -90,3 +90,9 @@ max_velocity: 300
 max_accel: 3000
 max_z_velocity: 5
 max_z_accel: 100
+
+[board_pins]
+aliases:
+  EXP1_1=PC6,EXP1_3=PB10,EXP1_5=PB14,EXP1_7=PB12,EXP1_9=<GND>
+  EXP1_2=PB2,EXP1_4=PB11,EXP1_6=PB13,EXP1_8=PB15,EXP1_10=<5V>
+  PROBE_IN=PB0,PROBE_OUT=PB1,FIL_RUNOUT=PC6


KevinOConnor:
Thanks.  How did you obtain the pin mappings for this board?  The aliases don't look correct to me - there should be EXP1_x pins defined and they should have similar reserved pins as the definition in config/generic-bigtreetech-skr-mini-e3-v2.0.cfg

-Kevin

pelrun:
Beeped them out explicitly with a multimeter on my board, and tested with the ender 3 lcd. Strictly speaking the pin naming is screwed up already, as the unnamed header on the mainboard is connected to exp3 on the lcd,  and exp1 is completely unused.  But if the board *was* connected to exp1 via a standard cable, the definitions are right.

pelrun:
Note that the definitions are correct for the header footprint, but the definitions are in numerical order, not footprint order.

KevinOConnor:
Thanks.  I think it would help if the pins were listed out in the same order as generic-bigtreetech-skr-mini-e3-v2.0.cfg and if the reserved pins are also explicitly listed.

Also, FWIW, I don't understand the comment added to sample-lcd.cfg .  If a change is needed there I think it would be best to do that separately from the change to this config.

Cheers,
-Kevin

pelrun:
The problem is the "EXPn" naming scheme is only applicable for the physical connectors on certain lcds, not the signals on the cable or the connectors on the mainboards. But the configs are all using them as a global label instead.

So you get misleading situations like with the Ender 3 LCD, which has three headers EXP1/2/3, and the definition in sample-lcd.cfg has the pin layout for EXP3 but using "EXP1_x" labels. Someone trying to use it without already understanding the problem is going to think that they should plug into EXP1 on that LCD, and it's not going to work. There is no way to rewrite the definition to work with the actual EXP1, because not all the signals are present on it.

Short of fixing the naming problem, which would require changing all "EXPn_x" aliases across the board to something completely independent of any board/lcd connector name, a comment in the definition to explicitly tell the user how to connect it up is the simpler fix.

KevinOConnor:
The purpose of the EXP1_x aliases is to describe the physical layout of the pins on the micro-controller board.  It is not intended to describe the pins on the lcd.  When the sample-lcd.cfg file has a line with `sclk_pin: EXP1_6` it is saying that the LCD's "sclk" wire is routed to the "6th pin" on the micro-controller board's EXP1 connector.  It makes no comment on where the "sclk" wire is on the lcd plugs.

I think I now understand the comment you added to the sample-lcd.cfg file.  I think maybe something like the following could be added:
```
# This section is used for a Creality "12864" display with a single
# ribbon cable between the display's EXP3 plug and the
# micro-controller board's EXP1 connector.
```

Again, though, I would do that separate from this config change.

Cheers,
-Kevin

KevinOConnor:
Okay, is there interest in providing an EXP1 definition for this board?

-Kevin

pelrun:
It's not really useful; only EXP3 has all the signals.

I've updated my commit now to reorder the pin definition as per the other files.

KevinOConnor:
The change looks fine to me, but I'm confused at the state of this PR.  Is this change useful (and I should commit it) or is this change not useful (and I should not commit it)?

-Kevin

pelrun:
Absolutely it's worth committing! Any discussion about naming problems is completely orthogonal (and much more difficult to resolve satisfactorily) to actually having a working configuration, and I'm actively using this as it stands.

KevinOConnor:

-Kevin

PR #4312, opened 2021-05-21
gcode_macro: Disallow whitespace in macro name

FHeilmann:
This change disallows additional spaces in `gcode_macro` names.

Specifying a gcode macro block like `[gcode_macro MACRO NAME]` does not cause errors or warnings, until the user attempts to use the macro, (at which point the macro will be available under `MACRO` instead). This change simply checks whether the section header contains additional spaces, and raises an error if it does.

Signed-off-by: Florian Heilmann <Florian.Heilmann@gmx.net

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -115,7 +115,7 @@ class GCodeMacro:
     def __init__(self, config):
         if len(config.get_name().split()) > 2:
             raise config.error(
-                    "Name of section '%s' contains illegal whitespace" 
+                    "Name of section '%s' contains illegal whitespace"
                     % (config.get_name()))
         name = config.get_name().split()[1]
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.

-Kevin

swalsh11:
Hi, This breaks this :  'gcode_macro START_PRINT T_BED T_EXTRUDER'   where arguments on temperatures were passed to the printer, How do I update my Printer. CFGG

FHeilmann:
Parameters were never meant to be defined this way. Basically the thing that now triggers the error was ignored by klipper before. You can just remove them and your macro should function as intended. See this document in case you want to set defaults for your parameters.

swalsh11:
thanks @FHeilmann  that worked,

PR #4313, opened 2021-05-21
qgl/z-tilt: expose an 'applied' status to allow macros to skip QGL/Z_TILT_ADJUST

Xorlev:
This is useful for macros that'd like to skip QUAD_GANTRY_LEVEL/Z_TILT_ADJUST if already leveled,
e.x.:

```
    {% if not printer.quad_gantry_level.applied %}
        QUAD_GANTRY_LEVEL
    {% endif %}
```

```
    {% if not printer z_tilt.applied %}
        Z_TILT_ADJUST
    {% endif %}
```

Signed-off-by: Michael Rose <elementation@gmail.com>

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -217,6 +217,13 @@ is defined):
   template expansion, the PROBE (or similar) command must be run prior
   to the macro containing this reference.
 
+# quad_gantry_level
+
+The following information is available in the `quad_gantry_level` object
+(this object is available if quad_gantry_level is defined):
+- `leveled`: True if the gantry leveling process has been run and completed
+  successfully.
+
 # query_endstops
 
 The following information is available in the `query_endstops` object
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -221,7 +221,7 @@ is defined):
 
 The following information is available in the `quad_gantry_level` object
 (this object is available if quad_gantry_level is defined):
-- `leveled`: True if the gantry leveling process has been run and completed
+- `applied`: True if the gantry leveling process has been run and completed
   successfully.
 
 # query_endstops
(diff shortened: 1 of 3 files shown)


KiloQubit:
I think this would be very useful.  It's already possible to check homing status with `printer.toolhead.homed_axes`, it would be nice to check the QGL status as well.  I think something similar should also exist for `Z_TILT_ADJUST` status.

joshmurrah:
I would love to see this functionality to be able to skip un-needed GQL.  I have one question, would this variable transition to false when idle timeout is reached or steppers are disabled?

Xorlev:
> I have one question, would this variable transition to false when idle timeout is reached or steppers are disabled?

It does. It listens to the `stepper_enable:motor_off` event (issued when the motors are entirely disabled with M18/M84 or via the default idle_timeout), and resets the leveled status to false.

zellneralex:
@Xorlev you should add a description to https://github.com/KevinOConnor/klipper/blob/master/docs/Status_Reference.md

Xorlev:
Good call, done.

KevinOConnor:
Thanks.  In general it seems useful functionality.  I have a few high-level comments:
1. I'm not sure about the name "leveled".  How about "active" or "applied" as an alternative?
2. I think it may be preferable to move this logic into ztilt.py:ZAdjustHelper - that way both z_tilt and quad_gantry_level can share the code (the qgl get_status() method can just call self.z_helper.get_status() ).  See #4326 .

Cheers,
-Kevin

vladbabii:
If we're doing this on a more generic level, could we also save the deviation value in a variable?

Xorlev:
> I'm not sure about the name "leveled". How about "active" or "applied" as an alternative?

`applied` seems fine, and generic across QGL and Z_TILT_ADJUST.

> I think it may be preferable to move this logic into ztilt.py:ZAdjustHelper - that way both z_tilt and quad_gantry_level can share the code (the qgl get_status() method can just call self.z_helper.get_status() ). See #4326 .

I made a separate status object (`ZAdjustStatus`) used by QuadGantryLevel and ZTilt. I don't think it belongs in ZAdjustHelper: it doesn't have any information about the status of retries performed by the RetryHelper in each of the leveling processes. The status object centralizes _some_ of the logic with minimal hooks by QGL and ZTilt to make it work.

PTAL.

KevinOConnor:
Thanks.

-Kevin

PR #4315, opened 2021-05-22
Added second CNC fan to generic-fysetc-cheetah-v1.2.cfg

jericho-adler:
Added second CNC fan, PB0

Signed-off-by: Jericho Adler jericho.adler@gmail.com


KevinOConnor:
Thanks.

-Kevin

PR #4317, opened 2021-05-22
gcode_macro: add description property

zellneralex:
This add a description property to gcode_macro in the same way as used for klipper internal macros. That should help to identify the function of a own macro while using help or the autocomplete function in an console

### DIFF
--- a/docs/Command_Templates.md
+++ b/docs/Command_Templates.md
@@ -29,8 +29,8 @@ beginning.
 
 ### Add a description to your macro
 
-The help indetify the functionality a short description can be added. To
-do that add `description:` with a short text.
+To help indetify the functionality a short description can be added.
+Add `description:` with a short text to describe the functionality.
 Default is "G-Code macro" if not specified.
 For example:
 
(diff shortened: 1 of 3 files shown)


docs/Command_Templates.md
```diff
@@ -27,6 +27,25 @@ Note how the `gcode:` config option always starts at the beginning of
 the line and subsequent lines in the G-Code macro never start at the
 beginning.
 
+### Add a description to your macro
+
+The help indetify the functionality a short description can be added. To
```
KevinOConnor, line 32: Typo.

Otherwise the change is fine with me.  The `description` field will also need to be added to docs/Config_Reference.md .

-Kevin

zellneralex: I hope now everything is ok

KevinOConnor:
There is still "indetify" instead of "identify" in the added docs.

-Kevin

zellneralex:
Sorry overseen that. Done

KevinOConnor:
Thanks.

-Kevin

PR #3462, opened 2020-10-23
module: DS18B20, new module for 1-wire temperature sensor (WIP)

theopensourcerer:
Initial commit of code to support 1-wire (Dallas) sensors such
as the DS18B20. requires Linux kernel drivers to create a file
in /sysfs which is read by this module, and temperature
typically returned to a temperature_fan.

This code purely for testing/revew. No documentation as yet.

Signed-off-by: Alan Lord <alanslists@gmail.com>

PR #4320, opened 2021-05-23
Add DHT22 Sensor

TheJoshW:
DHT22 sensor provides temperature and humidity readings at ~2 sec polling interval.
RPi allows this sensor to be enabled as a userspace device using a device tree overlay per https://www.raspberrypi.org/forums/viewtopic.php?t=98543.
Add the overlay to the `/boot/config.txt` file:
```
dtoverlay=dht11,gpiopin=#
```
Where `#` is the RPi GPIO pin the DHT22 signal pin is connected to.

This allows the temperature to be read from file: `/sys/bus/iio/devices/iio:device<#>/in_temp_input`
And humidity to be read from file: : `/sys/bus/iio/devices/iio:device<#>/in_humidityrelative_input`
Where `#` is the device number (typically 0).

In Klipper, `temperature_host` allows the addition of a file based temperature sensor, however this is polled every second which is too fast and is not tolerant of the errors that may occur. Also it does not support humidity reading.

This PR proposes the addition of a DHT22 sensor plugin that polls at every 20 seconds and retries 9 times on error (2 sec between retries) and supports humidity.

Signed-off-by: Joshua Wherrett thejoshw.code@gmail.com


KevinOConnor:
Thanks.  However, the convention is to read sensor input from the "mcu" code and not directly from the host hardware.  There was a similar discussion on this at PR #3462 .  (And it may be possible to extend te ds18b20 code for the dh22 functionality.)

This distinction is also important because there are reports that the dht22 kernel code blocks the calling process during reads.  The host code can't tolerate blocking reads - it could cause subtle failures.

-Kevin

TheJoshW:
No worries. Will refactor as an MCU sensor.

KevinOConnor:

-Kevin

TheJoshW:
@KevinOConnor Sorry it's waiting on me to find time to turn it into an MCs sensor. Not a complex task. I just need to find time to do it.

KevinOConnor:
Okay - I will leave the "pending feedback" label on this then.

-Kevin

PR #4325, opened 2021-05-28
docs: Updated adxl345 wiring image

dmbutyugin:
ADXL345 3.3v and GND pins can be wired closer to the other
pins on RPi, which is often more convenient.


KevinOConnor:
Thanks!

-Kevin

PR #4326, opened 2021-05-28
Z_TILT_ADJUST status for macros

vladbabii:
This is useful for macros that'd like to skip Z_TILT_ADJUST if already
leveled, e.g.:

```
    {% if not printer.z_tilt.leveled %}
        Z_TILT_ADJUST
    {% endif %}
```

Signed-off-by:  Vlad Babii <vlad@babii.ro>


KevinOConnor:
I think this should be addressed with the merge of #4313 .

-Kevin

vladbabii:
I'll close it. Many thanks for getting this feature in klipper.

PR #4328, opened 2021-05-29
annexed_probe: Support for probes that don't live on the toolhead

mental405:
There are many new probe designs utilizing microswitches that attach to the 
toolhead via magnets or some other mechanical coupling. When they are not in 
use, these probes are stowed in a dock located somewhere in the printer. Due 
to the ever growing complexity required for homing macros, it was determined 
that these types of probes would be ultimately better suited to a module. 

I have attempted to handle as many different scenarios as I have seen, but 
people still continue to surprise me with what they have created. While this
module won't handle _all_ configurations, it should handle _most_ 
configurations with a large degree of flexibility and customization.


MarcPot:
Awesome work Mental, with the probe, and with cleaning up my config from the macro’s. Thank you 😊

mental405:
If anyone is going to try this branch, I suggest reading the documentation and making sure you understand it as well as keeping a hand on the estop when first trying things.

Sugaroverdose:
First off: awesome thing

But it's not clear from doc what state of pins(probe_sense_pin and dock_sense_pin) is expected, for exception of check_open_attach - it's self describing

i've tried to enable manual_verify but detach_probe and attach_probe commands always raises errors twice, for some reason, also i've tried to combine check_open_attach and dock_sense_pin, but it always fails on detach_probe, not sure is it hardware failure or this options are mutually exclusive?

mental405:
> But it's not clear from doc what state of pins(probe_sense_pin and dock_sense_pin) is expected, for exception of check_open_attach - it's self describing

I think I can probably clarify that a bit in the docs. Basically you want the pins to be configured in printer.cfgso they report the correct states when GET_PROBE_STATUS is called. 

> I've tried to enable manual_verify but detach_probe and attach_probe commands always raises errors twice, for some reason

I will look at how those exceptions are being raised... what frontend are you using?

> UPD: called detach_probe when toolhead was above probe dock and script didn't move toolhead out, instead it decided to drop toolhead right on it

If the tool is within the minimum safe distance it will not move away from the dock as that could potentially yank the probe out of the dock. Instead it calculates an intercept vector perpendicular to the approach angle and moves along that and then moves to the dock coordinates. If the tool is already over top of the dock coordinates, the calculations for that probably net to zero tool movement.

Sugaroverdose:
> I will look at how those exceptions are being raised... what frontend are you using?
>
>>
i'm running fluidd
> I think I can probably clarify that a bit in the docs. Basically you want the pins to be configured in printer.cfgso they report the correct states when GET_PROBE_STATUS is called.
>
>>
at some point i've inverted dock and it's started to work properly, so when probe is docked expected state is TRIGGERED - that kind of counter-intuitive, at least for me

> If the tool is within the minimum safe distance it will not move away from the dock as that could potentially yank the probe out of the dock. Instead it calculates an intercept vector perpendicular to the approach angle and moves along that and then moves to the dock coordinates. If the tool is already over top of the dock coordinates, the calculations for that probably net to zero tool movement.
>
>>
Well if i'd call attach_probe this sounds reasonable, but i've called detach_probe and as a result - dock has been destroyed :)

mental405:
i'm running fluidd

I have noticed that fluidd duplicates error messages being raised in certain cases.

```
22:30:50
// Klipper state: Ready
22:30:51
!! Error on 'SET_LED LED=ring RED=0 GREEN=1 BLUE=0 INDEX=0 TRANSMIT=1': INDEX must have minimum of 1
22:30:51
!! Error on 'SET_LED LED=ring RED=0 GREEN=1 BLUE=0 INDEX=0 TRANSMIT=1': INDEX must have minimum of 1
```
 I don't know for sure the cause of it, but the exceptions I am raising in the probe module are `printer.command_error()` messages which should halt execution of the current running command.

---
> at some point i've inverted dock and it's started to work properly, so when probe is docked expected state is TRIGGERED - that's kind of counter-intuitive, at least for me

Internally, I am treating the `probe_sense_pin` and `dock_sense_pin` like endstops.  It is exceedingly difficult to make a state handler with 3 separate pins that is reliable, which is why only `check_open_attach` _or_ `probe_sense_pin` should be used. Additionally, the `probe_sense_pin` should, in theory, be more reliable than `check_open_attach`. 

If _only_ `check_open_attach`  is in use, the checks follow this logic

```
if  open_attach = OPEN   then probe_state = ATTACHED
if  open_attach = TRIGGERED then probe_state = DOCKED
```
When `probe_sense_pin` is supplied, it follows the same

```
if  probe_sense = TRUE  then probe_state = ATTACHED
if  probe_sense = FALSE then probe_state = DOCKED
```
when `dock_sense_pin` is used in conjunction with either `check_open_attach` or `probe_sense_pin`

```
if  probe_sense = TRUE  and dock_sense = FALSE then probe_state = ATTACHED
if  probe_sense = FALSE and dock_sense = TRUE  then probe_state = DOCKED
if  probe_sense = FALSE and dock_sense = FALSE then probe_state = UNKNOWN
if  probe_sense = TRUE  and dock_sense = TRUE  then probe_state = DOCKED
```

The TRUE/FALSE part is where the confusion is probably occurring. The pin definition determines how TRUE/FALSE/HIGH/LOW is being reported from the sense pins. So if the pin is defined with `!` it will be expecting the "switch" to be in a LOW state in order to report true. So for instance, if you are using the annex probe with the screws through the switch body and a bit of wire connecting them... Assuming that one magnet goes directly to an IO pin and the other magnet is wired into a GND pin... The definition for the pin should be something like `^!P1.25`. 

Therefor when the probe is _not_ in the dock, the pin reads high, when the pin _is_ in the dock, the pin reads LOW. I probably said that backwards somewhere and my answer is totally wrong because I get confused by this stuff too most of the time.

The reason I included the `GET_PROBE_STATUS` gcode command was so that this could be more easily debugged when configuring it. 

**Do you think it would be helpful to include additional information in the output?**

Presently it outputs
```
Probe Status: UNKNOWN|ATTACHED|DOCKED
```
I could have it include (if the pins are configured
```
check_open_attach: ATTACHED|UNATTACHED
probe_sense_pin: ATTACHED|UNATTACHED
dock_sens_pin: DOCKED|UNDOCKED
```
---

> Well if i'd call attach_probe this sounds reasonable, but i've called detach_probe and as a result - dock has been destroyed :)
Could you provide your config file so I can understand what is happening here? I am assuming the dock is mounted to one of the bed support extrusions on the 2.4 in the rear, correct?

If you are using a bed mounted dock, the Z axis should always be homed before attempting to do anything with the probe and the module should be throwing an error if it isn't. It is possible it is trying to Z hop or something because the probe is registering that it is `ATTACHED` since it is sitting over it in the dock.

I am thinking I need to try to replicate this behavior on my own and add an additional boundary check so that it will not move the Z axis unless it is outside of the minimum safe distance in a sphere around the dock instead of just a radius.

Sugaroverdose:
> If you are using a bed mounted dock, the Z axis should always be homed before attempting to do anything with the probe and the module should be throwing an error if it isn't. It is possible it is trying to Z hop or something because the probe is registering that it is `ATTACHED` since it is sitting over it in the dock.
>>
let me clarify:
1. Everything homed, qgl'ed, etc
2. Probe dock is mounted near bed(you've got it right)
3. Toolhead is right above dock, but not touching it
4. Probe is mounted on toolhead and probe status is 'attached'
5. After calling detach_probe toolhead does not bother to move out of dock safe distance and goes down right on it, only by doing z movement, which leads to collision

> The TRUE/FALSE part is where the confusion is probably occurring. The pin definition determines how TRUE/FALSE/HIGH/LOW is being reported from the sense pins. So if the pin is defined with `!` it will be expecting the "switch" to be in a LOW state in order to report true. So for instance, if you are using the annex probe with the screws through the switch body and a bit of wire connecting them... Assuming that one magnet goes directly to an IO pin and the other magnet is wired into a GND pin... The definition for the pin should be something like `^!P1.25`.
>>
As i said, i've got it after some time :) Yet it would be good to have this clarification in docs, something like: if your dock pin is closed when probe is docked, than you need to invert pin 

> The reason I included the `GET_PROBE_STATUS` gcode command was so that this could be more easily debugged when configuring it.
>>
Unfortunately get_probe_status does not give clear visibility what exactly happens to pins, with clarify of pins states in doc it should not be a problem ofc w/o it it's kinda not clear does it hardware failure or wrong configuration(i could move probe by hands so it's become 'detected' with wrongly configured pin, but i didn't actually seen is it shorted dock pins or not)


As additional suggestion: it would be good to have an option to disable 'return to previous position' after attach/detach action(at least as command argument) since in a lot of cases it's just makes unnecessary movements

mental405:
Ok. I think it would be overall beneficial for me to add individual pin states in for the `GET_PROBE_STATUS` command for debugging purposes. I will look into the positioning after detach is called. It might need some adjusting.

KevinOConnor:
Thanks.  This does seem like useful functionality.  Alas, I haven't gotten a chance to fully review it and I think it will be weeks before I can review it.

Some comments from my limited review:
1. I can't tell if "annexed" is meant to be related to the physical hardware or is a reference to the 3d printer community.  In either case, I think a different name would be preferable.  Perhaps something like "probe_dock".
2. As high-level feedback, I fear this module/documentation may be too "generic" - in that it's customizability may limit its utility for real-world users.  Specifically, I fear the configuration complexity may be too great for users to configure it optimally.  If possible, I think selecting one or two common dock designs and coding/documenting for them may help.  Also, if there are config options that are "optimization" it may help to eliminate them for simpler use.  (For example, is it necessary to define a dock_angle and dock_safe_distance as opposed to defining a rendezvous coordinate that the toolhead always moves to on dock and undock?)
3. I'd be leery of having the probe "auto attach".  Could we just require the user run ATTACH_PROBE prior to any probing attempts, and require them to run DETACH_PROBE when complete?  (And, raise an error if the user issues any "probe" command when the probe is not actually attached.)

Thanks again,
-Kevin

mental405:
>Thanks. This does seem like useful functionality. Alas, I haven't gotten a chance to fully review it and I think it will be weeks before I can review it.

It is not at all ready for merge so no rush. I opened the PR primarily so people could test it and to elicit feedback from the community.

---

>I can't tell if "annexed" is meant to be related to the physical hardware or is a reference to the 3d printer community. In either case, I think a different name would be preferable. Perhaps something like "probe_dock".

It's a bit of both actually. The dictionary definition for annexed is 
`attached or added, especially to something larger or more important`
I will be more than happy to rename it.

---

>As high-level feedback, I fear this module/documentation may be too "generic" - in that it's customizability may limit its utility for real-world users. Specifically, I fear the configuration complexity may be too great for users to configure it optimally.

The "default" configuration requires the following values be set

```
pin:
z_offset:
sample_retract_dist:
dock_position:
dock_angle:
detach_angle:
dock_safe_distance:
check_open_attach:
dock_fixed_z:
```

`dock_position`, `dock_angle`, `detach_angle`, and `dock_safe_distance` must be supplied. `detach_angle` could be inferred using a perpendicular vector to the `dock_angle`  by default, but it would need to be able to be specified in the even that it guesses incorrectly. Additionally a `dock_safe_distance` of 50mm could be defaulted as well. These vectors could also easily be obtained via coordinates instead of angles. So a user could supply `departure_position`, `dock_position`, and `approach_position` instead. 

The two distinct positions for approach and departure _must_ be used in most cases so a single rendezvous point would not be ideal. In most cases lifting the toolhead away from the probe can, over time, pull the magnets out of their pockets and this results in probe accuracy degrading. To properly remove the probe from the toolhead, the probe is "slid" off the magnets in a side to side motion. If a single rendezvous point is used, the probe would never actually detach.

`sample_retract_dist` must be supplied as well due to the switch actuation and attachment checks. During _most_ probing routines the calling module will retract the toolhead after the probe is triggered. Certain modules leave the toolhead down after the probing move and the switch remains triggered. Due to how many of these switches are wired, this falsely reports that the probe is detached when it is not. I have added a check after probing moves that ensures the toolhead is at or above the `sample_retract_dist` so that this does not occur. It could be defaulted to 3mm since most probes use a variation of the omron switch with around a 2mm actuation distance.

It would also be possible to remove `check_open_attach` as a config option and leave it on by default. In the event someone defines `probe_sense_pin` it is already switching detection to use the pin instead. 

---

>If possible, I think selecting one or two common dock designs and coding/documenting for them may help. Also, if there are config options that are "optimization" it may help to eliminate them for simpler use. (For example, is it necessary to define a dock_angle and dock_safe_distance as opposed to defining a rendezvous coordinate that the toolhead always moves to on dock and undock?)

I can look into further scaling back the supported conditions. I initially did a survey of users to find out what and how they were using the probe. 68% of respondents said their dock was mounted to a moving gantry or to the frame with the remainder having bed mounted docks. I felt that limiting it only to moving gantries, however, limited it to "cube" printers so I included bed mounted docks as well. 

By my count there are 3 popular designs floating around, the Annex Quickdraw, the Voron Users Klicky-Probe, and the Euclid. These all use a variation of the same basic dock and probe design using magnets and a "fork", so the built in default toolhead movements were intended to handle those. The biggest difference between the 3 is the length of the forks on the dock which translates to the `dock_safe_distance` parameter.

The mounting locations of these varies and I have limited support for them to _positive_ Z positions. It is not feasible to "dip" the toolhead into negative space to pick up the probe. 

Outside of the dock, the only other common condition that affects operation is whether or not the probe is being used as an endstop or if a physical endstop is in use.

So without doing any custom gcode options  the "default" common dock configurations (`dock_fixed_z`) :
- Gantry/Frame mounted
- Z Positive Bed mounted with separate Z endstop

The 6 optional gcode sections that can be supplied exist for users with edge cases. There are a few printers with odd pathing issues that cannot be solved with straight lines so they need to be moved to a position prior to moving to the approach so fan ducts don't get ripped off etc... The other use case is to have the dock on a servo that flips it into position. I have heard that there are people wanting to build this type of dock for Switchwire printers but I have not seen it yet. I figured I would go ahead and add _something_ so they can deploy the servo at the appropriate time. They are in there as config options but should not cause. 

I use these sections to change the color of the LEDs on the toolhead to reflect the current probe action.



---

>I'd be leery of having the probe "auto attach". Could we just require the user run ATTACH_PROBE prior to any probing attempts, and require them to run DETACH_PROBE when complete? (And, raise an error if the user issues any "probe" command when the probe is not actually attached.)

As for the auto attach part. The goal was to have it behave like any other probe in that you command 'probe' from the gcode console and a complete probing action occurs without having to write macros around it. I have not noticed any adverse behavior with the automatic attach part, only the auto detach part. The adverse behavior reported so far is typically due to false positives with probe detection or user configuration. Even if it were made to only attach and detach when commanded, I believe it would still exhibit the same behavior since it would read the probe state incorrectly.  

In the case of probe calibration for instance, I believe it would cause out if DETATCH_PROBE was called from the terminal after the initial point was probed. At the present time, when `PROBE_CALIBRATE` is called, the probe is picked up, the first point is probed, the probe is docked, and the toolhead returns to the calibration point for the nozzle.

I am likely removing the `delayed_detach` entirely as it only seems to cause problems if `DETACH_PROBE` is not explicitly called at the end of a "start print" gcode sequence.

I also waffled around for days about whether or not I wanted to support scenarios where there was no internal means of probe detection and the user supplies the probe state via the gcode command. I think this is just going to cause more trouble that it is worth as well so it is probably going away as well. I think in order to use the module, there _must_ be some way configured to determine the probe state.

KevinOConnor:
> The two distinct positions for approach and departure must be used in most cases so a single rendezvous point would not be ideal.

Okay, that makes sense.

> 68% of respondents said their dock was mounted to a moving gantry or to the frame with the remainder having bed mounted docks. I felt that limiting it only to moving gantries, however, limited it to "cube" printers so I included bed mounted docks as well.

Okay, FWIW, I think it may be easier to target the main use case first, get the code merged, get lots of feedback, and then look to extend to other probes (if necessary).

> The biggest difference between the 3 is the length of the forks on the dock which translates to the dock_safe_distance parameter.

I'm a little confused on what `dock_safe_distance` is used for.  Wouldn't we always want to dock from the same `dock_rendezvous_coordinate` and undock to the same `undock_rendezvous_coordinate`?

> The goal was to have it behave like any other probe in that you command 'probe' from the gcode console and a complete probing action occurs without having to write macros around it.

So, as high-level feedback, we tried to do something similar in Klipper with the toolchange commands (eg, `T1`).  That is, the code tried to do the "right thing" and there was config customizability to handle "corner cases".  It didn't work well, as the config complexity to "do the right thing" was far greater than the config complexity of a macro that explicitly states the steps it takes.

FWIW, I expect users of a "docked probes" to introduce macros for their calibration routines (eg, `G29`) and I expect that they will do that regardless of how "smart" the dock code is.  It shouldn't be difficult to add an explicit `ATTACH_PROBE` command to these macros, and I suspect being explicit will actually make the macro more clear.  Having some of the actions in a `G29` macro, some in the code, some in `[annexed_probe]` config section, etc. I fear will be harder to understand.

> I also waffled around for days about whether or not I wanted to support scenarios where there was no internal means of probe detection and the user supplies the probe state via the gcode command. I think this is just going to cause more trouble that it is worth as well so it is probably going away as well. I think in order to use the module, there must be some way configured to determine the probe state.

That makes sense to me.  Again, it's always possible to add additional features at a later time.

Thanks,
-Kevin

Sugaroverdose:
> FWIW, I expect users of a "docked probes" to introduce macros for their calibration routines (eg, `G29`) and I expect that they will do that regardless of how "smart" the dock code is. It shouldn't be difficult to add an explicit `ATTACH_PROBE` command to these macros, and I suspect being explicit will actually make the macro more clear. Having some of the actions in a `G29` macro, some in the code, some in `[annexed_probe]` config section, etc. I fear will be harder to understand.

>>
Well, if code is smart enough it will dock probe whatever happens, at the same time if user does not have proper error handling in gcode(let’s be honest-it’s quite complex question) than he will face issues with probe attached in situations where it not expected to be on toolhead

mental405:
>Okay, FWIW, I think it may be easier to target the main use case first, get the code merged, get lots of feedback, and then look to extend to other probes (if necessary).


image

The overall design is centered around the magnetic probes so there isn't much to strip out there. Really the only things to restrict are 

- only supporting docks attached to the frame/gantry 
- only allowing the probe to be used as the Z endstop. 

If think restricting the dock location is fine, but I feel like there is going to be less adoption with restricting the endstop usage since a lot of Voron people feel they need the nozzle endstop to calibrate offsets.

---

>I'm a little confused on what dock_safe_distance is used for. Wouldn't we always want to dock from the same dock_rendezvous_coordinate and undock to the same undock_rendezvous_coordinate?

For all intents and purposes ATTACH_PROBE and DETACH_PROBE as gcode commands map directly to the attach_probe and detach_probe methods in the module. They perform the same position checks in order to do make sure the tool is properly aligned with the dock.

When either of those commands are called, there are some checks performed. 

The first check is to see whether or not the probe is already attached to the toohead. In the case of PROBE_ATTACH, it simply stops executing as it is already attached

The second check performed is to align the toolhead to the dock. The `minimum_safe_distance` parameter is supplied to make sure that the toolhead does not move in a direction that might attract the magnets prematurely. 

<p align="center"></p>

The distance from the toolhead to the `dock_position` is checked, if the toolhead is outside of that radius, the tool moves in a straight line to the approach coordinate. 

<p align="center"></p>

From a user standpoint, the angles aren't as important as setting the correct `minimum_safe_distance`. If the config settings are changed to coordinates, I would plan on using the angles internally because of the next bit. If for some reason the toolhead is close to the `dock_position` and ATTACH_PROBE is called, the tool would move in a straight line to the approach coordinates. If that straight line is misaligned with the dock, there is a possibility that the probe will jump out of the dock and not attach correctly. 

<p align="center"></p>

In an attempt to mitigate that, if the toolhead is within the `minimum_safe_distance`, a coordina is calculated that moves the toolhead perpendicular to the approach vector. This is also done to minimize toolhead movement

<p align="center">

</p>

After the toolhead is aligned to the correct approach vector, it then moves to the `dock_position` in a straight line  where the magnets snap to the mount. The tool then moves in a straight line back to the approach coordinate in the case of probe attach or it moves to the detach coordinate in the case of probe detach.

As I am typing this, I realize I should probably change the behavior for detach since it could potentially knock the probe off with the dock.

---


The angles vs coordinates thing seems to be confusing people. Initially when all of the probe handling was done in macros, it was all coordinate based.  There were certain docks being developed that were positioned 45 degrees

mental405:
> > FWIW, I expect users of a "docked probes" to introduce macros for their calibration routines (eg, `G29`) and I expect that they will do that regardless of how "smart" the dock code is. It shouldn't be difficult to add an explicit `ATTACH_PROBE` command to these macros, and I suspect being explicit will actually make the macro more clear. Having some of the actions in a `G29` macro, some in the code, some in `[annexed_probe]` config section, etc. I fear will be harder to understand.
> 
> > >
> 
> Well, if code is smart enough it will dock probe whatever happens, at the same time if user does not have proper error handling in gcode(let’s be honest-it’s quite complex question) than he will face issues with probe attached in situations where it not expected to be on toolhead

There are issues either way you look at it.

On one hand, automatic attaching/detaching means you don't have to remember to do it and can simplify macro writing a bit.
On the other hand there are a lot of scenarios that must be accounted for.

With manually doing it, you can make sure the toolhead is exactly where you want it to be when the macro executes, but any other time you want to command the probe, you have to remember to attach/detach it. Remembering to detach is probably more important that attaching.

A consideration with this type of probe that most people don't see right away is that the probe is below the nozzle. So if the probe is attached, any movement of the toolhead between Z0 and Z+Probe Offset will result in the probe getting buried in the bed. This is an easy thing do in macros if you aren't careful.

KevinOConnor:
Thanks.  Your description and pictures are very helpful.

> The overall code in the module is centered around theses magnetic probes so there isn't much to strip out there.

Okay.

> From a user standpoint, the angles vs coordinates isn't as important as setting the correct minimum_safe_distance.

I understand.  Just out of curiosity, if the config defined a `dock_rendezvous_coordinate` could the code infer the minimum_safe_distance from the distance between the `dock_position` and `dock_rendezvous_coordinate`?

> In an attempt to mitigate that, if the toolhead is within the minimum_safe_distance, a coordina is calculated that moves the toolhead perpendicular to the approach vector.

Okay.  If you want my $0.02, as a user, I'd prefer to get an error message than have the code "try to do the right thing".  I think it would be easier for me, as a user, to arrange a consistent toolhead position prior to docking than try to understand the special cases that occur otherwise.

> On one hand, automatic attaching/detaching means you don't have to remember to do it and can simplify macro writing a bit.
On the other hand there are a lot of scenarios that must be accounted for to make sure it does what it is supposed to do, when it is supposed to.

Indeed.  FWIW, as a user, I'd rather setup repeatable routines and get an error message when something unusual happens.

> Remembering to detach is probably more important than remembering to attach it.

FWIW, it should be possible for ATTACH_PROBE to register a "gcode move transform" that raises an error on Z movement.  (The transform doesn't apply to the probing code, which directly calls `toolhead.move()`.)

Cheers,
-Kevin

mental405:
I have changed the toolhead positioning in the configuration to discrete coordinates instead of angles. This may prove to be easier for users when configuring the module.

Additional feedback I have been receiving centers around the `dock_fixed_z` option. Admittedly it is a bit confusing and I am open to any suggestions regarding the naming of this parameter or anything that can be reworded in the documentation so that it is easier to understand.

I have also removed the ability to manually set the probe state via gcode command, this was just a bad idea to begin with.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+
+        return angle, magnitude
+
+    # Align z axis to prevent crashes
+    def _align_z(self):
+        if not self._last_homed:
+            curtime = self.printer.get_reactor().monotonic()
+            homed_axes = self.toolhead.get_status(curtime)['homed_axes']
+            self._last_homed = homed_axes
+        if self.dock_fixed_z:
```
dalegaard, line 374: This `if` statement is a bit unwieldy, could maybe help to split as two functions e.g. `_align_z_fixed` and `align_z_moving` or similar which get called based on `self.dock_fixed_z`?

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+                                        [None, None, self.dock_position[2]],
+                                        self.lift_speed)
+            else:
+                raise self.printer.command_error(
+                    "Cannot attach/detach probe, must home Z axis first")
+
+    # Hop z and return to un-homed state
+    def _force_z_hop(self):
+        this_z = self.toolhead.get_position()[2]
+        if self.last_z != this_z:
```
dalegaard, line 399: Could this be rewritten as early return?

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+                        if i in homing_state.get_axes()]
+        if self in [es for rail in rails for es, name in rail.get_endstops()]:
+            self.is_z_endstop = True
+
+        # Get homed axes now as they will report they are
+        # homed later even if they arent
+        curtime = self.printer.get_reactor().monotonic()
+        self._last_homed = self.toolhead.get_status(curtime)['homed_axes']
+        if self.get_probe_state() == PROBE_UNKNOWN:
+                self._force_z_hop()
```
dalegaard, line 523: Indent wrong?

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+        if self.get_probe_state() == PROBE_ATTACHED:
+            # If X and Y endstop are within the dock location,
+            # Move out of the dock before attempting to home Z
+            if any(a in ['x','y'] for a in homing_axes):
```
dalegaard, line 527: Maybe `if 'x' in homing_axes or 'y' in homing_axes:` is easier to read?

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+        if self._check_distance(dist=self.approach_distance):
+            self._align_to_vector(self.dock_angle)
+        self.toolhead.manual_move(
+            [self.approach_position[0], self.approach_position[1], None],
+             self.travel_speed)
+
+    # Detaching actions
+    def detach_probe(self, return_pos=None):
+        self.pre_detach_gcode.run_gcode_from_command()
+        if self.get_probe_state() == PROBE_ATTACHED:
```
Sugaroverdose, line 454: This part is quite harmful in combination with quad_gantry_level since qgl changes Z height of gantry which can become lower than before, i've actually got probe smashed into bed because of this.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+            self._align_z()
```
Sugaroverdose, line 455: This part is quite harmful in combination with quad_gantry_level since qgl changes Z height of gantry which can become lower than before, i've actually got probe smashed into bed because of this.

docs/Config_Reference.md
```diff
@@ -1704,6 +1704,96 @@ control_pin:
...
+#   The X,Y position where the toolhead needs to be prior to moving into the
+#   dock so that the probe is aligned properly for docking or attaching.
+#   This parameter is required.
+detach_position: 0,0
+#   Similar to the approach_position, the detach_position is the X,Y
+#   where the toolhead is moved after the probe has been docked.
+#   For magnetically coupled probes, this is typically perpendicular to
+#   the approach_position in a direction that does not cause the tool to
+#   collide with the printer.
+#   This parameter is required
```
Sugaroverdose, line 1734: it would be good to have an option to make them Z movement as well(for instance, make approach_position a little higher than probe dock, as additional safety)

klippy/extras/dockable_probe.py
Sugaroverdose: also it seems like that something wrong happened to intend levels since only Z movement is actually prevented by 'not attached' state of probe

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+        gcode.register_command('GET_PROBE_STATUS',
+                                self.cmd_GET_PROBE_STATUS,
+                                self.cmd_GET_PROBE_STATUS_help)
+
+        if not any([self.check_open_attach,
+                    self.probe_sense_pin,
+                    self.dock_sense_pin]):
+            raise self.printer.config_error(HINT_VERIFICATION_ERROR.format(
+                                            aProbe.name))
+        else:
```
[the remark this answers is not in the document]
mental405: Done.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+                probe_sense_helper = configEndstop(self.probe_sense_pin)
+                self.probe_sense_pin = probe_sense_helper.query_pin
+            else:
+                self.probe_sense_pin = ehelper.query_pin_inv
+
+            if self.dock_sense_pin:
+                dock_sense_helper = configEndstop(self.dock_sense_pin)
+                self.dock_sense_pin = dock_sense_helper.query_pin
+            else:
+                if probe_sense_helper is not None:
```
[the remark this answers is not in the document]
mental405: Makes sense.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+                e = "Unable to parse {0} in {1}"
+                raise config.error(e.format(self.name, name))
+            return p
+
+        self.approach_position = parseCoord('approach_position')
+        self.detach_position   = parseCoord('detach_position')
+        self.dock_position     = parseCoord('dock_position')
+        self.safe_z_position   = parseCoord('safe_z_position', 2)
+
+        self.dock_angle, self.approach_distance = self._get_vector(
```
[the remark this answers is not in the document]
mental405: I think it could be. I was planning on maybe using it later but I can always break it back out.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+                rmin, rmax = rail.get_range()
+                if hi.positive_dir:
+                    mid = (hi.position_endstop - rmin) *.5
+                else:
+                    mid = rmax * .5
+                for s in rail.get_steppers():
+                    a = s.get_name(short=True)
+                    if a == 'x':
+                        home_xy[0] = mid - self.x_offset
+                    if a == 'x':
```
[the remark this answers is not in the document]
mental405: why yes, yes it should.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+        dock = self.dock_position
+
+        if dist > sqrt((pos[0]-dock[0])**2 +
+                       (pos[1]-dock[1])**2 ):
+            return True
+        else:
+            return False
+
+    # Find a point on a vector line at a specific distance
+    def _get_point_on_vector(self, point, angle, magnitude=1):
```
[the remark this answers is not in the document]
mental405: The convention in Klipper I believe is to prefix _internal/private_ methods with "_" and it is up to whomever is writing code to not use it as it has an "_" an should be considered off limits to external access.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+    def _handle_connect(self):
+        self.toolhead = self.printer.lookup_object('toolhead')
+    def _handle_ready(self):
+        self.last_z = -9999
+        self.is_z_endstop = False
+        self.multi = MULTI_OFF
+        self._last_homed = None
+        self.init_pos = None
+        kin = self.printer.lookup_object('toolhead').get_kinematics()
+        if not self.safe_z_position:
```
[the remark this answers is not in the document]
mental405: yes

klippy/extras/dockable_probe.py
mental405: I think that can be arranged.

klippy/extras/dockable_probe.py
mental405: yes

klippy/extras/dockable_probe.py
mental405: seems to be

klippy/extras/dockable_probe.py
mental405: thats how I had it written before then decided it was more illegible this way.

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+            if not self._check_distance(return_pos, self.detach_distance):
+                self.toolhead.manual_move(
+                    [return_pos[0], return_pos[1], None],
+                    self.travel_speed)
+    def _do_detach(self):
+        if self._check_distance(dist=self.detach_distance):
+            self._align_to_vector(self.dock_angle)
+        else:
+            self.toolhead.manual_move(
+                [self.detach_position[0], self.detach_position[1], None],
```
zifnab06, line 482: I believe this should be approach_position (otherwise you're going detach -> dock -> detach, not approach -> dock -> detach). 
```suggestion
                [self.approach_position[0], self.approach_position[1], None],
```

ctso: 👍 I was also having problems with detach, though strangely only when the `x` detach position was the same as the `x` attach position.

Either way, the behavior was `detach_position` -> `dock_position` -> `detach_position`, which is not correct.

Making this change solved my problems.

klippy/extras/dockable_probe.py
mental405: I am looking at splitting the alignment into two separate flows entirely. Additionally, there will be a 'hop only' restriction where it will only move Z in a positive direction.

docs/Config_Reference.md
mental405: See previous comment

klippy/extras/dockable_probe.py
mental405: I was helping another user on discord the other day when we discovered this typo. I should have to patched out in the next day or so along with some other corrections.

zellneralex:
It would be helpful to have result of the attach/detach also as Status Reference Object
So something like:
` def get_status(self, eventtime):
        return {'probe_state': self.last_verify_state}`
would be helpful to use the state inside macros and for console output

docs/G-Codes.md
```diff
@@ -369,6 +369,23 @@ see the [BL-Touch guide](BLTouch.md)):
 - `BLTOUCH_STORE MODE=<output_mode>`: This stores an output mode in the
   EEPROM of a BLTouch V3.1 Available output_modes are: `5V`, `OD`
 
+## Annexed Probe
```
zellneralex, line 372: Should be Dockable Probe

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+            if self.dock_sense_pin:
+                dock_sense_helper = configEndstop(self.dock_sense_pin)
+                self.dock_sense_pin = dock_sense_helper.query_pin
+            else:
+                if probe_sense_helper is not None:
+                    self.dock_sense_pin = probe_sense_helper.query_pin_inv
+                else:
+                    self.dock_sense_pin = ehelper.query_pin
+    cmd_GET_PROBE_STATUS_help = ("Prints the current probe state, valid probe" +
+                                 " states are UNKNOWN, ATTACHED, and DOCKED")
```
zellneralex, line 134: Typo states are UNKNOWN, ATTACHED and DOCKED"

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        gcode = self.printer.lookup_object('gcode')
+        gcode.register_command('SET_PROBE_STATUS',
+                                self.cmd_SET_PROBE_STATUS,
+                                desc=self.cmd_SET_PROBE_STATUS_help)
+
+        self.probe_state = PROBE_UNKNOWN
+    cmd_SET_PROBE_STATUS_help = ("Manually specify probe status, valid probe " +
+                                 "states are UNKNOWN, ATTACHED, and DOCKED")
```
zellneralex, line 64: Typo states are UNKNOWN, ATTACHED and DOCKED"

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+        self.pre_detach_gcode = gcode_macro.load_template(
+            config, 'pre_detach_gcode', '')
+        self.detach_gcode = gcode_macro.load_template(
+            config, 'detach_gcode', '')
+        self.post_detach_gcode = gcode_macro.load_template(
+            config, 'post_detach_gcode', '')
+
+        #Gcode Commands
+        self.gcode.register_command('ATTACH_PROBE',
+                                    self.cmd_ATTACH_PROBE)
```
zellneralex, line 246: add a desc=self.cmd_ATTACH_PROBE_help otherwise the gcode will not show up in help. Could be as simple as "Check probe status and attach probe"

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+        self.gcode.register_command('DETACH_PROBE',
+                                    self.cmd_DETACH_PROBE)
```
zellneralex, line 248: add a desc=self.cmd_DETACH_PROBE_help otherwise the gcode will not show up in help. Could be as simple as "Check probe status and detach probe"

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+                    if a == 'x':
+                        home_xy[0] = mid - self.x_offset
+                    if a == 'x':
+                        home_xy[1] = mid - self.y_offset
+                    if a == 'z':
+                        if hi.positive_dir:
+                            self.z_homes_positive = True
+                        else:
+                            self.z_homes_positive = False
+            self.safe_z_position = home_xy
```
zellneralex, line 300: add cmd_ATTACH_PROBE_help = ("Check probe status and attach probe") see above

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+    def cmd_ATTACH_PROBE(self, gcmd):
+        return_pos = self.toolhead.get_position()
+        self.attach_probe(return_pos)
+    def cmd_DETACH_PROBE(self, gcmd):
```
zellneralex, line 304: add cmd_DETACH_PROBE_help = ("Check probe status and detach probe") see above

docs/G-Codes.md
mental405: I have made this correction on my local repo already I just have not pushed the commits to github. Thanks for catching it though.

klippy/extras/dockable_probe.py
mental405: will do

wile-e1:
Are there any plans to deal with a probe that is still attached before all axes are homed?  For instance.... if the probe is attached and the machine times out, or if an emergency shutdown is required during an operation with the probe installed.

Some probe/machine configurations allow for the Z axis to be measured with the probe still attached, however some configurations do not.

Would it be possible to home X and Y and then maybe have a "safety" probe location and use the probe offset or a discrete value to know what the Z height is in order to be able to dock the probe without manual intervention.

for example, if I'm trying to do a bed mesh and for some reason, I have to hit e-stop, the probe is now docked to the toolhead all X Y Z references are gone.  In this instance, the probe could be used as a temporary endstop.  Here's how I envision things going:

- User tries to home XYZ
- Klipper checks to see if probe is attached, and it is, so it raises a flag and overrides standard homing routine with a safety routine
- X axis is homed
- Y axis is homed
- Toolhead moves to "safe location" (maybe center of build plate, user’s choice...of course)
- Z axis is lowered until probe is triggered
- Z Offset is adjusted by the difference in height from the nozzle to the probe trigger height (either by using 'z_offset' parameter or a discrete variable
- Klipper uses temporary Z height to dock the probe
- Klipper then reverts back to standard operation and then re-homes Z axis without probe.
- XYZ are now fully homed and the machine is ready to go

I'm not using this module right now, but I would like to try it again now that things have changed a little.  I was finding the angle attach/dock stuff far too hard to follow, but I see that is all changed now, so I may have to give it a go.  In any case, the code that I am using, I have implemented some code similar to what I have written out and it works perfectly.

lbibass:
Any updates on getting this pulled in?

mental405:
I am working on it when I have time to do so. It has been busy around here the last couple of.... years...

Some people use the word literally in a figurative sense. I _literally_ just started a print for a probe dock with sensing capabilities. I need to take that and confirm it works as intended. 

Based on all of the other feedback, I am rewriting a significant portion of the code in an effort to separate the different styles of probes and docks out there.. I hope to have this updated soon.

Jasrags:
Thank you for your work @mental405.

wlhlm:
I'm trying out a dockable probe (gantry-docking [Klicky](https://github.com/jlas1/Klicky-Probe/)) for the first time on a newly built V2. Here is my inital configuration:

```ini
[dockable_probe]
pin: PG9
dock_position: 47.5, 309
approach_position: 47.5, 280
detach_position: 309, 80
dock_fixed_z: True
z_offset: 0
dock_retries: 3
check_open_attach: True
```

Unfortunately, running `ATTACH_PROBE` results in an internal error:

```
Internal error on command:"ATTACH_PROBE"
  File "/opt/klipper/klippy/gcode.py", line 182, in _process_commands
    handler(gcmd)
  File "/opt/klipper/klippy/gcode.py", line 120, in <lambda>
  File "/opt/klipper/klippy/extras/dockable_probe.py", line 303, in cmd_ATTACH_PROBE
    self.attach_probe(return_pos)
  File "/opt/klipper/klippy/extras/dockable_probe.py", line 415, in attach_probe
    self._align_z()
  File "/opt/klipper/klippy/extras/dockable_probe.py", line 376, in _align_z
    self.position_endstop +
IndexError: list index out of range
```

Peaking at the code, it looks like it expects `dock_position` to have three coordinates, but I only specified two under the assumption that a gantry-fixed dock is independent of Z.

`dock_position` documentation says the following:
> ```
> #   [...] Certain dock designs are independent of the z axis.
> #   If a z value is provided for these configurations, the z axis will
> #   be raised to that amount in order to prevent collisions.
> #   This parameter is required.
> ```

From the second to last sentence I figured that the Z-coordinate is optional. My suggestion is to either validate that `dock_position` has three components or make sure that the code works with only the X- and Y-coordinate.

[klippy.log.gz](https://github.com/KevinOConnor/klipper/files/7198803/klippy.log.gz)

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,596 @@
...
+            try:
+                p = [None] * 3
+                val = config.get(name, None)
+                if not val:
+                    return None
+                p[0:c] = map(lambda x: float(x.strip()),
+                             val.split(','))[0:c]
+            except:
+                e = "Unable to parse {0} in {1}"
+                raise config.error(e.format(self.name, name))
```
wlhlm, line 195: Arguments `{0}` (self.name) and `{1}` (name) are swapped. Currently, this outputs the following into the log:
```
configparser.Error: Unable to parse dockable_probe in approach_position
```

wlhlm:
This module seems to be incompatible with Python 3 (3.9.9 in my case). The log contains the following:
```
  File "/usr/lib/klipper/klippy/extras/dockable_probe.py", line 191, in parseCoord
    p[0:c] = map(lambda x: float(x.strip()),
TypeError: 'map' object is not subscriptable
```

klippy/extras/dockable_probe.py
mental405: I have not tested this on python 3 yet but that should be easy enough to resolve.

wlhlm: Yeah, easy to fix. Here what I did for my personal Klipper branch: https://github.com/wlhlm/klipper/commit/11e60b64b7ee5ecb5c8e6c9c8f06b33fe87f369a

Other than that it's working fine.

netweaver1970:
Out of interest, as I have a dockable probe (Mantis toolhead with Annex Quickdraw derivative) install coming up, what needs to be done to get this merged into the master? It seems most of the work was done mid 2021.
Is it worthwile for me to wait for a mainline solution or should I just install your separate Klipper module when I get to that part?
Thanks for you work on these!

KevinOConnor:
What's the state of this PR?

-Kevin

MrKumaPants:
Been almost a year and this PR still hasn't made it in. What gives?

jp8311:
In the docs , line 267 says "There are 4 gcode commands that can be used..." and then proceeds to list 3.

kdb424:
I've been using this on several of my printers, and it's been working fairly well for me. The only "complaint" I have is the ability to type commands like `ATTACH_PROBE` but I can't tab complete them. This appears to be something with this module and possibly the way that it's exposed. If this is intentional, then this comment can be ignored. Really appreciate the work on making a standard for all of these amazing probes.

pawlowskialex:
One thing that would really be useful is a way to attach a probe from G-Code and keep it attached until a manual `DETACH_PROBE` is called. This way you in your start G-Code you can home, heat the bed, do tilt adjustment, bed mesh, and only THEN detach the probe.
I know there's the delayed detach feature, but I found it to be weird, I'd rather prefer a deterministic solution that gives me manual control when I need it, but is still completely automatic by default, so that a `G28` fired outside of the start G-Code will not break my printer.

mental405:
> One thing that would really be useful is a way to attach a probe from G-Code and keep it attached until a manual `DETACH_PROBE` is called. This way you in your start G-Code you can home, heat the bed, do tilt adjustment, bed mesh, and only THEN detach the probe. I know there's the delayed detach feature, but I found it to be weird, I'd rather prefer a deterministic solution that gives me manual control when I need it, but is still completely automatic by default, so that a `G28` fired outside of the start G-Code will not break my printer.

I tried to automate this but it proved to be problematic for a number of reasons. I am trying to keep the module out of Parameter Hell by balancing functionality with use. I have found that it actually doesnt add much overall time to probing moves when attaching/detaching between each operation and only increases the overall print time by about a minute. Which, in the grand scheme of a 4, 6, 8 hour print, isn't really much time when you get down to brass tacks.

pawlowskialex:
You could remove the delayed detach feature and replace it with `ATTACH_PROBE FORCE=TRUE` and we'd end up with the same amount of parameters :) Unless somebody is using that, of course.
Repeated attachments/detachments don't add any real print time, but they're annoying and make it feel unpolished.

Sineos:
For the sake of the mental health of the people doing support, I'd very much appreciate this feature coming. The macro and log hell these probes are causing is a mature pain in the back 👍

Vassssko:
> strange issue with this. added a huvud to my trident and it will not z_tilt, mesh bed, auto z, etc... getting error Internal error on command:"Z_TILT_ADJUST" Traceback (most recent call last): File "/home/pi/klipper/klippy/gcode.py", line 198, in _process_commands handler(gcmd) File "/home/pi/klipper/klippy/gcode.py", line 135, in func = lambda params: origfunc(self._get_extended_params(params)) File "/home/pi/klipper/klippy/extras/z_tilt.py", line 145, in cmd_Z_TILT_ADJUST self.probe_helper.start_probe(gcmd) File "/home/pi/klipper/klippy/extras/probe.py", line 428, in start_probe pos = probe.run_probe(gcmd) File "/home/pi/klipper/klippy/extras/probe.py", line 166, in run_probe pos = self._probe(speed) File "/home/pi/klipper/klippy/extras/probe.py", line 124, in _probe epos = phoming.probing_move(self.mcu_probe, pos, speed) File "/home/pi/klipper/klippy/extras/homing.py", line 245, in probing_move epos = hmove.homing_move(pos, speed, probe_pos=True) File "/home/pi/klipper/klippy/extras/homing.py", line 86, in homing_move triggered=triggered) File "/home/pi/klipper/klippy/extras/dockable_probe.py", line 584, in home_start print_time, sample_time, sample_count, rest_time, triggered) File "/home/pi/klipper/klippy/mcu.py", line 195, in home_start trsync.start(print_time, self._trigger_completion, expire_timeout) File "/home/pi/klipper/klippy/mcu.py", line 106, in start expire_ticks, min_extend_ticks) TypeError: initializer for ctype 'struct trdispatch_mcu *' must be a cdata pointer, not NoneType Transition to shutdown state: Internal error on command:"Z_TILT_ADJUST"

I have the same behavior after adding CAN toolhead. "TypeError: initializer for ctype 'struct trdispatch_mcu *' must be a cdata pointer, not NoneType". Looks like no fix available so far.

mental405:
It was brought to my attention that there could be a rounding error somewhere when calculating the approach angle to the dock if the dock position is at or very near the axis max position for the printer.

Vassssko:
Indeed, I have my dock at Xmax and Ymax position. Currently I gave up and wrote hard-coded macro, that work flawlessly along with tricky home procedure, preventing probe to be dropped off the head. Looiking forward for this awesome work to be completed and merged into mail klipper code already :)

mental405:
It has also come to my attention that running this in Python 3 has some issues. So far I have identified an incompatibility on line 187 where it is trying to parse the coordinates out from the config file.  In python 3 the map function needs to be converted to a list first. I am told this is not particularly efficient but it only has to execute once when the config is loaded.

`        def parseCoord(name, c=3):
            try:
                p = [None] * 3
                val = config.get(name, None)
                if not val:
                    return None
                p[0:c] = list(map(lambda x: float(x.strip()),
                             val.split(',')))[0:c]
            except:
                e = "Unable to parse {0} in {1}"
                raise config.error(e.format(self.name, name))
            return p`

JRHeaton:
What's the status on this PR and its level of completion? I am willing to help if needed. This has been something I feel that a TON of klipper users have been using for a long time, and having to keep pulling in main/master to their local clones of this fork. 

I guess TL;DR -- what's holding up the PR merge and can I assist at all?

mental405:
`what's holding up the PR merge`
-- I am busy and lazy
`and can I assist at all?`
-- Yes.

Read the comments, make the changes, submit a PR to my repo and I will merge.

bwnance:
@mental405 this might have already been addressed, but what's the behavior when a probing fails? (samples exceed tolerance, etc) -- will this dock the probe afterwards?

kyleisah:
Hello! I'm just wondering why this hasn't been merged yet. I'm looking at Mental's PR and the files changed, and everything *seems* to be in order. Is there something I'm not seeing? Having to have a separate "BED_MESH_CALIBRATE" macro just to define more macros to move to the dock area is giving me a lot of headaches right now, and I feel like now a module for stuff like this is a requirement.

lbibass:
This still needs to be merged. Any updates?

LobodPl:
To everybody having:
`TypeError: initializer for ctype 'struct trdispatch_mcu *' must be a cdata pointer, not NoneType`
Just add:
` self.printer.register_event_handler('klippy:mcu_identify',
                                            self._handle_mcu_identify)`
between
` self.printer.register_event_handler('klippy:ready',
                                            self._handle_ready)`
and
`self.printer.register_event_handler('homing:home_rails_begin',
                                            self._handle_homing_rails_begin)`
It'd look like this:
```
        self.printer.register_event_handler('klippy:ready',
                                            self._handle_ready)
        self.printer.register_event_handler('klippy:mcu_identify',
                                            self._build_config)
        self.printer.register_event_handler('homing:home_rails_begin',
                                            self._handle_homing_rails_begin)
```

KenVanHoeylandt:
It error'ed because couldn't find the method, so I checked `probe.py` and the method is in the class `ProbeEndStopWrapper`.
I just added the method manually to `dockable_probe.py`:

```python
    def _handle_mcu_identify(self):
        kin = self.printer.lookup_object('toolhead').get_kinematics()
        for stepper in kin.get_steppers():
            if stepper.is_active_axis('z'):
                self.add_stepper(stepper)
```

Probably not the best approach, but it works now.

blalor:
Ugh, really?

KevinOConnor:

-Kevin

KevinOConnor:
What is the status of this PR?  I wonder if we should track this on Klipper Discourse instead of on github?

-Kevin

PR #4331, opened 2021-05-30
safe_power: Stepper Driver Anti-SNAFU Protection

popshansen:
safe_power: Stepper Driver Anti-SNAFU Protection

Add stepper driver protection for boards to Klipper in form of 
safe_power module. The "Stepper Driver Anti-SNAFU Protection"
was introduced by BIGTREETECH in the "SKR 2" board and "SKR SE BX"
board. The protection give Klipper the option to check if Stepper Driver
is plugged into boards in correct orientation.

Signed-off-by: Lars R. Hansen popshansen@hotmail.com

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3907,6 +3907,47 @@ host_mcu:
 #   (True sets CFG5 high, False sets it low). The default is True.
 ```
 
+## [safe_power my_mcu_safe_power]
+
+Stepper Driver Anti-SNAFU Protection.
+
+Some micro-controllers have protection against incorrect plugged
+stepper drivers.
+Applying this section to the config file will let Klipper check the
+correct orientation of configured stepper drivers.
+If incorrect orientation then Klipper raise an error with message like this:
+"[stepper_z] driver is backward!"
+If the message appears then turn off the printer before any attempt on
+resolving the issue.
+
+If one may use TMC drivers the safe_power section have to be before
+any TMC sections in printer.cfg. It will check and raise config error if any
+TMC section is before the safe_power section.
+
+**WARNING: Using this on micro-controllers without a Stepper Driver Anti-SNAFU
+Protection may lead to undesirable results.**
+
+See the [generic-bigtreetech-skr-2.cfg](../config/generic-bigtreetech-skr-2.cfg)
+file for an example.
+
+```
+[safe_power my_mcu_safe_power]
+pin:
+#   Control pin of driver/heater/fan power supply
+```
+
+## [safe_power my_extra_mcu_safe_power]
+
+Additional micro-controllers with Stepper Driver Anti-SNAFU Protection
+(one may define any number of sections
+with an "safe_power" prefix). Additional micro-controllers may introduce
+additional safe power pin.
+
+```
+[safe_power my_extra_mcu_safe_power]
+# See the "safe_power" section for configuration parameters.
+```
+
 # Other Custom Modules
 
 ## [palette2]
--- a/klippy/extras/safe_power.py
+++ b/klippy/extras/safe_power.py
@@ -1,0 +1,112 @@
+# Stepper Driver Anti-SNAFU Protection support
+#
+# Copyright (C) 2021  Lars R. Hansen <popshansen@hotmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging
+
+SF_INVERT_STEP = 1<<2
+
+class Safe_Power:
+    def __init__(self, config):
+        self.name = config.get_name()
+        self.printer = config.get_printer()
+        self.reactor = self.printer.get_reactor()
+        ppins = self.printer.lookup_object('pins')
+        pin_params = ppins.lookup_pin(config.get('pin'))
+        self.mcu = pin_params['chip']
+        self.oid = self.mcu.create_oid()
+        self.pin = pin_params['pin']
+        self.mcu_pin = self.mcu.setup_pin('digital_out', pin_params)
+        self.mcu_pin.setup_start_value(0, 0, False)
+        self.set_digital_out_cmd = self.read_digital_in_cmd = None
+        self.steppers = {}
+        for s in config.get_prefix_sections('stepper_'):
+            self._register_stepper(s, ppins)
+        for s in config.get_prefix_sections('extruder'):
+            self._register_stepper(s, ppins)
+        for s in config.get_prefix_sections('manual_stepper'):
+            self._register_stepper(s, ppins)
+        if config.has_section('dual_carriage'):
+            self._register_stepper(config.getsection('dual_carriage'), ppins)
+        #Check config order
+        for name, obj in self.printer.lookup_objects():
+            if name.startswith('tmc'):
+               raise self.printer.config_error(
+                     "[%s] section should be before"
+                     " any tmc section in printer.cfg" % (self.name,))
+        self.mcu.register_config_callback(self._build_config)
+        self.printer.register_event_handler("klippy:connect",
+                                            self._connect)
+    def _build_config(self):
+        self.set_digital_out_cmd = self.mcu.lookup_command(
+            "set_digital_out pin=%u value=%c")
+        self.read_digital_in_cmd = self.mcu.lookup_query_command(
+            "read_digital_in pin=%u pull_up=%c",
+            "read_digital_in_value pin=%u value=%c")
+    def _connect(self):
+        ucmd = self.set_digital_out_cmd.send
+        res = None
+        # Disabling mcu power
+        ucmd([self.pin, 0])
+        if len(self.steppers) > 0:
+            logging.info("%s starting checks for stepper"
+                         " driver backwards" % (self.name,))
+            for stepper in self.steppers.values():
+                res = self._check_driver(stepper)
+                if res is not None:
+                    break
+        if res is not None:
+            raise self.printer.command_error(res)
+        else:
+            # Enabling mcu power
+            ucmd([self.pin, 1])
+    def _register_stepper(self, config, ppins):
+        name = config.get_name()
+        enable_pin = config.get('enable_pin', None, note_valid=False)
+        step_pin = config.get('step_pin', None, note_valid=False)
+        # Only register stepper driver with enable and step pins
+        if enable_pin is not None and step_pin is not None:
+            enable_pin_params = ppins.parse_pin(enable_pin, can_invert=True)
+            step_pin_params = ppins.parse_pin(step_pin, can_invert=True)
+            # Only register stepper driver belongs to current mcu
+            if (step_pin_params['chip'] is self.mcu
+                    and enable_pin_params['chip'] is self.mcu):
+                self.steppers[name] = StepperTracking(
+                    name, step_pin_params, enable_pin_params)
+    def _check_driver(self, stepper):
+        logging.info("Checking [%s]" % (stepper.name,))
+        res = None
+        # Reconfig stepper driver pins
+        curtime = self.reactor.monotonic()
+        print_time = self.mcu.estimated_print_time(curtime)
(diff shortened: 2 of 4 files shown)


popshansen:
This pull request should replace the previous pull request #4234 by FHeilmann.
I have SKR 2 boards Rev A modified to Rev B by replacing the mosfet that causing the issues introduced in Rev A. I have test this safe_power module on my printer by plugging stepper drivers in wrong direction and let Klipper shutdown to prevent any damages on SKR 2 and stepper drivers (TMC2209).

- Lars

KevinOConnor:
Thanks.  However, my high-level feedback is that I don't think the benefit of supporting this feature is currently a good trade-off for the costs of maintaining additional micro-controller code.  The feature is limited to a couple of recent hardware boards from one manufacturer and it's not clear to me that many users desire this feature nor that manufacturers will widely deploy it in the future.

If it could be implemented as an isolated "extras module" (that is, without micro-controller code and without changes to other core components) then that would be preferable.  Is that possible?  Is the test just a matter of setting the step pin low on all drivers and then sampling the enable pin on all drivers?

-Kevin

popshansen:
Thanks for reply. I investigate the Klipper and could see the way pins is managed it does not allow reconfigure pins from output to input to do the test and then change it back to output easily.
For once pins config has been finalized into the mcu you cannot change the pins config and resend it to the mcu again. That is why I add additional micro-controller code to allow me to do that. I tried to add reconfig into the safe_power.py but it was not possible because of this. Yes, it is a good idea to keep the micro-controller code minimized as much as possible but it was the only workaround that was easily enough to implement.
The way BTT has introduced this check is that if the stepper driver has been inserted the wrong way around, the enable pin and step pin is directly connected thru the drivers ground pins, so by writing low to the step pin you can read the enable pin to see if it is high or low. If high the driver is inserted correct and if is low it is inserted incorrect. Basically, BTT has reintroduced the E-stop by adding mosfet to control main power of the board. But for how long they will do this in the future is a good question. If it is just making SKR 2 standout from the previous models (1.3 and 1.4) or they see a real problem with burnt boards/drivers.  They didn’t add it to the new octopus board. I am not sure if other manufacturers will add the feature. 
I have checked my cramps boards as they also have an E-stop, but it is not possible to do driver checks because limitations on hardware level.

Anyway, it was a fun way to learn the deep technology of Klipper 😊
So, no harm if it is dumped/not merged.

/Lars

KevinOConnor:
Yeah, there isn't a way for the host to change a pin from output to input currently.  That is a functionality we could add though (the internal capability is already present in the mcu via the `gpio_out_reset()` and `gpio_in_reset()` functions).  So, if you are interested in pursuing it, that is an option.

Cheers,
-Kevin

popshansen:
My first attempt was calling `reset_pin_sharing()`but it was a no go, thats why I went with add a module to the mcu.

I could give it a try by adding additional commands to `gpiocmds` to call those gpio resets and let `safe_power.py` use those commands.
I can see that there is already an `command_set_digital_out()` I can use.

I will make that changes, then we will see if it works.

/Lars

popshansen:
Updated to use gpiocmds.c, I have added 2 cmds `command_set_digital_in()` and `command_read_digital_in()`.
I am not quit happy with `command_read_digital_in()` as it need to call `gpio_in_setup()` in order to get `struct gpio_in`, I missed a lookup method in gpio to get a gpio_in for/from the pin in order to read the pin as it is already setup for input. The `oid_lookup()`  is not a option to get the gpio_in as gpiocmds in theory don't know the `struct stepper`.
But anyway it is working with stm32.

/Lars

KevinOConnor:
Thanks.  In general it looks okay to me.  Some comments:
1. Adding `read_digital_in` is fine, but I'm not sure I see the need to add `set_digital_in` as the host could always just use `read_digital_in` and ignore the results.
2. It's not valid to call `sleep()` in the host code (it will cause problems with event dispatch).  It also wont properly implement a delay because of the way commands are queued.  To implement a delay, use the `minclock` mechanism - see klippy/extras/display/uc1701.py:ResetHelper() as an example.  Note that when using this mechanisms delays should be large (eg, 100ms).
3. I'd prefer not to modify stepper.py for this feature.  The safe_power.py module can lookup other config sections itself via `stepper_config = config.getsection(stepper_name)`.
4. If I understand this code change, it would apply the change to all steppers on all micro-controllers.  That doesn't seem correct as the alternate micro-controllers may not have this feature.
5. Minor, but it would be preferable to raise an error on a "backwards stepper" as opposed to going into shutdown mode.

Cheers,
-Kevin

popshansen:
Thanks for feedback.

1. I will check if it is okay to remove the `set_digital_in`.

2. I'll have look at uc1701.py.

3. The check should be done as early as posible, at least before any enable power and configuration of TMC's. If TMC drivers dosen't have power before any configuration you will get TMC configurtion errors.
I have not seen anyway to prioritize modules, that's why I modified the stepper.py. But I have proberly misunderstand something, I will do some more debugging to understand when stuff get's executed.
I will change to use the `get_prefix_sections('stepper_')` as they have different names depended on kinematics type.


4. It would make sense to have mcu check, that is checking steppers belongs to the mcu for which the safe power pin is configured for.

5. I will change the error handling.

/Lars

KevinOConnor:
> The check should be done as early as posible, at least before any enable power and configuration of TMC's. 

FYI, you may be able to use the 'klippy:mcu_identify' callback (instead of 'klippy:connect') if you need a very early callback.

-Kevin

popshansen:
Got it to work independent of the stepper.py, added code for multiple mcu's. Each safe_power sections only check drivers that belongs to the mcu.
Shutdown is replaced by exceptions.
The order in the printer.cfg determine the execution order, so the code checks that the section is before any tmc sections, if the safe_power is after you will get errors from the configuration of the tmc as the is no power to the driver before the configuration is sent to the driver.
Replaced the sleep pause with the queue timings. I’ll guess it is correct.
It looks for all four types of stepper drivers (stepper_, extruder, manual_stepper and dual_carriage).
If there is no stepper for an mcu, it will just enable the power to make other functions work, on SKR 2 it is heater and fans.
The 'klippy:mcu_identify' was a little to early as no configuration was carried out, and there for no communications with mcu. So the best place is the ‘klippy:connect’.

/Lars

docs/Config_Reference.md
```diff
@@ -3895,6 +3895,46 @@ host_mcu:
 #   (True sets CFG5 high, False sets it low). The default is True.
 ```
 
+## [safe_power]
```
KevinOConnor, line 3898: I think the documentation would be simpler if it always had a name on the config section (eg, `[safe_power myname]`).

klippy/extras/safe_power.py
```diff
@@ -0,0 +1,116 @@
+# Stepper Driver Anti-SNAFU Protection support
+#
+# Copyright (C) 2021  Lars R. Hansen <popshansen@hotmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging, time
+
+class BackwardError(Exception):
```
KevinOConnor, line 8: Raising a BackwardError will effectively halt Klipper on an error.  I think it would be preferable for the code to raise a `printer.config_error()` or `printer.command_error` on a problem.  That provides better diagnostics.

klippy/extras/safe_power.py
```diff
@@ -0,0 +1,116 @@
...
+            self._register_stepper(s, ppins)
+        for s in config.get_prefix_sections('extruder'):
+            self._register_stepper(s, ppins)
+        for s in config.get_prefix_sections('manual_stepper'):
+            self._register_stepper(s, ppins)
+        if config.has_section('dual_carriage'):
+            self._register_stepper(config.getsection('dual_carriage'), ppins)
+        #Check config order
+        for name, obj in self.printer.lookup_objects():
+            if 'tmc' in name.lower():
```
KevinOConnor, line 38: A search for 'tmc' may conflict with arbitrary user names.  Consider using `name.startswith("tmc")`.

klippy/extras/safe_power.py
```diff
@@ -0,0 +1,116 @@
...
+        logging.info("Checking [%s]" % (stepper.name,))
+        res = None
+        # Reconfig stepper driver pins
+        curtime = self.reactor.monotonic()
+        print_time = self.mcu.estimated_print_time(curtime)
+        minclock = self.mcu.print_time_to_clock(print_time + .100)
+        self.set_digital_out_cmd.send([stepper.step_pin_params['pin'],0],
+                                      minclock=minclock)
+        # Do the stepper driver check
+        minclock = self.mcu.print_time_to_clock(print_time + .100)
```
KevinOConnor, line 91: This `minclock` line will return an identical value as the previous `minclock` calculation.  Maybe you want `print_time + 0.200` to ensure a 100ms delay since the pin was set as output?

klippy/extras/safe_power.py
```diff
@@ -0,0 +1,116 @@
...
+                    break
+        if res is not None:
+            raise self.error(res)
+        else:
+            # Enabling mcu power
+            ucmd([self.pin, 1])
+    def _register_stepper(self, config, ppins):
+        name = config.get_name()
+        enable_pin = config.get('enable_pin', None)
+        step_pin = config.get('step_pin', None)
```
KevinOConnor, line 71: When inspecting config values from other config sections, be sure to use `config.get(..., note_valid=False)` - otherwise it can confuse the config parameter checks.

KevinOConnor (commented):
Thanks.  In general it looks fine to me.  I have a few comments below.  Also, this has a merge conflict with the master branch - it will probably need a rebase in order for me to commit.

-Kevin

docs/Config_Reference.md
popshansen: I will tighten up the documentation

/Lars

klippy/extras/safe_power.py
popshansen: I had my thought on execption types. As I understand the config_error is issues in printer.cfg. But I can change the code to use it. If it raise an error you need to switch off you printer and rotate the stepper drive to the correct orientation. Maybe the switch off printer message should be part of the error message, so the user know's what todo to resolve the issue.

/Lars

popshansen:
> Thanks. In general it looks fine to me. I have a few comments below. Also, this has a merge conflict with the master branch - it will probably need a rebase in order for me to commit.
> 
> -Kevin

Will it solve the conflict if I update my fork with a fresh copy and apply my changes to it before commit it to the pull request?

/Lars

KevinOConnor:
> Will it solve the conflict if I update my fork with a fresh copy and apply my changes to it before commit it to the pull request?

Probably.  Another option is to use `git rebase`.

-Kevin

popshansen:
@Kevin, All your suggestions is applied. I tried with git rebase but everything went wrong for me, I am still too green on git, so I had to grab a fresh fork and applied my changes to it and it works.

-Lars

KevinOConnor:
Thanks.  In general it looks fine.  However, it's breaking the PRU build due to increase in code size on that platform.  I have a rough plan to improve the PRU build so there is more code space available.  I guess this change will need to wait until that infrastructure work can be completed.

-Kevin

KevinOConnor:
Thanks for working on this.  The changes to free up space on the PRU were completed a couple of months back.  So, it should now be possible to add `command_read_digital_in()` to all mcu boards.

However, as a high-level comment, I'm not sure adding this feature is still worthwhile - it seems the SKR 2 wasn't a particularly well selling board, there is already a workaround, and it doesn't seem like BTT is reusing this mechanism on other boards.

-Kevin

popshansen:
That is also my thoughts about this. I think it is better to scrap this request and use the free space to something more useful. I don't think there will be any new MCU's with this feature. BTT has put several new boards on the market and none contains this mechanism. For my part it was great learning of the Klipper code. So no regreat from my part if you close/delete the request.

PR #4335, opened 2021-05-31
Update gcode descriptions

meteyou:
This is necessary, that the gcodes are displayed in the HELP response and that they are forwarded to the GUIs via the API.


KevinOConnor:
Thanks.

Separately, it may make sense to extend the webhooks API to list all available commands (even those without help).

-Kevin

meteyou:
or you could use this condition as a "filter" to not send gcodes like "accept" to the client. sometimes it might even make sense not to list it in help.

PR #4339, opened 2021-06-02
neopixel/dotstar: add sync parameter to prevent waking toolhead

mental405:
The addressable LED code resets the idle timeout when the `SET_LED` command is executed. This results in resetting the idle_timeout. When used in macros and delayed_gcode timers to react to things like temperature outside of a running print, this can result in the idle timeout never being reached and LCD screens to repeatedly switch from idle=>printing every time the macro executes in the background.

The new `SYNC=0` parameter changes the scheduling of the `SET_LED` command to where it won't sync with print time.

I also somewhat  refactored dotstar.py so that the class members and parameters are the same as neopixel.py.


KevinOConnor:
Thanks.

-Kevin

PR #4340, opened 2021-06-02
Create printer-tronxy-x1-2017.cfg

cron410:
All features and buttons are working with this config. Extruder max_temp has been set to 230C for PTFE hotend toxin safety. Kids are the most common user of this bargain printer.


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit directly (see https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md ).  If that's an issue, I can commit a basic config file using my signed-off-by line though (if so, let me know).

-Kevin

cron410:
I need to update the config anyway, I'll try again.

PR #4341, opened 2021-06-02
Controller fan with temperature sensor

mnigbur:
This PR adds a target temperature to temperature sensors, which in return can be used with controller_fan. A controller fan is active when either aheater is active, or a temperature sensor has reached it's target.

This is quite useful for the controller fans in the Voron 2 lower deck, whereyou want air circulation for the steppers the RPi / any other controller board.

Sample config: 
```yaml
[temperature_sensor pi]
sensor_type: temperature_host
min_temp: 0
max_temp: 85
target_temp: 55

[temperature_sensor spider]
sensor_type: temperature_mcu
min_temp: 0
max_temp: 85
target_temp: 40

[controller_fan controller_fan1]
pin: PB2
kick_start_time: 0.5
heater: heater_bed, extruder
fan_speed: 0.30
idle_timeout: 60
temperature_sensor: pi, spider
```

### DIFF
--- a/docs/Code_Overview.md
+++ b/docs/Code_Overview.md
@@ -331,12 +331,11 @@ Useful steps:
    seconds) to a cartesian coordinate (in millimeters), and then
    calculate the desired stepper position (in millimeters) from that
    cartesian coordinate.
-4. Implement the `calc_tag_position()` method in the new kinematics
-   class. This method calculates the position of the toolhead in
-   cartesian coordinates from the position of each stepper (as
-   returned by `stepper.get_tag_position()`). It does not need to be
-   efficient as it is typically only called during homing and probing
-   operations.
+4. Implement the `calc_position()` method in the new kinematics class.
+   This method calculates the position of the toolhead in cartesian
+   coordinates from the position of each stepper. It does not need to
+   be efficient as it is typically only called during homing and
+   probing operations.
 5. Other methods. Implement the `check_move()`, `get_status()`,
    `get_steppers()`, `home()`, and `set_position()` methods. These
    functions are typically used to provide kinematic specific checks.
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2010,6 +2010,8 @@ temperature sensors that are reported via the M105 command.
 #max_temp:
 #   See the "extruder" section for the definition of the above
 #   parameters.
+#target_temp:
+# Specifies the target temperature when used in an controller_fan section
 #gcode_id:
 #   See the "heater_generic" section for the definition of this
 #   parameter.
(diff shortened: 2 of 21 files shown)


meteyou:
I would prefer to define the target_temp in the controller_fan like this:

```
[controller_fan controller_fan1]
pin: PB2
kick_start_time: 0.5
heater: heater_bed, extruder
fan_speed: 0.30
idle_timeout: 60
temperature_sensor:
    pi:55
    spider:40
```

mnigbur:
It actually might make sense to have different targets for different fans and the same temperature_sensor,
Moved everything into the controller fan.

New sample config:
```yaml
[temperature_sensor pi]
sensor_type: temperature_host
min_temp: 0
max_temp: 85

[temperature_sensor spider]
sensor_type: temperature_mcu
min_temp: 0
max_temp: 85

[controller_fan controller_fan1]
pin: PB2
kick_start_time: 0.5
heater: heater_bed, extruder
fan_speed: 0.30
idle_timeout: 60
temperature_sensor:
    pi: 55,
    spider: 40
```

KevinOConnor:
I'm not sure what the state of this PR is (it doesn't pass the build test cases).

There have been a few proposed changes to controller_fan (eg, PR #4447 ).  I'm not sure how this would interact with that.

I'm not sure what the best way to accomplish the goals of this PR, but FWIW, I find the proposed config syntax to be complex.

-Kevin

mnigbur:
> I'm not sure what the state of this PR is (it doesn't pass the build test cases).
> 
> There have been a few proposed changes to controller_fan (eg, PR #4447 ). I'm not sure how this would interact with that.
> 
> I'm not sure what the best way to accomplish the goals of this PR, but FWIW, I find the proposed config syntax to be complex.
> 
> -Kevin

I simply pushed into the wrong branch, that's why my fixes never showed up..
In order to include the target temperature for the additional temperate sensors I had to get a little creative with the config syntax. Decided for yaml style, but we could change it as it's used with the board pins?

However: Looks like this does not work properly with the handle_connect yet and I need to change the command to be controller_fan aware.

KevinOConnor:
FWIW, I'm not sure I understand the "big picture" of this PR.  Is the goal to have a "heater_fan" with a little more customizability?  I'd be inclined to say this may not be a good match for the Klipper master branch - as I'm not sure additional configuration and documentation complexity is a good trade off for an uncommon use case.

-Kevin

mnigbur:
@KevinOConnor the idea is to have e.g. the skirt fans of the Voron V2 not only running when the the defined heaters are running, but also when any of the defined temperature sensors reach a certain limit.
Even though the configuration might look a little bit more complex than usual, it actually reduces the needed configuration **and** hardware required to accomplish the same thing otherwise.

edddeduck:
This seems like an extremely valuable feature. 

For example being able to control an electronics bay fan based on various temperature sensors on different hardware inside the bay would be very useful so this PR seems worth investigating / reopening.

PR #4342, opened 2021-06-03
Change M204 behavior: update `max_accel_to_decel` too

lijgame:
Set max_accel_to_decel to max_accel/2 when M204 is called.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -18,6 +18,7 @@ Klipper supports the following standard G-Code commands:
   - Note: If S is not specified and both P and T are specified, then
     the acceleration is set to the minimum of P and T. If only one of
     P or T is specified, the command has no effect.
+  - Note2: This command will set `max_accel_to_decel` to `value/2`
 - Get extruder temperature: `M105`
 - Set extruder temperature: `M104 [T<index>] [S<temperature>]`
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks, but as a general rule we don't alter the behavior (nor add new behavior) for "legacy" G-Code commands.  That is, the old M123 style commands "do what they do" and no more.  If some other behavior is desired then consider using SET_VELOCITY_LIMIT or use a gcode_macro config section to override the behavior of the M204 command.

-Kevin

lijgame:
> Thanks, but as a general rule we don't alter the behavior (nor add new behavior) for "legacy" G-Code commands. That is, the old M123 style commands "do what they do" and no more. If some other behavior is desired then consider using SET_VELOCITY_LIMIT or use a gcode_macro config section to override the behavior of the M204 command.
> 
> -Kevin

Thanks Kevin.
That makes a lot of sense.
I will close this PR.

In case someone else is looking for it, here is the macro to achieve the same effect:

````
[gcode_macro M204]
rename_existing: M204.old
gcode:
    {% set ACC = params.S|default(1000)|float %}
    {% set ACC2DECEL = params.S|float/2 %}
    SET_VELOCITY_LIMIT ACCEL={ACC} ACCEL_TO_DECEL={ACC2DECEL}

PR #4346, opened 2021-06-03
bed_mesh: add support for X/Y offsets

Arksine:
This implements `BED_MESH_OFFSET`, which can be used to apply offsets to the mesh lookup for correct adjustment on printers with multiple extruders.  The offset should be relative to the primary extruder, so generally it should be the additive inverse of the gcode offset.

The first commit in this series contains a minor refactor of how the `fade_target` is handled. The initial offset is no longer pre-calculated when the mesh is set, instead the offset is applied when the final adjustment is calculated.  This made adding the X and Y offsets less confusing.

Signed-off-by:  Eric Callahan <arksine.code@gmail.com>.

### DIFF
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -430,3 +430,14 @@ probing the "Probe" points will refer to both the tool and nozzle locations.
 `BED_MESH_CLEAR`
 
 This gcode may be used to clear the internal mesh state.
+
+### Apply X/Y offsets
+
+`BED_MESH_OFFSET [X=<value>] [Y=<value>]`
+
+This is useful for printers with multiple independent extruders, as an offset
+is necessary to produce correct Z adjustment after a tool change.  Offsets
+should be specified relative to the primary extruder.  That is, a positive
+X offset should be specified if the secondary extruder is mounted to the
+right of the primary extruder, and a positive Y offset should be specified
+if the secondary extruder is mounted "behind" the primary extruder.
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -434,6 +434,10 @@ The following commands are available when the
   supplied name from persistent memory.  Note that after SAVE or
   REMOVE operations have been run the SAVE_CONFIG gcode must be run
   to make the changes to peristent memory permanent.
+- `BED_MESH_OFFSET [X=<value>] [Y=<value>]`:  Applies X and/or Y
+  offsets to the mesh lookup.  This is useful for printers with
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Thanks.  Looks fine to me and I'll merge when you're ready.  A couple of minor questions:

Does this alter the behavior of previously saved meshes?

What is the new BED_MESH_OFFSET relative to?  As an example, lets say I have `mesh_min: 20,20`, `mesh_max: 180:180`, a probe 10mm to the right of the primary nozzle (`x_offset: 10`), and a secondary nozzle 20mm to the right of the primary nozzle - what parameters should be sent to BED_MESH_OFFSET when switching to the secondary hotend?

-Kevin

Arksine:
> Does this alter the behavior of previously saved meshes?

No, they won't be affected.  The profiles contain the probed z values along with the parameters necessary to generate the mesh.  When a profile is loaded, a new mesh is generated from it.  The X and Y offsets will initialize to 0.  The fade target will be applied during the z lookup just as it would with a mesh generated by BED_MESH_CALIBRATE.

> What is the new BED_MESH_OFFSET relative to? As an example, lets say I have mesh_min: 20,20, mesh_max: 180:180, a probe 10mm to the right of the primary nozzle (x_offset: 10), and a secondary nozzle 20mm to the right of the primary nozzle - what parameters should be sent to BED_MESH_OFFSET when switching to the secondary hotend?

The offset is relative to the primary extruder.  In your scenario `BED_MESH_OFFSET X=20` would be necessary when changing from the primary to the secondary extruder.  Essentially we just need to cancel out the offset applied by `SET_GCODE_OFFSET`.  If `BedMesh.move()` receives X=10, the X offset will result in a lookup at X=30, which is where the secondary extruder is relative to the bed.  If this is confusing we can make the offset relative to the secondary extruder.  In that case, the values for `BED_MESH_OFFSET` would just match the values for `SET_GCODE_OFFSET`.

@shanewarner has agreed to test this PR.  I think it would be a good idea to make sure it behaves correctly on a printer with multiple extruders proceed with a merge if everything goes as planned.

shanewarner:
I tested this over the weekend with several large flat prints on the second extruder, and it appears to be working very well. I incorporated the BED_MESH_OFFSET commands into my T0/T1 commands as per your documentation. Thanks for the fix!

Arksine:
Great, thanks Shane.  Kevin, this is ready whenever you are, thanks!

KevinOConnor:
Thanks.

-Kevin

PR #4354, opened 2021-06-06
Add support for the Prusa Mini+ and the Prusa Buddy board

matthewlloyd:
The Buddy board will likely be used by the MK4, so has a separate generic config file.

### DIFF
--- a/config/generic-buddy.cfg
+++ b/config/generic-buddy.cfg
@@ -1,0 +1,159 @@
+# This file contains common configurations and pin mappings for the Prusa Buddy
+# board. The LCD is not currently supported by Klipper, so the touchscreen will
+# permanently display the bootloader screen after the Klipper firmware is flashed;
+# use Fluidd, Mainsail, or OctoPrint etc. to control the printer.
+
+# To use this config, the firmware should be compiled for the STM32F407. When
+# running "make menuconfig", enable "extra low-level configuration setup",
+# select the "128KiB + 512 byte offset" bootloader, and USB communication.
+# Connect the printer to your Raspberry Pi using the printer's micro-USB port.
+# If you prefer to remove Prusa's stock bootloader entirely, select the
+# "No bootloader" option.
+
+# When flashing for the first time, you will need to break the "appendix"
+# on the Buddy board, then put the device into DFU mode by moving the jumper
+# on the 3-pin header (older boards) or shorting the 2-pin header (newer boards)
+# and resetting, and finally use "make flash" to install Klipper. Once Klipper is
+# installed, you no longer need the jumper - just use "make flash" which will
+# automatically put the device into DFU mode.
+
+# Note that if you were previously running Prusa firmware, you must fully
+# power cycle the board after flashing. Otherwise, Klipper will be unable to
+# communicate with the TMC2209s due to the abrupt change in the baud rate,
+# and will show this error: "Unable to read tmc uart register IFCNT".
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[stepper_x]
+step_pin: PD1
+dir_pin: PD0
+enable_pin: !PD3
+microsteps: 16
+rotation_distance: 32
+endstop_pin: tmc2209_stepper_x:virtual_endstop
+position_endstop: 200
+position_min: 0
+position_max: 200
+homing_speed: 50
+homing_retract_dist: 0
+
+[stepper_y]
+step_pin: PD13
+dir_pin: PD12
+enable_pin: !PD14
+microsteps: 16
+rotation_distance: 32
+endstop_pin: tmc2209_stepper_y:virtual_endstop
+position_endstop: 0
+position_min: 0
+position_max: 200
+homing_speed: 50
+homing_retract_dist: 0
+
+[stepper_z]
+step_pin: PD4
+dir_pin: !PD15
+enable_pin: !PD2
+microsteps: 16
+rotation_distance: 4
+endstop_pin: probe:z_virtual_endstop
+position_min: 0
+position_max: 200
+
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Thanks.

Some high-level comments:
1. The printer-prusa-mini-plus-2020.cfg (and the associated build changes) looks fine to me.
2. Is it necessary to add a generic-buddy.cfg file?  It seems the comments in the file are geared towards the mini-plus anyway (eg, the probe comments are specific to the printer, not the controller board).
3. The changes to tmc2209.py don't look correct to me.  It's the norm for the tmc2209 to run at 12Mhz and the 9000 baud works well on thousands of these chips.  A tmc2209 clock rate slower than 20Mhz would actually permit an even slower baud rate than 9000.  Can you go into more detail on the issue you were seeing with the default baud?  I'm leery of adding a config option for this as I don't think it's reasonable for users to select a valid value (given various tmc and mcu processing constraints).

Cheers,
-Kevin

matthewlloyd:
Hi Kevin,

Thanks for the quick response. I'm hoping, given the large base of Mini+ owners, this will help make Klipper even more popular! As of ~2 hours ago I have the LCD and encoder wheel working too, but will put those changes in a separate PR.

2. generic-buddy.cfg.

The upcoming Prusa MK4 will almost certainly share the Buddy board (and LCD) with the Mini+: there are references to the MK4 in the Buddy firmware repo. I added this board config just following perceived convention, as for example the Creality printers have generic-creality* configs for their associated boards. It's entirely your call, and I don't mind either way. I expect the Mini+ and the MK4 will be the only off-the-shelf printers using this board for the forseeable future. Some hobbyists may be using clone boards but likely not many.

3. TMC baud.

The error consistently produced over many hard resets was the standard "Unable to read tmc uart 'stepper_X' register IFCNT" which seems common when the UART isn't working. I checked the wiring, pins, tried variations of the UART addresses, I double and triple checked, but the firmware just wouldn't talk to the TMC2209s. It was by luck that I happened to try changing the baud rate setting (and nothing else), and it worked. I even changed it back and it stopped working again - and repeated the experiment twice.

I agree that a slower clock rate than 20MHz should allow a slower baud than 9000. It seems odd this was necessary. Perhaps my board was some a bad state for some reason, or I was hitting some as yet unknown bug. Let me do some further testing to make sure this is really needed, and get back to you tomorrow evening.

Cheers,
Matthew

KevinOConnor:
> I agree that a slower clock rate than 20MHz should allow a slower baud than 9000. It seems odd this was necessary. Perhaps my board was some a bad state for some reason, or I was hitting some as yet unknown bug. Let me do some further testing to make sure this is really needed, and get back to you tomorrow evening.

Ah, okay.  I think I know what you hit.  If anything communicates with the tmc drivers at a high baud rate then it wont accept a baud rate significantly lower.  From the spec: `Remark, that due to this mechanism an abrupt reduction of the baud rate to less than 15 percent of the previous value is not possible.`  So, if you were using some other software (with a high baud rate) prior to installing Klipper than it would be necessary to completely power down the printer (both USB and motor power off) before Klipper will communicate with the chips.  I've run into this and it is very confusing.

-Kevin

matthewlloyd:
> it would be necessary to completely power down the printer (both USB and motor power off) before Klipper will communicate with the chips

Yes, further testing indicates this was the issue! Flashing Prusa firmware and homing, flashing Klipper, then hitting the reset button, with the config set to the default 9000 baud, consistently produced the IFCNT error on my two attempts. The same sequence, using a full power cycle with 5 second wait instead of a reset, again at 9000 baud, yields no IFCNT error. The board seems to retain some power for a few seconds after power-down (an LED stays lit at least), so that may have been masking the issue previously.

I will update this PR to remove the baud rate changes, and add a note about fully power-on resetting after flashing to the Buddy and Mini+ config files. I'll also add a note in the TMC stepper documentation in case anyone else runs into it.

Relatedly... I was originally hoping to be able to add support for dual boot (Klipper / Prusa) through a custom bootloader. Prusa is running the steppers at 115200 baud (https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/ce8840291035daea3cff700bc5e307037a6bcf60/src/main.cpp#L822). It did make me wonder whether there is room for Klipper to increase the TMC baud rate a little.

KevinOConnor:
> I will update this PR to remove the baud rate changes, and add a note about fully power-on resetting after flashing to the Buddy and Mini+ config files. I'll also add a note in the TMC stepper documentation in case anyone else runs into it.

Okay, thanks.

> Prusa is running the steppers at 115200 baud. It did make me wonder whether there is room for Klipper to increase the TMC baud rate a little.

Klipper supports single wire operation and can use any gpio pin for the TMC UART.  Most other firmware require that the TMC UART wire be routed to a MCU hardware UART block and that there be two wires routed (rx and tx).  Klipper's implementation is more flexible and more portable, but it does require mcu processor time when transmitting messages.  A lower baud helps reduce cpu time and there isn't any particular reason to transmit messages to the stepper drivers faster.

That said, I have considered increasing the baud rate on faster mcus while leaving a low baud on the slower AVR chips.  So far, however, there just hasn't been any compelling reason to make a change.

Cheers,
-Kevin

KevinOConnor:
Thanks.  I committed this change and renamed "generic-buddy.cfg" to "generic-prusa-buddy.cfg" on top (commit  6dcb902d).  (A "generic buddy" sounds odd to me.)

Separately, it would be nice if you could inspect the mini-plus extruder gears and add a `gear_ratio` for the printer - as we generally prefer to list the rotation_distance as the circumference of the bolt that pushes the filament.

Cheers,
-Kevin

PR #4356, opened 2021-06-07
Draft: Replicate Prusa MK3 hotbed silencer

ayufan:
Prusa MK3 have this nosiy problem of clicking bed when switching.

Prusa fixed that to some extent by introducing a 62.5kHz increasing duty cycle on switching.
It was well documented in https://blog.prusaprinters.org/dev-diary-2-how-we-made-the-heatbed-silent_30946/.

This PR does:

- makes problem of clicky bed go away
- replicate what Prusa does in their's firmware
- does it in a much simpler form retaining soft PWM behavior
- this generates a `STEPS` amount of `ISR` for each edge change at 62.5kHz, exactly how the mentioning blog post does (significantly better, as Prusa uses ISR on regular intervals either be 62.5kHz or 7.8kHz)
- it was validated using `simavr` and with osciloscope

Problems:

- I have no idea how best integrate this. This likely makes sense as a compile-type of behavior
- It is hardcoded to PG5 a heatbed PIN

### DIFF
--- a/src/avr/bed_timer.c
+++ b/src/avr/bed_timer.c
@@ -1,0 +1,39 @@
+#include "autoconf.h" // CONFIG_MACH_atmega644p
+#include "gpio.h" // gpio_out_write
+
+#define STEPS 16
+#define DUTY_CYCLE (256/STEPS)
+
+ISR(TIMER0_OVF_vect)
+{
+	if (OCR0B + DUTY_CYCLE >= 255) {
+		OCR0B = 255;
+		TIMSK0 &= ~(1 << TOIE0);
+	} else {
+		OCR0B += DUTY_CYCLE;
+	}
+}
+
+void timer0_restart(uint8_t val)
+{
+	TCNT0 = 0;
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Interesting, thanks.

I'm not sure how to integrate it either.  It's very specific to the printer hardware and to the micro-controller.  I'd probably look at creating a new C module with its own commands (eg, see how src/linux/pca9685.c creates its own PWM commands) and then create a new "extras" module that implements the host PWM interface (eg, see how klippy/extras/replicape.py implements the pca9685_pwm class).

Cheers,
-Kevin

KevinOConnor:
This does look interesting.  Due to time constraints on my side though, I'd prefer not to track draft PRs in the main Klipper repository.  Perhaps Klipper Discourse would be an alternative?

-Kevin

ayufan:
I'm closing this as even though it works, in this current form it introduces a temperature fluctuations due too much power passed to heatbed.

image

PR #4359, opened 2021-06-08
menu: auto-refresh menu visible items

ayufan:
This makes to dynamically refresh menu items on a heartbeat
according to current print state.

To make things slightly more stable, we always keep
the current selected item always visible.

This is a follow-up after the: https://github.com/KevinOConnor/klipper/pull/4336.

### DIFF
--- a/klippy/extras/display/menu.py
+++ b/klippy/extras/display/menu.py
@@ -346,9 +346,15 @@ class MenuContainer(MenuElement):
         self.send_event('populate', self)
 
     def update_items(self):
+        selected_item = self.selected_item()
+
         _a = [(item, name) for item, name in self._allitems
-              if item.is_enabled()]
+            if item.is_enabled() or item == selected_item]
         self._items, self._names = zip(*_a) or ([], [])
+
+        # ensure that prior item is selected
+        if selected_item:
+            self.select_item(selected_item)
 
     # select methods
     def init_selection(self):
@@ -597,6 +603,10 @@ class MenuList(MenuContainer):
         self._viewport_top = 0
         #  add back as first item
         self.insert_item(self._itemBack, 0)
+
+    def heartbeat(self, eventtime):
+        super(MenuList, self).heartbeat(eventtime)
+        self.update_items()
 
     def draw_container(self, nrows, eventtime):
         display = self.manager.display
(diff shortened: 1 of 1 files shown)


KevinOConnor:
FYI, on all the menu changes I'll be looking for feedback from @mcmatrix.

Separately, if you're on Klipper Discord, let me know.

-Kevin

mcmatrix:
Hi @ayufan,
The problem is not only the currently selected item.

The selection will start jumping (position-based) when some menu items
in the open list will appear or disappear. 

Btw, In the beginning, it was updating the open menu items but in some situations, 
this made the use of the menu erratic. So I had to disable it.

For example, you have the following list and some items 
before selection or after selection will appear or disappear.
You can play it thru and see how the menu behaves.
```
  ..
  item 1
  item 2
  item 3
 >item 4
  item 5
  item 6
  item 7
```
Therefore, I'm not sure if this simple solution is stable enough to use.

-Janar

ayufan:
@mcmatrix 

Correct :) I tried two ways:

1. always recalculate visible ones, and pick the first prior visible to retain cursor
2. make the current one as a special case that it is always visible regardless of the flag

The first one, was pretty bad, as it created a lot of randomness that you mention. I was getting cases that `Move ?` would result in majority of settings to disappear, and loose the correct focus.

This is why I tried this one. I'm currently running this one, and it works decent with some caveats. It is dynamic, but appears to be stable enough. It means that if you move cursor, your option will disappear...

One case where I notice it could be a problem:

1. SDCard menu
2. Pause / resume printing are currently two separate options
3. The best if they would be a single one that changes type of action depending on system state
4. This way you get pretty good stable interface where: if you pause it it changes into resume, and otherwise

If you would have a moment to try it, and see how it behaves in different scenarios.

ayufan:
@mcmatrix @KevinOConnor 

I'm happily running this, and it is significantly better experience than what we have now, and does not create that much chaos by dynamically changing items :) 

This especially when used in a context of https://github.com/KevinOConnor/klipper/pull/4336 actually makes the display significantly better for quick adjustments. Now, you need to navigate very deep to do something useful, and it also gets outdated as system state changes.

mcmatrix:
@ayufan I'll try to find a time to test this approach more deeply.
Let you know.

KevinOConnor:
Thanks. As a high-level comment though, due to time constraints on my side, I don't think I'll be looking to merge display/menu changes to Klipper in the next several months.

-Kevin

KevinOConnor:

-Kevin

ayufan:
This PR is constantly updated to keep with master. It is fully functional.

PR #4363, opened 2021-06-08
Add link in SKR Pro 1.1 section to crystal startup bug

zbrozek:
The SKR Pro 1.1 (and likely others in the BTT family that use the same MCU) have incorrect crystal loading capacitors that make it impossible to enter DFU mode via the BOOT0 and BOOT1 straps. This PR adds a link to the issue in case folks would like to understand and rework their boards to fix that functionality.


KevinOConnor:
Thanks, but commits would need to have a signed-off-by line for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

PR #4364, opened 2021-06-08
virtual_sdcard: add `file_path` and `file_size` to `status`

ayufan:
This provides a comprehensive information if currently we have a file loaded.

Previously:

- we used `print_stats` to know if something is loaded, which is not always correct nor true
- this provides a clear information if a `virtual-sdcard` file is loaded and currently printing
- provides a clear purpose of `v-sd` to track a file progress, `print_stats` to track print progress (which might be coming from manual stream of commands [not yet possible])

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -306,7 +306,9 @@ The following information is available in the
 - `is_active`: Returns True if a print from file is currently active.
 - `progress`: An estimate of the current print progress (based of file
   size and file position).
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  Looks fine to me.  If there are no further comments, I'll look to commit in a few days.

-Kevin

KevinOConnor:
Thanks.

-Kevin

Issue #3895, opened 2021-02-06
pause_resume registers undocumented "CANCEL_PRINT" gcode


yschroeder:
According to the [documentation](https://www.klipper3d.org/G-Codes.html#pause-resume) the `[pause_resume]` registers 3 new gcodes. However in reality it also registers `CANCEL_PRINT`.

This caught me off-guard as I have my own `CANCEL_PRINT` defined, resulting in a weird error message about redefining the `CANCEL_PRINT` gcode.

Relevant line in source:
https://github.com/KevinOConnor/klipper/blob/a7b50b6002d2982c6b426f180a5f6bf9db976e1f/klippy/extras/pause_resume.py#L20

yschroeder:
I think this issue should stay open, as I think it needs fixing.

KevinOConnor:
This should be fixed with #4370.

-Kevin

Issue #3666, opened 2020-12-22
Heaters Stay on after "Move exceeds maximum extrusion"  with hotend melting the part.


bung69:
Hi,  Just had my printer stop due to a "Move exceeds maximum extrusion" error.  - my fault as I forgot to change the nozzle size in the config.  no problem there.  

However the printer just stopped without disabling heaters.  is that expected behaviour?  I assume / hope the heaters will eventually time out?

Thanks

Matt

klipper-gitissuebot:
Hi @bung69,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

Unfortunately, too many people have opened tickets without providing the log. That consumes developer time; time that would be better spent enhancing the software. If this ticket references an event that has occurred while running the software then the Klipper log must be attached to this ticket. Otherwise, this ticket will be automatically closed in a few days.

For information on obtaining the Klipper log file see: https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md

The log can still be attached to this ticket - just add a comment and attach the log to that comment.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

bung69:
[KlipperLog- move exceeds maximum extrusion.txt](https://github.com/KevinOConnor/klipper/files/5733608/KlipperLog-.move.exceeds.maximum.extrusion.txt)

jman1o1:
Just experienced the exact same scenario. Seems like a major safety issue.

monoxide13:
> Just experienced the exact same scenario. Seems like a major safety issue.

Same

KevinOConnor:
Responding to an error like this is typically handled by OctoPrint.  However, I agree Klipper could do a better job now that many users are using moonraker and mainsail/fluidd.  Adding a better default behaviour is something we will look at adding.

-Kevin

dudeofawesome:
+1 from me too. Just had it destroy a build platform due to leaving the heater and bed on with the nozzle oozing PETG onto the bed. Definitely should fail-safe with the heaters turning off in this event.

vladbabii:
Could we have an option to run some gcode on errors that stop the print, with an error code as a parameter? This way we can handle any possible issues in the best way for that printer

Sineos:
[...]

PR #4366, opened 2021-06-08
print_stats: add `cancelled` when `CANCEL_PRINT` used

ayufan:
Before this change, a `CANCEL_PRINT` would set a `print_stats` to `paused`
that would later be workaround-ed with `moonraker`/`fluidd`/`mainsail` to re-define
`CANCEL_PRINT` to transition to `standby` and forcefully unload file.

This adds a proper `canceled` state. This additonally closes a file
from a `virtual_sdcard` context for `canceled`/`error`, as such state is no longer
resumable from this point.

This removes workarounds needed by `fluidd`/`mainsail` as previously they would ask to:
- https://docs.fluidd.xyz/configuration/initial_setup#cancel_print
- https://docs.mainsail.xyz/necessary-configuration#for-pause--resume--cancel-functionality
- https://github.com/Arksine/moonraker/blob/master/moonraker/components/history.py#L191
- manually call `CLEAR_PAUSE` to mark `pause_resume` as not paused before the call to `SDCARD_RESET_FILE`
- manually call `SDCARD_RESET_FILE` to reset `print_stats` to `standby` and close file in `virtual_sdcard` context
- used a workaround to discover `canceled` state by looking at `print_stats` transition from `paused/printing` to `standby`

This removes all the above workarounds, as either generic `CANCEL_PRINT` can be used, or can be extended with `TURN_OFF_HEATERS`.

This connected with https://github.com/KevinOConnor/klipper/pull/4364/files, does provide a clear distinction about `virtual_sdcard` holding information of currently loaded file, `print_state` holding information about the most recent print result.

Related:
- https://github.com/Arksine/moonraker/pull/176

### DIFF
--- a/klippy/extras/pause_resume.py
+++ b/klippy/extras/pause_resume.py
@@ -80,7 +80,9 @@ class PauseResume:
         self.is_paused = self.pause_command_sent = False
     def cmd_CANCEL_PRINT(self, gcmd):
         self.cmd_PAUSE(gcmd)
-        if not self.sd_paused:
+        if self.sd_paused:
+            self.v_sd.do_cancel()
+        else:
             gcmd.respond_info("action:cancel")
         self.cmd_CLEAR_PAUSE(gcmd)
 
--- a/klippy/extras/pause_resume.py
+++ b/klippy/extras/pause_resume.py
@@ -38,12 +38,14 @@ class PauseResume:
         return {
             'is_paused': self.is_paused
         }
+    def is_sd_active(self):
+        return self.v_sd is not None and self.v_sd.is_active()
     def send_pause_command(self):
         # This sends the appropriate pause command from an event.  Note
         # the difference between pause_command_sent and is_paused, the
         # module isn't officially paused until the PAUSE gcode executes.
         if not self.pause_command_sent:
-            if self.v_sd is not None and self.v_sd.is_active():
+            if self.is_sd_active():
                 # Printing from virtual sd, run pause command
                 self.sd_paused = True
                 self.v_sd.do_pause()
@@ -79,8 +81,7 @@ class PauseResume:
     def cmd_CLEAR_PAUSE(self, gcmd):
         self.is_paused = self.pause_command_sent = False
     def cmd_CANCEL_PRINT(self, gcmd):
-        self.cmd_PAUSE(gcmd)
-        if self.sd_paused:
+        if self.is_sd_active() or self.sd_paused:
             self.v_sd.do_cancel()
         else:
             gcmd.respond_info("action:cancel")
--- a/klippy/extras/print_stats.py
+++ b/klippy/extras/print_stats.py
@@ -41,11 +41,15 @@ class PrintStats:
             self._update_filament_usage(curtime)
         if self.state != "error":
             self.state = "paused"
-    def note_error(self, message):
-        self.state = "error"
-        self.error_message = message
     def note_complete(self):
-        self.state = "complete"
+        self._note_finish("complete")
+    def note_error(self, message):
+        self._note_finish("error", message)
+    def note_cancel(self):
+        self._note_finish("canceled")
+    def _note_finish(self, state, error_message = None):
+        self.state = state
+        self.error_message = error_message
         eventtime = self.reactor.monotonic()
         self.total_duration = eventtime - self.print_start_time
         if self.filament_used < 0.0000001:
--- a/klippy/extras/print_stats.py
+++ b/klippy/extras/print_stats.py
@@ -46,7 +46,7 @@ class PrintStats:
     def note_error(self, message):
         self._note_finish("error", message)
     def note_cancel(self):
-        self._note_finish("canceled")
+        self._note_finish("cancelled")
     def _note_finish(self, state, error_message = None):
         self.state = state
         self.error_message = error_message
--- a/klippy/extras/print_stats.py
+++ b/klippy/extras/print_stats.py
@@ -47,7 +47,7 @@ class PrintStats:
         self._note_finish("error", message)
     def note_cancel(self):
         self._note_finish("cancelled")
-    def _note_finish(self, state, error_message = None):
+    def _note_finish(self, state, error_message = ""):
         self.state = state
         self.error_message = error_message
         eventtime = self.reactor.monotonic()
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -98,6 +98,13 @@ class VirtualSD:
         self.must_pause_work = False
         self.work_timer = self.reactor.register_timer(
             self.work_handler, self.reactor.NOW)
+    def do_cancel(self):
+        if self.current_file is not None:
+            self.do_pause()
+            self.current_file.close()
+            self.current_file = None
+            self.print_stats.note_cancel()
+        self.file_position = self.file_size = 0.
     # G-Code commands
     def cmd_error(self, gcmd):
         raise gcmd.error("SD write not supported")
@@ -212,6 +219,7 @@ class VirtualSD:
         gcode_mutex = self.gcode.get_mutex()
         partial_input = ""
         lines = []
+        error_message = None
         while not self.must_pause_work:
             if not lines:
                 # Read more data
@@ -245,7 +253,7 @@ class VirtualSD:
             try:
                 self.gcode.run_script(line)
             except self.gcode.error as e:
-                self.print_stats.note_error(str(e))
+                error_message = str(e)
                 break
             except:
                 logging.exception("virtual_sdcard dispatch")
@@ -265,7 +273,9 @@ class VirtualSD:
         logging.info("Exiting SD card print (position %d)", self.file_position)
         self.work_timer = None
         self.cmd_from_sd = False
-        if self.current_file is not None:
+        if error_message is not None:
+            self.print_stats.note_error(error_message)
+        elif self.current_file is not None:
             self.print_stats.note_pause()
         else:
             self.print_stats.note_complete()


ayufan:
cc @Arksine @cadriel @meteyou

KevinOConnor:
Okay, a change here is fine with me, but I would need feedback from the frontend developers.

Also, if we're changing this area of the code, it's worth mentioning that #3666 and #3895 are also pending.

-Kevin

zellneralex:
Hello Kevin,

for me #3666 is a general problem and should not be mix'ed with a user forced CANCEL_PRINT. At the moment any error during a print using the virtual_sd will not stop the print and/or switch of the heater. E.g. a QGL fail in your PRINT_START will exit your PRINT_START but the print of the virtual_sd card will continue. Maybe there is a need of an "general error handler" while printing from the virtual_sd.

meteyou:
i still have to test it, but from a purely visual point of view, i see the modification as a good idea. it makes it easier to separate user forced cancellations and klipper errors, and it also makes it possible to display them differently/seperate in the gui (reprints after canceling for example)

ayufan:
@zellneralex 

Likely this can be solved by a sane defaults: on error or cancel always do `TURN_OFF_HEATERS` as this is a security feature.

zellneralex:
@ayufan but that solve only part of the problem I would expect the print to error and stop if e.g a QGL fails as this means the gantry is not leveled and has a potential risk of a nozzle crash. That's the point why I suggest a more generally look in an error handling when using the virtual_sd

ayufan:
@zellneralex 

> potential risk of a nozzle crash

We have no clue how to move gantry. What we can do, is to offer a way to allow you to hook on error. This could be achieved with:

```
[print_stats] or [pause_resume] or [something_else]
on_complete_gcode: ...
on_cancel_gcode: ...
on_error_gcode: ...
```

zellneralex:
@ayufan btw the CLEAR_PAUSE is in the macro as the proposed flow is: PAUSE -> CANCEL_PRINT or RESUME. Later that was changed that you can also use CANCEL_PRINT direct in both FE.
The CLEAR_PAUSE simply avoids that message 'Print already paused' if you want to avoud it in both cases https://github.com/KevinOConnor/klipper/blob/master/klippy/extras/pause_resume.py#L91 would need to change to  `if not self.is_paused:
    self.cmd_PAUSE(gcmd)`

ayufan:
@zellneralex Good point. I think `cmd_PAUSE` was retained, to ensure that we set a proper state, but you are right, this can be avoided.

zellneralex:
> @zellneralex
> 
> > potential risk of a nozzle crash
> 
> We have no clue how to move gantry. What we can do, is to offer a way to allow you to hook on error. This could be achieved with:
> 
> ```
> [print_stats] or [pause_resume] or [something_else]
> on_complete_gcode: ...
> on_cancel_gcode: ...
> on_error_gcode: ...
> ```

That's my point I expect klipper to handle it thats the reason I say that should been mixed with the PR here. There are errors where it is simple enough to been reported but there are others that need to at least stop the print. As Kevin mentioned in the past octoprint handles most of them by simply sending a M112. So as more and more users switching away from octo and therfore use the virtual_sd for there printing task's it might be time to get that better sorted

zellneralex:
I will check your changes against the proposed FE macros later today as both FE uses currently macros I did.

ayufan:
> That's my point I expect klipper to handle it thats the reason I say that should been mixed with the PR here

We can solve that, but I would not bundle too many things at once, as they might require more discussion. This PR acknowledges a missing state, and does not truly change status-quo that was present.

I'm happy tho to follow-up, but to keep things easy to discuss I would focus on one thing at a time, it makes it easier to reason about problems :)

ayufan:
> I will check your changes against the proposed FE macros later today as both FE uses currently macros I did.

Thanks. My current macro is:

```ini
[gcode_macro CANCEL_PRINT]
rename_existing: BASE_CANCEL_PRINT
gcode:
    PRESENT_PRINT
    BASE_CANCEL_PRINT

[gcode_macro PRESENT_PRINT]
gcode:
  SAVE_GCODE_STATE NAME=PRESENT_PRINT_state
  M83 ; extruder relative mode
  {% if printer.extruder.temperature > 175 %} G1 E-10 F500 {% endif %} ; Retract a bit
  TURN_OFF_HEATERS ; turn off extruder & heatbed
  M107 ; turn off fan
  {% if printer.toolhead.homed_axes == "xyz" %} G1 X0 Y200 F3000 {% endif %} ; home X axis
  RESTORE_GCODE_STATE NAME=PRESENT_PRINT_state
```

It works fine with latest frontend, even though it does not show `cancelled` as expected. With Moonraker change mentioned, it does track in history cancelled properly. This change should also be backwards compatible for products that use `printing/paused > standy` flow.

zellneralex:
For me it is more important that the work backwards so I will check it against the current proposal in FluiddPi and MainsailOS. We do not want all people coming to the discord asking why there macros do not work anymore. Of coarse I will do also modified version that would fully use your changes

ayufan:
> For me it is more important that the work backwards so I will check it against the current proposal in FluiddPi and MainsailOS

Thank you. I agree with this being important.

cadriel:
I've taken a look, and I think this change is good. It'll remove some unnecessary logic client side, and helps better demarcate these states.

I have not tested this yet - and I would like to prior to merge. I'm not sure what impact a change like this is going to have on the current releases in circulation, and I think we may need to consider in what order, and how we release this - if it does require client updates.

@ayufan Have you tested the clients with it in place, any concerns?

A couple of things;

ayufan:
@cadriel 

I tested:
- Moonraker history plugin breaks without `cancelled`
- Fluidd appears to work fine with `cancelled` and non-cancelled
- I did not yet test Mainsail, but will try
-

ayufan:
I tested the following:

## With patch (with default `CANCEL_PRINT`)

```ini
[gcode_macro CANCEL_PRINT]
rename_existing: BASE_CANCEL_PRINT
gcode:
    BASE_CANCEL_PRINT
```

```
17:03:40 $ SDCARD_PRINT_FILE FILENAME="wait_print.gcode"
17:03:40 File opened:wait_print.gcode Size:450
17:03:40 File selected
17:03:46 $ CANCEL_PRINT
```

Screen Shot 2021-06-09 at 16 46 53
Screen Shot 2021-06-09 at 16 47 02

## With patch with old `CANCEL_PRINT`

```ini
[gcode_macro CANCEL_PRINT]
rename_existing: BASE_CANCEL_PRINT
gcode:
    CLEAR_PAUSE
    SDCARD_RESET_FILE
    BASE_CANCEL_PRINT
```

```
17:07:35 $ SDCARD_PRINT_FILE FILENAME="wait_print.gcode"
17:07:35 File opened:wait_print.gcode Size:450
17:07:35 File selected
17:07:40 $ CANCEL_PRINT
17:07:41 // action:cancel
```

- The `action:cancel` appear here due to `CLEAR_PAUSE`, `SDCARD_RESET_FILE` resetting active virtual sdcard context, and `BASE_CANCEL_PRINT` => this is harmless

Screen Shot 2021-06-09 at 16 49 15
Screen Shot 2021-06-09 at 16 49 20

## Summary

All appears to work fine. The Moonraker patch is required for History component: https://github.com/Arksine/moonraker/blob/master/moonraker/components/history.py#L191

KevinOConnor:
> That's my point I expect klipper to handle it thats the reason I say that shouldn't been mixed with the PR here.

Just to be clear, my goal with mentioning #3666 and #3895 was to raise awareness of two other pending issues.  I'm not suggesting that this PR should fix them.

Cheers,
-Kevin

zellneralex:
Did a few test on mainsail and fluidd
1) no CLEAR_PAUSE AND SDCARD_RESET_FILE -> works status cancel file still shows as loaded
2) current FE macros (CLEAR_PAUSE AND SDCARD_RESET_FILE) -> works status standby file unloaded
3) only CLEAR_PAUSE -> print can not been canceled until you run SDCARD_RESET_FILE manual
4) only SDCARD_RESET_FILE same as 2) but message already paused

used FE macro 
```
[gcode_macro CANCEL_PRINT]
rename_existing: CANCEL_PRINT_BASE
gcode:    
  TURN_OFF_HEATERS
  CLEAR_PAUSE    
  SDCARD_RESET_FILE    
  CANCEL_PRINT_BASE
```

Chase 3 is what worries me

ayufan:
@zellneralex

Do you mean `CLEAR_PAUSE` being run alone? 

I didn't see a usage of `CLEAR_PAUSE` alone. This is based on a fact that `print_stats` is connected with `virtual_sdcard` and cannot be used outside of this context for other purposes.

I would aim to ensure that we can maintain backward compatibility with what is running for majority and a new preferred one.

zellneralex:
Chase 3 means 
```
[gcode_macro CANCEL_PRINT]
rename_existing: CANCEL_PRINT_BASE
gcode:    
  TURN_OFF_HEATERS
  CLEAR_PAUSE    
  CANCEL_PRINT_BASE
``` as said this is used to avoid the message already in PAUSE when using the FE macro out of a paused stare

ayufan:
Did you test latest commit? This should not be needed :)

zellneralex:
Oh no I prepared at 12h ago

zellneralex:
So need to redo it with the last changes

zellneralex:
Already integrated you last changes, started the next test cycle. I will redo all 4 chases

zellneralex:
redone all 4 cases again. I can confirm there are working now.

I will update the Mainsail docu and MainsailOS macros as soon that is merged and @cadriel can than take them from mainsailOS. The change will require no change it both FE

ayufan:
Thank you @zellneralex !

cadriel:
Thank you all, given this isn't breaking at all then I think its good from my side.

KevinOConnor:
Thanks.

-Kevin

PR #4374, opened 2021-06-09
display: Add support for the ST7789V LCD used by the Prusa Mini+

matthewlloyd:
Demo: https://youtu.be/PKduopITGcU

Note: the change in spi.c was needed because the bootloader uses the display, so the SPI interface's pclock is already enabled but it still needs initializing because the config is different. Without this change the display doesn't accept any SPI commands.

### DIFF
--- a/klippy/extras/bus.py
+++ b/klippy/extras/bus.py
@@ -103,13 +103,13 @@ class MCU_SPI:
 
 # Helper to setup an spi bus from settings in a config section
 def MCU_SPI_from_config(config, mode, pin_option="cs_pin",
-                        default_speed=100000, share_type=None):
+                        default_speed=100000, share_type=None, use_cs_pin=True):
     # Determine pin from config
     ppins = config.get_printer().lookup_object("pins")
     cs_pin = config.get(pin_option)
     cs_pin_params = ppins.lookup_pin(cs_pin, share_type=share_type)
     pin = cs_pin_params['pin']
-    if pin == 'None':
+    if pin == 'None' or not use_cs_pin:
         ppins.reset_pin_sharing(cs_pin_params)
         pin = None
     # Load bus parameters
--- a/klippy/extras/display/display.py
+++ b/klippy/extras/display/display.py
@@ -6,18 +6,21 @@
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import logging, os, ast
-from . import hd44780, hd44780_spi, st7920, uc1701, menu
+from . import hd44780, hd44780_spi, st7920, uc1701, menu, st7789v
 
 # Normal time between each screen redraw
 REDRAW_TIME = 0.500
 # Minimum time between screen redraws
 REDRAW_MIN_TIME = 0.100
+# Minimum time between calls to flush() that did not have time to complete
+INCOMPLETE_FLUSH_MIN_TIME = 0.075
 
 LCD_chips = {
     'st7920': st7920.ST7920, 'emulated_st7920': st7920.EmulatedST7920,
     'hd44780': hd44780.HD44780, 'uc1701': uc1701.UC1701,
     'ssd1306': uc1701.SSD1306, 'sh1106': uc1701.SH1106,
-    'hd44780_spi': hd44780_spi.hd44780_spi
+    'hd44780_spi': hd44780_spi.hd44780_spi,
+    'st7789v': st7789v.ST7789V
 }
 
 # Storage of [display_template my_template] config sections
@@ -200,19 +203,22 @@ class PrinterLCD:
             self.redraw_request_pending = False
             self.redraw_time = eventtime + REDRAW_MIN_TIME
         self.lcd_chip.clear()
-        # update menu component
-        if self.menu is not None:
-            ret = self.menu.screen_update_event(eventtime)
-            if ret:
-                self.lcd_chip.flush()
-                return eventtime + REDRAW_TIME
-        # Update normal display
         try:
-            self.show_data_group.show(self, self.display_templates, eventtime)
+            # Render the menu if running
+            menu_rendered = False
+            if self.menu is not None:
+                menu_rendered = self.menu.screen_update_event(eventtime)
+            # Otherwise render the normal display
+            if not menu_rendered:
+                self.show_data_group.show(
+                    self, self.display_templates, eventtime)
         except:
             logging.exception("Error during display screen update")
-        self.lcd_chip.flush()
-        return eventtime + REDRAW_TIME
+        if not self.lcd_chip.flush():
+            # The flush did not have time to complete.
+            return eventtime + INCOMPLETE_FLUSH_MIN_TIME
+        else:
+            return eventtime + REDRAW_TIME
     def request_redraw(self):
         if self.redraw_request_pending:
             return
--- a/klippy/extras/display/hd44780.py
+++ b/klippy/extras/display/hd44780.py
@@ -87,6 +87,7 @@ class HD44780:
                 self.send([fb_id + chip_pos])
                 self.send(new_data[pos:pos+count], is_data=True)
             old_data[:] = new_data
+        return True
     def init(self):
         curtime = self.printer.get_reactor().monotonic()
         print_time = self.mcu.estimated_print_time(curtime)
--- a/klippy/extras/display/hd44780_spi.py
+++ b/klippy/extras/display/hd44780_spi.py
@@ -78,6 +78,7 @@ class hd44780_spi:
                 self.send([fb_id + chip_pos])
                 self.send(new_data[pos:pos+count], is_data=True)
             old_data[:] = new_data
+        return True
     def init(self):
         curtime = self.printer.get_reactor().monotonic()
         print_time = self.mcu.estimated_print_time(curtime)
--- a/klippy/extras/display/logo.py
+++ b/klippy/extras/display/logo.py
@@ -1,0 +1,18 @@
+# Utility to read and resize the Klipper logo.
+#
+# Copyright (C) 2021  Matthew Lloyd <github@matthewlloyd.net>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import os
+from PIL import Image
+
+LOGO_IMAGE = None
+
+def make_logo(width):
+    global LOGO_IMAGE
+    if LOGO_IMAGE is None:
+        LOGO_IMAGE = Image.open(
+            os.path.join(os.path.dirname(__file__),
+                         '../../../docs/img/klipper-logo.png'))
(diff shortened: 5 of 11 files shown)


KevinOConnor:
Thanks.  Looks like there's overlap with this PR and #3724 .  @pienjo - do you have any comments on this?

I think the spi changes should be moved to a separate commit (having that separate makes it easier to identify issues with `git bisect` and revert if necessary).  Also, if we've got a bootloader leaving a peripheral in an unusual state, then I think it would be preferable to address it by resetting APBxENR at startup and perform explicit device resets (as was done in 689c2f9b).

I'm not sure what the implications are for adding dependencies to the build.  (On some platforms it can dramatically increase the Klipper install time.)  Can you describe what the two dependencies are being used for?

Also, can you give a high-level description on the changes to display.py?

-Kevin

matthewlloyd:
> I think the spi changes should be moved to a separate commit (having that separate makes it easier to identify issues with `git bisect` and revert if necessary). Also, if we've got a bootloader leaving a peripheral in an unusual state, then I think it would be preferable to address it by resetting APBxENR at startup and perform explicit device resets (as was done in [689c2f9](https://github.com/KevinOConnor/klipper/commit/689c2f9baf835b55cdbb68d78b0913d8ca6df317)).

Done.

> I'm not sure what the implications are for adding dependencies to the build. (On some platforms it can dramatically increase the Klipper install time.) Can you describe what the two dependencies are being used for?

My first pass attempts at this tried to avoid using Pillow or NumPy, but they were horrendously slow, causing loss of MCU sync, slow display refresh, and failed prints. Juggling 240x320 framebuffers is just too slow on a Pi with regular Python lists/strings/bytearrays. Tried my best to avoid the deps but eventually accepted their inevitability. The code in its current form works very quickly and reliably - I have had no display-related crashes so far in my week or so of use and ~20hrs print time.

Used for:

* Pillow: Fast image manipulation. RGB bitmap to/from bytes, image paste, crop, diff, and for the logo, PNG import and high quality resampling.
* NumPy: Much faster (orders of magnitude) than Python loops for large array manipulation due to vectorization and low-level optimization. Here used for fast bit twiddling (RGB888 to RGB565) and array summarization (which cols contain changed pixels).

Time to install Klippy reqs in a fresh venv on my Pi 4B:

```
$ time pip2 install --no-cache-dir -r ~/klipper/scripts/klippy-requirements.txt
Successfully installed Jinja2-2.11.3 MarkupSafe-1.1.1 aenum-3.1.0 cffi-1.12.2 greenlet-0.4.15 pycparser-2.20 pyserial-3.4 python-can-3.3.4 typing-3.10.0.0 wrapt-1.12.1

real    0m30.280s
user    0m24.175s
sys     0m2.983s
```

With Pillow and numpy:

```
Successfully installed Jinja2-2.11.3 MarkupSafe-1.1.1 Pillow-5.4.1 aenum-3.1.0 cffi-1.12.2 greenlet-0.4.15 numpy-1.16.2 pycparser-2.20 pyserial-3.4 python-can-3.3.4 typing-3.10.0.0 wrapt-1.12.1

real    4m48.738s
user    7m53.829s
sys     0m47.932s
```

Quite an increase, almost all of it due to NumPy. Some options:

- Accept it. The input shaper calibration script which many (most?) users will want to run requires NumPy anyway and it might prove useful for other things.
- Remove the dependencies from klippy-requirements.txt, and instead put a note in the Prusa Mini+ and Buddy configs telling users to install them manually. (Assuming Klippy will run and pass tests with an unused module missing its dependencies.)

> Also, can you give a high-level description on the changes to display.py?

* Allow a display's flush() method to signal to display.py that it needs more time to complete. If so we will call it again after 75ms instead of the usual 500ms. Splitting up the flush prevents it stalling the MCU, and calling it back more quickly makes display refresh much smoother.
* Small rearrangement of code around drawing the menu, so there is only one call to flush() to worry about. Behavior is same as before.

pienjo:
> Thanks. Looks like there's overlap with this PR and #3724 . @pienjo - do you have any comments on this?

There is definate overlap here - it is the same display after all, only the attachment interface is different, but this is relatively easy to address. The inherent problems are the same though: There is a lot of data that needs to be sent to the display, which ultimately needs to go through the microcontroller, via the serial connection between klippy and the controller - and since display.py was written for monochrome displays it doesn't really make effective use of all those colours.

This PR took a different approach than I did (I was carefully avoiding pulling in external dependencies). My attempt eventually got stuck trying to reduce the overhad on the system by compressing the data between klippy and controller. This worked fine, but that created a bottleneck that I haven't been able to solve.  I'm not sure the bytearrays are the largest culprit though.

matthewlloyd:
> trying to reduce the overhead on the system by compressing the data between klippy and controller.

We followed a similar path - I also tried RLE compression. For my specific board, it was actually slower. I think it may depend on the specifics of the Klippy/MCU communication interface: Prusa Buddy uses the STM32's USB pins, and the configured UART baud rate is effectively ignored because the packets are transmitted at very high USB speeds directly on the USB bus. The serial transfer time is negligible, and the bottlenecks are preparing and sequencing the packets on the Klippy side (hence Pillow + NumPy) and the actual SPI transmission on the MCU side (fast enough as it is, but would be good to use DMA eventually). The overhead of RLE compression/decompression made it actually slower than just sending the raw data over the fast bus. For other boards with regular baud-rate UARTs, it seems to be a different story (Could connect the LCD to the Pi directly instead? This option not so good for Prusa since it breaks the stock firmware for reverting or dual boot.)

The dependency on NumPy can be made run-time, I found this existing code elsewhere and think it's suitable: https://github.com/KevinOConnor/klipper/blob/e520fb78782f7148123fd26dfc9511313d016111/klippy/extras/shaper_calibrate.py#L177

Alternatively I can write a small C module to handle the bit twiddling and array summarization.

I'm hoping to add full color support to the existing display templates at some point. It's useful for highlighting which heaters are turned on etc. I also hope to add some predefined templates that take advantage of larger screens, and allow the menu and some status info to be displayed simultaneously.

pienjo:
> We followed a similar path - I also tried RLE compression. For my specific board, it was actually slower. I think it may depend on the specifics of the Klippy/MCU communication interface: Prusa Buddy uses the STM32's USB pins, and the configured UART baud rate is effectively ignored because the packets are transmitted at very high USB speeds directly on the USB bus. 

Interesting; I used a blue pill for development (as it made updating the firmware so much easier), which also uses the STM's USB interface. Not using a compression scheme resulted in a *lot* of messages, causing serious problems because it interfered with other parts of the system. In fact, your pull request all but admits that: it tries to throttle the display updates for this very reason.

I'm not trying to do a "my PR is better than yours" here - far from it. I got stuck since it turned out that the optimizations backfired once I moved back to the meager OrangePi I was intending to use (which is admittedly quite a bit slower than a raspi 4B). It doesn't help that the printer has other unrelated issues (I haven't been able to complete a single print for the last few months due to EMI issues that I can't seem to tackle), which is a bit demotivating :(

matthewlloyd:
> Not using a compression scheme resulted in a _lot_ of messages, causing serious problems because it interfered with other parts of the system. In fact, your pull request all but admits that: it tries to throttle the display updates for this very reason.

It's not so much that the messages overload or interfere with the system - the MCU seems to be able to handle a high volume of display-related messages. That's not too surprising, because based on my limited testing, the actual transmission of messages to the MCU is very fast (USB speeds), and the transmission of the display data by the MCU over the SPI bus to the display is also very fast (3MHz baud). Python can send these messages fast enough to keep up. The problem is whether Klippy's event loop gets to run when it needs to.

The issues I encountered were when the `flush` method took too long to return. Klippy runs on an event loop, and the loop needs to cycle frequently to handle clock sync, dispatching timers, and so on. If I choke the event loop by preventing it from running with a long-running `flush`, the MCU shuts down with a "Timer too close" error. This happens when Klippy sends a message to the MCU asking to schedule a timer at a time that is in the past. It happens even if you just put a time.sleep() in the `flush` method, with no display updates at all.

To illustrate this another way, I tried removing my code that forces the `flush` method to limit the display update batch size (which I added so the event loop gets a chance to run at least once every 50ms). Doing this makes the MCU reliably shut down with a "Timer too close" error. However, it still shuts down in the same way even if you remove the code in `flush` that sends the actual display data messages to the MCU.

So the problem I ended up solving here isn't related to the volume or size of the messages - it's how long it takes for Python to compute the deltas and encode the updates in the `flush` method. That's why I ended up using PIL and NumPy.

KevinOConnor:
Thanks. As high-level feedback, installing numpy would notably alter the installation time of Klipper.  It's certainly an option, but I'd guess that is unlikely to happen in 2021.

Also, separately, due to time constraints on my side, I don't think I'll be looking to merge display/menu changes to Klipper in the next several months.

-Kevin

matthewlloyd:
No problem - Klipper's rise to popularity has created more pull requests than any one developer could handle.

Let me know in a few months when you're ready to review display/menu changes, and I'll modify this PR with a note for users to install PIL/NumPy manually, rather than including it in requirements.txt. 

Meanwhile, for others who stumble on this PR looking for Prusa Mini support, here is a ready-to-use fork: https://github.com/matthewlloyd/klipper

Matthew

KevinOConnor:
Thanks for working on this.  Unfortunately, due to time constraints on my side it is unlikely I'll get to review this in the near future.  The menu system in general is an area where I think we'll need to find volunteers to help review and lead development going forward.

-Kevin

KevinOConnor:
Thanks for working on this.  Unfortunately, due to time constraints on my side it is unlikely I'll get to review this in the near future.  The menu system in general is an area where I think we'll need to find volunteers to help review and lead development going forward.

-Kevin

christofkern:
Hey @matthewlloyd, I was using your Branch of Klipper for a while now for my Prusa Mini, but unfortunately I had to re-setup my raspberry pi (had a legacy version for the camera, which I switched out now) and instead of saving my octoprint-installation I was just thinking, I would setup everything again from new (cause I had some outdated plugins). But I just noticed that your repository  https://github.com/matthewlloyd/klipper does not exist anymore... and when trying to use the normal klipper installation I get the error, "Choice 'st7789v' for option 'lcd_type' in section 'display' is not a valid choice". Can you help me out, or am I stuck and have to resetup the standard prusa firmware (or llama or smth)?
Thanks a lot
-Christof

Pravv:
@christofkern As long as you don't mind lack of LCD support, master branch of Klipper works perfectly fine out of the box as long as your config doesn't have `[display]` section. 
Alternatively you could use my fork, that merges this PR into main Klipper repo master branch. I tested it for a couple of hours now  and it seems to work just fine. 

Thankfully not much work had to be done, as i'm not nearly knowledgeable enough to recreate any of this on my own, if need would arise.

PR #4377, opened 2021-06-10
bltouch: Throw error if z_offset is not positive

Yoyolick:
I installed klipper for the first time yesterday and struggled with the error shown below:
before
Eventually, someone pointed out that my z_offset was negative when it should be positive, which fixed my problem, but it would be nice for any newcomers who encounter this to immediately know the problem as opposed to look around for hours as I did.

This is a very small change, it just adds the "above=0" argument to the getconfig request for the z_offset in the bltouch.py file. With this change, when you attempt to use a negative z_offset you are given this error message as opposed to a unrelated one:
after

Signed-off-by: Ryan Zmuda ryanzmuda@gmail.com

### DIFF
--- a/klippy/extras/bltouch.py
+++ b/klippy/extras/bltouch.py
@@ -28,7 +28,7 @@ class BLTouchEndstopWrapper:
         self.printer = config.get_printer()
         self.printer.register_event_handler("klippy:connect",
                                             self.handle_connect)
-        self.position_endstop = config.getfloat('z_offset')
+        self.position_endstop = config.getfloat('z_offset', above=0)
         self.stow_on_each_sample = config.getboolean('stow_on_each_sample',
                                                      True)
         self.probe_touch_mode = config.getboolean('probe_with_touch_mode',


KevinOConnor:
Thanks.  I'm not sure the bltouch z_offset is always above zero for all printers.  (It certainly is for all normal printers, but I couldn't say if someone has a special setup with probe points outside the normal bed plane.)  Maybe @FanDjango would know.

-Kevin

Yoyolick:
Hmm, you raise a good point. But the fact that the first error is shown with a negative value means there is no case in which people who have points outside the bed can configure it that way currently without getting this error anyways right? Not too sure but just thought I would throw this PR out there, do with it as you will 😃 

-Ryan

FanDjango:
> Thanks. I'm not sure the bltouch z_offset is always above zero for all printers. (It certainly is for all normal printers, but I couldn't say if someone has a special setup with probe points outside the normal bed plane.)

I know of none.

>Eventually, someone pointed out that my z_offset was negative when it should be positive, which fixed my problem, but it would be nice for any newcomers who encounter this to immediately know the problem as opposed to look around for hours as I did.

Why does a negative z-offset produce that error? Probes (other types) can have negative z-offsets. What were the ..._max and ..._min values etc.?

Yoyolick:
> Why does a negative z-offset produce that error? 

This is likely the thing that needs figured out as I don't know for sure, somebody in a discord who was helping me troubleshoot mentioned this after I attached my cfg

> What were the ..._max and ..._min values etc.?

Attached below is my .cfg, as it is now it works completely fine, but if you add a negative before the z offset it reproduces the error shown in my very first screenshot.

[printer.txt](https://github.com/KevinOConnor/klipper/files/6656765/printer.txt)

Extra info on my setup:

- ender 3 v1
- SKR 1.3
- tmc2208's
- genuine bltouch 3.1
- dotbit bmg extruder

FanDjango:
Ok, I understand. And thx for the config.

Well, only a quick thought, but one would need to look at stepper code, bltouch code and where from is "position_endstop" inferred. As there is none defined in this case for Z.

KevinOConnor:
> But the fact that the first error is shown with a negative value means there is no case in which people who have points outside the bed can configure it that way currently without getting this error anyways right? 

As far as I know, if the position_min is less than the z_offset, then it would be possible.  That is, z_offset effectively gets copied into position_endstop, and then position_endstop is verified to be within position_min to position_max.

There's a merge conflict with this branch, but I went ahead and committed the change - commit 6c8d1dda.

-Kevin

PR #4378, opened 2021-06-11
config: Add generic-bigtreetech-skr-cr6-v1.0.cfg

KoeRt888:
Adds SKR CR6 Config with support for loadcell tare

Signed-off-by: Carlo Hoffmann <KoeRtGit@outlook.com>


KevinOConnor:
Thanks.  In general, the config looks fine.  However, it seems the comments and config sections are geared towards a particular printer - the "generic" config files are intended to describe the board, so it doesn't make sense to add peripherals that may not be present.  See: https://www.klipper3d.org/Example_Configs.html

-Kevin

KoeRt888:
I see your point of keeping it a base configuration since the board could be used for a different printer. I updated the PR accordingly.

KevinOConnor:
Thanks.  I committed this change and made some minor changes on top (commit e520fb78).

-Kevin

PR #4379, opened 2021-06-11
config: Create printer-monoprice-mini-delta-2018.cfg

boilerbots:
Tuned for production 1/16 step hardware
Tuned pressure advance with two values.
Increased max_velocity and max_accel.
Configure fan as a heater_fan to keep heatbreak cool.
Configure probing to make 3 measurements at a slower speed.
Simplify START_PRINT to be compatible with Prusa Slicer gcode.


KevinOConnor:
Thanks.  In general the config looks good to me.  However, I noticed it has a `pressure_advance` value defined.  (As well as changes to `square_corner_velocity`, `max_accel_to_decel`, and retraction settings.)  That setting is specific to the filament and not the printer so should not be specified in a default config file - see: https://www.klipper3d.org/Example_Configs.html

-Kevin

boilerbots:
I will re-submit but I found a problem with the config, several technical errors.

KevinOConnor:
Thanks.  How does this differ from the existing `printer-monoprice-mini-delta-2017.cfg` config file?

Otherwise, this change adds a "printer" config file and those files should always end in a year (see https://www.klipper3d.org/Example_Configs.html ).  Also, it's not valid to call DELTA_CALIBRATE from a START_PRINT macro, as DELTA_CALIBRATE doesn't do anything unless `SAVE_CONFIG` is also run (which in turn does a restart, and thus can not be scripted).

-Kevin

PR #4381, opened 2021-06-12
spi_flash: Make SD updates more tolerant of flaky boards

jschuh:
* Unconditionally deletes firmware upload image after reset
* Adds command line switch to ignore validation failure

### DIFF
--- a/klippy/msgproto.py
+++ b/klippy/msgproto.py
@@ -408,6 +408,8 @@ class MessageParser:
         except Exception as e:
             logging.exception("process_identify error")
             self._error("Error during identify: %s", str(e))
+    def get_raw_data_dictionary(self):
+        return self.raw_identify_data
     def get_version_info(self):
         return self.version, self.build_versions
     def get_messages(self):
--- a/scripts/spi_flash/spi_flash.py
+++ b/scripts/spi_flash/spi_flash.py
@@ -788,6 +788,7 @@ class MCUConnection:
         self.connect_completion = None
         self.connected = False
         self.enumerations = {}
+        self.raw_dictionary = None
 
     def connect(self):
         output("Connecting to MCU..")
@@ -814,6 +815,7 @@ class MCUConnection:
                 "MCU Type mismatch: Build MCU = %s, Connected MCU = %s"
                 % (build_mcu_type, mcu_type))
         self.enumerations = msgparser.get_enumerations()
+        self.raw_dictionary = msgparser.get_raw_data_dictionary()
 
     def _do_serial_connect(self, eventtime):
         endtime = eventtime + 60.
@@ -949,29 +951,56 @@ class MCUConnection:
             % (fw_path, sd_size, sd_chksm))
         return sd_chksm
 
-    def verify_flash(self, req_chksm):
+    def verify_flash(self, req_chksm, old_dictionary):
         output("Verifying Flash...")
-        cur_fw_sha = hashlib.sha1()
-        cur_fw_path = self.board_config.get('current_firmware_path',
-                                            "FIRMWARE.CUR")
-        try:
-            with self.fatfs.open_file(cur_fw_path, 'r') as sd_f:
-                while True:
-                    buf = sd_f.read(4096)
-                    if not buf:
-                        break
-                    cur_fw_sha.update(buf)
-        except Exception:
-            msg = "Failed to read file %s" % (cur_fw_path,)
-            logging.exception(msg)
-            raise SPIFlashError(msg)
-        cur_fw_chksm = cur_fw_sha.hexdigest().upper()
-        if req_chksm == cur_fw_chksm:
-            output_line("Done")
-            output_line("Firmware Flash Successful")
+        validation_passed = False
+        msgparser = self._serial.get_msgparser()
+        cur_dictionary = msgparser.get_raw_data_dictionary()
+        # Check that the version changed
+        if cur_dictionary != old_dictionary:
+            output("Version updated...")
+            validation_passed = True
         else:
-            raise SPIFlashError("Checksum Mismatch: Got '%s', expected '%s'"
-                                % (cur_fw_chksm, req_chksm))
+            output("Version unchanged...")
+        # If the version didn't change, look for current firmware to checksum
+        cur_fw_sha = None
+        if not validation_passed:
+            cur_fw_path = self.board_config.get('current_firmware_path',
+                                                "FIRMWARE.CUR")
+            try:
+                with self.fatfs.open_file(cur_fw_path, 'r') as sd_f:
+                    cur_fw_sha = hashlib.sha1()
+                    while True:
+                        buf = sd_f.read(4096)
+                        if not buf:
+                            break
+                        cur_fw_sha.update(buf)
+            except Exception:
+                msg = "Failed to read file %s" % (cur_fw_path,)
+                logging.debug(msg)
+                output("Checksum skipped...")
+            if cur_fw_sha is not None:
+                cur_fw_chksm = cur_fw_sha.hexdigest().upper()
+                if req_chksm == cur_fw_chksm:
+                    validation_passed = True
+                    output("Checksum matched...")
+                else:
+                    raise SPIFlashError("Checksum Mismatch: Got '%s', "
+                                        "expected '%s'"
+                                        % (cur_fw_chksm, req_chksm))
+        if not validation_passed:
+            raise SPIFlashError("Validation failure.")
+        output_line("Done")
+        # Remove firmware file if MCU bootloader failed to rename.
+        if cur_fw_sha is None:
+            try:
+                fw_path = self.board_config.get('firmware_path', "firmware.bin")
+                self.fatfs.remove_item(fw_path)
+                output_line("Found and deleted %s after reset" % (fw_path,))
+            except Exception:
+                pass
+        output_line("Firmware Flash Successful")
+        output_line("Current Firmware: %s" % (msgparser.get_version_info()[0],))
 
 class SPIFlash:
     def __init__(self, args):
@@ -988,6 +1017,7 @@ class SPIFlash:
         self.firmware_checksum = None
         self.task_complete = False
         self.need_upload = True
+        self.old_dictionary = None
 
     def _wait_for_reconnect(self):
         output("Waiting for device to reconnect...")
@@ -1022,6 +1052,7 @@ class SPIFlash:
         # Reconnect and upload
         if not self.mcu_conn.connected:
             self.mcu_conn.connect()
+        self.old_dictionary = self.mcu_conn.raw_dictionary
         self.mcu_conn.configure_mcu(printfunc=output_line)
         self.firmware_checksum = self.mcu_conn.sdcard_upload()
         self.mcu_conn.reset()
@@ -1031,7 +1062,7 @@ class SPIFlash:
         # Reconnect and verify
         self.mcu_conn.connect()
         self.mcu_conn.configure_mcu()
-        self.mcu_conn.verify_flash(self.firmware_checksum)
+        self.mcu_conn.verify_flash(self.firmware_checksum, self.old_dictionary)
         self.mcu_conn.reset()
         self.task_complete = True
 


jschuh:
Just for some more context, I have several  btt-skr-mini-e3-v1.2 boards, and it's reasonably well known that they don't reliably rename firmware.bin to firmware.cur after a successful firmware update from SD. This change just guarantees that the firmware flash file gets cleaned up (so it doesn't keep trying to flash on restart) and gives the option of ignoring the missing validation file.

Adding @Arksine, since he wrote the SD flash code.

Arksine:
Hi, sorry for missing this before.  FWIW, I have an SKR Mini e3 v1.2 and I have never seen it do this.  Perhaps its due to a bug in the bootloader that was fixed before mine shipped?

I'm ok with adding an option to ignore the file, but I'm not sure that we want to delete firmware.bin.  On most boards its presence is an additional indicator that something went wrong with the flash.

FWIW, there may be a better approach to validating the flash that doesn't involve hashing firmware.cur.   When Klipper builds the firmware it creates a version string.  It should be possible to parse it out from the build, then check it against the version stored on the MCU after we reconnect.

jschuh:
No worries about the delay. I wasn't in a hurry and figured I'd ping again if I didn't get a reply soon. And oddly enough, I have three printers with SKR Mini e3 v1.2 boards (all purchased at different times) that exhibit this behavior to some degree. That said, they were purchased in the first six months of its release, so maybe they fixed the issue in later stock.

Either way, using the reported version string sure sounds like a good idea, but I have a few questions:
- Seems the best way to grab the compiled MCU version is from `out/klipper.dict` (which would need to be command line configurable)?
- Looks like I can compare that against the version returned by `MCUConnection._serial.msgparser.get_version_info()`?
- My instinct would be to remove the checksum verification as part of this change, since we would already be verifying that the MCU is running and reporting the correct version, but I defer to your call on that.
- Would it be reasonable to delete the file firmware.bin if the version string check succeeds, but still echo a warning back to the user (or maybe add that behavior as an option)?

Arksine:
> Seems the best way to grab the compiled MCU version is from out/klipper.dict (which would need to be command line configurable)?

Yes, its stored in the `version` key of `klipper.dict`. What I would recommend is to provide a command line option to specify the path to it, defaulting to the `klipper.dict` relative to `klipper.bin`.  Its possible that a user want to flash a prebuilt binary and has no klipper.dict, in that scenario we should fall back to validating the hash.

> Looks like I can compare that against the version returned by MCUConnection._serial.msgparser.get_version_info()?

Yes, something like this could be done in `verify_flash()`if klipper.dict was available:
```python
if self.klipper_dict_version is not None:
  msgparser = self._serial.get_msgparser()
  cur_version = msgparser.get_version_info()[0]
  if self.klipper_dict_version != cur_version:
     raise SPIFlashError("Version Mismatch: Got '%s', expected '%s'"
                         % (cur_version, self.klipper_dict_version))
  output_line("Firmware Flash Successful")
else:
   # Validate hash of .CUR file...
```

> My instinct would be to remove the checksum verification as part of this change, since we would already be verifying that the MCU is running and reporting the correct version, but I defer to your call on that.

As mentioned above, I think we should keep that as a fallback in the event that `klipper.dict` is not found.

>  Would it be reasonable to delete the file firmware.bin if the version string check succeeds, but still echo a warning back to the user (or maybe add that behavior as an option)?

If the firmware version is validated I do think its reasonable to remove the `.bin` file if it exists.  I don't believe that we need to worry about an option here.

KevinOConnor:
Just some random comments:
1. If `firmware.cur` exists, is there any harm in validating it?  Seems like a useful check with little downside.
2. Note that the version string wont encode any options the user may have changed in "make menuconfig".
3. In most cases, it should be possible to check if the mcu version (or mcu constants) change between pre update and post update.  That is, if version/constants change then almost certainly a software update completed successfully.

-Kevin

Arksine:
> If firmware.cur exists, is there any harm in validating it? Seems like a useful check with little downside.

Agreed.  There is certainly no harm in keeping it and it should generally be a reliable check.  Its unfortunate that BTT has a habit of shipping broken bootloaders.

> Note that the version string wont encode any options the user may have changed in "make menuconfig".

Ah, good point.  Also it might be possible that the user is simply flashing the same build to test that `flash-sdcard.sh` is working correctly.  Would you be opposed to adding the build time to the version string?  That should guarantee a unique build. 

Somewhat related to this, I am currently working on adding "beta" and "stable" channels to Moonraker's update manager, this involves packaging and deploying Moonraker and Klipper in a zip file.  I will need to submit a PR that makes a small change to `build_commands.py` that falls back on the `klippy/.version` if `git describe` fails.   If you don't think it would cause an I issue I could append the build time to the version string in that PR.

If you would prefer not to change the version string then I agree that checking for a delta in the constants seems like the appropriate solution.

KevinOConnor:
> Would you be opposed to adding the build time to the version string? That should guarantee a unique build.

I recently removed the build time from "pristine" builds (commit 947ffe70).  The main issue with adding the timestamp is that it results in every build being slightly different - not just the version string, but the layout of the binary itself (eg, different code and data locations).  With the latest code, most users are likely to deploy identical binaries, which helps in troubleshooting.

Ironically, I made the above change because of flaky bootloaders - being able to deploy identical binaries helps confirm if an issue is with a binary or with the bootloader.

FWIW, I suspect the vast majority of updates will be to a new version.

Cheers,
-Kevin

jschuh:
Okay, I think I incorporated all the feedback, and the last commit message should be the accurate one for a commit squash. Here are the changes:
- Accepts an optional klipper.dict (the shell script tries a default) to compare its version string against the running firmware version.
- Deletes the firmware.bin image on the SD card if still present and the version strings matched.
- Always attempts the checksum comparison, but ignores it if firmware.cur is missing and the version strings already matched.

Please let me know if there's anything else you'd like me to change.

Regarding making the version string more unique, what about something like incorporating a hash of the build config? You could even filter it first, to e.g. sort lines and trim whitespace and comments to prevent non-impacting changes from resulting in a different hash. (That stated, I'm not too familiar with the build system, so there may be an obvious reason this is a bad idea.)

jschuh:
Apologies, I fixed a few nits after my earlier comment, and then had to fix a fix. 🤷‍♂️ Really though, it should be good to take a look at now.

Arksine:
> I recently removed the build time from "pristine" builds (commit 947ffe7). The main issue with adding the timestamp is that it results in every build being slightly different - not just the version string, but the layout of the binary itself (eg, different code and data locations). With the latest code, most users are likely to deploy identical binaries, which helps in troubleshooting.

Ah, this is what confused me earlier, I missed this change.  I had recalled that the date was part of the version string, when you pointed out that it wasn't there I assumed that I was just mistaken.  Your reasoning is sound.

@jschuh With Kevin's feedback I think that the conclusion is that we don't need parse `klipper.dict` or add any additional command line options.  Instead we need to fetch the MCU constants on the first connection, then compare them on reconnect if the checksum verification on FIRMWARE.CUR fails.  If the MCU constants have changed then we can reliably assume that the update succeeded and remove `firmware.bin` if it exists.  Otherwise we raise an exception that no firmware changes were detected.

jschuh:
Oh, sorry, I misread the third bullet in Kevin's first comment yesterday. But my question is if it hurts to just always try all three validation checks and error out only in both the absence on a hard failure and if no check returns a success?

What I'm thinking is:
1. Always check if the version string changed between restarting the MCU
    - success: version string change
    - soft fail: version didn't change
2. If firmware.dict is present, check if the version strings match
    - success version string match
    - soft fail: missing firmware.dict
    - hard fail: expected version mismatch
3. If firmware.cur is present, run the checksum comparison
    - success: checksum match
    - soft fail: missing firmware.cur
    - hard fail: checksum mismatch
4. Delete firmware.bin when 1 and/or 2 succeeds, but 3 soft fails.

To explain my preference for keeping the dict version comparison, I've personally reflashed MCUs with the same version on numerous occasions, because with these cheap boards reflashing is an easy way to reduce variables when troubleshooting. And since I already have the dictionary check working reliably, it doesn't seem like it hurts to keep it as an extra validation step to cover this scenario.

jschuh:
I went ahead and implemented what I described above, since it was easy to just add the check for a version change to the PR. Let me know if that works, or if you don't think there's a good enough reason to keep the firmware.dict part and I should just remove it entirely.

Arksine:
I think its close, however as Kevin mentioned it probably isn't enough to check the version string since the build date is no longer included.  The msgparser has a `get_constants()` method that can be used to retrieve MCU Constants.  By comparing a tuple of the version string and the constants we can identify a change to the menuconfig when the user hasn't pulled a new update.

With regard to klipper.dict, I'm not sure it buys us anything since we are comparing the actual MCU versions.  It may be an opportunity for user error.  For example,  the scenario where  `old_version != new_version` and `new_verison != dict_version` would suggest to me that the flash was successful, however the user supplied a stale or incorrect dict file.

KevinOConnor:
FWIW, we can add:
```diff
--- a/klippy/msgproto.py
+++ b/klippy/msgproto.py
@@ -408,6 +408,8 @@ class MessageParser:
         except Exception as e:
             logging.exception("process_identify error")
             self._error("Error during identify: %s", str(e))
+    def get_raw_data_dictionary(self):
+        return self.raw_identify_data
     def get_version_info(self):
         return self.version, self.build_versions
     def get_messages(self):
```
to msgproto.py.  That's probably a better way to check for a change between two sessions, as it encodes the full data dictionary in a single string.

-Kevin

jschuh:
Here's another round using the raw dictionary, as Kevin suggested. I've also made it so that any of the following checks is treated as an _immediate validation success_ (in order):
1. Raw dictionary changed
2. Checksum matched
3. Raw dictionary matched

I'm also unconditionally deleting `firmware.bin` if validation succeeded but there was no checksum match. 

Now I guess I'm going to make a last ditch case for keeping the raw dictionary check in as a last fallback.

First, I think I've sufficiently mitigated the potential for user error from the explicit dictionary matching. It's now used only when the version didn't change and the checksum was unavailable, so it applies only to flaky bootloader scenarios. It's also now clearly labeled as a fallback in the help text, and a default dictionary is not provided unless the shell script is also using the default klipper.bin path.

As for why I see this as useful, I cannot count the number of times in the last few hours that I've had to walk into the other room and manually power cycle the printer I was testing this on due to an aborted SD flash. Really, the whole reason I authored a fix was to not have to do that. And yes, without the dictionary comparison fallback I won't have issues so long as I don't accidentally reflash with the same version while the bootloader is simultaneously flaking on renaming `firmware.bin`. However, I also feel like reflashing with the same version shouldn't be treated as an error if it's reasonably possible to avoid it (particularly not an error that often requires manual power cycling).

Anyway, hopefully that was a convincing case. If not, I can remove the raw dictionary matching from this PR (although I will likely carry a patch locally on my flaky boards).

Arksine:
It looks good to me.  I have no issue with the `klipper.dict` check.

Are you able to test these changes on boards with bootloaders that work correctly?  If not I'll try to do so.

jschuh:
Yup, I have an SKR Mini e3 v2.0 that's never been flaky, and it flashes just fine with this change. Although the weird thing about the flaky board I was primarily testing with is that it reliably renames `firmware.bin` on every other successful flash, so it was good for testing the correct behavior exactly half the time.

I also made sure to test all the scenarios I could on both the v1.2 and v2.0. On the v1.2 I was able to verify: 1) version change; 2) checksum validation; 3) dictionary fallback; and 4) failure (1 and 2 didn't pass, and I blocked 3 by not supplying a dictionary). On the v2.0 board I was only able to verify 1 and 2, but didn't go to the extent of forcing an artificial failure to test 3 and 4.

KevinOConnor:
Thanks.  It looks fine to me as well.  I think it would help though, if this could be broken up into two separate commits - one commit that changes the success check to include dictionary comparison, and one commit that adds support for a user supplied klipper.dict file.  This is useful because if we run into an issue, it's easier to manage `git bisect` and `git revert` if feature additions are separate from process changes.

Also, it seems odd to me that if a user specifies a klipper.dict file that the code doesn't always check it.  Perhaps always run that check if that option is provided?

-Kevin

jschuh:
Happy to split it into two patches, where this one adds only the check for the MCU reported dictionary changing after restart. I expect I can do that later today.

For the second patch that compares against a user supplied dictionary, I originally required a positive match when a dictionary is provided, but switched it to being only a fallback after @Arksine raised concerns about user error. That stated, in the same change I also made the shell script wrapper much less aggressive about providing a default dictionary. Now the shell script uses `out/klipper.dict` only when the default `out/klipper.bin` is used, otherwise there is no default dictionary. Perhaps that's enough to mitigate the user error concern?

jschuh:
Okay, I made the changes so that this patch now adds only the dictionary change comparison and removes the firmware.cur file on verifying a successful flash. I'm putting up a new PR that adds the dictionary comparison as specified by @KevinOConnor in his last comment.

The log was getting a bit noisy for me when I split the changes, so I squashed the commits, but copied the full history over here: https://github.com/jschuh/klipper/tree/sd_ignore_fail_old

KevinOConnor:
I think this got merged with #4471.

-Kevin

PR #4385, opened 2021-06-13
 Refactored tsl1401cl and hall_filament into filament_width_sensor

dawidmpunkt:
module: klippy/extras/filament_width_sensor.py

Refactored tsl1401cl and hall_filament_sensor into filament_width_sensor.
Included extruder.find_past_position() method.

Signed-off-by: Dawid Murawski <dawid.m@gmx.net>

@KevinOConnor @nopbit


KevinOConnor:
FYI @nopbit .

Separately, as high-level feedback, there seems to be a lot of duplication between tsl1401cl_filament_width_sensor.py and hall_filament_sensor.py .  The code really should be refactored - I'm not sure we want to continue to duplicate code between them.  (At a high-level, both modules measure an ADC value and use an offset and factor to obtain the filament width.)  Also, it is now possible for the code to obtain an accurate filament position using the extruder `find_past_position()` method and neither module uses this functionality.

-Kevin

dewi-ny-je:
Just as generic feedback, it was pointed out in the Discord by Dawidm that the SKR 2 has a noisy ADC, around 0.075 V noise. He solved it with an external ADC MCP3425, 0.01 V noise. It may be useful, for this kind of sensors, to allow external ADC, since a small error may result in big apparent diameter changes

nopbit:
> 
> 
> FYI @nopbit .
> 
> Separately, as high-level feedback, there seems to be a lot of duplication between tsl1401cl_filament_width_sensor.py and hall_filament_sensor.py . The code really should be refactored - I'm not sure we want to continue to duplicate code between them. (At a high-level, both modules measure an ADC value and use an offset and factor to obtain the filament width.) Also, it is now possible for the code to obtain an accurate filament position using the extruder `find_past_position()` method and neither module uses this functionality.
> 
> -Kevin

Dear Kevin,

I agree with you no need to have duplicated code. Due to my work load i have sadly no time to compare two addons codes in a sooner time period. As you mentioned main functionality is same. Input voltage equals to width of the filament always.

Best regards
Mustafa

dawidmpunkt:
> 
> 
> Just as generic feedback, it was pointed out in the Discord by Dawidm that the SKR 2 has a noisy ADC, around 0.075 V noise. He solved it with an external ADC MCP3425, 0.01 V noise. It may be useful, for this kind of sensors, to allow external ADC, since a small error may result in big apparent diameter changes

I will try to write a MCP3425/MCP3428 module. External ADCs are useful for extra thermocouples etc. After that, we could modify both filament sensor modules to import either the mainboard's ADC module or the MCP3425/MCP3428 module.

dewi-ny-je:
> > Just as generic feedback, it was pointed out in the Discord by Dawidm that the SKR 2 has a noisy ADC, around 0.075 V noise. He solved it with an external ADC MCP3425, 0.01 V noise. It may be useful, for this kind of sensors, to allow external ADC, since a small error may result in big apparent diameter changes
> 
> I will try to write a MCP3425/MCP3428 module. External ADCs are useful for extra thermocouples etc. After that, we could modify both filament sensor modules to import either the mainboard's ADC module or the MCP3425/MCP3428 module.

The best would be to be able to define any arbitrary number of "ADC" sections in the config, more or less like an MCU, where you define the ADC chip and the number of channels directly in them. At that point those external ADCs could be called anywhere an ADC pin is expected!
No need to change any module, it's just a matter of changing the pin definitions to take advantage of the external ADC!

Once such a module is defined, adding further ADC chips should also be easy with little coding. For example, there are cheap 24 bit ADCs which offer two channels each and can get sub-mV noise.
For example, I'm not able to do all the stuff I said here, but if there were an ADC module I just need to modify to adapt the SPI commands for a different chip, that I would be able to do.

dawidmpunkt:
> The best would be to be able to define any arbitrary number of "ADC" sections in the config, more or less like an MCU, where you define the ADC chip and the number of channels directly in them. At that point those external ADCs could be called anywhere an ADC pin is expected!
> No need to change any module, it's just a matter of changing the pin definitions to take advantage of the external ADC!
> 
> Once such a module is defined, adding further ADC chips should also be easy with little coding. For example, there are cheap 24 bit ADCs which offer two channels each and can get sub-mV noise.
> For example, I'm not able to do all the stuff I said here, but if there were an ADC module I just need to modify to adapt the SPI commands for a different chip, that I would be able to do.

There already seem to be some modules e.g. for the BME280 connected via I2C. Maybe parts of the code could also be used for the MCP342x ADCs. It will take me some time. I'm quite new to Python.
https://github.com/KevinOConnor/klipper/blob/master/docs/Config_Reference.md#bmp280bme280bme680-temperature-sensor

The code I (noobishly) clapped together to test the filament width sensor with the MCP3425 runs on Python 3 and uses the smbus package. I dont quite understand yet, how to make it work for Python 2 (Klipper) and hope that maybe we can write something simpler.

KevinOConnor:
I'm not sure what the state of this PR is (it doesn't pass the regression test cases).

As for smbus, it should be possible to use the bus.I2C() class to communicate with devices.  See klippy/extras/mcp4451.py for an example.

-Kevin

dawidmpunkt:
> 
> 
> I'm not sure what the state of this PR is (it doesn't pass the regression test cases).
> 
> As for smbus, it should be possible to use the bus.I2C() class to communicate with devices. See klippy/extras/mcp4451.py for an example.
> 
> -Kevin

Thank you for the feedback! Apparently I forgot some whitespaces. All checks have passed now. I will work on an update for the config reference, as I have added options for logging, runout detection etc.

As for the module for mcp342x adcs, I am working on some code and I found the existing mcpxxxx and htu21d modules helpful. But I2C via klipper works somehow different than smbus. e.g. sometimes the write/read command only accepts decimal values and sometimes it accepts hex. I was not able to address the adc and get a reading via klipper yet. 

After this is done, I want to add a virtual adc pin so the filament sensor and other modules can be connected to the external adc.

dawidmpunkt:
Here ist the code I am working on:
https://github.com/dawidmpunkt/for_klipper/blob/main/mcp342x.py
If anyone is interested in this: I'd be happy about some feedback/help.

KevinOConnor:
As high level feedback, I think this change should probably come after some refactoring of the tsl1401cl and hall_filament_width sensor code.

-Kevin

dawidmpunkt:
Ok, so what ist there to do? 
Would it be sufficient, to restructure the code of one module, so that the other module can call the (otherwise duplicate) functions from the first one?

KevinOConnor:
When I last looked at this, I got the impression one module could handle both tsl1401cl and hall_filament - both take an ADC reading and use that to determine a filament diameter.  I got the impression that a single module that took "adc_value1 / filament_width1", and "adc_value2 / filament_width2" configuration settings would be sufficient.  The code should also use the (recently added) `extruder.find_past_position()` method to get a more accurate extruder position.  Several of the existing commands for configuration should no longer be necessary (eg, the configuration commands in hall_filament_sensor could use QUERY_ADC instead).  The hall_filament_sensor uses two ADC pins and averages them, but that could be handled by the code by detecting if an `adc2_pin` is defined in the config.

I don't own this hardware, so I could be wrong.  In any case, it seems like there's a lot of code duplication and both modules could use the improved `extruder.find_past_position()`.

-Kevin

dawidmpunkt:
That should be possible. I will write a module that supports both sensors, then we'll see. It'll take some time, though. I need to read up on what extruder.find_past_position() does and how to use it.

dawidmpunkt:
@dewi-ny-je: I wrote a module for external ADCs, if you are interested: https://github.com/KevinOConnor/klipper/pull/4681

PR #4388, opened 2021-06-14
Add debugging tool to permit a pin to be used multiple times in the config file

KevinOConnor:
Normally, if one uses the same pin multiple times in a config file it is the result of an error, and the Klipper code will report a "pin XXX used multiple times in config" error.  However, sometimes it is useful to use the same pin multiple times for debugging purposes.

This PR adds a new config option to override the default error check.

-Kevin


th3fallen:
Any chance this could be adapted to work for a thermistor ;)

PR #4389, opened 2021-06-14
Update Bed_Level.md

gregmatic:
Line 64 mentions paper 'width' to be 100 micron...should be thickness


KevinOConnor:
Okay, but the term "width" is used several times in that paragraph.  I don't think "width" is that confusing, because from the context it is clear which dimension is being referred to.

Separately, commits need to have a signed-off-by line for me to directly commit -see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

KevinOConnor:
Thanks, but I'm going to close this as I don't think this is a good match for the Klipper master branch at this time.

-Kevin

PR #4390, opened 2021-06-14
More config examples

Aeronaut:
Brings:
* Mappings for GT2650 Ver 3.0 
* Mappings for Arduino UNO CNC Shield V3.0
* Printer config for Geeetech A20
* Macros for Geeetech A20 and Web-Interface Fluidd

I also tried to make the configs as clean, modular and reusable as possible.
Please let me know if there are more concepts on how to make things more clear.


zellneralex:
Please remove the fluidd.cfg from your list as they will be changed soon after #4366 is merged

Aeronaut:
> Please remove the fluidd.cfg from your list as they will be changed soon after #4366 is merged

Done.

KevinOConnor:
Thanks.  Please see the guide for example config files at: https://www.klipper3d.org/Example_Configs.html .  Also, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

It looks like you have a few configs that wouldn't fit the normal config categories.  It's possible we could extend the categories, but it might be simpler to just use "sample" files for that purpose.  In any case, probably better to look at merging the standard configs first.

-Kevin

Aeronaut:
> Thanks. Please see the guide for example config files at: https://www.klipper3d.org/Example_Configs.html . Also, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md
> 
> It looks like you have a few configs that wouldn't fit the normal config categories. It's possible we could extend the categories, but it might be simpler to just use "sample" files for that purpose. In any case, probably better to look at merging the standard configs first.
> 
> -Kevin
Thanks for your feedback. 
* Renamed the files

Signed-off-by: Andreas Clausnizer <andreas@clausnizer.de>

Aeronaut:
> Thanks. Please see the guide for example config files at: https://www.klipper3d.org/Example_Configs.html . Also, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md
> 
> It looks like you have a few configs that wouldn't fit the normal config categories. It's possible we could extend the categories, but it might be simpler to just use "sample" files for that purpose. In any case, probably better to look at merging the standard configs first.
> 
> -Kevin

Thanks for your feedback. 

* Renamed the files
* Moved printer sections out of generic board definition
* Added printer.test sections entry

I'm not sure if this commit now 100% fits the guidelines, let me know if there is still something to fix.

Signed-off-by: Andreas Clausnizer <andreas@clausnizer.de>

PR #4391, opened 2021-06-14
Update FAQ.md

nick-parker:
module: Add UART0 tip to FAQ

Adding an important tip for the hard-wired UART0 pins to FAQ. There isn't a great
reason to use these other than compatibility with existing wiring in a system, but
for those that need it this will save several hours of debugging. The standard
raspbian build we recommend installing Klipper on has the bluetooth chip enabled
and routed through UART0, so this serial port will show up in the OS but not work
until bluetooth is disabled.

Signed-off-by: Nick Parker <nmp67@cornell.edu>

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -71,6 +71,14 @@ Follow the directions in the
 "[Where's my serial port?](#wheres-my-serial-port)" section to prevent
 this from occurring.
 
+### My micro-controller is wired to UART0 pins and refuses to flash.
+
+You likely need to add `dtoverlay=disable_bt` to `/boot/config.txt`. Raspberry
+pi reuses UART0 for the bluetooth host chip, so the serial port will show up as
+`/dev/AMA0` but be flooded with garbage until bluetooth is disabled. UART0 is
+not recommended in general, but works reliably with this fix if your system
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  We can add a new entry to the FAQ, but it would need to be added to the table-of-contents at the top of the file.  Also, we should make it clear this applies to the raspberry pi 3 (and not rpi1/2 nor other SOCs), and note that it is often necessary to disable the rpi kernel logging (via raspi-config) as well.

If I get a chance I'll take a look at it.

-Kevin

mental405:
@nick-parker , we should be able to merge this if you bring it up to date with the current FAQ.md and add a link to the table of contents.

nick-parker:
@mental405 sorry for the glacial response here, I've had a busy and covid-extended holiday season.

I've just updated my master branch to Klipper3d:master with a new FAQ commit on top of it. @KevinOConnor I'm not familiar with the kernel logging issue you mentioned and google didn't turn anything up, so I haven't included that yet.

I added the note in the section of the FAQ relating to flashing, but if you would prefer I append it at the end to avoid modifying so much of the table of contents let me know.

PR #4392, opened 2021-06-14
Add STEVAL3DP001-v1 board support

h-oussama:
This adds support for the STEVAL3DP001-V1 board from STMicroelectronics.

Changed files: 
- stm32f4xx.h
- kconfig
- adc.c
- stm32f4.c  

added files:
- generic-steval-3dp001v1.cfg 
- L6474.py (support for the l6474 motor drivers)
- l647x.py  (support for the l6474 motor drivers)

Signed-off-by: HOUIDAR Oussama  houidar.ousama@gmail.com


KevinOConnor:
Thanks.  In general, it looks useful to me.  I do have some high-level comments.

1. The change to the lib/stm32f4/include/stm32f4xx.h file does not look correct.  In general, we try not to modify the upstream code that we import.  It shouldn't be necessary to modify that file as the stm32 Makefile should define STM32F401xE anyway.
2. It should not be necessary to distinguish between the f401xc and f401xc as Klipper never uses more than about 12KiB of ram anyway.  The build can safely use the smaller values for both chips.
3. I don't understand the changes to the stm32 code.  Is there a reason the clock frequencies need to be different on this particular chip?  In general, we want to use the same settings on as many chips as possible.
4. It looks like there is code copied from tmc.py to the new L647x.py file.  Could the original FieldHelper (and similar) classes in tmc.py be used instead?  Can you describe what changes were made to this (and the other) classes in the new L647x.py file.
5. If we add new config sections (eg, L6474) then it will be necessary to update docs/Config_Reference.md with a description of the new config options.  Also, it would be preferable for the example configs to not have individual field descriptions.

Cheers,
-Kevin

h-oussama:
Thanks @KevinOConnor  for the feedback.

1. You are right, no need to modify `lib/stm32f4/include/stm32f4xx.h` , so i did retrieve back this modif in commit : https://github.com/KevinOConnor/klipper/pull/4392/commits/9c1f7824fc0c725f9824c539a2ab9ce4edd67404
2. Indeed as Klipper does hold no more than 12Kib of Ram, Using the smaller value for both chips is not critical, I will make needed changes in order to have a common configuration for both chips.
3. In what concerns the STM32 code, the stm32f401 microcontrollers have actually the same clock configuration, so I did arrange code to have a common clock configuration for both chips. (commit: https://github.com/KevinOConnor/klipper/pull/4392/commits/9c1f7824fc0c725f9824c539a2ab9ce4edd67404 )
         - I did rectify the peripheral clock frequencies (APB1 up to 42Mhz & APB2 up to 84Mhz).
         - I did also rectify the APB Prescaler values ( 2 for APB1 & 1 for APB2 ).
4. For the L647x module, indeed, the FiledHelper class could be imported as it's from the tmc module, I will make necessary changes, and come up with more explanations about the changes made to other classes in the few days to come.
5. Yes, adding new sections to the config file is necessary, I will update the `docs/Config-Reference.md` with a description of the new config options, and clean the config file from any individual field descriptions.

-Oussama

KevinOConnor:
> In what concerns the STM32 code, the stm32f401 microcontrollers have actually the same clock configuration, so I did arrange code to have a common clock configuration for both chips. (commit: 9c1f782 )

Is it necessary to make a change to the clocks at all (as compared to the clock configuration on the current Klipper master branch)?

-Kevin

h-oussama:
> Is it necessary to make a change to the clocks at all (as compared to the clock configuration on the current Klipper master branch)?
> 
> -Kevin

Actually, on the current `stm32f4.c` code in the master branch, the APB2 peripheral clock is set to 42Mhz which not the maximum freq that the bus can support. This is not critical by itself, but I think it's better to take advantage of the higher speed of the bus in question.
This is why I have made the following changes to the `stm32f4.c` code :
 - Added a section to the `get_pclock_frequency(uint32_t periph_base)` function to return the standard (maximum) frequencies of the different peripheral clocks for the `stm32f401` chips.
```C
get_pclock_frequency(uint32_t periph_base)
{
#if CONFIG_MACH_STM32F401
    if (periph_base < APB2PERIPH_BASE) {
         return FREQ_PERIPH;           // APB1 clock frequency  (up to 42 MHz)
    } else {
         return FREQ_PERIPH * 2;       // APB2 clock frequency  (up to 84 MHz)
    }
#else
    return FREQ_PERIPH;
#endif
}
```
 - Set the standard APB Prescalers values for the `stm32f401` chips.
```C
// Switch system clock to PLL
    if (FREQ_PERIPH_DIV == 2)
        RCC->CFGR = RCC_CFGR_PPRE1_DIV2 | RCC_CFGR_PPRE2_DIV1 | RCC_CFGR_SW_PLL;  //APB1 set to 2 & APB2 set to 1
    else
        RCC->CFGR = RCC_CFGR_PPRE1_DIV4 | RCC_CFGR_PPRE2_DIV4 | RCC_CFGR_SW_PLL;
```

In the image below, I highlighted the values in question for `stm32f401` chips.

image

KevinOConnor:
> Actually, on the current stm32f4.c code in the master branch, the APB2 peripheral clock is set to 42Mhz which not the maximum freq that the bus can support. This is not critical by itself, but I think it's better to take advantage of the higher speed of the bus in question.

Okay, but there's nothing on that bus that would utilize the higher frequency with Klipper.  So, my high-level feedback is that I think we should prioritize code simplicity here and continue to use the same frequency for both APB1 and APB2.

-Kevin

h-oussama:
Hello @KevinOConnor ,
- I have added the [L6474] section to the `docs/Config-Reference.md`, and removed any descriptions from the `generic-steval-3dp001v1` config file. 
- In what concerns the FieldHelper class, i have imported it from the TMC module, as it can be used as it is. For the other classes, i tried to let the same structure as for the tmc's, but did change the methods, especially for the build, read, write methods, as the L6474 motor drivers use an 8-bit SPI while tmc's use 40-bit SPI. the two also differ on the initialization process.
- Code has been tested using the STEVAL board with the SOVOL Sv01 printer.

KevinOConnor:
Thanks.  Unfortunately, I'm a bit behind on review of this PR (and several others).  What is the state of the changes to the src/stm32/ directory?  This PR still makes changes there (as well as changes to lib/stm32f4/ ).

-Kevin

PR #4395, opened 2021-06-14
stm32: SPI changes in preparation for ST7789V support

matthewlloyd:
As requested in https://github.com/KevinOConnor/klipper/pull/4374#issuecomment-860868153

* Use APB1RSTR etc. to reset peripherals when enabling them, as we already do for STM32F1.
* Always enable SPI buses and configure GPIOs for them, even if they appear already enabled (e.g. by the bootloader). It's cheap and harmless, and prevents misconfiguration.


matthewlloyd:
On second thought, I will just arrange these as separate commits in my original PR rather than opening a new PR here.

PR #4396, opened 2021-06-14
neopixel: add a get_status() method

Arksine:
This request adds a get_status() method for both the neopixel and dotstar modules.  Status will return a `color_data` field that contains a list of RGB(W) values for each led in the chain.  For example, fetching the red value for the first led in a chain might look like the following:
```
printer["neopixel my_led"].colordata[0].R
```

This is primarily useful for clients that wish to provide a UI for changing LED settings, however it could have some use for templates as well.


KevinOConnor:
Thanks.

-Kevin

PR #4398, opened 2021-06-15
klippy: add `klippy:interrupt` used by `pause_resume/cancel`

ayufan:
This allows to implement a graceful interrupt of a blocking
commands that do implement `klippy:interrupt` to break
busy wait loops.

Example implemented here break wait for temperature on heaters
to make cancel be almost immediate when called via API.


KevinOConnor:
Thanks.  In general it seems like useful functionality.  However, I'd suggest some changes:
1. It seems this is intended to be a generic mechanism to interrupt long running g-code commands.  As such, I think I'd add it to the "gcode.py" module and not the klippy.py module.
2. I would not alter the behavior of the existing cancel_print webhook - as that would require deprecation notices.  Might be better to introduce a new webhook in the gcode.py module.
3. I think it should be possible to introduce a new `gcode.check_long_running()` helper function so as to reduce duplication in the callers that have long running g-code commands.  I'll try to look at this when I get a chance.

-Kevin

ayufan:
> It seems this is intended to be a generic mechanism to interrupt long running g-code commands. As such, I think I'd add it to the "gcode.py" module and not the klippy.py module.

Yes, ideally we should be able to interrupt all, like: `RESONANCE TEST`, `HEATERS OFF`, etc. Do you think that this is wrong that it is a default behavior of `cancel print`? I would not treat this as breaking/deprecation rather like an expected outcome. Cancel should be in most cases graceful and immediate as much as possible, so this fulfil that goal. We can add additional interrupt webhook, but I would still assume that this is a default behavior of cancel.

> I think it should be possible to introduce a new gcode.check_long_running() helper function so as to reduce duplication in the callers that have long running g-code commands. I'll try to look at this when I get a chance.

The tricky part is to some figure out a mutex or conditional wait to perform this gracefully. Likely the best would be to implement a busy loop function that would accept lambda.

garethky:
I looked though the code and I think it is generally OK. I think this is high impact because the behavior of "Cancel Print" is so often misunderstood by users that it has its own FAQ: https://github.com/Klipper3d/klipper/blob/master/docs/FAQ.md#how-do-i-cancel-an-m109m190-wait-for-temperature-request. There have been previous feature requests for this and I want it on my printers.

1) I second Kevin's comment about putting it in gcode.py. I would call the event `gcode:request_abort`. We already have `gcode:request_restart` in there.

2) I cant say if the original cancel behavior needs to be kept. My intuition is that it doesn't, people would generally prefer instant cancel. If you made a new web hook, wouldn't you just go back to Moonraker and change the default Cancel behavior to call that? That's the same effect as not making a new one. I think if we can get @Arksine to comment we can make a decision.

3) I don't know that I grock the comment about `gcode.check_long_running()` but here is my stab at it: I take it to mean removing the need to keep the `is_waiting` flag, related code and event handler in each module. Basically you just call `gcode.check_long_running()` in the busy loop conditional. I think this could work if we assume that only 1 of these busy loops is ever executing (which is the case because only 1 gcode command runs at a time). The flag gets set when the event occurs and it gets re-set when the method is called. But maybe that comment went right over my head 😅.

You probably want update or delete that FAQ in this PR 😉

KevinOConnor:
FYI, I came up with an implementation of the high-level idea from this PR, but using helper code in gcode.py .  The code can be found at https://github.com/KevinOConnor/klipper-dev/tree/work-cancel-20220322 .

However, I'm not happy with the results and I'm no longer sure if a "cancel" system is a good idea.  There are two issues that I found:
1. The behavior of a "cancel" request was not intuitive.  For example, a wait for bed heat request of `M190 S80` would block the g-code command queue, and I could successfully cancel that blocking state with the `gcode/cancel` webhook.  However, cancelling the wait *did not* cancel the temperature change.  That was unintuitive and I fear it could lead to notable grief for users.  It's not clear to me what behavior should be taken on a cancel of a `M190` wait for bed temperature command (should it turn the heater off, should it revert to the previous bed temperature, should it keep the new temperature but not pause new commands, etc.).
2. It's surprisingly difficult to abort some commands.  I think it would be unintuitive for the user that a "cancel" request only stops some commands.  For example, `PROBE` and `G28` commands can take a long time, but deconstructing the state on a user cancel is error prone and could lead to large errors; the `delta_calibrate` command can run for minutes, but stopping the background math process would require notable work.

The existing `M112` command doesn't have these subtle issues.  It's certainly a little annoying to use, but I think its behavior is easier to understand and provides a more consistent experience.  Cancelling is not a common event and I'm concerned that trying to optimize it could lead to a more complex system and increase the chance of larger failures.

-Kevin

PR #4400, opened 2021-06-15
config: Specify gear_ratio for Prusa Mini+ extruder

matthewlloyd:
Also clarifies the documentation for rotation_distance in the
stepper docs.


KevinOConnor:
Thanks, I committed this change.

If you get a chance, though, it's nice to actually count the teeth on the gears and use that number in the config.  A lot of the "3:1" gears aren't really 3 to 1 - for example, the bmg is actually 50:17.

-Kevin

PR #4404, opened 2021-06-17
Probe & manual_probe: add ability to save babystepping

shiftingtech:
Created two new extended gcodes: PROBE_UPDATE_OFFSET, and Z_ENDSTOP_UPDATE_POSITION
These use the z gcode offset to revise the probe offset, or z endstop position
allowing users to make a frequently used babystepping value permanent without
manual config editing.

Signed-off-by: Ben Eastep <shifting@shifting.ca>

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -207,6 +207,10 @@ The following standard commands are supported:
   for calibrating a Z position_endstop config setting. See the
   MANUAL_PROBE command for details on the parameters and the
   additional commands available while the tool is active.
+  `Z_ENDSTOP_UPDATE_POSITION`: Take the current Z Gcode offset (aka, 
+  babystepping), and subtract it from the stepper_z endstop_position. 
+  This acts to take a frequently used babystepping value, and "make 
+  it permanent".  Requires a `SAVE_CONFIG` to take effect.
 - `TUNING_TOWER COMMAND=<command> PARAMETER=<name> START=<value>
   FACTOR=<value> [BAND=<value>]`: A tool for tuning a parameter on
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  At a high-level it looks fine to me.  Some minor notes:
1. The convention in Klipper is for code to not peek at member variables of other classes.  So, the code should call gcode_move.get_status() and not directly access `gcode_move.base_position`.
2. I'm not sure about the new command names.  Although I'm not sure I have any suggestions either.  I'll give it some more thought.

-Kevin

shiftingtech:
APPLY_Z_OFFSET/APPLY_PROBE_OFFSET or COMMIT_Z_OFFSET/COMMIT_PROBE_OFFSET have been suggested as other possible names.

shiftingtech:
I've opted for Meteyou's suggestion of 

z_offset_apply_endstop
z_offset_apply_probe

seems clear, and not overly wordy

shiftingtech:
@KevinOConnor   you do you feel about the names I chose here?  Would you rather go a different direction with it?

KevinOConnor:
Thanks, and sorry for the delay.

-Kevin

zellneralex:
@KevinOConnor Extended GCODE names are wrong G-Codes.md. How we should solve this? 
Do we need a new PR to correct that

KevinOConnor:
Ah - I missed that the documentation had not been updated.  We'll need to fix the documentation.

-Kevin

shiftingtech:
I completely forgot I hadn't done that yet: I'll fix it tonight

PR #4405, opened 2021-06-17
Add controller fan pin

th3fallen:
Add pinout for controller fan socket


KevinOConnor:
Needs a signed-off-by line and the definition should probably be commented out as the end user may not have actually have a fan attached to it.

-Kevin

KevinOConnor:
I'm going to close this as it looks like it has gone stale.

Separately, you have opened several incomplete PRs on the Klipper repo.  It takes time for me to review and track these PRs.  Please make sure any future PRs are complete before opening them.

-Kevin

PR #4406, opened 2021-06-18
Support Arduino style to force restart into bootloader mode.

henkyprayoga:
I just use Seeduino XIAO as additional board for my Klipper setup. Somehow, after I flashed Klipper into it, it refuse to enter bootloader mode by using baud rate 1200 mechanism. Resetting the board twice still works. Turns out that XIAO bootloader ignore double tap magic or probably has different magic set. 
Luckily, this problem doesn't occurred when we use Arduino IDE. So after some digging, Arduino erase application area to force the MCU enter bootloader mode after reset.
This PR add option for that mechanism.


KevinOConnor:
Okay, but I'm leery of having the Klipper code reset the flash.  We also try to avoid "ifdefs" in the Klipper mcu code where possible.  As an alternative, could those with this board flash one of the publicly available bootloaders that does not have this limitation?

-Kevin

PR #4408, opened 2021-06-18
config: Update gear_ratio for Prusa Mini+ extruder based on counting teeth

matthewlloyd:


KevinOConnor:
Okay, thanks.  That's certainly a surprising gear ratio - it's not that close to 3:1 and it doesn't actually mesh the teeth well either.

-Kevin

PR #4412, opened 2021-06-22
Add support for STM32F072 (used in the TurboCAN board)

henrikssn:
See https://github.com/henrikssn/TurboCAN

Temp compensation logic is common across STM32F0: https://techoverflow.net/2015/01/13/reading-stm32f0-internal-temperature-and-voltage-using-chibios/

### DIFF
--- a/klippy/extras/temperature_mcu.py
+++ b/klippy/extras/temperature_mcu.py
@@ -66,9 +66,9 @@ class PrinterTemperatureMCU:
             ('samd21', self.config_samd21), ('samd51', self.config_samd51),
             ('stm32f1', self.config_stm32f1), ('stm32f2', self.config_stm32f2),
             ('stm32f4', self.config_stm32f4),
-            ('stm32f042', self.config_stm32f042),
+            ('stm32f042', self.config_stm32f0x2),
             ('stm32f070', self.config_stm32f070),
-            ('stm32f072', self.config_stm32f070),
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .

Separately, I'm not sure why the stm32f070 specific temperature code was removed.  That chip is different from the stm32f042/72.

-Kevin

henrikssn:
> Thanks. Commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .
> 
> Separately, I'm not sure why the stm32f070 specific temperature code was removed. That chip is different from the stm32f042/72.
> 
> -Kevin

You are correct, the stm32f0x0 seems to be special compared to the stm32f0x1/0x2/0x8. I'll add that code back and rename the methods so make this clear.

henrikssn:
> Thanks. Commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .
> 
> Separately, I'm not sure why the stm32f070 specific temperature code was removed. That chip is different from the stm32f042/72.
> 
> -Kevin

I have applied both of your suggestions.

-Erik

KevinOConnor:
Thanks.

-Kevin

PR #4414, opened 2021-06-22
Add support for Raspberry Pi Pico micro-controllers

KevinOConnor:
This adds initial support for the rp2040 micro-controller (as found on the Raspberry Pi Pico board).

This commit adds basic support for UART communication, GPIO pins, internal timers, and the watchdog.  I plan to work on ADC and USB support in the near future.  (I don't have any immediate plans to work on SPI, I2C, nor hard_pwm support.)  This code only uses one of the two cortex-m cores on the chip (it's not clear that utilizing both cores would result in an improvement given Klipper's IO driven usage patterns).

Performance looks very good on these chips - the benchmarks reach 727K steps/s (with 4 active steppers).  That's near the highest Klipper mcu performers, and this chip is notably less expensive.

-Kevin


raymondh2:
Glad to see support for this. This will make a lot of expander boards more readily available for manufacture.

KevinOConnor:
I've added basic ADC and USB support to this branch.  I'm inclined to commit this branch in its current state, as additional features can be added later.

Things to consider in the future:
1. It would probably be better to run all the code from ram instead of flash.  The chip has a cache for the flash chip, but cache misses may introduce less predictable code execution.  Given this chip has more memory than we need, there shouldn't be any harm in running the code from ram.
2. It would be useful to support unique ids - in particular for giving each usb device a unique name.  The rp2040 doesn't directly have a unique id, but each flash chip attached to it should.
3. It would be useful to support rebooting into the builtin bootloader.  It would also be useful to extend scripts/flash_usb.py to support this chip (so that `make flash` works).
4. The chip has a "usb reset errata workaround" that has not been implemented.  It probably should be.
5. The USB code works for me in basic tests, but it could use additional testing.
6. Hardware spi, i2c, and hard_pwm support.

-Kevin

Issue #3682, opened 2020-12-26
Error when compiling firmware on Raspberry Pi Zero W


Xkonti:
When following the klipper installation process the firmware compilation step fails. After sending following the command to the Raspberry Pi the process ends with an error:

`make flash FLASH_DEVICE=/dev/serial/by-id/usb-1a86_USB_Serial-if00-port0`

Console output:

```
  Building out/autoconf.h
  Compiling out/src/sched.o
  Compiling out/src/command.o
  Compiling out/src/basecmd.o
  Compiling out/src/debugcmds.o
  Compiling out/src/initial_pins.o
  Compiling out/src/gpiocmds.o
  Compiling out/src/stepper.o
  Compiling out/src/endstop.o
  Compiling out/src/adccmds.o
  Compiling out/src/spicmds.o
  Compiling out/src/thermocouple.o
  Compiling out/src/i2ccmds.o
src/i2ccmds.c:14:23: error: field ‘i2c_config’ has incomplete type
     struct i2c_config i2c_config;
                       ^
src/i2ccmds.c: In function ‘command_config_i2c’:
src/i2ccmds.c:23:23: warning: implicit declaration of function ‘i2c_setup’ [-Wimplicit-function-declaration]
     i2c->i2c_config = i2c_setup(args[1], args[2], addr);
                       ^
src/i2ccmds.c: In function ‘command_i2c_write’:
src/i2ccmds.c:35:5: warning: implicit declaration of function ‘i2c_write’ [-Wimplicit-function-declaration]
     i2c_write(i2c->i2c_config, data_len, data);
     ^
src/i2ccmds.c: In function ‘command_i2c_read’:
src/i2ccmds.c:48:5: warning: implicit declaration of function ‘i2c_read’ [-Wimplicit-function-declaration]
     i2c_read(i2c->i2c_config, reg_len, reg, data_len, data);
     ^
make: *** [Makefile:64: out/src/i2ccmds.o] Error 1
```

There are no problems when connecting the OctoPrint with the printer.

Platform: Raspberry Pi Zero W
System: OctoPi 0.17.0
Printer: Anycubic Mega Zero 2.0

klipper-gitissuebot:
Hi @Xkonti,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

[...]

PR #4415, opened 2021-06-22
Add Make Clean Instruction

### LINKED ISSUE #4409: ‘i2c_config’ has incomplete type      struct i2c_config i2c_config;
Compiling out/src/i2ccmds.o
src/i2ccmds.c:14:23: error: field ‘i2c_config’ has incomplete type
     struct i2c_config i2c_config;

TheFanatr:
Adds the instruction to run `make clean` after `make menuconfig` if `make` will be run afterwards, otherwise the build will fail as described in the issues below.

This behavior is unexpected and confusing, and there is no official documentation or instruction regarding the issue itself <s>nor what must be done to allow the build to continue successfully</s>. This step is shown [in the FAQ, under "How Do I Upgrade"](https://github.com/KevinOConnor/klipper/blob/master/docs/FAQ.md#how-do-i-upgrade-to-the-latest-software) ([thanks @ShohninDmitriy](https://github.com/KevinOConnor/klipper/pull/4415#issuecomment-866246091)) but is not explained, and is still absent from the installation instructions, which currently fail without this step, causing a lot of confusion for people starting out (see the referenced issues at the bottom, and potentially others).

This aims to at least partially resolve this insufficiency by just telling viewers of the installation documentation to simply clean up after `menuconfig` manually.

This should ideally also (or probably only) be resolved in the implementation of the `menuconfig` script, but until that happens, this should suffice and help get people going with a little less hassle than reading through the issues and finding one of the comments with this suggested fix.

Fixes #4409, #3682

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -50,7 +50,16 @@ make menuconfig
 ```
 
 Select the appropriate micro-controller and review any other options
-provided. Once configured, run:
+provided.
+
+Once configured, clean up after the configuration script (important!):
+
+```
+make clean
+```
+
+Then build the firmware:
+
 
 ```
 make


ShohninDmitriy:
https://github.com/KevinOConnor/klipper/blob/master/docs/FAQ.md#how-do-i-upgrade-to-the-latest-software

KevinOConnor:
As indicated, the "make clean" step is documented during updates.  It isn't needed during initial installation.  As far as I know, the i2c error occurs when updating from really old version of Klipper - it should not occur between recent updates.  If someone wants to track down why the error is occurring, we can certainly look at making the build more resilient.  However, I'm reluctant to complicate the installation instructions at this time.

Separately, there are plans (see #4307) to automate the mcu compilation to avoid issues like this.

-Kevin

TheFanatr:
> It isn't needed during initial installation.

I encountered the `i2c`-related build error when following the instructions from a freshly-downloaded copy of the master branch, after running `make menuconfig`, then trying to just run `make`.

I'll double check again tonight to see if I am mistaken, but from my experience it happens every time you don't run `make clean` after `make menuconfig`.

brycestejskal:
@KevinOConnor @TheFanatr -

Just noting that I had hit this on two separate clean installs.

1. Pi3b making for creality 32bit board
2. Pi4 making for octopus 1.1

In both cases, it was the latest Klipper version, brand new pi, and the MCUs had never seen Klipper before.

I am about to do a third in a week or so, and can report back.

PR #4416, opened 2021-06-23
Config: Calibrate E3Pro ESteps

rbncrthms:
I calibrated my rotation_distance to a high degree of accuracy as the
value provided was a bit on the high side.

Signed-off-by: Robyn Thomas <robyn.ceri.thomas@gmail.com>

### DIFF
--- a/config/printer-creality-ender3pro-2020.cfg
+++ b/config/printer-creality-ender3pro-2020.cfg
@@ -54,7 +54,15 @@ step_pin: PB4
 dir_pin: PB3
 enable_pin: !PC3
 microsteps: 16
-rotation_distance: 34.406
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Okay, but each printer and filament tend to have a slightly different rotation_distance.  Which is why we recommend everyone run the calibration procedure at https://www.klipper3d.org/Rotation_Distance.html#calibrating-rotation_distance-on-extruders .  I don't think we need to modify the example config file with a different value.

-Kevin

rbncrthms:
It is your choice, of course, but where did the original value come from?
Perhaps this value would be a better starting point for people than the
other? Although I accept that the original value is still in the valid
range.

KevinOConnor:
The original config file was contributed by @mpfl .

-Kevin

PR #4417, opened 2021-06-23
docs: added an apt update to measuring_resonance

shiftingtech:
Signed-off-by: Ben Eastep shifting@shifting.ca

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -76,8 +76,9 @@ CPU, it may take *a lot* of time, up to 10-20 minutes. Be patient and wait
 for the completion of the installation. On some occasions, if the board has
 too little RAM, the installation may fail and you will need to enable swap.
 
-Next, run the following command to install the additional dependencies:
+Next, run the following commands to install the additional dependencies:
 ```
+sudo apt update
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks - looks fine to me.  @dmbutyugin , do you have any comments?

-Kevin

dmbutyugin:
@shiftingtech I'm fine with the change either, but could you describe what kind of scenario you are trying to avoid/fix? My only tiny concern is that the update may pull in, well, updates of the package index, and the subsequent install command may need to pull a lot of updated dependencies. But perhaps this is exactly intentional?

shiftingtech:
I'm trying to avoid package not found errors.

The thing is, when you don't run apt update,  you have the potential to be
using stale packages indexes.  Stale packages indexes don't result in
getting older packages,  they result in getting no package at all (because
the old file is no longer on the mirror)

Also, apt install is pretty good about only pulling updates that are
directly required, so I don't see it as likely to cascade out of control


Ben

dmbutyugin:
Yes, I suppose that can be a real risk (if the index is too stale). So OK, thanks!

KevinOConnor:
Thanks.

-Kevin

PR #4418, opened 2021-06-24
Add hardware PWM documentation

mrvanes:
This addition explains how to enable and use hardware pwm support on a Raspberry Pi using the Linux host mcu.

### DIFF
--- a/docs/RPi_microcontroller.md
+++ b/docs/RPi_microcontroller.md
@@ -154,3 +154,39 @@ gpiochip1 - 8 lines:
         line   6:      unnamed       unused   input  active-high
         line   7:      unnamed       unused   input  active-high
 ```
+
+Optional: Hardware PWM
+=======================================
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Looks fine but needs a signed-off-by line.

-Kevin

Alexdad76:
Is it possible to use this method to control the laser instead of using a PWM fan?
Instead of this [https://github.com/KevinOConnor/klipper/pull/4128](url)

mrvanes:
No, I don't think this will help controlling laser. The challenge in 4128 is to execute pin commands quicker than the current queueing system allows. This PR only describes how to controll a hardware PWM pin, which is subject to the same queueing mechanism currently implemented in Klipper.
@KevinOConnor I signed the commit, I think everything is in place to merge?

KevinOConnor:
Thanks.

-Kevin

PR #4420, opened 2021-06-24
Add GD32F103RCT6 config (Flashforge Voxelab Aquila board)

NokHook:


DaStivi:
i think the correct name of the GD µP is GD32F103R**CT** ...

NokHook:
> i think the correct name of the GD µC is GD32F103R**CT** ...
> 
> https://www.gigadevice.com/products/microcontrollers/gd32/arm-cortex-m3/mainstream-line/gd32f103-series/?fwp_microcontroller_part_number=GD32F103RCT6

Oh, thanks! That's my fault.

DaStivi:
@NokHook does the Flashforge Voxellab Board work with that  change? as far as i understand everything you just "display" the board type in the make menuconfig.... nothing else special done... 
theoretically it should be compatible with the STM32F103 Chips....

BUT, i got another GD32F103 board and the issue is, that the boards is not getting found via USB, no CDC is gonna reported to Pi... there is a group working on this issue, within Marlin, as when using the STM-Core dev envs there its the same thing.... but i would wonder it works on your side...

connecting the board via UART to the pi does work btw!

NokHook:
> @NokHook does the Flashforge Voxellab Board work with that change? as far as i understand everything you just "display" the board type in the make menuconfig.... nothing else special done...
> theoretically it should be compatible with the STM32F103 Chips....
> 
> BUT, i got another GD32F103 board and the issue is, that the boards is not getting found via USB, no CDC is gonna reported to Pi... there is a group working on this issue, within Marlin, as when using the STM-Core dev envs there its the same thing.... but i would wonder it works on your side...
> 
> connecting the board via UART to the pi does work btw!
> 
> and another thing i noticed, your Flash and RAM values also seams to be way off?! not sure how klippers hex values represents to real numbers, but basically what i saw on the datasheets, the same STM32F103 types have the same flash/ram amounts, besides the GD µC can offer more clock frequency...
> 
> i would asume it should be the same as with the STM32F103 definition, i guess?!
> image


I changed ram and flash size. Becouse GD32F103RCT6 have only 48kb of ram, and 256kb of flash. My Aquila work with that changes.

DaStivi:
what connection do you use? uart or USB?

the STM/GD-32F103RE would have 512K flash...

NokHook:
> what connection do you use? uart or USB?
> 
> the STM/GD-32F103RE would have 512K flash...

Aquila has GD32F103RCT6 so it has 256K flash. I have UART connection via MicroUSB on stock board, just changed baud rate to 230400, because CH341 doesn't support 250000.

DaStivi:
is it possible you did also some change to the lib/stm32f1/include/stm32f1xx.h ??
image
cant use the GD definitions...

NokHook:
> is it possible you did also some change to the lib/stm32f1/include/stm32f1xx.h ??
> image
> cant use the GD definitions...

I have made .config with fixed memory size. Maybe just made menu where you can choose memory size of your STM32/GD32 chip. Right now I think that will be better.

KevinOConnor:
Okay, but I don't think we need to add a new menu selection for this chip.  If it's compatible with the stm32f103 then users can select the existing stm32f103 during "make menuconfig".

Separately, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

DaStivi:
> Okay, but I don't think we need to add a new menu selection for this chip. If it's compatible with the stm32f103 then users can select the existing stm32f103 during "make menuconfig".
> 
> Separately, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md
> 
> -Kevin

@KevinOConnor I tried the changes and it doesnt even work, see my Screenshots... The stm libs of course cant work with the GD descriptions...


The issue is, I am having another Board with ah GD32F103(Re = 512k Flash but Flash size shouldnt matter at all)

Compiled Klipper with the STM32F103 settings does not work on this Board with USB connection! It does not displayed, not even recognizee in pi (or Windows) at all...

There where the same issue with marlin and there is some single/double Buffer issue in CDC/MSC parts...

So this needs to be fixed and I guess also made then as seperated menu in make config... There are Boards comming with these processors! ;)

KevinOConnor:
FYI, I'm told that there are several reports of success using the gd32f103 chips with Klipper on UART.  I'm also told that USB works with the following Klipper change:
```diff
--- a/src/stm32/stm32f1.c
+++ b/src/stm32/stm32f1.c
@@ -119,10 +119,12 @@ gpio_peripheral(uint32_t gpio, uint32_t mode, int pullup)
     else if (pullup < 0)
         regs->BSRR = 1 << (pos + 16);
 
-    if (gpio == GPIO('A', 13) || gpio == GPIO('A', 14))
+    if (gpio == GPIO('A', 13) || gpio == GPIO('A', 14)) {
         // Disable SWD to free PA13, PA14
+        enable_pclock(AFIO_BASE);
         stm32f1_alternative_remap(AFIO_MAPR_SWJ_CFG_Msk,
                                   AFIO_MAPR_SWJ_CFG_DISABLE);
+    }
 
     // STM32F1 remaps functions to pins in a very different
     // way from other STM32s.
```

At best guess, the above change is forcing a reset of the AFIO block when changing software debug pin settings.

In any case, this PR does not look like a good match for Klipper master branch at this time.

-Kevin

PR #4426, opened 2021-06-26
#Issue 4425 - bed meh display swrong id of overlapping maulty regions

craiglink:
Fix displayed index of overlapping faulty region to match the correct region which is indexed starting at 1.


Arksine:
Ah, good catch.  I did indeed forget to account for the offset index when logging the error.

KevinOConnor:
Okay, thanks.  Commits need to have a signed-off-by line for me to directly commit - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .  If you're just reporting the bug, though, then I can make the change using my signed-off-by line.

-Kevin

KevinOConnor:
I made the suggested change - commit 1aa4becc.

Thanks,
-Kevin

PR #3974, opened 2021-02-21
Periodically check TMC drivers for errors

KevinOConnor:
There are several high-level changes in this PR:
1. Every enabled TMC stepper is now polled once per second to see if the driver reports an error (eg, over-current or over-temp).  If the driver reports an error (or can not be contacted) then a shutdown is invoked.
2. TMC temperature warnings are also checked once per second.  Warnings are written to the log.
3. The full initialization sequence is now sent every time the driver is enabled.
4. It is no longer an error if a TMC2208/TMC2209 driver is not available on startup.  This may simplify setups where motor power is externally controlled.  If the software cant contact the driver when it is actually enabled, then a shutdown will be invoked.
5. The SET_TMC_CURRENT command has been refactored.

The above changes apply to all of the tmc drivers (tmc2130, tmc2208, tmc2209, tmc2660, and tmc5610).  However, I do not have all of those drivers to test.

Help with testing would is appreciated.

-Kevin

PR #4429, opened 2021-06-27
TMC driver improvements

leptun:
This PR improves the usability of TMC drivers with klipper.

The configuration of tmc2130, tmc2208, tmc2209 and tmc5160 was expanded to include more fields that were already there, just not parsed in \_\_init\_\_. Helper functions were made more generic (for `TPWMTHRS`, `TCOOLTHRS` and `THIGH`). More about this is described in the first commit message. As a result, stallGuard homing is more reliable, coolStep can be used and fullstepping at high speeds is possible.

The TMC5160 runtime current update functions were fixed so that `GLOBALSCALER` is updated whenever `IRUN` and `IHOLD` is changed. This is done in the second commit.

The homing current is now independent of the running current. In combination with `TCOOLTHRS`, it helps the homing reliability of tmc2130, tmc2209 and tmc5160 drivers. More about this in both commits.

Homing before and after:
- 0.3A run current, SGT=3, TCOOLTHRS=max (0mm/s): https://photos.app.goo.gl/wKCMmKFDk4c5vZBy7
- 0.3A run current, 0.15/0.18 homing current, SGT=3, TCOOLTHRS=413 (20mm/s) (exactly the same values as in Prusa-Firmware for the MK3S): https://photos.app.goo.gl/NTZsQnqxKvH6brV46
Notice how in the first video the second bump is never attempted (stallguard triggering too early).

Older config files work with this PR without changing anything. Updating the config will make use of the new helper functions.
I'd like some feedback on these changes before I update the config reference as this is the first time I use python code. Be advised that there might be rookie mistakes :P

### DIFF
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -316,7 +316,7 @@ class TMCCommandHelper:
 
 # Helper class for "sensorless homing"
 class TMCVirtualPinHelper:
-    def __init__(self, config, mcu_tmc):
+    def __init__(self, config, mcu_tmc, tmc_freq):
         self.printer = config.get_printer()
         self.mcu_tmc = mcu_tmc
         self.fields = mcu_tmc.get_fields()
@@ -334,6 +334,12 @@ class TMCVirtualPinHelper:
         self.en_pwm = False
         self.pwmthrs = 0
         self.tcoolthrs = 0
+
+        velocity = config.getfloat(
+            'sensorless_homing_stallguard_threshold', None, minval=0.)
+        self.homing_tcoolthrs = TMCtstepHelper(
+            config, mcu_tmc, tmc_freq, velocity, 0xfffff)
+
         # Register virtual_endstop pin
         name_parts = config.get_name().split()
         ppins = self.printer.lookup_object("pins")
@@ -378,7 +384,7 @@ class TMCVirtualPinHelper:
             self.fields.set_field("en_pwm_mode", 0)
             gconf_val = self.fields.set_field(self.diag_pin_field, 1)
             self.mcu_tmc.set_register("GCONF", gconf_val)
-        tc_val = self.fields.set_field("TCOOLTHRS", 0xfffff)
+        tc_val = self.fields.set_field("TCOOLTHRS", self.homing_tcoolthrs)
         self.mcu_tmc.set_register("TCOOLTHRS", tc_val)
     def handle_homing_move_end(self, hmove):
         if self.mcu_endstop not in hmove.get_mcu_endstops():
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -351,10 +351,9 @@ class TMCVirtualPinHelper:
         reg = self.fields.lookup_register("en_pwm_mode", None)
         if reg is None:
             self.en_pwm = not self.fields.get_field("en_spreadCycle")
-            self.pwmthrs = self.fields.get_field("TPWMTHRS")
         else:
             self.en_pwm = self.fields.get_field("en_pwm_mode")
-            self.pwmthrs = 0
+        self.pwmthrs = self.fields.get_field("TPWMTHRS")
         self.coolthrs = self.fields.get_field("TCOOLTHRS")
         self.printer.register_event_handler("homing:homing_move_begin",
                                             self.handle_homing_move_begin)
@@ -374,6 +373,8 @@ class TMCVirtualPinHelper:
             self.mcu_tmc.set_register("GCONF", gconf_val)
         else:
             # On earlier drivers, "stealthchop" must be disabled
+            tp_val = self.fields.set_field("TPWMTHRS", 0xfffff)
+            self.mcu_tmc.set_register("TPWMTHRS", tp_val)
             self.fields.set_field("en_pwm_mode", 0)
             gconf_val = self.fields.set_field(self.diag_pin_field, 1)
             self.mcu_tmc.set_register("GCONF", gconf_val)
@@ -388,6 +389,8 @@ class TMCVirtualPinHelper:
             self.mcu_tmc.set_register("TPWMTHRS", tp_val)
             val = self.fields.set_field("en_spreadCycle", not self.en_pwm)
         else:
+            tp_val = self.fields.set_field("TPWMTHRS", self.pwmthrs)
+            self.mcu_tmc.set_register("TPWMTHRS", tp_val)
             self.fields.set_field("en_pwm_mode", self.en_pwm)
             val = self.fields.set_field(self.diag_pin_field, 0)
         self.mcu_tmc.set_register("GCONF", val)
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -482,3 +482,42 @@ def TMCcoolStepHelper(config, mcu_tmc, tmc_freq):
         'coolStep_min_velocity_threshold', None, minval=0.)
     fields.set_field("TCOOLTHRS",
         TMCtstepHelper(config, mcu_tmc, tmc_freq, velocity, 0))
+
+
+class TMCHomingCurrentHelper:
+    def __init__(self, config, mcu_tmc, current_helper):
+        self.printer = config.get_printer()
+        self.name = config.get_name().split()[-1]
+        self.mcu_tmc = mcu_tmc
+        self.current_helper = current_helper
+        self.run_current = 0
+        self.hold_current = 0
+        self.homing_current = config.getfloat(
+            'homing_current', None, minval=0.)
+        self.printer.register_event_handler("homing:home_rails_begin",
+                                            self.handle_home_rails_begin)
+        self.printer.register_event_handler("homing:home_rails_end",
+                                            self.handle_home_rails_end)
+
+    def handle_home_rails_begin(self, homing, rails):
+        if self.homing_current is None:
+            return
+        names = [stepper.get_name() for rail in rails
+            for stepper in rail.get_steppers()]
+        if self.name not in names:
+            return
+        logging.info("homing_begin for %s", self.name)
+        self.run_current, self.hold_current, max_current = (
+            self.current_helper.get_current())
+        self.current_helper.set_current(
+            self.homing_current, self.homing_current, None)
+    def handle_home_rails_end(self, homing, rails):
+        if self.homing_current is None:
+            return
+        names = [stepper.get_name() for rail in rails
+            for stepper in rail.get_steppers()]
+        if self.name not in names:
+            return
+        logging.info("homing_end for %s", self.name)
+        self.current_helper.set_current(
+            self.run_current, self.hold_current, None)
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -333,6 +333,7 @@ class TMCVirtualPinHelper:
         self.mcu_endstop = None
         self.en_pwm = False
         self.pwmthrs = 0
+        self.tcoolthrs = 0
         # Register virtual_endstop pin
         name_parts = config.get_name().split()
         ppins = self.printer.lookup_object("pins")
@@ -354,6 +355,7 @@ class TMCVirtualPinHelper:
         else:
             self.en_pwm = self.fields.get_field("en_pwm_mode")
             self.pwmthrs = 0
+        self.coolthrs = self.fields.get_field("TCOOLTHRS")
         self.printer.register_event_handler("homing:homing_move_begin",
                                             self.handle_homing_move_begin)
         self.printer.register_event_handler("homing:homing_move_end",
@@ -368,12 +370,13 @@ class TMCVirtualPinHelper:
             # On "stallguard4" drivers, "stealthchop" must be enabled
             tp_val = self.fields.set_field("TPWMTHRS", 0)
             self.mcu_tmc.set_register("TPWMTHRS", tp_val)
-            val = self.fields.set_field("en_spreadCycle", 0)
+            gconf_val = self.fields.set_field("en_spreadCycle", 0)
+            self.mcu_tmc.set_register("GCONF", gconf_val)
         else:
             # On earlier drivers, "stealthchop" must be disabled
             self.fields.set_field("en_pwm_mode", 0)
-            val = self.fields.set_field(self.diag_pin_field, 1)
-        self.mcu_tmc.set_register("GCONF", val)
+            gconf_val = self.fields.set_field(self.diag_pin_field, 1)
+            self.mcu_tmc.set_register("GCONF", gconf_val)
         tc_val = self.fields.set_field("TCOOLTHRS", 0xfffff)
         self.mcu_tmc.set_register("TCOOLTHRS", tc_val)
     def handle_homing_move_end(self, hmove):
@@ -388,7 +391,7 @@ class TMCVirtualPinHelper:
             self.fields.set_field("en_pwm_mode", self.en_pwm)
             val = self.fields.set_field(self.diag_pin_field, 0)
         self.mcu_tmc.set_register("GCONF", val)
-        tc_val = self.fields.set_field("TCOOLTHRS", 0)
+        tc_val = self.fields.set_field("TCOOLTHRS", self.coolthrs)
         self.mcu_tmc.set_register("TCOOLTHRS", tc_val)
 
 
@@ -423,18 +426,29 @@ class TMCMicrostepHelper:
         mscnt = self.fields.get_field(field_name, reg)
         return 1023 - mscnt, 1024
 
-# Helper to configure "stealthchop" mode
-def TMCStealthchopHelper(config, mcu_tmc, tmc_freq):
+# Helper for calculating TSTEP based values from velocity
+def TMCtstepHelper(config, mcu_tmc, tmc_freq, velocity, default_val):
     fields = mcu_tmc.get_fields()
-    en_pwm_mode = False
-    velocity = config.getfloat('stealthchop_threshold', 0., minval=0.)
-    if velocity:
+    if velocity is None:
+        return default_val
+    elif velocity > 0:
         stepper_name = " ".join(config.get_name().split()[1:])
         stepper_config = config.getsection(stepper_name)
         step_dist = stepper.parse_step_distance(stepper_config)
         step_dist_256 = step_dist / (1 << fields.get_field("MRES"))
         threshold = int(tmc_freq * step_dist_256 / velocity + .5)
-        fields.set_field("TPWMTHRS", max(0, min(0xfffff, threshold)))
+        return max(0, min(0xfffff, threshold))
+    else:
+        return 0xfffff
+
+# Helper to configure stealthChop-spreadCycle transition velocity
+def TMCStealthchopHelper(config, mcu_tmc, tmc_freq):
+    fields = mcu_tmc.get_fields()
+    en_pwm_mode = False
+    velocity = config.getfloat('stealthchop_threshold', None, minval=0.)
+    tpwmthrs = TMCtstepHelper(config, mcu_tmc, tmc_freq, velocity, 0xfffff)
+    fields.set_field("TPWMTHRS", tpwmthrs)
+    if velocity is not None:
         en_pwm_mode = True
     reg = fields.lookup_register("en_pwm_mode", None)
     if reg is not None:
@@ -442,3 +456,20 @@ def TMCStealthchopHelper(config, mcu_tmc, tmc_freq):
     else:
         # TMC2208 uses en_spreadCycle
         fields.set_field("en_spreadCycle", not en_pwm_mode)
+
+# Helper to configure spreadCycle-highVelocity transition velocity
+# It is also the top speed for coolStep and stallGuard operation
+def TMCTHIGHHelper(config, mcu_tmc, tmc_freq):
+    fields = mcu_tmc.get_fields()
+    velocity = config.getfloat(
+        'high_velocity_threshold', None, minval=0.)
+    fields.set_field("THIGH",
+        TMCtstepHelper(config, mcu_tmc, tmc_freq, velocity, 0))
+
+# Helper to configure coolStep and stallGuard lower velocity limit
+def TMCcoolStepHelper(config, mcu_tmc, tmc_freq):
+    fields = mcu_tmc.get_fields()
+    velocity = config.getfloat(
+        'coolStep_min_velocity_threshold', None, minval=0.)
+    fields.set_field("TCOOLTHRS",
+        TMCtstepHelper(config, mcu_tmc, tmc_freq, velocity, 0))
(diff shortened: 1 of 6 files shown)


KevinOConnor:
Thanks.  In general it looks good, but I have a few high-level comments.

On the first commit:
1. It looks like there are new fields available in the config file which should be added to docs/Config_Reference.md.
2. I'm not sure about the name `spreadcycle_threshold`, as I fear it could be confused with `stealthchop_threshold`.  Maybe something like `fullstep_threshold` instead?
3. I don't understand the change to TMCVirtualPinHelper .  It's not possible to do sensorless homing without setting TCOOLTHRS, so unless I'm missing something, this change would break existing users of sensorless homing (anyone that hasn't tuned the new `stallguard_threshold` setting).  Also, it might be worthwhile to separate out this change to its own commit.

On the second commit:
1. FWIW, the tmc5160 documentation has a bunch of warnings about changing the GLOBALSCALER value at runtime.  I don't have these drivers, so I don't know first hand.
2. If we're going to introduce a homing_current config option I think it should apply to all homing with that driver - not just sensorless homing.  Also, in my tests, changing the current while the carriage was pressed against the rail resulted in poor sensorless homing behaviour.  This was discussed briefly at https://github.com/KevinOConnor/klipper/pull/1957#issuecomment-782784385 and is mentioned at https://www.klipper3d.org/TMC_Drivers.html#sensorless-homing .
3. Why is the homing current handled directly in TMC5160CurrentHelper() and TMCCurrentHelper() instead of in TMCCommandHelper() or TMCVirtualPinHelper?

Cheers,
-Kevin

leptun:
> It looks like there are new fields available in the config file which should be added to docs/Config_Reference.md.

Yes. I'll do this last after everything else is sorted out.

> I'm not sure about the name spreadcycle_threshold, as I fear it could be confused with stealthchop_threshold. Maybe something like fullstep_threshold instead?

`fullstep_threshold` is misleading since it doesn't really enable fullstepping on its own. You have to enable `vhighfs` or `vhighchm` (or both) to alter the behaviour of the driver. Using `TSTEP` will only set the upper velocity threshold for a few things:
- stallguard enabled below this threshold. Default is 0x00000, so maximum velocity.
- fullstepping above the threshold if `vhighfs` is set.
- high velocity chopper mode is activated if `vhighchm` is set. From the datasheet: `This bit enables switching to chm=1 and fd=0, when VHIGH is exceeded. This way, a higher velocity can be achieved. Can be combined with vhighfs=1. If set, the TOFF setting automatically becomes doubled during high velocity operation in order to avoid doubling of the chopper frequency.`
I propose that we name this `highvelocity_threshold` or something similar (the datasheet calls it `high velocity` mode). Maybe this name can be made a bit more readable by some camelCase. So something like `highVelocity_threshold`. It's uglier, but at least it's clear what it does.

> I don't understand the change to TMCVirtualPinHelper . It's not possible to do sensorless homing without setting TCOOLTHRS, so unless I'm missing something, this change would break existing users of sensorless homing (anyone that hasn't tuned the new stallguard_threshold setting). Also, it might be worthwhile to separate out this change to its own commit.

It can be separated if you want, but it might cause conflicts with the second commit. Please let me know if I should continue. In any case, `TCOOLTHRS` is set on `__init__` by `TMCTCOOLTHRSHelper()`. it defaults to 0xFFFFF just like it is needed for sensorless homing. No need to set it to 0 when homing is not performed.

> FWIW, the tmc5160 documentation has a bunch of warnings about changing the GLOBALSCALER value at runtime. I don't have these drivers, so I don't know first hand.

I can test the behavior of tmc5160 drivers at a later time as I have a few silentstepsticks and an skr1.4 board. The tmc5160 changes weren't tested yet, just the tmc2130 ones. I might be able to also test the behaviour of tmc2209 drivers. I don't have tmc2208 and tmc2660 however. On that note, how does the tmc2660 do sensorless homing with no diag pin? Or am I missing something?
As for changing `GLOBALSCALER` at runtime, I don't see a problem in that. The datasheet says that the automatic stealthchop parameters won't get updated until the motors move again, but that's as expected as when the driver is powered up for the first time (or enabled via de nEN pin). I wouldn't worry too much about this. In any case, changing GLOBALSCALER at runtime is mandatory for homing_current.

> If we're going to introduce a homing_current config option I think it should apply to all homing with that driver - not just sensorless homing. Also, in my tests, changing the current while the carriage was pressed against the rail resulted in poor sensorless homing behaviour. This was discussed briefly at TMC2130, TMC2209, TMC5160 homing current #1957 (comment) and is mentioned at https://www.klipper3d.org/TMC_Drivers.html#sensorless-homing .

Why would anyone need different homing current if they are using something different than sensorless homing? I don't get this.

> Why is the homing current handled directly in TMC5160CurrentHelper() and TMCCurrentHelper() instead of in TMCCommandHelper() or TMCVirtualPinHelper?

I read this from the thread linked above and noticed this: https://github.com/KevinOConnor/klipper/pull/1957#issuecomment-539550192 . Wouldn't this be exactly what I tried to do? Please correct me if I'm wrong. With my approach the different registers for current are modified for each driver type individually (ie, `vsense` for tmc2130/2209 and `GLOBALSCALER` for tmc5160). As an added bonus to how it is now, changing the current of the driver via `SET_TMC_CURRENT` will update the stored current inside TMCCurrentHelper(), so any new homing moves will restore the correct current when the move finishes.

ReXT3D:
> Why would anyone need different homing current if they are using something different than sensorless homing? I don't get this.

The run current is normally selected to maximize printer performance at high speed.  Homing is normally done at slow speeds.  It is a great idea to reduce homing current to absolute minimum in order to minimize or even avoid mechanical damage to the printer in case of single point failures.  For example, one broken wire or failed switch can be enough to cause the print head to crash into the bed or the motion system to crash into the mechanical stops.

Question for Kevin - does Klipper limit the length of homing moves to **position_max** - **position_min** or are the homing moved fully "open loop" until the **endstop_pin** changes state?  Pardon my ignorance but I have not gone digging through the code.  Such a feature would provide another layer of mechanical protection during homing.

KevinOConnor:
> I propose that we name this highvelocity_threshold or something similar (the datasheet calls it high velocity mode). Maybe this name can be made a bit more readable by some camelCase. So something like highVelocity_threshold. It's uglier, but at least it's clear what it does.

high_velocity_threshold is fine.  Separately, is there anyone that is actively planning to use this capability?

> In any case, TCOOLTHRS is set on __init__ by TMCTCOOLTHRSHelper(). it defaults to 0xFFFFF just like it is needed for sensorless homing. No need to set it to 0 when homing is not performed.

Okay, but does that mean "coolstep" is unconditionally enabled?  If so, that would be a notable behavior change and I'm not sure it would be correct.  (As the specs warn against enabling coolstep at low velocity.)

> Why would anyone need different homing current if they are using something different than sensorless homing? I don't get this.

To be clear, it's fine with me if the feature is for sensorless homing only.  However, in that case I think the config name should be something like `sensorless_homing_current`.  That is, I find the config name to be confusing - not the functionality.

Separately, there is still the issue of the sensorless homing current being increased at the wrong time (while the carriage is pressed against the rail end).

> I read this from the thread linked above and noticed this: #1957 (comment) . Wouldn't this be exactly what I tried to do? Please correct me if I'm wrong. With my approach the different registers for current are modified for each driver type individually (ie, vsense for tmc2130/2209 and GLOBALSCALER for tmc5160). As an added bonus to how it is now, changing the current of the driver via SET_TMC_CURRENT will update the stored current inside TMCCurrentHelper(), so any new homing moves will restore the correct current when the move finishes.

I'm not sure I understand.  The code here adds a `set_homing_current()` method to both the tmc2130 and tmc5160 code.  My question is, is there a reason why TMCVirtualPinHelper doesn't call the existing `self.current_helper.get_current()`, set the homing current with `self.current_helper.set_current()`, and then restore the previous current after homing with `self.current_helper.set_current()`?

Thanks,
-Kevin

KevinOConnor:
> does Klipper limit the length of homing moves to position_max - position_min or are the homing moves fully "open loop" until the endstop_pin changes state?

It depends on the kinematics, but in most cases it is `1.5 * (position_max - position_min)`.

-Kevin

leptun:
Coolstep is enabled by SEMIN, which defaults to 0 (disabled), so TCOOLTHRS by default only influences stallguard min velocity threshold.

Btw, I have some exams right now, so please wait a few more days for the changes.

leptun:
Out of curiosity, is there such a thing in klipper as aligning to fullstep after homing for a more precise home? I know prusa firmware does this and it moves away from the endstop after homing.

As for skipping a step when homing while already at home position, would it be possible to specify a certain distance to move away from the endstop before homing? Again, PrusaFirmware moves away something like 1-2mm before homing. It looks weird, but it makes stallguard homing work reliably. Together with a reduced homing current, it might actually be doable without risking damage if the oposite end of the axis is touched just a tiny bit. Of course, such a feature would be optional, but useful in most cases.

KevinOConnor:
> Coolstep is enabled by SEMIN, which defaults to 0 (disabled), so TCOOLTHRS by default only influences stallguard min velocity threshold

Ah, okay.

Shouldn't the code still set TCOOLTHRS=0xfffff during sensorless homing though?

> Out of curiosity, is there such a thing in klipper as aligning to fullstep after homing for a more precise home? I know prusa firmware does this and it moves away from the endstop after homing.

There is the endstop_phases module ( https://www.klipper3d.org/Endstop_Phase.html ) and that module can assist in arranging for Z layers to always be on a full step position.  (Not sure if that's what you're asking though.)

> As for skipping a step when homing while already at home position, would it be possible to specify a certain distance to move away from the endstop before homing? Again, PrusaFirmware moves away something like 1-2mm before homing. It looks weird, but it makes stallguard homing work reliably. Together with a reduced homing current, it might actually be doable without risking damage if the oposite end of the axis is touched just a tiny bit. Of course, such a feature would be optional, but useful in most cases.

It's possible to change the code to do that.  The difficulty is that Klipper supports many different printers and it's not clear if that's a good default in general.  Currently we're documenting that macros should be used to accomplish that task - https://www.klipper3d.org/TMC_Drivers.html#using-macros-when-homing .

To be clear, the issue I've seen is skipping steps after homing - not before.  That is, the carriage moves to the end of the rail, raises stallguard flag, stepper stops, motor current is increased, and then the carriage moves because of an increase in current.  In particular, on corexy.

Separately, in case anyone is curious, moving after homing also comes up in #3956 where we have to account for overshoot during homing due to comms delays.

Cheers,
-Kevin

ReXT3D:
> > does Klipper limit the length of homing moves to position_max - position_min or are the homing moves fully "open loop" until the endstop_pin changes state?
> 
> It depends on the kinematics, but in most cases it is `1.5 * (position_max - position_min)`.
> 
> -Kevin

Ah, wonderful thank you.  To be honest this is what I expected based on my experience with Klipper.

leptun:
> Shouldn't the code still set TCOOLTHRS=0xfffff during sensorless homing though?

If `TCOOLTHRS` is set on startup to the desired value, then there is no need to update the register before and after homing. The purpose of setting something lower than 0xfffff is that during the acceleration and deceleration ramps, the stallguard might trigger too early (at really low speeds), which is undesirable. Setting it to something equivalent to let's say 20mm/s when you home at 50mm/s will make using lower homing current and more sensitive SGT values without generating false triggers. This could be useful for coreXY printers with long belts as the stallguard could trigger earlier.

> There is the endstop_phases module ( https://www.klipper3d.org/Endstop_Phase.html ) and that module can assist in arranging for Z layers to always be on a full step position. (Not sure if that's what you're asking though.)

That's pretty neat. This is what I was asking about, but I have a question about it: Does it also move to the fullstep or just silently adjust the "current_position" after home? It would be neat if there was an option to actually move to a fullstep away from the endstop after homing. It could be enough to release the tension on the belts. Maybe have the option to specify the minimum distance to move after home and then align to fullstep (to handle the weird case where moving to the first fullstep moves very little).

> Separately, in case anyone is curious, moving after homing also comes up in #3956 where we have to account for overshoot during homing due to comms delays.

It would be nice if something like this could be implemented for sensorless homing where the "delay" is caused by the tesnion in the belts "slowly" (relatively speaking :) ) building up.

KevinOConnor:
> The purpose of setting something lower than 0xfffff is that during the acceleration and deceleration ramps, the stallguard might trigger too early (at really low speeds), which is undesirable. Setting it to something equivalent to let's say 20mm/s when you home at 50mm/s will make using lower homing current and more sensitive SGT values without generating false triggers.

Interesting.  A bit dangerous though - if the carriage is close to the endstop near the beginning of homing it will bang against the carriage until it gets to full speed, and if the user sets a stallguard_threshold speed higher than their homing speed then it'll bang repeatedly against the end of rail.

> Does it also move to the fullstep or just silently adjust the "current_position" after home?

It adjusts Klipper's internal tracking of the current position to account for the over/undershoot.  (As does #3956 for its overshoot.)

> It would be neat if there was an option to actually move to a fullstep away from the endstop after homing.

It's possible to change the code to do that.  I don't think one full step is sufficient (on my Voron2 I've tested Z max sensorless homing and it's necessary to backoff further to release the belt tension).  Across all printers, it's unclear to me if backing off is, in general, a good default.  It may be.

-Kevin

MechanizedMedic:
Hi, I wanted to pop in and say thank you to @leptun for implementing this. I've spent the last few days trying to hack something like this together... My printer is driven by 5mm pitch lead screws so CoolStep/dcStep should make a big difference in peak performance.

Thanks again!
~Matt

leptun:
> Interesting. A bit dangerous though - if the carriage is close to the endstop near the beginning of homing it will bang against the carriage until it gets to full speed, and if the user sets a stallguard_threshold speed higher than their homing speed then it'll bang repeatedly against the end of rail.

Yes, it might bang at the beginning. But that’s where the “move a bit away before homing” comes into play. Not to mention that even without tcoolthrs adjustments, the axis still skips if it is already at the home position and you try to home again. Using tcoolthrs is optional. If people know what they want from their drivers, they should be able to use it. And having a helper for the velocity to tcoolthrs is nicer than writing the raw register value in the config.
There should be a check for the tcoolthrs velocity to be below homing velocity by at least a small arbitrary margin, but I didn’t know how to handle kinematics other than cartezian where there is compound stepper movement (or rotation for polar)

KevinOConnor:
> Using tcoolthrs is optional. If people know what they want from their drivers, they should be able to use it.

FWIW, it's very subtle.

I recommend changing the first commit to backup and restore TCOOLTHRS in sensorless homing.  If desired, have a separate commit introduce something like `sensorless_homing_stallguard_threshold`.  I think it would also be worthwhile to get feedback from a bunch of testers on how that setting impacts their various printers (eg, how much does it increase the viable "sensitivity range").

Cheers,
-Kevin

leptun:
Ok. I will have the commits split and the config name changed. In the meantime, observe this:

This is the most sensitive I can go before it starts being unreliable. It barely makes any sound when it touches the axis end, which is really nice tbh.
homing_current: 0.15
driver_SGT: -2
homing_speed: 50
sensorless_homing_stallguard_threshold: 20
https://photos.app.goo.gl/v8vYJbWS7GMDYbwx9
If I wouldn't use TCOOLTHRS, then the axis would instantly trigger the endstop as soon as any movement occured because of the reduced current and the very low SGT value.
Please ignore the extra tiny movement after the homing move completes. It is not a skipped step. I manually moved a full step away from the 0 position.

Here is a prototype of the homing belt detensioning. It was set to move 2mm (defaults to 0mm) at a configurable speed (defaults to homing_speed). That is an extreme value, but it's easier to see than a fullstep: https://photos.app.goo.gl/AERKcB6NrMsdYVSU9
After the detensioning is done, the endstop position is applied, so you can't tension the belt again unless you go out of bounds somehow.

KevinOConnor:
Great.  I agree it seems like a useful feature.  I don't think it will be difficult to get testers.

Thanks,
-Kevin

ReXT3D:
If your commits make the value of homing_current available during all homing (not just sensorless), you will have many people, including me of course, eternally grateful!

leptun:
@KevinOConnor How does this look for the belt detensioning?
https://github.com/KevinOConnor/klipper/compare/master...leptun:belt_detensioning?expand=1

leptun:
@ReXT3D @MechanizedMedic What do you say of these homing_current changes? They work for all tmc drivers. Tested them in my tmc2130 setup. Maybe you can test these changes with some other drivers as well. I might be able to test tmc2209 today as well.

@KevinOConnor I rewrote most of the commits. Did you see anything wrong? (especially in the last commit). I left a bit of debugging info in that commit. It can be removed if you confirm the rest is ok.

ReXT3D:
@leptun I did a simple test with TMC2209 and I can confirm that the homing current appears to be set correctly prior to each axis homing and then restored correctly following the homing.

MechanizedMedic:
> @ReXT3D @MechanizedMedic What do you say of these homing_current changes? They work for all tmc drivers. Tested them in my tmc2130 setup. Maybe you can test these changes with some other drivers as well. I might be able to test tmc2209 today as well.

@leptun I _really_ like these changes! I'll give it a whirl this weekend and let you know how it goes. I also have TMC2130s but setup driving stiff ball-screws, so should be a good stress test. :)

MechanizedMedic:
@leptun & @KevinOConnor,

I had the chance to test all of these new goodies over the weekend and did not experience any bugs. Here is my 
[tmc-example.cfg](https://github.com/KevinOConnor/klipper/files/6773524/tmc-example.cfg.txt)
 if y'all would like to audit my settings. It may be of note that I do not enable steathChop due to the demands of my lead screw setup.

The only issue I encountered was some confusion with the name/function of "sensorless_homing_stallguard_threshold" as it does not adjust the StallGuard threshold (SGT). If this parameter is going to stay then it might be better named as "sensorless_homing_coolstep_threshold" to match "coolstep_threshold", since they both deal with TCOOLTHRS.  Also, please consider automatically setting TCOOLTHRS to be offest below "homing_speed" (or set to 0) during homing moves instead of having users specify a velocity.

I do believe there is a strong case for a true "sensorless_homing_sgt" to be added to this PR. SGT is affected by changes in current, so using an SGT value that is calibrated to "homing_current" is likely to interfere with stallGuard operation when at "run_current"... I do plan on using StallGuard during normal operation in addition to sensorless homing. Recent PR #3974 added a mechanism to poll TMC drivers for errors and I intend on adapting this to check for missed steps then initiate rehoming.

Anyhoo, another big thanks to @leptun and @KevinOConnor - this PR has my printer performing well above what I anticipated. Enabling dcStep has increased my max velocity from 120 to 250 mm/s. I can now leave run_current at 1.2A without overheating due to the current scaling of coolStep. These features should be very nice for anyone looking to get more performance and/or reliability from their Trinamic drivers.

Cheers,
Matt

KevinOConnor:
Thanks.  Some feedback:
1. First commit ( tmc: Expanded config) looks fine.  Just needs the changes to docs/Config_Reference.md .
2. Can you expand on the second commit ( tmc: Save and restore TPWMTHRS during homing for both stealthChop modes  ) - why alter TPWMTHRS if it's not necessary?  This may be confusing to users that alter that field at runtime via `SET_TMC_FIELD`.
3. The third commit ( tmc: sensorless_homing_stallguard_threshold ) looks fine to me.  However, it would need to be documented in docs/Config_Reference.md .  It would also be useful to explain how it should be used in docs/TMC_Drivers.md .
4. The fourth commit ( tmc: Handle TMC5160 GLOBALSCALER correctly ) looks fine to me.  I think it would be worthwhile to add a note to docs/Config_changes.md though - so as to alert users that the behaviour of SET_TMC_CURRENT has changed on tmc5160.
5. I'm unsure on the final commit and I'm currently think it is not a good match for the Klipper master repo.  My main concern is that its primary use case is with sensorless homing, and using it that way has subtle flaws (changing current while the carriage is pressed against the end of the rail).  I'm reluctant to maintain/document an option and then basically say "don't use this option".

> How does this look for the belt detensioning?

Changing the toolhead position at that point would break the endstop_phases module, as that module current expects the "homing:home_rails_end" event to occur while the rails are at their endstop positions.

At a high-level though, my feedback is that I'm unsure adding homing_detension_speed, homing_detension_dist, homing_current is an improvement over the current method of defining macros ( https://www.klipper3d.org/TMC_Drivers.html#using-macros-when-homing ).  I regularly receive feedback that the Klipper config is complex and daunting for new users.  I suspect users may find it easier to explicitly state the steps to take during homing than to read, understand, and test the various config options needed to take those steps.

Thanks again for looking at this.
-Kevin

Sineos:
>  I regularly receive feedback that the Klipper config is complex and daunting for new users.

Sorry, I just needed to reply to this:
Whoever says this never has configured and build a vanilla Marlin

BlackStump:
Or RRFirmware with Gcode everywhere in the config, imho Klipper config is simpler

leptun:
@KevinOConnor 
1) Yes, I will document all the new fields.
2) I think I can discard this commit. I'm not entirely sure what was my idea behind it. Looking again at the code, changing the TPWMTHRS to 0xFFFFF and then restoring it changes nothing on the tmc2130/tmc5160 drivers since the `en_pwm_mode` is set to 0.
3) Yes, I will write documentation. Does it need to be the same commit or can I add new commits for documentation?
4) I'll edit Config_changes.md and mention that the TMC_SET_CURRENT command was not working properly for the tmc5160. and that now it should work.
5) I was hoping this could be used alongside the belt detensioning after I fix the `endstop_phases` module so that this current changing while pressed flaw can be eliminated.

I don't want to do the belt detensioning and the home current with macros since I was unable to find a way to make homing individual axis possible with the edited scripts. Let's say I use [homing_override], set axes to xyz and write a macro that does the homing for all axes that also adjusts currents before the move. How exactly would I be supposed to make the script know that I want to home the X axis only for example and not home all axes as the script in homing_override would dictate?
As for making the config overcomplicated, I'm not sure that I agree with this. I think all kinematics would benefit from such a feature and having the user find macros online for homing is not exactly user friendly. Having something builtin (not enabled by default) would allow users to just find the option in the documentation and use it accordingly. Not to mention that when you use [homing_override], some of the homing parameters for the steppers get overridden or ignored altogether (which is not documented).

I'm still hoping that I didn't break other kinematics with the homing changes. Cartesian printers seem to work, but I still have to somehow test CoreXY and maybe even Delta as both home with more than one stepper at the same time. I hope that the `TMCHomingCurrentHelper` sets and restores the current for all the steppers that are involved in a homing movement (say X and Y for X/Y homing on CoreXY or all 3 steppers on a Delta).

KevinOConnor:
> Does it need to be the same commit or can I add new commits for documentation?

Either is fine.  I tend to put my changes in a single commit.  It should be in the same PR.

> I don't want to do the belt detensioning and the home current with macros since I was unable to find a way to make homing individual axis possible with the edited scripts.

I understand.  However, my current high-level feedback is that I'm not sure that change is worthwhile.  We can certainly revisit - getting feedback from many different users with different types of printers would help.  In any case, I suspect it is unlikely something like that would be merged in the next 6 months or so.

-Kevin

KevinOConnor:
Is there interest in merging the config, TPWMTHRS, and tmc5160 parts of this PR?  Those changes look fine to me.

Thanks,
-Kevin

zellneralex:
Yes at least me would be interested in that

PR #4431, opened 2021-06-27
resonance_tester: Arbitrary AXIS in TEST_RESONANCES command

dmbutyugin:


KevinOConnor:
Thanks.  Looks good to me.  Unless anyone screams, I'll give a few days and then commit.

-Kevin

charlespick:
Can this be used to input shape an idex printer?

eddietheengineer:
I tested this branch and it worked well!

dmbutyugin:
@eddietheengineer OK, that's great to hear!

@charlespick in general, I suspect it should be possible to run input shaper on an idex printer. I think all the tooling is in place to do that. Though the tuning process may get a bit more elaborate. In fact, it would be great to have a section in one of the docs describing the suggested setup and the tuning process. Sadly, I don't own one, so I can try to write one, but I cannot verify myself that it works as expected. That said, if the question is about this PR, then no, this PR does not help with that.

PR #4439, opened 2021-06-28
tuning_tower: Add STEP and SKIP parameters as alternate tuning tower syntax

matthewlloyd:
This change is fully backwards compatible, and merely adds alternate syntax that is easier to configure for certain types of tuning tower. Proposed here: https://klipper.discourse.group/t/feedback-easier-tuning-tower-syntax/438

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -213,13 +213,17 @@ The following standard commands are supported:
   This acts to take a frequently used babystepping value, and "make
   it permanent".  Requires a `SAVE_CONFIG` to take effect.
 - `TUNING_TOWER COMMAND=<command> PARAMETER=<name> START=<value>
-  FACTOR=<value> [BAND=<value>]`: A tool for tuning a parameter on
-  each Z height during a print. The tool will run the given COMMAND
-  with the given PARAMETER assigned to the value using the formula
-  `value = start + factor * z_height`. If BAND is provided then the
-  adjustment will only be made every BAND millimeters of z height - in
-  that case the formula used is `value = start + factor *
-  ((floor(z_height / band) + .5) * band)`.
+  FACTOR=<value> [BAND=<value>] [SKIP=<value>]`: A tool for tuning
+  a parameter on each Z height during a print. The tool will run
+  the given COMMAND with the given PARAMETER assigned to the value
+  using the formula `value = start + factor * z_height`. If BAND is
+  provided then the adjustment will only be made every BAND
+  millimeters of z height - in that case the formula used is
+  `value = start + factor * ((floor(z_height / band) + .5) * band)`.
+  If `SKIP=<value>` is specified, the tuning process doesn't begin
+  until Z height `<value>` is reached, and below that the value will
+  be set to `START`; in this case, the `z_height` used in the
+  formulas above is actually `max(z - skip, 0)`.
 - `SET_DISPLAY_GROUP [DISPLAY=<display>] GROUP=<group>`: Set the
   active display group of an lcd display. This allows to define
   multiple display data groups in the config,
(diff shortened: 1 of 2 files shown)


charlespick:
I like the idea of skipping a bit of the print but I think additional parameters that simply replace existing ones by adding some math should rather be added using a gcode macro with params

Sineos:
I appreciate the idea. 
While the current implementation is logical and flexible, it is complex for the user. 

Telling the user to deploy a "complex" macro to hide the complexity of the core command makes absolutely no sense in my eyes unless I would loose flexibility in the core command. Maybe I'm missing the some aspects but I do not see any loss of flexibility in the proposed change. 

I personally have created a spread-sheet to avoid thinking through the logic over and over again whenever I use the command.

KevinOConnor:
Thanks.  At a high-level, I'm okay with the PR, but I would request some changes:
1. Separate out the SKIP code to its own commit, as that functionality seems logically separate from STEP.
2. I find the interaction between FACTOR, BAND, and STEP to be confusing.  I suggest leaving FACTOR unchanged, and allow STEP to be a replacement for BAND (that is, BAND is unchanged, but with STEP the formula doesn't do the half-step rounding).

-Kevin

KevinOConnor:
Any further updates on this?

-Kevin

crof2003:
I was just running into similar confusion today that would have been easier with the changes in this PR.

Assuming the original author has abandoned these changes, what's the etiquette here?  Leave it as abandoned?  Or can someone (aka me) pick it up and run with it?  Make another fork/copy the existing changes/incorporate the requested updates from the repo owner/create a new PR?  Or some other process?

I've never contributed to an open source project before, but would love to help get these changes across the finish line if I can...

matthewlloyd:
Not abandoned, just got distracted building my Voron V2.4 :) Give me a few days and I'll make some progress.

> * Separate out the SKIP code to its own commit, as that functionality seems logically separate from STEP.

Will do. I'll arrange them as separate commits in this PR.

> * I find the interaction between FACTOR, BAND, and STEP to be confusing.  I suggest leaving FACTOR unchanged, and allow STEP to be a replacement for BAND (that is, BAND is unchanged, but with STEP the formula doesn't do the half-step rounding).

I'm not clear on what you are suggesting as alternate syntax. There is a fundamental change in semantics here, which is that the user wants to specify the rate of change as per band, rather than per mm, so the FACTOR argument no longer applies.

Let's take a canonical example - the infamous gaaZolee temperature tower (https://www.thingiverse.com/thing:2729076). As this PR currently stands, the parameters for this tower can be determined directly from looking at the image (starts at 225C, decreases by 5C each band) and the description (one floor is 10mm, stand is 1.4mm):

```START=225 STEP=5 BAND=10 SKIP=1.4```

Kevin, what specific parameters are you proposing for this tower with your suggestion?

KevinOConnor:
> There is a fundamental change in semantics here, which is that the user wants to specify the rate of change as per band, rather than per mm, so the FACTOR argument no longer applies.

Okay, thanks.  I think I understand.  Perhaps we can choose two entirely new parameter names then?  Maybe something like `STEP` and `STEP_HEIGHT`?  Not sure.

-Kevin

Sineos:
`STEP` and `INTERVAL`?
or `STEP` and `EVERY_MM`?
Instead of `STEP` use `DELTA`?

KevinOConnor:
It's fine with me.

-Kevin

matthewlloyd:
Done. I've also rebased on latest master, and put SKIP into a separate commit as requested.

KevinOConnor:
Thanks.

-Kevin

Markus-1984:
Hi,

I´ve got an Error when start the Tuning Tower Command by this:
TUNING_TOWER COMMAND=“SET_HEATER_TEMPERATURE HEATER=extruder” PARAMETER=TARGET START=250 STEP_DELTA=-5 STEP_HEIGHT=5 SKIP=1.2

The Error Message is:
Error on "TUNING_TOWER COMMAND=“SET_HEATER_TEMPERATURE HEATER=extruder” PARAMETER=TARGET START=250 STEP_DELTA=-5 STEP_HEIGHT=5 SKIP=1.2": missing FACTOR

If I understand the Command right, I don´t need a FACTOR if I use STEP_DELTA and STEP_HEIGHT.
Am I wrong?

PR #4447, opened 2021-07-01
controller_fan: Add "stepper" config option

LadyBoonami:
Allows contoller_fan sections to monitor only certain steppers instead of
all of them, similar to how heaters are currently handled. The default is
to monitor all steppers except extruders for backwards compatibility. Note
that this module (accidentally?) never actually monitored extruder
steppers.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2374,7 +2374,7 @@ a shutdown_speed equal to max_power.
 
 Controller cooling fan (one may define any number of sections with a
 "controller_fan" prefix). A "controller fan" is a fan that will be
-enabled whenever its associated heater or any configured stepper
+enabled whenever its associated heater or its associated stepper
 driver is active. The fan will stop whenever an idle_timeout is
 reached to ensure no overheating will occur after deactivating a
 watched component.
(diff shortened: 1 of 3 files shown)


charlespick:
I think it might be worth adding extruder steppers to the mix but maybe there's a reason they were never part of it 🤷‍♂️

LadyBoonami:
Not sure if it was intentional. The `kin.get_steppers()` call (obviously) returns only the steppers known to the kinematics subsystem. From my poking around, it seems that querying extruder steppers is a bit more involved, and manual steppers would maybe also need to be considered? Again, the original code does none of these, that's why I didn't try to hack something up. I guess nobody noticed the discrepancy between documentation and behaviour because extruder steppers are usually only running when a heater is active, which would enable the associated fan...

KevinOConnor:
Thanks.  In general the change looks fine to me.  However, it does seem a bit odd to check the kinematic steppers when the `stepper_enable` module (used later in controller_fan) has the full list of possible steppers.

Also, just for my understanding, can you describe the high-level use case for this change?

Cheers,
-Kevin

LadyBoonami:
I will look into that.

Use case: my carriage is so heavy that it drops as soon as the Z stepper is powered down, so I keep that one enabled at all times. But that enables my case fans, which is unnecessary for the tiny load. I'd like to specify "track every stepper except Z" for the fans, so I can save power, prevent fan wear, and most importantly keep the room quiet.

LadyBoonami:
The code is now tracking all steppers. I had to add a `get_steppers` function to `stepper_enable.py` to avoid accessing internal members from another module, but the behaviour should be more consistent now.

KevinOConnor:
Thanks.  I committed this change with some minor updates on top (commit 103dcf27).

-Kevin

LadyBoonami:
Very nice, thanks =)

charredchar:
I've been using the `controller_fan` to use the `idle_timeout` for a `heater_generic` to have it stay on for a couple minutes after the heater turns off. Something I found out using this though is that if `stepper` isn't define it defaults to all but there is no option for `none`. If `heater_fan` has an `idle_timeout` as well we wouldn't need to use `controller_fan` for this but it might be good to also allow defining `none` for `stepper`.

PR #4450, opened 2021-07-02
docs: Measureing resonances

charlespick:
Adds a script to make setting up auto calibration slightly easier, fixes some typos, and clarifies a few things. 

1 commit was...
Signed-off-by: Charles Pickering <charles.pickering@live.com>


docs/Measuring_Resonances.md
```diff
@@ -65,36 +65,28 @@ the system that may damage the electronics.
 
 ## Software installation
 
-Note that resonance measurements and shaper auto-calibration require additional
-software dependencies not installed by default. First, you will have to run on
-your Raspberry Pi the following command:
+Because setting up the software for auto-calibration takes significant
```
dmbutyugin, line 68: As a small nit, it is also possible to just measure the resonances of the printer without doing shaper auto-calibration. And the doc name is Measuring_Resonanced.md. So I would update the changes (and perhaps the script name) to reflect that (e.g. here "for resonance measurements").

docs/Measuring_Resonances.md
```diff
@@ -65,36 +65,28 @@ the system that may damage the electronics.
...
-sudo apt install python-numpy python-matplotlib
-```
-
-Afterwards, check and follow the instructions in the
-[RPi Microcontroller document](RPi_microcontroller.md) to setup the
-"linux mcu" on the Raspberry Pi.
-
-Make sure the Linux SPI driver is enabled by running `sudo
-raspi-config` and enabling SPI under the "Interfacing options" menu.
+If you intend to connect the ADXL345 to your Raspberry Pi, check and follow the
```
dmbutyugin, line 80: TBH, the docs don't describe other possibilities in details. It's on my TODO-list to write a short section describing the possibility of connecting the ADXL345 to the printer board with potential pitfalls and gotchas, but it doesn't exist yet :) And the way it is written now may leave the reader puzzled (Like, oh, do I? But what are other options? But the docs don't describe other possibility). If you want to make it explicit, maybe it's worth mentioning that this is the suggested setup? E.g.

If you intend to connect the ADXL345 to your Raspberry Pi, which is the suggested setup, ...

docs/Measuring_Resonances.md
```diff
@@ -65,36 +65,28 @@ the system that may damage the electronics.
...
+`sudo raspi-config` and enabling SPI under the "Interfacing options" menu.
 
 Add the following to the printer.cfg file:
 ```
-[mcu rpi]
-serial: /tmp/klipper_host_mcu
-
 [adxl345]
-cs_pin: rpi:None
+cs_pin: AB0 # The CS pin on your sensor that may be connected to your
```
dmbutyugin, line 88: The big advantage of the previous 'rpi:None' was that this part was copy-pasteable. Maybe it's indeed better to remove [mcu rpi] section, but could you please keep 'rpi:None' as the default option here? Otherwise, the correct way to define the adxl345 connected to the RPi is not defined anywhere. And that 'rpi:None' syntax is very much not obvious. The comment is fine though.

scripts/setup-shaper-calibrate.sh
```diff
@@ -0,0 +1,10 @@
+# This script instals the prerequisites for measureing resonances on a standard, single instance install of Klippy
+
+# Step 1: Install numpy in the virtualenv so it can be used by Klippy
```
dmbutyugin, line 3: It might be great to actually echo these messages to the command line. Otherwise, almost no-one would read these comments.

scripts/setup-shaper-calibrate.sh
```diff
@@ -0,0 +1,10 @@
+# This script instals the prerequisites for measureing resonances on a standard, single instance install of Klippy
```
dmbutyugin, line 1: I imagine this script needs a header (Shebang).

dmbutyugin (commented):
Thanks! I got a few minor comments.

docs/Measuring_Resonances.md
charlespick: Yes, however the existing instructions have you verify the sensor is working with `MEASURE_AXES_NOISE` before taking measurements which requires `numpy`

docs/Measuring_Resonances.md
charlespick: I think I'll add that a pi is the suggested solution for a permanent installation with the possibility to use your primary mcu as well.

scripts/setup-shaper-calibrate.sh
charlespick: Good idea

docs/Measuring_Resonances.md
charlespick: the old documentation is now back in it's own section

docs/Measuring_Resonances.md
```diff
@@ -65,29 +65,29 @@ the system that may damage the electronics.
...
+too little RAM, the installation may fail, and you will need to
+[enable swap](https://nebl.io/neblio-university/enabling-increasing-raspberry-pi-swap/).
 
-Next, run the following commands to install the additional dependencies:
-```
-sudo apt update
-sudo apt install python-numpy python-matplotlib
-```
+## Hardware Configuration
+You can connect your ADXL345 to any spi bus supported by Klipper. Regular printer
```
dmbutyugin, line 81: Unfortunately, "to any spi bus supported by Klipper" is not really correct. This is because ADXL345 is itself non-compliant to the SPI protocal, and therefore it cannot share a bus with any other device - otherwise there is a risk it will interfere with the communications with that device. On Raspberry Pi this is almost never an issue - there is almost never anything else connected to the SPI bus. However, on the printer MCU there is often something else - SD Card reader, SPI bus is shared with the TMC drivers, SPI goes to screen (where is also an SD Card reader), etc. Connecting ADXL345 to such SPI bus will lead to a confusing and surprising behavior. For example, connecting to the SPI bus shared with SD Card typically leads to boot failures and inability to install a firmware update (from the SD Card) for the boards that use SD Card method. Then, another big consideration is 5V tolerance. Most Chinese boards only accept 5V _as a power source_. They are not 5V tolerant on other pins, and connecting them to many printer MCU boards will simply fry them. ADXL345 from Adafruit and SparkFun have logic-level shifter built-in, so they are fine, but depending on the MCU, it could probably get the other way around and damage the MCU in turn (if the MCU SPI pins are not 5V tolerant). Overall, this is a complicated topic, and I see connecting the ADXL345 to a printer MCU only suitable for the expert users who can figure out those nuances. Otherwise, it's really playing a russian roulette - either it works and fine, or you fry some of your electronics.

So, I cannot recommend adding this and the MCU sections to the Klipper documentation as it is. These things need to be mentioned explicitly. Note that I was planning to write on this topic to the Klipper documentation, but didn't get to it yet (and I wouldn't be able to commit to some very near time). I'm more than happy if somebody else writes that, but I think those considerations and gotchas above need to be explicitly written to help users avoid damaging their hardware.

docs/Measuring_Resonances.md
```diff
@@ -102,6 +102,38 @@ accel_chip: adxl345
 probe_points:
     100,100,20  # an example
 ```
+
+### Primary MCU
+Add the following to the printer.cfg file:
+```
+[adxl345]
+cs_pin: AB0 # The MCU pin connected to the CS pin on your sensor
```
dmbutyugin, line 110: In my experience, it is frequently necessary to also specify spi_bus. Unfortunately, what's currently lacking in Klipper documentation is a list of the possible values for this parameter for the supported MCUs. So I had to dig up the names from the source code directly. Maybe there is something available? Even if not, it would probably make sense to mention this requirement to the users and direct them in the right direction how to figure out the appropriate values.

docs/Measuring_Resonances.md
```diff
@@ -65,29 +65,29 @@ the system that may damage the electronics.
...
+their own advantages and disadvantages.
 
-Afterwards, check and follow the instructions in the
-[RPi Microcontroller document](RPi_microcontroller.md) to setup the
-"linux mcu" on the Raspberry Pi.
-
-Make sure the Linux SPI driver is enabled by running `sudo
-raspi-config` and enabling SPI under the "Interfacing options" menu.
+### Raspberry Pi MCU
+The reccomended way to connect the sensor to the printer in a permenant installation
```
dmbutyugin, line 86: A couple of typos here: 'reccomended' and 'permenant'. BTW, I wouldn't require a permanent installation. I'd just write something along the lines of "The recommended way to connect the sensor to the printer is to connect it directly to the SPI bus on the Raspberry Pi, as per [wiring](#wiring) instructions. Then, check and follow ..."

BTW, "## Wiring" part only describes wiring to the RPi. If you want to keep the sections about connecting the ADXL345 to the printer MCUs, I'd update that section too somehow. Otherwise, the user may be confused as to how the ADXL345 must be physically connected to the MCU.

docs/Measuring_Resonances.md
dmbutyugin: Yes, of course. But sorry, I did not understand your argument. I agree that everything is needed even for the basic `MEASURE_AXES_NOISE` test.

Maybe I wrote a bit confusingly, I was simply saying that this software is needed not just for auto-calibration (aka `SHAPER_CALIBRATE`), but for any measurement, and I was suggesting to rephrase the sentence to something like:

Because setting up the software for resonance measurements takes significant ...

dmbutyugin (commented):
As a high-level feedback, it seems that this PR now contains two parts. One part is adding a convenience script to install the software and updating the documentation to use that script. + fixing some typos. And then there is another part about connecting the ADXL345 to the printer MCU(s). That part deals with rather complicated topic. I would advice to split the PR into the two PRs accordingly: the first part is pretty much a no-brainer for integration, and I think it can be merged smoothly, because all the concerns there have been addressed. The second part requires some more work, and I wouldn't block merging the first part just because of the second part. But that's up to you.

docs/Measuring_Resonances.md
charlespick: I'll add a reference to the SPI section of the config document. I think this line settles it: 
```
#   If the micro-controller supports multiple SPI busses then one may
#   specify the micro-controller bus name here. The default depends on
#   the type of micro-controller.
```

charlespick: If it's worth it, I think the names of available spi buses should be included in config files for boards

docs/Measuring_Resonances.md
charlespick: I had no idea that the adxl345 had weird spi. :( I think the smart option at this point is to then have separate sections for RPi, secondary MCU, and primary MCU. Each would have it's own wiring directions, notes about power, and config snippets. RPi would be the recommended way, with secondary mcu recommended for if you don't want to install the linux mcu or you have something using your pi's gpio pins, and primary mcu for advanced users.

docs/Measuring_Resonances.md
charlespick: what I mentioned in the comment on line 81 should address this

docs/Measuring_Resonances.md
dmbutyugin: Yep, unfortunately, it is like that. To be fair, primary and secondary MCU are almost equally tough, maybe with just less stuff connected to the secondary MCU. But it is less common scenario to have that secondary MCU in the first place. So we could have 2 different sections for primary/secondary MCUs, but it could also be a single section, because all the considerations and concerns are equally valid in both cases.

docs/Measuring_Resonances.md
dmbutyugin: Yes, maybe having that in the board config files would be the best. In fact, those configs could have examples of configs for the connected ADXL345, similar to having display sections. Then maybe the folks who successfully connected the accelerometers to their MCUs could share their findings there, e.g. which SPI buses are appropriate for connection and under which conditions, and whether the 5V-tolerant accelerometer required for a given board.

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
 
 Add the following to the printer.cfg file:
+
 ```
 [mcu rpi]
 serial: /tmp/klipper_host_mcu
 
 [adxl345]
-cs_pin: rpi:None
+cs_pin: rpi: None
```
dmbutyugin, line 95: I'm not sure about the space here. I think most of examples don't have a space between the mcu name and the pin name. So let's not introduce it here.

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
 
 [resonance_tester]
 accel_chip: adxl345
 probe_points:
-    100,100,20  # an example
+    100,100,20  # It is advised to start with 1 probe point, in the middle
+    of the print bed, slightly above it.
```
dmbutyugin, line 101: As currently formatted, it won't be copy-pastable. Could you add '#' before the continuation of the comment? And indent it to match the comment indentation on the previous line?

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
 ```
-It is advised to start with 1 probe point, in the middle of the print bed,
-slightly above it.
+
+Restart Klipper via the `RESTART` command.
+
+## Wiring Option 2: MCU Connection (Advanced Users)
+If you want to avoid installing the Linux MCU software on the Raspberry Pi, your Pi's
```
dmbutyugin, line 107: To be honest, I wouldn't list 'avoid installing the Linux MCU software' as a reason to use MCU connection. Because, frankly, the alternative is a lot more bother.

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
+GPIO is already in use, or some other reason prevents you from using option 1, you can
+connect the accelerometer to your Primary MCU, or even a secondary MCU. There are some
+caveats, however:
+* As most printer boards output 5V power for accessories, you'll likely need an ADXL345
+board with a built in voltage level shifter.
+* The ADXL345 does not adhere to the SPI spec and cannot share an SPI bus with another
+device. (Thanks @dmbutyugin for pointing that out!) The sensor will need to be the only
```
dmbutyugin, line 114: No need to give credit to me in the docs :) Also, it was actually Kevin who first pointed out this defect of ADXL345.

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
+
+Restart Klipper via the `RESTART` command.
+
+## Wiring Option 2: MCU Connection (Advanced Users)
+If you want to avoid installing the Linux MCU software on the Raspberry Pi, your Pi's
+GPIO is already in use, or some other reason prevents you from using option 1, you can
+connect the accelerometer to your Primary MCU, or even a secondary MCU. There are some
+caveats, however:
+* As most printer boards output 5V power for accessories, you'll likely need an ADXL345
+board with a built in voltage level shifter.
```
dmbutyugin, line 112: Could you please expand this sentence a bit and add another sentence here:

... board with a built in voltage level shifter, besides a voltage regulator. You must check the schematics of your MCU board, the documentation on the chip the board uses, and the schematics of your particular ADXL345 board to see if they match by voltage **on all pins**.

I just want to be crystal clear about what the users need to check.

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
+device. (Thanks @dmbutyugin for pointing that out!) The sensor will need to be the only
+device connected to its SPI bus.
+* The sensor requires a reasonably fast SPI bus
+
+Identify the MOSI, MISO, SCLK, CS, Ground, and power (see above) pins to be used for
+your sensor anc connect your sensor as follows:
+
+| ADXL345 pin | MCU |
+| :--:        | :--: |
+| 3V3/5V/VCC  | 3V3/5V/VCC |
```
dmbutyugin, line 123: Note that 3V3 pin is rarely available on the MCU boards, and then the voltages on MISO/MOSI/SCLK/CS should match the input voltage. I'd suggest to write here something like

| VCC/VIN | 5V |
| or 3V3/VIN | 3V3* |

\* If 3.3V is available on the MCU board and compatible with other pins.

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
+```
+[adxl345]
+cs_pin: AB0 # The MCU pin connected to the CS pin on your sensor
+spi_bus: # This may be necessary if your board has multiple busses available
+
+[resonance_tester]
+accel_chip: adxl345
+probe_points:
+    100,100,20  # It is advised to start with 1 probe point, in the middle
+    of the print bed, slightly above it.
```
dmbutyugin, line 145: The same about '#' comment here.

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
+caveats, however:
+* As most printer boards output 5V power for accessories, you'll likely need an ADXL345
+board with a built in voltage level shifter.
+* The ADXL345 does not adhere to the SPI spec and cannot share an SPI bus with another
+device. (Thanks @dmbutyugin for pointing that out!) The sensor will need to be the only
+device connected to its SPI bus.
+* The sensor requires a reasonably fast SPI bus
+
+Identify the MOSI, MISO, SCLK, CS, Ground, and power (see above) pins to be used for
+your sensor anc connect your sensor as follows:
```
dmbutyugin, line 119: s/anc/and/ 
(typo)

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
-Note that resonance measurements and shaper auto-calibration require additional
-software dependencies not installed by default. First, you will have to run on
-your Raspberry Pi the following command:
+Because setting up the software for measuring resonances takes significant
+time and resources, it is not installed by default. Run the following
+script to setup `numpy` and `matplotlib` to get started:
+
 ```
-~/klippy-env/bin/pip install -v numpy
+~/klipper/scripts/setup-shaper-calibrate.sh
```
dmbutyugin, line 49: Maybe also rename the script to something like `setup-accelerometer-software.sh` or `setup-accelerometer-deps.sh`?

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
+Double-check your wiring, especially the voltage tolerated by your sensor and that
+provided by it's connection, before powering up your printer to prevent damaging it
+or the accelerometer.
+
+Add something like the following to the printer.cfg file:
+
+```
+[adxl345]
+cs_pin: AB0 # The MCU pin connected to the CS pin on your sensor
+spi_bus: # This may be necessary if your board has multiple busses available
```
dmbutyugin, line 139: I think 'buses' is a bit more mainstream :)

docs/Measuring_Resonances.md
```diff
@@ -65,45 +41,116 @@ the system that may damage the electronics.
...
+
+[resonance_tester]
+accel_chip: adxl345
+probe_points:
+    100,100,20  # It is advised to start with 1 probe point, in the middle
+    of the print bed, slightly above it.
+```
+
+For more information about how to configure the SPI behavior of Klipper, review the
+[common SPI settings](config_reference.md#common_SPI_settings).
```
dmbutyugin, line 149: I believe this link and the one below don't work. I think the right references are Config_Reference.md#common-spi-settings and Config_Reference.md#mcu-my_extra_mcu

dmbutyugin (commented):
@charlespick, Thanks a lot for your efforts to improve this doc! It really looks very good now. I just had a few more minor comments, most of them being nits really.

KevinOConnor:
FYI, many of the accelerometer boards have a power regulator on them, and few seem to have level shifters.

This means, in practice, there are a few reasonable setups:
1. On a RPi mcu, it's desirable to wire 3v3 to the sensor and not use 5V at all.  This is important because 5V could easily damage the rpi or the adxl345.  All the common adxl345 boards (that support SPI in general) should work in this setup.
2. On an AVR MCU (running at 5V) you need an adxl345 board with the power regulator and level shifters.  As far as I know, this realistically means buying the genuine adafruit adxl345 sensor.  (Because, as far as I know, most clones don't deploy the level shifters.)  Otherwise, without that equipment, it's easy to damage the adxl345.  With the proper board, 5V should be wired to it, and the SPI pins wired via the level shifters.
3. On modern 32bit MCUs (running at 3.3V), you need a board with a power regulator and _without_ level shifters.  This is actually pretty common on the clones.  In this case, you can wire 5V to the adxl345 board and directly wire the SPI pins to the micro-controller.

One of the reasons it's easier to recommend the rpi as host is because it easier to document.  (Just wire 3.3V and don't worry about the rest.)

FWIW, if there is desire to expand on this, it might be worthwhile to have a separate document just on ADXL345 wiring.  As it's not that closely related to the resonance document.  (This is not a requirement for merging.)

-Kevin

dmbutyugin:
Kevin, I agree with all your points. But since there is some interest to have a documentation for ADXL345 - MCU wiring, I think we can give it a go. There is probably not enough information to justify writing a dedicated doc, and I tried to move the documentation in the direction such that MCU - ADXL345 wiring is only for advanced users who can read the schematics of the boards and the data sheet for the MCU chip and figure out the voltage requirements. There are many nuances here and no "one size fits all" recommendations, unlike the RPi setup, which is recommended.

Separately, I think it would make sense to have per-board documentation in the example configs. We have already examples of the configs (and wiring) for LCD screens, and I'm thinking that we could have the same for the accelerometers. Something like
```
# If you want to connect ADXL345 to this board, you can connect it to the SPI pins available on
# EXP_N connector as SDO - PIN_A, SDA - PIN_B, ... This board is 5V including MCU, so
# the ADXL345 with a logic level shifter is required (e.g. from Adafruit or SparkFun).
# [adxl345]
# cs_pin: PIN_CS
# spi_bus: ...
```
It is not realistic that a single person can provide the documentation for all boards, but if there is enough interest from the community, perhaps different folks could contribute their findings for various boards. I could contribute RUMBA and SKR 1.4 configs from my side. E.g. if I read the LPC1769 spec correctly,

> 3. On modern 32bit MCUs (running at 3.3V), you need a board with a power regulator and _without_ level shifters.

it seems that it is 5V tolerant on the IO pins, so boards with and without logic level shifters should work in this instance. Or maybe I misunderstood it and just got lucky. But not all MCUs are like that.

docs/Measuring_Resonances.md
charlespick: oops! forgot how comments worked for a second there

docs/Measuring_Resonances.md
charlespick: I think which is more of a bother is a matter of opinion that will change depending on who is reading the document. I prefer to run a very light linux installation because installing anything on any computer always introduces more possibility that something could go wrong. Now, is the likelihood that the linux mcu will be cause for an issue with the Klipper system down the line high? Probably not. But installing the linux mcu requires modifying your Klipper start script, and there seems to be no clear way to uninstall the linux mcu so it could also be argued that the ability to simply unplug a secondary mcu makes the linux mcu more of a bother. We've already made it a completely separate option and labeled it for advanced users, I think it's good enough.

charlespick:
It seems there's more to this whole voltage thing than I was aware of. I simply bought a board that advertised 5v compatibility on Amazon and it worked with my Arduino Uno. If more documentation is needed, could someone else add it? I'll accept pull requests to my branch and contributors can commit directly.

KevinOConnor:
What's the current state of this PR?

-Kevin

charlespick:
I'm fairly happy with it. The main point of contention has been about connecting to the MCU. I think that because of the prevalence of 5v tolerant adxl345's and it clearly being tagged as the advanced way to go about doing this with an explicit warning to check your voltages and wiring, it's safe to merge. Someone else seemed to want to add more details on that but they never did so...

KevinOConnor:
Okay, but this PR would currently regress some parts of the documentation - in particular the chapter section headers have been converted to a new format for use in mkdocs.

Also, looks like there is some unrelated change to logextract.py .  Although that change itself is fine, it's important that separate changes be done in separate commits - otherwise it causes merge conflicts and makes tools like `git revert` and `git bisect` harder to use.  I'm also not sure about adding a new script here - I don't think copy-and-pasting a few commands is more work than copy-and-pasting the command to run a script - and the actions of the script are less transparent.  Finally since @dmbutyugin wrote the initial documentation, I think it's ultimately his call on whether or not to merge this.

Thanks,
-Kevin

charlespick:
I can't figure out what is wrong with the chapter section headers, please explain

I can't remember why I decided to put the typo fix in `logextract.py` in this PR. I'm guessing there isn't a reason. I guess I can revert that diff. 

I have to say, I'm not all that enthusiastic towards getting this closed anymore. I originally wrote this because I don't like the idea of installing the linux MCU unless it's absolutely necessary, I know a lot of people have their raspberry pis deep inside their machines, and I found the documentation a little confusing the first time I followed it. But from the conversation here, it looks like these aren't really concerns so 🤷‍♂️

KevinOConnor:
> I can't figure out what is wrong with the chapter section headers, please explain

When I look at the complete changes I see the following:
```diff
+ Installation Instructions
+ ===========================
 
- ## Installation instructions
```

So, somehow this series is changing the chapter headers.

> I have to say, I'm not all that enthusiastic towards getting this closed anymore.

I understand.  FWIW, there's some good stuff in this PR - for example the grammar fixes.  Alas, there's an assortment of changes in this PR (eg, grammar fixes, section movement, script changes, config changes) and that makes it hard to review and commit.  When someone spots an issue with any of the changes it holds the whole thing up.

-Kevin

PR #4451, opened 2021-07-03
Add support for gcode:command_exec event handlers.

BlueDragonX:
This creates a new event "gcode:command_exec" which is sent on execution of each line of gcode. Handlers are able to modify or remove the line or add new commands to execute immediately after the given command. This enables a "middleware" style of gcode filtering via the extras module.

This is immediately useful for working around some slicer quirks. In particular disabling/enabling PA for multimaterial prints in SuperSlicer as the generated wipe tower gcode causes a fatal error in Klipper when PA above a certain value is used. See [superslicer_multimaterial.py](https://github.com/BlueDragonX/klipper/blob/superslicer_multimaterial/klippy/extras/superslicer_multimaterial.py) as an example.

### DIFF
--- a/klippy/gcode.py
+++ b/klippy/gcode.py
@@ -157,7 +157,20 @@ class GCodeDispatch:
     # Parse input into commands
     args_r = re.compile('([A-Z_]+|[A-Z*/])')
     def _process_commands(self, commands, need_ack=True):
-        for line in commands:
+        # Commands will be treated as a stack in order to allow exec callbacks to add new commands
+        # to the top of the stack.
+        commands.reverse()
+        while commands:
+            for cb in self.printer.event_handlers.get("gcode:command_exec", []):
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, I'm not sure this is a good candidate for the Klipper master repository.  The `_process_commands()` method is on the "fast path" and adding extra overhead to it would impact Klipper's overall performance.  Given the performance implications, it's not clear to me that having generic G-Code filtering is worthwhile.  Note that, it is possible to override existing commands using the gcode_macro capability (see docs/Config_Reference.md for details).

-Kevin

BlueDragonX:
In the case I'm trying to solve for it's not filtering specific commands, but rather it's looking for a particular comment in the file that's generated by SuperSlicer.

KevinOConnor:
Okay, I understand.  However, I don't think this is a good candidate for the Klipper master branch.

-Kevin

PR #4453, opened 2021-07-03
stm32: Fix stm32f207 USB problems.

boilerbots:
The Clock and USB setup wasn't working on stm32f207 boards. This
area is tricky to debug with gdb because it is timing sensitive. To
isolate my fixes I move more of the armcm_main() code into the processor
specific functions because I do not have the ability to test and verify
on other hardware beside stm32F207 and stm32F070.

The earlyBoot() enables the power domains for USB and Watchdog.

In watchdog_init() added a write to DBGMCU->APB1FZ to disable
the watchdog when debugging. Maybe the watchdog should be a compile
time option instead because you can't debug with it enabled.

Added scripts/elf2dfuse.py so I can reprogram the stm32f207 using dfu-util
over USB using "dfu-util -R -a 0 -d 0483:df11 -D out/klipper.dfuse"

### DIFF
--- a/scripts/elf2dfuse.py
+++ b/scripts/elf2dfuse.py
@@ -1,0 +1,107 @@
+#! /usr/bin/env python2
+from __future__ import with_statement
+
+
+# usage: elf2dfuse.py infile outfile
+
+import os, subprocess, tempfile
+import sys
+import struct
+import subprocess
+import zlib
+
+def mk_uint8(x):
+    return struct.pack("<B", x)
+
+def mk_uint16(x):
+    return struct.pack("<H", x)
+
+def mk_uint32(x):
+    return struct.pack("<I", x)
+
+def compute_crc(data):
+  return 0xFFFFFFFF & -zlib.crc32(data) -1
+
+def read_file(file):
+    with open(file) as f:
+        return f.read()
+
+def write_file(file, content):
+    with open(file, "w") as f:
+        f.write(content)
+    
+def try_delete_file(file):
+    try:
+        os.remove(file)
+    except OSError, e:
+        if e.errno != 2:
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Thanks.

Due to the low-level nature of these changes, I think it will be important to break them up into individual commits - each doing one small and complete change.  This is important, because if we run into an issue we'll need to use tools like `git bisect` and `git revert`.

Also, as a general policy, we don't include mcu specific debugging or development tools in the Klipper github repo - as there are just too many of them.  So, scripts/elf2dfuse.py doesn't look like a good match for the Klipper master repo.

-Kevin

boilerbots:
I can divide things up if the changes seem reasonable.

The elf2dfuse.py is a utility required to convert compiled elf code into the format that the STM32 built in factory boot loader understands. IE. short boot0 pin to gnd and flash what ever you want. I am not running 3rd party boot loaders because my hardware didn't have functional ones like the monoprice mini delta or my own personal STM32F207 board. When I run a bootloader, I follow the STM design and also require the DFUSE format.

I admit that I don't know how your script can load binary code using dfu-util because I don't have an STM32F4xx board yet but I  just got a Bigtreetech Octopus and might look into it.

KevinOConnor:
> I can divide things up if the changes seem reasonable.

Well, I don't really understand all the changes here.  Splitting them up and providing a description of each part would also help with review.

> I admit that I don't know how your script can load binary code using dfu-util 

All of the stm32 chips I'm aware of do not actually require dfu format - the dfu-util tool will accept a raw .bin file.

-Kevin

boilerbots:
The changes have been split up and new requests generated.

PR #4454, opened 2021-07-03
heaters: Allow PID parameters to be adjusted at runtime

awigen:
This commit allows for the PID parameters to be adjusted at runtime.
This can be useful when different PID tunings are needed for different
temperatures. For example a heated bed running at 60c vs 100c.

This commit also allows for settle delta to be adjusted which is very
useful during print start macros. And allows settle delta to be loaded
from configuration files.

Signed-off-by: Alexander Wigen <alex@wigen.net>

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -148,6 +148,8 @@ The following standard commands are supported:
   is specified then the toolhead move will be performed with the given
   speed (in mm/s); otherwise the toolhead move will use the restored
   g-code speed.
+- PID HEATER=<config_name> [P=<proportional>] [I=<integral>] [D=<derivative>]
+  [SETTLE_DELTA=<temperature>] : Adjust PID parameters.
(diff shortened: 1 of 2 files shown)


StrikeEagleCC:
I would be grateful for this.

KevinOConnor:
Thanks, and sorry for the delay in responding.  Can you give a high-level overview on how the `PID` command is intended to be used?  Could similar functionality be obtained with alternate mechanisms?

The change to `pid_settle_delta` would need to be in a separate commit from a change that adds a new pid command.  However, as high-level feedback, I'm not really looking to add that type of functionality to the main Klipper branch.  The new config/command would solely alter the behavior of `M109`/`M190` commands and there isn't really any interest in adding "new capabilities" to "old g-code".  New functionality should go into new commands like `TEMPERATURE_WAIT` (eg, #4237).

-Kevin

PR #4456, opened 2021-07-04
Specify ATSAMD sercom explicitly in a samd_sercom config option

KevinOConnor:
The current setup doesn't make it possible to configure multiple samd sercom instances on multiple micro-controllers.  This updates the code to make it possible, but it does make a non-backwards compatible change (for those users using a samd_sercom config).

@FHeilmann - FYI.

-Kevin


FHeilmann:
Thanks, LGTM.

PR #4458, opened 2021-07-04
Update generic-bigtreetech-skr-2.cfg

lfarrand:
Updated SKR2 configuration following my install. Some of the pins have been changed. The original SKR 2 config didn't work -  the axes were inverted.

### DIFF
--- a/config/generic-bigtreetech-skr-2.cfg
+++ b/config/generic-bigtreetech-skr-2.cfg
@@ -12,11 +12,13 @@
 # Note: The initial revision of this board has a flaw that can cause
 # damage to itself and other boards. Be sure to verify the board is
 # not impacted by this flaw before using it.
+# See the following link for more information:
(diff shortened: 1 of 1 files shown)


Sineos:
According to 
https://raw.githubusercontent.com/bigtreetech/SKR-2/master/Hardware/BIGTREETECH%20SKR%202-Pin.pdf
https://raw.githubusercontent.com/bigtreetech/SKR-2/master/Hardware/BIGTREETECH%20SKR%202-SCH.pdf
the changes do not look correct. 
Maybe you switched your axes in your printer install.

Also the BLTouch information mostly make no sense also, since they are printer dependent and should not go into a generic board definition.

dawidmpunkt:
@lfarrand as Sineos mentioned, you swapped your x and y axes in your printer install. Pins for X and Y in the default config work fine on my SKR 2.  dir_pin on stepper_z needs to be inverted, though (!PA8 is correct).

dawidmpunkt (commented):
@lfarrand as Sineos mentioned, you swapped your x and y axes in your printer install. I advise against those changes. Pins for X and Y in the default config work fine on my SKR 2. dir_pin on stepper_z needs to be inverted, though (!PA8 is correct).

PR #4460, opened 2021-07-04
rp2040 bootloader request, make flash, and chipid support

dalegaard:
This adds a bootloader request (using the standard 1200 baud method), then builds `make flash` support on top of this.

To better support multiple chips, chipid support is also added. The rp2040 chips are all identical, and have no chip ID, but their flash chips do. At startup we decouple the XIP layer, ask the flash chip for UID, then re-engage the XIP layer.

Finally a small linker fix is included, as newer binutils(>=2.36) can't link without this change, see the corresponding commit.

/Lasse

### DIFF
--- a/scripts/flash-uf2.sh
+++ b/scripts/flash-uf2.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# This script installs a uf2 imag to a given rp2040 BOOTSEL disk
+
+DISK=$1
+BIN_FILE=$2
+
+if [[ "$DISK" = "" || "$BIN_FILE" = "" ]]; then
+    echo "Usage: $0 <disk> <uf2 file>" >&2
+    exit -1
+fi
+
+MOUNT_POINT="$(mktemp -d)"
+function cleanup {
+    umount "$MOUNT_POINT" &>/dev/null
+    rm -rf "$MOUNT_POINT"
+}
+trap cleanup EXIT
+
+mount "/dev/${DISK}1" "$MOUNT_POINT"
+cp "$BIN_FILE" "$MOUNT_POINT/"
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks!  As discussed, I merged in the build fix, usb boot, and chipid commits.  The flash_usb.py change looks good, but I'd like additional time to review it.

FYI, I also added in an additional change to use a `_ramfunc` macro - mostly so that each function gets its own linker section - which helps with the `--gc-sections` ld feature that Klipper uses.

-Kevin

dalegaard:
I've rebased and force-pushed just the `make flash` commit for easier review.

/Lasse

KevinOConnor:

It would certainly be nice to have `make flash` support for the rp2040.  However, I'm a bit concerned about running `sudo mount ...` on a board because of the havoc that could cause if the mount point isn't accurate for the particular OS the user is running.  It's unlikely to be an issue on an rpi with debian, but users run Klipper on many different machines.  That said, I don't have a good alternative.  So, I'm not sure what the best way to proceed is.

-Kevin

dalegaard:
Hi Kevin,

Yes I've been a bit concerned about it myself, and tried to avoid any issues in a few ways:
- The mount point is a temporary directory, created just for this mount
- The block device is verified to be an rp2040 in bootsel mode by looking at `/sys/`
- The disk name comes from `/sys/block/`

My hope was that this would make it pretty foolproof. Also `flash-uf2.sh` is only supposed to be called by `flash_usb.py`, not directly by a user(to avoid havoc, as you said). I'm not sure how to best indicate this? I put the actual mounting inside `flash-uf2.sh` to avoid having to run more stuff as root than absolutely necessary, as the probing step can be done without root.

dalegaard:
This works for my use case but I understand it may not for Klipper master so I'll close this. Feel free to re-open or use the code.

PR #4461, opened 2021-07-04
Allow STM32F401 boards to use STM32 HID Bootloader

henkyprayoga:
W/o this patch, STM32F401 can't select bootloader offset. This is necessary for scripts/usb-flash.py use STM32 HID bootloader instead of DFU.

I use STM32F401CC board from WeAct Studio. It comes with preloaded HID bootloader. However, the preloaded bootloader is not compatible with HID Bootloader from Serasidis. If anyone else turns out at the same situation, I made some patch on WeAct bootloader so it compatible with klipper flashing mechanism. 

https://github.com/henkyprayoga/WeAct_HID_Bootloader_F4x1/releases/tag/v1.1-stm32-hid-bootloader

Just download the zip and flash it using DFU.

### DIFF
--- a/docs/Benchmarks.md
+++ b/docs/Benchmarks.md
@@ -360,6 +360,32 @@ micro-controller.
 | 1 stepper (no delay) | 42    |
 | 3 stepper (no delay) | 194   |
 
+### RP2040 step rate benchmark ###
+
+The following configuration sequence is used on the RP2040:
+
+```
+allocate_oids count=4
+config_stepper oid=0 step_pin=gpio25 dir_pin=gpio3 invert_step=0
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  Change looks fine, but it needs a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md 

-Kevin

henkyprayoga:
Done. Thank you.

KevinOConnor:
Thanks.

-Kevin

xromansx:
The device is not always detected as HID need multiple time to replug it. changed different cables same issue

xromansx:
compiled the code and bootloader doesn't work led doesn't even light up.

PR #4463, opened 2021-07-05
rp2040: hardware PWM support

dalegaard:
This implements hardware PWM support for the rp2040. The maximum value(100% duty) is set to 255 to match the other controllers. Cycle time is clamped automatically, and uses the full 8.4 fractional range of the rp2040 PWM block. This allows a maximum PWM frequency of 490kHz and a minimum frequency of 1915 Hz.

Because the rp2040 timer runs at only 12MHz and the PWM system runs at full 125 MHz, the `cycle_time` received from Klippy needs to be scaled up, not down as is the case with e.g. the LPC176x and STM32 targets. To maximize the precision of this step, a little bit of extra math is done. This is documented in the code.

/Lasse

### DIFF
--- a/src/rp2040/hard_pwm.c
+++ b/src/rp2040/hard_pwm.c
@@ -1,0 +1,101 @@
+// Hardware PWM support on rp2040
+//
+// Copyright (C) 2021  Lasse Dalegaard <dalegaard@gmail.com>
+//
+// This file may be distributed under the terms of the GNU GPLv3 license.
+
+#include "autoconf.h" // CONFIG_CLOCK_FREQ
+#include "command.h" // DECL_CONSTANT
+#include "gpio.h" // gpio_pwm_write
+#include "sched.h" // sched_shutdown
+#include "internal.h" // get_pclock_frequency
+#include "hardware/structs/pwm.h" // pwm_hw
+#include "hardware/structs/iobank0.h" // iobank0_hw
+#include "hardware/regs/resets.h" // RESETS_RESET_PWM_BITS
+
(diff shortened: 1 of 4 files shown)


src/rp2040/hard_pwm.c
```diff
@@ -0,0 +1,97 @@
+// Hardware PWM support on lpc176x
```
KevinOConnor, line 1: Copy-and-paste error.

src/rp2040/main.c
```diff
@@ -144,7 +144,8 @@ clock_setup(void)
     watchdog_hw->tick = 1 | WATCHDOG_TICK_ENABLE_BITS;
 
     // Enable GPIO control
-    enable_pclock(RESETS_RESET_IO_BANK0_BITS | RESETS_RESET_PADS_BANK0_BITS);
+    enable_pclock(RESETS_RESET_IO_BANK0_BITS | RESETS_RESET_PADS_BANK0_BITS
+                  | RESETS_RESET_PWM_BITS);
```
KevinOConnor, line 148: I think it would be preferable to enable the PWM clock in hard_pwm.c (as this follows the common behaviour on rp2040 and other architectures).

KevinOConnor (commented):
Thanks.  In general, it looks fine, but I have a few minor comments (see below).

-Kevin

dalegaard:
Hi Kevin,

Thanks for the comments, the issues should be fixed now :ok_hand: 

/Lasse

KevinOConnor:
Thanks.

-Kevin

PR #4471, opened 2021-07-09
spi_flash: Support firmware dictionary validation

jschuh:
Updates firmware validation to use a dictionary if provided (or found
when updating from the default out/ directory).

Validation without a dictionary still checks the following (in order):
1. Active firmware's raw dictionary changed after update
2. Checksum of firmware.cur matches expected

This PR depends on #4381. Hopefully I did this right, since it's the first time I've stacked PRs. I would have expected the diff to be against the other PR, but the UI seems to display a much larger diff against master.

### DIFF
--- a/klippy/msgproto.py
+++ b/klippy/msgproto.py
@@ -408,6 +408,8 @@ class MessageParser:
         except Exception as e:
             logging.exception("process_identify error")
             self._error("Error during identify: %s", str(e))
+    def get_raw_data_dictionary(self):
+        return self.raw_identify_data
     def get_version_info(self):
         return self.version, self.build_versions
     def get_messages(self):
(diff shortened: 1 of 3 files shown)


jschuh:
@Arksine, I wanted to check if you're okay with the way I split this out from #4381. Kevin suggested that the script should always validate the dict if one is passed, but I know you're concerned about user error as well. So, I kept the dictionary as an optional switch and have the shell script provide a default `out/firmware.dict` only if it's also using the default `out/firmware.bin`. That seemed like a good way to address both concerns, but I'm open to other suggestions.

Arksine:
Looks ok to me, it seems that the user will have to explicitly enter the wrong dictionary path for it to be a problem.

jschuh:
Thanks, that was my take as well. If you're okay with it, then I think this patch and #4381 should be good to go?

KevinOConnor:
Thanks.

-Kevin

PR #4477, opened 2021-07-10
toolhead: Add `RESPOND` attribute to `SET_VELOCITY_LIMIT`

meteyou:
This attribute can be used to hide the debug output of SET_VELOCITY_LIMIT, e.g. when overwrite the M204 gcode or with post processing of the ACC values. 
You could also hide / filter the outputs in the GUI, but then the moonraker gcode/respond cache (https://moonraker.readthedocs.io/en/latest/web_api/#request-cached-gcode-responses) is still overfilled with these outputs and important outputs could be overlooked / kicked out.
The normal behavior of SET_VELOCITY_LIMIT is not affected by this PR.

M204 gcode_macro example:
```
[gcode_macro M204]
rename_existing: M204.1
gcode:
    {% set ACC = params.S|default(1000)|float %}
    {% set ACC2DECEL = params.S|float/2 %}
    SET_VELOCITY_LIMIT ACCEL={ACC} ACCEL_TO_DECEL={ACC2DECEL} RESPOND=0
```
(the macro originally comes from here: https://github.com/KevinOConnor/klipper/pull/4342#issuecomment-855192746)


KevinOConnor:
Thanks.  However, my high-level feedback is that I'm not sure of this change.  My main concern is that the new parameter is a bit "goofy" - no other command has a similar option to control user feedback.  We could remove all user feedback from the command, but that could be considered a regression for some use cases.

I understand that output is not desirable when the command is invoked frequently from a slicer.  However, from my perspective, changing core printer kinematic settings thousands of times in a print is also a "goofy" thing to do.  (It's not clear to me that frequently changing these parameters results in improved print quality.)  So, to me, it seems almost like we'd be adding a "goofy" command parameter to make a "goofy" configuration work slightly better..

-Kevin

jordan-woyak:
I agree that this is a out-of-place solution but frequently changing these parameters is not a strange thing to do.
e.g. It's sensible to use more aggressive acceleration for hidden infill.

As far as I can tell, `SET_VELOCITY_LIMIT` is the only way to adjust `ACCEL_TO_DECEL` and `SQUARE_CORNER_VELOCITY`.

Currently, if I override `M204` or `M205` to use `SET_VELOCITY_LIMIT` and control the acceleration/jerk in the slicer based on print feature there is so much spam in Mainsail's console output it starts to make my browser hang.

zellneralex:
Hello Kevin,

over the last 3 month, people have tried to change M204 and you told please use an rename_existing. That solution has the problem that you have an overflowing console. It is quit normal to change your accel limit during a print to allow different accel for different things like Infill or Permitter. 
As accel_to_decel must be changed also in this situation SET_VELOCITY_LIMIT is the only solution we have.
So than please tell us what you would accept as an solution.

meteyou:
thx for you feedback. and maybe it was a too quick and too dirty solution... i checked some other extended gcodes and found maybe a "better" solution.

i checked these gcodes:
- SET_VELOCITY_LIMIT
- SET_EXTRUDER_STEPS_DISTANCE
- SET_PRESSURE_ADVANCE
- SET_GCODE_OFFSET (no respond)
- SET_HEATER_TEMPERATURE (no respond)
- SET_STEPPER_ENABLE (no respond)
- SET_DISPLAY_GROUP (no respond)
- SET_IDLE_TIMEOUT
(there are much more SET_ commands in klipper)

i have checked these for a respond and have especially checked the behavior when no new value is passed.
I could think of 2 possible solutions. the first and probably simpler solution would be to return a respond only if no new value is set.

for example:
```
SET_VELOCITY_LIMIT > respond the current values
SET_VELOCITY_LIMIT ACCEL=2000 > no respond
```

or the second solution would be to remove all respond from "SET_" commands and add a second command with "GET_" at the beginning.

for example:

```
SET_VELOCITY_LIMIT ACCEL=2000 > no respond
GET_VELOCITY_LIMIT > respond of all current values
```

would you prefer one of these solutions? if so, i could modify all the gcodes...

meteyou:
small addition...

there are already some extended gcodes with SET_ & GET_

for example:
- SET_SKEW & GET_CURRENT_SKEW
- SET_RETRACTION & GET_RETRACTION

KevinOConnor:
FWIW, I'd recommend against adding `GET_XXX` commands.  It's preferable to access Klipper state via the "command templates" system (eg, `{printer.toolhead.max_velocity}`) and not via formatted g-code responses.

I'd be okay with changing SET_VELOCITY_LIMIT to only respond when there are no parameters.  Such a change would also need an additional entry in docs/Config_Changes.md .

-Kevin

strayr:
If changes are being ruled out because they don't stick to The Style, I think a very clear decision on what The Style should be is required. 

This solution is stateless and makes no changes to behaviour. GET/SET would increases the exposed functions, and a quiet SET unless no (useful) parameters changes and both make changes to info provided to user.

A much bigger change would be a config (or realtime) variable that tracks a user's chattiness preferences and could be consulted by any output, but I think that will confuse a lot of people and is more work.

Sorry if it feels like i'm jumping in, i'm using this PR because it's helpful and an idea of _the right way to do things_ will help me with what I'm trying to get done.

meteyou:
i have now modified SET_VELOCITY_LIMIT so that it only create a respond when a value is set. 

i personally think it's a very nice solution and doesn't require another command. should i modify all the other SET_ commands to be similar?

docs/G-Codes.md
```diff
@@ -160,8 +160,9 @@ The following standard commands are supported:
   Wait until the given temperature sensor is at or above the supplied
   MINIMUM and/or at or below the supplied MAXIMUM.
 - `SET_VELOCITY_LIMIT [VELOCITY=<value>] [ACCEL=<value>]
-  [ACCEL_TO_DECEL=<value>] [SQUARE_CORNER_VELOCITY=<value>]`: Modify
-  the printer's velocity limits.
+  [ACCEL_TO_DECEL=<value>] [SQUARE_CORNER_VELOCITY=<value>]
+  [RESPOND=0]`: Modify the printer's velocity limits. If RESPOND=0 is
```
blalor, line 164: This new param is no longer supported.

blalor (changes requested):

docs/G-Codes.md
meteyou: Oh... Yes! I only changed the command for testing.

meteyou:
here a gcode example (sliced with PS 2.3.1-rc), why we should remove the respond:

```
;TYPE:Gap fill
;WIDTH:0.20993
G1 F1800.000
G1 X172.958 Y169.174 E0.00624
M204 S4000
M204 S2000
;WIDTH:0.209929
G1 X172.843 Y169.251 E0.00225
M204 S4000
M204 S2000
;WIDTH:0.197997
G1 X172.408 Y169.626 E0.00861
M204 S4000
M204 S2000
;WIDTH:0.211698
G1 X172.320 Y169.712 E0.00203
M204 S4000
M204 S2000
;WIDTH:0.198145
G1 X171.967 Y170.151 E0.00847
M204 S4000
M204 S2000
;WIDTH:0.201827
G1 X171.961 Y170.160 E0.00016
M204 S4000
M204 S2000
;WIDTH:0.204171
G1 X171.949 Y170.177 E0.00032
M204 S4000
M204 S2000
;WIDTH:0.20417
G1 X171.708 Y170.595 E0.00758
M204 S4000
M204 S2000
;WIDTH:0.203778
G1 X171.692 Y170.627 E0.00056
M204 S4000
M204 S2000
;WIDTH:0.199543
G1 X171.688 Y170.636 E0.00015
M204 S4000
M204 S2000
;WIDTH:0.210932
G1 X171.542 Y170.993 E0.00634
M204 S4000
M204 S2000
;WIDTH:0.210932
G1 X171.491 Y171.138 E0.00253
M204 S4000
M204 S2000
;WIDTH:0.198743
G1 X171.488 Y171.148 E0.00015
M204 S4000
M204 S2000
;WIDTH:0.211576
G1 X171.441 Y171.325 E0.00303
M204 S4000
M204 S2000
;WIDTH:0.211578
G1 X171.370 Y171.674 E0.00588
M204 S4000
M204 S2000
;WIDTH:0.195761
G1 X171.368 Y171.683 E0.00014
M204 S4000
M204 S2000
;WIDTH:0.20792
G1 X171.353 Y171.799 E0.00188
M204 S4000
M204 S2000
;WIDTH:0.20792
G1 X171.328 Y172.179 E0.00614
M204 S4000
M204 S2000
;WIDTH:0.20281
G1 X171.327 Y172.185 E0.00008
M204 S4000
M204 S2000
;WIDTH:0.207527
G1 X171.325 Y172.262 E0.00124
M204 S4000
M204 S2000
;WIDTH:0.207525
G1 X171.325 Y177.232 E0.07986
M204 S4000
M204 S2000
;WIDTH:0.207527
G1 X171.329 Y177.331 E0.00159
M204 S4000
M204 S2000
;WIDTH:0.200113
G1 X171.329 Y177.338 E0.00010
M204 S4000
M204 S2000
;WIDTH:0.212083
G1 X171.372 Y177.823 E0.00807
M204 S4000
M204 S2000
;WIDTH:0.212083
G1 X171.400 Y177.997 E0.00293
M204 S4000
M204 S2000
```

i think this is also a PS bug, but with a gcode like this, you have a really big klippy.log and only not important lines in the console.

KevinOConnor:
Thanks.  However, the `set_rollover_info()` function needs to run if any value is set.  Getting the latest requested kinematics information in the log is important, should an error occur and we need to analyze the log.  Only the `respond_info()` call should be conditional.

-Kevin

meteyou:
I just used the `M204` command as a guide and it doesn't have an output either. should I add the ACCEL in the log in the `M204` as well?

zellneralex:
Its running on my machine at every print. I do setup current, PA, accel/accel_to_decel at every print

meteyou:
@KevinOConnor i have modified all the commands i found with the same behaviour. is there anything else todo?

KevinOConnor:
Thanks.  The change to toolhead.py looks fine to me.  However, the changes to the other commands would all need to be in separate commits (so that we can use `git bisect` and `git revert` should a problem occur).  I'd also like to clear those changes with each of the authors of the given modules.

FWIW, it's probably simpler to limit this PR to the change to toolhead.py as that will be simpler to review and merge.

-Kevin

meteyou:
i splitted the first 3 moduls and will finish it during this week.

PR #4480, opened 2021-07-11
docs: Fix M300 beeper example.

s-holst:
The M300 example macro now sets the PWM cycle time to generate the requested
frequency. 'scale' did not have the desired effect. Duty cycle is now
fixed at 0.5. All S parameter values less than 1 generate a 1Hz 'tone'
to avoid division-by-zero.

Signed-off-by: Stefan Holst <mail@s-holst.de>


KevinOConnor:
Thanks.  Looks fine to me.  @ndupont-net do you have any comments on this?

-Kevin

ndupont-net:
All good for me.
Nicolas

PR #4482, opened 2021-07-12
scripts: Convert compiled elf to dfuse format.

boilerbots:
For STM32 controllers you can use the factory ROM bootloader and
dfu-util to program them if the elf file is put into the correct format.

The advantage is that the code loads into the correct flash memory location.
The symbol "VectorTable" is used to set the image address in the dfuse file.

### DIFF
--- a/scripts/elf2dfuse.py
+++ b/scripts/elf2dfuse.py
@@ -7,47 +7,62 @@ from __future__ import with_statement
 #
 # usage: elf2dfuse.py infile outfile
 
-import os, subprocess, tempfile
+import os
+import subprocess
+import tempfile
 import sys
 import struct
 import subprocess
 import zlib
 
+
 def mk_uint8(x):
     return struct.pack("<B", x)
+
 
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks, but I'm not sure this change is a good candidate for the Klipper master branch.  The dfu-util program, as far as I know, accepts raw binaries.  So, I'm unsure of the advantage of adding a tool to convert to dfu format.  Also, FYI, the existing lib/rp2040/elf2uf2 tool may also be capable of creating dfu binaries (I'm not sure).

-Kevin

boilerbots:
I am pretty sure you are relying on the printer boards bootloader to accept your binary file. If you boot any of the STM32 processors up in DFU mode using the boot 0 pin they will only accept uploaded in the proper dfu format as I mentioned in the comment of my script. I run an F0, F2 and F4 without aftermarket bootloaders simply using the boot 0 pin and this conversion script.

PR #4483, opened 2021-07-12
stm32: For stm32f0 set OSPEEDR to high-speed

boilerbots:
The hex value 2 matches the x0 rule which
is low speed, use 0x3 instead for high speed.

This helps with long traces or fanout.


KevinOConnor:
Thanks.  However, I'm not sure of this change.  This would cause the f0 code to deviate from the other stm32 chips.  It's not clear to me why one would want to alter the "gpio speed" or why one would need more than a 10mhz update rate.

-Kevin

boilerbots:
That’s fine to not take it but this register sets the pins drive strength which affects the signals rise and fall times.

boilerbots:
The code was setting the speed to an undefined state, however I wanted the strongest drive with the highest speed because some pins where being fanned out to drive multiple drivers and I am not concerned with electrical emissions. Please close if you don't want this.

KevinOConnor:
Okay, thanks.  I missed that 0x2 on stm32f0 is the same as 0x0, which definitely seems like a poor choice.  I updated the code with some additional comments (1b801508) and changed the stm32f0 code to match the speeds set for the stm32f1 code (97d36ac4).

-Kevin

PR #4484, opened 2021-07-12
stm32: Stop the watchdog clock when the core is halted.

boilerbots:
Since the watchdog is always enabled it is impossible to debug using
stlink or jtag. Set the DBG_IWDG_STOP bit so the watchdog clock stops
when the core is halted.

An alternative would be to provide a compile time option to disable
the watchdog.


boilerbots:
The debugger can't do this because break points are implemented in the hardware and the watchdog have safety mechanisms to prevent it from being easily disarmed. In my projects I would have a debug build and a release build and enable things like this for the debug build along with assertion checks and other nice things that might otherwise slow the code down.

Anyway I guess almost no one really uses JTAG or STLink debugging. Close it if you don't want it.

PR #4485, opened 2021-07-12
stm32: Fix USB startup problems

boilerbots:
My stm32f207 board would not startup, no USB connection. Debugging over
STLink with gdb shows problems with clock configuration however using the
debugger was changing the behaviour of the system.

This change move clock and power domain configuration into the specific
processor sections to isolate the changes. Then I enable the USB power
and clock domain and also add a wait for the HSERDY flag.

I removed writing to SCB->VTOR because it is done by SystemInit().

I skip the sysbase write also because it prevents proper startup on my boards.

### DIFF
--- a/src/stm32/stm32f4.c
+++ b/src/stm32/stm32f4.c
@@ -124,20 +124,42 @@ enable_clock_stm32f20x(void)
 {
 #if CONFIG_MACH_STM32F207
     uint32_t pll_base = 1000000, pll_freq = CONFIG_CLOCK_FREQ * 2, pllcfgr;
+    __IO uint32_t StartUpCounter = 0, HSEStatus = 0;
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Okay, thanks.  However, there are too many changes in this commit for me to feel comfortable with it.  I also don't understand the core issue that is being fixed here.  I think it would help if you could narrow down the change to just what fixes the issue and not introduce any other changes (or place the other changes in separate commits).  This is important so that we can use `git bisect` and `git revert` if necessary.

Separately, the change to VTOR would introduce a regression - the SystemInit() code sets the VTOR to an incorrect location when a bootloader is used, which is why the Klipper code sets it back to a correct location.

-Kevin

PR #4486, opened 2021-07-12
added JSON based temperature sensor to heaters.py

br-edward:
Adds JSON based local file and remote (urllib) sensors to heaters.py

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2252,6 +2252,20 @@ sensor_type: temperature_host
 #   system file on a Raspberry Pi computer.
 ```
 
+## Any temperature sensor
+
+Temperature via URL-JSON source or any file (improved version).
+
+```
+sensor_type: sensor_any
+#sensor_path:
+#   URL or file path to the file that contains the temperature
+#conversion_factor:
+#   if specified, factor to multiply (temperature) with
+#json_path:
+#   if specificed, with e.g. ["Status"]["TEMPERATURE"]["Value"]
+#   sets the correct position in the JSON data for your temperature
+
 ## DS18B20 temperature sensor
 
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  However, this code looks like it could cause internal errors - if the socket reading code were to block, it would cause internal errors in the Klipper host code.  That is, the "event reactor" that Klipper uses can't tolerate arbitrary delays.  In general, we want the host code to aggregate information from the micro-controller code - and do all the timing intensive work in the micro-controller code.

-Kevin

br-edward:
> Thanks. However, this code looks like it could cause internal errors - if the socket reading code were to block, it would cause internal errors in the Klipper host code. That is, the "event reactor" that Klipper uses can't tolerate arbitrary delays. In general, we want the host code to aggregate information from the micro-controller code - and do all the timing intensive work in the micro-controller code.
> 
> -Kevin

Hello, Kevin.

Thanks for your answer, my initial thought was that reactor timers where async non-blocking queue objects. Nevertheless what is the max. execution time (delay) in millis you believe a sensor module is allowed to run for their respective sampling routines? I just now wrote a version where this is limited to 70ms, by timeout handling. Do you suggest a background threading model for the sampling routines instead? That would be easy to implement, as well.

Thanks!

KevinOConnor:
The main host event code shouldn't block for more than a millisecond I'd say.

-Kevin

br-edward:
> The main host event code shouldn't block for more than a millisecond I'd say.
> 
> -Kevin

Which function of the class  `Sensor_ANY` do you consider part of the "main host event code"? What about the reactor timer as queue objects?

KevinOConnor:
All the Python code runs in the Klipper "reactor".  We don't do any blocking IO calls anywhere in the code.

-Kevin

br-edward:
>  We don't do any blocking IO calls anywhere in the code.

Are your sure? I thought on Python2 file IOs are usually blocking calls. Here an example in the existing code: https://github.com/KevinOConnor/klipper/blob/master/klippy/extras/temperature_host.py#L55

KevinOConnor:
The temperature_host.py code is designed to read from a `/sys/` file with the understanding that the Linux kernel would never block on a simple read.

-Kevin

PR #4488, opened 2021-07-14
Update printer-flsun-q5-2020.cfg

ddacunha:
module: Config file

The MKS Robin Nano v1.2 expects the firmware file to be called `Robin_nano.bin`.

Signed-off-by: Daniel Da Cunha <github@ddc.im>

### DIFF
--- a/config/printer-flsun-q5-2020.cfg
+++ b/config/printer-flsun-q5-2020.cfg
@@ -8,7 +8,7 @@
 
 # Note that the "make flash" command does not work with MKS Robin
 # boards. After running "make", run the following command:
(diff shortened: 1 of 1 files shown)


KevinOConnor:
This was just changed to nano by @CobraPi in PR #4237 .  @CobraPi - can you confirm if that change was intentional or not?

-Kevin

CobraPi:
That change was unintentional. I only added the qqs config, but I copied the format of the q5 config. I must have edited the wrong file by accident.

KevinOConnor:
Thanks.

-Kevin

PR #4464, opened 2021-07-06
kinematics: add duplication/mirror modes to idex_mode

Tircown:
Add duplication and mirror modes to idex_modes.
Allows hybrid-corexy and hybrid-corexz to use these modes.

Idex_mode avoids collisions if the user correctly sets the position_min and position_max for stepper_x and dual_carriage.

Do not merge before PR ~~#4445~~ #4489. I will add a sample configuration and a document on how to set the IDEX parameters.

Signed-off-by: Fabrice GALLET tircown@gmail.com

PR #4445, opened 2021-06-30
sync_heater and sync_extruder

Tircown:
Add the ability to synchronize heaters temperatures with an optional offset.
Add the ability to synchronize extruders steppers and heaters with an optional offset.

This PR prepares duplication/mirror modes for IDEX printers. Even if this code is fully functional the PR is mainly for discussion about the best way to synchronize the extruders. Heaters synchronisation can eventually be used for big heatbed that use multiple heatpads.

Signed-off-by: Fabrice GALLET tircown@gmail.com

PR #4489, opened 2021-07-14
Add sync_extruder_steppers

Tircown:
Add the possibility to synchronize multiple extruder steppers.
Typical usage is to mimic extruder with extruder1 on an IDEX printer printing in duplication or mirrored mode.
Related to PR #4464, replace #4445. 

Signed-off-by: Fabrice GALLET tircown@gmail.com

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -170,6 +170,9 @@ The following standard commands are supported:
 - `ACTIVATE_EXTRUDER EXTRUDER=<config_name>`: In a printer with
   multiple extruders this command is used to change the active
   extruder.
+- `SYNC_EXTRUDER_STEPPERS EXTRUDER=<config_name> [TO=<config_name>]`:
+  Synchronize extruders steppers. To unsynchronize an extruder, omits the TO
+  parameter.
 - `SET_PRESSURE_ADVANCE [EXTRUDER=<config_name>] [ADVANCE=<pressure_advance>]
   [SMOOTH_TIME=<pressure_advance_smooth_time>]`: Set pressure advance
   parameters. If EXTRUDER is not specified, it defaults to the active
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  As high-level feedback, though, I think this change should wait until we've reworked the code so that the existing `SYNC_STEPPER_TO_EXTRUDER` command can be used with extruder stepper motors.  That is, I'd prefer to not add a new command that does something very similar to an existing command.

Also, the help text on the new command mentions that the command syncs "heaters and stepper", but that doesn't seem to be the case?

-Kevin

Tircown:
Thank you for the comment.

The mention of heating is an oversight from the original version (PR 4445).
 
I do understand that it's not right to add and then remove a G-code command. Duplication and mirroring modes require the extruder and extruder1 to make the same movements. Unless to declare dummy pins in extruder1 and an extruder_stepper section with the real pins and use the `SYNC_STEPPER_TO_EXTUDER` command, I don't know how to perform this task from g-code commands without wasting pins. I assume that dummy pins must exist for the given mcu.

Have you defined a schedule for the extruder.py rework? Can I help on some points?

KevinOConnor:
Just to be clear, I'm not suggesting that one define an `[extruder_stepper]` config section to synchronize multiple extruders - I'm saying I think the existing `SYNC_STEPPER_TO_EXTUDER` command should work for either extruder or extruder_stepper.

> Have you defined a schedule for the extruder.py rework? Can I help on some points?

I haven't looked at it in sufficient detail.  I'll try to take a closer look at it.

-Kevin

Tircown:
Thank you for the comment.
Got it.

KevinOConnor:
I do think it would be useful to merge this support into Klipper.  If there is still interest in working on this then please update it to the latest code and I will prioritize a review of this work.

Thanks,
-Kevin

Tircown:
Thank you. The PR is ready to be reviewed and hopefully merged.

KevinOConnor:
Thanks.  At a high-level, I think it makes sense to add this support.  However, I think it would be preferable to refactor the existing SYNC_STEPPER_TO_EXTRUDER than to introduce a new SYNC_EXTRUDER_STEPPERS command.

I have attempted to do this in PR #5143 .  Unfortunately, I don't have a good way to test that PR.  Would you be able to review that PR and see if it meets the requirements here?

-Kevin

Tircown:
Thanks. I think I can close this PR since #5143 seems to work perfectly now for what we need in idex modes.

PR #4491, opened 2021-07-15
New printer profile for Eryone ER20 and minor update for Eryone Thinker Series

dapostol73:
As always, please let me know if you have any feedback on the changes.  Thanks,


KevinOConnor:
Thanks.  Config looks fine to me - I would ask that we try to place author information in the commit message and not in the config file though (mentioned briefly at https://www.klipper3d.org/Example_Configs.html ).

-Kevin

sihawken:
The Eryone ER-20 requires a 16KiB bootloader as well, which isn't currently an option in make menuconfig. 

https://github.com/KevinOConnor/klipper/pull/4476

KevinOConnor:
Thanks.

-Kevin

PR #4492, opened 2021-07-15
rp2040: implement SPI

dalegaard:
This implements SPI for the rp2040 target. All output groupings of both SPI blocks are available for use.

Because each blocks has multiple outputs, I called the first output of each block `spi0a` and `spi1a`. If you prefer another naming scheme, let me know!

/Lasse


KevinOConnor:
Thanks.

-Kevin

PR #4493, opened 2021-07-15
Add 16KiB bootloader option for STM32F103

sihawken:
The Eryone ER-20 is a STM32F103 based 3D printer which ships with a 16KiB bootloader. Their official version of Marlin starts at 0x08004000.

This can be seen here. https://github.com/Eryone/STM32/blob/master/Marlin_ER20/buildroot/share/PlatformIO/scripts/eryone_mini_STM32f103.py

Fixed signoff issue with previous pull request: https://github.com/KevinOConnor/klipper/pull/4476


KevinOConnor:
Thanks.

-Kevin

PR #4496, opened 2021-07-17
St7920 cmd and sync delay config option

sihawken:
Necessary to increase the reliability of the st7920 display with certain manufacturer's 3D printers. It is common in Marlin for the manufacturer to have added their own display delay in the form of: #define ST7920_DELAY_1... ST7920_DELAY_3. Adding this as a configurable option will theoretically help increase the reliability of certain displays across various manufacturers.

On the Eryone ER-20, there are some printers which work well with the default delay options, and others which don't. Marlin works well for these printers.

This is a user who used the default config on the ER-20, and was getting really odd LCD issues. We were unsuccessful in solving the issue by getting him to test multiple USB cables and cutting the signal from the 5V pin on the USB cable. None of it worked.
 
image

Default Marlin for the ER-20 has the following delays configured in the pins file

```
#ifndef ST7920_DELAY_1
  #define ST7920_DELAY_1 DELAY_NS(125)
#endif
#ifndef ST7920_DELAY_2
  #define ST7920_DELAY_2 DELAY_NS(125)
#endif
#ifndef ST7920_DELAY_3
  #define ST7920_DELAY_3 DELAY_NS(125)
#endif
```

We would like to get him to test different LCD delays in Klipper, but that is much easier for less technical users if there it is a configurable option. So we made this change to test to see if it makes a difference and at the very least to set the delays to the same value as stock Eryone ER-20 Marlin.

Signed-off-by: Simon Hawkenson <simon@hawkenson.ca>

### DIFF
--- a/klippy/extras/display/st7920.py
+++ b/klippy/extras/display/st7920.py
@@ -142,9 +142,8 @@ class ST7920(DisplayBase):
                 for name in ['cs', 'sclk', 'sid']]
         mcu = None
         # Spec says 72us, but faster is possible in practice
-        
-        delay_cmd = config.getfloat('st7920_delay_cmd', .000020, minval=.000020)
-        delay_sync = config.getfloat('st7920_delay_sync', .000045, minval=.000045)
+        delay_cmd = config.getfloat('delay_cmd', .000020, minval=.000020)
+        delay_sync = config.getfloat('delay_sync', .000045, minval=.000045)
         for pin_params in pins:
             if mcu is not None and pin_params['chip'] != mcu:
                 raise ppins.error("st7920 all pins must be on same mcu")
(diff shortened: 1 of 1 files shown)


sihawken:
image

Config options to set the different delays

KevinOConnor:
Thanks.  However, I don't think we want to add config options for these delays, as it isn't something a user can reasonably configure.  Set it too low and the display just wont work, set it too high and the code will throw random errors.  In practice, there are only a handful of chips out there - I think it would be preferable to find a set of delays that work well with most chips.

Separately, any time a config option is added, it is necessary to add the corresponding documentation to docs/Config_Reference.md.  (Though, as above, I don't think we should do that.)

-Kevin

PR #4502, opened 2021-07-20
pca9533: Enhanded RGB support for the mightyboard

DeanCording:
Added ability to set led brightness levels between 0 and 1

Signed-off-by:  Dean Cording <dean@cording.id.au>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2585,10 +2585,12 @@ PCA9533 LED support. The PCA9533 is used on the mightyboard.
 #initial_GREEN: 0
 #initial_BLUE: 0
 #initial_WHITE: 0
-#   The PCA9533 only supports 1 or 0. The default is 0. On the
-#   mightyboard, the white led is not populated.
-#   Use GCODE to modify led values after startup.
-#   set_led led=my_pca9533 red=1 green=1 blue=1
+#   The PCA9533 supports any value between 0 and 1. The default is 0 for all leds.
+#   The PCA9533 has one on/off channel and two PWM channels for controlling led brightness
+#   so some combinations of led settings are approximated.
+# . On the mightyboard, the white led is not populated.  Using it may result in odd colors.
+#   Use GCODE to modify led values after startup.  Leds may be controlled individually.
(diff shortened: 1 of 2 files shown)


KevinOConnor:
@le-Bark - do you have any comments on this?

-Kevin

le-Bark:
Good to see that someone is improving my contribution.

The reason that I didn't implement the dimming feature in the beginning is because the driver IC only has 2 timers.
I think that it is a good way to implement it.

Is [this ](https://github.com/KevinOConnor/klipper/pull/4150#issuecomment-816992032) comment still pertinent?

-Marc-André

DeanCording:
I've added the code to support the SYNC option referred to [above](https://github.com/KevinOConnor/klipper/pull/4502#issuecomment-883499235)

KevinOConnor:
Thanks.  I don't know enough about this chip to provide much review feedback.  That said, I'm fine with merging.  However, one thing I noticed is the copyright line - in general, it's preferable to not add a copyright with every change, but if you do feel a copyright is needed, please make it look like the other copyright lines:
```
# Copyright (C) 2021  Dean Cording <dean@cording.id.au>
```

Cheers,
-Kevin

DeanCording:
I've removed the copyright lines as there was little of @le-Bark 's original code left and my code was a re-implementation of the example use from the PCA9533 data sheet.

KevinOConnor:
Thanks.  However, I have the same comment as before - I'd prefer all the code to follow the same copyright header format - one line description, one blank line, one line copyright, one blank line, one line license statement.  I don't think we can remove the existing copyright from this file.  If you feel you need to add your copyright, then that is okay, but please add it using the same format as the other files so that is clear you are claiming a copyright.  (And in that case, it would be one line description, one blank line, two copyright lines, one blank line, one line license statement - for example see klippy/extras/display/display.py .)  If you don't feel you need to add your copyright to this file, then please don't modify the copyright block at the top of the file.

-Kevin

KevinOConnor:
I'm going to mark this as inactive as it looks like there hasn't been a response to the comments at https://github.com/Klipper3d/klipper/pull/4502#issuecomment-902384117 .

-Kevin

PR #4503, opened 2021-07-20
docs: Make sure SPI is enabeld

strayr:
Module: Documentation

Whilst this line is in Measuring_Resonances.md, it's causing some confusion for users and it should be in the RPi Microcontoller guide because not all RPi users are measuring resonances, and SPI appears to be a hard requirement for the RPi mcu

Signed-off-by: Rowland Straylight rowlandstraylight@gmail.com


KevinOConnor:
Thanks.

-Kevin

PR #4504, opened 2021-07-20
New motion_report module

KevinOConnor:
This adds a new module that adds diagnostic reporting on Klipper's motion system.  This initial PR adds support for dumping out the motion queues (the trapq history and the stepper queue_step history) on a shutdown event.

I plan to continue adding additional support to the module.  My goal is to make it possible to export the trapq and step queues in near real-time via the "api server" for analysis purposes.

Separately, I've also locally attempted to extend this new module to report the current "live" toolhead position (ie, the position of the toolhead at the current time - not the last requested gcode position).  Unfortunately, my current code produces misleading results after a `toolhead.set_position()` call.  If I get that working though, I'll push it in a separate PR.

-Kevin

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -189,6 +189,18 @@ The following information is available in
 - `last_stats.<statistics_name>`: Statistics information on the
   micro-controller connection.
 
+## motion_report
+
+The following information is available in the `motion_report` object
+(this object is automatically available if any stepper config section
+is defined):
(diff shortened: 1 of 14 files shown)


KevinOConnor:
FYI, I've updated this PR to include a `get_status()` method that reports the current requested toolhead position.  This is the "live" position of the toolhead - where the toolhead is actually at - which should be nicer for displays.

-Kevin

PR #4505, opened 2021-07-20
Trinamic stepper motor UART updates

KevinOConnor:
An investigation by @ReXT3d found that the tmc2209 UART behavior is greatly improved by setting SENDDELAY=2.  ( https://klipper.discourse.group/t/intermittent-tmc-uart-weirdness/454 ) As a guess, the additional delay helps the chips identify each others responses on shared UART lines - which may improve their ability to keep synchronized with the UART traffic.

This PR also increases the default UART baud rate to 40000 for 32bit MCUs.  There shouldn't be any harm in using a faster speed on more modern MCUs (the slower AVR chips will continue to use a baud of 9000).

-Kevin

### DIFF
--- a/klippy/extras/tmc2209.py
+++ b/klippy/extras/tmc2209.py
@@ -59,6 +59,9 @@ class TMC2209:
         self.fields = tmc.FieldHelper(Fields, tmc2208.SignedFields,
                                       FieldFormatters)
         self.mcu_tmc = tmc_uart.MCU_TMC_uart(config, Registers, self.fields, 3)
+        # Setup fields for UART
+        self.fields.set_field("pdn_disable", True)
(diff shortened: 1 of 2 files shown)


ReXT3D:
Wonderful!  Thanks very much for the suggestions with the troubleshooting and the investigation!

Peter.

KevinOConnor:
> would the comment # Avoid tx errors on shared uart be better to reference rx errors since the actual manifestation of the issue seemed to be rx retries?

The diagnostic information from the logs seemed to indicate the root issue was the drivers not accepting the query request (and thus not responding at all).  So, it seemed to be a host->driver tx problem.

-Kevin

ReXT3D:
Ah, makes perfect sense now, thank you.

I'm very happy to help whenever I can, especially with awesome projects like Klipper!

Thank you.

PR #4508, opened 2021-07-21
docs: dual_carriage

charlespick:
Added an STL for calibration, some gcode macros to the sample idex config, and an entire document on configuring an idex printer.

Signed-off-by: Charles Pickering <charles.pickering@live.com>


charlespick:
If anyone knows how dual carriage works with bed level, please let me know. That's the other big question I wanted to answer but couldn't because I don't have one on my idex printer.

KevinOConnor:
Thanks.  As high-level feedback, it looks good to me.  I think this change should wait until after #4512 though.  As for idex work, I don't have that hardware, but @Tircown has done a lot of work in that area recently.

-Kevin

Tircown:
I didn't test a bed_mesh or z_tilt myself on a IDEX but for what I know it works. Ankurv, tester of hybrid-corexz, uses the bed_mesh.

I am a little embarrassed because #4464 will overwrite the M106 macro totally. In addition I don't agree on some points:
- if P param is not set, the fan speed should be applied to both fans especially if there is no switching for active fan in T0/T1 macros.
- if S is not defined the current macro will raise an error. Despite the G-Codes.md doc the S parameter is optionnal. `M106` is equivalent to `M106 S255`. Klipper's base M106 allows it (in terms of code).
- `M107` overide is needed too.

Concerning the documentation I have a few comments too:
- Dual_carriage on Y-Axis is only available for cartesian printers.
- I would use the command `SET_GCODE_OFFSET X= Y= Z=` instead of dealing with the position_endstop of the dual_carriage section to adjust the position. But it's just a personnal preference since offsets are still needed for Y and Z. Also it can be adjusted and tested without restarting the firmware. So it's easier to perform an adjustement < 1mm as early as the step1.
- step1: After making a few movements to overlay the real position of the second nozzle on the first, `M114` is helpful to get the firmware position and calculate the offsets. It would be nice to mention it.
- Most calipers have an accuracy of 0.1 mm. This is not enough for fine adjustement. I use ad-hoc stl files to calibrate the offsets i.e.  https://www.thingiverse.com/thing:3873434. This perform the step2 and step3 of the guide in more or less 1 print and is pretty accurate. Few tests prints are still needed after to confirm.

charlespick:
Can you tell me more about how #4464 will overwrite the M106 macro? I just took a look at it and I can see [this comment](https://github.com/KevinOConnor/klipper/pull/4464#issuecomment-880115393) mentions the fan things but I didn't see anything about where or how to configure multiple part cooling fans so each extruder can have it's own fan speed. 
Otherwise I agree that my macro is a little stupid. I wrote it in a rush about 3 weeks ago for my own printer and didn't really test all the possible execution flows. I'll fix that and add an M107 if it's still needed after #4464. 

At this time, `[dual_carriage]` only supports cartesians period according to config_referance.md. If #4464 merges first we'll update this branch to reflect that, if this merges first then #4464 will need to update this file. 

Using g-code offsets to calibrate an IDEX printer seems like the wrong tool for the job and is a bad idea in my mind for a couple of reasons. 
1. I am not sure if Klipper would properly prevent the carriages from running into each other 
2. If you use g-code offsets for any other reason then you'd have problems. 
3. I don't believe this persists across restarts of the firmware. 

I believe all 3d-printer owners should have calipers that go down to .01mm anyway for calculating flow. But even if you didn't already have them, I don't think the solution is a printable vernier. First of all, I'm not sure how the part you linked would achieve any accuracy beyond .1mm so step 2 could accomplish that without any need to slice and with much less filament. Without a .1mm nozzle or some really fine slicer tuning, or a lot of filament and a big bed, I'm not sure a .01mm printed vernier could achieve the accuracy of a calipers either. No matter whether you're going for multiple colors in a model, or soluble supports in a functional part, you're gonna need the precision that you can only get with a dedicated measuring tool. And tbh if you have a printer with dual extruders, I think you can afford[ a .01mm calipers.](https://www.amazon.com/gp/product/B000GSLKIW/ref=ppx_yo_dt_b_search_asin_title?ie=UTF8&psc=1)

Tircown:
> Can you tell me more about how #4464 will overwrite the M106 macro? I just took a look at it and I can see [this comment](https://github.com/KevinOConnor/klipper/pull/4464#issuecomment-880115393) mentions the fan things but I didn't see anything about where or how to configure multiple part cooling fans so each extruder can have it's own fan speed.

4464 is still a work in progress concerning the macros and documentation. You can have a preview of the macros in DrumClock's comment. We are currently testing and improving the macros in order to use all idex modes.

> At this time, `[dual_carriage]` only supports cartesians period according to config_referance.md.

That's a forget, thank you to pointed out. Since #4296 is merged, hybrid-corexy and hybrid-corexz have a dual_carriage ability too.

> g-code offsets to calibrate an IDEX printer seems like the wrong tool [...]

This is however the typical use of SET_GCODE_OFFSET according to the documentation G-Codes.md. G-Code offsets are still needed for Y and probably Z, so why not X too?

> 1. I am not sure if Klipper would properly prevent the carriages from running into each other

True. In your documentation you advise the user to modify the limits after doing the adjustments. This last step should avoid collision.

> 2. If you use g-code offsets for any other reason then you'd have problems.

One can use the `X_ADJUST`,  `Y_ADJUST`,  `Z_ADJUST` parameters for that purpose.

> 3. I don't believe this persists across restarts of the firmware.

No need, it is typicaly set in the T1 macro. You will find this command in the current sample-idex.cfg.

To be clear, I don't advise to set all the X adjustement in offset. IMHO the endstop_position adjustment should be the tool for the precision range of the tape measure (>2mm), then use the G-code offset for fine tuning.

charlespick:
I haven't tested these new macros yet, don't merge

charlespick:
Using the adjust commands for g-code offsets might be a bit easier if you were expecting to make multiple adjustments in a row but you shouldn't need to. You should start with an idea of how big your printer is I think, after the first test you'll likely need to configure the macro for step 2, which will require a restart, and if you were to put your tuning into gcode offsets in t0 and t1 macros, you'd still need to restart again to get your new macros to take effect before running step 3 anyway. 

Where it says gcode offsets are useful in tool change macros is referencing when you have 2 extruders fixed relative to each other, not an idex. 

I think you should use the config options for what they're meant to do. Position_endstop exists so Klipper knows where the endstop is and can accurately place the limits that protect the printer from itself and yourself, and who knows what else will be dependent on the position_endstop value down the line. You wouldn't write a macro to replace M104 or M140 to compensate for a weird thermistor, you would define a new thermistor.  

I've fixed the fan macros I think, will test them in about an hour when my FlashForge is done printing something.

charlespick:
they work

Frix-x:
Hi, just read this issue as I'm myself interested into IDEX machines.

Coming from the professional CNC industry world, I can only second that using SET_GCODE_OFFSET in tool change macros (like T0, T1, Tx) is THE WAY to do it properly if you want to follow industrial standards from for example Fanuc or Siemens.
Machines limits are the physical one that your cariage should not go over. Offsets are the one you use to set up and calibrate and fine tune the machine. If you change the tool (spindle or hotend here) or just dismount/remount it, you need to be able to calibrate it again easily without having to change the axis limits -> in klipper, I think modifying the tool change "T0" or "T1" macro is good enough as there is not proper "tool HMI".

And I also second that using a calibration part is easier if you don't have the proper precise caliper tool. Everybody is able to print a part but not everyone have the proper caliper.
It would follow the same logic than calibrating pressure advance or flow for example.

charlespick:
> And I also second that using a calibration part is easier if you don't have the proper precise caliper tool.

I never said that was a bad idea. In fact that is what step 2 is. What I said is to achieve the precision necessary to make an IDEX actually make sense, you need to calibrate the machine down to tolerances that you cannot see with the naked eye and would require too large a build area for something like a printed vernier in some cases. You have to remember that typically on an idex, both extruders cannot reach the entire build area which shrinks your available build area. 

Just like how Marlin and Klipper have different suggested ways of tuning linear/pressure advance, I'm sure someone else is going to come up with another way to do this. Maybe you print a vernier that isn't all in one line. Maybe you print a vernier that goes up the Z axis and uses Klipper's tuning tower command. Maybe you use senseless homing to home the right one into the left one. There are other docs here that recommend using a calipers for various things and I still stand behind that if you have money for an IDEX printer you should have a calipers anyway. 

> Machines limits are the physical one that your cariage should not go over

This is literally the reason I recommend using the `position_endstop` config setting and not gcode offsets. At the end of the day here, we're using the printheads to print objects that allow us to calibrate said limits and I think we need to stop thinking about tool changes entirely. Imagine if you had 4 extruders, 2 on each carriage. That's potentially 4 gcode macros you'd have to update and play around with as belts tighten and your printer does other weird things. 

If we wanted to truly match the best practices of the CNC world then we should have something like another configuration option for dual_carriage just for this that is tuned like a probe offset. This would be managed by Klipper with the `SET_DUAL_CARRIAGE` command leaving no way for the printer to enter an invalid state. But this is not a code pr, nor are cnc machines with multiple toolheads actually that common to my knowledge - they just eject one tool and mount a new one.

charlespick:
Alright, I have an idea that might make people happy but it's gonna take a while to test so I'm converting to draft for now

eddietheengineer:
I can provide a little feedback! I just went through setting up a hybridcorexy IDEX printer in the last few days and love that this functionality is being improved in Klipper. 

I did initially use the X_max endstop position as a way to get the toolheads relatively close to aligned (within 0.5mm or so), but after that I used the vernier test print to get X and Y offsets. I wasn't expecting to need a Y offset but I ended up needing one--likely just tolerance stickup and assembly of both my toolheads. It took me a try or two to figure out the test print but it was super easy to read and visually see what offsets I needed. I wouldn't have wanted to try to measure with a caliper, once you're in the fractions of a mm range it's hard to be very precise.

The results are pretty good I think--I probably could adjust alignment a little bit (I was streaming on YouTube and just wanted to get to a test print) but it's functional.

charlespick:
Same, I am realizing that a Y offset is also necessary. So I am working on a system that uses [save_variables] to define the offsets and apply them as necessary.

KevinOConnor:
What's the status of this PR?

-Kevin

grigi:
I just had to go through figuring out the fan situation for fans on an IDEX, and right now the best solution is to have a sacrificial `[fan]` section so that all the "normal" fan feedback works as expected.
This isn't ideal as we are now PWM-ing to a pin that isn't connected.

```ini
# Sacrificial "main" fan
[fan]
pin: rpi:gpio20

[fan_generic fan_extruder]
pin: PC6

[fan_generic fan_extruder1]
pin: yz:PC6

[gcode_macro _SET_TOOL_FAN]
variable_speed: 0
gcode:
    SET_FAN_SPEED FAN=fan_{printer.toolhead.extruder} SPEED={params.SPEED}
    # Store the fan speed so that we can carry it over during toolchanges.
    SET_GCODE_VARIABLE MACRO=_SET_TOOL_FAN VARIABLE=speed VALUE={params.SPEED}

[gcode_macro M106]
rename_existing: M106.1
gcode:
    {% set s = params.S|default(255)|int %}
    _SET_TOOL_FAN SPEED={s / 255}
    # Set the original fan speed pin, else Klipper itself thinks the fan isn't on.
    M106.1 S{s}

[gcode_macro M107]
rename_existing: M107.1
gcode:
    M106 S0
```

I did not yet use #4464, but once that lands these macro's will have to be updated to sync the fans when in duplicate/mirrored mode.

charlespick:
@KevinOConnor -
I am working on macros to set the gcode offsets of the dual_carriage and set them as needed using save_variables. I'm almost done

regarding fans, I've been printing with double fan_generic sections and custom m106 and m107 macros and it's worked great for me. @grigi did you have an issue with this implementation that forced you to go about it the way you did?

grigi:
> regarding fans, I've been printing with double fan_generic sections and custom m106 and m107 macros and it's worked great for me. @grigi did you have an issue with this implementation that forced you to go about it the way you did?

The issue for me is that without having a sacrificial/dummy `[fan]` that I drive, feedback in the Klipper display interface and in Mainsail isn't working properly.

For example, mainsail only shows the separate fan controls, but not a combo one, and in the display the fan speed is always reported as 0% (and the menu options to change fan speed is missing). Thus just restoring the functionality that is there standard for single-tool printers.

Ideally I would like to completely virtualise it, instead of having to have a real pin. Hence me using a pin on the RPi host, as I don't have those connected.

Another thing is I wanted my default fan to automatically apply to the active toolhead only, for when printing gcode that has no knowledge of multiple toolheads. (e.g. not using P) This may be a bad assumption on my part, or not.

Now that I think of it, I should:
* Clean up my macros (can simplify/clean them some more)
* Provide a sample of how fan interaction works with the existing sample T0/T1 commands. (right now I swap the fan speed on tool swap)
* Extend the M106/M107 to support all the cases of:
  * Specifying P
  * Not specifying P and in `SET_DUAL_CARRIAGE_MODE MODE=FULL_CONTROL` (default) mode (swop fan between toolheads)
  * Not specifying P and in `SET_DUAL_CARRIAGE_MODE MODE=DUPLICATION|MIRRORED` modes (update all fans like yours) (This would depend on #4464 so I'll comment this block out)
  * M107 should stop everything for all above cases (so, basically calling M106 S0 for P=<unset> and each P to always cover all cases)

I will have a go at this, this weekend.

Disclaimer: Note this is my first IDEX (and its built from spare parts), and I don't have much experience on how best an IDEX should be used yet. If you believe that I'm missing a point here, please tell me.

grigi:
Minor update to a misunderstanding: According to spec here: https://www.reprap.org/wiki/G-code#M107:_Fan_Off M107 should just be an alias to M106 S0, but that it needs to disable all fans if P is not specified.

grigi:
Hi, Apologies for not getting back when I said I would. I have been reworking my config a lot to try and keep the macros I propose as simple as possible.

For getting a polished IDEX experience there is docs that wasn't clear re:
* Getting the display to show dual extruder data (it still doesn't show which tools are "active", and could be neater, etc...)
* The sample T0/T1 macros will destroy print state, causing out or movement range errors (Still need to fix that)
* Probing where one of the tools have the probe would fail disastrously if the wrong toolhead was active (fixed this already)

I feel the IDEX docs should make these gotchas clearer.

Another thing, having "perfect" docs is not a requirement, this PR significantly improves the situation already.
So I'd vote for getting this "good enough" and in so we can have other PR's that address the other shortcomings (like Menu/Probing) and improve things a little.

charlespick:
@grigi

> For example, mainsail only shows the separate fan controls, but not a combo one, and in the display the fan speed is always reported as 0% (and the menu options to change fan speed is missing). Thus just restoring the functionality that is there standard for single-tool printers.

When I started working on this, `rename_existing:` wasn't an option for `gcode_macro`s. Now that it is, I will look into ways to make this look more elegant. I also know that #4464 introduces something about syncing extruders together. I am not sure how or if that touches fans, but since I said today that I think 4464 should merge first, I'll revisit this after that happens. 

Regarding your comments on screens, I do not have a display on my printer, I use klipperscreen. I assume this will involve custom glyphs or menus or something but I literally have no knowledge of how the klipper menus work so someone else will have to introduce config examples/documentation in another pr. 

Regarding m107, I am trying to follow the reprap spec so when I revisit the fans macros after 4464 I will review those before undrafting this pr. Apart from that, I'm sure there's someone who will want the macro to work differently, and they can do that in their own config, which is the beauty of Klipper.

config/sample-idex.cfg
```diff
@@ -94,3 +100,70 @@ gcode:
...
+            SET_FAN_SPEED FAN=part_fan_1 SPEED={klipperFanSpeed}
+    {% else %}
+            SET_FAN_SPEED FAN=part_fan_0 SPEED={klipperFanSpeed}
+            SET_FAN_SPEED FAN=part_fan_1 SPEED={klipperFanSpeed}
+    {% endif %}
+
+[gcode_macro M107]
+description: Turn off all fans
+gcode:
+    M106 S0
```
grigi, line 169: The fan macros I have. It does a few things differently:
* I follow the `_extruder`/`_extuder1` convention started in the provided sample T0/T1 macros
* I save the fan speed so that you can swop the active extruder's fan when you do a T0/T1
* I set the active extruders fan only by default (This coincides with the auto-park functionality in the provided T0/T1 sections)
* I always disable all fans if `M106 S0` is passed.
* I extended M107 to take an optional `Pn` parameter, to comply with the Marlin usage of it.
* I tried to explain that if you need existing tool fan compatibility, you should use a sacrificial `[fan]` section.

These macros are tested working on my system.

You can change it as you wish

```suggestion
#######################################################################
# Part cooling fans
#######################################################################

# The following is for if your carraiges have their own part cooling fans.
# If your printer only has one part cooling fan, don't copy this section

# Sacrificial "main" fan
# It is switching on an unused pin
# This is so that fan speed is reported correctly on the display and klipper-api(moonraker) interfaces
# If you don't care about that, you can remove this and the "rename_existing" sections in M106 & M107
[fan]
pin: rpi:gpio20

[fan_generic fan_extruder]
pin: PC6

[fan_generic fan_extruder1]
pin: yz:PC6

[gcode_macro M106]
# Only rename_existing if you have a sacrificial [fan] section
rename_existing: M106.1
variable_speed: 0
gcode:
    {% set s = [[params.S|default(255)|int, 255]|min, 0]|max %}
    {% set p = params.P|default(-1)|int %}
    {% set speed = s / 255 %}

    # Save speed variable so you can use it in other macros
    # Access via {printer["gcode_macro M106"].speed}
    SET_GCODE_VARIABLE MACRO=M106 VARIABLE=speed VALUE={speed}

    {% if p == -1 %}
      # Set current active extruder fan
      {% if speed == 0 %}
        # Always turn off al fans if S0 is specified without a specific fan
        SET_FAN_SPEED FAN=fan_extruder SPEED=0
        SET_FAN_SPEED FAN=fan_extruder1 SPEED=0
      {% else %}
        SET_FAN_SPEED FAN=fan_{printer.toolhead.extruder} SPEED={speed}
      {% endif %}
    {% else %}
      # Set specified active extruder fan
      {% if p == 0 %}
        SET_FAN_SPEED FAN=fan_extruder SPEED={speed}
      {% else %}
        SET_FAN_SPEED FAN=fan_extruder1 SPEED={speed}
      {% endif %}     
    {% endif %}

    # Update core Klipper's fan speed
    # Only do this if you have a sacrificial [fan] section
    M106.1 S{s}

[gcode_macro M107]
rename_existing: M107.1
gcode:
    {% set p = params.P|default(-1)|int %}
    M106 S0 P{p}
```

config/sample-idex.cfg
```diff
@@ -94,3 +100,70 @@ gcode:
...
+gcode:
+    G28
+    G90
+    M83
+    T0 ; test T0
+    G1 X120 Y150 Z.2 F4800
+    G1 Y75 E10
+    T1 ; test T1
+    G1 X120 Y0 Z.2
+    G1 Y75 E10
```
grigi, line 119: A modified macro to test for alignment on both X & Y axis, as mentioned by @eddietheengineer 

You'll probably have to have a fine-tuning print in step 3 as well.

```suggestion
#######################################################################
# Tool alignment Calibration
#######################################################################

[gcode_macro calibrate_separation]
# YOU MUST CALIBRATE YOUR ROTATION_DISTANCE BEFORE THIS IS USEFUL!!!
# Configure this to draw 2 line segments perpendicular to the dual_carraige axis
# and 2 line segments parallel. Both sets of lines should have the same coordinate
# on the axis of the dual_carraige. If they don't print inline, adjust your endstop
# settings, or specify an offset on a carriage to line them up.
gcode:
    G28
    G90
    M83

    # X axis alignment
    T0 ; test T0
    G1 X120 Y140 Z.2 F4800
    G1 Y80 E10
    T1 ; test T1
    G1 X120 Y20 Z.2
    G1 Y80 E10

    # Y axis alignment
    T0 ; test T0
    G1 X140 Y160 Z.2 F4800
    G1 X80 E10
    T1 ; test T1
    G1 X20 Y160 Z.2
    G1 Y80 E10
```

grigi (commented):
Just some proposals :smile:

docs/Dual_Carriage.md
```diff
@@ -0,0 +1,98 @@
...
+# General configuration
+Start with studying, and possibly using parts of the [example config](../config/sample-idex.cfg)
+for IDEX printers. It has some distinct items that you may or may not
+wish to use.
+* Definition for `dual_carriage` on the X-axis, which you can change to
+be on the Y axis if that is how your printer is configured.
+* Sample macros for parking and switching toolheads. You should check
+these to make sure they make sense for your setup and update the
+position values with those that match your printer
+* Configuration for multiple part cooling fans. The `[fan]` module in
```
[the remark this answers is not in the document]
grigi: Could you also add the following about a `[display]` section:
* If you have a `[display]` section of a st7920 or uc1701 type, set the `display_group: _multiextruder_16x4` to get a dual-extruder summary view.

docs/Dual_Carriage.md
```diff
@@ -0,0 +1,98 @@
...
+Klipper only supports a single controllable fan. If you would like to
+have independently controlled part cooling fans, do not configure a
+`[fan]` section and use the template provided in the file.
+* A macro for calibrating carriage separation described later in this
+document.
+
```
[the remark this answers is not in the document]
charlespick: can you tell me more? I was actually wondering earlier if senseless homing would be possible. For example you home the left extruder and then you home the right extruder into the left extruder. This would solve the endstop_phase issue I'm investigating

docs/Dual_Carriage.md
```diff
@@ -0,0 +1,98 @@
...
+While the measuring process of this step accounts for symmetric
+horizontal expansion, you will likely achieve better results by
+calibrating the flow, temperature, and pressure advance for the
+filament you'll be using first. You'll also do better with a filament
+that is known to have better dimensional stability. Look for filaments
+that can be printed without a heated bed.
+
+Use a slicer to generate g-code for the multi-part print found in
+[docs/prints/calibrate_idex.stl](prints/calibrate_idex.stl).
+Align the slot down the middle with the axis of your `dual_carriage`.
```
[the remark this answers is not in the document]
charlespick: I'm throwing out that stl anyway so it doesn't matter 😂

docs/Dual_Carriage.md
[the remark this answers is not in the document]
charlespick: > My comment and proposed solution was to have a `[fan]` section but then override the M106 & M107 gcodes. This way the display will show the "active extruder's" fan speed.

how does the system know which fan is tied to the active extruder and switch that? how are you configuring the second fan? can you send me your config?



> Could you also add the following about a `[display]` section:
> 
> * If you have a `[display]` section of a st7920 or uc1701 type, set the `display_group: _multiextruder_16x4` to get a dual-extruder summary view.

I'll consider it. Can you send me pictures of what that looks like and do you know what to do about other displays or are those the common display types?

config/sample-idex.cfg
charlespick: OK I like the idea of a sacrificial fan section and I think we can use the T0 and T1 macros to update this value to the active extruder when that changes as well

config/sample-idex.cfg
charlespick: I'm not sure what the advantage of this macro is over the one I wrote. I was also planning on writing a macro that takes the axis limits of the carriages and divides them by 2 automatically so the user doesn't have to setup the macro. 

I also need to put in a thing about calibrating the Z axis. It won't be part of this macro, but it's something that'll need to happen

docs/Dual_Carriage.md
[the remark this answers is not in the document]
grigi: I tried sensorless homing on my custom printer, and I just couldn't get any consistency better than ~0.2mm. which is extremely noticeable on the alignment prints. I'd get it lined up, then on the next home it would not line up anymore.

In this nice doc: https://duet3d.dozuki.com/Wiki/Stall_detection_and_sensorless_homing#Section_Limitations_of_stall_detection
Note the +-2 full step accuracy. Multiply this variance by 2 (one for each side) and you see.

Part of the inaccuracy (as I understand it) is that the EM feedback changes depending on motor coil heat. So cold motors, once heated up, seems to home less sensitively.

I suppose some motors/setup might achieve acceptable accuracy, but my repurposed A8 motors have a very high inductance, and is more prone to variations. (as I understand it)

docs/Dual_Carriage.md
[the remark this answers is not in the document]
grigi: Those are the common 12864 and 12864 mini display types. I have a repurposed CR10 display which is a Reprapdiscount 12864 clone without an SD card reader.

I'll attach the T0/T1 macros.  they are heavily based on the sample ones.

config/sample-idex.cfg
grigi: All this was is to try and see if your hotends are vastly out in Y offset. Mine is (probably due to plastic print warping, or assembly variance) out about 0.12mm. But this seems to be very static.
The issue that had me hesitating on this is that the same-axis variations are very small, so we need maybe only a step3-equivalent for it.
You can probably ignore this proposal.

Calibrating Z, I basically did a paper test separately. The issue is my hotends are about 0.5mm different in Z which is crazy high, and the eva2 hotend system doesn't have an easy way of adjusting this.
I would like to get them within 0.05mm so that simultaneous printing is possible, and need a way of easily adjusting them. but that's my design flaw.
(but bed uneven-ness will in any case require a fat 0.3mm first layer to even have a chance of success in any case)

docs/Dual_Carriage.md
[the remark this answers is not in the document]
grigi: It might be better to say if it's a common 12864 or 12864 mini display.
Here is a picture:

image

config/sample-idex.cfg
grigi: Here is the T0/T1 sections I have that swop the fans between active extruders:
```ini
#######################################################################
# Tool changing
#######################################################################
[gcode_macro _PARK_extruder]
gcode:
    SAVE_GCODE_STATE NAME=park0
    G90
    G1 X283 F10800
    SET_FAN_SPEED FAN=fan_extruder SPEED=0
    RESTORE_GCODE_STATE NAME=park0

[gcode_macro T0]
gcode:
    {% if printer.toolhead.extruder != 'extruder' %}
      SET_FAN_SPEED FAN=fan_extruder SPEED={printer["gcode_macro M106"].speed}
      _PARK_{printer.toolhead.extruder}
      ACTIVATE_EXTRUDER EXTRUDER=extruder
      SET_DUAL_CARRIAGE CARRIAGE=0
      #SET_GCODE_OFFSET Y=0
    {% endif %}

[gcode_macro _PARK_extruder1]
gcode:
    SAVE_GCODE_STATE NAME=park1
    G90
    G1 X-63 F10800
    SET_FAN_SPEED FAN=fan_extruder1 SPEED=0
    RESTORE_GCODE_STATE NAME=park1

[gcode_macro T1]
gcode:
    {% if printer.toolhead.extruder != 'extruder1' %}
      SET_FAN_SPEED FAN=fan_extruder1 SPEED={printer["gcode_macro M106"].speed}
      _PARK_{printer.toolhead.extruder}
      ACTIVATE_EXTRUDER EXTRUDER=extruder1
      SET_DUAL_CARRIAGE CARRIAGE=1
      #SET_GCODE_OFFSET Y=15
    {% endif %}
```

grigi:
> Getting the display to show dual extruder data (it still doesn't show which tools are "active", and could be neater, etc...)

On this, I was thinking a separate PR that would add an asterisk (or something) after the toolhead's temperature on the standard idex display, and do a separate PR if I find a clean solution for it. It would have to support #4464 modes to understand if its either/both.

docs/Dual_Carriage.md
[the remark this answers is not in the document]
grigi: Oh have them all home against each other? Hmm, could work...
But I have started crimping wires for microswitch endstops, so will probably use those with phase assistance.

KevinOConnor:
> I tried sensorless homing on my custom printer, and I just couldn't get any consistency better than ~0.2mm. which is extremely noticeable on the alignment prints. I'd get it lined up, then on the next home it would not line up anymore.

Out of curiosity, have you tried using sensorless homing with endstop_phases?

-Kevin

config/sample-idex.cfg
charlespick: Interesting. I am leary of setting the fan speed on tool change. I was thinking the custom m106 remains the only thing that modifies the actual fan speeds and the tool change macro simply updates the sacrificial fan to match the speed of the currently selected extruder.

grigi:
> Out of curiosity, have you tried using sensorless homing with endstop_phases?

I did not. The variations felt too large to me, and the duet3d docs stating accuracy of +-2 full steps made me doubt that that it would be able to pull it in (as how I understand it endstop_phase would fix +-0.5 full steps only)

I did however do a home, move 10mm away then 10mm back, and sometimes I can easily slide a paper between the head and the block, and some other times it bangs quite hard.

config/sample-idex.cfg
grigi: I think, these macros should only do it if a Pn was not specified. 
I also have no idea how this will work with the duplication/mirror modes, we should probably not touch the fans in that case.

Possibly need a gcode_variable that states whether fanspeed should be swapped (and controlled from M106).
I think we could reset `{printer["gcode_macro M106"].speed}` to `-1` (and default to that) to indicate that no fan swapping should occur.
Should then rename it to swap_speed or something for clarity.

grigi: >  the tool change macro simply updates the sacrificial fan to match the speed of the currently selected extruder.
I missed this. It's a great idea!

:thinking: Now how do I merge both those ideas in a clean way?

grigi: I updated the tool changing & M106 macros to work as I summarized.
It will now only swap the tool fans if the last call was a `M106 S[>0]` and no P was specified.
Else it will update the sacrifical fan speed to the speed of the current toolhead for informational purposes.

```ini
#######################################################################
# Tool changing
#######################################################################
[gcode_macro _PARK_extruder]
gcode:
    SAVE_GCODE_STATE NAME=park0
    {% if printer["gcode_macro M106"].swap_speed != -1 %}
      SET_FAN_SPEED FAN=fan_extruder SPEED=0
    {% endif %}
    G90
    G1 X283 F10800
    RESTORE_GCODE_STATE NAME=park0

[gcode_macro T0]
gcode:
    # Only do anything if the current tool is not T0
    {% if printer.toolhead.extruder != 'extruder' %}
      {% set fan_speed = printer["gcode_macro M106"].swap_speed %}
      {% if fan_speed != -1 %}
        SET_FAN_SPEED FAN=fan_extruder SPEED={fan_speed}
      {% else %}
        # Update core Klipper's fan speed to the fan speed of the active toolhead
        # Only do this if you have a sacrificial [fan] section
        M106.1 S{printer["fan_generic fan_extruder"].speed * 255}
      {% endif %}
      _PARK_{printer.toolhead.extruder}
      ACTIVATE_EXTRUDER EXTRUDER=extruder
      SET_DUAL_CARRIAGE CARRIAGE=0
      #SET_GCODE_OFFSET Y=0
    {% endif %}

[gcode_macro _PARK_extruder1]
gcode:
    SAVE_GCODE_STATE NAME=park1
    {% if printer["gcode_macro M106"].swap_speed != -1 %}
      SET_FAN_SPEED FAN=fan_extruder1 SPEED=0
    {% endif %}
    G90
    G1 X-63 F10800
    RESTORE_GCODE_STATE NAME=park1

[gcode_macro T1]
gcode:
    # Only do anything if the current tool is not T1
    {% if printer.toolhead.extruder != 'extruder1' %}
      {% set fan_speed = printer["gcode_macro M106"].swap_speed %}
      {% if fan_speed != -1 %}
        SET_FAN_SPEED FAN=fan_extruder1 SPEED={fan_speed}
      {% else %}
        # Update core Klipper's fan speed to the fan speed of the active toolhead
        # Only do this if you have a sacrificial [fan] section
        M106.1 S{printer["fan_generic fan_extruder1"].speed * 255}
      {% endif %}
      _PARK_{printer.toolhead.extruder}
      ACTIVATE_EXTRUDER EXTRUDER=extruder1
      SET_DUAL_CARRIAGE CARRIAGE=1
      #SET_GCODE_OFFSET Y=15
    {% endif %}


#######################################################################
# Part cooling fans
#######################################################################

# The following is for if your carraiges have their own part cooling fans.
# If your printer only has one part cooling fan, don't copy this section

# Sacrificial "main" fan
# It is switching on an unused pin
# This is so that fan speed is reported correctly on the display and klipper-api(moonraker) interfaces
# If you don't care about that, you can remove this and the "rename_existing" sections in M106 & M107
[fan]
pin: rpi:gpio20

[fan_generic fan_extruder]
pin: PC6

[fan_generic fan_extruder1]
pin: yz:PC6

[gcode_macro M106]
# Only rename_existing if you have a sacrificial [fan] section
rename_existing: M106.1
# The variable that controls fan speed swopping if not specifying P parameter
# -1 means the control is disabled, a value of 0-1 is the requested fan speed.
# Access via {printer["gcode_macro M106"].swap_speed}
variable_swap_speed: -1
gcode:
    {% set s = [[params.S|default(255)|int, 255]|min, 0]|max %}
    {% set p = params.P|default(-1)|int %}
    {% set speed = s / 255 %}

    # Set speed to -1 by default
    SET_GCODE_VARIABLE MACRO=M106 VARIABLE=swap_speed VALUE=-1

    {% if p == -1 %}
      # Set current active extruder fan
      {% if speed == 0 %}
        # Always turn off al fans if S0 is specified without a specific fan
        SET_FAN_SPEED FAN=fan_extruder SPEED=0
        SET_FAN_SPEED FAN=fan_extruder1 SPEED=0
      {% else %}
        # Opt into fan speed swop control
        SET_GCODE_VARIABLE MACRO=M106 VARIABLE=swap_speed VALUE={speed}
        SET_FAN_SPEED FAN=fan_{printer.toolhead.extruder} SPEED={speed}
      {% endif %}
    {% else %}
      # Set specified active extruder fan
      {% if p == 0 %}
        SET_FAN_SPEED FAN=fan_extruder SPEED={speed}
      {% else %}
        SET_FAN_SPEED FAN=fan_extruder1 SPEED={speed}
      {% endif %}     
    {% endif %}

    # Update core Klipper's fan speed
    # Only do this if you have a sacrificial [fan] section
    M106.1 S{s}

[gcode_macro M107]
rename_existing: M107.1
gcode:
    {% set p = params.P|default(-1)|int %}
    M106 S0 P{p}
```

config/sample-idex.cfg
grigi: I tried to make a generic macro for this that reads the axis from the config, but then realised that `printer.dual_carriage` doesn't exist for cartesion IDEX, only hybrid_core* IDEX :sob: 
And #4464 doesn't address it either. Meaning it doesn't add that functionality to cartesian IDEX printers either.

charlespick:
hey everyone who's watching this, 

as I get ready to wrap this up, if someone could answer my question on just how endstop phases work that'd be great: https://klipper.discourse.group/t/endstop-phase-questions/888 

thanks

grigi:
> as I get ready to wrap this up, if someone could answer my question on just how endstop phases work that'd be great: https://klipper.discourse.group/t/endstop-phase-questions/888

I think you're using them as they are intended.
As I understand it, you do a home, check the endstop phases, repeat ~5 times.
Take the average values and write them to the config.

It should now correct to that endstop-phase when homing.
Note that endstop phases can only correct for ±½ step error. Meaning your regular endstops need to be accurate to ~0.1mm under normal circumstances.

I tried to do endstop_phases on my IDEX I'm building now, and it was weird in that it doesn't feel like it was doing anything.
I chalked that up to sensorless homing not being accurate enough, but if you're getting the same issue with regular endstops, then maybe there's been a regression?

config/sample-idex.cfg
```diff
@@ -44,13 +50,21 @@ gcode:
 
 # Activate the primary extruder
 [gcode_macro T0]
+variable_offset_applied: 0
 gcode:
-    PARK_{printer.toolhead.extruder}
+    {% set svv = printer.save_variables.variables %}
```
grigi, line 55: I can see how saved variables could be useful, but this is complicating the understanding of what is required for setup.
It should be kept quite simple I think.
Possibly split it into basic, and advanced?

It's a flaw in my example macros as well. You really need succinct examples for basic docs, and everything needs to be obvious or well explained.

config/sample-idex.cfg
```diff
@@ -44,13 +50,21 @@ gcode:
...
+    {% if "x" in printer.toolhead.homed_axes %} # this check ensures compatibility with Cura
+        PARK_extruder1
```
grigi, line 57: I prefer the `PARK_{printer.toolhead.extruder}` over this, as that will scale to more than 2 extruders if you have such a setup. (e.g. this docs can still be useful for people doing toolchangers).

config/sample-idex.cfg
charlespick: I can add comments but I think this is the correct way to go here. Every commercial IDEX printer has a way to adjust the offset of the secondary extruder at runtime and the only way to do that and have the change persist across restarts is to use save variables. I also don't know how it's really complicating things at all. I think having a basic and advanced example would (1) just make the whole thing more complicated and (2) be unnecessary because I think everything else about IDEX printers is more complex anyway.

config/sample-idex.cfg
charlespick: I will switch back when I go through this variableifying everything. I've been hesitant to so far however because they don't always seem to work the way I expect. For example, I wanted to use the axis.maximum value to park the right extruder, but even when eh dual_carraige is enabled, the axis.maximum value is that of the primary carriage.

config/sample-idex.cfg
grigi: It's a valid thing that basically every IDEX does things in a certain way, but remember if the macro is scary long, people are just going to copy it and not read it.
What are the purposes of this PR? To deliver a perfect solution or give people enough information to help themselves?

charlespick: > What are the purposes of this PR? To deliver a perfect solution or give people enough information to help themselves? 

Great question. My hope is to have enough documentation so that someone can *easily* setup an IDEX printer with Klipper. 

Now, defining "easily" is a complex thing to do for something that is inherently difficult. I would rather have a macro that works and that people won't read than a macro like the current one. The current T0 and T1 macros only have a Y offset, and require a whole restart just to update the x and y alignment. There is no mention of Z alignment. 

Basically, `if the macro is scary long, people are just going to copy it and not read it` yes, but if the macro doesn't work or is difficult to use then they aren't gonna use it at all. I'd rather them use it and then understand it later as opposed to spend 2 months gathering information from around the internet and experimenting and wasting a lot of pla like I did.

grigi: Fair point. So you're aiming for a solution that will work sanely out the box.

I do think that it would be worth it to split it in 2 parts (PR's):
1. Get the docs scaffold in re IDEX
2. Work on the adding stuff to the scaffold.

Because that allows other people, e.g. like me, to help.

charlespick: Every time people mention splitting PRs I cringe because I have no idea how to split a branch without just branching and reverting different changes which seems dumb. 

I'll accept pull requests against my branch if you want... 

I think this conversation branch is getting away from where it started however which is the Save Variables element.

charlespick: @grigi Alright, a person on the internet is admitting they were wrong...

I have macros working but they still aren't perfect. I'm looking at things like z hop when changing extruders and stuff for reducing stringing but that is really getting nitty gritty and I'm probably not going to have a perfect printer for a couple months. 

I'm going to clean this up to the point you described above and I think we'll be ready for final review

grigi: 🤣  No worries 
Cura does hop/retract/temp/fan for you,  only need to park other toolhead on Tx.
Prusaslicer does nearly nothing, so for now I'm just using Cura 🤷 

Time is always a rarity

charlespick: I was having retraction quality problems with Cura :( I was able to get retraction to work with the same speed and distance in every slicer but Cura for some reason. I like the plugins that are available for cura, I like the interface for the most part, and I like the theory behind some of their decisions in structuring profiles but in the end tuning print quality was taking more time than just learning how to use superslicer

charlespick:
@grigi can you check that I integrated your fan work correctly?

grigi:
@charlespick The fan macros look good :+1: 

One other comment, I found that printing this: https://www.thingiverse.com/thing:6632 and both measuring the height and then looking at the first layer visually, so that they are identical, helped me dial in the Z offset very well.

It's important to also mention Y offset tuning, as for example, even though I have 2 v6-clone hotends (probably different manufacturers) and mounted in the same way, I have a colossal Y offset of 1.23mm

charlespick:
I mentioned y offset tuning a couple of times 👌

grigi:
You know after the work that I put in to help @charlespick to get this ready, and it just getting ignored like this is quite discouraging.

KevinOConnor:
> You know after the work that I put in to help @charlespick to get this ready, and it just getting ignored like this is quite discouraging.

I apologize.  The ticket should not have been closed as there was feedback provided after I asked for it.

Unfortunately, I'm severely backlogged on reviews and I fear I have to face the reality that I can not review all the PRs that Klipper receives.  At this point, I'm going to try to focus on changing the PR review process (ideally by getting others to help with reviews) instead of trying to review all the current PRs.  I know this is slow and discouraging, but I don't know of a good alternative at this time.

-Kevin

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+        SAVE_VARIABLE VARIABLE=yoffset VALUE={ newZ|float }
+    {% endif %}
+
+[gcode_macro calibrate_separation]
+gcode:
+    G28
+    G90
+    M83
+    T0 ; test T0
+    G1 X120 Y150 Z.2 F4800
```
dmbutyugin, line 184: The way the tuning guide goes is that the Z height for the hotends is calibrated (step 4) after this macro is used (step 2). And the step 4 says "To print with low layer thicknesses, you'll probably find that the error between your nozzles in the Z axis is more than your layer height itself." I'm thinking if the layer height .2 is too small at this point?

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+    G1 Y75 E10
```
dmbutyugin, line 185: With the currently given parameters this will print a line with ~1.6 mm width
```
pi * (1.75/2)^2 * 10 / (.2 * (150 - 75)) ~= 1.6
```
Is that intentional? Would it make sense to reduce the suggested E value?

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+
+# For completeness, you can add a [fan] section with an unused pin
+[fan]
+pin: rpi:gpio20
+
+[gcode_macro M106]
+# Only rename_existing if you have a sacrificial [fan] section
+rename_existing: M106.1
+# The variable that controls fan speed swopping if not specifying P parameter
+# -1 means the control is disabled, a value of 0-1 is the requested fan speed.
```
dmbutyugin, line 207: It could be just me, I got confused a bit at first by a description of P parameter. Maybe you could say that P is the index of a fan (0-1)? *And -1 is the active extruder fan.

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+    # Only do this if you have a sacrificial [fan] section
+    M106.1 S{s}
+
+[gcode_macro M107]
+rename_existing: M107.1
+gcode:
+    {% set p = params.P|default(-1)|int %}
+    M106 S0 P{p}
+
+[gcode_macro M107]
```
dmbutyugin, line 248: This is duplicate?

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+      {% if p == 0 %}
+        SET_FAN_SPEED FAN=fan_extruder SPEED={speed}
+      {% else %}
+        SET_FAN_SPEED FAN=fan_extruder1 SPEED={speed}
+      {% endif %}
+    {% endif %}
+
+    # Update core Klipper's fan speed
+    # Only do this if you have a sacrificial [fan] section
+    M106.1 S{s}
```
dmbutyugin, line 240: @grigorig so, this is only useful for reporting?

config/sample-idex.cfg
```diff
@@ -44,13 +50,30 @@ gcode:
...
+variable_offset_applied: 0
 gcode:
-    PARK_{printer.toolhead.extruder}
+    {% set svv = printer.save_variables.variables %}
+    {% if "x" in printer.toolhead.homed_axes %} # this check ensures compatibility with Cura
+        PARK_{printer.toolhead.extruder}
+    {% endif %}
+    {% set fan_speed = printer["gcode_macro M106"].swap_speed %}
+      {% if fan_speed != -1 %}
+        SET_FAN_SPEED FAN=fan_extruder SPEED={fan_speed}
```
dmbutyugin, line 61: I wonder what is supposed to happen to the fan speed of the other extruder? Would it make sense to really swap their speeds, so that the fan of the parked extruder get the fan speed of the previously parked extruder?

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+[fan]
+pin: rpi:gpio20
+
+[gcode_macro M106]
+# Only rename_existing if you have a sacrificial [fan] section
+rename_existing: M106.1
+# The variable that controls fan speed swopping if not specifying P parameter
+# -1 means the control is disabled, a value of 0-1 is the requested fan speed.
+# Access via {printer["gcode_macro M106"].swap_speed}
+variable_swap_speed: -1
```
dmbutyugin, line 209: Maybe this would be more like 'active_fan_speed' ? Because the active fan gets this speed. But I do not insist on this.

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+    {% set p = params.P|default(-1)|int %}
+    M106 S0 P{p}
+
+[save_variables]
+filename: ~/klipper_config/variables.klip
+# this is used for saving and restoring the idex offsets
+# create the file and insert the folloring content (without the #) to start
+#
+# [Variables]
+# xoffset = 0.0
```
dmbutyugin, line 260: I didn't try it myself, but would it be possible to use `|default(0.0)` at the place of their use? Or `is defined`. So that the user does not have to manually add this section to the save variable file. But maybe it doesn't work that way.

dmbutyugin (commented):
Please be advised that I do not own an IDEX printer. So, I looked at the PR from the theoretical point of view. So, take my comments with a grain of salt :)

config/sample-idex.cfg
```diff
@@ -44,13 +50,30 @@ gcode:
...
+        # Update core Klipper's fan speed to the fan speed of the active toolhead
+        # Only do this if you have a sacrificial [fan] section
+        M106.1 S{printer["fan_generic fan_extruder"].speed * 255}
+    {% endif %}
     ACTIVATE_EXTRUDER EXTRUDER=extruder
     SET_DUAL_CARRIAGE CARRIAGE=0
-    SET_GCODE_OFFSET Y=0
+    {% if printer["gcode_macro T0"].offset_applied == 1 %}
+        SET_GCODE_OFFSET  X_ADJUST={ -(svv.xoffset) } Y_ADJUST={ -(svv.yoffset) }
+        SET_GCODE_OFFSET Z_ADJUST={ -(svv.zoffset) } MOVE=1
```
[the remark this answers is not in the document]
charlespick: Readability

config/sample-idex.cfg
charlespick: Idk it helps with getting the line to stick to the bed

config/sample-idex.cfg
```diff
@@ -90,7 +114,149 @@ gcode:
...
+    {% endif %}
+
+    {% if params.Y_ADJUST is defined %}
+        {% set newY = params.Y_ADJUST|float + oldY %}
+        SAVE_VARIABLE VARIABLE=yoffset VALUE={ newY|float }
+    {% endif %}
+
+    {% if params.Z_ADJUST is defined %}
+        {% set newZ = params.Z_ADJUST|float + oldZ %}
+        SAVE_VARIABLE VARIABLE=yoffset VALUE={ newZ|float }
```
kodonnell, line 175: `yoffset` should probably be `zoffset` = )

config/sample-idex.cfg
grigi: If you are talking of the unused fan pin, then it also allows you to alter the fan speed for the active tool from the display menu(or a webUI).

Most of the benefit is seeing the fan speed correctly reflect on the 12864 display for me though.

config/sample-idex.cfg
grigi: So, using CURA, if you tell it you have separate tools it will set the fan speed of each manually and the variable_swap_speed will stay -1 (Because P is set) and let CURA control everything.

If using PrusaSlicer, it won't emit the P{n} so it expects the active tool's part fan to be controlled.

Currently having the part fan of the parked extruder run will cool somewhere on the bed that isn't where printing is happening. In theory this unnecessary draft could cause part warping so I set it to turn off entirely to minimize risks.
This is based on the assumption that the heatsink fan is adequate to not cause the tool to melt.

config/sample-idex.cfg
grigi: I do like that name convention. Ultimately @charlespick will have to alter this PR to have that change, so it's up to him.

DrumClock:
Hio,
there are a lot of bugs in **sample-idex.cfg** ..... e.g.

you define:
[fan_generic part_fan_0]
pin: PE5

[fan_generic part_fan_1]
pin: PD13

but you use:
  SET_FAN_SPEED FAN = fan_extruder SPEED = 0
  SET_FAN_SPEED FAN = fan_extruder1 SPEED = 0

charlespick:
this pull request has gotten really confusing to me as more and more code snippets keep getting piled on. right now I am too busy with work and school to keep contributing to this pr that I started in the summer, especially since my idex printer is the one I use for school so I can't be tinkering around with it constantly. 

when spring break comes around I might make one last push to get it to a workable state.

charlespick:
My IDEX is not working after sitting in the closet. Not exactly sure what's going on and I don't know when I'll be getting back to it. I'm going to close this now but I encourage @grigi and @dmbutyugin to take what they are comfortable working on from here and charlespick/klipperconf to start a new discussion.

Sorry it has to end this way :( I really appreciate everyone else who contributed to this.

grigi:
Sigh, timing isn't great. I'll only have a look at it in probably 3 months. Goods and people are traveling together during emigration :shrug: 
Maybe I'll update documentation or something. A subset of this PR.

grigi:
Finally basically settled again. I have started on this, can I just copy what I like and credit you in the PR?

charlespick:
@grigi Go ahead I don't have a dual carriage printer anymore so I don't have anything to contribute beyond what I have already published here

PR #4515, opened 2021-07-23
CONFIGFILE: Added BACKUP and RESTART to SAVE_CONFIG

randellhodges:
Added optional BACKUP and RESTART parameters to prevent creating the
backup file and/or restarting klipper. If unspecified it maintain current
behavior (backing up and restarting)

Signed-off-by: Randell L Hodges <rhodges@taxfodder.com>

### DIFF
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -329,11 +329,13 @@ class PrinterConfig:
             f = open(temp_name, 'wb')
             f.write(data)
             f.close()
-            os.rename(cfgname, backup_name)
+            if gcmd.get_int('BACKUP', 1):
+                os.rename(cfgname, backup_name)
             os.rename(temp_name, cfgname)
         except:
             msg = "Unable to write config file during SAVE_CONFIG"
             logging.exception(msg)
             raise gcode.error(msg)
         # Request a restart
-        gcode.request_restart('restart')
+        if gcmd.get_int('RESTART', 1):
+            gcode.request_restart('restart')


zellneralex:
You need also to describe your add one here https://github.com/KevinOConnor/klipper/blob/master/docs/G-Codes.md

KevinOConnor:
Thanks.  Can you provide a high-level use case for this change?  As high-level feedback, it does not look like a good candidate for the Klipper master branch, as I can't see a compelling reason to not backup or to not restart.

-Kevin

randellhodges:
Thanks for the feedback.

I'm a save early, save often guy. I guess that goes back to my days of working with Visual Basic 6 where you could spend the day writing code, debug and crash the editor and lose everything. Now days, Ctrl-S is hardwired in my brain and the idea of anything hanging around for any length of time unsaved really tries my OCD.

RESTART: Creating meshes, renaming them, deleting them, etc leaves a lot of data "floating" in the session until you save. It can get annoying that every time you make a small change, such as a probe calibrate or a bed mesh, that to save it you have to restart everything, then home it all, qgl, etc just to ensure you don't loose all you changes. I've been hit a few times with leaving things around in session and forgetting to save them because I didn't want to go thru that hassle. Hours (many, many hours sometimes) later, I forget that I didn't save and power off the printer or whatever, and boom. Gone. Then after a fail print or whatever later, I remember, oh, I didn't save my changes last time.

BACKUP: This sort of goes with the above statement. I save a lot and I have seen others making similar requests to prevent the backups. Having it default to creating a backup is great. For most it is best option, but for some, those backups build up and having to go in and delete them by hand is an annoyance. I know, not the end of the world, but a small convenience parameter seemed like a low impact addition.

If these changes are something that will be accepted, I'll be sure to update the G-Code documentation zellneralex mentioned.

MoellerDi:
I love this PR and would like to support it. My use case is as follows.....

In my "START_GCODE" macro I simply call a "LOAD_OR_CREATE_MESH BED_TEMPERATURE={BED_TEMP}" macro to either load or create a mesh-profile (in case the profile is not already in my printer.cfg). A normal call of "SAVE_CONFIG" ( w/o RESTART=0 ) at that time would restart the system and I would need to restart the print again. So one would need to be aware or have to remember at the end of a 3-9h print that a new mesh was created in the beginning of the print and that ""SAVE_CONFIG" should be called before you (or octoprint) will be switching off the printer.

I'm including parts of my macros for better understanding below:
```
[gcode_macro START_GCODE]
description: Just START_GCODE
default_parameter_PREHEAT_TEMP: 0
default_parameter_EXTRUDER_TEMP: 0
default_parameter_BED_TEMP: 0
<...>
  M117 Homing...
  G28 ;homing XYZ

  M117 Calibrating Z tilt...
  Z_TILT_ADJUST
  G28Z ;homing Z again

  M117 Mesh bed leveling...
  LOAD_OR_CREATE_MESH BED_TEMPERATURE={BED_TEMP} # FORCE=1

  M117 Preparing for print...
<...>

[gcode_macro LOAD_OR_CREATE_MESH]
description: Load or probe mesh for given temperature
default_parameter_BED_TEMPERATURE: 0
default_parameter_FORCE: 0
gcode:
  {% if BED_TEMPERATURE|int < 30 %}
    M118 Your bed temp is to low, make sure it is at least 30 or higher
  {% else %}
     {% if printer.configfile.config["bed_mesh " + BED_TEMPERATURE] is defined and FORCE|int == 0 %}
        BED_MESH_PROFILE LOAD={BED_TEMPERATURE}
        M118 Succesfully loaded bed_mesh {BED_TEMPERATURE}
     {% else %}
        {% if printer.configfile.config["bed_mesh " + BED_TEMPERATURE] is defined and FORCE|int == 1 %}
          BED_MESH_PROFILE REMOVE={BED_TEMPERATURE}
        {% endif %}
        M118 bed_mesh not defined, your bed temperature will go up!
        M118 We will probe the bed and save the mesh as bed_mesh {BED_TEMPERATURE}
        ADD_BED_MESH TARGET_TEMP={BED_TEMPERATURE}
    {% endif %}
  {% endif %}

[gcode_macro ADD_BED_MESH]
description: Probe and save Mesh for Temperature
default_parameter_TARGET_TEMP: 30
gcode:
    M190 S{TARGET_TEMP} ;Wait for the bed to hit TARGET_TEMP
    Z_TILT_ADJUST ;Calibrating Z tilt...
    G28Z ;homing Z only
    BED_MESH_CALIBRATE
    BED_MESH_PROFILE SAVE={TARGET_TEMP}
    SAVE_CONFIG RESTART=0
```

zellneralex:
@MoellerDi that can easily be done with an Variable that you set e.g to 1 and a delayed_gcode that you run in your print_end macro. See `https://github.com/zellneralex/klipper_config/blob/master/bed_mesh.cfg#L166 ` as an example

MoellerDi:
@zellneralex thanks for sharing. I'll give it a try. I'm new and just started with klipper but I already like how powerful it is and how much freedom one got in writing nice delayed_gcode and gcode_macros.

KevinOConnor:
Thanks.  I understand.  However, I'm not sure this is a good candidate for the Klipper master repository at this time.  My main concern is that I get feedback that SAVE_CONFIG is often confusing to users.  (And, indeed, it is a bit confusing the way it overwrites the existing config with an unusual text block at the end of the file.)  I'm leery of making the SAVE_CONFIG command (and documentation) more complicated with infrequently used options.

-Kevin

randellhodges:
If the concern is writing of the save data at the end of the file, maybe by default, it should write it out to a different file that is included. (printer.cfg and then settings.cfg (or a different name))

I go about it the other way at the moment. My printer.cfg basically just includes a config.cfg where I put all my real config so that SAVE_CONFIG can write to the end of a file that I never change.

Having to restart, re-home, qgl, etc, just to flush changes from memory to disk feels like there is a lot of room for improvement.

So much of Klipper is designed around the end user experience but this part is very lacking. It's like opening notepad and doing a file->save rebooting your computer. You could run all day/week and never save that notepad, but if you lose power or forget to save before you power off, you lose your work.

I'm sure we can figure out a way that satisfies both sides of the issue.

KevinOConnor:
Yes, it's entirely possible there's another way forward or that I've "made a bad call" here.  I suspect the near future will involve maintenance of this feature outside of the main Klipper repo though.

Cheers,
-Kevin

PR #4516, opened 2021-07-24
correct pins for ender3max display

shawnifoley:
knob/clicker doesnt work with current values. Switching to values from printer-creality-ender3pro-2020.cfg work as expected. 

Signed-off-by: Shawn Foley shawnifoley@gmail.com

### DIFF
--- a/config/printer-creality-ender3max-2021.cfg
+++ b/config/printer-creality-ender3max-2021.cfg
@@ -146,5 +146,5 @@ lcd_type: st7920
 cs_pin: PB12
 sclk_pin: PB13
 sid_pin: PB15
-encoder_pins: ^PD2, ^PD3
-click_pin: ^!PC0
+encoder_pins: ^PB14, ^PB10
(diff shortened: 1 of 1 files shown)


KevinOConnor:
@ChadDevOps - do you have any comments on this change?

-Kevin

ChadDevOps:
@KevinOConnor @shawnifoley I'm no longer running klipper on my Ender 3 Max. I have no objections to the change if it's working on stock with the new values.

shawnifoley:
Ill just keep both values and a person can uncomment out based on there board

KevinOConnor:
Okay, thanks.  However, I don't understand the new comments that were added - the entire config is only valid for a 32bit board.  If the pins were incorrect before, then best to just correct them.

-Kevin

shawnifoley:
Corrected

KevinOConnor:
Thanks.

-Kevin

PR #4520, opened 2021-07-26
Update htu21d.py: Fix humidity reading

dawidmpunkt:
module: htu21d.py

Just a quick fix for the humidity reading.
Line 202: should be self.humidity instead of self.temp

Signed-off-by: Dawid Murawski <dawid.m@gmx.net>

### DIFF
--- a/klippy/extras/htu21d.py
+++ b/klippy/extras/htu21d.py
@@ -199,7 +199,7 @@ class HTU21D:
                 if (self.humidity < 0):
                     #due to RH accuracy, measured value might be
                     # slightly less than 0 or more 100
-                    self.temp = 0
+                    self.humidity = 0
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Change looks fine, but would need a signed-off-by line if you want me to commit directly (see https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md ).

Otherwise, I will make the change using my signed-off-by line.

-Kevin

dawidmpunkt:
I changed the comment above. Sorry, for the inconvenience. Now that I read the contributing guide again, it is clear to me.

KevinOConnor:
Thanks.

-Kevin

PR #4521, opened 2021-07-26
bed_screws: Finish bed_screws_adjust after 4 (corrected: n screws) consecutive accepts

chrenderle:
It always bothered me that on my Ender 3 Pro with 4 bed screws I had to accept 7 times if I adjusted the first screw and after that screw every screw was fine. So I rewrote the code that it would finish the bed_screws_adjust command when you accepted all screws without a adjust in between no matter which screw you accepted first.

In the commit title I wrote 4 consecutive accepts but I meant as many as the printer has screws but just thought of my Ender 3.

I tested the code (except for the fine tuning) on my Ender 3 where it worked fine. 

This is my first contribution to a open-source project ever so please let me know if I did anything wrong.

### DIFF
--- a/klippy/extras/bed_screws.py
+++ b/klippy/extras/bed_screws.py
@@ -17,7 +17,8 @@ class BedScrews:
         self.printer = config.get_printer()
         self.state = None
         self.current_screw = 0
-        self.adjust_again = False
+        self.accepted_screws = 0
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #4522, opened 2021-07-26
bed_mesh_calibrate: Added PROFILE name option

randellhodges:
Added an optional parameter to allow bed_mesh_calibrate directly into the named profile instead of `default`. If not provided, it will default to the current behavior of `default`.

A use case would be when a person creates a mesh for a different temperature or maybe a different build surface, they can specify the name then. Otherwise they have to remember to run bed_mesh_profile save=name when they are done and potentially overwrite their `default` mesh if they already have one.

### DIFF
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -345,13 +345,17 @@ are identified in green.
 
 ### Calibration
 
-`BED_MESH_CALIBRATE METHOD=[manual | automatic] [<probe_parameter>=<value>]
+`BED_MESH_CALIBRATE PROFILE=name METHOD=[manual | automatic] [<probe_parameter>=<value>]
  [<mesh_parameter>=<value>]`\
+_Default Profile:  default_\
 _Default Method:  automatic if a probe is detected, otherwise manual_
 
-Initiates the probing procedure for Bed Mesh Calibration.  If `METHOD=manual`
-is selected then manual probing will occur.  When switching between automatic
-and manual probing the generated mesh points will automatically be adjusted.
+Initiates the probing procedure for Bed Mesh Calibration.
+
+The mesh will be saved into a profile specified by the `PROFILE` parameter,
+or `default` if unspecified. If `METHOD=manual` is selected then manual probing
+will occur.  When switching between automatic and manual probing the generated
+mesh points will automatically be adjusted.
 
 It is possible to specify mesh parameters to modify the probed area.  The
 following parameters are available:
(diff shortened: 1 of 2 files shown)


Arksine:
Looks like a good addition, however I wonder if we could do something like this:
```diff
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -571,9 +571,12 @@ class BedMeshCalibrate:
         return adj_pts
     cmd_BED_MESH_CALIBRATE_help = "Perform Mesh Bed Leveling"
     def cmd_BED_MESH_CALIBRATE(self, gcmd):
+        profile_name = gcmd.get('PROFILE', "default")
         self.bedmesh.set_mesh(None)
         self.update_config(gcmd)
         self.probe_helper.start_probe(gcmd)
+        gcmd.respond_info("Mesh Bed Leveling Complete")
+        self.bedmesh.save_profile(profile_name)
     def probe_finalize(self, offsets, positions):
         x_offset, y_offset, z_offset = offsets
         positions = [[round(p[0], 2), round(p[1], 2), p[2]]
@@ -690,8 +693,6 @@ class BedMeshCalibrate:
         except BedMeshError as e:
             raise self.gcode.error(str(e))
         self.bedmesh.set_mesh(z_mesh)
-        self.gcode.respond_info("Mesh Bed Leveling Complete")
-        self.bedmesh.save_profile("default")
     def _dump_points(self, probed_pts, corrected_pts, offsets):
         # logs generated points with offset applied, points received
         # from the finalize callback, and the list of corrected points
```

Minor modification, however it eliminates the need for the profile name to be a class attribute.

randellhodges:
I didn't see your edit until after I updated the code per your original suggestion. The comment about using the class attribute, does that mean the use of a local variable won't work?

If so, maybe since probe_helper.start_probe is passed the gcmd object it could pass it back to the finalize method so it could be accessed there?

`def probe_finalize(self, offsets, positions, gcmd):`

Or is that too large of a change for something this simple and the use of the class attribute is the better choice?

Arksine:
Sorry, that was my mistake.  The issue is that during manual calibration the call to `start_probe` won't block until calibration is complete, so we would save the profile prematurely.  I simply thought that we could get away with doing everything in the gcode handler without thinking through all scenarios. 

Using the class attribute, as you did in the original implementation, is likely the best option in this scenario.  Changing `probe_finalize` to include the gcode command would require modification to several modules.

randellhodges:
No problem. I'll change it back, and include the suggestion about initing it as well. It'll be a little bit before I get to it.

randellhodges:
Changes made

KevinOConnor:
Thanks.  @Arksine are the changes okay with you?

-Kevin

Arksine:
Looks good to me.

KevinOConnor:
Thanks.

-Kevin

PR #4525, opened 2021-07-26
neopixel: increase BIT_MAX_TIME to increase compatibility on AVR

ayufan:
It appears that LED strips are more permissive about bit time.
This makes use of that to actually make it possible to support
LED strips on AVR. Without that, the bit time is too short
due to hardware irq (on timer) being fired and return early
from neopixel update.

In my case. I connected LED strip to PJ3 (the J19 on Einsy Rambo 1.1a of Prusa MK3S+ https://github.com/ultimachine/Einsy-Rambo/blob/1.1a/board/Project%20Outputs/Schematic%20Prints_Einsy%20Rambo_1.1a.PDF).

### DIFF
--- a/klippy/extras/neopixel.py
+++ b/klippy/extras/neopixel.py
@@ -7,7 +7,7 @@ import logging
 
 BACKGROUND_PRIORITY_CLOCK = 0x7fffffff00000000
 
-BIT_MAX_TIME=.000004
+BIT_MAX_TIME=.000030
 RESET_MIN_TIME=.000050
 
 MAX_MCU_SIZE = 500  # Sanity check on LED chain length


KevinOConnor:
Alas, the neopixel hardware is "all over the place".  There were reports that some older ws2811 chips would reset after just a few microseconds.  I added the BIT_MAX_TIME and RESET_MIN_TIME settings to make this more flexible.  However, I think we'd need some wider testing before changing this.  (If we make the code change here and users have older neopixel chains, then they'll lose the current retransmit capability.)

FWIW, when I tested the neopixel led on my adafruit feather board (the only neopixel I own), I found it tolerated up to ~38us.

-Kevin

ayufan:
@KevinOConnor This is very valid point.

Actually, it is all over the place, but we can likely categorise that by variants. Others doing neopixel are actually allow you to choose variant, each with different timings.


One design that we could go with is to add a `variant =` to `[neopixel ...]` with a different timings.
Default being `800kbps`. So, far esphome.io did work for me for all my different strips.

Ref.:
- https://esphome.io/components/light/neopixelbus.html
- https://github.com/Makuna/NeoPixelBus/blob/master/src/internal/NeoEsp8266UartMethod.h#L280

KevinOConnor:
Yes, it may make sense to have some kind of "variant" option.  However, it would be preferable to see if we can get some feedback from neopixel users first.  It may be the case that the old neopixels are too rare to worry about.  In general, I prefer not to add config options for esoteric settings - in particular config options that normal users wont know how to set.

-Kevin

ayufan:
@KevinOConnor I will get back to it in 2 wks time. Currently, I'm traveling.

4rianton:
I have a 256 LED Neopixel panel that I am using and it works but only with changing the BIT MAX to 0.000040. Otherwise, the LEDs are too unreliable. Hoping that this makes it into the main branch.

cpasoft:
So do I... I have a 60 LED Neopixel panel and I must to set BIT_MAX to 0.000030. Now always have the "dirty" signal under update manager or invalid under klipperscreen... Is there any workaround to this other than make this into the main branch??

PR #4527, opened 2021-07-27
Update generic-mks-robin-nano-v2.cfg

mks-viva:
Update generic-mks-robin-nano-v2.cfg EXP2_4 and EXP2_6  pin

Signed-off-by: Qiuyanhua 1224833100@qq.com


KevinOConnor:
Thanks.  The change looks fine to me.  Is "Qiuyanhua" your full real name?  https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

PR #4528, opened 2021-07-27
Add some missing documentation for attributes on [extruder].

alexsomesan:
Was just going through the docs to setup Klipper for the first time and I noticed these attributes missing from the reference for the `[extruder]` section. They are mentioned in several example configuration and they also work as expected when configured in my use-cases.


alexsomesan:
Closing to redo formatting.

PR #4529, opened 2021-07-27
docs: missing extruder attributes docs

alexsomesan:
Add some missing documentation for attributes on [extruder].

Signed-off-by: Alex Somesan <alex.somesan@gmail.com>


KevinOConnor:
Thanks.  I don't think we need to copy the stepper field description into the extruder config section.  However, I did add a comment on the two available stepper fields in commit 4edab5b3.

-Kevin

Issue #4538, opened 2021-07-28
Documentation anchors not working on Safari, macOS and iOS


blalor:
After the (fantastic) site update #4512, anchors aren't working correctly when loading the page on Safari on iOS (14.7) and macOS (11.4, Safari 14.1.1).  For example, if I open https://www.klipper3d.org/Config_Reference.html#heater_bed in a new tab, the page doesn't scroll to the anchor, but if I click it from the nav on the right, it does.  Firefox 89.0.2 does not show this behavior.  This means that linking between sections on different pages doesn't work.

damlobster:
Thanks @blalor for spotting this, I'll look into how to solve it.

KevinOConnor:
Thanks.  Hopefully this is resolved with the merge of #4539 .

-Kevin

blalor:
It is!  Thank you!

PR #4539, opened 2021-07-28
klipper3d : anchor links not working on Safari

damlobster:
Solves issue #4538 by disabling the `navigation.instant` feature of the mkdocs-material theme which does not work well with Safari.


KevinOConnor:
Thanks.

-Kevin

PR #4540, opened 2021-07-28
filament_motion_sensor: Add time-based runout calculation

dragorn:
Add optional time-based runout calculation to filament motion sensors.
This is an alternate calculation method for detecting runout and jams
and addresses problems that can occur with some motion sensors like the
BTT sensor.

Timeout mode is disabled by default.

Signed-off-by: Mike Kershaw <dragorn@kismetwireless.net>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3677,6 +3677,16 @@ detection_length: 7.0
 #   The minimum length of filament pulled through the sensor to trigger
 #   a state change on the switch_pin
 #   Default is 7 mm.
+# time_based: False
+#   When set to true, use a time-based calculation to trigger runout
+#   events instead of distance travelled.  This is less precise than
+#   length measurement, but addresses issues with some movement sensors.
+# timeout_seconds: 30
+#   When time_based is set to True, the number of seconds of no
+#   filament movement before a runout is triggered.
+#   This should be adjusted based on the length of your filament path,
+#   tolerance of jams, etc.
+#   Default is 30 seconds.
 extruder:
 #   The name of the extruder section this sensor is associated with.
 #   This parameter must be provided.
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3687,6 +3687,11 @@ detection_length: 7.0
 #   This should be adjusted based on the length of your filament path,
 #   tolerance of jams, etc.
 #   Default is 30 seconds.
+# arm_seconds: 120
+#   The amount of time from the start of printing before the filament
+#   sensor is used.  For sensors which incorrectly trigger during the
+#   homing and heatup sequence.
+#   Default is 0 seconds, immediately armed.
 extruder:
 #   The name of the extruder section this sensor is associated with.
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks, however I'm not sure I understand what this does differently and what the intended audience is.  How would a user set these new config options?  Is there a way to tweak the existing config (and/or existing code) so that the hardware works reliably without adding new config options?

@TheJoshW - do you have any comments?

-Kevin

dragorn:
> Thanks, however I'm not sure I understand what this does differently and what the intended audience is.

This treats a pulse-based motion sensor as a timer rather than a distance sensor.

The intended audience is anyone with a pulse-style (instead of presence-style single switch) sensor which does not accurately report distance.

After inquiring on the discord server if anyone has successfully gotten the BTT "smart sensor" to work with the length estimation, I got several replies that it wasn't just me.

> How would a user set these new config options?

Not sure I understand - the same way they set any other options? The filament motion sensor can either be monitored via estimated distance, which doesn't seem to work with some sensors like the BTT smartsensor, or by time, which has worked now for weeks with my tests.

> Is there a way to tweak the existing config (and/or existing code) so that the hardware works reliably without adding new config options?

It would appear the BTT smartsensors aren't reliable in their distance reporting, since it constantly flags an underrun/jam.

This alternate movement indicator offers a way to make them function.

zellneralex:
@dragorn for me the BTT sensor is working as is today. What I needed to do was switch it off while doing load or unload.

dragorn:
> ***@***.***(https://github.com/dragorn) for me the BTT sensor is working as is today. What I needed to do was switch it off while doing load or unload.

I've tested 3 on 2 different printers; none of them consistently work in length mode. The faults happen anywhere from 5 minutes to multiple hours into printing, unrelated to load/unload cycles.

TheJoshW:
Hi  @dragorn , are you able to describe the exact issue you are having?  I am also trying to track down a random runout issue, however mine only occurs on the second print, never on the first (no matter how long the first print is).  I have been avoiding it by restarting everything between prints until I can figure it out.  I have made a change (attached) based on Kevin's original query here: https://github.com/KevinOConnor/klipper/pull/3857#issuecomment-799033040.  I also updated get_status() to output additional useful info.  I haven't seen any issues since these changes, however I haven't been printing much lately, so if you have a chance could you try running it for me and let me know the status values if you do get a false trigger.
Thanks.
[filament_motion_sensor.py.txt](https://github.com/KevinOConnor/klipper/files/7024733/filament_motion_sensor.py.txt)

tjengbudi:
@TheJoshW i already open it as issued but the bot force to closed it

https://github.com/KevinOConnor/klipper/issues/4554

the second printing is not stable at all. 

however after first detect, the sensor will set to error state, until you move the sensor manually (with changing filament if used btt) if not after resume it will not detect the jam anymore. 
i think it is a bug. if it resume the state of the sensor will be forced to working state. so if it is not replace the filament it will said it error again and paused the printing.

PR #4543, opened 2021-07-29
Update printer-alfawise-u30-2018.cfg

Schlunze:
module: printer-alfawise-u30-2018.cfg,

added BLtouch and Filament sensor.

Signed-off-by: Schlunze <ts_reload@gmx.net>

### DIFF
--- a/config/printer-alfawise-u30-2018.cfg
+++ b/config/printer-alfawise-u30-2018.cfg
@@ -40,8 +40,9 @@ dir_pin: !PB9
 enable_pin: !PE1
 microsteps: 16
 rotation_distance: 8
-endstop_pin: !PE6
-position_endstop: 0.0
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks, but I'm not sure I understand this change.  Did the alfawise u30 come standard with a filament runout sensor and come standard with a bltouch?  Otherwise, see the guide on example configs at https://www.klipper3d.org/Example_Configs.html .

Separately, commits need to have a signed-off-by line containing your real full name - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

Schlunze:
Yes alfawise u30 has standard filament runout sensor. 
BL touch is optional and commented out. Just for the case tested pin names for stm32 are directly in the sample config.

Keep up the good work.
Greetings Schnabulator

PR #4553, opened 2021-08-02
Probe: bugfix to z_offset_apply_probe

shiftingtech:
command was accessing the wrong variable for the existing z offset,
leading to bad behavior.
--signed-off-by: Ben Eastep <shifting@shifting.ca>


KevinOConnor:
Thanks.

-Kevin

PR #4555, opened 2021-08-03
Data logging and motion analysis

KevinOConnor:
This PR is a follow up to PR #4504 - it extends the motion_report.py module to support dumping Klipper's internal motion queues over the Klipper API Server.  This PR also adds a new `data_logger.py` tool for capturing and compressing that data, along with a new `motan_graph.py` tool for performing motion analysis on the captured data.

There are some brief usage docs on the new system in the docs/Debugging.md file on this PR - https://github.com/KevinOConnor/klipper/blob/work-motan-20210801/docs/Debugging.md#motion-analysis-and-data-logging .

The support here was inspired by the `scripts/graph_motion.py` (and similar) tool.  The main advantage of this system is that it can analyze printer data from real prints.  The `data_logger.py` tool is intended to be flexible enough that it could run for hours and capture large amounts of data while a test system is in regular use.

I'm still working on this PR.  There are some areas of the analysis software that I think could be improved.  (And, there are likely many bugs.)

As an example of the functionality, I generated some graphs of test data.  Here's a graph of movement with an input_shaper change - it's from the following command sequence:
```
G1 X50 F6000
SET_INPUT_SHAPER SHAPER_FREQ_X=49.4 SHAPER_TYPE_X=ei
G1 X150
G1 X50
```
input_shaper
The blue line is from Klipper's generated queue_step commands.  The graph shows that the queue_step commands don't deviate significantly from the command position prior to input_shaper being enabled, but then does deviate noticeably on the move after input_shaper is enabled.

Here's a zoom in of the second 180 degree turn with input_shaper enabled, along with a calculated velocity from the queue_step commands:
input_shaper_zoom
Again, this clearly shows how input_shaper alters the stepper velocity.

Here's a graph showing the effects of Pressure Advance - taken from this gcode snippet:
```
G1 X150 E4.989020
SET_PRESSURE_ADVANCE ADVANCE=0.30
G92 E0
G1 X50 E4.989020
```
pa
And here's a zoom-in of that first non-PA to PA transition, along with the calculated velocity from queue_steps:
pa_zoom

Going forward, I plan to extend this functionality to the adxl345.py module.  I also plan to add similar support for hall magnetic angle sensors.  The goal is to be able to analyze these sensors side-by-side with the Klipper generated motion commands.

@dmbutyugin - FYI.

-Kevin

### DIFF
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -278,6 +278,49 @@ window" interface. Parsing content from the G-Code terminal output is
 discouraged. Use the "objects/subscribe" endpoint to obtain updates on
 Klipper's state.
 
+### motion_report/dump_stepper
+
+This endpoint is used to subscribe to Klipper's internal stepper
+queue_step command stream for a stepper. Obtaining these low-level
+motion updates may be useful for diagnostic and debugging
+purposes. Using this endpoint may increase Klipper's system load.
+
+A request may look like:
+`{"id": 123, "method":"motion_report/dump_stepper",
+"params": {"name": "stepper_x", "response_template": {}}}`
+and might return:
+`{"id": 123, "result": {"header": ["interval", "count", "add"]}}`
+and might later produce asynchronous messages such as:
+`{"params": {"first_clock": 179601081, "first_time": 8.98,
+"first_position": 0, "last_clock": 219686097, "last_time": 10.984,
+"data": [[179601081, 1, 0], [29573, 2, -8685], [16230, 4, -1525],
+[10559, 6, -160], [10000, 976, 0], [10000, 1000, 0], [10000, 1000, 0],
+[10000, 1000, 0], [9855, 5, 187], [11632, 4, 1534], [20756, 2, 9442]]}}`
+
+The "header" field in the initial query response is used to describe
+the fields found in later "data" responses.
+
+### motion_report/dump_trapq
+
+This endpoint is used to subscribe to Klipper's internal "trapezoid
+motion queue". Obtaining these low-level motion updates may be useful
+for diagnostic and debugging purposes. Using this endpoint may
+increase Klipper's system load.
+
+A request may look like:
+`{"id": 123, "method": "motion_report/dump_trapq", "params":
+{"name": "toolhead", "response_template":{}}}`
+and might return:
+`{"id": 1, "result": {"header": ["time", "duration",
+"start_velocity", "acceleration", "start_position", "direction"]}}`
+and might later produce asynchronous messages such as:
+`{"params": {"data": [[4.05, 1.0, 0.0, 0.0, [300.0, 0.0, 0.0],
+[0.0, 0.0, 0.0]], [5.054, 0.001, 0.0, 3000.0, [300.0, 0.0, 0.0],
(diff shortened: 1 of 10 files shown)


dmbutyugin:
Thanks Kevin for letting me know! This looks very interesting. I wonder what level of precision we could achieve for adxl345? Will the data be coming from the trapq queue? Or from the generated stepper steps?

BTW, some time ago I wrote a prototype for `adxl345_simulated`, which I recently updated [adxl345_simulated.py](https://github.com/dmbutyugin/klipper/blob/resonance-test-methods/klippy/extras/adxl345_simulated.py). It is possible to configure in the config as a 'normal' ADXL345 sensor
```
[adxl345_simulated]

[resonance_tester]
accel_chip: adxl345_simulated
...
```
And then use it in the simulation runs via the usual commands `ACCELEROMETER_MEASURE` or `TEST_RESONANCES`.

But the real application I am trying to use it now is to design a different resonance test. Then one could use it as
```
real_accelerometer.start_measurements()
simulated_accelerometer.start_measurements()
# Run some moves
real_data = process_accelerometer_data(real_accelerometer.finish_measurements())
simulated_data = process_accelerometer_data(simulated_accelerometer.finish_measurements())
real_data.subtract(simulated_data)  # Now we only have resonances here
```
I think I am getting some promising results here, at least on my printer.

So, Kevin, do you think that such a usecase can be covered by what you propose? Or would it be better to have a dedicated `adxl345_simulated` module that intercepts the moves and generates the output on its own?

KevinOConnor:
Thanks for reviewing.

> I wonder what level of precision we could achieve for adxl345?

That's one of the questions I want to answer with this new infrastructure.  I think with some careful coding it should be possible to time each adxl345 measurement to within a microsecond or so.  The queue_step stepper commands should also have similar precision.  It's an open question (for me at least) how noisy the adxl345 is on each measurement.

I suspect getting really good accuracy will require calibration software - a mechanism to determine the angle between printer axes and adxl345 axes, determine the gravity vector, and possibly a magnitude correction.

> Will the data be coming from the trapq queue? Or from the generated stepper steps?

My current plan is to give the adxl345 its own API Server "endpoint" with its own data messages.  Basically, an array containing a list of (time, measured_acceleration) pairs.

If you're curious, I have some working (but quite raw) code for hall magnetic angle sensors.  I've uploaded it to the work-angle-20210722 branch.  The high-level flow is similar to the existing adxl345 data flow - except instead of writing all the data to a CSV file, the data is incrementally written via JSON messages on the API Server.  The key code on that branch is in `klippy/extras/angle.py`.

> But the real application I am trying to use it now is to design a different resonance test.

Interesting!  I was thinking something similar - instead of analyzing the raw accelerometer data, analyze the difference between the expected and measured acceleration.

If I understand your code correctly, you've hooked into the movement dispatch code path to get access to the commanded acceleration.  The code on this branch takes a slightly different approach - it keeps the trapq history around for later analysis.

> So, Kevin, do you think that such a usecase can be covered by what you propose?

Yes, I think so.  My current code is focused on off-line analysis though.

That said, If doing analysis within Klipper, I think using the trapq history may be preferable to hooking into the main movement code flow.

Cheers,
-Kevin

PR #4557, opened 2021-08-04
Config for Anycubic 4MaxPro_2.0-2021.cfg

int-0x03:
New config for printer.

This config I created and fully tested on my own printer Anycubic 4Max Pro 2.0. This works fine! Please note that this printer is different from the Anycubic 4MAX.

Signed-off-by: Alexander Usachev <int_0x21@protonmail.com>

### DIFF
--- a/config/printer-anycubic-4MaxPro_2.0-2021.cfg
+++ b/config/printer-anycubic-4MaxPro_2.0-2021.cfg
@@ -1,0 +1,179 @@
+# Klipper firmware config file for Anycubic 4Max Pro 2.0. To use this config,
+# the firmware should be compiled for the AVR atmega2560.
+
+# You can build and flash firmware with commands in klipper directory:
+# $ make clean
+# $ make menuconfig
+# $ make
+# $ make flash FLASH_DEVICE={your serial port}
+
+# For more precision, don't forget calibrate PIDs, extruder step_distance, and axises step_distance (if needed) individually for your printer.
+
+# For Anycubic 4Max Pro (not 2.0) owners:
+#       Be careful when using this config! This config tested only on Anycubic 4Max Pro 2.0 with klipper v0.9.1-667-g31761500! At first, you should set homing_speed on 5,
+#       and run homing and click on the endstops with your fingers. It is necessary to make sure that all the motors are spinning in the right direction,
+#       all the temperature sensors show the correct values, and all the endstops are triggered correctly.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[stepper_x]
+step_pin: PF0
+dir_pin: PF1
+enable_pin: !PD7
+step_distance: 0.01
+endstop_pin: ^!PE5
+position_min: -8
+position_endstop: -8
+position_max: 270
+homing_speed: 50.0
+homing_retract_dist: 5.0
+second_homing_speed: 5.0
+
+[stepper_y]
+step_pin: PF6
+dir_pin: !PF7
+enable_pin: !PF2
+endstop_pin: ^!PD2
+position_max: 215
+position_min: -1
+position_endstop: -1
+homing_speed: 50.0
+homing_retract_dist: 5.0
+second_homing_speed: 5.0
+step_distance: 0.0125
+
+[stepper_z]
+step_pin: PL3
+dir_pin: PL1
+enable_pin: !PK0
+endstop_pin: ^!PD3
+position_endstop: 0.0
+position_max: 195
+homing_speed: 5.0
+homing_retract_dist: 3.0
+second_homing_speed: 0.8
+homing_speed: 6.0
+step_distance: 0.00125
+
+[extruder]
+step_pin: PA4
+dir_pin: !PA6
+enable_pin: !PA2
+step_distance: 0.00234
+nozzle_diameter: 0.400
+filament_diameter: 1.750
+instantaneous_corner_velocity: 1.5
+max_extrude_only_distance: 500
+max_extrude_only_velocity: 50
+max_extrude_only_accel: 3000
+heater_pin: PB4
+sensor_type: ATC Semitec 104GT-2
+sensor_pin: PK5
+smooth_time: 2
+control: pid
+pid_kp: 29.735                            # calibrate these PID values for your printer individually!
+pid_ki: 1.304                             # calibrate these PID values for your printer individually!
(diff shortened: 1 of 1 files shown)


WhatIsMyOldUserName:
It's missing the display info?

int-0x03:
@WhatIsMyOldUserName , yes dude. Because in this printer used proprietary display with own MCU. Such one, reading files from SD card and send bytes to the trigorilla (main board) through UART interface. In "normal" printers (e.g. Ender 3, Prusa i3 etc, with text lcd), screen doesn't have own MCU and controlled by main board, but in this case we have opposite situation, touchscreen+display with MCU control printer's mainboard through UART. This logical circuit between main board and display is bullshit, but vendors race on "pretty and user friendly interfaces".

When we install Klipper, we can't use built-in display, unfortunately. But you can replace this screen with another one, like a LCD 12864 and connect it with mainboard. Next step edit Klipper config for your printer.

KevinOConnor:
Thanks.  However, the config would need to follow the config guidelines at https://www.klipper3d.org/Example_Configs.html (no deprecated features, leave values at defaults where possible, no host settings, etc.).

-Kevin

int-0x03:
@KevinOConnor, ok. I fixed this. That's all? May I skip something, sry.

KevinOConnor:
Thanks.  In general the config looks fine to me.  I did notice a few minor things:
1. A fan `kick_start_time` of 1 second is known to cause havoc.  Also, it doesn't make sense to use a kick_start_time with heater_fan nor controller_fan sections.  Can you confirm that this setting is necessary for the `fan` section (the default of 100ms does not work)?
2. The `instantaneous_corner_velocity` setting looks high to me.  Unless you found it was necessary in testing, it would be preferable to leave that at the default.
3. It would be preferable to define a `gear_ratio` for the extruder config (by counting the gears in the extruder mechanism) as it is generally preferable to use a rotation_distance that matches the circumference of the hobbed bolt that engages the filament.

-Kevin

KevinOConnor:
Ah, if it's a BMG extruder then one would use:
```
gear_ratio: 50:17
rotation_distance: 22.024
```
(7.488 * 50 / 17 == 22.024)

-Kevin

KevinOConnor:
Thanks.

-Kevin

int-0x03:
> Thanks.
> 
> 
> 
> -Kevin

Hi. Sorry, I am too late. But I will test gear ratio today on my printer, and write you.

int-0x03:
I checked this gear ratio, and this works fine! How I can upload new version of config, with fixes accordingly guidelines?

KevinOConnor:
Create a new PR with the change.  Typically this would be creating a branch on your repo, committing the change to that branch, and then making a pull request of that branch.

-Kevin

Piazzi68:
Hi, sorry for the inconvenience, I have installed klipper on 4MaxPro V2.0, is it possible to use the automatic shutdown at the end of printing? what commands do I have to give to do it? thanks if you can help me. also in mail piazzi1968@gmail.com

JonnyHaystack:
I notice this still has `kick_start_time = 1.0` - is that okay?

> Thanks. In general the config looks fine to me. I did notice a few minor things:
> 
>     1. A fan `kick_start_time` of 1 second is known to cause havoc.  Also, it doesn't make sense to use a kick_start_time with heater_fan nor controller_fan sections.  Can you confirm that this setting is necessary for the `fan` section (the default of 100ms does not work)?
> 
>     2. The `instantaneous_corner_velocity` setting looks high to me.  Unless you found it was necessary in testing, it would be preferable to leave that at the default.
> 
>     3. It would be preferable to define a `gear_ratio` for the extruder config (by counting the gears in the extruder mechanism) as it is generally preferable to use a rotation_distance that matches the circumference of the hobbed bolt that engages the filament.
> 
> 
> -Kevin



> Thanks. In general the config looks fine to me. I did notice a few minor things:
> 
>     1. A fan `kick_start_time` of 1 second is known to cause havoc.  Also, it doesn't make sense to use a kick_start_time with heater_fan nor controller_fan sections.  Can you confirm that this setting is necessary for the `fan` section (the default of 100ms does not work)?
> 
>     2. The `instantaneous_corner_velocity` setting looks high to me.  Unless you found it was necessary in testing, it would be preferable to leave that at the default.
> 
>     3. It would be preferable to define a `gear_ratio` for the extruder config (by counting the gears in the extruder mechanism) as it is generally preferable to use a rotation_distance that matches the circumference of the hobbed bolt that engages the filament.
> 
> 
> -Kevin

These first two questions were never answered as far as I can see, so I'm not sure if I should change those before trying to use this config.

JonnyHaystack:
I notice this still has `kick_start_time = 1.0` - is that okay?

> Thanks. In general the config looks fine to me. I did notice a few minor things:
> 
>     1. A fan `kick_start_time` of 1 second is known to cause havoc.  Also, it doesn't make sense to use a kick_start_time with heater_fan nor controller_fan sections.  Can you confirm that this setting is necessary for the `fan` section (the default of 100ms does not work)?
> 
>     2. The `instantaneous_corner_velocity` setting looks high to me.  Unless you found it was necessary in testing, it would be preferable to leave that at the default.
> 
>     3. It would be preferable to define a `gear_ratio` for the extruder config (by counting the gears in the extruder mechanism) as it is generally preferable to use a rotation_distance that matches the circumference of the hobbed bolt that engages the filament.
> 
> 
> -Kevin



> Thanks. In general the config looks fine to me. I did notice a few minor things:
> 
>     1. A fan `kick_start_time` of 1 second is known to cause havoc.  Also, it doesn't make sense to use a kick_start_time with heater_fan nor controller_fan sections.  Can you confirm that this setting is necessary for the `fan` section (the default of 100ms does not work)?
> 
>     2. The `instantaneous_corner_velocity` setting looks high to me.  Unless you found it was necessary in testing, it would be preferable to leave that at the default.
> 
>     3. It would be preferable to define a `gear_ratio` for the extruder config (by counting the gears in the extruder mechanism) as it is generally preferable to use a rotation_distance that matches the circumference of the hobbed bolt that engages the filament.
> 
> 
> -Kevin

These first two questions were never answered as far as I can see, so I'm not sure if I should change those before trying to use this config.

int-0x03:
> 4MaxPro

https://github.com/Klipper3d/klipper/pull/5629

Hello man, check this PR.

PR #4564, opened 2021-08-06
docs: Fix reference to neopixel color_data

mattheath:
The example given for accessing the existing color values of a neopixel chain in the [status reference documentation](https://www.klipper3d.org/Status_Reference.html#neopixel-dotstar) is missing an underscore, so it doesn't work as expected (which confused me while trying to read the values from some neopixels). This change adds it in so the example works!

Hopefully I've submitted this change correctly — please let me know if there's anything I need to update 😊


KevinOConnor:
Thanks.

-Kevin

PR #4566, opened 2021-08-07
Add mixing extruder support

matyay:
This PR adds support for color mixing extruders. It is a continuation of https://github.com/KevinOConnor/klipper/pull/3920 with fixed bugs and added features.

It especially adds automatic retraction processing for mixing extruders as it requires special handling to be done correctly.

### DIFF
--- a/config/printer-geeetech-301.cfg
+++ b/config/printer-geeetech-301.cfg
@@ -1,0 +1,158 @@
+# This file contains common pin mappings for the GTM32 PRO board in the Geeetech 301 printer
+# To use this config, the firmware should be compiled for the
+# STM32F103 with "No bootloader" and with "Use USB for communication"
+# disabled.
+
+# The "make flash" command does not work on the GTM32 PRO. Instead,
+# after running "make", run the following command to flash the board:
+#  stm32flash -w out/klipper.bin -v -i rts,-dtr,dtr /dev/ttyUSB0
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[board_pins gtm32_pro_vb]
+aliases_probes: X_MIN=PE5,X_MAX=PE4,Y_MIN=PE3,Y_MAX=PE2,Z_MIN=PE1,Z_MAX=PE0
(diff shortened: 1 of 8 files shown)


binarylefmart:
Hi @matyay, i tested the Branch but when i put the sample macro on my printer.cfg, this kill my mcu communication, i the nothing on klippy.log.
Have an idea ?

matyay:
@binarylefmart Hmm, Can you attach your config and `klippy.log`? When does the communication with MCU breaks? right from the beginning or when you issue some commands?

binarylefmart:
https://paste.ofcode.org/DkdYWRHdGPKkvVExP2xxix

here is the file,it's weird it is like klipper lost the serial or have wrong script.

binarylefmart:
always have the same issue, and have one too when i enable the extended g1 line.

binarylefmart:
PrtScr capture
PrtScr capture_2
PrtScr capture_3

matyay:
@binarylefmart Yeay, try removing the `extended_g1` option completely. It shouldn't be there.

binarylefmart:
When i remove it like i said, have the mcu issue

KevinOConnor:
Thanks.  My high-level feedback at this time is that we need to implement some architectural changes to Klipper's treatment of extruder stepper motors before we can add enhancements in this area.  Currently the klipper `extruder` config section defines the parameters to a hotend, a heater, a temperature sensor, and an extruder stepper motor.  I think we need a more clear abstraction so that it is possible to better control "extruder stepper motors".  That work should improve future mixing extruder support as it would not require the user to define multiple dummy "extruder" sections.

-Kevin

KevinOConnor:
FYI, this architectural work also effects #4489, #4082, and #3920.

-Kevin

KevinOConnor:

-Kevin

matyay:
Hi, I'm willing to continue the work on this feature but I'm short on time right now. Besides that I need to do some work on my 3D printer which has the mixing hotend, otherwise I won't be able to test the code.

ETE-Design:
@matyay Good to hear that your back, and can work on again mabye in a not so far away future... Really would like to have this feature added to Klipper

KevinOConnor:
FYI, #5143 makes it possible to call SET_EXTRUDER_STEP_DISTANCE on extruder_stepper objects, which may make it possible to implement mixing extruders via extruder_stepper objects.

-Kevin

PR #4569, opened 2021-08-08
output_pin: fix time difference calculation

Cirromulus:
This fixes the time_difference calculation for rescheduling a resend-timer.
Previously, any `output_pin` with activated `maximum_mcu_duration` would trigger two set_pin messages.


KevinOConnor:
Thanks.

-Kevin

PR #4575, opened 2021-08-10
config: Adds generic config for BigTreeTech E3 RRF 1.1

pedrolamas:
Mostly based on existing configurations and official [pin docs](https://github.com/bigtreetech/BTT-E3-RRF/blob/master/BTT%20E3%20RRF%20V1.1/Hardware/BTT%20E3%20RRF%20V1.1-PIN.pdf), I've tested this in my own printer


pedrolamas:
Made a small adjustment on the documentation to remove all "SKR" mentions as this board is actually called "BigTreeTech E3 RRF 1.1".

KevinOConnor:
Thanks.

-Kevin

PR #4578, opened 2021-08-11
Endstop linking

dalegaard:
This adds the ability to, for motors on a rail, specifying another motor to use as endstop, effectively 'linking' the motor to the endstop.

This is useful for printers with multiple motors per rail, e.g. if a machine has two belts for one axis, and two motors for each belt. Using linking, it's possible to, in software, let the motors that share belts, also share their endstop pin.

We've been testing this for some time on the Annex Engineering Tycho printer, which has 8 motors on the XY gantry in just such a configuration(4 belts, 2 motors per belt).

To allow the general case to work, where endstops from both higher and lower indexed rail steppers can be used on any other rail stepper, a two-pass approach is used. All steppers are added, and steppers with endstop pointers are then linked to the correct endstops.

### DIFF
--- a/klippy/stepper.py
+++ b/klippy/stepper.py
@@ -264,6 +264,7 @@ class PrinterRail:
         self.stepper_units_in_radians = units_in_radians
         self.steppers = []
         self.endstops = []
+        self.endstop_map = {}
         self.add_extra_stepper(config)
         mcu_stepper = self.steppers[0]
         self.get_name = mcu_stepper.get_name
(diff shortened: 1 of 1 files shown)


Annex-Engineering:
just wanted to throw my 2c in here

Ive been running this on my personal machine for atleast a month with zero issues to speak of. Hopefully it can get merged into base line klipper, as I can see lots of teams wanting to try this setup in the future. (especially on a z drive with no gearbox)

KevinOConnor:
Thanks, and sorry for the delay in responding.  Alas, I'm missing the "big picture" here - I don't understand what this change is intended to do.  How does it differ from the existing multi-stepper support?

-Kevin

Annex-Engineering:
The current implementation requires one endstop per stepper. 

If your running 2 steppers on the same belt, it is advantageous to only use one endstop to control these to keep even tension on the belt on both sides. Even with step locking, there is still a high chance of this differential, especially with sensorless homing.

This solution just stops both motors at the same time, keeping them as close to in sync as possible.

Youll see in the photo below: 
 - the X axis is the Yellow and Red Lines
 - the Y axis is the Blue and Green Lines

If you wanted to home these sides independently, such that they self square, it would be impossible without this modification to the code
unknown

KevinOConnor:
Erhh, you can use the same endstop for all steppers on a "shared rail".  If an endstop is not specified for stepper_x1 it will use the endstop from stepper_x.

Or, are you saying you want stepper_x and stepper_x1 to share an endstop and stepper_x2 and stepper_x3 to share a second endstop?

-Kevin

Annex-Engineering:
X shares with X1, X2 shares with X3.

Tycho - 8 Motor XY Homing 
https://youtu.be/HMGsMNoHM4o

KevinOConnor:
Okay, I think I understand.

Instead of introducing a new `endstop_of` parameter, could the PrinterRail class detect a duplicate `endstop_pin` declaration and then call the existing `endstop.add_stepper()` method?  That is, could we avoid introducing `endstop_of` by detecting duplicate `endstop_pin`?

-Kevin

dalegaard:
Hi Kevin,

Sure, we could definitely do that instead. I'll get on that :ok_hand:

dalegaard:
I've updated the code to detect duplicated `endstop_pin` settings.

I decided to add a dict that simply maps endstop_pin -> MCU_endstop, but retain the `endstops` list. I didn't want to change the `get_endstops` interface. If you prefer that I re-write this to keep a single `endstops` list that also includes the `endstop_pin` in the tuple, and then map back to the 2-tuple in `get_endstops`, let me know and I'll change this. Personally I think this is clearer, but it's your code base of course :-)

KevinOConnor:
Thanks.  I agree with your approach.

One thing I noticed, is that I think it would be helpful if the code checked that there isn't a conflict with the pullup/invert settings if the same endstop pin is used twice.  Just as a sanity check.

-Kevin

dalegaard:
Hi Kevin,

I had completely forgotten about the details of pullups and inverts, so the code was subtly wrong as well. Good catch!

I've changed the code to use the original `endstop_pin` value when setting up the endstop(to ensure pullup and invert information is passed along correctly) and in the `endstop_map` I now save the pullup and invert information - this is then compared to the first instance of the same pin, and an error is raised if the flags set are not the same.

KevinOConnor:
Thanks.

-Kevin

PR #4579, opened 2021-08-11
Adding files to support ar100

eliasbakken:
module: ADD SUPPORT FOR AR100 CPU

This pull-request adds support for real-time control using the AR100 
CPU that is present in some Allwinner SoCs (A64, H3 etc). The amount 
of SRAM present for code and variables is limited, so some minor changes 
to the codebase is necessary. It should not affect other controllers. 

Signed-off-by: Elias Bakken <elias@iagent.no>


KevinOConnor:
Thanks.

Due to time constraints on my side I haven't been able to fully review this change.

In a quick review, it generally looks good to me, but I did notice a few minor items:
1. For the benchmarks, please provide the results of tests with the standard step delay (and use that number in Features.md).  The primary purpose of the benchmarks is for code evaluation and platform comparison - so we want to try to use the same tests on all the platforms.
2. The `pin_map` config option is deprecated - it was replaced with `[board_pins]` config sections.  So, there shouldn't be any changes to the aliases in the pins.py file.
3. I would prefer the copyright to be the same format for all the files (one line file description, one blank comment line, one line copyright statement, one blank comment line, one line license statement).
4. It looks like several files were added from an external library (eg, spr.h, macro.S, etc.).  It would be preferable to place these files in the lib/ directory and to update lib/README with the upstream source information (along with a description of any changes - ideally no changes).
5. We don't want to make changes to the src/generic/ files for a particular platform - instead if a change is needed, copy the relevant code to the src/ar100/ directory and make the modifications there.  For example, see how src/pru/main.c implements a custom `dynmem_start()` function.
6. I don't understand the TMC_UART changes.  In general, we want to use the same code on all the platforms.  The existing src/tmcuart.c code should run fine on any board that supports gpio pins.  If it doesn't work on this platform, then I think we need to discuss those changes in a PR separate from the ar100 support.

Thanks again,
-Kevin

Issue #4571, opened 2021-08-09
DELTA_CALIBRATE uses incorrect parameter name


DeanCording:
After doing a `DELTA_CALIBRATE` Klipper reports the results as:

    stepper_a: position_endstop: 404.870784 angle: 208.981638 arm: 250.659000

The reported arm length should have the label `arm_length` instead of `arm`

KevinOConnor:
Should be fixed - commit d7b04aa1.

-Kevin

PR #4580, opened 2021-08-11
Update delta.py

conlank:
Resolved issue with how arm length is reported to the user, as found in #4571.

Signed-off-by: Conlan Kreher <ConlanKreher@gmail.com>

### DIFF
--- a/klippy/kinematics/delta.py
+++ b/klippy/kinematics/delta.py
@@ -221,9 +221,9 @@ class DeltaCalibration:
                            "%.6f" % (self.endstops[i],))
         gcode = configfile.get_printer().lookup_object("gcode")
         gcode.respond_info(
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, if you'd like me to commit directly then the commit will need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

Otherwise, I can make the change using my signed-off-by line.

-Kevin

KevinOConnor:
Thanks - I made the change (commit d7b04aa1).

-Kevin

KevinOConnor:
Ah - I missed that you updated the comment with a signed-off-by line.  Oh well.

-Kevin

PR #4582, opened 2021-08-12
Create Creality-ender3-pro-4.2.7-with-bltouch

jasonehines:
I was unable to find a configuration for the particular upgrades my Ender 3 Pro has, this is a combination of several configurations and what I had to figure out on my own. Hopefully it will help someone else with the newer Ender 3 Pros.


KevinOConnor:
Thanks, but configs in the main Klipper repository need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html .

If you're interested in discussing and sharing example configs, the Klipper Discourse Server may be a better alternative - https://community.klipper3d.org/

-Kevin

Issue #4560, opened 2021-08-05
[FR] Alternative resonance testing methods with the ADXL345


dmbutyugin:
For some time, I have been looking into ways of improving the current resonance test offered by Klipper, which runs vibrations of different frequencies and measures the printer response. I had some ringing on the X axis of my Ender 3 with the direct drive extruder, and the input shapers that were recommended as a result of the resonance testing measurements didn't remove the ringing very effectively. 

Overall, the current approach is quite decent and robust, but it may have some shortcomings:

- It is ultimately limited by the stepper motor resolution (microstepping only helps so much) at the high end of the spectrum.
- It may exaggerate the actual impact of the higher-frequency resonances.
- It finds all the resonances of the frame and the motion system, including some that are potentially not excited during the normal printing.
- Puts a lot of stress on the printer.
- It is not possible to properly test the effect of the input shaping on the printer kinematics and resonances, so it is not possible to compare the different input shaper configurations with the accelerometer and choose one that is the best.

So, in this FR I want to discuss and invite to test a different resonance testing procedure. I don't think it is ready for PR, but I did get some promising results on my printer, so it is worth further exploration and testing. If you want to check it out, the code is available in the branch
https://github.com/dmbutyugin/klipper/tree/resonance-test-methods

In order to enable the new test method, replace `[resonance_tester]` section in the `printer.cfg` with the following one:
```
[resonance_tester]
accel_chip: ....  # or separate accel_chip_x and accel_chip_y, this is also supported
method: pulses
max_accel: 10000
probe_points: ... # The usual probing point(s)
```
There are, in principle, more parameters, but they have reasonable defaults (I think), so no need to change them. `max_accel = 10000` works well for Ender 3, I suppose it will work for most printers too. If the printer has a too heavy bed, maybe `max_accel` here will need to be reduced to something like 5000, until the motor does not skip steps.

There is one more thing that must to be configured for the acceleromer(s). Previously, the ADXL345 axes orientation did not really matter. With this test, it matters a bit. Namely, one axis of the accelerometer must align with the tested axis, and you must configure the `axes_map` parameter for the accelerometer ([config reference](https://github.com/KevinOConnor/klipper/blob/master/docs/Config_Reference.md#adxl345)). To give you an idea, I have mounted the ADXL345 on the hotend in this manner:
[...]

PR #4585, opened 2021-08-14
Update adxl345 to use new motion reporting system

KevinOConnor:
This PR updates the adxl345 code to support logging of accelerometer data via the new "motion report" system.  This is based on top of PR #4555 .

The main advantages of this support are:
1. The new data logging should scale to very large captures.  It should be possible to capture data from real prints and to build captures lasting many hours.
2. The data log includes other useful information on Klipper's motion (such as requested movement).

This change does require a recompile of the micro-controller.  New timing messages were added to the micro-controller code so that the host software can calculate accurate accelerometer timestamps.  Previously, the code would calculate timestamps at the end of a capture.  The new code queries the buffer/fifo state every 100ms and reconstructs timestamps from that information.

The "motion analysis" graph system (motan_graph.py) has also been extended to support drawing of the accelerometer information.  Here's an example graph:
accelgraph

At first glance, the acceleration readings on that graph looks like gibberish.  However, if we zoom in on the acceleration spikes we can see a regular pattern:
accelmagnetgraph
That pattern correlates with four full steps of the stepper motor (one phase transition).  If we further correlate those spikes with stepper magnetic angle sensor data (separate experimental support not on this branch) as shown in the bottom graph of the above picture, then we can see that the accelerometer pulses are aligning with pulses on the stepper motor.  So, I conclude from this data that the accelerometer is accurate and the steppers are themselves moving with uneven velocity.

I think this does show the potential for detailed timing analysis.  However, it also indicates to me that simple analysis of accelerometer data will not be easy due to complex interactions with other components in the system.

@dmbutyugin - FYI.

-Kevin

### DIFF
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -321,6 +321,26 @@ and might later produce asynchronous messages such as:
 The "header" field in the initial query response is used to describe
 the fields found in later "data" responses.
 
+### adxl345/dump_adxl345
+
+This endpoint is used to subscribe to ADXL345 accelerometer data.
+Obtaining these low-level motion updates may be useful for diagnostic
+and debugging purposes. Using this endpoint may increase Klipper's
+system load.
+
+A request may look like:
+`{"id": 123, "method":"adxl345/dump_adxl345",
+"params": {"sensor": "adxl345", "response_template": {}}}`
+and might return:
+`{"id": 123,"result":{"header":["time","x_acceleration","y_acceleration",
+"z_acceleration"]}}`
+and might later produce asynchronous messages such as:
+`{"params":{"overflows":0,"data":[[3292.432935,-535.44309,-1529.8374,9561.4],
+[3292.433256,-382.45935,-1606.32927,9561.48375]]}}`
+
+The "header" field in the initial query response is used to describe
+the fields found in later "data" responses.
+
 ### pause_resume/cancel
 
 This endpoint is similar to running the "PRINT_CANCEL" G-Code command.
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20210830: The adxl345 ACCELEROMETER_MEASURE command no longer supports
+a RATE parameter.  To alter the query rate, update the printer.cfg
+file and issue a RESTART command.
+
 20210821: Several config settings in `printer.configfile.settings`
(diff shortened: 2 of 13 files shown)


KevinOConnor:
I've been doing further work on this branch and am getting some interesting results with the accelerometer.

As before, the accel values themselves are showing very large swings.  However, I've found that integrating the results (going from accel to velocity) shows promising data:
adxl-graph-vel
The above graph shows the extreme acceleration spikes, but after integration the resulting velocity actually looks pretty good.  It's inline with the velocity derived from the magnetic angle sensors I have on this printer.  I did need to do some tricks during the integration though to account for accelerometer offsets and measurement drift.  (Code is on the work-angle-20210722 branch - commit a1446c46).

It's also possible to take the integral again to obtain a position:
adxl-graph-pos
The results are generally inline with the deviations reported by the angle sensor.  (Though, the reference weighting in the integral is keeping the position from drifting.)

Interesting, it's possible to zoom in to see the individual pulses - for example:
adxl-graph-pos2
The total offset and drift on the accelerometer aren't particularly meaningful due to the integral calculations, but it is interesting that the timing of the deviations and the magnitude of the deviations match so closely to the angle sensors.

I'm inclined to merge this adxl345 branch.

Cheers,
-Kevin

dmbutyugin:
Thanks Kevin! That looks very interesting! I find it especially promising that the timing of the adxl345 readings aligns very well with the expected acceleration data.

> As before, the accel values themselves are showing very large swings. However, I've found that integrating the results (going from accel to velocity) shows promising data:

I think this noise is real - adxl345 detects the non-smooth motion due to stepper motor operation, bearings, and such, and it also picks up some ambient noise and the noise of the fans. It also happens when the printer is motionless. But it doesn't matter for the subsequent frequency analysis, so I don't see that as a problem.

I find this PR especially interesting in connection with my FR #4560. I hope we will be able indeed to collect the actual and expected readings, subtract one from the other, and get a reasonable diff that we can use to analyze the resonances.

I did take a very brief look at the latest developments in the code, and if you'd like, I could take a bit more close look in the next day or two. What I am especially interested in is how ClockSyncRegression works, how the missing data is handled (I do see occasionally data losses between the MCU and the RPI, when the sensor is connected to the MCU; and occasionally missed data reads when RPi is having some scheduling problems, I suppose), and how I could plug the simulated adxl345 (or get the expected acceleration values some other way) for the new resonance test implementation.

PR #4592, opened 2021-08-19
Elegoo Neptune 2 configuration

wsippel:
module: Elegoo Neptune 2 example configuration

Example config for the Elegoo Neptune 2, with the standard MKS Robin Nano 1.2 clone board. Some Neptune 2's use a Robin Nano 2.0 clone, this isn't for those machines. 

Signed-off-by: Willie Sippel <wsippel@me.com>


KevinOConnor:
Thanks.  In general, it looks fine to me, but "printer" config filenames need to end in a year as described at https://www.klipper3d.org/Example_Configs.html

-Kevin

wsippel:
Sorry, I renamed the file. Should I also add a static_digital_output line to disable the fsmc display?

KevinOConnor:
Thanks.  The config looks fine to me.  Alas, I don't know enough about the display to give advice there.  When you decide what you think is best, let me know.

-Kevin

KevinOConnor:
Thanks, I committed this change with some minor fixes on top (commit 9014677c).

-Kevin

PR #4598, opened 2021-08-20
Add config reading helpers for parsing lists in the config file

KevinOConnor:
This change adds new helper functions to the main `config` object for parsing out lists of numbers, lists of strings, and lists of lists.  I've also gone through and updated many places in the code that manually parse lists to use these new helper functions.

The main goal of this change is so that the `printer.configfile.settings` object will now contain lists for these existing config settings (instead of the encoded strings they currently have).  So, for example, where `printer.configfile.settings.extruder.gear_ratio` might have previously had something like `"50:17"`, it would now have `[[50.0, 17.0]]`.  This should make it easier for API clients to access the core configuration.

@Arksine , @cadriel, @meteyou - FYI, this PR changes the behaviour of `printer.configfile.settings`.  I didn't notice any regressions in Mainsail or Fluidd, but it is possible there is an impact.

@Arksine - the changes to bed_mesh.py were more invasive than the other changes.  You might want to double check them.

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20210821: Several config settings in `printer.configfile.settings`
+will now be reported as lists instead of raw strings.  If the actual
+raw string is desired, use `printer.configfile.config` instead.
+
 20210819: In some cases, a `G28` homing move may end in a position
 that is nominally outside the valid movement range.  In rare
 situations this may result in confusing "Move out of range" errors
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -74,6 +74,34 @@ class ConfigWrapper:
             raise error("Choice '%s' for option '%s' in section '%s'"
                         " is not a valid choice" % (c, option, self.section))
         return choices[c]
+    def getlists(self, option, default=sentinel, seps=(',',), count=None,
+                 parser=str, note_valid=True):
+        def lparser(value, pos):
+            if pos:
+                # Nested list
+                parts = [p.strip() for p in value.split(seps[pos])]
+                return tuple([lparser(p, pos - 1) for p in parts if p])
+            res = [parser(p.strip()) for p in value.split(seps[pos])]
+            if count is not None and len(res) != count:
+                raise error("Option '%s' in section '%s' must have %d elements"
+                            % (option, self.section, count))
+            return tuple(res)
+        def fcparser(section, option):
+            return lparser(self.fileconfig.get(section, option), len(seps) - 1)
+        return self._get_wrapper(fcparser, option, default,
+                                 note_valid=note_valid)
+    def getlist(self, option, default=sentinel, sep=',', count=None,
+                note_valid=True):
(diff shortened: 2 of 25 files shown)


zellneralex:
@KevinOConnor had no time to check it manual do you do that only for printer.configfile.settings. or also for printer.configfile.config?

KevinOConnor:
`printer.configfile.config` always reports the raw strings from the config and is not changed here.

-Kevin

Arksine:
Nice addition, definitely can see how this will be useful for clients.  Bed Mesh looks good to me, I'll give it a test run to be sure.

meteyou:
@KevinOConnor i just found an issue in the heightmap in mainsail... do you also want to convert the points in the bed_mesh? with this PR they looks like this:

```YAML
"bed_mesh energetics": {
    "tension": 0.2,
    "min_x": 30.0,
    "min_y": 30.0,
    "y_count": 5,
    "mesh_y_pps": 2,
    "x_count": 5,
    "mesh_x_pps": 2,
    "points": "\n0.070000, 0.068750, 0.078750, 0.076250, 0.058750\n0.040000, 0.041250, 0.031250, 0.046250, 0.030000\n0.002500, 0.016250, 0.000000, 0.017500, -0.003750\n0.025000, 0.023750, 0.021250, 0.027500, 0.016250\n0.075000, 0.077500, 0.072500, 0.095000, 0.083750",
    "algo": "lagrange",
    "version": 1,
    "max_x": 220.0,
    "max_y": 220.0
}
```

Arksine:
Ah, that is where bed_mesh is reading saved profile from the config.  I think the best we could get would be a list of strings since its a 2 dimensional array.

meteyou:
it doesn't matter if its a 2 dimension array or just one array with strings in it...
I only noticed this case because I was expecting it to have also changed, but it hasn't changed at all... i can work with all possible outputs.

internal i have to convert this value to this:
```YAML
"points": [
    [0.1, 0.2, 0.3, ...],
    [0.1, 0.2, 0.3, ...],
    [0.1, 0.2, 0.3, ...],
    ...
]
```

KevinOConnor:
Ah, I missed that one.  It should be a simple change.  I added it to this PR (commit 98f9ce03a).

Would making this change break existing mainsail / fluidd users though?

-Kevin

meteyou:
it don't break mainsail v1.6.0 (current stable release) but it would break V2-RC, but i can publish a new release any time of it...

i checked fluidd bed_mesh function via app.fluidd.xyz and is looks good.

KevinOConnor:
Sounds good.  Thanks.

Separately, it's possible I missed other cases of lists in the config.  If so, let me know, and we can update them as well.

-Kevin

meteyou:
i checked 2 printers and everything looks nice. i just found this:
```YAML
"tmc2209 extruder": {
    "microsteps": "16"
}
````
microsteps are formated as string instead of number. all other values are nice.

KevinOConnor:
Thanks.  I made the microstep change as well and committed this branch.

-Kevin

PR #4599, opened 2021-08-20
docs: Consistency Improvements

dingyifei:
Affect Rendering

- Added empty line above/below the fenced code blocks
- Added language specifiers to improve code block rendering
- Correct things like gcode to G-Code
- remove the double space after the period (do not show up in render).
- Added some empty lines I missed earlier
- 
Signed-off-by: Yifei Ding <yifeiding@protonmail.com>

# Please review the rendered text on Github!

### DIFF
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -149,7 +149,7 @@ transition to a "shutdown" state. It behaves similarly to the G-Code
 ### register_remote_method
 
 This endpoint allows clients to register methods that can be called
-from klipper.  It will return an empty object upon success.
+from Klipper. It will return an empty object upon success.
 
 For example:
 `{"id": 123, "method": "register_remote_method",
@@ -168,7 +168,7 @@ gcode:
   {action_call_remote_method("paneldue_beep", frequency=300, duration=1.0)}
 ```
 
-When the PANELDUE_BEEP gcode macro is executed, Klipper would send something
+When the PANELDUE_BEEP G-Code macro is executed, Klipper would send something
 like the following over the socket:
 `{"action": "run_paneldue_beep",
 "params": {"frequency": 300, "duration": 1.0}}`
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -8,7 +8,8 @@ control the Klipper host software.
 
 In order to use the API server, the klippy.py host software must be
 started with the `-a` parameter. For example:
-```
+
+```bash
 ~/klippy-env/bin/python ~/klipper/klippy/klippy.py ~/printer.cfg -a /tmp/klippy_uds -l /tmp/klippy.log
 ```
 
@@ -20,13 +21,15 @@ Klipper.
 
 Messages sent and received on the socket are JSON encoded strings
 terminated by an ASCII 0x03 character:
-```
+
+```json
 <json_object_1><0x03><json_object_2><0x03>...
 ```
 
 Klipper contains a `scripts/whconsole.py` tool that can perform the
 above message framing. For example:
-```
+
+```bash
 ~/klipper/scripts/whconsole.py /tmp/klippy_uds
 ```
 
@@ -158,7 +161,8 @@ will return:
 The remote method `paneldue_beep` may now be called from Klipper. Note
 that if the method takes parameters they should be provided as keyword
 arguments. Below is an example of how it may called from a gcode_macro:
-```
+
+```gcode
 [gcode_macro PANELDUE_BEEP]
 gcode:
   {action_call_remote_method("paneldue_beep", frequency=300, duration=1.0)}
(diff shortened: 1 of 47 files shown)


dingyifei:
My confidence in the last commit is not high...  Github does not support LaTeX and the mathematical equations in LaTeX look kinda "meh". 

图片

$value = start + factor \times (\lfloor{\frac{z_{height}}{band}}\rfloor + .5) \times band$

dingyifei:
I revert the LaTeX commit and merged the recent changes. I think it's ready.

KevinOConnor:
Thanks.

My main feedback is similar to my comments at https://github.com/KevinOConnor/klipper/pull/4577#issuecomment-902376709 .  In order to go forward with this change, I think we really want to have a separate commit for each high-level change.  So, for example, a commit with just capitalization changes across many files.  Or, alternatively, a single commit with many updates to just one document.

I know breaking up the commits is a lot of additional work, but I think it ultimately leads to time savings.  I've seen this on other projects and in my experience a "mega commit" can cause significant pain later on.  (In particular, if it results in merge conflicts or a revert is needed.)

There's a bunch of good improvements in this series.  However, they are mixed in with things I'm not sure about:
1. I'm not sure about the changes to whitespace.  Certainly it's worthwhile if there is an outlier in a file, but if the entire document is consistent with whitespace then I'm not sure it is worthwhile to make such an invasive change.  We can do it, but this kind of change I'd really like to see entirely separate and after any other content changes.
2. I'm not sure about the change to add `gcode`, `bash`, etc. identifiers to all the code blocks.  It certainly doesn't hurt, but I'm not sure it has much gain either.
3. I'm not sure I understand why blank lines are added in various places (MD022, MD012, MD040, MD031)?  Is this a rendering issue or a lint thing?  If the latter, I'm not sure it's worthwhile to make changes just to make a lint tool happy - the main reason to use markdown for the documentation is to encourage writing of free form text for documents - I don't think we want to adopt a strict style that would make it difficult for new contributors to add documents.

Again, I realize this is adding more work for you, which I don't want to do.  Maybe we could discuss the changes that aren't controversial, put together commits for them, get them merged, and then go on to the changes that are more involved.

Thanks again for looking at this.
-Kevin

dingyifei:
> I'm not sure about the changes to whitespace. Certainly it's worthwhile if there is an outlier in a file, but if the entire document is consistent with whitespace then I'm not sure it is worthwhile to make such an invasive change. We can do it, but this kind of change I'd really like to see entirely separate and after any other content changes.

Agree, it is certainly a massive and strange change which makes little to no difference in 90% the cases. I'll try to do some cherry picks and make it into the last commit.

The problem is the double white space will almost always be rendered as single white space, where none of the documents have exclusively single or exclusively double whitespace. Thus I'm thinking if we could make everything single space since the raw text will show the difference(in fluid for example) and it often add a strange extra space in the Weblate automatic suggestions.

> I'm not sure about the change to add gcode, bash, etc. identifiers to all the code blocks. It certainly doesn't hurt, but I'm not sure it has much gain either.

The identifiers change the way code blocks are rendered, where high lights will be added depending on the regular expressions. (I admit I'm probably a potato level identifier picker and CSS/regex editor)

It certainly makes a difference in how everything looks:

 image

> I'm not sure I understand why blank lines are added in various places (MD022, MD012, MD040, MD031)? Is this a rendering issue or a lint thing? If the latter, I'm not sure it's worthwhile to make changes just to make a lint tool happy - the main reason to use markdown for the documentation is to encourage writing of free form text for documents - I don't think we want to adopt a strict style that would make it difficult for new contributors to add documents.

They are for both linting and rendering, but are certainly at a very low priority(can be even lower than the double whitespaces), because it's more of making every parser happy(which might just be "over engineering"). 

---

Instead of pressing stylings to the new contributors, the styling should be a source string review side topic since they are hard to spot with a default parser in many editors like VScode and  many are impossible to spot unless you are comparing different parser output or reading the super duper long and frustrating markdown specifications. Translators can add string reviews on Weblate. Maybe instead of these "preventive measures" on rendering, only fix the known visible rendering and raw text issues can be sufficient to the readers. 

The large formatting changes are certainly not important. I think the following level of importance from higher(will affect the Gettext strings and translation output) to lower(parser comparability) like this:

1. Kib and KiB, list nesting issues, inconsistent/incorrect highlighting
2. Wording, Goode G-Code gcode Klipper klipper etc
3. Whitespaces and different syntax for bold text/italics/lists (sometimes will affect Gettext but not render
4. Markdown parser comparability (no effect to Gettext and translations)
5. Improvements(like highlighting)

The Gettext affecting changes(1,2, and 3) are more vital since if a doc contributor finds one and attempts to fix it after it's translated, all the translated strings correspond to the changed paragraph will require manual reviews (impossible to avoid). For example the Korean translation is at 54%, if all the white space changes are applied it will drop like 5 to 10% because these 5 to 10% will need review.

KevinOConnor:
Okay, that's fine with me.  Given the upcoming release, I don't think we want to make changes 4-6 at this time.  If you want to fix up 1-3 that would be great.

-Kevin

dingyifei:
Incredibly busy, also it seems like I need to merge from upstream and cherry-pick my commits

PR #4601, opened 2021-08-21
webhooks: log client requests on shutdown

Arksine:
This dumps a log of of requests for each connected client when Klippy enters the shutdown state.  The dump is modeled after the gcode log for consistency.  I chose to store 20 requests in the deque, however I can easily change that if necessary.


KevinOConnor:
Thanks!  I think this could be very helpful in tracking down issues that pop up, so lets merge now.

-Kevin

PR #4605, opened 2021-08-22
lpc176x: Fix wrong inital value at PWM setup

### LINKED ISSUE #4559: [output_pin (pwm)] only outputing 1 or 0 as initial value when 0.X is provided
I specified my LED_pin as follows, and it works great (also dimming with `0.X` values)

> [output_pin LED_pin]
> pin: P2.4
> pwm: True
> value: 0.1  # The value to **initially** set the pin to during MCU configuration.
> shutdown_value: 0.01 #<-- this works
> hardware_pwm: True
> cycle_time: 0.001 #s

I'd like the LEDs to be dim at startup (~`0.1`). If I understand correctly, the `value` parameter should do that. But if I provide `0.1` (or any other `0.X`) the LED startup full brightness (=`1`). If I provide `0` they start with `0`.

The only way I can get the `0.1` from the config is by sending `RESTART` to restart the host. Then the "MCU configuration" (don't know much about that process) presumably gets executed. But shouldn't this not also take place at the initial startup?

Is there something I can do to make it work?

Thanks in advance!

[klippy.log](https://github.com/KevinOConnor/klipper/files/6934439/klippy.log)

1am3r:
This fixes a PWM going to full scale output when a initial value != 0 was
given. The output was on 100% until another update occurred.

This change enables the PWM counter before setting the channel values.

This fixes #4559

### DIFF
--- a/src/lpc176x/hard_pwm.c
+++ b/src/lpc176x/hard_pwm.c
@@ -61,31 +61,30 @@ gpio_pwm_setup(uint8_t pin, uint32_t cycle_time, uint8_t val) {
         enable_pclock(PCLK_PWM1);
     }
 
-    if (LPC_PWM1->TCR & 1) {
+    if (LPC_PWM1->TCR & 0x1) {
         if (LPC_PWM1->PR != prescaler) {
             shutdown("PWM already programmed at different speed");
         }
     } else {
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks!

-Kevin

PR #4613, opened 2021-08-23
SAVE_CONFIG supports symlinked printer.cfg

RomRider:
Adding support for symlinked printer.cfg file.
Current behavior breaks the symlink and replaces it with a non symlinked file

### DIFF
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2016-2021  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import os, glob, re, time, logging, ConfigParser as configparser, StringIO
+import os, glob, re, time, logging, ConfigParser as configparser, StringIO, shutil
 
 error = configparser.Error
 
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks, but I'm not sure this is a good candidate for the Klipper master branch at this time.  In general, I'm leery of taking an automated action that could potentially result in the loss of the user's config file.  That's why the code goes through the mulitple rename steps - it is there to try and ensure the config isn't lost even in the event of OS errors.

Separately, commits need to have a signed-off-by line in order for me to commit them - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

randellhodges:
Doesn't the proposed code perform the same rename steps as the original code? Just asking from a learning/understanding point of view.

RomRider:
Thanks, for the feedback.

I'm not sure why the config would be lost?
Instead of a rename, I do a copy of the current config to a backup file and then I rename the new config (temp file) to be the current config.
The only difference is that today it does a rename of the current config to the backup config instead of a copy.

If copy is the issue, I'll make it work with a rename instead if you prefer to have atomic operations.

(I'll sign the commits later)

RomRider:
So it's closed but no feedback... Interesting approach 🤷‍♂️

KevinOConnor:
Thank you for the feedback.  I provided the best feedback I could at https://github.com/KevinOConnor/klipper/pull/4613#issuecomment-909727249 .  If you'd like to continue this support I think the next step would be to raise it with others in the Klipper community on Klipper Discord or Klipper Discourse.  If some consensus is found on how SAVE_CONFIG should work then that's great.

-Kevin

RomRider:
Thanks, I've also answered to that feedback, thus my last comment.

What I can tell you is that in my case, SAVE_CONFIG creates data loss because it replaces a perfectly valid symlink with a plain file.

I'm going to head over to discord but that creates a lot of overhead for such a small feature...

PR #4615, opened 2021-08-24
toolhead: change SET_VELOCITY_LIMIT respond behavior

meteyou:


KevinOConnor:
Thanks.

-Kevin

KevinOConnor:
On a closer look, I think there may have been a merge error (eg, `self.max_velocity` is assigned unconditionally prior to a conditional assignment).  I've reverted this change for further review (commit 75183bfb).

@DMSimone - fyi.

-Kevin

meteyou:
it was my fault. an error crept in when splitting the PR.

https://github.com/meteyou/klipper/commit/b2aa55a10502462df4c2eb09969271af17aa69a3

should i make a new pr? i tested it now.

PR #4616, opened 2021-08-24
extruder: change SET_PRESSURE_ADVANCE and SET_E_STEP_DISTANCE respond behavior

meteyou:


zellneralex:

KevinOConnor:
Okay.  FWIW, I'm not really sure it's worth changing these commands - mostly due to the possibility that the code change could introduce a regression.  Same feedback for #4617, #4618, and #4619.  Should we decide to merge these, it would be after the v0.10.0 release.

-Kevin

PR #4617, opened 2021-08-24
tmc: change SET_TMC_CURRENT respond behavior

meteyou:
Returns only the current values if no new ones have been passed and add a line break between run and hold current.


zellneralex:

PR #4618, opened 2021-08-24
input_shaper: change SET_INPUT_SHAPER respond behavior

meteyou:
Returns only the current values if no new have been passed.

### DIFF
--- a/klippy/extras/input_shaper.py
+++ b/klippy/extras/input_shaper.py
@@ -105,13 +105,13 @@ class InputShaper:
     cmd_SET_INPUT_SHAPER_help = "Set cartesian parameters for input shaper"
     def cmd_SET_INPUT_SHAPER(self, gcmd):
         damping_ratio_x = gcmd.get_float(
-                'DAMPING_RATIO_X', self.damping_ratio_x, minval=0., maxval=1.)
+                'DAMPING_RATIO_X', None, minval=0., maxval=1.)
         damping_ratio_y = gcmd.get_float(
-                'DAMPING_RATIO_Y', self.damping_ratio_y, minval=0., maxval=1.)
+                'DAMPING_RATIO_Y', None, minval=0., maxval=1.)
         shaper_freq_x = gcmd.get_float(
-                'SHAPER_FREQ_X', self.shaper_freq_x, minval=0.)
+                'SHAPER_FREQ_X', None, minval=0.)
         shaper_freq_y = gcmd.get_float(
-                'SHAPER_FREQ_Y', self.shaper_freq_y, minval=0.)
+                'SHAPER_FREQ_Y', None, minval=0.)
 
         def parse_shaper(shaper_type_str):
             shaper_type_str = shaper_type_str.lower()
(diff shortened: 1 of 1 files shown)


zellneralex:

klippy/extras/input_shaper.py
```diff
@@ -123,25 +123,41 @@ def parse_shaper(shaper_type_str):
...
-            shaper_type_y = gcmd.get('SHAPER_TYPE_Y', self.shaper_type_y,
+            shaper_type_y = gcmd.get('SHAPER_TYPE_Y', None,
                                      parser=parse_shaper)
         else:
             shaper_type_x = shaper_type_y = shaper_type
 
-        self._set_input_shaper(shaper_type_x, shaper_type_y,
-                               shaper_freq_x, shaper_freq_y,
-                               damping_ratio_x, damping_ratio_y)
+        if damping_ratio_x is not None:
```
dmbutyugin, line 133: I'm afraid that writing the new values directly into the member variables will break new_delay / old_delay  calculations and the subsequent call to `toolhead.note_step_generation_scan_time` in `_set_input_shaper` method. This may manifest itself as the errors in the stepcompress code upon changes of the shaper frequency or the shaper type. So, this isn't really the right way to do it.

dmbutyugin:
I'm generally curious how useful this change is? Are there scenarios where the input shaper parameters are changed very frequently so there is a desire to reduce the logging?

FWIW, in my experience, it is very useful to get a full response on the command like `SET_INPUT_SHAPER SHAPER_FREQ_X=45` so that one could see what the shaper type is, and whether the respective shaper frequency makes sense. Then, arguably, for a command like `SET_INPUT_SHAPER SHAPER_TYPE_X=EI` it is even more important to get the rest of the parameters, because it is very easy to misconfigure the input shaper that way. Note that I'm talking more about the cases when the input shaper parameters are changed manually, and not as a result of an automatic gcode post-processing script, for instance.

In general, I see that there are some differences between the `SET_INPUT_SHAPER` and `SET_VELOCITY_LIMIT` commands - the arguments of the latter are mostly independent, so the command `SET_VELOCITY_LIMIT ACCEL=3000` is more or less self-contained. But this is not quite true for the `SET_INPUT_SHAPER` command.

klippy/extras/input_shaper.py
meteyou: oh ok. so if i reverse it to:

```python
if damping_ratio_x is None:
  damping_ratio_x = self.damping_ratio_x
```
and check before for the output, it would fit?

meteyou:
i just tested all gcode commands and prepared a PR for all those who immediately make a respond (also when setting) to have a basis for conversation. if we would merge these PRs, then we would have all gcodes like "linux bash commands".

you are absolutely right that SET_INPUT_SHAPER will definitely not overflow the console in practice. i could not think of any use case where this could happen.

klippy/extras/input_shaper.py
dmbutyugin: Yeah, that should work.

dmbutyugin:
> i just tested all gcode commands and prepared a PR for all those who immediately make a respond (also when setting) to have a basis for conversation. if we would merge these PRs, then we would have all gcodes like "linux bash commands".

Well, this API in general seems a bit odd: `SET_VELOCITY_LIMIT` with parameters changes those parameters (but reports nothing), and the same command without any parameters returns the current parameters. It looks like it would be better to have a dedicated command `GET_VELOCITY_LIMIT` for that case instead? In fact, we already have precedents of such commands: `GET_RETRACTION` (vs `SET_RETRACTION`), `GET_POSITION`, `GET_CURRENT_SKEW`, `BED_MESH_OUTPUT`, etc. 

> you are absolutely right that SET_INPUT_SHAPER will definitely not overflow the console in practice. i could not think of any use case where this could happen.

Well, I was worried that merging this PR will make some current valid workflows (changing the input shaper manually) more error prone. So, I was trying to understand what's wrong with the command responding the status of the operation right away? Is having all gcodes behaving like "linux bash commands" beneficial in some way?

zellneralex:
@dmbutyugin we tried so many different approaches. Add a parameter and defaulted it so that the behavior stays the same, add a new GCode like GET_VELOCITY_LIMIT. All these was disliked by Kevin. How it is done now was the recommendation from him.
All the other GCODE will not flood the console like GET_VELOCITY_LIMIT does but to have a consistent behavior PRs for all other SET_… GCodes are prepared

dmbutyugin:
@zellneralex OK, thanks for your explanation about `SET_VELOCITY_LIMIT`. That makes sense.

I'm still not sure if we should change the behavior of `SET_INPUT_SHAPER`, mainly because it is used to manually adjust the input shaper (AFAICT), and in that case reporting all other parameters helps avoiding mistakes which are, otherwise, hard to spot.

PR #4619, opened 2021-08-24
idle_timeout: change SET_IDLE_TIMEOUT respond behavior

meteyou:
Returns only the current values if no new have been passed.


zellneralex:

PR #4620, opened 2021-08-24
toolhead: change SET_VELOCITY_LIMIT respond (fixed)

meteyou:
fixed the issue from PR https://github.com/KevinOConnor/klipper/pull/4615


zellneralex:
I tested it and this time it works

KevinOConnor:
Thanks.

-Kevin

PR #4621, opened 2021-08-25
add support for creality sermoon D1 printer

peterpanstechland:
created a config support for creality sermoon D1 printer with option of adding bltouch.


peterpanstechland:
this printer come with a Creality-4.3.1 board.

KevinOConnor:
Thanks.  In general, the config looks good to me.  However, commits need to have a signed-off-by line containing your full real name - https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

Also, all the configs are added to the regression test case (as describe at https://www.klipper3d.org/Example_Configs.html ) and so the PID parameters need to be defined in the heater sections.

-Kevin

peterpanstechland:
Thanks Kevin,

I think the latest commit should have fixed all the requirements and followed the contributing guide lines, hopefully this config file could finally get merged.😉

-Peter

KevinOConnor:
Thanks, but this doesn't look like it will pass the regression test cases as the pid parameters aren't defined in the heater_bed config section.  Also, the PID parameters should use `option: value` instead of `option = value` syntax (as described in the document above).

-Kevin

peterpanstechland:
> Thanks, but this doesn't look like it will pass the regression test cases as the pid parameters aren't defined in the heater_bed config section. Also, the PID parameters should use `option: value` instead of `option = value` syntax (as described in the document above).
> 
> -Kevin

Sorry, I think I've used someone's sample fork, didn't pay attention on the format, will fix that by next commit.

-Peter

KevinOConnor:
Thanks.

-Kevin

PR #4622, opened 2021-08-25
Fix fan feedback with max_power != 1

AzazKamaz:


AzazKamaz:
I agree that it can be fixed at UI level. But at the klipper level logic will be strange. When setting value it is scaled by max_power but it is not when getting. Like you ask it to rotate on full speed but it rotates only on 50%.

If it is necessary to show user that full speed is not 100% then we can just show max_power instead of 100% in UI. We can have it scaled in both directions in klipper itself or not scaled at all (but ensure that value is less that max_power)

When making fix to this in UI we will keep an inadequate interface in klipper. It is not possible even to get value and set it back without unobvious actions

Vhab:
Fixing it at the klipper level does feel more appropriate.

When requesting a 100% fan, klipper adjusts it to 50%, so it makes sense to report it back as a 100%.
If the true value is needed, a simple calculation can be made, or an additional value like "true_speed" could be exposed.

When I ran into this issue in Fluidd (which doesn't yet adjust for max_power), my instinct was to fix it at moonraker or klipper level. I only went for the Fluidd level fix because that's what Mainsail did.

zellneralex:
As said then make it consistent we have the same behavior e.g at the heater.

in my opinion the behavior is correct. You add a limiter with max power and .speed shows the actual speed and that e.g 50%

Vhab:
Do note max_power is not a limit, but it rescales the value, If you set it to 0.5, then requesting 100% will result in 0.5, but requesting 50% will result in 0.25.

That said, this is not a hill I'm willing to die on, I'm ok with fixes on either end. Klipper's side just feels more intuitive.

zellneralex:
We can discuss that the next year or two, as there is no right or wrong. Yes I know that M106 is rescaled to max power to be backwards compatible.

For me .fan.speed is the actual speed based on full scale. Both definitions are true as they are only definitions.

And as said that is the same for heaters. So if we change it on fans we need also to change it at heaters to stay consistent. So if you set e.g. for your bed_heater a max_limit of 60 you will see that it shows a maximum of 60%. 
So please explain why you believe that definition is wrong at a fan but ok for an heater or a temperature_fan.

KevinOConnor:
Thanks.  I agree that the `speed` name is misleading, however I think the current behaviour should be clarified in the documentation and not via a code change.  That is, I think we should just document that the `speed` parameter returns the fan power as a value between 0.0 and 1.0.

-Kevin

PR #4626, opened 2021-08-28
Export tmc driver status

KevinOConnor:
This PR adds a `get_status()` method to the TMC driver objects.  The stepper phase information and recently queried DRV_STATUS information is reported there.

@Arksine , @meteyou , @cadriel - FYI, this allows access to the driver temperature warning flag (eg, `printer["tmc2209 stepper_x"].drv_status.otpw`).  I suspect that information may be of interest to the front-ends.

-Kevin

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -311,6 +311,21 @@ objects:
   temperature seen by the sensor since the Klipper host software was
   last restarted.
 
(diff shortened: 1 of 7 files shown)


meteyou:
sry for the late respond. tested it and it works fine for me and this are very nice values to know!

PR #4631, opened 2021-08-29
Add files via upload

gcastle2007:
Example default config for printer Tronxy XY-2 Pro.

### DIFF
--- a/config/printer-Tronxy XY-2-Pro-(no-Titan,-no-BMG).cfg
+++ b/config/printer-Tronxy XY-2-Pro-(no-Titan,-no-BMG).cfg
@@ -1,0 +1,153 @@
+# This is a Klipper configuration for Tron XY-2 Pro, with
+# CXY-V6 motherboard. no-Titan, no-BMG.
+
+
+#            === FLASHING WITH STOCK BOOTLOADER ===
+# You should make firmware for STM32F103 with bootloader offset
+# at 0x8008800 (Chitu v6 Bootloader). Uncheck USB, and leave default
+# serial settings.
+#
+# Use "./scripts/update_chitu.py ./out/klipper.bin ./out/update.cbd" after make to generate update.cbd.
+# Put `update.cbd` onto SD card, and reboot the printer.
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Example configs would need to follow the guidelines described at: https://www.klipper3d.org/Example_Configs.html

-Kevin

gcastle2007:
I have updated file and filename.

KevinOConnor:
Thanks.  The printer-tronxy-xy-2-Pro-2020.cfg looks fine to me, but it looks like this PR is adding both the old and new files.  Likely needs a `git rebase` to resolve.

-Kevin

gcastle2007:
Unfortunately, I had some unfriendliness with the github, so I deleted the old file via the Web interface.
Please check the result.

KevinOConnor:
Thanks.  In general, it looks fine.  However, commits need to have a signed-off-by line with your full real name - https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md .

Out of curiosity, why does the bed_screws section have different screw locations from the bed_screws_adjust section?

It's also generally nice to define a gear_ratio for the extruder (by inspecting the number of gears in the mechanism) so that the defined rotation_distance is the circumference of the bolt that engages with the filament.

-Kevin

gcastle2007:
I have removed section bed_screws and bed_screws_adjus. Whoever needs it - they can create for themselves on their own.
About gear_ratio: default printer have CR-10 extruder.
I'm not undestand whats is CONTRIBUTING.md and and where should I put this file.

Kevin, if all last question this is the problem for adding the config to the klipper distribution, then close the request and don't do anything. Who needs it - they themselves will search for the config on the Internet or create it themselves.

gcastle2007:
Kevin,

Printer use motor 1.8 degress.

Step Angle: 1.8deg-> https://aliexpress.ru/item/4000220942066.html?spm=a2g0o.store_pc_groupList.8148356.43.58db47c8BZI0mF

PR #4633, opened 2021-08-29
Add get_status() method for screws_tilt_adjust

theophile:
This adds a get_status() method to the screws_tilt_adjust.py module so that gcode macros and menu elements can read the results from the `screws_tilt_calculate` command. This creates a new `printer.screws_tilt_adjust` object with two properties:

`error`: False by default, but True if the `screws_tilt_calculate` command triggered the "bed level exceeds configured limits" error.
`results`: Until the `screws_tilt_calculate` command is run, this is an empty python list. But once `screws_tilt_calculate` has completed successfully, this is a list containing a python dictionary for each configured bed screw. Each of those dictionaries contains the following keys:

- **name**: Value is the name of the screw from the config file
- **x**: Value is the configured X coordinate of the screw
- **y**: Value is the configured Y coordinate of the screw
- **z**: Value is the measured Z probe height above the screw
- **sign**:* Value is either "CW" for clockwise or "CCW" for counter-clockwise 
- **adjust**:* Value is the amount the screw must be adjusted, expressed in the same "time" format as is output in the console. (e.g. `01:15` for one full turn plus a quarter turn)

*_The dictionary for the "base" screw does not have the **sign** or **adjust** keys because the calculation assumes the base screw will not be adjusted._

I have successfully used this new get_status() method together with a handful of macros, custom menus, display groups, and even a couple of custom glyphs to enable Z-probe bed tramming directly from the printer LCD, with a rudimentary GUI and everything.


theophile:
Sorry, messed up the PR. Will try again...

Issue #4623, opened 2021-08-26
Klipper fails to install


shadow7412:
Starting from an octopi image as per [this page](https://www.klipper3d.org/Installation.html), I ran into the following error when running `make`.

```
src/i2ccmds.c:14:23: error: field ‘i2c_config’ has incomplete type
     struct i2c_config i2c_config;
                       ^
src/i2ccmds.c: In function ‘command_config_i2c’:
src/i2ccmds.c:23:23: warning: implicit declaration of function ‘i2c_setup’ [-Wimplicit-function-declaration]
     i2c->i2c_config = i2c_setup(args[1], args[2], addr);
                       ^
src/i2ccmds.c: In function ‘command_i2c_write’:
src/i2ccmds.c:35:5: warning: implicit declaration of function ‘i2c_write’ [-Wimplicit-function-declaration]
     i2c_write(i2c->i2c_config, data_len, data);
     ^
src/i2ccmds.c: In function ‘command_i2c_read’:
src/i2ccmds.c:48:5: warning: implicit declaration of function ‘i2c_read’ [-Wimplicit-function-declaration]
     i2c_read(i2c->i2c_config, reg_len, reg, data_len, data);
     ^
make: *** [Makefile:64: out/src/i2ccmds.o] Error 1
```

PR #4635, opened 2021-08-29
Allow release info changes when installing on Octoprint

shadow7412:
Follow on from issue #4623, to make installation a little more seamless from a fresh octopi image.

### DIFF
--- a/scripts/install-octopi.sh
+++ b/scripts/install-octopi.sh
@@ -21,7 +21,7 @@ install_packages()
 
     # Update system package info
     report_status "Running apt-get update..."
-    sudo apt-get update
+    sudo apt-get update --allow-releaseinfo-change
 
     # Install desired packages
     report_status "Installing packages..."


KevinOConnor:
I'm not sure what this change does, but it hasn't been needed for me.  I'm inclined to say that this isn't currently a good match for the master Klipper repository.

Separately, commits need to have a signed-off-by line - https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

zellneralex:
@KevinOConnor the change is needed as buster changed state to old-stable. It’s already added to the moonraker updater.
@Arksine can confirm that

shadow7412:
@KevinOConnor I've edited the commit to include the line.

It's mentioned in the description - but you should be able to see what this does by installing a fresh octopi image then attempting to run `./klipper/scripts/install-octopi.sh` (after cloning of course). Without it, the script errors.

If you don't want this included in the klipper install I understand - but the docs would need to be updated to include this instruction.

shadow7412:
@zellneralex I've confirmed this - it's [here](https://github.com/Arksine/moonraker/blob/master/scripts/install-moonraker.sh#L44).

guysoft:
Hey, just found this, had the same issues.
Related:
https://github.com/guysoft/CustomPiOS/issues/143
https://github.com/RPi-Distro/pi-gen/issues/546

KevinOConnor:
Thanks.  I understand that this is an issue for installations.  However, am not comfortable with the Klipper installation scripts silently making an OS level change.  So, this isn't a change I'm comfortable making.

-Kevin

KevinOConnor:
Thanks.  I understand that this is an issue for installations.  However, I am not comfortable with the Klipper installation scripts silently making an OS level change.  So, this isn't a change I'm comfortable making.

-Kevin

shadow7412:
Understandable. Please consider updating the documentation instead.

guysoft:
Hey, I suggest that instead changing the OS, you can switch an OS, its also much easier to install this way, I made a dedicated distro.
https://github.com/Klipper3d/klipper/pull/4895

PR #4639, opened 2021-08-30
Correct typo of ADXL345

fredrikbaberg:


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit directly - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .  I made the fix though - commit 1feea0cfe.

-Kevin

PR #4644, opened 2021-09-01
heater fan: Allow for a minimum target temperature

SeanReg:
Allows the user to set a minimum target temperature before the heater fan will turn on. This allows for things like exhaust fans to turn on only when higher temp filament is being used such as ABS (Like heater: heater_bed, min_target_temp: 80)

### DIFF
--- a/klippy/extras/heater_fan.py
+++ b/klippy/extras/heater_fan.py
@@ -14,6 +14,7 @@ class PrinterHeaterFan:
         self.printer.register_event_handler("klippy:ready", self.handle_ready)
         self.heater_names = config.getlist("heater", ("extruder",))
         self.heater_temp = config.getfloat("heater_temp", 50.0)
+        self.min_target_temp = config.getfloat("min_target_temp", 0.0)
         self.heaters = []
         self.fan = fan.Fan(config, default_shutdown_speed=1.)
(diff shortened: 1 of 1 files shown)


dalegaard:
Isn't this what `temperature_fan` is for? The `heater_fan` name indicates that it's for use with heaters and in that case the only safe choice is enabling the fan when there's power to the heater.

SeanReg:
> Isn't this what `temperature_fan` is for? The `heater_fan` name indicates that it's for use with heaters and in that case the only safe choice is enabling the fan when there's power to the heater.

Only "downside" to using `temperature_fan` is that the sensor and pin will be redefined and registered twice. I guess this really isn't an issue though

SeanReg:
@dalegaard 

`temperature_fan` can't be used because Klipper will prevent the redefinition/reuse of the bed sensor pin. Another option is to make these changes/additions to `controller_fan` instead so that a `controller_fan` can be turned off/on at a certain temp of the specified heater.

KevinOConnor:
Thanks.  However, I'm not sure this is a good candidate for the master Klipper repo.  As mentioned above, one of the main purposes of the heater_fan module is that it will enable the fan whenever the associated heater is enabled.   (In particular, changing fan speeds based on heater temperature can cause poor interactions with the heater PID.) If that behavior isn't desired, it would seem heater_fan isn't the right fan object to use.

-Kevin

PR #4652, opened 2021-09-03
atsamd: Add i2c_read() implementation

simonkuehling:
I have added an implementation of the i2c_read() function in the atsamd firmware. The code is tested and running well with Honeywell ABP pressure sensors and was also checked for protocol conformity with a logic analyzer.

Signed-off-by: Simon Kühling <mail@simonkuehling.de>


KevinOConnor:
Thanks.  The changes look fine to me.  However, out of an abundance of caution, I'm inclined to look at merging this after the v0.10.0 release.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #4658, opened 2021-09-04
Fix: inverted rotation for screw_tilt_adjust

Biorn1950:
- I just reversed the instruction for the adjust sens (was writing to tighten the screws to up the bed)
- Example of the issue with the original code with thread CW-M4 where CW = tighten the screw and CCW = loosen the screw:
```
// front left screw (base) : x=7.0, y=9.0, z=2.33000
// front right screw : x=177.0, y=9.0, z=2.31250 : adjust CW 00:01
// rear right screw : x=177.0, y=179.0, z=2.29500 : adjust CW 00:03
// rear left screw : x=7.0, y=179.0, z=2.34500 : adjust CCW 00:01
```

 - Biorn Walliser (biorn@me.com)

### DIFF
--- a/klippy/extras/screws_tilt_adjust.py
+++ b/klippy/extras/screws_tilt_adjust.py
@@ -93,9 +93,9 @@ class ScrewsTiltAdjust:
                 else:
                     adjust = diff / threads_factor.get(self.thread, 0.5)
                 if is_clockwise_thread:
-                    sign = "CW" if adjust >= 0 else "CCW"
-                else:
                     sign = "CCW" if adjust >= 0 else "CW"
+                else:
+                    sign = "CW" if adjust >= 0 else "CCW"
                 adjust = abs(adjust)
                 full_turns = math.trunc(adjust)
                 decimal_part = adjust - full_turns


celtare21:
Did you declare your screw_thread in your config?
For me it's CW-M4 and I don't need to modify the source

Biorn1950:
> Did you declare your screw_thread in your config?

> For me it's CW-M4 and I don't need to modify the source

So for you turn a CW screw in the CW direction mean unscrew?

JamesH1978:
I assume this has been written for looking top down onto the bed and wheels, so CW and CCW means as you look at it. But what Biorn means this is not mechanically correct, as mechanically the screw thread is inverted when observing from above, therefore the instructions given are inverted when mechanically observing the wheels from below.

KevinOConnor:
Okay, but this doesn't look like something we should change in the code - as it would break everyone's existing configs.  This looks instead like something that should be clarified in the documentation.

@rmcbc , @matthewlloyd - do you have any comments on this?

-Kevin

Biorn1950:
That won't really break the existing config but give the good instruction. But yeah that will change the habits of currents users.

As you prefer ;)

matthewlloyd:
I agree with Kevin:

* This just needs to be clarified in the documentation. The code is "correct", the docs are not clear enough.
* Inverting the sense of "CW" and "CCW" will break the configs of anyone who has configured and is already using this, since presumably they will already have inverted the orientation in their config if needed.

This is the current doc:

```
#screw_thread: CW-M3
#   The type of screw used for bed level, M3, M4 or M5 and the
#   direction of the knob used to level the bed, clockwise decrease
#   counter-clockwise decrease.
```

We need to:
* Fix the double "decrease", that's either a typo or needs rewording.
* To clarify whether "decrease" means to increase or decrease the height of the _bed_, or the _distance from the nozzle to the bed_.

The "CW" or "CCW' referred to in the screw_thread parameter does not refer to the handedness of the thread on the screw itself, (since almost all are right-handed anyway), rather it specifies the direction the screw needs to be turned to decrease the bed-nozzle distance. I don't think this code was written from any specific perspective in terms of whether the adjustment screws are on the top surface (e.g. most Prusas with bed adjust mods) or accessed from underneath (Ender 3s etc.).

I used this regularly on my Mini+, where tightening the screw on the top of the bed by turning it clockwise would lower the bed and increase the bed-nozzle distance. The CW and CCW output was made correct by setting `screw_thread: CCW-M3`.

To summarize, I think we just need doc clarification here. How about:

```
#screw_thread: CW-M3
#   The size of screw used for bed leveling, M3, M4 or M5, and the
#   direction the screw or knob should be turned to raise the bed
#   and move the surface closer to the nozzle. The direction depends
#   on whether the screw is accessed from above or below the bed,
#   and whether the screw is right-handed (standard) or left-handed.
#   For example, for a standard right-handed M3 screw accessed
#   from below, with an attached knob which when turned clockwise
#   (tightened) raises the bed, set this to CW-M3. For a similar screw,
#   accessed from above, which when turned clockwise (tightened)
#   lowers the bed, set this to CCW-M3.
```

Biorn1950:
> #   For example, for a standard right-handed M3 screw accessed
> #   from below, with an attached knob which when turned clockwise
> #   (tightened) raises the bed, set this to CW-M3.

@matthewlloyd that exactly where the problem is. If you turned clockwise the knob that will lower the bed. (with an ender 3 for example)

matthewlloyd:
> @matthewlloyd that exactly where the problem is. If you turned clockwise the knob that will lower the bed. (with an ender 3 for example)

I see, I don't own an Ender 3 so I did not know. It is unfortunate that the default value has the opposite sense to what would work on one of the most popular printers. Nevertheless, inverting the sense in the code will break a lot of people's configurations, while the documentation can simply be clarified to make it correct (even though most users will still have to set this to CCW-M3 to get the expected behavior).

Changing the default value of screw_thread would also break the configurations of anyone who uses this but hasn't overridden it. This is definitely better but still not ideal.

If you can think of a solution that makes the default more appropriate for the majority of the userbase, but doesn't break existing configurations, I'm all ears...

Biorn1950:
> If you can think of a solution that makes the default more appropriate for the majority of the userbase, but doesn't break existing configurations, I'm all ears...

Maybe just change the instructions with something like that? (with the words you prefer, english isn't my first language)

```
// front left screw (base) : x=7.0, y=9.0, z=2.33000
// front right screw : x=177.0, y=9.0, z=2.31250 : Distance increase 00:01
// rear right screw : x=177.0, y=179.0, z=2.29500 : Distance increase 00:03
// rear left screw : x=7.0, y=179.0, z=2.34500 : Distance decrease 00:01
```

Biorn1950:
> Changing the default value of screw_thread would also break the configurations of anyone who uses this but hasn't 
> overridden it. This is definitely better but still not ideal.

Yeah that why I make this PR, if something is wrong, it should be fixed a day. Editing the doc is just a way around the problem in my opinion.

> If you can think of a solution that makes the default more appropriate for the majority of the userbase, but doesn't break existing configurations, I'm all ears...

@matthewlloyd  Maybe just change the instructions with something like that? (with the words you prefer, english isn't my first language)

```
// front left screw (base) : x=7.0, y=9.0, z=2.33000
// front right screw : x=177.0, y=9.0, z=2.31250 : Distance increase 00:01
// rear right screw : x=177.0, y=179.0, z=2.29500 : Distance increase 00:03
// rear left screw : x=7.0, y=179.0, z=2.34500 : Distance decrease 00:01
```

Biorn1950:
Since English is not my native language, I'm not sure to use the good words for a clear prompt. After all people could set the screw type in CCW like matthew did. Or if you prefer maybe 'raise 00:42' 'lower 01:54' could be the solution ?

Tell me if you want that and I'll edit the pr including docs or close the pr.

Biorn1950:
The raise_direction is a nice trick. And the prompt is perfect.

@KevinOConnor if it's ok for you, I'll edit the pr in this way.

PR #4659, opened 2021-09-04
Export config file warnings for deprecated options

KevinOConnor:
This PR will add a new `printer.configfile.warnings` entry containing a string message reporting any deprecated config options.  This should be useful for warning users about needed config updates before they become errors.

@Arksine , @meteyou , @cadriel - it would be great if the frontends could prominently display this information.

-Kevin

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -28,6 +28,12 @@ The following information is available in the `configfile` object
   setting as read by Klipper during the last software start or
   restart. (Any settings changed at run-time will not be reflected
   here.) All values are returned as strings.
+- `save_config_pending`: Returns true if there are updates that a
+  `SAVE_CONFIG` command may persist to disk.
+- `warnings`: A list of warnings about config options. Each entry in
+  the list will be a dictionary containing a `type` and `message`
+  field (both strings). Additional fields may be available depending
+  on the type of warning.
 
 ## display_status
 
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -116,6 +116,17 @@ class ConfigWrapper:
     def get_prefix_options(self, prefix):
         return [o for o in self.fileconfig.options(self.section)
                 if o.startswith(prefix)]
+    def deprecate(self, option, value=None):
+        if not self.fileconfig.has_option(self.section, option):
+            return
+        if value is None:
+            msg = ("Option '%s' in section '%s' is deprecated."
+                   % (option, self.section))
+        else:
+            msg = ("Value '%s' in option '%s' in section '%s' is deprecated."
+                   % (value, option, self.section))
+        pconfig = self.printer.lookup_object("configfile")
+        pconfig.deprecate(self.section, option, value, msg)
 
 AUTOSAVE_HEADER = """
 #*# <---------------------- SAVE_CONFIG ---------------------->
@@ -127,8 +138,10 @@ class PrinterConfig:
     def __init__(self, printer):
         self.printer = printer
         self.autosave = None
+        self.deprecated = {}
         self.status_raw_config = {}
         self.status_settings = {}
+        self.status_warnings = []
         self.save_config_pending = False
         gcode = self.printer.lookup_object('gcode')
         gcode.register_command("SAVE_CONFIG", self.cmd_SAVE_CONFIG,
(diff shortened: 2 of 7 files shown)


Arksine:
I have a PR (arksine/Moonraker#223) that will add this to the list of warnings.  As soon as its merged it will show in Mainsail, it doesn't appear that Fluidd shows warnings at the moment.

Below is an example.
image

It might be a good idea to render newlines so that the warnings are easier to read.  Also clients might wish to check the warning list each time Klipper reports "ready" so it can update the UI if the user fixes config.  Alternatively clients could just fetch Klipper's warnings directly from the configfile and create a separate section, in which case the Moonraker PR would be largely unnecessary.

FrYakaTKoP:
I hope you guys are prepared for the big support wave coming out of this 😊

(don't get me wrong, i like it. sooner or later people need to get rid of their legacy config)

Arksine:
I'm sure there will be plenty of "what do I do" questions, but this is a pretty easy fix for the community.  It also gives us a chance to get people moved over before those options are invalid.

The remaining holdouts on rotation distance will probably be the most difficult.  It might be useful if the frontends provided a rotation distance calculator.

FrYakaTKoP:
don't worry most of the support questions will land on the frontend discords, where we are now trained after the first "auth enable" wave. 

What i would like to see is, instead a long string block with warnings some kind of list of warnings (it's already in a json why not add another layer or two). This would allow the frontend to render the warnings better also may directly hint the user to a guide. What in my opinion would help alot.

Didn't discuss this with mete, but i think it is worth the effort. Atleast we could reduce some of the questions.

meteyou:
i discussed this topic with @FrYakaTKoP and the best option for the guis would be a array with a object of each warning like this:

```json
[
  {
    type: "deprecated",
    section: "gcode_macro PANELDUE_BEEP",
    option: "default_parameter_frequency",
    value: 1000,
    message: "Option 'default_parameter_duration' in section 'gcode_macro_PANELDUE_BEEP' is deprecated."
  },
  {
    type: "deprecated",
    section: "gcode_macro PANELDUE_BEEP",
    option: "default_parameter_layer_height",
    value: 0.2,
    message: "Option 'default_parameter_layer_height' in section 'gcode_macro_PANELDUE_BEEP' is deprecated."
  }
]
```
with this option we can integrate a translation for warnings and for the fallback we have the message itself. This is just a suggestion, we could implement the best possible output in mainsail.

Arksine:
If it were possible to specify the file name of the config file containing the deprecated (or invalid) option it would probably be useful.  Not sure that is easily done though, might be something to consider at a later date if the configfile module needs significant refactoring to track that

KevinOConnor:
Thanks.  I made some changes based on the feedback at https://github.com/KevinOConnor/klipper/pull/4659#issuecomment-915039252 .  As recommended, `printer.configfile.warnings` will now be a list containing dictionaries - where each dictionary will have (at a minimum) `type` and `message` fields.  I did introduce two different types right now - `deprecated_option` and `deprecated_value`.  Only deprecated_value types will have a `value` parameter.

-Kevin

KevinOConnor:
> If it were possible to specify the file name of the config file containing the deprecated (or invalid) option it would probably be useful. Not sure that is easily done though, might be something to consider at a later date if the configfile module needs significant refactoring to track that

It might be possible to track that, but it doesn't seem to be a simple change.  So, I'd recommend going forward with this smaller change before the v0.10.0 release - if desired, we can add a `filename` field later.

-Kevin

meteyou:
this works fine for me! now i have to fix some deprecated options...

image

is it possible to add the value to each option? i think it is easier for many users to understand if to find / recognize the line (as it is in their config). since from time to time certain options can be named very similar to sections.

KevinOConnor:
It's possible to include the `value` field, but that seems confusing to me when the entire option is deprecated.  I fear it would be confused with a `deprecated_value` report (`Value 'rpi_temperature' in option 'sensor' in section 'controller_fan host_fan' is deprecated.`).

-Kevin

KevinOConnor:
FYI, I merged this branch.  We can add additional fields in the future if desired (after the v0.10.0 release).

Thanks,
-Kevin

Issue #3026, opened 2020-06-27
[FR] Adaptive acceleration support in Klipper


dmbutyugin:
Adaptive acceleration control can reduce acceleration of the toolhead on short moves and increase it on long moves. One possible implementation uses 'jerk-limiting' to control acceleration (implemented in [scurve-shaping](https://github.com/dmbutyugin/klipper/tree/scurve-shaping) and [scurve-smoothing](https://github.com/dmbutyugin/klipper/tree/scurve-smoothing) branches). It seems to additionally reduce vibrations vs plain input shaping or smoothing - at the expense of slowing down the prints to some degree - and might be beneficial at least to some users.

This is a dedicated ticket to discuss this feature instead of #57, which got too large to manage.

klipper-gitissuebot:
Hi @dmbutyugin,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

Unfortunately, too many people have opened tickets without providing the log. That consumes developer time; time that would be better spent enhancing the software. If this ticket references an event that has occurred while running the software then the Klipper log must be attached to this ticket. Otherwise, this ticket will be automatically closed in a few days.

For information on obtaining the Klipper log file see: https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md

The log can still be attached to this ticket - just add a comment and attach the log to that comment.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

BlackStump:
scurve-smoothing is my go to branch for normal printing. It has been for some time. scurve-shaping looks promising on the test prints at least.
When I have a need to print something it is generally set to scurve-smoothing, I will change to scurve-shaping for normal printing to see how it compares on general usage.

DroneMang:
> scurve-smoothing is my go to branch for normal printing. 

Is this branch as updated as master?

BlackStump:
@dmbutyugin 
Is it possible to update the scurve-smoothing branch to current master?

Cheers

dmbutyugin:
@BlackStump Well, input_shaper was integrated into the mainline Klipper. smooth_axis feature is sort of incompatible with it. I would suggest you to use scurve-shaping branch instead. Is there a reason you cannot use that one?

BlackStump:
[...]

PR #2030, opened 2019-10-01
Experiment with s-curves

KevinOConnor:
This branch contains the code to implement an experiment with s-curves to try and reduce "ringing" (aka "ghosting") on prints.  This is just a test, and is likely only of interest to other developers.

The current Klipper code uses a standard "trapezoid" velocity motion planner.  PR #1776 changes that planner to implement "s-curves" during acceleration:

trapezoid-bezier svg

This branch attempts a different approach:

trapezoid-antires svg

That is, instead of implementing the curves within the time allotted for normal acceleration, the code permits acceleration leading up to and after the nominal acceleration time.  This does result in the steppers taking a "path" that deviates from the commanded path - but it is done with the high-level thinking that it actually results in the toolpath more closely following the commanded path.

For example, it is currently believed that "ringing" is the result of the toolhead not following the "stepper path":

ringing

And thus commanding the steppers to take a deviated path could result in an improved overall toolhead path:

ringing-avoid

This code is an experiment.  It defines new `spring_x`, `spring_x_smooth_time`, `spring_y`, `spring_y_smooth_time` config settings that define the new s-curves.  The duration of the s-curve is controlled by the smooth_time settings.  The amount of deviation is determined by the spring parameters and the amount of estimated acceleration.  (The code uses estimated acceleration instead of the nominal acceleration, and thus does not have the "double acceleration" type issues with the bezier s-curve implementation.)  Different s-curves may be used for x and y axes.

There are several caveats with this code:
- it only works on cartesian printers
- it does not work with dual_carriage or multiple x/y/z stepper rails
- it is based on the PR #1997 code and thus has all of its limitations
- likely other bugs and limitations

I was able to tune this on my Makergear M2 printer (a "bed slinger").  The results were a noticeable, but small improvement.  I suspect an improved calibration routine may yield better results.  The calibration method I used was: print a tuning tower and measure the resonance, set the smooth_time parameters based on 1/4th the resonance time, print a tuning tower with changing spring value for each level (eg, `TUNING_TOWER COMMAND=SET_SPRING PARAMETER=SPRING_Y START=0 FACTOR=.000010`, measure the point of least resonance, and finally test on typical xyz calibration cubes.  

-Kevin

PR #1776, opened 2019-07-01
Add support for s-curve acceleration

KevinOConnor:
This is a pull request for the work-scurve-20180620 branch that has been under discussion in issue #57 for some time.  This branch has last been rebased on 20190701.

To test with this branch ssh into the host machine and run:
```
cd ~/klipper ; git fetch ; git checkout origin/work-scurve-20180620 ; sudo service klipper restart
```
You will then need to update the config file to specify an acceleration_order of 4 or 6.  (Alternatively, one may enable via the ACCEL_ORDER parameter in the SET_VELOCITY_LIMIT command.)

To the best of my knowledge, this support works as intended.  However, more people have reported a degradation in print quality (or no change) than people that have reported an improvement in quality.  It is believed that fundamentally s-curves improve performance at very low acceleration (eg, 500mm/s^2) and actually make quality worse at normal/high acceleration values (eg, 1000+ mm/s^2).

There is no immediate plan to merge this support into the master Klipper branch.  In order to merge the support, I believe comprehensive documentation would need to be written that describes how to test if s-curves improve quality on a printer.  I fear that without that documentation users may enable s-curves (because they're "better"), obtain poor results, and it would reflect poorly on Klipper.

To comment on this branch (or s-curves in general), please use the existing issue #57.

-Kevin

PR #4661, opened 2021-09-06
cartesian: Initial implementation of independent X/Y accel/velocity

rdbahm:
Extend the existing Z acceleration/velocity limits to work on the other axes as well, allowing the user to independently control the maximum acceleration and velocity of each axis. This was driven in large part by a desire to push my bed slinger further - testing on a resonance tower showed that X acceleration could be pushed much further than Y acceleration. While Input Shaper can largely deal with the non-uniform axes, being able to tune acceleration on each axis individually should allow users to achieve greater speeds without sacrificing quality. 

Testing has shown that it functions as expected, allowing independent accelerations on each axis. I think the biggest challenge will be seeing if users are able to make good use of the option.

I targeted this initial implementation at the cartesian kinematic since that covers the common bed slinger design, which is likely to have the most asymmetrical acceleration constraints. From a glance at the code, it looks like it would be relatively trivial to implement in most other non-delta kinematics - CoreXY, CoreXZ, Hybrid CoreXY, andd Hybrid CoreXZ all use the same (or nearly the same?) check_move code.

New parameters have been added to the docs.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -202,6 +202,24 @@ available parameters.
 ```
 [printer]
 kinematics: cartesian
+max_x_velocity:
+#   This sets the maximum velocity (in mm/s) of movement along the x
+#   axis. This setting can be used to restrict the maximum speed of
+#   the x stepper motor. The default is to use max_velocity for
+#   max_x_velocity.
+max_x_accel:
+#   This sets the maximum acceleration (in mm/s^2) of movement along
+#   the x axis. It limits the acceleration of the x stepper motor. The
+#   default is to use max_accel for max_x_accel.
+max_y_velocity:
+#   This sets the maximum velocity (in mm/s) of movement along the y
+#   axis. This setting can be used to restrict the maximum speed of
+#   the y stepper motor. The default is to use max_velocity for
+#   max_y_velocity.
+max_y_accel:
+#   This sets the maximum acceleration (in mm/s^2) of movement along
+#   the y axis. It limits the acceleration of the y stepper motor. The
+#   default is to use max_accel for max_y_accel.
 max_z_velocity:
 #   This sets the maximum velocity (in mm/s) of movement along the z
 #   axis. This setting can be used to restrict the maximum speed of
--- a/klippy/kinematics/cartesian.py
+++ b/klippy/kinematics/cartesian.py
@@ -23,9 +23,19 @@ class CartKinematics:
                                             self._motor_off)
         # Setup boundary checks
         max_velocity, max_accel = toolhead.get_max_velocity()
-        self.max_z_velocity = config.getfloat('max_z_velocity', max_velocity,
+        self.max_velocities = [max_velocity]*3
+        self.max_accels = [max_accel]*3
+        self.max_velocities[0] = config.getfloat('max_x_velocity', max_velocity,
                                               above=0., maxval=max_velocity)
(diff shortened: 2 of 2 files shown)


Piezoid:
I've made some [figures for torque/velocity limiting on CoreXY](http://bl.ocks.org/Piezoid/raw/368e4ca48c65724e419cfb8198cfee0e/#CoreXY) (sorry for the bad english :grin:), and a [draft implementation](https://github.com/Piezoid/klipper/blob/work-peraxis/klippy/kinematics/limited_corexy.py).

Another [cartesian implementation](https://github.com/Piezoid/klipper/blob/work-peraxis/klippy/kinematics/cartesian_slinger.py).

KevinOConnor:
Thanks.  As high-level feedback, I think this change would need to come along with documentation describing how to tune these new variables and testing results from a variety of printers demonstrating its utility.

FWIW, in initial versions of Klipper (before even its first release) using separate acceleration/velocity limits was possible and I got poor results with it on the Makergear M2 bed-slinger that I owned.  It took me several days to figure out what was going on, and when I did finally track it down I concluded that the ability to specify independent xy acceleration/velocity was highly prone to misconfiguration.  Specifically, when there are separate limits, every diagonal move (and thus every curve) effectively has a different acceleration and velocity.  This results in surprising behavior - every move has a different flow rate, every move has a different pressure advance behavior, alternate layers have different infil print times, etc.  I also concluded that the critical printer interaction is between nozzle and print - the stepper motors typically have oodles of torque and tuning to them generally does not make sense.

I'm leery of adding config options that make it "very easy" to obtain poor quality.  (To wit, I'm leery of adding "landmine options".)  That said, if there is a significant number of users with printers that can improve quality with these options then that's great.  That is, showing a level of print quality in significantly less time, which can't be obtained without these options, would be a very convincing demonstration.

Thanks again,
-Kevin

rdbahm:
I can appreciate avoiding configuration footguns. While I had considered the possibility of diagonal moves having different velocities/accelerations at every different angle, I had not considered the effects of the non-uniform acceleration on other features, like pressure advance. Thoughtful documentation to demonstrate how to use this capability conservatively would be important.

This feature isn't a hill I need to die on, and I simply don't have the variety of printers required to prove if the concept has merit. I'll leave some info on my findings here in case someone in the future thinks that independent acceleration is the hill they would like to die on. :)

On my project printer (Geeetech A10, an Ender 3 clone, modded half to death), I found that the independent accelerations let me achieve an additional 500mm/s/s (5700 to 6200) on my X axis due to the heft of the Y axis with no impact to print quality. This was also nice because it prevents my printer from flinging things off tables when the Y axis gets going. It worked nicely in combination with accelerometer-measured resonance compensation, since RC outputs suggest max accelerations for each axis and algorithm. 

More comprehensive testing would be required - I was largely testing for functionality of the change, not for "is this a thing that people should do?" If you're working on independent axis accelerations in the future, reach out to me if you'd like someone to run some additional validations.

KevinOConnor:
Okay.  Just to be clear, I'm not against exploring this change; I do think more documentation and testing is necessary before merging though.

Just as context, the Klipper S-Curve acceleration branch went through discussions for a few years (eg, #2030, #1776, #57, #3026, etc.).  Some people saw improvement with the support, while some people saw a degradation in quality.  Ultimately we ended up merging input-shaping (#3025).  So, if this change can be shown to notably improve quality then that would be great.

In any case, thanks for working on this.
-Kevin

PR #4664, opened 2021-09-06
Added raw response functionality

chestwood96:
module: RESPOND, allow sending raw messages without prefix

Some octoprint plugins require quite specific messages, for example the telegram plugin requiring no space between "echo:" and the message or the bed visualizer plugin requiring no space between "echo:" and the message. 

This allows sending completely raw messages back to the host.

Signed-off-by: Adrian Joachim <adi.joachim12@gmail.com>

### DIFF
--- a/klippy/extras/respond.py
+++ b/klippy/extras/respond.py
@@ -38,17 +38,21 @@ class HostResponder:
     def cmd_RESPOND(self, gcmd):
         respond_type = gcmd.get('TYPE', None)
         prefix = self.default_prefix
-        if(respond_type != None):
-            respond_type = respond_type.lower()
-            if(respond_type in respond_types):
-                prefix = respond_types[respond_type]
-            else:
-                raise gcmd.error(
-                    "RESPOND TYPE '%s' is invalid. Must be one"
-                    " of 'echo', 'command', or 'error'" % (respond_type,))
-        prefix = gcmd.get('PREFIX', prefix)
-        msg = gcmd.get('MSG', '')
-        gcmd.respond_raw("%s %s" % (prefix, msg))
+        raw = gcmd.get('RAW', None)
+        if raw:
+          gcmd.respond_raw(raw)
+        else:
+            if(respond_type != None):
+                respond_type = respond_type.lower()
+                if(respond_type in respond_types):
+                    prefix = respond_types[respond_type]
+                else:
+                    raise gcmd.error(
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  My high-level feedback is that I'm not sure this is a good candidate for the master Klipper repo.  I'm leery of making it possible for macros to respond with arbitrary responses because of the possibility of it confusing the g-code protocol (in particular, sending an "ok" message or similar).  This is the reason that the current respond module explicitly checks that the prefix is in a limited set of, generally safe, options.

Also, as a general trend, we want to move interfacing functionality to the Klipper API Server - https://www.klipper3d.org/API_Server.html - and avoid complex g-code parsing.

-Kevin

chestwood96:
While I don't personally agree i can understand your reservations, however the main goal of this PR is to enable mainline klipper to work with some common Octoprint plugins that do some imo pretty sloppy gcode parsing. The telegram plugin soes a contains check so I can just put the whole "echo:something" as message of an echo message which is kind of janky but works, other plugins on the other hand match the whole line which makes it impossible to interact witch some features from klipper. The api server looks pretty cool but imo it's a bit overkill to have to fork every octoprint plugin that you want to interact with.

I initially made just a toggle for space/no space but then thought why I should just not let the user send whatever they want, would something like this fit mainline better? (either space enable/disable or just a variable for the spacer defaulting to " ") .

There is of course also the option of just removing the space entirely to bring the formating inline with marlin but that would change (and potentially break) the output from existing gcode macros so I am not a huge fan of this way.

KevinOConnor:
I don't see a harm in adding support for a new prefix with "echo:" and no space.

Looking closer though, it looks like the respond module already supports arbitrary prefixes.  Not sure that was intended - @alecplumb - do you know if that was the original intent?

-Kevin

chestwood96:
Arbitrary prefixes are a thing but they do include the forced space between preix and msg so it's not possible to get rid of the space with the existing setup.

`gcmd.respond_raw("%s %s" % (prefix, msg))`

I did considder removing the forced space and just adding it to the default prefixes but that would still affect pepole that allready use custom prefixes.

chestwood96:
The existing predefined echo is already without space, the issue is the space in the template that puts the prefix and msg together.

If I just move the space to the predefined prefixes and take it out of the templates that would cause that space to be missing for people that already use arbitrary prefixes. If that is acceptable that would be an easy fix, if not I have no idea how to cleanly fix this.

PR #3912, opened 2021-02-10
WIP stm32h7 port, "Rescheduled timer in the past" error

D4SK:
I'm currently trying to port klipper to stm32h750 mcus and running into a "Rescheduled timer in the past" error.

I think I have most of it implemented, whats missing is SPI, USB and ADC control. I haven't looked much into the timer and irq handeling aspect, so this could be the problem. All the abstractions armcm_timer needs are provided by the cmsis library, and startup/communication works fine, so not sure where to start there. timer_read_time seems correct as well.

[klippy.log](https://github.com/KevinOConnor/klipper/files/5959837/klippy.1.log)
[klippy_shutdown00037.log](https://github.com/KevinOConnor/klipper/files/5959838/klippy_shutdown00037.log)

I already tried running at 200mhz instead of 400, and increasing the timeout in timer_dispatch_many from 1000us to 1000000us, both created the same error.
When connecting with console.py get_config returns is_shutdown=1

What could cause this, are there any other irregularities in the log I should look at?
Other than the clocks, are all times in the log in seconds?

PR #4666, opened 2021-09-07
stm32h7 port - set adc to 12bit mode

D4SK:
This is a continuation of work on #3912
 
Set the adc to 12 bit mode, as promised.
I also merged the work-stm32h7-20210902 branch into my fork, and everything still works. So the changes should be fine (I cant easily test the work-stm32h7-20210902 branch without reinstalling on python2).


KevinOConnor:
Thanks - I merged this into the work-stm32h7-20210902 branch.  We'll target merging of that branch post v0.10.0 release.

-Kevin

Issue #4427, opened 2021-06-26
Bed temp overshoots and never recovers


RafaelSchridi:
Me and some other users on the Artillery Discord have been trying to solve the problem of the bed temperature overshooting and basically "soft locking" the print because it never goes back down.

(manually stopped here)
![firefox_2021-06-26_01-03-52](https://user-images.githubusercontent.com/11460682/123522957-0ff9e680-d6c1-11eb-9b0f-fbd99da4b32c.png)

When I use the PID values I still have of my marlin installation it also overshoots but swiftly recovers.
![image](https://user-images.githubusercontent.com/11460682/123523045-7ed73f80-d6c1-11eb-9498-555857334be6.png)

The values produced by each PID tune are also VERY different as seen here.
![firefox_2021-06-26_16-43-13](https://user-images.githubusercontent.com/11460682/123523060-a201ef00-d6c1-11eb-96b6-5a35ebe0cd68.png)

tanaes:
Chiming in, this seems to be a pretty common problem among users on the Artillery Discord, and in most cases, replacing the Klipper-tuned PID values with their previous Marlin-tuned values gets rid of the problem.

At a bit of a loss as to why this might be. The only meaningful difference I can find in the code between the two tuning loops is that the Marlin code has a Kp adjustment factor that changes between hotend (0.6) and bed (0.2) tuning, where Klipper seems to use 0.6 for all. (Caveat: I am easily lost in C code!)

If the problem really is an overestimated Kp due to asymmetric temperature response (as suggested in the aforementioned PR), is it possible that this (presumably empirically-derived) correction is worth implementing?

I'll try to get some other users who've experienced this to contribute data, and if anyone has a suggestion as to a preferred style of solution I can try whipping up a branch with it and have folks test it out.

KevinOConnor:
Interesting.  Klipper uses the basic Ziegler–Nichols "classic PID" (see https://en.wikipedia.org/wiki/Ziegler%E2%80%93Nichols_method).  It looks like in 2019 Marlin changed the bed calibration to use Ziegler–Nichols "no overshoot" (Marlin commit 1db7013e).

I'd be reluctant to hard-code the bed to use a different algorithm, but I can see where using "no overshoot" would be useful for heaters that heat much faster than they cool.  So, I agree Klipper could be improved here.

[...]

PR #4667, opened 2021-09-07
adding two additional pid control algorithms:  no_overshoot and ponm …

bjones14:
http://brettbeauregard.com/blog/2017/06/introducing-proportional-on-measurement/

https://github.com/KevinOConnor/klipper/issues/4427

### DIFF
--- a/klippy/extras/heaters.py
+++ b/klippy/extras/heaters.py
@@ -43,7 +43,10 @@ class Heater:
         self.next_pwm_time = 0.
         self.last_pwm_value = 0.
         # Setup control algorithm sub-class
-        algos = {'watermark': ControlBangBang, 'pid': ControlPID}
+        algos = {'watermark': ControlBangBang, 
+                 'pid': ControlPID,
+                 'pid_no_overshoot': ControlPIDNoOvershoot,
+                 'pid_ponm': ControlPIDPonM}
         algo = config.getchoice('control', algos)
         self.control = algo(self, config)
         # Setup output heater pin
@@ -166,7 +169,7 @@ class ControlBangBang:
 
 
 ######################################################################
-# Proportional Integral Derivative (PID) control algo
+# Proportional Integral Derivative (PID) control algo (Classic)
 ######################################################################
 
 PID_SETTLE_DELTA = 1.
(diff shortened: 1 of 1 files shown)


susisstrolch:
Doesn't work at all...
Steps to reproduce:
1. removed the pid sections from printer.cfg (SAVE_CONFIG)
2. set 'control: pid_ponm' for both extruder and heatbed
3. run PID_TUNE for extruder and heatbed

after running PID_TUNE 'control' was reset to 'pid' (by klipper).
So I manually set 'control' to 'ponm' for both extruder and heatbed (in the SAVE_CONFIG section of printer.cfg), restarted firmware and klipper and and tried heating with the new algorith.
Heating simply doesn't work - power stays at 0% - and klipper errors out.

KevinOConnor:
Thanks.  I'm not sure I understand the "big picture" of this PR.  If a user wanted to use these methods, couldn't they just tweak their pid_Kp, pid_Ki, pid_Kd settings in the config?  What's the gain of altering the code?

-Kevin

bjones14:
> Thanks. I'm not sure I understand the "big picture" of this PR. If a user wanted to use these methods, couldn't they just tweak their pid_Kp, pid_Ki, pid_Kd settings in the config? What's the gain of altering the code?
> 
> -Kevin

The big picture is the proportional on measurement PID method - it changes how the math is done in the PID algorithm - though you are correct in that the only different between the original two are the scaling/gains on the PID terms so it doesn't matter as much there - it was more for adding additional PID control types than anything, key one being PonM.

bjones14:
> Doesn't work at all...
> Steps to reproduce:
> 
> 1. removed the pid sections from printer.cfg (SAVE_CONFIG)
> 2. set 'control: pid_ponm' for both extruder and heatbed
> 3. run PID_TUNE for extruder and heatbed
> 
> after running PID_TUNE 'control' was reset to 'pid' (by klipper).
> So I manually set 'control' to 'ponm' for both extruder and heatbed (in the SAVE_CONFIG section of printer.cfg), restarted firmware and klipper and and tried heating with the new algorith.
> Heating simply doesn't work - power stays at 0% - and klipper errors out.
> 
> General note:
> I'm missing doku about the usage and hints how to tune the ponm with klipper.

Running PID auto tune will run through the standard Z-N PID variables which will not be correct for PonM control - there is not an auto-tune function in place for this control type.

More documentation/info on PonM can be found here - http://brettbeauregard.com/blog/2017/06/introducing-proportional-on-measurement/

KevinOConnor:
Thanks.  As high-level feedback, a merge of PONM support would need come after test results on a variety of printers demonstrating its utility.  I think key to obtaining that wouId be user documentation explaining how and when to utilize PONM.

So, if there is interest in pursuing this, I'd say the next steps would be to try and find testers and come up with a procedure that testers should run to quantify the overall impact.

FWIW, I'd be surprised that proportional on measurement would work well in a 3d printer as the amount of power required to heat to a particular temperature is not linear.  For example, heating to 120 degrees is not twice the power needed to heat to 60 degrees and not half the power needed to heat to 240 degrees.  Higher temperatures need significantly higher power to maintain.  Happy to be shown to be wrong though.

Cheers,
-Kevin

bjones14:
> Thanks. As high-level feedback, a merge of PONM support would come after test results on a variety of printers demonstrating its utility. I think key to obtaining that wouId be user documentation explaining how and when to utilize PONM.
> 
> So, if there is interest in pursuing this, I'd say the next steps would be to try and find testers and come up with a procedure that testers should run to quantify the overall impact.
> 
> FWIW, I'd be surprised that proportional on measurement would work well in a 3d printer as the amount of power required to heat to a particular temperature is not linear. For example, heating to 120 degrees is not twice the power needed to heat to 60 degrees and not half the power needed to heat to 240 degrees. Higher temperatures need significantly higher power to maintain. Happy to be shown to be wrong though.
> 
> Cheers,
> -Kevin

I think that makes perfect sense, I'll work to provide and gather more information and tests to verify the impact.

You pointed out the exact reason why PonM is good - it's an integrating process.  That means that the heater output controls the rate of temperature change, but there is not a linear relationship to X power = X temperature, as we are always adding heat as it is taken away by ambient air, cooling fans, etc.

acwest:
> Running PID auto tune will run through the standard Z-N PID variables which will not be correct for PonM control - there is not an auto-tune function in place for this control type.
> ... 
> More documentation/info on PonM can be found here - http://brettbeauregard.com/blog/2017/06/introducing-proportional-on-measurement/

Can you document HOW to determine the correct values for the PonM case, if autotune won't work?

jwkicklighter:
If #4427 is definitely waiting on this ticket, can we open it back up?

Issue #3496, opened 2020-11-02
Support for load cell based probe


mhier:
I am in the process of writing the support for my 3D printer (Renkforce RF1000) which has a special type of bed level probes based on a load cell to measure the force when the nozzle touches the build plate. I am wondering what the best approach is to integrate this. I find the modular approach of the Klipper firmware very appealing and I would like to embrace it as much as possible. This is what I got so far:

I have written a module to support the I2C-connected ADC for measuring the load cell signal:
https://github.com/RF1000community/klipper/blob/master/klippy/extras/ads1100.py

This already works, I can add the ADC as pin for a a fake temperature sensor to get it into the system. Then I see the debug log messages printing, and the query adc command works as well.

Next I was thinking to write another implementation of the PrinterProbe class and register it as a "probe" object, so it is found by the ProbePointsHelper. That part (finding the object) also seems to work.

Now I wonder I best interface the new PrinterProbe implementation with the ADC module without breaking the abstraction. My idea was to register a callback function (via setup_adc_callback) to retrieve new samples when arriving and process them inside the run_probe() function. This doesn't seem to work though, because the callback function is no longer serviced while staying inside run_probe() - presumably because both would run in the same thread.

Is there another generic method to obtain the ADC samples (preferably with a timing based on the ADC sampling time)? If possible I would like to avoid introducing a private interface between the new PrinterProbe implementation class and the ADC module, because this would prevent to use another ADC type with the probe implementation.

PS: I am intending to send pull requests with the printer support once it is "complete". Is there any policy how do to this? Individual pull requests for each feature/module, or better one big pull request with everything? The printer support also involves a new stepper driver module which will even require modifications to the SPI code (it expects an inverted CS signal... so far I just hacked this...).

klipper-gitissuebot:
Hi @mhier,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

Unfortunately, too many people have opened tickets without providing the log. That consumes developer time; time that would be better spent enhancing the software. If this ticket references an event that has occurred while running the software then the Klipper log must be attached to this ticket. Otherwise, this ticket will be automatically closed in a few days.

For information on obtaining the Klipper log file see: https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md

The log can still be attached to this ticket - just add a comment and attach the log to that comment.

Best regards,
[...]

PR #4670, opened 2021-09-07
z_endstop_set_calibration: Calibrate endstop based on current position

mhier:
Add new module which allows to calibrate the z-axis endstop position
based on the current position. This can be used with load cell-based
or accelerometer-based probes, as discussed in #3496.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -402,6 +402,15 @@ see the [BL-Touch guide](BLTouch.md)):
 - `BLTOUCH_STORE MODE=<output_mode>`: This stores an output mode in the
   EEPROM of a BLTouch V3.1 Available output_modes are: `5V`, `OD`
 
+
+### Z Endstop Calibration by Current Position
+
+The following commands are available when the z_endstop_set_calibration config
+section (which has no further options) is enabled:
+- `Z_ENDSTOP_SET_CALIBRATION Z=<value>`: This command will change the
+  `position_endstop` value for the Z stepper such that the current Z position
+  will equal the given Z coordinate. Requires a `SAVE_CONFIG` to take effect.
+
 ### Delta Calibration
 
 The following commands are available when the
--- a/klippy/extras/z_endstop_set_calibration.py
+++ b/klippy/extras/z_endstop_set_calibration.py
@@ -1,0 +1,39 @@
+# Allow self-calibration of Z endstop using a Z probe.
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+
+class ZEndstopSetCalibration:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        gcode = self.printer.lookup_object('gcode')
+        gcode.register_command('Z_ENDSTOP_SET_CALIBRATION',
+                               self.cmd_Z_ENDSTOP_SET_CALIBRATION,
+                               desc=self.cmd_Z_ENDSTOP_SET_CALIBRATION_help)
+
+        zconfig = config.getsection('stepper_z')
+        self.z_position_endstop = zconfig.getfloat('position_endstop', None,
+                                                   note_valid=False)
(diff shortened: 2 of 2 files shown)


KevinOConnor:
Thanks.  What functionality does this provide over the new Z_OFFSET_APPLY_ENDSTOP command?  That is, could a user interested in this functionality implement it via a macro and Z_OFFSET_APPLY_ENDSTOP?

-Kevin

mhier:
Honestly, I wasn't aware of the new `Z_OFFSET_APPLY_ENDSTOP` (was likely introduced after I wrote this code - it just took me too long to create this PR, sorry... :-)). It looks like both are doing pretty much the same. What worries me a bit is that `Z_OFFSET_APPLY_ENDSTOP` is part of the manual probe module - while I need to use it with the load cell based probe.

Can I use this without the manual_probe? Or do I need to move it into a separate module for this?

KevinOConnor:
The manual_probe module is automatically loaded - its commands are always available.

-Kevin

mhier:
> The manual_probe module is automatically loaded - its commands are always available.
> 
> -Kevin

I see now one difference: `Z_OFFSET_APPLY_ENDSTOP` uses the "gcode homing", while `Z_ENDSTOP_SET_CALIBRATION` uses the current position. I do not see how I can set the gcode homing based on the current position. `G92` does not change the gcode homing, but the gcode base. `SET_GCODE_OFFSET` sets the gcode homing, but not based on the current position.

I tried setting the homing_origin from a gcode macro, like this:
```
[gcode_macro SET_GCODE_Z_ORIGIN_FROM_CURRENT_POSITION]
gcode:
    {% set printer.gcode_move.homing_origin.z = printer.toolhead.position.z %}
```
but this results in a syntax error... Is there any other way to achieve this without code change?

If not, I see two choices:
- extend `SET_GCODE_OFFSET` command such that it can optionally work relative to the current position, e.g. by introducing `X_RELATIVE`, `Y_RELATIVE`, `Z_RELATIVE` parameters, or
- keep the pull request as is.

My feeling is, the first approach (extend `SET_GCODE_OFFSET`) seems more elegant. What do you think?

KevinOConnor:
Well, you don't want to use the current position though, because with the latest code it is possible to overshoot slightly during a probe attempt.  Best is to extract the probe position as reported by `{printer.probe.last_z_result}`.

Setting the Z offset should be possible with something like: `SET_GCODE_OFFSET Z={printer.probe.last_z_result}`  This is totally untested though - it might need adjustment based on the current probe Z offset.

-Kevin

mhier:
Ok I understand. My probe algorithm currently moves to the determined Z=0 position to avoid this kind of overshoot. It can (should) also set `last_z_result`, then this approach can work. I will test this this evening, and then likely close this PR.

(The load-cell probe does not have any offsets, so this shall work without issues.)

mhier:
Ok, it works, but not as easy as I thought:
- I do not know how to alter `{printer.probe.last_z_result}`, so I introduced my own variable which I can access via `{printer["load_cell_probe"].last_z_result}`.
- Since the macro is parsed completely before executing, I have to introduce a helper macro "z_offset_from_probe_result" to set the z offset after running the probe in the same macro.

So my macro looks like this now:
```
[gcode_macro z_offset_from_probe_result]
gcode:
    SET_GCODE_OFFSET Z={printer["load_cell_probe"].last_z_result}

[gcode_macro bed_calibrate]
gcode:
    G0 Z10
    G0 X100 Y100
    G0 Z2
    PROBE
    Z_OFFSET_FROM_PROBE_RESULT
    Z_OFFSET_APPLY_ENDSTOP
    GET_POSITION
    SET_KINEMATIC_POSITION Z=0
    GET_POSITION
    G0 Z2
    BED_MESH_CALIBRATE
    SAVE_CONFIG
```

I think this is acceptable, hence I close this PR (and prepare the next one ;-)).

PR #4673, opened 2021-09-09
Update generic-bigtreetech-octopus.cfg

ShohninDmitriy:
Fix Pins EXP2


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md .  However, I can make this fix using my signed-off-by line if you prefer.

-Kevin

ShohninDmitriy:
Make it under your signature, it's not particularly important to me, the main thing is to let them use it with a full configuration

KevinOConnor:
Thanks.  Commit cd864488.

-Kevin

PR #4677, opened 2021-09-10
Document motor_power pin

truckershitch:
module: config/generic-bigtreetech-skr-2.cfg

This is just a short inline comment about Bigtreetech's implementation of the "anti-reversal stepper protection" and the way it should be handled in the Klipper printer configuration.

This was discussed in the Klipper discord around 8 to 9 AM UTC September 10, 2021.

Signed-off-by: Jerry Chapman <truckershitch@hambone.e4ward.com>


KevinOConnor:
Thanks.

-Kevin

PR #4681, opened 2021-09-11
External ADC via I2C Module

dawidmpunkt:
module: klippy/extras/i2c_adc.py

A module for external ADCs (ADS1013 to ADS1015 and MCP3421 to MCP3428) via I2C (and a rPi). 
Voltage can be read via a virtual chip/pin.

PR for documentation:
https://github.com/Klipper3d/klipper/pull/4762
PR for Config_Reference:
https://github.com/Klipper3d/klipper/pull/4756


dawidmpunkt:
I made some mistakes while trying to create a PR for this module. Did it work?

KevinOConnor:
Thanks.  In general it looks fine to me, but I do have some high-level comments:
1. This is something that we'd look to merge after the v0.10.0 release.
2. A new module would require the corresponding documentation update to docs/Config_Reference.md .
3. I'm confused what the gain, rate, resolution, and report_time parameters are used for.  Shouldn't the code know what these need to be set to based on the type of chip?
4. I noticed that `eval()` is used in the code - I'm not sure what it is being used for, but `eval()` has a number of security implications.  Some method other than using `eval()` will need to be found.
5. It looks like the code is reading a min_temp and max_temp settings, but is not verifying those values.  The associated checks and shutdown tests would be needed, as not implementing those safety checks could result in an unsafe condition.

-Kevin

dawidmpunkt:
@KevinOConnor:
1: sounds good
3: depending on the voltage to be measured or depending on the adc-chip, the parameters gain (amplifies/reduces the measured voltage to get an adequate reading within e.g. a 12 bit number), rate (according to the manual, the more time the chip has for the sampling, the more precise the measurement is), resolution (e.g. 12 bit, 14 bit) etc. different chips also support parameters (some only 12 bit some 18 bit etc.) can be adjusted. The module uses default values, but they can be changed if necessary. the report_time parameter gets passed from other modules (e.g. filament width sensor) via callback to the adc-module. here, it tells the adc-module how frequent it should sample voltage. since the other modules pass those parameter, i decided to include it here.
2, 4, 5: I will change/add those.

PR #4683, opened 2021-09-12
Add make clean before make as without build fails

kbehren:


KevinOConnor:
This looks like a dup of #4415 .  Also, separately, commits need to have a signed-off-by line - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

kbehren:
Ahh OK didn't search long enough for the duplicate it seems.

PR #4684, opened 2021-09-12
Add support for half-stepping encoders

rufo:
Adds support for half-stepping encoders (encoders that only emit two steps per detent, instead of four). This PR incorporates the feedback from [@susisstrolch's PR](https://github.com/KevinOConnor/klipper/pull/4202), which was itself built upon [a previous PR from @nickbrennan01](https://github.com/KevinOConnor/klipper/pull/730), but I also reworded the documentation a bit. It also uses [the `HALF_STEP` table](https://github.com/brianlow/Rotary/blob/6b784cca67c5f1ce5e11d757a540fc4c0311efca/Rotary.cpp#L21-L40) from the Rotary Arduino library linked in buttons.py, so I presume using it here won't be a problem.

I'm not sure _why_ my Fysetc display is special this way, but in researching a solution, it seems like a few other people would have found this handy over the years. I'm only starting to dabble in Python, so I'd appreciate feedback from anyone on if there's anything that goes against conventions or should be tweaked. Thanks!


KevinOConnor:
Thanks.  In general, it looks fine to me.  I do have a few minor comments:
1. Out of an abundance of caution, I'd be inclined to look at merging this after the v0.10.0 release.
2. Very minor, but I don't think `pass` is needed in either of the sub-classes.
3. Very minor, but multiple definitions of R_START, R_DIR_CW, R_DIR_CCW, R_DIR_MSK seems unnecessary and a bit confusing to me.

-Kevin

KevinOConnor:
Oh, also as another very minor comment, it's not necessary to lookup the pins module to raise an error - just use `self.printer.config_error(...)`.

For `config.getchoice()` it is now preferable to use integer keys when the value is inherently an integer - this was done in commit 84ac5b01.

-Kevin

rufo:
Thanks, that all makes sense. The variable duplication in particular is a good catch - I'd intended to fix that, but when I deleted `R_START`/etc. in the subclasses the IDE hinted those would be undefined, and figuring out a better solution slipped my mind in the midst of other cleanup. I'll figure out a way to fix that (probably by referring to `BaseRotaryEncoder` directly, unless changing `ENCODER_STATES` to a method would be better) and take care of the other points you mentioned.

rufo:
Just got back around to this and pushed some updates that should address this feedback. Let me know if any more changes are needed.

PR #4686, opened 2021-09-13
mcu: Add support for RTS MCU restart method.

Quantum-cross:
mcu: Add support for RTS MCU restart method.

Summary
------------
This lets you use RTS line on serial to reboot the mcu.
This has been tested on a pi3 and SKR Mini v2 installed in a Voron 0.

Motivation
-------------
I have connected my Pi to my MCU via the Pi's UART0 because it makes the wiring much more compact than a bulky USB cable, I also feel that the Pi and STM have perfectly good serial interfaces, and adding a USB interface in the middle is just complicating things.

I had a lot of errors until I realized that I had to set `restart_method` to `command` for the board to properly reset. I am uncomfortable using this method though, as in the case with a complete firmware lockup, it will not receive the restart command. 
A better solution would be to pulse the MCU nRST line.

As the raspberry pi UART does not contain a hardware DTR pin, I decided to use RTS, as it doesn't seem to be used for anything else, and it should be compatible with other klipper hosts that implement hardware RTS to pySerial.

Effects
---------
When `restart_method` is set to `rts` in your klipper config, the RTS serial line will remain high during normal operation. During a `FIRMWARE_RESTART`, RTS will be pulled low momentarily, in the same manner as `arduino` restart will pull DTR low.

Wiring/implementation
-----------------------------
Set `restart_method` to `rts` in klipper config.

On the RPi3, run:
`raspi-gpio set 17 a3`
This sets GPIO 17 as RTS0

Connect GPIO 17 (RPi header pin 11) to nRST on SKR Mini v2. I'm using the nRST pin exposed on the SWD port.

After running `FIRMWARE_RESET` g-code command, the board will probably reboot and klipper host will correctly re-establish connection!

---------------------------

Signed-off-by: Robert Cross <quantumcross@gmail.com>

### DIFF
--- a/klippy/mcu.py
+++ b/klippy/mcu.py
@@ -542,7 +542,8 @@ class MCU:
                     or self._serialport.startswith("/tmp/klipper_host_")):
                 self._baud = config.getint('baud', 250000, minval=2400)
         # Restarts
-        restart_methods = [None, 'arduino', 'cheetah', 'command', 'rpi_usb']
+        restart_methods = [None, 'arduino', 'cheetah', 'command', 'rpi_usb',
+                           'rts']
         self._restart_method = 'command'
         if self._baud:
             rmethods = {m: m for m in restart_methods}
@@ -758,7 +759,8 @@ class MCU:
                 elif self._baud:
                     # Cheetah boards require RTS to be deasserted
                     # else a reset will trigger the built-in bootloader.
-                    rts = (resmeth != "cheetah")
+                    # Also deassert RTS if using RTS as a reset method.
+                    rts = not ((resmeth == "cheetah") or (resmeth == "rts"))
                     self._serial.connect_uart(self._serialport, self._baud, rts)
                 else:
                     self._serial.connect_pipe(self._serialport)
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Interesting, thanks.  However, I'm not sure this is currently a good match for the Klipper master repo, because I'm not sure there are enough users that would currently utilize this functionality.  As a general "rule of thumb", I generally target a minimum user size of 100 or so for any given code feature.  It's not clear to me if there are that many people with RTS wired to an mcu reset.

Also, if you do have RTS wired to reset, it may be possible to use Moonraker's power plugin to obtain similar results (but I'm not sure about that).

-Kevin

Quantum-cross:
Hm, fair enough. I just thought it was strange that there are 4 reset methods: 'arduino', 'cheetah', 'rpi_usb', 'command'.
The former three are only really useful for USB, but all perform a hard MCU reset.
The latter is the only option available for hardware serial on a raspberry pi (because it does not support DTR), but it requires the firmware to be responsive to the command.

Granted I have not had klipper firmware lock up on me, and I don't expect it to. I just like planning for the unexpected :)

PR #4688, opened 2021-09-13
extras: Smart Effector support

dmbutyugin:
This PR adds support of programmable sensitivity for Smart Effector, if the effector control pin is wired to a GPIO pin,
as well as a few other features useful in combination with the custom sensitivity: probing acceleration and recovery time.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1743,6 +1743,58 @@ control_pin:
 #   See the "probe" section for information on these parameters.
 ```
 
+### [smart_effector]
+Smart Effector. One may define this section instead of `[probe]` to enable
+the SmartEffector-specific features. This also enables
+[runtime commands](G-Codes.md#smart-effector) to adjust the parameters of
+the Smart Effector at run time.
+
+```
+[smart_effector]
+pin:
+#   Pin connected to the Smart Effector Z Probe output pin (pin 5). Note that
+#   pullup resistor on the board is generally not required. However, if the
+#   output pin is connected to the board pin with a pullup resistor, that
+#   resistor must be high value (e.g. 10K Ohm or more). Some boards have a low
+#   value pullup resistor on the Z probe input, which will likely result in an
+#   always-triggered probe state. In this case, connect the Smart Effector to
+#   a different pin on the board. This parameter is required.
+# control_pin:
+#   Pin connected to the Smart Effector control input pin (pin 7). If provided,
+#   Smart Effector sensitivity programming commands become available.
+# probe_accel:
+#   If set, limits the acceleration of the probing moves (in mm/sec^2).
+#   A sudden large acceleration at the beginning of the probing move may
+#   cause spurious probe triggering, especially if the hotend is heavy.
+#   To prevent that, it may be necessary to reduce the acceleration of
+#   the probing moves via this parameter.
+# recovery_time: 0.4
+#   A delay between the travel moves and the probing moves in seconds. A fast
+#   travel move prior to probing may result in a spurious probe triggering.
+#   This may cause 'Probe triggered prior to movement' errors if no delay
+#   is set. Value 0 disables the recovery delay.
+#   Default value is 0.4.
+#x_offset:
+#y_offset:
+#   Should be left unset (or set to 0).
+z_offset:
+#   Trigger height of the probe. Start with -0.1 (mm), and adjust later using
+#   `PROBE_CALIBRATE` command. This parameter must be provided.
+#speed:
+#   Speed (in mm/s) of the Z axis when probing. It is recommended to start
+#   with the probing speed of 20 mm/s and adjust it as necessary to improve
+#   the accuracy and repeatability of the probe triggering.
+#samples:
+#sample_retract_dist:
+#samples_result:
+#samples_tolerance:
+#samples_tolerance_retries:
+#activate_gcode:
+#deactivate_gcode:
+#deactivate_on_each_sample:
+#   See the "probe" section for more information on the parameters above.
+```
+
 ## Additional stepper motors and extruders
 
 ### [stepper_z1]
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1053,6 +1053,34 @@ profile matching the supplied name from persistent memory. Note that
 after SAVE or REMOVE operations have been run the SAVE_CONFIG gcode
 must be run to make the changes to persistent memory permanent.
 
+### [smart_effector]
+
+Several commands are available when a
+[smart_effector config section](Config_Reference.md#smart-effector) is enabled.
+Be sure to check the official documentation for the Smart Effector on the
+[Duet3D Wiki](https://duet3d.dozuki.com/Wiki/Smart_effector_and_carriage_adapters_for_delta_printer)
+before changing the Smart Effector parameters. Also check the
+[probe calibration guide](Probe_Calibrate.md).
+
+#### SET_SMART_EFFECTOR
+`SET_SMART_EFFECTOR [SENSITIVITY=<sensitivity>] [ACCEL=<accel>]
+[RECOVERY_TIME=<time>]`: Set the Smart Effector parameters. When `SENSITIVITY`
+is specified, the respective value is written to the SmartEffector EEPROM
+(requires `control_pin` to be provided). Acceptable <sensitivity> values are
+0..255, the default is 50. Lower values require less nozzle contact force to
+trigger (but there is a higher risk of false triggering due to vibrations
+during probing), and higher values reduce false triggering (but require larger
+contact force to trigger). Since the sensitivity is written to EEPROM, it is
+preserved after the shutdown, and so it does not need to be configured on
+every printer startup. `ACCEL` and `RECOVERY_TIME` allow to override the
+corresponding parameters at run-time, see the
+[config section](Config_Reference.md#smart-effector) of Smart Effector for
+more info on those parameters.
+
+#### RESET_SMART_EFFECTOR
+`RESET_SMART_EFFECTOR`: Resets Smart Effector sensitivity to its factory
+settings. Requires `control_pin` to be provided in the config section.
+
 ### [stepper_enable]
 
 The stepper_enable module is automatically loaded.
--- a/klippy/extras/smart_effector.py
+++ b/klippy/extras/smart_effector.py
@@ -1,0 +1,154 @@
+# SmartEffector support
+#
+# Copyright (C) 2021  Dmitry Butyugin <dmbutyugin@google.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+import logging
+from . import probe
+
+# SmartEffector communication protocol implemented here originates from
+# https://github.com/Duet3D/SmartEffectorFirmware
+BITS_PER_SECOND = 1000.
+
+class ControlPinHelper:
+    def __init__(self, pin_params):
+        self._mcu = pin_params['chip']
+        self._pin = pin_params['pin']
+        self._start_value = self._invert = pin_params['invert']
+        self._oid = None
+        self._set_cmd = None
+        self._mcu.register_config_callback(self._build_config)
+    def _build_config(self):
+        self._mcu.request_move_queue_slot()
+        self._oid = self._mcu.create_oid()
+        self._mcu.add_config_cmd(
+            "config_digital_out oid=%d pin=%s value=%d default_value=%d"
+            " max_duration=%d" % (self._oid, self._pin, self._start_value,
+                                  self._start_value, 0))
+        cmd_queue = self._mcu.alloc_command_queue()
+        self._set_cmd = self._mcu.lookup_command(
+            "queue_digital_out oid=%c clock=%u on_ticks=%u", cq=cmd_queue)
+    def write_bits(self, start_time, bit_stream):
+        bit_step = 1. / BITS_PER_SECOND
+        last_value = self._start_value
+        bit_time = start_time
+        for b in bit_stream:
+            value = (not not b) ^ self._invert
+            if value != last_value:
+                clock = self._mcu.print_time_to_clock(bit_time)
+                self._set_cmd.send([self._oid, clock, value])
+                last_value = value
+            bit_time += bit_step
+        # After the last bit, the signal on the control pin must go back
+        # to its start value.
+        if value != self._start_value:
+            clock = self._mcu.print_time_to_clock(bit_time)
+            self._set_cmd.send([self._oid, clock, self._start_value])
+            bit_time += bit_step
+        return bit_time
+
+class SmartEffectorEndstopWrapper:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.gcode = self.printer.lookup_object('gcode')
+        self.probe_accel = config.getfloat('probe_accel', 0., minval=0.)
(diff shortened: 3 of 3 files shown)


dmbutyugin:
With this PR, I was able to get +/- 0.025 mm Z level accuracy within the delta calibration radius of 140 mm on a ~300 mm build plate diameter delta printer (after a few runs of `DELTA_CALIBRATE`). YMMV, of course.

The relevant part of the Smart Effector configuration on my printer (with a volcano hotend):
```
[smart_effector]
pin: PC4
control_pin: PC7
probe_accel: 400
recovery_time: 0.4
speed: 12
samples: 5
sample_retract_dist: 2
samples_tolerance: 0.0075
samples_tolerance_retries: 10
```
The Smart Effector sensitivity was programmed as 15 vs the default 50 (smaller values indicate smaller force to trigger - so 'higher' sensitivity):
`SET_SMART_EFFECTOR SENSITIVITY=15`

schnello:
Hello. Is this also useable with an Piezo Orion?

dmbutyugin:
@schnello If I understand correctly, Piezo Orion doesn't have programmable sensitivity feature, right? You could, in principle, use this for other features: `recovery_time` and `probe_acceleration` (you can define `smart_effector` section instead of `probe` for piezo orion too, `control_pin` and related functionality is optional). But you could also achieve those effects with `activate_gcode` and `deactivate_gcode` (e.g. G4 P400, though it gets more complicated for acceleration). In general, I think if there is sufficient interest/demand, `recovery_time` and `probe_acceleration` could be moved into the generic `probe` module too.

KevinOConnor:
Thanks.  I have some high-level questions and comments:
1. I think this would be something we'd look at merging after the Klipper v0.10.0 release.
2. This doesn't pass the test cases because it increases the size of the PRU code.  I have a high-level plan to free up space on the PRU, but that's likely some weeks/months out.  Can you give a high-level overview of what the new mcu code is intended for?
3. I'm a little confused on the copyrights in the new smart_effector.py file.  If David Crocker has a copyright to that file I think it would be preferable to state that unambiguously; conversely if the comments are just of historical interest I think we shouldn't mix it with the file copyrights.

-Kevin

dmbutyugin:
@KevinOConnor, thanks!


> Thanks. I have some high-level questions and comments:
> 
>     1. I think this would be something we'd look at merging after the Klipper v0.10.0 release.

i think it's reasonable, and I'm OK with that. I still wanted to get the feedback on the approach I chose (both on the user-facing API and the MCU code) so that I could make adjustments if necessary.
 

>     2. This doesn't pass the test cases because it increases the size of the PRU code.  I have a high-level plan to free up space on the PRU, but that's likely some weeks/months out.  Can you give a high-level overview of what the new mcu code is intended for?

Yes, I've noticed that. Sadly, I wasn't able to deal with it yet because I had some troubles building pru-gcc, and it would be a pain to try to fix it without this cross-compilation toolchain working.

The new MCU function allows to write a short sequence of bits into the digital pin output with a specified frequency (toggling it high and low following the bit pattern, in other words). As I previously asked you on Discord, I had timer scheduling issues when trying to toggle the digital pin at 1 kHz using MCU_digital_out.set_digital method. This 1kHz is the required communication frequency by the Smart Effector.

>     3. I'm a little confused on the copyrights in the new smart_effector.py file.  If David Crocker has a copyright to that file I think it would be preferable to state that unambiguously; conversely if the comments are just of historical interest I think we shouldn't mix it with the file copyrights.

Hmm, I see. To be clear, none of the code from RRF or SmartEffector firmware was used in my implementation, you could see for instance
https://github.com/Duet3D/RepRapFirmware/blob/master/src/Endstops/LocalZProbe.cpp#L169-L226
Only the description of the communication protocol from there and from the firmware itself:
https://github.com/Duet3D/SmartEffectorFirmware/blob/master/src-v1/main.cpp#L386-L395

So the Copyright only applies to the RRF and the SmartEffectorFirmware really. Maybe I should just keep a link to, say, SmartEffectorFirmware, stating that the implementation follows the protocol described there?

KevinOConnor:
> The new MCU function allows to write a short sequence of bits into the digital pin output with a specified frequency (toggling it high and low following the bit pattern, in other words).

Ah, okay.  Another goofy bit-banging protocol.

As high-level feedback, I'd avoid modifying gpiocmds.c for this.  The general pattern is to add a new c file for each bit-banging protocol that we need to support (eg, see src/tmcuart.c).  That means creating a new oid specific config_XXX command, associated struct, and scheduling code.  Also, note that it's not valid to call move_queue_push() in the firmware without the host tracking it as well, as the host needs to ensure there is sufficient space available in the move_queue before sending commands that use it.  (Otherwise, it can result in "move queue overflow" errors.)

That said, it might be possible to reuse tmcuart.c for this purpose.  It already implements a fixed frequency bit transmission protocol.  You wont be able to reuse its host python code, but you may be able to call the existing `config_tmcuart` and `tmcuart_send` commands directly from your new module.

Note that this also avoids a conflict with the PRU build as the bit-banging protocol code isn't compiled in when building on the PRU.

> Maybe I should just keep a link to, say, SmartEffectorFirmware, stating that the implementation follows the protocol described there?

Okay.

Cheers,
-Kevin

tcj-lev:
Hi,
is there any chance to have this implemented soon?
I have just "klipperized" my Anycubic Predator with a Duet Maestro and Smart Effector.
And I am missing the feature to program the sensitivity.

dmbutyugin:
> Hi, is there any chance to have this implemented soon? I have just "klipperized" my Anycubic Predator with a Duet Maestro and Smart Effector. And I am missing the feature to program the sensitivity.

I looked at the possibility of using tmcuart module for this purpose. It is likely possible, but requires some work. As such, I don't expect to make it in the very near future.

However, the code itself is in working condition. Since the sensitivity setting persists in EEPROM, you only need to set the desired sensitivity once. Then, you have two options: either configure the required sensitivity before converting the printer to Klipper, or take this branch, tune the sensitivity and write to SmartEffector, then switch back to the mainline Klipper branch. Do the latter only if you are experienced enough to deal with git and branches and such.

tcj-lev:
> > Hi, is there any chance to have this implemented soon? I have just "klipperized" my Anycubic Predator with a Duet Maestro and Smart Effector. And I am missing the feature to program the sensitivity.
> 
> I looked at the possibility of using tmcuart module for this purpose. It is likely possible, but requires some work. As such, I don't expect to make it in the very near future.
> 
> However, the code itself is in working condition. Since the sensitivity setting persists in EEPROM, you only need to set the desired sensitivity once. Then, you have two options: either configure the required sensitivity before converting the printer to Klipper, or take this branch, tune the sensitivity and write to SmartEffector, then switch back to the mainline Klipper branch. Do the latter only if you are experienced enough to deal with git and branches and such.

That's not the answer that I was hoping for ;-)
But meanwhile I've found a working sensitivity with RRF and will soon "re-klippperize" the printer

KevinOConnor:
Okay, makes sense.  I'm not sure what you want to do with this PR though.  I agree getting tmcuart to work doesn't seem like it is worth the effort.  However, adding code to all the micro-controllers for this one quirky device also doesn't seem like a good long-term plan.

Separately, I'm not sure why this code does add new mcu code.  It seems the mcu code just does the equivalent of calling `queue_digital_out` many times.  Couldn't the host code just do that instead?  (There would still be the issue of overflowing the move queue, but if the host called `toolhead.wait_moves()` before and after each transmission then I'd guess it would be unlikely to cause an issue in practice.)

So, how do you want to proceed?

-Kevin

dmbutyugin:
> However, adding code to all the micro-controllers for this one quirky device also doesn't seem like a good long-term plan.

That's true. I thought that having the possibility to write an arbitrary sequence of bits to the output port might be helpful for other devices, but maybe it isn't.


> Separately, I'm not sure why this code does add new mcu code. It seems the mcu code just does the equivalent of calling `queue_digital_out` many times. Couldn't the host code just do that instead? (There would still be the issue of overflowing the move queue, but if the host called `toolhead.wait_moves()` before and after each transmission then I'd guess it would be unlikely to cause an issue in practice.)

I think I tried that previously, and ran into 'Timer too close' errors. I recall you mentioned on Discord that these commands are not buffered (on host?), and I suppose that writes of these commands at 1 kHz rate are unattainable. FWIW, I wasn't running these commands in parallel with anything else.


> So, how do you want to proceed?

In light of that, it might be worthwhile to modify tmcuart implementation. I could close this PR for now though until I have some code ready.

KevinOConnor:
> I think I tried that previously, and ran into 'Timer too close' errors. I recall you mentioned on Discord that these commands are not buffered (on host?), and I suppose that writes of these commands at 1 kHz rate are unattainable.

Hrmm.  I don't recall that conversation.  Could it have occurred before the mcu buffering of digital out commands was implemented?

As far as I know, it should now work - just need to make sure all the `queue_digital_out` commands are queued in advance on the mcu.  Something like (totally untested):
```python
...
def build_commands(self):
    self._mcu.add_config_cmd(
        "config_digital_out oid=%d pin=%s value=%d"
        " default_value=%d max_duration=%d"
        % (self._oid, self._pin, 0, 0, 0))
    cmd_queue = self._mcu.alloc_command_queue()
    self._set_cmd = self._mcu.lookup_command(
        "queue_digital_out oid=%c clock=%u on_ticks=%u", cq=cmd_queue)
...
def _send_bit_stream(self, bit_stream):
    # Convert stream to a long integer
    bits = 0
    for i, b in enumerate(bit_stream):
        bits |= b << (i * 8)
    num_bits = len(bit_stream) * 8
    # Send bits
    last_bit = 0x00
    bit_step = 1. / BITS_PER_SECOND
    toolhead = self.printer.lookup_object("toolhead")
    toolhead.wait_moves()
    start_time = bit_time = toolhead.get_last_move_time()
    for i in range(num_bits):
        bit_time += bit_step
        b = (bits >> i) & 0x01
        if b != last_bit:
            self._set_cmd.send([self._oid, bit_time, b])
            last_bit = b
    toolhead.dwell(bit_time - start_time)
    toolhead.wait_moves()
```

Basically, queue up a bunch of timed actions in the mcu, and the mcu should be able to run them with excellent precision.

-Kevin

dmbutyugin:
Thank you, Kevin!

> Hrmm. I don't recall that conversation. Could it have occurred before the mcu buffering of digital out commands was implemented?

Hmm, quite possible. It was back in August:
> Kevin
> — 
> 08/10/2021
Are you trying to send the writes from the host code?  If so, you can run into issues with comms - as digital updates aren't queued in the micro-controller right now.  For code running on the mcu itself, it should be possible to reach 100K+ pin updates per second.

> As far as I know, it should now work - just need to make sure all the queue_digital_out commands are queued in advance on the mcu. Something like (totally untested):

OK, I'll look up my first version of the code. I actually had something implemented very close to your snippet (just without last_bit optimization IIRC). I'll give that a go again then, and if it works - then it will be even better.

KevinOConnor:
Ah, I didn't realize what you were trying to do.  It is true that the host MCU_digital_out abstraction will not queue updates in the mcu today (it would need to extend the move_queue buffer management which hasn't been implemented yet).  It is, however, possible to queue them manually as in the snippet above (you can't use the MCU_digital_out abstraction).  The support for queuing in the mcu was added with 9cdf9bb6.

-Kevin

dmbutyugin:
@KevinOConnor, OK, I've done what you've suggested, and it seems to work. I've verified on my printer that SmartEffector sensitivity can be programmed and reset back to factory settings.

@tcj-lev If you know how, you can test the code on your SmartEffector and see if it works. I will also make a post on Discourse for better visibility to invite interested people for testing.

bernhardberger:
What's the status of this? Any chance this PR can be rebased and conflicts resolved?

Would be a huge addition to Klipper for Delta users. Especially the recovery time is doing my head in at the moment, would be nice to have that configurable.

dmbutyugin:
@bernhardberger The code should be working, but I suppose it needs more testing by interested users. So far, only I have tested it and confirmed that it is working. I now rebased the code and resolved the conflicts, though the conflicts were really only in the documentation.

bernhardberger:
Thank you for rebasing and resolving the conflicts. Will try the PR either tonight if I find the time or at latest sometime next week.

Your work is really appreciated, hopefully this will get merged at some point 👌

bernhardberger:
Finally got around to properly testing it - works flawlessly for me. +1

Can't wait to have this in the master.

dmbutyugin:
OK, thanks a lot for testing! Then there is an extra confirmation that the code works, and I think it can be merged to Klipper. When someone reviews the code, that is :)

bernhardberger:
> OK, thanks a lot for testing! Then there is an extra confirmation that the code works, and I think it can be merged to Klipper. When someone reviews the code, that is :)

unfortunately I just found a bug during `PROBE_CALIBRATE`:

```
22:34:54
smart_effector: z_offset: -0.151
The SAVE_CONFIG command will update the printer config file
with the above and restart the printer.
```

after `SAVE_CONFIG` I get an error:
22:35:02
SAVE_CONFIG section 'smar
t_effector' option 'z_offset' conflicts with included value

Log files attached.

[moonraker.log](https://github.com/Klipper3d/klipper/files/8457717/moonraker.log)
[klippy (1).log](https://github.com/Klipper3d/klipper/files/8457715/klippy.1.log)

dmbutyugin:
> I know too little about the structure of the Klipper codebase, but from what I gathered by quickly skimming through the code the [smart_effector] module 'extends' the [probe] module - correct?

Yes, it extends [probe] module and provides extended functionality.

> Is there a way to retain both modules and just move the smart effector related options into it's own module (like control pin and recovery time?). This would allow to keep the [probe] section streamlined with upstream and the docs?

No, right now you should either have [smart_effector] or [probe] module configured, but not both. Why would you want both? Besides, this was implemented in the same manner as [bltouch], which also replaces [probe] and they should not be configured simultaneously.

Separately, I suspect that the error

```SAVE_CONFIG section 'smart_effector' option 'z_offset' conflicts with included value```

may have a different origin. Can it be that you include a part of your config with `[smart_effector]` section from another file into your printer.cfg using `[include]`? If yes, then `SAVE_CONFIG` will not be able to automatically save an option override. Unfortunately, the klippy.log does have this information, so I cannot check it myself.

bernhardberger:
> Separately, I suspect that the error
> 
> `SAVE_CONFIG section 'smart_effector' option 'z_offset' conflicts with included value`
> 
> may have a different origin. **Can it be that you include a part of your config with `[smart_effector]` section from another file into your printer.cfg using `[include]`?** If yes, then `SAVE_CONFIG` will not be able to automatically save an option override. Unfortunately, the klippy.log does have this information, so I cannot check it myself.

That's correct. I have offloaded my complete smart_effector.cfg and [include] it in the printer.cfg 

I don't have any [probe] or [smart_effector] sections defined outside of smart_effector.cfg though.

If that's a known limitation - then sorry for crying wolf here!

KevinOConnor:
Thanks.  I committed this with some minor docs changes on top (commit 05330fd3).

-Kevin

dmbutyugin:
> If that's a known limitation - then sorry for crying wolf here!

Yes, it is a known limitation of the config include system in Klipper.

> Thanks. I committed this with some minor docs changes on top (commit https://github.com/Klipper3d/klipper/commit/05330fd3e5ba45a8a694f16fd2e4863858133a51).

Thank you, Kevin!

PR #4689, opened 2021-09-13
max z-accel and speed ratio calculation  for deltas

Hywelmartin:
Signed-off-by: Martin Malmqvist volcomosq@gmx.com


KevinOConnor:
Thanks.  I think I'm missing the "big picture" on this change.  Why would a user want to limit the Z acceleration on a delta?

Also, the new docs state `It limits the acceleration of the z stepper motor` but a delta doesn't have a z stepper motor?

-Kevin

Hywelmartin:
Now with input shaping, I got a recommended acc at 24000mm/s² and it works nice… except z-movements… that sounds like a machine gun… If I dull it down a bit the printer behaves in Z… <10k
The joined movement of the whole system at the same time in the same direction is only connected to clean z movements like z-hops and layer changes.

Hywelmartin (commented):
removed the text on z motor

Hywelmartin:
made a little movie with or without the clonk.. stock vs reduced z acc..
https://youtu.be/kLnJhON5woU

KevinOConnor:
Interesting.  I'm not sure why that would be though.  It seems odd to me that a delta would be able to handle XY moves at the given acceleration while not being able to handle it on the Z!?  In particular, a move on the XY could result in a carriage acceleration of up to 3 times the specified acceleration ( https://www.klipper3d.org/Kinematics.html#stepper-motor-acceleration-limits ).

So, not understanding the "why", I guess I'm not sure there is a wide enough audience of users that would benefit from this change.  (As a general "rule of thumb", I try to target code features with a minimum of 100 users.)  If this change helps only a handful of printers then it might be better for those users to maintain the code/docs for it - or possibly add macros to manually change the accel on layer changes emitted by the slicer.

Separately, if you're still running your delta on an AVR - I'd be curious if the code change at https://klipper.discourse.group/t/intermittent-tmc-driver-crash-unable-to-read-tmc-uart-registers/932/14 has any impact on this.

Finally, note that the input shaper code does not make a recommendation on acceleration.  It only recommends *not going above* a particular acceleration.

Cheers,
-Kevin

Hywelmartin:
It's the clonks that I'm trying to mitigate. I have been stiffen up the machine and this "clean" z moves has been louder the stiffer the motion system have been. 
The rather new Flsun SR is a capable Dela out of the box, so I thing that we will see lots of users that will benefit from this.

I don't see any harm in adding these features. The speed ratio part is mostly usable for bigger deltas(not flat beds)  and the Z acc part could just be same as max acc for XY..

I do use different acc for different features and I know that the "suggested" is max..

I do run AVR but TMC2130... SPI- so no Uart.. sorry

dalegaard:
The existing code does not, as far as I can tell, work well with bed meshing. The speed of any move with _any_ Z movement will be capped at the Z movement speed, even if most of the speed of the move is in the XY plane. No z-ratio scaling is done like for e.g. cartesian and coreXY.

KevinOConnor:
> The existing code does not, as far as I can tell, work well with bed meshing. The speed of any move with any Z movement will be capped at the Z movement speed, even if most of the speed of the move is in the XY plane. No z-ratio scaling is done like for e.g. cartesian and coreXY.

Ah, yeah, that's not ideal.  That should probably be changed in Klipper.

-Kevin

dmbutyugin:
@KevinOConnor 

> Interesting. I'm not sure why that would be though. It seems odd to me that a delta would be able to handle XY moves at the given acceleration while not being able to handle it on the Z!?

After some thinking, I think that could be quite possible. When input shaping is enabled, the system can move much smoother in X/Y direction (at least, the motion is perceived to be much smoother). E.g. moving a bed on an Ender 3 at 5000 mm/sec^2 feels very 'jerky', each direction change feels like a kick, creates a lot of noise, and the stepper can skip steps. With input shaping, there are no sounds of hits and kicks. On a delta printer, it is pretty much the same in X/Y direction. However, Z motion is different between Cartesian/CoreXY printers and Delta printers. On Cartesian/CoreXY there is a separate acceleration limit, and even though input shaping is not enabled for Z axis, it's still OK. On a Delta, input shaping is also not enabled for Z movement, but the acceleration limit is the same as in X/Y. So, the effector can easily experience resonances when moving in Z direction. So, I would agree that limiting Z acceleration could be beneficial on a Delta printer (I just never printed at more than 5K accel on my delta, but I can confirm this behavior when I was setting max_accel to 7-10K for resonance testing with a test model, for instance - homing becomes very noisy and shakes the whole printer).

KevinOConnor:
> When input shaping is enabled, the system can move much smoother in X/Y direction

Ah, that's a good point.  I hadn't thought of that.

Still, if that's the case, I think adding a Z input shaper option might be more appropriate for deltas.

Separately, I agree we should scale the delta max z velocity limit, but I'd prefer to do that after the v0.10.0 release.

-Kevin

dmbutyugin:
> > When input shaping is enabled, the system can move much smoother in X/Y direction
> 
> Ah, that's a good point. I hadn't thought of that.
> 
> Still, if that's the case, I think adding a Z input shaper option might be more appropriate for deltas.

That could definitely be an option. Just calibrating it and finding the resonance frequency would be pretty much impossible with a print test model - only with an accelerometer. And while I expect that the resonance frequencies should be the same in X/Y plane if the belts are equally tightened (at least, I can see why X and Y input shaper configuration should be the same), I do not think it is reasonable to assume that the resonance frequencies in Z direction will be the same as in X/Y.

KevinOConnor:
> I do not think it is reasonable to assume that the resonance frequencies in Z direction will be the same as in X/Y.

Hrmm.  I would have thought Z would be the same as X and Y on a delta.  On a delta, we're ignoring the carriages and assuming the toolhead is the main source of resonance.  I'd have thought that resonance would occur no matter what direction you move the toolhead.  I could be wrong though.

If I get a chance, I'll dust off my delta, attach the accelerometer, and see what it does.

-Kevin

Hywelmartin:
@KevinOConnor my Delta isn't dusty what are you going to test..

dmbutyugin:
FWIW, I pushed a commit to my https://github.com/dmbutyugin/klipper/tree/resonance-test-methods branch that adds support  `TEST_RESONANCES AXIS=Z` method, for the default vibrations test and the other pulses test from that branch. I ran both tests on my old delta printer (with the smart effector). While I generally find the resonance testing results on it currently somewhat inconclusive, they still suggest that the resonances on Z axis could be different than on the other axes:

vibrations_test
pulses_test

KevinOConnor:
@Hywelmartin 
> my Delta isn't dusty what are you going to test..

The tests I would run are what Dmitry has coded and run.  I'd probably also try coding up the Z input shaping to see if the "machinegun noises" went away when it is applied.  Alas, I don't have a timeframe for when I'd be able to look at that.

@dmbutyugin 
Interesting, thanks.

-Kevin

Hywelmartin:
I can test it as well when I get my ADXL back...

dmbutyugin:
FWIW, this PR now has conflicts with the mainline Klipper branch, it would make sense to resolve them if you want to go forward with this PR. While the input shaping on Z axis perhaps makes sense, it's not clear whether it is worth the effort. And limiting max z accel is easy and not something unheard of in Klipper (e.g. it is a thing for other kinematics).

KevinOConnor:
> While the input shaping on Z axis perhaps makes sense, it's not clear whether it is worth the effort. And limiting max z accel is easy and not something unheard of in Klipper 

Okay, we can go ahead with a merge.  I think it may be worthwhile to make the documentation more specific on why one would want to set this value on a delta - maybe something like:
```
max_z_accel:
#   This sets the maximum acceleration (in mm/s^2) of movement along
#   the z axis. Setting this may be useful if the printer can reach higher acceleration
#   on XY moves than Z moves (eg, when using input shaper).  The default
#   is to use max_accel for max_z_accel.
```

-Kevin

Hywelmartin:
I'm ok with that ..

klippy/kinematics/delta.py
```diff
@@ -30,6 +30,8 @@ def __init__(self, toolhead, config):
         self.max_z_velocity = config.getfloat(
             'max_z_velocity', self.max_velocity,
             above=0., maxval=self.max_velocity)
+        self.max_z_accel = config.getfloat('max_z_accel', self.max_accel,
```
dmbutyugin, line 33: I'm sorry, but this appears to be unused (maybe the code was lost during rebase/merge)?

docs/Config_Reference.md
```diff
@@ -250,6 +250,11 @@ max_z_velocity:
 #   maximum speed of up/down moves (which require a higher step rate
 #   than other moves on a delta printer). The default is to use
 #   max_velocity for max_z_velocity.
+#max_z_accel:
+#   This sets the maximum acceleration (in mm/s^2) of movement along
+#   the z axis. Setting this may be useful if the printer can reach higher acceleration
+#   on XY moves than Z moves (eg, when using input shaper).  The default
```
dmbutyugin, line 256: Would you mind removing the spurious extra spaces?

klippy/kinematics/delta.py
Hywelmartin: how do you mean unused..

docs/Config_Reference.md
Hywelmartin: I found one space.. and rearranged it to have a nicer right "border"

klippy/kinematics/delta.py
Hywelmartin: removed

dmbutyugin: Sorry, I meant it the other way around: you created this member variable (self.max_z_accel), which is good, but the code to actually make use of this variable was accidentally removed (it seems). Now if I look at the diffs:
https://github.com/Klipper3d/klipper/pull/4689/files
I see that only documentation is updated. But the actual code to limit Z accel on Delta is gone. Please add it back :)

Hywelmartin: sorry I'm new at this.. =)

Hywelmartin: think I got it right.

KevinOConnor:
Thanks.

-Kevin

PR #4691, opened 2021-09-14
config: Added the Mellow Super Infinity HV board

tehmaze:
This adds a reference configuration file for the Mellow Super 8 High-Voltage board.


KevinOConnor:
Thanks.  In general, it looks fine, but commits need to have a signed-off-by line containing your full real name - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

Also, minor, but I think it might be preferable to just say "infinity" than have the emoji in the file comments ( `Mellow Super ♾️  HV Board`).

-Kevin

tehmaze:
PR updated, Mellow themselves abbreviate Infinity to "8" in their product listings so I adapted that.

KevinOConnor:
Thanks.

-Kevin

PR #4696, opened 2021-09-16
tmc: add current values to get_status()

phil65:
Adds current values to get_status() for TMC motors.
I noticed that the current values slightly differ from the ones set in the config. I guess that this is intended?


zellneralex:
@phil65 yes that`s intended.  The actual value is for e.g a TMC2209 max_current (defind by rsense) /31 x scale value. That means you have only 32 dedicated values and the scale value will be set to get the actual value as close as possible to the requested value

phil65:
Thanks for the explanation!

KevinOConnor:
Thanks.  The code change looks fine, but any `get_status()` update needs to have a corresponding change to docs/Status_Reference.md.

Also, commits need to have a signed-off-by line containing your full real name - see: https://github.com/KevinOConnor/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

phil65:
Done. 👍

KevinOConnor:
Thanks.  It looks fine to me.  This is something I'll look at merging after the Klipper v0.10.0 release ( https://klipper.discourse.group/t/upcoming-v0-10-0-release/791 ).

-Kevin

phil65:
Sure. Thx for reviewing.

PR #4701, opened 2021-09-17
Add support for BLTouch probe on the Wanhao D9 MK2

mdshw5:
I'm adding a BLTouch probe to my MK1 printer, so needed to determine which pin is connected for servo control. Based on the pin assignments in `pins_I3PLUS3030.h` in the archive extracted from https://github.com/garychen99/D9MK2 I can determine that the `control_pin` is `PH4`. I'll test this configuration and report success before requesting a merge.

### DIFF
--- a/config/printer-wanhao-duplicator-9-2018.cfg
+++ b/config/printer-wanhao-duplicator-9-2018.cfg
@@ -1,6 +1,7 @@
 # This file contains pin mappings for the Wanhao Duplicator 9 MK1,
-# also sold as the Monoprice Maker Pro MK1. To use this config,
-# the firmware should be compiled for the AVR atmega2560.
+# also sold as the Monoprice Maker Pro MK1. Pin assignments for the BLTouch
+# Z probe included on the Wanhao Duplicator 9 MK2 must be uncommented to use.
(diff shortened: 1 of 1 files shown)


mdshw5:
@KevinOConnor The printer comes in several revisions. Originally the MK1 revision came with an inductive probe and a plug on a daughter board to break out pins for this application:
MK1 D9 splitter board

The MK2/MK3 revisions come with a BLTough probe and a replacement daughter board to break out the appropriate pins:

MK2 D9 splitter board

Depending on the configuration a user will need to include the `[probe]` or the `[bltouch]` and `[safe_z_home]` configuration blocks. I've updated the pull request with more necessary config and comments to instruct appropriate configuration. If you would rather have these as separate config templates I  can understand and would make this change.

KevinOConnor:
Okay, thanks.  The change looks fine to me.  One minor request would be to not modify the top comment block and instead put the comments about bltouch next to the bltouch section (so as to focus the top comment block on firmware and flashing).

Separately, commits need to have a signed-off-by line containing your full real name - see: https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

PR #4705, opened 2021-09-19
config: Add [board_pins] for generic-duet3-mini.cfg

joaobarros:
Add EXP1 / EXP2 display pins to [board_pins] for the Duet 3 Mini 5+
Mention the requirement of software SPI pin declaration on [display]

Signed-off-by: Joao Barros <joao.barros@gmail.com>


KevinOConnor:
Thanks.  In general, it looks fine to me.  However, I'm not sure I understand the comments on the SPI bus - as some LCD displays don't use SPI and thus can't define those options.

@lukeslaboratory - do you have any comments on this?

-Kevin

lukeslaboratory:
The Mini 5+ doesn't support "normal" LCD's via its EXP1+2 headers as it has a reset pin that makes the MCU reset if connected with "simpler" LCD's.

The pins aren't correct in this example - SPI2MISO is PC15 on the header, MOSI is PC12, and SCK is PC13
image
image


its intended for the Mini 12864 boards (uc1701) and similar. I use the following settings for the fystek mini12864:
```
[display]
lcd_type = uc1701
cs_pin = PC6
a0_pin = PA2
contrast = 63
encoder_pins = ^PC11, ^PD1
click_pin = ^!PB9
spi_software_miso_pin = PC15
spi_software_mosi_pin = PC12
spi_software_sclk_pin = PC13

[output_pin beeper]
pin = PA9

[neopixel fysetc_mini12864]
pin = PB12
chain_count = 3
color_order = RGB
initial_red = 0.4
initial_green = 0.4
initial_blue = 0.4
```

KevinOConnor:
Okay.  It seems like defining the EXP1 and EXP2 aliases would be useful, but I guess the pin definitions on this PR are not correct?

-Kevin

joaobarros:
Luke is right, it should be spi_software_miso_pin = PC15 SPI2_MISO and not PA4 which is SPI0_MISO
Curiously enough my Fysetc Mini 12864 v2.1 works with both settings.

I have corrected the pin. Thank you Luke!

KevinOConnor:
FYI, if that's the case, you should be able to use hardware spi with:
```
[samd_sercom exp2_sercom]
sercom: sercom7
rx_pin: PC15
tx_pin: PC12
clk_pin: PC13

[display]
spi_bus: sercom7
...
```

Separately, I find the current comments added to generic-duet3-mini.cfg to be confusing.  I'm not sure what it's trying to convey.

-Kevin

joaobarros:
@KevinOConnor I actually tried the sercom config but didn't work most probably because of my wrong pin.
I just tested now and it's working :)

I have edited the comment about the SPI pins to be similar to other config examples.

KevinOConnor:
Thanks.  I committed this and made a minor change on top (commit 31ac9dc6).

-Kevin

PR #4713, opened 2021-09-22
F429 32KiB bootloader

bigtreetech:
Signed-off-by: Alan from BigTreeTech tech@biqu3d.com

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -135,7 +135,7 @@ config STACK_SIZE
 ######################################################################
 
 choice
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  The change looks fine, but commits need to have the full real name of the submitter - https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md .  I can make this change under my signed-off-by line though.

Out of an abundance of caution I'll look to make this change after the Klipper v0.10.0 release.  https://klipper.discourse.group/t/upcoming-v0-10-0-release/791

-Kevin

PR #4716, opened 2021-09-23
Exclude Objects

troy-jacobson:
module: Exclude Objects

Adding Klipper functionality to support cancelling objects while printing.

This module keeps track of motion in and out of objects and adjusts movements as needed.  It also
tracks object status and provides that to clients.

The Klipper module is relatively simple, and only provides one piece of the workflow.  Support from Moonraker is underway to pre-process gcode files from various slicers with the extended gcode commands supplied by this module.  UI's such as Fluidd and Mainsail will provide the user facing controls.

There has been a small group sharing code.  In addition to the Moonraker work, I have Fluidd code that will be submitted shortly.  Mainsail support is also underway.

Signed-off-by: Troy Jacobson <troy.d.jacobson@gmail.com>
Co-authored-by: Franklyn <voron@tackitt.net>
Co-authored-by: Eric Callahan <arksine.code@gmail.com>


kageurufu:
A reference implementation of the preprocessor is available at https://github.com/kageurufu/cancelobject-preprocessor and works for Cura, IdeaMaker, and Slic3r-family. Other slicers are fairly trivial to implement at this point, and I plan on adding a conversion for M486 -> our syntax. The benefit over macros are getting inferred POLYGON and CENTER.

Integration into moonraker is also in progress, to simplify the overall workflow. Native slicer support is also in discussion with a few slicer developers.

wlhlm:
`exclude_object.py` has execute permissions. There is no need to run it directly, so it's probably better to change it to 0644 like the rest of the files in `klippy/extras`.

Arksine:
I'm going to try to do a detailed review of the module soon.  One thing I would recommend is to rebase your branch against master so the PR doesn't have 173 commits, it makes the review a bit easier.

troy-jacobson:
@wlhlm Fixed the permission.
@Arksine Fixed the merge/rebase issue.

Arksine:
I left a few comments, I'm mostly unsure about the extrusion offset.  Once I have a good understanding of how your approach is working, there are some other things to consider:

1) What happens when an object is excluded while the tool is printing in that location?  This is likely going to affect retraction, and IIRC different slicers handle it in different ways.
2) If my understanding of the extrusion offset is correct, we likely need to offset individual extruders in for setups with multiple extruders.

KevinOConnor:
Thanks.  What's the current state of this PR?

-Kevin

kageurufu:
@KevinOConnor 

Sorry for the delay. It's pretty well ready to go as far as I've been aware, Moonraker has full support for processing uploaded G-Code for cancellation, the python module there is fully functional standalone for processing within your slicer.

Cancellation works correctly in testing, and the fork has been tested by a variety of users to great success so far. I believe @Arksine was intending to review this as well when they have the time

wlhlm:
There is currently a problematic interaction between this module and `tuning_tower` caused by the handling of the move_transformers. @troy-jacobson is aware of this.

Arksine:
Sorry it took me a while to get to this.  I went ahead and looked at the updates.  I think we are close, and I know its testing well.  I had some minor comments and the question about a "what if" when attempting to compensate for retraction.  Once that is hammered out we can rebase this and it will be approved from my perspective.

troy-jacobson:
Hey everyone.  Sorry for the lapse in communication.  I haven't had as much time to dedicate to this since the holidays.  Here's the current status:

- Overall, the cancel object functionality has been working great.  In addition to the testing done through the Mainsail beta, it's been on my system for ages and I've yet to have an issue.
- I believe the current MR has a fix for the retraction issues Arksine mentioned.  I'll check in with him soon now that I have some time to work on this.
- What I consider the main blocker right now is the issue with registering the transform.  When used with other modules, say the tuning tower, the transforms are not unregistered in the reverse order they are registered resulting in the chain of transforms getting corrupted.  I've looked at some ideas to fix this which will work most of the time, but may not be correct in all situations (especially if transforms other than the tuning tower can cause the same issue).

Arksine:
> I believe the current MR has a fix for the retraction issues Arksine mentioned. I'll check in with him soon now that I have some time to work on this.

Ok, I'll wait on pending feedback for the retraction issue.  IIRC, you mentioned that this was to accommodate SuperSlicer.  I'm curious, does this affect PrusaSlicer as well?  If not, should we just consider it a bug in SuperSlicer and request that it get fixed there rather than attempt to compensate in Klipper?

> What I consider the main blocker right now is the issue with registering the transform. When used with other modules, say the tuning tower, the transforms are not unregistered in the reverse order they are registered resulting in the chain of transforms getting corrupted. I've looked at some ideas to fix this which will work most of the time, but may not be correct in all situations (especially if transforms other than the tuning tower can cause the same issue).

The issue we are bumping into is that Tuning Tower is registered before Exclude Object is registered, then unregistered before Cancel Object is unregistered.  This is effectively "breaking the chain", because when Tuning Tower is removed Exclude Object's `next_transform` is still referencing it.  I think its best to get Kevin's feedback on this, I'm not sure how to fix it without refactoring how transforms are registered.  We could document that `exclude_object` is not compatible with `tuning_tower`.  To my knowledge there are no other "dynamic" transforms, `bed_mesh`, `bed_skew`, and `bed_tilt` are never unregistered...but that doesn't mean that more won't be added in the future.

kageurufu:
We could also just make `exclude_object` not dynamic, and short-circuit it's own logic when not needed

klippy/extras/exclude_object.py
```diff
@@ -0,0 +1,276 @@
...
+                      self.max_position_excluded)
+        logging.debug("EXO: New position: %s", " ".join(str(x) for x in newpos))
+        logging.debug("EXO: Offset: %s", " ".join(str(x) for x in offset))
+        self.log_next_moves = 10
+
+        # This adjustment value is used to compensate for any retraction
+        # differences between the last object printed and excluded one.
+        self.extruder_adj = self.max_position_excluded \
+            - self.last_position_excluded[3] \
+            - (self.max_position_extruded - self.last_position_extruded[3])
```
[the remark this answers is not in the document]
troy-jacobson: Yes, that is the intended logic around retraction, and the intent is to ensure that the filament is in the expected position in the extruder prior to moving to the next object.

You're correct that this is assuming a particular behavior on the part of the slicers, but this behavior has been consistent between slicers so far.  This is relatively simple logic and I'm comfortable with that it's doing.  If there's a slicer or something that needs different behavior, that could probably be addressed in preprocessing.

troy-jacobson:
Just pushed changes to address the Tuning Tower issue.  The change is basically a) don't load the exclude_object transform if a tuning tower test is active and b) don't unload the exclude_object transform if it's not the at the head of the list.  I think this both fixes the Klipper crashes and maintains the expected behavior for users.  
I thought about some alternatives, but I felt they may either make the tuning tower more difficult to use, figuring out how to unregister the exclude_object transform, or navigate non-intuitive side effects of how the two modules interact.
Here are the two scenarios:
A: I've been actively printing have have been using the exclude_object module.  It's still loaded when starting a tuning tower.  The tuning tower transform is loaded ahead of exclude object.  Given that the tuning tower doesn't actually transform the movements, I can't see any reason why this state needs to be avoided.  When the SD card reset is called at the start of the print, the exclude_object transform is not unloaded like it normally would be.  Again, I'm not seeing any functional issues with this.  After tuning is finished, a regular print is started.  Again, the exclude_object transform is reused because at this time, the tuning tower transform is still enabled.  Once the print starts extruding filament, however, the tuning tower recognizes that the test is done and unloads itself.  The exclude_object transform is now at the front of the transform list.  Subsequent prints will behave as normal with regards to registering and un-registering the exclude object move transform.  During this time, the user will be able to cancel objects, including the tuning tower.
B) Neither the tuning tower nor exclude object transforms are loaded.  A tuning test is started, loading that transform.  If the STL for the test print has objects defined, the attempt to load the exclude object transform will be rejected and the exclude object functionality will not be enabled.  After the tuning print, a regular print is started.  At this point, the tuning test is still active so the exclude object transform is still not enabled.  Once that print starts extruding, the tuning tower will detect that the test is finished and unload its transform.  The next regular print will be able to enable the exclude object functionality as expected.
Does this sound OK to you?  The second scenario has the quirk around the first print after the tower, but honestly given what's usually happening when doing tuning towers, I bet it would hardly be noticed.  Also, I think the first scenario is going to be the most common one anyway.

klippy/extras/exclude_object.py
```diff
@@ -52,7 +51,11 @@ def __init__(self, config):
             desc=self.cmd_LIST_EXCLUDED_OBJECTS_help)
     def _setup_transform(self):
         if not self.next_transform:
-            logging.debug('Enabling ExcludeObject as a move transform')
+            tuning_tower = self.printer.lookup_object('tuning_tower', None)
+            if self.gcode_move.move_transform == tuning_tower:
```
Arksine, line 55: We can't peek into another object's members, so we either need to add a `get_move_transform()` method to `gcode_move`, or add the `is_testing()` method to `tuning_tower` and perform the check that way.  This would apply to the check on line 80 as well.

scripts/klippy-requirements.txt
```diff
@@ -7,3 +7,4 @@ pyserial==3.4
 greenlet==1.1.2
 Jinja2==2.11.3
 python-can==3.3.4
+markupsafe<2.1
```
Arksine, line 10: It looks like this got unintentionally got pulled in.  I think Kevin plans to pin markupsafe to 1.1.1.

Arksine:
I added a couple of minor comments, but overall I think it looks good and ready to go.

Arksine:
Everything looks good.  My last ask is a bit tedious but it will help Kevin out.  It would be good if we can get this into four commits, for example they could look like the following:
```
virtual_sdcard:  add reset_file event

Signed-off-by: Troy Jacobson <troy.d.jacobson@gmail.com>
```
```
tuning_tower: implement is_testing() method

Signed-off-by: Troy Jacobson <troy.d.jacobson@gmail.com>
```
```
exclude_objects: initial implementation

<you can decide how you want to describe>

Signed-off-by: Troy Jacobson <troy.d.jacobson@gmail.com>
```
```
docs: add exclude_object documentation

Signed-off-by: Troy Jacobson <troy.d.jacobson@gmail.com>
```

That way Kevin can just hit rebase and merge when he's ready.  That said, before you do that we might want to see if Kevin has any additional input.  Also, you might find that its easier to do this in a new branch/PR rather than to modify this one.  If you want to give it a go in this one via `git rebase` my only suggestion would be to make a backup first :smile:

KevinOConnor:
Thanks.  At a high-level I agree this is useful functionality that would be good to have in Klipper.  I have some high-level questions, some high-level observations, and some minor code comments.

The documentation states that it is expected one would use a slicer or post-processor to use this functionality.  In practice, what slicers and/or tools are available today, and what are users expected to use today?  What special steps (if any) do users need to do today to make this work?

As a high-level comment on the implementation, I'm concerned about the host cpu efficiency.  If I understand this functionality correctly, users would be expected to enable this module for all prints, but rarely exclude an object.  This module looks like it will register a move transform for all prints - even if the user never aborts an object in practice.  The move code path is a "fast path", and it seems like this would add some notable overhead to normal moves.  Have you performed benchmarks on this code?  Are there alternatives to registering the transform even when no object is excluded?

I agree that it would be good to separate out the virtual_sdcard.py and tuning_tower.py changes to separate commits.  As to exclude_object.py and docs - I generally prefer to have docs in the same commit as the code, but if you want to separate them out it is also fine.  However I have a couple of comments on the modifications to the external modules:
1. I think it would be preferable to use a name like `is_active()` in tuning_tower.py as I think `in_testing()` could be ambiguous.
2. The change to virtual_sdcard.py seems a little odd to me.  We can certainly export an event for the sdcard.  However, wouldn't it make more sense for the new print g-code to explicitly call reset on start and finish?  Seems odd that sdcard prints would be handled differently from prints originating directly from OctoPrint.  If we do export an event, note that the prefix should match the module name - `virtual_sdcard:xxx` and not `sdcard:xxx`.

I have some comments on the command naming.  I think "OBJECT" may be too generic - how about changing the command names to something like:
- DEFINE_OBJECT -> EXCLUDE_OBJECT_DEFINE
- START_CURRENT_OBJECT -> EXCLUDE_OBJECT_START
- END_CURRENT_OBJECT -> EXCLUDE_OBJECT_END
- LIST_OBJECTS -> EXCLUDE_OBJECT_DEFINE with no parameters
- LIST_EXCLUDED_OBJECTS -> EXCLUDE_OBJECT with no parameters
- EXCLUDE_OBJECT_RESET -> EXCLUDE_OBJECT RESET=1

I have some minor code comments that I will send separately.  I'm still trying to understand the move transforms so I may follow up with additional comments in the coming days.

Thanks again,
-Kevin

docs/Exclude_Object.md
```diff
@@ -0,0 +1,106 @@
+# Exclude Obects
+
+The `[exclude_object]` module allows Klipper to exclude objects while a print is in progress.
```
KevinOConnor, line 3: It would be preferable to wrap documentation files to 80 columns.

Also, new documents should be added to overview.md and to docs/_klipper3d/mkdocs.yml  (as mentioned in docs/CONTRIBUTING.md ).

docs/Exclude_Object.md
```diff
@@ -0,0 +1,106 @@
...
+components and users have many options to choose from.  Therefore, in order to provide a
+a consistent user experience, the `[exclude_object]` moudle will establish a contract or API
+of sorts.  The contract covers the contents of the gcode file, how the internal state of the
+module is controlled, and how that state is provided to clients.
+
+## Workflow Overview
+A typical worfklow for printing a file might look like this:
+1. Slicing is completed and the file is uploaded for printing.  During the upload, the file
+   is processed and `[exclude_object]` markup is added to the file.
+1. When printing starts, Klipper will reset the `[exclude_object]` status.
```
KevinOConnor, line 17: It would be preferable to expand the numbering (eg, 1, 2, 3) so the document is readable via text files.

docs/Status_Reference.md
```diff
@@ -60,6 +60,39 @@ The following information is available in the
   forward direction minus the total number of steps taken in the
   reverse direction since the micro-controller was last restarted.
 
+## exclude_object
+
+The following inforation is avaialbe in the [exclude_object](Exclude_Object.md) object:
```
KevinOConnor, line 65: Spell check (eg, available).

klippy/extras/exclude_object.py
```diff
@@ -0,0 +1,263 @@
...
+            'DEFINE_OBJECT', self.cmd_DEFINE_OBJECT,
+            desc=self.cmd_DEFINE_OBJECT_help)
+        self.gcode.register_command(
+            'LIST_OBJECTS', self.cmd_LIST_OBJECTS,
+            desc=self.cmd_LIST_OBJECTS_help)
+        self.gcode.register_command(
+            'LIST_EXCLUDED_OBJECTS', self.cmd_LIST_EXCLUDED_OBJECTS,
+            desc=self.cmd_LIST_EXCLUDED_OBJECTS_help)
+    def _setup_transform(self):
+        if not self.next_transform:
```
KevinOConnor, line 52: It's preferable to use `if self.next_transform is None` when checking an external object because the external object could override the truth check.

klippy/extras/exclude_object.py
```diff
@@ -0,0 +1,263 @@
...
+    def get_position(self):
+        extr_idx = self.toolhead.get_extruder().get_name()[8:]
+        if extr_idx:
+            self.extruder_idx = int(extr_idx)
+        else:
+            self.extruder_idx = 0
+        len_diff = self.extruder_idx + 1 - len(self.extrusion_offsets)
+        if len_diff > 0:
+            self.extrusion_offsets.extend([[0., 0., 0., 0.]] * len_diff)
+        offset = self.extrusion_offsets[self.extruder_idx]
```
KevinOConnor, line 100: I'm not sure what this code does, but it looks a bit unusual to me.  I suspect things would be more clear (and more efficient) if `self.extrusion_offsets` was a dict instead of a list - then the code could just be `offset = self.extrusion_offsets[self.toolhead.get_extruder().get_name()]`.

klippy/extras/exclude_object.py
```diff
@@ -0,0 +1,263 @@
...
+        # Inside cancelled object
+        return self.current_object in self.excluded_objects
+
+    def get_status(self, eventtime=None):
+        status = {
+            "objects": list(self.objects.values()),
+            "excluded_objects": list(self.excluded_objects),
+            "current_object": self.current_object
+        }
+        return status
```
KevinOConnor, line 176: Why is the `objects` field a list instead of the more natural dictionary?  That is, wouldn't it be more useful to return `"objects": self.objects`?  I'm a little concerned about the overhead of translating the object dict to a list on every get_status invocation (many UIs subscribe to everything, so internally we have to build the status every 250ms and check for changes).

klippy/extras/exclude_object.py
```diff
@@ -0,0 +1,263 @@
...
+        else:
+            object_list = " ".join(self.objects.keys())
+        gcmd.respond_info(object_list)
+    cmd_LIST_EXCLUDED_OBJECTS_help = "Lists the excluded objects"
+    def cmd_LIST_EXCLUDED_OBJECTS(self, gcmd):
+        object_list = " ".join (str(x) for x in self.excluded_objects)
+        gcmd.respond_info(object_list)
+    cmd_DEFINE_OBJECT_help = "Provides a summary of an object"
+    def cmd_DEFINE_OBJECT(self, params):
+        self._setup_transform()
```
KevinOConnor, line 236: The docs indicate that one does not need to "define objects", but as this is the only place the transform is registered, it would seem one must define at least one object.

klippy/extras/exclude_object.py
```diff
@@ -0,0 +1,263 @@
...
+            gcmd.respond_info("END_CURRENT_OBJECT called, but no object is"
+                              " currently active")
+            return
+        name = gcmd.get('NAME', default=None)
+        if name != None and name.upper() != self.current_object:
+            gcmd.respond_info("END_CURRENT_OBJECT NAME=%s does not match the"
+                              " current object NAME=%s" %
+                              (name.upper(), self.current_object))
+
+        is_excluded = self._test_in_excluded_region()
```
KevinOConnor, line 211: Dead code?

klippy/extras/exclude_object.py
```diff
@@ -0,0 +1,263 @@
...
+        self.last_position_excluded[:] =self.last_position
+        self.max_position_excluded = max(self.max_position_excluded, newpos[3])
+
+    def _move_into_excluded_region(self, newpos, speed):
+        self.in_excluded_region = True
+        self._ignore_move(newpos, speed)
+
+    def _move_from_excluded_region(self, newpos, speed):
+        self.in_excluded_region = False
+        offset = self.extrusion_offsets[self.extruder_idx]
```
KevinOConnor, line 157: Dead code?

KevinOConnor (commented):
Some minor code comments.

-Kevin

Arksine:
> The documentation states that it is expected one would use a slicer or post-processor to use this functionality. In practice, what slicers and/or tools are available today, and what are users expected to use today? What special steps (if any) do users need to do today to make this work?

Hi Kevin, I can help answer some of this.  With regard to the post-process, @kageurufu wrote a [python library](https://github.com/kageurufu/cancelobject-preprocessor) that handles it.  Moonraker can be configured to use this library to perform the post-process during its metadata extraction phase (just after a file is uploaded).  Specifically, it looks to see if the file is supported, has object labeling enabled, and hasn't been previously processed.  FWIW, this post processor generates bounding boxes for each object and injects "DEFINE_OBJECT" commands so that Klipper may read and store them.  It would be possible for Moonraker to extract this data, however I do think its better if Klipper reports it.

I also went ahead and did a couple of initial benchmarks in batch mode on a Pi 3A+.  We can do more to get more data, but I thought I would put something together to get started.  I sliced a series of 9 boxes with 15% gyroid infill and did two runs.  The first without `exclude_object` enabled on a file that did not go through the post process (so Klippy only sees the object labels as comments).  The second with `exclude_object` enabled on a post processed file.  The results were:
```
# Exclude Object Disabled

INFO:root:Exiting (print time 18614.240s)

real	2m42.426s
user	2m40.522s
sys	0m3.127s
```
```
# Exclude Object Enabled

INFO:root:Exiting (print time 18614.240s)

real	2m51.474s
user	2m48.595s
sys	0m3.244s

```
So roughly 9s longer, or a 5.6% hit running as fast as we can go.  That said, I don't think it would be an issue to enable the transform when an object is cancelled rather than when they are defined.

klippy/extras/exclude_object.py
Arksine: This was my suggestion, the goal is to track the amount of extrusion we are excluding per extruder.  I agree that your suggestion is a cleaner way of handling it.

troy-jacobson:
@KevinOConnor I see the logic behind renaming the commands and like how they all group nicely.  Unfortunately that's going to be the most problematic change to make.  I'll check with @kageurufu and @meteyou and come up with a plan for that.  We could probably configure some macros to map between old and new for a while.

Do we need to handle the OctoPrint path?  It has its own cancel object functionality and is implemented quite differently.  I think there would have to be a Klipper specific plugin if OctoPrint wanted to use this implementation, right?

I think I can make a change to address your concern about having the module always loaded.  The transform could be registered when the user requests to cancel an object with `EXCLUDE_OBJECT`.  That way it won't be running for prints where it is not needed.  It would be unregistered at the reset at the beginning of the the next print.  I was trying to think if there was a way to only load the transform while actively excluding objects, but there are complications with that idea.  I'll also take a look and see if there are ways to optimize the critical path through `move()`.

klippy/extras/exclude_object.py
troy-jacobson: I'll likely change to to load when an `EXCLUDE_OBJECT` command is received.

klippy/extras/exclude_object.py
troy-jacobson: Even though this code uses a map, it seemed that for a status, a simple list made more sense.  It's been a long time since I changed this code, but I remember having issues with status not updating when passing those values without wrapping them in the `list()` calls.

KevinOConnor:
> I see the logic behind renaming the commands and like how they all group nicely. Unfortunately that's going to be the most problematic change to make.

I understand.  However, it seems likely that a change after merging into the main Klipper repo will be even harder.  So, if there is agreement that more consistent naming will improve maintainability then I think we need to do that before merging.

> Do we need to handle the OctoPrint path? It has its own cancel object functionality and is implemented quite differently. I think there would have to be a Klipper specific plugin if OctoPrint wanted to use this implementation, right?

Well, these are questions I would ask you!

There is no requirement to handle OctoPrint from my side.  It does seem odd that there is a special case for virtual_sdcard though - why wouldn't the post processor just add the reset to the gcode at the start/finish on its own?  That is, it's not a requirement from me, but it seems odd to exclude the capability for no particular reason.  I may be missing something though.  This is not a blocker.

If I'm understand the replies here correctly, today the only practical method to utilize this module is with moonraker/mainsail.  That isn't clear from the documentation in this PR - the documentation seems to imply that there are slicers and post processing tools that can prepare the g-code.  I think the documentation should be more clear - if there's only one reasonable solution then lets just state that in the documentation.  (If I've missed something though, please correct me.)

> I think I can make a change to address your concern about having the module always loaded. The transform could be registered when the user requests to cancel an object with EXCLUDE_OBJECT. 

Okay.  To be clear, a 6% performance degradation isn't a blocker.  We can improve performance (if necessary) after merging.  However, if we think we need to change the behavior of the DEFINE_OBJECT / EXCLUDE_OBJECT commands then I'd like to settle that before merging.

Thanks,
-Kevin

klippy/extras/exclude_object.py
KevinOConnor: Right - if the dict is exported from `get_status()` then it needs to be treated as immutable internally.  See the recent conversation at #5222 .  Basically, make copies of the dict on every change to the dict instead of on every `get_status()` call.

JRHeaton:
This looks great. A feature I'm long awaiting coming from Marlin. Great work, and let me know if I can be of service in any way to help facilitate testing or other means of assisting the completion and adoption of this.

CODeRUS:
@JRHeaton you already can. copy exclude_object.py to klippy/extras, configure and restart klipper service.

Arksine:
Since we are post processing gcode files I think we could add something like a `CLEAR_OBJECTS` gcode at both the begging (before all `DEFINE_OBJECT` calls) and at the end of the file.  We could also modify the default behavior of `CANCEL_PRINT` to run `CLEAR_OBJECTS` it if `exclude_object` is configured.

Another alternative would be to instruct users to override `SDCARD_RESET_FILE`.  I see where Kevin is coming from and agree we shouldn't limit the ability to clear objects to the reset event.

Arksine:
That is typical behavior of exclude object.  The gcode on the excluded portion of the layer must still be processed up through the `exclude_object` module, this takes some time and introduces a pause.

CODeRUS:
@Arksine out of curiosity: have you considered fast-forwarding sdcard position from start to end object macro?

Arksine:
I'm not the lead dev on this, but I don't think that is feasible.  You couldn't fast forward as you still need to check each line of G-Code to determine when the object ends.  It is also necessary to process the excluded gcode moves so we can track how much extrusion is being excluded and offset the tool appropriately.

kageurufu:
@CODeRUS We have to do some processing of skipped G-Code to handle some edge cases, mostly around toolchanging and absolute extrusion

See https://github.com/troy-jacobson/klipper/blob/4c5e052d46f875a52da18b59f77bba99a1789bbd/klippy/extras/exclude_object.py#L142-L149

Fast-forwarding could be done in controlled situations, although it would probably require native slicer integration to get "right". If the preprocessor included the jump and extrusion offsets in the `START/END_CURRENT_OBJECT` it could possibly be done?

I'll experiment sometime, but the priority right now is getting this working in general.

CODeRUS:
@Arksine sounds like something can be done on moonraker preprocessor side, end_object macro could contain all necessary values itself, with no need to process chunks in realtime again.

@kageurufu no problem, just my silly question? mostly out of context :)

CODeRUS:
btw, @Arksine is it possible to have macro for calling object processing for file manually? i have couple of scenarios. 1. is when long file is saved by chunks, object processing may be called to early. 2. is when i dont want to have exculude object being active for all my prints for time savings and etc., so i can call processing manually per file. having separate slicing profiles can be pita in this case.

Arksine:
Moonraker imports @kageurufu's library to handle the processing.  That said, even if the preprocessor made these calculations and provided the correct location to jump to, there are still outstanding issues:
1) It isn't possible pre-calculate the portion of extrusion skipped if the object is canceled while its printing inside of that object, so you would need logic to process out the object at that point.
2) There may be some gcode embedded in the object that should still be executed, ie M73
3) This would require modification to Klipper's `virtual_sdcard` module, increasing complexity and maintenance in that portion of the code.  You would have to sell Kevin on whether or not such a modification is worth the change there.

Its not possible to manually call the preprocessor from Moonraker, the purpose of adding that support was to automate the process.  It should be possible to work with the library directly if you want to determine when and what files to process.

CODeRUS:
@Arksine thanks, just found debug output with correct command 

```
~/moonraker-env/bin/python ~/moonraker/moonraker/components/file_manager/metadata.py -p ~/gcode_files -f "$1" --check-objects
```

can work with that. asked that question because i already had issues with not properly processed files, when processing was started while file is still uploading manually.

glowtape:
I wonder if it would make sense to be able to define things like optional E- and Z-offsets to apply and undo when object exclusion starts and ends respectively, e.g. raise the nozzle a millimeter and retract the filament some to prevent oozing during stalls.

troy-jacobson:
@wlhlm I would like to see the gcode file, if you still have it.  That's a lot more of a pause than I'm used to seeing.
I've excluded large objects with gyroid infill and didn't seem to have issues.  SS does end up with an extra move or two when switching layers in a situation like this, but it shouldn't be that long unless you're running with a lot of retraction.

troy-jacobson:
@glowtape I don't think it should be necessary to have extra moves like that.  There could be a situation, though, where we should tweak the preprocessing in order to clean up the between object moves for an individual slicer.  I feel that this portion in klipper should be as simple as possible.  If we need to provide hints or slicer specific work-arounds, they should be directed by the preprocessor.

wlhlm:
> @wlhlm I would like to see the gcode file, if you still have it.

Sure thing.

[wlhlm-samples-prusament-asa.gcode.gz](https://github.com/Klipper3d/klipper/files/8157902/wlhlm-samples-prusament-asa.gcode.gz)

troy-jacobson:
@JRHeaton If you're interested in trying this out, the best way is through Mainsail.  There is documentation from the beta cycle that includes directions for klipper and mainsail.  The currently released Mainsail has the UI side of the functionality.  It's best to discuss any install questions there.

troy-jacobson:
I've been processing the comments here for a while, and feel like I have a good path forward for all of the outstanding items except around the sd card reset.   Here are some of my current thoughts and things that influenced the implementation.

I'm hesitant to delegate this to the preprocessor.  If someone is in the middle of configuration, tuning, slicer changes, etc. they could easily run into a situation where the UI could be showing object information that is from an older print.  That's why I looked for an automated solution.  

I had thought about having the UI handle this, but then starting a print from an LCD could cause issues.

From a user's perspective, it's nice not to reset the state at the end of a print.  It lets them review what happened.  Especially if the print was cancelled.

I'm not sure if this has any bearing on the reset implementation, but I'd like to have flexibility to tie into the Moonraker metadata in the future.  One example could be reprinting only the cancelled objects.

KevinOConnor:
> I've been processing the comments here for a while, and feel like I have a good path forward for all of the outstanding items except around the sd card reset.

Okay.  I don't object to an automated reset on an sdcard 
reset.

As previously mentioned though, the event should use "virtual_sdcard" prefix and it would be preferable for the change to be in a separate commit.

Thanks.
-Kevin

Arksine:
Perhaps it would still be a good idea to add a G-Code command that resets the transform.  That way `exclude_object` isn't exclusively locked to Moonraker.  For example, OctoPrint's `cancel object` plugin could implement it as an alternative, Repetier could implement it if they so chose, etc.

gurusonwheels:
im not a developer . but i have installed this yesterday on my voron 2.4  with mainsail os and mainsail web UI . it works very well and i dont see any pauses when i exclude an object .  im just a end user and wanted to give you my 2 cents for whatever its worth . 
i use superslicer on windows 10 .  my voron has a 5 inch lcd with klipperscreen if you need something tested .  as i say im not a developer so my linux skills is very week . 
one thing i notice is the popup interface in mainsail web for excluding objects . it highlights the object on the left but not the actual name on the right .   please make it highlight the name as well . makes it very hard to track down which one to click the exclude button when there small items being printed as it switches to fast .   other than that kudos for a job well done so far .

kageurufu:
Alright, I have a updated branch for this PR ready at https://github.com/kageurufu/klipper/tree/exclude-object-pr, I think Troy got a bit busy, would it be acceptable to open a new PR, referencing this one?

troy-jacobson:
@kageurufu Did a fantastic job rewriting the commit history, formatting the documentation, and renaming the commands.
Between his work there and some final tweaks I made, I thing we have all of the requested changes covered.

I've been testing these changes for the past couple of days, including his preprocessor changes that use the new commands.

I opened a new pull request with these updates: https://github.com/Klipper3d/klipper/pull/5326

gurusonwheels:
did this get abandoned ?

maz0r:
> did this get abandoned ?

No its moved to #5413

KevinOConnor:
PR #5413 was merged.

-Kevin

PR #4720, opened 2021-09-24
Add min_power option for fans

HairingX:
An alternative to off_below, min_power raises the fan speeds of commands below min_power up to min_power.
This ensures the fan will always run when commanded to.

Example: My fan starts spinning slowly at 15%. Anything below it and it does not spin.
Using off_below I have to use different filament settings for each of my printers (one starts at 7%, another at 15% and a third around 25%).
With min_power i can adjust each printer fan to start spin at 1% and ramp up to max_power.

Thereby I can use the same filament profile for all printers, as 15% fan is approximately the same across the lineup.
I know there is scaling that differs between them - one fan at 40% blows like another at 60%. - But I can live without that feature for now ;-) 
As long as the fan spins, when we tell it to without having to worry about minimum power needed when working in the slicer.

## Scaling:
This pull request is all about the min_power option that will help me and others I have spoken with about this option.

But I would like to touch upon the scaling option mentioned above, which I of course would love to see in the future - or merged into this branch before merging it into Klipper.
The power scaling should work as such:

#### Machine 1: (5015 blower fan)
- min_power = 0.2
- max_power = 1.0
- **power_scaling = 1.0**

#### Machine 2: (4020 blower fan)
- min_power = 0.2
- max_power = 1.0
- **power_scaling = 1.5**

>Without power_scaling:
  
When machine 1 runs with 50% fan, it will move more air than my 4020 fan at 50%.

>With the power_scaling:

machine 1 - 50% will be translated to 50% * power_scaling = 50%
maching 2 - 50% will be translated to 50% * power_scaling = 75%

It will be easier to have machines match the same output with the scaling functionality.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2302,6 +2302,14 @@ Print cooling fan.
 [fan]
 pin:
 #   Output pin controlling the fan. This parameter must be provided.
+#min_power: 0.0
+#   The minimum power (expressed as a value from 0.0 to 1.0) that makes
+#   the fan spin. If this value is more than 0.0 then fan speed requests
+#   will be scaled between min_power and max_power (for example, if
+#   min_power is .1 and max_power is 1 a fan speed of 10% is requested then the fan
+#   power will be set to 19%)
+#   This setting may be used to prevent fan stalls at low speeds.
+#   The default is 0.0.
 #max_power: 1.0
 #   The maximum power (expressed as a value from 0.0 to 1.0) that the
 #   pin may be set to. The value 1.0 allows the pin to be set fully
@@ -2309,8 +2317,8 @@ pin:
 #   pin to be enabled for no more than half the time. This setting may
 #   be used to limit the total power output (over extended periods) to
 #   the fan. If this value is less than 1.0 then fan speed requests
-#   will be scaled between zero and max_power (for example, if
-#   max_power is .9 and a fan speed of 80% is requested then the fan
+#   will be scaled between min_power and max_power (for example, if
+#   max_power is .9 and min_power is 0 a fan speed of 80% is requested then the fan
 #   power will be set to 72%). The default is 1.0.
 #shutdown_speed: 0
 #   The desired fan speed (expressed as a value from 0.0 to 1.0) if
@@ -2369,6 +2377,7 @@ a shutdown_speed equal to max_power.
 ```
 [heater_fan my_nozzle_fan]
 #pin:
+#min_power:
 #max_power:
 #shutdown_speed:
 #cycle_time:
@@ -2405,6 +2414,7 @@ watched component.
 ```
 [controller_fan my_controller_fan]
 #pin:
+#min_power:
 #max_power:
 #shutdown_speed:
 #cycle_time:
@@ -2450,6 +2460,7 @@ additional information.
 ```
 [temperature_fan my_temp_fan]
 #pin:
+#min_power:
 #max_power:
 #shutdown_speed:
 #cycle_time:
@@ -2497,6 +2508,7 @@ with the SET_FAN_SPEED
 ```
 [fan_generic extruder_partfan]
 #pin:
+#min_power:
 #max_power:
 #shutdown_speed:
 #cycle_time:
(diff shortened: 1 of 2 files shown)


Vhab:
Regarding power scaling, thinking out loud here, would the same not be possible to accomplish by allowing max_power to exceed 1.0, but not actually put more than 1.0 worth of voltage out on the hardware end when the calculated result exceeds 1.0?

Pizzahd88:
Max_power above 1.0 without actually sending more than 1 would do the same yes.

Pizzahd88:
Just a note:
When/If merging this, I should make a pull request for https://github.com/meteyou/mainsail since it uses Klipper configs to show sliders for fan speed, and this needs to know the min_power value to scale properly.

HairingX:
After considering the proposal of using max_power above 1 for scaling a bit more, I would recommend against it. 
If scaling is supported it should be obvious when using it. We therefore still catch typos (value above 1). 
A user might want to limit a fan, and still use scaling. 
So the scaling property is my recommended solution.

HairingX:
@KevinOConnor I have tried changing off_below to be on the raw power instead of user input and had some friends test it out.
We all think having it work on the input value instead of the raw power makes it easier to understand and set values.

## Setup
- min_power is the actual min raw power needed to spin fan.
- max_power is the upper limit raw power
- off_below works on the requested fan speed input

Example:
- User requests fan speed = 0.5 (50%)
Setting min_power = 0.2:
- Power to the fan will be 0.6
Setting off_below = 0.1
- Fan will turn on only if the requested fan speed is 0.1 or above.

## Testing on 3'rd party members
When off_below worked on the power to the fan rather than the input value, the setup was not as easy as working with input values. 
The reason is that they could easily identify the min_power using 0 as initial value, and cranking the fan up till it can only just spin.
After setting min_power it is easy to think of off_below to be the minimum fan spin percentage, as we now have .01 (1%) to be the fan spinning at 1 % speed. Setting off_below = 0.05, will cut the fan off if we request 5% or lower fan speed.

When using off_below on the fan power (raw power), most of them would calculate the spin percentage using the formular min_power is using, which would confuse them. 
To assign the off_below a good value, you would have to set min_power = 0 and ramp up the fan to a speed where you would like the fan to start at, then persist that value and reassign the min_power to the earlier value. This has to be done again if fan over time needs more power to spin at 1% due to degradation. Having off_below work on user input, only min_power needs to be readjusted if that happens.

## Conclusion

I believe keeping off_below as is today (working requested value), will be more intuitive and require less maintenance (if any).
> \#off_below: 0.0
> \#   The minimum input speed which will power the fan (expressed as a
> \#   value from 0.0 to 1.0). When a speed lower than off_below is
> \#   requested the fan will instead be turned off.

HairingX:
@KevinOConnor I have added feature request for this to Mainsail Web UI as the test-group using this new feature is growing a lot.

No errors found, and people are loving the min_power.
Everyone of the 30+ test subjects wants the off_below to be implemented on the user input, as it seems more intuitive.
The web UI's we test on use the off_below to indicate the turn off point on sliders, which works with current implementation.

KevinOConnor:
Thanks.  However, I'm still concerned about its interaction with `off_below`.

The `off_below` option was added as a safety feature to prevent fans from burning themselves out (if power is provided to some fans, but below the minimum power needed to fully turn, then they may vibrate in place, and generate excessive internal heat and/or generate excessive wear on the internal bearings).

It seems to me some refactoring is needed here.  As I don't see value in a safety feature with such subtle behavior.

-Kevin

TheRealOrange:
Alternatively, in order to keep the old configs working, why not introduce min_power which defaults to zero, and apply the off_below threshold to the output generated after the min_power and max_power scaling?

Let me elaborate, if I command a fan speed of F (assume from 0-1 for simplicity) through Gcode, the actual PWM duty cycle is D = (F x (max_power - min_power) + min_power). Now apply the off_below check to this duty cycle value, such that if D is below the threshold in off_below, the PWM output is disabled.

This keeps the safety feature, while introducing a very useful min_power setting.  I noticed that hardware PWM is not recommended for fans, I believe that a much faster cycle time would reduce the buzzing noise as well as allow for smoother fan speed control, but the minimum duty cycle for hardware PWM would be much higher than 0, which I believe necessitates the introduction of min_power.

PR #4722, opened 2021-09-25
docs: Add document for RP2040 spi bus

ChipCE:
Add document for the pins configuration of RP2040 microcontroller.

Signed-off-by: John Smith <chip@snakeoildev.com>

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -10,9 +10,9 @@ documents.
 - [FAQ](FAQ.md): Frequently asked questions.
 - [Releases](Releases.md): The history of Klipper releases.
 - [Config changes](Config_Changes.md): Recent software changes that
-may require users to update their printer config file.
+  may require users to update their printer config file.
 - [Contact](Contact.md): Information on bug reporting and general
-communication with the Klipper developers.
+  communication with the Klipper developers.
 
 ## Installation and Configuration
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  The information looks useful, however I'm not sure if we should add a new document for it.  What would you think about adding this as a section under "Common SPI settings" in docs/Config_Reference.md ?

-Kevin

ChipCE:
@KevinOConnor Thank for the feed back!
I think about that before but it is device specified. So instead of flooding the common config, I made a new file for it. 
I will move it to Config_Reference.md if you think that it the better place.

ChipCE:
@KevinOConnor Does this location sound good to you?
Screenshot_2021-10-10_14-01-47

KevinOConnor:
I would add it to that page but under the existing "Common spi settings" section.

-Kevin

KevinOConnor:
Thanks.  What is the state of this PR?

-Kevin

ChipCE:
> Thanks. What is the state of this PR?
> 
> -Kevin

Sorry, I found out not just the RP2040 but there are no document about the bus name for other MCU. I will submit other PR include document for all supported MCU later. 
I will close this PR for now.
Thank you!

PR #4723, opened 2021-09-25
Add TIMEOUT to TEMPERATURE_WAIT

TimCobra:
Added TIMEOUT parameter to TEMPERATURE_WAIT

I use the TEMPERATURE_WAIT gcode to wait until my chamber's temperature is reached (I only use the bed as a heater).
This will cause my printer to always leave the heaters on, if I for example leave the doors open or accidentaly request a temperature which is unreachable. The timeout should save me from forgetting that I pre-heated the chamber aswell.

The TIMEOUT is off by default so no makros get broken.

I have no real experience with programming but it works for me. Maybe someone can improve / find a better solution.
Or maybe I'm the only one who is that paranoid and forgetful :)

Thanks everyone for your awesome software!
Singed off by: Tim Severin TimCobra@web.de


zellneralex:
With the shutdown you make a very hard event out of the watchdog. Would it not be better and suit much more use cases if you simple leave the command in the same way as it would when reached the target but print a message that timeout was reached.
In that way you could use it in the regular flow as max waiting time before the flow goes on in any case

TimCobra:
Yes you are probably right. My thought was that it should react like a normal heating command, when the thermostat falls out of the block for example. But it might be better your way and I can still stop my printer inside a macro as soon as the timeout elapses.

Doesn't the idle_timeout only work if the printer is "idle"? It's "busy" in this case and would heat forever in my understanding

zellneralex:
Normally if you only hear the printer is idle but I did not check the wait command

TimCobra:
Yeah it's kind of wierd. I't locks up everything and you have to shutdown klipper if you want to cancel, so klipper is defintely is busy (Fluidd  also says so)

dalegaard:
I've been running a patch similar to this on my own Klipper fork and been very happy with it.

Issue #4675, opened 2021-09-10
Losing communication with MCU during resonance testing


charlespick:
I'm getting a number of issues running the resonance tester after the recent adxl345 commits. I rolled back to 6999ff6 and it began working again. By X axis tests fine but my Y axis always crashes somewhere around 80-100hz. 
[klippy.log](https://github.com/KevinOConnor/klipper/files/7140552/klippy.log)
<img width="2086" alt="Screen Shot 2021-09-09 at 5 48 54 PM" src="https://user-images.githubusercontent.com/17918019/132782295-b84ca1a6-87b9-4fa2-b4b0-4bed35bc12fd.png">
<img width="2086" alt="Screen Shot 2021-09-09 at 5 49 08 PM" src="https://user-images.githubusercontent.com/17918019/132782317-caa23a3d-5ba2-4083-bd1c-af8c4b73954c.png">

Talha909:
now a days every has the same issue. May be it will solved in coming days.  Which type or raspbery pi u r using.

charlespick:
rpi 4

KevinOConnor:
That's unfortunate and I don't know why that would be.  We don't use github to help diagnose problems with a printer though.  Best to see if someone on Discord or Discourse can assist.

-Kevin

charlespick:
@KevinOConnor I don't think this is a problem with my printer. This is only happening after the recent changes to adxl345. Rolling back and changing nothing about the printer solves the problem.

charlespick:
thanks, bot

dmbutyugin:
@KevinOConnor, today I've finally upgraded my installation to the the latest ADXL345 measuring code, and it seems that we are maybe on to something here. I checked the performance and the system load at the current head (59bd51d207747efcdd37b5577f817cd95ff5edbb) and before the changes to the adxl345 code (6999ff6256514474d2a309c76cef18e26acf2969). I ran the same resonance testing sequence
```
G28
TEST_RESONANCES AXIS=Y OUTPUT=raw_data,resonances NAME=...
M112 ; after all test processing is finished
```
after rebuilding and reflashing the MCU code for all MCUs, deleting `klippy.log` and restarting Klipper with the aforementioned versions of the code. The ADXL345 is connected to the Raspberry Pi 3 Model B+ (printer is Ender 3). Then I used the `scripts/graphstats.py -s` command to generate the sytem load charts, which are:

[...]

PR #4725, opened 2021-09-26
adxl345: small fixes of the API

dmbutyugin:


dmbutyugin:
Kevin, this is a first pass making some really small fixes and improvements. I didn't do any work to improve error reporting in the console, as I wasn't yet sure how you [prefer](https://github.com/Klipper3d/klipper/issues/4675#issuecomment-926676057) to proceed with the error reporting.

KevinOConnor:
> If we believe that after the fix to the error propagation it is no longer possible to occur that the measurement results are empty, then these checks are probably unnecessary, and we could remove them all for consistency (otherwise it does look weird that they are used inconsistently).

I believe it's technically still possible to complete with no data, but it would be rare.  All the initial chip configuration would need to succeed and there would need to be errors on every sample read.

> Then get_samples is indeed called in the main thread, but now it is much more lightweight (that's one of the reasons I renamed it, because it does not do any 'decoding' anymore).

Okay.  I'm not sure it would be lightweight though, because Python would still need to iterate through every sample - which could be 100s of thousands.  Unfortunately, Python is pretty slow - in particular with large lists.

> However, if you prefer, if not aclient.get_samples(): could be changed to a more light-weight check to see if there are any valid samples within [self.request_start_time, self.request_end_time] range.

I think that might be safer, but it's up to you.

-Kevin

dmbutyugin:
So I wrote a light-weight test that should be good enough, and made the `shaper_calibrate.py` processing a bit more robust. At this state, the code should never generate any unintended run-time errors (like out-of-bound errors), and instead generate helpful error messages.

PR #4727, opened 2021-09-27
gcode_macro: Add "rawparams" pseudo-variable

pedrolamas:
This adds a new `rawparams` pseudo-variable to the `gcode_macro` template context.

The main usage for this is to allow easier override of an existing G-Code like `M117`, where we don't have any parameters (so `params` is useless) but instead take the whole message after the gcode.

Here's an example of it:

```yaml
[gcode_macro M117]
rename_existing: M117.1
gcode:
  M117.1 { rawparams }
  M118 { rawparams }
```

In the above example, any message sent via `M117` would also be mirrored with `M118`.

### DIFF
--- a/docs/Command_Templates.md
+++ b/docs/Command_Templates.md
@@ -129,6 +129,20 @@ gcode:
   M140 S{bed_temp}
 ```
 
+### The "rawparams" variable
+
+The full unparsed parameters for the running macro can be access via the `rawparams` pseudo-variable.
+
+This is quite useful if you want to change the behavior of certain commands like the `M117`. For example:
+
+```
+[gcode_macro M117]
+rename_existing: M117.1
+gcode:
+  M117.1 { rawparams }
+  M118 { rawparams }
+```
+
 ### The "printer" Variable
 
 It is possible to inspect (and alter) the current state of the printer
--- a/klippy/extras/display_status.py
+++ b/klippy/extras/display_status.py
@@ -35,19 +35,8 @@ class DisplayStatus:
         curtime = self.printer.get_reactor().monotonic()
         self.expire_progress = curtime + M73_TIMEOUT
     def cmd_M117(self, gcmd):
-        msg = gcmd.get_commandline()
-        umsg = msg.upper()
-        if not umsg.startswith('M117'):
-            # Parse out additional info if M117 recd during a print
-            start = umsg.find('M117')
-            end = msg.rfind('*')
-            if end >= 0:
-                msg = msg[:end]
-            msg = msg[start:]
-        if len(msg) > 5:
-            self.message = msg[5:]
-        else:
-            self.message = None
+        msg = gcmd.get_raw_command_parameters() or None
+        self.message = msg
 
 def load_config(config):
     return DisplayStatus(config)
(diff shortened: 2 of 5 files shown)


pedrolamas:
Took me a while to get the whole "line number" thing, but I think I understand now!

Something like `N1 M117 test*123` would be the case, correct?

Given that, this is how one could code it in a template:

```yaml
[gcode_macro M117]
rename_existing: M117.1
gcode:
  {% set message = commandline %}
  {% if not message.startswith("M117") %}
    {% set start = message.find("M117") %}
    {% set end = message.rfind('*') %}
    {% if end >= 0 %}
      {% set message = message[:end] %}
    {% endif%}
    {% set message = message[start:] %}
  {% endif %}
  M117.1 { message }
  M118 { message }
```

But that does seem way to much verbose to me, so I think something better should be done on Klipper side... I will follow up on your comments and see how I can improve this!

pedrolamas:
@KevinOConnor I've added a new `get_sanitizedcommandline()` method that takes the command and removes the line number and checksum value (plus trims the spaces) so that can be used directly for the M117, M118 and to easier command override.

I considered changing the existing `_commandline` and `get_commandline()` method directly, but I was a bit afraid that it could break something so I opted for adding a separate one instead.

pedrolamas:
Given that, I changed approach: instead of having the full command line as a variable, I'm just adding a `rawparams` that has the unparsed parameters (using the M117/M118 approach with a few minor touches)

pedrolamas:
I've updated the PR title and description to match these changes.

pedrolamas:
@KevinOConnor just wanted to check if you think this needs more work or even a different approach to the current way I've implemented it?

whi-tw:
Hmm, turns out I didn’t see this when making #4883. I went for the least intrusive way by not adding any extra functions and just passing the whole command line in, leaving it to the macro to sanitise the params

pedrolamas:
@whi-tw that was actually my initial approach but after some discussion, @KevinOConnor pointed out some problems with that as the commandline (gcode) can include line numbers and checksums, hence why I ended with the proposed implementation in this PR.

whi-tw:
@pedrolamas for sure, makes sense, although it depends where the responsibility for parsing the variable lies. I went for the responsibility being on the macro author, vs Klipper itself, as it’s feasible the line number could be helpful to someone?

KevinOConnor:
Thanks.  Looks fine to me.

One question - is there a reason that `gcmd.get_raw_command_parameters()` returns None when there are no parameters?  I would think that would be awkward for macros to handle.  The special M117 case could just do `if not raw_commands: raw_commands = None`?

-Kevin

pedrolamas:
> is there a reason that `gcmd.get_raw_command_parameters()` returns None when there are no parameters?

`None` can easily be handled inside the macro with `{ rawparams | default }`, but I am ok to change it to something else if you think there's a better way here!

KevinOConnor:
I'd say always returning a string would be preferable (and add the additional None logic to the M117 code).  Otherwise, we'd need to document that `{rawparams}` can return None.

Thanks,
-Kevin

pedrolamas:
No worries, I just pushed the necessary changes to make `get_raw_command_parameters()` return `''` as discussed!

KevinOConnor:
Thanks.

-Kevin

PR #4731, opened 2021-09-29
Update BLTouch.md to prevent position_endstop error

Agilato:
module: BLTOUCH DOCUMENTATION

Klipper with Fluid complains about `position_endstop` line when `endstop_pin: probe:z_virtual_endstop` is used.

Signed-off-by: Ilia Rebane <Agilatosay@gmail.com>


KevinOConnor:
Thanks.  Change looks fine, but commits need to have a signed-off-by line containing your full real name - see: https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

I can change the documentation with my signed-off-by line if you prefer though.

-Kevin

PR #4732, opened 2021-09-29
Update generic-printrboard.cfg to mention different Y endstop pin on RevF hardware

creyc:
Just a simple comment which may help others with a RevF or F4 board - the Y endstop pin is different from RevD boards.

Since there's already comments for other RevF changes I figured this might be appropriate to include as well.

### DIFF
--- a/config/generic-printrboard.cfg
+++ b/config/generic-printrboard.cfg
@@ -29,6 +29,8 @@ enable_pin: !PE6
 microsteps: 16
 rotation_distance: 40
 endstop_pin: ^PB0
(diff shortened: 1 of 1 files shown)


creyc (commented):
removed extra whitespace.

KevinOConnor:
Thanks.  Commits need to have a signed-off-by line containing your full real name for me to commit directly - see: https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md .

I can add a comment to this config using my signed-off-by line if you prefer though.

-Kevin

PR #4548, opened 2021-07-30
Anycubic Vyper

mneuhaus:
This adds support and example config for the Anycubic Vyper.

Discussion is happening in Discord (#misc-mainboard)

PR #4733, opened 2021-09-29
Anycubic Vyper

mneuhaus:
This adds support and example config for the Anycubic Vyper.

Continuation of: https://github.com/Klipper3d/klipper/pull/4548

Thanks to @Crykor90 for determining the gear_ratio and verifiying on his machine!


sstickel:
Z1 endstop pin is commented out.

mneuhaus:
> Z1 endstop pin is commented out.

Anycubic Marlin actually uses only one of the Endstops (https://github.com/ANYCUBIC-3D/Vyper/blob/vyper/Marlin/src/pins/stm32f1/pins_TRIGORILLA_PRO.h#L61)

I tried to use both but had problems with triggering the other one when homing with the other, that's why i only use one like the Marlin firmware.

PR #4736, opened 2021-09-29
docs: background color fix for bed_mesh svgs

Arksine:
This adds an opaque background to the 3 svgs in `Bed_Mesh.md` that were previously transparent.


KevinOConnor:
Thanks!

-Kevin

PR #4737, opened 2021-09-29
removed forced spaces in respond

chestwood96:
Different take on #4664

I moved the spaces into the default prefixes so it is possible to send responses without a forced space. This is required for some octoprint plugins.

Side-effect:  For people that used custom prefixes before it will slightly change the output of their responses (no space between prefix and msg)

Signed-off-by: Adrian Joachim adi.joachim12@gmail.com

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1407,6 +1407,7 @@ Enable the "M118" and "RESPOND" extended
 #   Sets the default prefix of the "M118" and "RESPOND" output to one
 #   of the following:
 #       echo: "echo: " (This is the default)
+#       echo_no_space: "echo:"
 #       command: "// "
 #       error: "!! "
 #default_prefix: echo:
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -593,6 +593,8 @@ The following commands are availabe when the
   configured default prefix (or `echo: ` if no prefix is configured).
 - `RESPOND TYPE=echo MSG="<message>"`: echo the message prepended with
   `echo: `.
+- `RESPOND TYPE=echo_no_space MSG="<message>"`: echo the message prepended with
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Thanks.  It shouldn't be necessary to change the existing users - just add a space to the user defined prefix - something like:
```python
if config.get("default_prefix", None) is not None:
    self.default_prefix = config.get("default_prefix") + " "
```
and similar for the gcode PREFIX parameter.

You can add also add a new "echo_no_space" type for your use case.

Separately, I'd be inclined to remove the "default_prefix" and PREFIX capabilities, as I'm very leery of generating "ok" messages or similar that could confuse the g-code protocol.  It's a "can of worms" I don't want to open.

-Kevin

chestwood96:
My bad, I misunderstood your previous comment. That does indeed work too.

KevinOConnor:
Thanks.  Unless I'm missing something, this documents that `echo_no_space` is valid in the config but the updated doesn't actually support that?

-Kevin

chestwood96:
Unless I messed up something, adding `echo_no_space` as a valid option is pretty much the only thing this update does.

chestwood96:
@KevinOConnor Can this be reopened? I was just waiting for your feedback, it is otherwise done as far as I know.

KevinOConnor:
It doesn't look like my comment at https://github.com/Klipper3d/klipper/pull/4737#issuecomment-974148880 has been addressed.  Specifically, the documentation says one can use `default_type: echo_no_space`, but it seems to me that would raise an error.  Perhaps I've missed something.

-Kevin

chestwood96:
>Specifically, the documentation says one can use `default_type: echo_no_space`, but it seems to me that would raise an error. Perhaps I've missed something.

Oh right, default prefix does not work, I'll remove that from the doc next week, other than that, can it get reopened?

chestwood96:
@KevinOConnor Fixed the config doc but it does not update here, do I need to make a new pull request?

PR #4738, opened 2021-09-29
introduce load cell-based probe

mhier:
As announced in #3496, this PR introduces support for probes based on load cells.


KevinOConnor:
Interesting, thanks.  I took a look through the code and have some high-level comments.

My primary observation is that this code "feels" like it is still an experimental feature.  For example, the code and docs mention some unusual use cases - probing xy, "force subscriptions", measuring extruder force.  And there are config options that seem like development/debugging capabilities (eg, fit parameters, oversampling options).  It's also not clear to me if this software would work on "arbitarary load cell" systems or if it is specific to a particular hardware implementation.  As per the Klipper guidelines ( https://www.klipper3d.org/CONTRIBUTING.html ) we don't want to merge experimental features (or sub-features) into the master Klipper repository.

I know it has been some time since you submitted this module, andI'm not sure if you're still interested in merging this code.  If you do want to merge, I think it would be helpful to separate out the changes into multiple commits - for example, start with the ads1000.py support (which needs docs and needs to implement important calls like setup_minmax() ), and then a commit with the core load cell probe support.  Additional features can always be added in later PRs.

Thanks again,
-Kevin

KevinOConnor:
> The point is a bit what "experimental" means in this context. The feature is IMO not experimental when running on a Renkforce RFx000 printer.

To be clear, I did not mean to imply that I think the code "needs testing".  I understand that you and others are using this module and that it has utility.  I agree that extending Klipper to support load cell based probes would be a useful addition.

During development it's common to code debugging, tuning, and experimental processes to see what capabilities the hardware can support.  When I look at this PR, I get the impression that lots of development and experimentation went into getting a solid working system.  For the Klipper master branch, though, I wonder if it is possible to refactor all that work to the "core functionality".

Said another way, while reviewing I try to keep the core goal of the submission in mind (eg, "load cell based probe"), then try to reason how each config option supports that goal, how users will set/tune each config option, and how each line of code facilitates the goal.  I've written a little about this at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review .  So, as an example, there's a lot of config options in this PR and it's not clear to me how a user would set each of them.  I suspect many of the config options could be replaced with Python module code constants.

> Well, that's the point of the many parameters, to make it work on (more or less) arbitrary load cell systems. If those parameters were hard coded to a specific value, it would probably not work on different setups.

FWIW, I'm confident it will be a "developer" that does the work of getting it running on another machine.  In my experience, lots of user facing tuning options make that developer's job harder - because if some code change is necessary then not only do they need to make that code change, but they have to test it with every possible user facing tuning parameter..

> So if you agree that this feature is not considered experimental and hence can be merged in principle as is, I would split the PR into smaller chunks. If it helps, I can also leave out the XY probing parts at first and hand them in separately (although the algorithm is to some extend designed with XY probing in mind already, which I of course can't change).

Thanks.  If it's possible to break this PR up into multiple chunks I think that would help a lot.  We can always add features on top.

I don't want to "spin your wheels" though.  Maybe we could go over the core user-facing commands and config parameters, so that we're both on the "same page".

> I was afraid that just providing e.g. the ads1100.py would seem pointless without the rest of the code, especially since it provides a "bigger" interface (get_last_value() and read_single_value()) - which is important as the load cell probe needs to synchronise the ADC reads with the movement.

Can you elaborate on why a change to the "adc pin" api is needed?  Could the load_cell_probe.py code subscribe to the adc updates and locally record the last measurement/time?  Maybe I'm missing something.

Cheers,
-Kevin

mhier:
I think I get your point now. The module offers many parameters which are not user parameters once running on a given machine. It may also output some information to the gcode console which is uninteresting for normal users. On the other hand, if I completely remove that, I don't see how someone (the developer you mentioned) could possibly integrate the module with a printer model without changing Python code.

Is there some way to introduce configuration parameters which are model/machine specific rather than user options? Let's make a simple example: depending on the used type of load cells and ADC, the signal response to a small nozzle-to-bed contact force could be 5 digits or 5000 digits. There has to be some parameter reflecting such difference. In principle all parameters (maybe except for the 4 "general" parameters in the config reference documentation) are on that level. They likely depend on the hardware setup, which will be different in every printer model, but same for everyone with the same hardware. (Of course there may be people retrofitting load cells to their printer, which kind of will create one-of-its-kind printer models... We should not exclude them, IMO Klipper is the perfect choice for such highly customised printers!)

One alternative could be to try to reduce as many as possible parameters to either constants or derived values from very few higher level parameters. This might create problems when trying to use the module on different hardware, but maybe it is sufficient to introduce new configuration parameters only when actually found necessary later. E.g. there could be one common scaling factor for all parameters which are in units of ADC counts. The ratios between those parameters would then be fixed (via Python constants). If I did the counting right, this would unify already 5 parameters into one. If found later that one if the 5 parameters needs to be independent, we could add a true config parameter for its ratio to the common scaling factor. Other parameters like the fit quality requirement could be just completely fixed. Please note that this would make this module less flexible, but maybe that's the lesser evil?

About the console output: It might be helpful to introduce a configuration parameter (this time for the user) to enable debug output. There will be some output necessary though, since the scans take quite some time and e.g. Octoprint will run into a timeout without any visible output - but this could be some simple "still working..."  message or so.


> Can you elaborate on why a change to the "adc pin" api is needed? Could the load_cell_probe.py code subscribe to the adc updates and locally record the last measurement/time? Maybe I'm missing something.

This is unfortunately owed to a peculiarity of the RFx000 printer series's hardware: The used ADS1100 ADC is very slow, it can do only 8 Hz at full resolution. On the other hand, each conversion can be triggered by the software at any desired point in time. When using the standard subscription mechanism, the load_cell_probe code has no influence on when the conversion would take place, as it happens asynchronously in a different module (that's my understanding at least). Hence the load_cell_probe code would have to wait always one extra conversion to be sure that the measurement was started after the motion has stopped. Therefore I have introduced a synchronous interface which allows to control when the conversion is started.

I am not fully satisfied with this solution either. Recently I have looked into using a HX711 (mainly to simplify using this module for others, as one can get ready-to-use boards with that chip which can be connected to most printer mainboards - I would not like to abandon the original RFx000 printer hardware). That chip can only do continuous conversions, so one cannot make use of the synchronous interface anyway. On the other hand it does not matter so much with that chip as it can do a much faster frequency (80 Hz).

The "perfect" solution would probably be to have the conversions somehow scheduled properly in a similar way as the motor steps. That might even eliminate the need for waiting until the movement has stopped. It will be a lot more work though, as it involves writing MCU drivers for the ADS1100 resp. HX711. If this is ok for you I might do this later as a feature upgrade (otherwise nothing would be merged until in two years from now or so :-D I have too many "construction sites" already...).

KevinOConnor:
> Is there some way to introduce configuration parameters which are model/machine specific rather than user options? Let's make a simple example: depending on the used type of load cells and ADC, the signal response to a small nozzle-to-bed contact force could be 5 digits or 5000 digits.

It's a challenge for me to give suggestions, because I'm not familiar with all the details of load cell probing.

I'm not understanding your point above - what do you mean by a "digit" in this context?  In Klipper today we normalize ADC readings to a relative value between 0.0 and 1.0.

I would tend to agree that there would need to be some kind of user facing configuration for a "force threshold" to distinguish between bed not pressed and pressed.

A lot of config options in this PR I don't understand.  Some examples:  Why would a user want to configure `adc_n_average` and how would they do so?  Wouldn't they always want to average all the valid readings?  Why (and how) would they configure `step_size`?  If the module has been tested at 0.1mm what gain is there in configuring an untested distance?

> Please note that this would make this module less flexible, but maybe that's the lesser evil?

I'm not sure if that was a question to me - if so I don't really understand what you're proposing.

> It might be helpful to introduce a configuration parameter (this time for the user) to enable debug output. 

FWIW, I think the g-code console is basically broken.  It's not a user-friendly interface nor a developer friendly interface.  That said, maybe you could describe what information you envision a user would obtain and what they would do with that information.  Again, it's a challenge for me to give feedback because I'm not that familiar with the low-level details.

> This is unfortunately owed to a peculiarity of the RFx000 printer series's hardware: The used ADS1100 ADC is very slow, it can do only 8 Hz at full resolution. [...] Therefore I have introduced a synchronous interface which allows to control when the conversion is started.

Okay, I understand.  However, I'm a little confused on the goals here.  It seems the ads1100 would only be used for load cell probing, the load cell code can't use "mcu adc pins", and it seems the Klipper "mcu adc pin" interface is insufficient for load cell probing.  But why implement the "mcu adc pin" interface for the ads1100 then?  Could one implement a separate "load cell adc" interface that is sufficient?

Separately, just out of curiosity, is the ads1100 really more accurate at 8sps or is it just implementing oversampling?  I guess it doesn't really matter for measurements during standstill, but I'd have thought 16x more measurements would be better than 4-bits of additional precision.

Thanks,
-Kevin

KevinOConnor:
Let me see if I understand the "core algorithm" of this load cell probing implementation: the nozzle starts away from the bed and descends in steps, if the measured force increases by a predetermined threshold then the nozzle is considered to have made contact with the bed, the nozzle descends an additional 4 steps, a linear regression of the force per distance over the "in contact steps" is calculated, the average force of "pre-contact steps" is calculated, the resulting bed position is the intersection of those two calculated "lines".

Did I understand that correctly?

It seems there are also some "secondary goals" - faster descent speeds prior to initial contact (as an optimization to reduce probing time), XY probing, etc.

If I've understood that correctly, as a user, I guess I'd expect to see the following main config options:
1. The force threshold increase (ideally in Newtons or Kilograms) that determines a contact has been made.  I'd expect I'd need to set this as it could vary from machine to machine.
2. The maximum distance (in mm) to depress the bed after contact is detected.  A larger value may provide more accurate readings, but a too high value may cause damage to the machine.
3. Maybe a descent speed (in mm/s) and/or an initial descent speed.  A higher speed may reduce probing time, but increase risk of damage.  Though I guess it may not be worthwhile to configure this if the overall speed is already so slow that trying to optimize it isn't practical for regular users.

Am I missing something, or are there other fundamental physical parameters to the "core algorithm"?

FWIW, I don't see how regular users can meaningfully configure the "fit", "average", and other math parameters.  I don't see why a user would want to configure some of the distance parameters.  These seem like they should be code constants to me.  It's pretty trivial to change a python code constant, so it's not outside the reach of power users should they need to investigate something less common.

> In my 16 bit ADC setup, the typical forces during the probing measurements are several 100 "digits", or maybe a few 1000.

I'd avoid using "arbitrary units" for configuring.

If you don't want to use 0.0 to 1.0 then maybe add a scaling config parameter so users can scale the value to Newtons or Kilograms and use those "real units" in the rest of the config.  Working with "real physical units" is much easier to understand and work with.

> Basically the question is whether we want this module to be as independent as possible of the hardware setup, or whether we should make it as simple as possible at first and later see how to extend it to different hardware setups.

In general, I'd say the latter.  Adding code for possible future hardware often increases the effort to support new hardware - as the future developers not only need to add support for their hardware, but also understand and test what may be unused code.

That said, I'd certainly try to keep the code reasonably modular (eg, keeping the ads1000 code separate) and use config settings or code constants for obvious physical parameters of the system.

> Yes we could do that. On the other hand, the ADS1100 chip is a general purpose ADC.

Okay.  Just be sure to fully implements the interface.  (Otherwise, it raises the risk of a user configuring it for a heater while it silently does not implement important safety checks.)

-Kevin

mhier:
> Let me see if I understand the "core algorithm" of this load cell probing implementation: the nozzle starts away from the bed and descends in steps, if the measured force increases by a predetermined threshold then the nozzle is considered to have made contact with the bed, the nozzle descends an additional 4 steps, a linear regression of the force per distance over the "in contact steps" is calculated, the average force of "pre-contact steps" is calculated, the resulting bed position is the intersection of those two calculated "lines".
> 
> Did I understand that correctly?

Partially correct. When the first contact is found, the nozzle backs up a bit until it has no contact any more. Only then the (minimum) 5 additional steps are made to do the linear regression (more steps will be used if the first measurements stay below a threshold, because the nozzle moved away a bit too far). Also important: Each measurement step used for the linear regression is "compensated" with a second measurement where the nozzle has no contact with the bed. So the nozzle will jump up and down for every measurement.

I have tried to explain this in more detail here (not sure if you have seen it already):
https://github.com/RF1000community/klipper/blob/master/docs/LoadCellProbe.md#working-principle-of-the-probe-algorithm


> It seems there are also some "secondary goals" - faster descent speeds prior to initial contact (as an optimization to reduce probing time), XY probing, etc.

Exactly. Especially speed is a concern, that is the reason for these 3 phases: First phase finds rough position as fast as possible, second phase finds start position of the fit (as close as possible above the bed), third phase is slow but gives the full precision.


> If I've understood that correctly, as a user, I guess I'd expect to see the following main config options:
> 
>     1. The "force increase" threshold (ideally in Newtons or Kilograms) that determines a contact has been made.  I'd expect I'd need to set this as it could vary from machine to machine.

So far I have never calibrated this to Newtons or Kilograms, but I slowly start thinking it might be a good idea to do that. It would probably be much easier to determine the parameters then.

So this parameter is the `threshold` parameter.


>     2. The maximum distance (in mm) to depress the bed after contact is detected.  A larger value may provide more accurate readings, but a too high value may cause damage to the machine.
> 
>     3. Maybe a descent speed (in mm/s) and/or an initial descent speed.  A higher speed may reduce probing time, but increase risk of damage.  Though I guess it may not be worthwhile to configure this if the overall speed is already so slow that trying to optimize it isn't practical for regular users.

These two are actually somewhat connected. The movement of the initial decent is not continuous, since the measurements need to be done at a defined position (and we have latencies between Python and MCU code). Hence a very important quantity is the step size of the initial approach. The speed of each step is basically not important (the faster the better) and the wait/measurement time in between cannot really be influenced.

So basically this is the `step_size` parameter.


> Am I missing something, or are there other fundamental physical parameters to the "core algorithm"?

Well, we have a smaller step size for the steps done in the fit (-> `fit_step_size`). That one might be difficult to derive from more fundamental parameters, as in my experience there might be some non-linearities in some cases even when staying far away from any dangerous force (My guess is: I have a removable build plate on top of the actual heating bed, and there may be small gaps in some areas between the build plate and the bed below. If the build plate is pressed down until the gap is no longer there, the effective spring constant will change. It is important not to reach that point, because that spoils the precision completely. I think many people will have this problem, so it's not a super special case...)

Also there is a safety parameter `max_abs_force`, which will cause the algorithm to stop if the forces get too big.

Apart from this, I think only the Z lift distance for the compensation measurement in case of XY probing is important, since this really depends on the shape of the probing tool and sometimes even on the shape of the workpart. This is an advanced use case anyway, I think we can leave it out of the discussion for now.

I think the other fit parameters can be derived from the others or constant. fit_threshold is not unimportant but could be some fraction of the threshold, maybe the ratio `fit_step_size/step_size`.


> FWIW, I don't see how regular users can meaningfully configure the "fit", "average", and other math parameters. I don't see why a user would want to configure some of the distance parameters. These seem like they should be code constants to me. It's pretty trivial to change a python code constant, so it's not outside the reach of power users should they need to investigate something less common.

I agree.


> > In my 16 bit ADC setup, the typical forces during the probing measurements are several 100 "digits", or maybe a few 1000.
> 
> I'd avoid using "arbitrary units" for configuring.
> 
> If you don't want to use 0.0 to 1.0 then maybe add a scaling config parameter so users can scale the value to Newtons or Kilograms and use those "real units" in the rest of the config. Working with "real physical units" is much easier to understand and work with.

I also agree :-)


> > Basically the question is whether we want this module to be as independent as possible of the hardware setup, or whether we should make it as simple as possible at first and later see how to extend it to different hardware setups.
> 
> In general, I'd say the latter. Adding code for possible future hardware often increases the effort to support new hardware - as the future developers not only need to add support for their hardware, but also understand and test what may be unused code.

Ok, I understand.


> That said, I'd certainly try to keep the code reasonably modular (eg, keeping the ads1000 code separate) and use config settings or code constants for obvious physical parameters of the system.

Sure.


> > Yes we could do that. On the other hand, the ADS1100 chip is a general purpose ADC.
> 
> Okay. Just be sure to fully implement the interface. (Otherwise, it raises the risk of a user configuring it for a heater while it silently does not implement important safety checks.)

All right. Thanks a lot for the input! I think I now know, what I have to do :-) It will take a while though to get through all this...

KevinOConnor:
> I have tried to explain this in more detail here (not sure if you have seen it already):
https://github.com/RF1000community/klipper/blob/master/docs/LoadCellProbe.md#working-principle-of-the-probe-algorithm

I did see that and it is helpful.  My message above was an attempt to summarize the above in "my own words" to see if I understood it.

I missed that there is a Z lift on each "fit measurement".  Interesting.

> Hence a very important quantity is the step size of the initial approach.

Take this with "a grain of salt" as I have little knowledge of this system.  However, as a user, if I supplied the "maximum distance I'm willing to depress the bed" then I'd have thought the code could infer a reasonable "fast descent step size".

> Well, we have a smaller step size for the steps done in the fit (-> fit_step_size). That one might be difficult to derive from more fundamental parameters

If we know 5 fit samples are needed, know that maybe the first 3 fit_step_size may not yet have depressed the bed, and know the "maximum distance to depress the bed", then could one reasonably calculate `fit_step_size = max_depress_distance / 8`?  Again, feel free to completely disregard this if it doesn't make sense.

> It will take a while though to get through all this...

I understand.  I appreciate your work and feedback.

-Kevin

mhier:
> I missed that there is a Z lift on each "fit measurement". Interesting.

That is important at least as long as the total time of the measurements used for the fit is _not_ much shorter than the typical time the measurements are stable enough (without drift) At least on my printer this is clearly not the case. (For XY probing, it is not only about time but also about the movements.)


> > Hence a very important quantity is the step size of the initial approach.
> 
> Take this with "a grain of salt" as I have little knowledge of this system. However, as a user, if I supplied the "maximum distance I'm willing to depress the bed" then I'd have thought the code could infer a reasonable "fast descent step size".

True. This is more or less the same parameter, maybe with something like a fixed factor of 2 or so to provide some safety margin. For the fast decent, we want the step size as big as safely possible to achieve the fastest possible speed.


> > Well, we have a smaller step size for the steps done in the fit (-> fit_step_size). That one might be difficult to derive from more fundamental parameters
> 
> If we know 5 fit samples are needed, know that maybe the first 3 fit_step_size may not yet have depressed the bed, and know the "maximum distance to depress the bed", then could one reasonably calculate `fit_step_size = max_depress_distance / 8`? 

No, because for the fit we want the step size resp. the total measurement range to be as small as possible without degrading the precision. The fit is assuming that the measured force is proportional to the distance, which is not perfectly the case. Hence (in contrast to what intuition tells us) the smaller we make the step size the better will be our precision. This is of course up to the point where we cannot resolve the difference of the force any more or where the Z axis cannot precisely do the right step size (-> microsteps are usually not equidistant inside the full steps, so going below a full step on the Z axis may not be a good idea).

This is based on my observations on my printer model. I can probably provide a prescription how to determine this step size on an arbitrary printer model. Maybe I should create a calibration command which does this automatically?

mhier:
still active and working on it ;-)

KevinOConnor:
Okay, thanks.  I'll put it back in the open state.  But, note that we can always reopen it when you are ready as well.

-Kevin

mhier:
I am getting close to finish a very much improved implementation. I can now derive most parameters in a semi-automatic way. Mainly, only one parameter has to be chosen fully by the "user" (i.e. the person setting up the configuration for the printer model), which is the maximum acceptable force (in physical units). This works by executing a series of special calibration commands. IMO this feels now a lot more mature than with "guessed" parameters, and even the speed has increased :-)

I have now some questions though:
- Can I rely that all ADCs in Klipper give values in the range 0 to 1 or -1 to 1?
- I need to know the value of an LSB (least significant bit) to understand the quantisation effects. I don't think there is an automated way to get this from the ADC module currently. Should I add this to the ADC interface, or expect this as an input from the "user"?
- Is it ok to expect numpy to be installed? Shall I just add it to the `klippy-requirements.txt`?

KevinOConnor:
> Can I rely that all ADCs in Klipper give values in the range 0 to 1 or -1 to 1?

The ADC code always reports values between 0.0 and 1.0 today.

> I need to know the value of an LSB (least significant bit) to understand the quantisation effects.

This is exported by the mcu code as `ADC_MAX`; it isn't exposed beyond `mcu.py:MCU_ADC()` though.

> I need to know how long to wait until the next sample.

This is the report_time parameter of `setup_adc_callback()`.

> Is it ok to expect numpy to be installed?

It's not normally installed, as the installation time on Python2 builds takes too long.  A few modules require it today, but they are normally the exception.

Cheers,
-Kevin

mhier:
> > Can I rely that all ADCs in Klipper give values in the range 0 to 1 or -1 to 1?
> 
> The ADC code always reports values between 0.0 and 1.0 today.

So there are probably no signed ADCs supported yet, right? I guess -1.0 to 1.0 for them would be the right choice then?


> > I need to know the value of an LSB (least significant bit) to understand the quantisation effects.
> 
> This is exported by the mcu code as `ADC_MAX`; it isn't exposed beyond `mcu.py:MCU_ADC()` though.

Ok, maybe we can simply extend the Python interface to expose it. The current ADS1100 implementation does not even have any MCU code, and there may be other ADCs like that I guess?


> > I need to know how long to wait until the next sample.
> 
> This is the report_time parameter of `setup_adc_callback()`.

This is in my understanding "the wrong way round". The load cell probe module calls setup_adc_callback() on the ADC module to subscribe to the value and can specify the wanted report_time. I need to know the actual report_time, which may differ. Or should I expect from the "user" (again I mean the one who sets up the probe parameters for the printer model) to specify a correct report time / frequency as a parameter?



> > Is it ok to expect numpy to be installed?
> 
> It's not normally installed, as the installation time on Python2 builds takes too long. A few modules require it today, but they are normally the exception.

Ok, I understand (speed is not so much an issue for me, since I am working on a desktop machine ;-)). Do you think it is better to avoid numpy then and instead provide my own implementation e.g. in mathutil.py? This is rather simple stuff like standard deviation. I already implemented a linear fit in mathutils.py, but I am wondering whether it is better to use the numpy implementation instead...

If you are saying, build time on Python2 is slow: Is it much better on Python3? I guess in some not too distant future Klipper will switch to Python3 anyway, so this may be an argument in favour of numpy...

KevinOConnor:
> So there are probably no signed ADCs supported yet, right? I guess -1.0 to 1.0 for them would be the right choice then?

FWIW, I suspect trying to mix the existing "mcu ADC" code with your new "force sensor ADC" code is not going to work well.  I suspect it will be simpler and easier to maintain if the two are kept separate.

> Do you think it is better to avoid numpy then and instead provide my own implementation e.g. in mathutil.py? This is rather simple stuff like standard deviation.

For simple things (like standard deviation) it's probably better to just implement it directly.  If it's a complex math algorithm (eg, FFT) then I'd use numpy.

-Kevin

mhier:
> FWIW, I suspect trying to mix the existing "mcu ADC" code with your new "force sensor ADC" code is not going to work well. I suspect it will be simpler and easier to maintain if the two are kept separate.

Generally, there are signed (differential) ADCs and unsigned (single-ended) ADCs. So far, the ADCs used in Klipper are unsigned, because this is sufficient for temperature monitoring. For load cell readout, signed ADCs need to be used, because the signal is differential by nature. Apart from that distinction, the ADCs are pretty much standard. Even the HX711, which is designed specifically for weight scales, can be used for other purposes. Moreover, it should be possible to read out a load cell even with a single-ended ADC, if one uses a proper analogue circuitry to move the signal to some reference point in the middle of the ADC input range.

Splitting up the interface would in my opinion make everything unnecessarily complicated. Consider this: There are microcontrollers which have on-chip ADCs which can be configured differential as well as single-ended. If we one day might have a driver which can use such ADC in both configurations (this is typically a single bit in some register), it could be used either as a temperature read out or as a load cell interface, or maybe for some entirely different purpose. If we would decide now to introduce a new interface incompatible with the current ADC interface in Klipper, we would have to write two drivers for the same ADC.

On the other hand I see no benefit from breaking the interface. The biggest problem for me right now is to understand how the current ADC interface is actually working ;-) I think I have understood that now mostly. This is a question of documentation mainly - maybe we should add some internal interface description to the developer documentation of Klipper. 


> For simple things (like standard deviation) it's probably better to just implement it directly. If it's a complex math algorithm (eg, FFT) then I'd use numpy.

Perfect, thanks!


PS: There is one last "fishy thing" left I want to attack before re-opening this PR. In the past I just discarded one ADC value after waiting for all moves to finish. This is actually not so nice. I am right now looking into ways to check the timestamp of the ADC value and compare it against the timestamp of the last move. This is a big more complicated than it sounds, since the ADC timestamp is probably the time of the digital readout, while the analogue sample might be taken earlier, but this is something which I can even measure. In the end, this might even allow to avoid waiting for the moves to finish, so everything can be sped up.

mhier:
I would like to revive this pull request now :-) I think the module is now a lot more mature than before. The number of parameters is greatly reduced, and most of them are determined in a semi-automatic procedure, which is described step-by-step in the documentation. I have tested this now with two different ADC chips (ADS1100 on the original mainboard of my printer, and a HX711 with some preliminary driver module). It should run with any ADC supported by the Klipper ADC interface.

KevinOConnor:
Sounds good.  Github wont let me reopen this PR (likely something to do with the upstream branch).  Feel free to open a new PR or rebase this one.

-Kevin

mhier:
It is rebased already - hence I will open a new one.

PR #4740, opened 2021-09-30
Anycubic Vyper

Crykor90:
This adds support and example config for the Anycubic Vyper.

Continuation of: #4733

Thanks to @mneuhaus for this work 

I have tested the config and everything works fine!


Crykor90 (commented):
Signed-off-by: Manuel Fichtner <manuel.fichtner@me.com> in agreement with @mneuhaus

mneuhaus:
Has anything in here changed compared to the pull request #4733 i opened with the gear_ratio information from you?

Crykor90:
> Has anything in here changed compared to the pull request #4733 i opened with the gear_ratio information from you?

Yes i added z1 endstop + commits + formal edits (like you agreed to me via discord ;) )

mneuhaus:
And then you said you couldn't which is why i created a new pull-request.
Did you test the homing with both endstops? i ran into issues with Z1 endstop enabled.
But ok, then you can take over here and i'll close my other pull-request since there is no reason for 2 🤷‍♂️

sstickel:
I am running a cfg based on this with both end stops and it works as expected. Tilted the X Carriage and Homing fixed it.

So can confirm it works.

sstickel:
did one of you guys try to bring the LED to life? I wasn't successful. According to the Marlin configs it should be on PA13, but I had no luck here
```
[output_pin bedled]
pin: PA13
value: 0


[gcode_macro LEDOFF]
gcode:
    SET_PIN PIN=bedled VALUE=0

[gcode_macro LEDON]
gcode:
    SET_PIN PIN=bedled VALUE=1
```

Crykor90:
> And then you said you couldn't which is why i created a new pull-request. Did you test the homing with both endstops? i ran into issues with Z1 endstop enabled. But ok, then you can take over here and i'll close my other pull-request since there is no reason for 2 🤷‍♂️

Endstops works great - and i opened this pull request because you said its ok and u have no time actually - so i fixed the formal mistakes.

Crykor90:
> did one of you guys try to bring the LED to life? I wasn't successful. According to the Marlin configs it should be on PA13, but I had no luck here
> 
> ```
> [output_pin bedled]
> pin: PA13
> value: 0
> 
> 
> [gcode_macro LEDOFF]
> gcode:
>     SET_PIN PIN=bedled VALUE=0
> 
> [gcode_macro LEDON]
> gcode:
>     SET_PIN PIN=bedled VALUE=1
> ```

Sorry i think this wont work because the led is connected to the display and i have no idea to handle this with klipper

sstickel:
Interesting. Because I was able to switch that LEDs on Marlin via the proper M command and the pin is configured in the board file there.

No, if it's really connected to the Display this will never work. Where you got that information from? Haven't seen schematics of that trigorilla board yet

Crykor90:
> Interesting. Because I was able to switch that LEDs on Marlin via the proper M command and the pin is configured in the board file there.
> 
> No, if it's really connected to the Display this will never work. Where you got that information from? Haven't seen schematics of that trigorilla board yet

PA13 is the Caselight Pin this isnt what you are searching for

The LED Pin is defined under `#define LED_PIN  -1`

sstickel:
And using [M355](https://marlinfw.org/docs/gcode/M355.html) I was able to control that LED when running Marlin FW so to me it looks like it's connected to the case light pin.

Where you got that information from, that the LED is controlled via Display?

austinsc:
I can confirm that the LED is not responding to SET_PIN. One thing you may want to add to your config:
`[output_pin beeper_pin]
pin: PB15
pwm: True
value: 0
shutdown_value: 0
cycle_time: 0.001

[gcode_macro M300]
gcode:
    # Use a default 1kHz tone if S is omitted.
    {% set S = params.S|default(1000)|int %}
    # Use a 10ms duration is P is omitted.
    {% set P = params.P|default(100)|int %}
    SET_PIN PIN=beeper_pin VALUE=0.85 CYCLE_TIME={ 1.0/S if S > 0 else 1 }
    G4 P{P}
    SET_PIN PIN=beeper_pin VALUE=0
`

Also, this was my case light attempt:
`[output_pin case_light_pin]
pin: PA13
pwm: False
value: 0

[gcode_macro M355]
gcode:
    {% set S = params.S|default(0)|int %}
    SET_PIN PIN=case_light_pin VALUE={S}`

FrYakaTKoP:
> Where you got that information from, that the LED is controlled via Display?

I was helping setup Crykor90's printer with klipper and tried to reverse engineer the LED Pin from Marlin. Since he told me he used a plain [M42](https://marlinfw.org/docs/gcode/M042.html) with only defining the S parameter (0-255) and the Marlin docu states that M42 without P parameter will fallback to LED_PIN which is defined as "-1" (or disabled). So i assumed that the Display also reads the gcode (how most those displays work) and turn on the Led from the M42. 

It's interesting that you are able to control it with M355 (Case Light Control), since we tried the PA13 Pin too (defined in Marlin as Case light), without luck (same as austinsc). But it is also possible that the Display catches both M42 and M355. 

Without closer look at the cabling we will not know for sure, but since there is no other pin defined in Marlin (other then PA13 which does not work). I still assume it is connected to the Display. Somebody owning a vyper and a multimeter needs to check the cabling, to be sure though.

sstickel:
Thank you for clarification. Yes, my start and end gcode for the stock Marlin FW just used M355 S1/0 and that worked as expected. I never tried with M42. As the printer is otherwise running flawlessly and right now printing almost 24/7 with Klipper I am have no plans to futher investigate for now.

Regarding this PR: I think it's a good base config for this printer and could be merged into master. Thanks everybody.

KevinOConnor:
Thanks.

-Kevin

PR #4748, opened 2021-10-01
rp2040: implement fix for RP2040-E5 errata

dalegaard:
The rp2040 USB controller needs a forced J state after reset in order to go in to connected mode. This patch implements that.

The rp2040 datasheet better explains the issue, but in a nutshell the USB controller requires 800us of J-state(D+ high, D- low) to transition from RESET to CONNECTED. On a busy USB bus, e.g. on a single-TT hub with other 12Mbps USB devices, this might never happen. Using some internal debug logic we can force this to happen(as seen from the USB controller).

Both I and @Coffee0297 have had issues with this on setups with multiple Picos connected to the same Raspberry Pi or hub. In our testing this patch solves the issue.

The pico SDK implements this same fix in `rp2040_usb_device_enumeration.c`. The method used in this patch is a re-implementation of that fix to be compatible with the Klipper codebase.

### DIFF
--- a/src/rp2040/usbserial.c
+++ b/src/rp2040/usbserial.c
@@ -10,8 +10,11 @@
 #include "board/usb_cdc.h" // usb_notify_ep0
 #include "board/usb_cdc_ep.h" // USB_CDC_EP_BULK_IN
 #include "board/usbstd.h" // USB_ENDPOINT_XFER_INT
+#include "board/misc.h" // timer_read_time
 #include "hardware/structs/resets.h" // RESETS_RESET_USBCTRL_BITS
 #include "hardware/structs/usb.h" // usb_hw
+#include "hardware/structs/iobank0.h" // iobank0_hw
+#include "hardware/structs/padsbank0.h" // padsbank0_hw
 #include "internal.h" // enable_pclock
 #include "sched.h" // DECL_INIT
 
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I agree it would be good to add this workaround.  As high-level feedback, though, I'm concerned about the complexity of this implementation.  I'm concerned that the various timers could cause more issues than the "initial pins" stuff.  In particular, it's not well defined if a timer can occur while still in the "init" phase, and it's not well defined what happens if a hardware irq calls `sched_add_timer()`.  The additional code complexity is also a bit daunting.

It seems to me that it would be simpler if we didn't have the timers and instead call `usbserial_init()` directly from `armcm_main()`, and then do all work directly in `usbserial_init()`.

-Kevin

dalegaard:
Hi Kevin,

Okay, I fear the workaround will actually end up more complex, not simpler, but I'll take a stab at it. I'll need to force `armcm_init` to initialize the timer first, then the USB, and busy wait until the USB is fully connected.

I think that this will also break the ability to reconnect to the USB on a busy bus?(which it will be if Klippy is talking to other devices). I don't know if this is an issue?

Best regards,
Lasse

dalegaard:
Hi Kevin,

After taking a step back and considering my use case, I've determined that for my use case I need initial pins quite quickly after the controller starts up, and can't live with an arbitrarily long wait (until the USB host comes up). The MCU is powered separately and so might come up long before the Raspberry Pi.

As the current works for me, I've therefore decided I won't be looking into converting this to the busy-wait solution we discussed, so I'll close this.

You or someone else are of course free to base a future fix for this errata issue on the code for this PR if you decide it's helpful.

Best regards,
Lasse

dalegaard:
Hi Kevin,

So I had some time to kill and decided to play around a bit with this anyway - I can't really wrap my head around how the rp2040 is reset after a USB communication failure? In my setup the Picos and Raspberry Pi are powered on at the same time from the same PSU, so the Picos will come up before the Pi is finished booting and can start talking to them over USB. I have found a work around for my hardware setup such that I can get away with not using `initial_pins` anymore, so now I'm just trying to understand what happens during a reset.

Basically:
Imagine that the Raspberry Pi reboots but the Picos do not. How do I ensure that the USB errata fix is run in that case, if implemented as a `usbserial_init` busy wait loop? Can you point me to the code that makes the MCU reset in this case, so I can understand what is going on?

Best regards,
Lasse

PR #4751, opened 2021-10-02
Richardjm/wled module

richardjm:
Add support for WLED to klipper
- Extra commands to drive WLED presets
- SET_WLED for easy conversion from SET_LED of neopixels
- Uses background thread for processing
- Uses urllib2 for POST requests


richardjm:
I think there is a case to be made for using the Requests library
- It's far simpler to code
- It is measurably faster - about twice as fast
- I did try using basic connection pooling use httplib but either I was doing it wrong or it made no difference

PR #4764, opened 2021-10-03
Module: Toolhead, implements independent X-Y axis acceleration

jasonhamilton:
Adds the ability to allow for independent x-y axis acceleration.  This allows for bedslingers or printers with different resonance characteristics on the X and Y axis to further improve print quality/speed.

Signed-off-by: Jason Hamilton <jasonscotthamilton@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -93,6 +93,12 @@ max_velocity:
 max_accel:
 #   Maximum acceleration (in mm/s^2) of the toolhead (relative to the
 #   print). This parameter must be specified.
+max_accel_x:
+#   Maximum acceleration (in mm/s^2) of the toolhead in the direction
+#   of the X axis (relative to the print).  The default is max_accel.
+max_accel_y:
+#   Maximum acceleration (in mm/s^2) of the toolhead in the direction
+#   of the Y axis (relative to the print).  The default is max_accel.
 #max_accel_to_decel:
 #   A pseudo acceleration (in mm/s^2) controlling how fast the
 #   toolhead may go from acceleration to deceleration. It is used to
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -17,6 +17,8 @@ class Move:
         self.start_pos = tuple(start_pos)
         self.end_pos = tuple(end_pos)
         self.accel = toolhead.max_accel
+        if toolhead.max_accel_x != toolhead.max_accel_y:
+            self.accel = self.calc_split_x_y_accel()
         self.timing_callbacks = []
         velocity = min(speed, toolhead.max_velocity)
         self.is_kinematic_move = True
@@ -104,6 +106,20 @@ class Move:
         self.accel_t = accel_d / ((start_v + cruise_v) * 0.5)
         self.cruise_t = cruise_d / cruise_v
         self.decel_t = decel_d / ((end_v + cruise_v) * 0.5)
+    def calc_split_x_y_accel(self):
+        # get the max acceleration based on the ratio of the direction
+        # between x,y axis
+        if self.start_pos[1] == self.end_pos[1]:
+            ratio_y_x = 0
+        else:
+            angle_rad = abs(math.atan2(self.end_pos[1]-self.start_pos[1],
+                                       self.end_pos[0]-self.start_pos[0]))
+            if angle_rad > math.pi/2:
+                ratio_y_x = (math.pi-angle_rad)/(math.pi/2)
+            else:
+                ratio_y_x = angle_rad/(math.pi/2)
+        return  (self.toolhead.max_accel_y * ratio_y_x
+                 + self.toolhead.max_accel_x * (1-ratio_y_x) )
 
 LOOKAHEAD_FLUSH_TIME = 0.250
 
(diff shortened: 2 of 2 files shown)


KevinOConnor:
See the discussion at #4661 .

-Kevin

jasonhamilton:
Oh wow.  I should have looked there first!

jasonhamilton:
I’m happy to run some tests on my printers for some data points.  Is there a specific model, slicer settings, etc that would make comparing tests easier?

JRHeaton:
I think I speak for a lot of people with bedslingers running Klipper who are, whether it's an ideal build or not, limited in our max acceleration to what our insanely heavy Y axis can handle. This and previous attempts to address per-axis treatment of accelerations I believe to have been and be being blown off while requesting a nebulous test/documentation goal must be reached, and somewhat ignoring the inherent good this would do to machines with such weight disparities across X/Y axes. What's more, the other two primary recognized production 3d printer firmwares in the industry support independent control over acceleration on this granular of a level. Lastly, why is(was) it seen as inherently beneficial to separate out Z accel/velocity limits from X/Y in Klipper? I submit that the same principle applies, your bed slinger axes may be weighted Z > Y > X, and frankly the Y will be more important to your print moves, so it feels odd that we cannot control each.

People in the SpeedBoatRace discord channels and on YouTube have been making private/personal forks in order to achieve the speeds you're seeing on video. There _is_ demonstrated merit to this on those mediums. What I think we need are clearly defined documentation goals and a testing strategy for demonstrating whatever data is required to show this merit and that it is going to make this provably "non-problematic" to existing features or UX. Just saying "Good questions, and I don't know the answer to them. Figuring out all that stuff is part of the work that I think someone would need to do." (@KevinOConnor) is correct in 99% of OSS circumstances but who but you is opposing this change? And if so, it likely need be you that define the barrier-to-entry.

Here is a fork currently being used(it looks like by a participant on the linked-to discussion of old) here: https://github.com/Piezoid/klipper/blob/work-peraxis/klippy/kinematics/limited_cartesian.py and discussion around this and others is that is is a necessary evil at the moment to use a potentially less trustworthy fork like this to achieve the performance we want out of the modern "speed printing era." As a developer myself I see more potential harm and confusing support on Klipper's end if an unexperienced person resorts to using one and then having that code potentially NOT meet your rigorous standards (which I respect of this project) yield them unexpected behavior for which they may come submit an issue _here_ not knowing whether it's a core Klipper issue or a fork issue, etc. I have seen it before; I think SuperSlicer is a good example of an OSS project where it started as something we all loved (PrusaSlicer), then the community needs outpaced/misaligned with Prusa and so it was effectively superseded by community authors. I would really not like to see any more dividing of our limited efforts in this space.

jasonhamilton:
I can appreciate the need for simplicity and ease of troubleshooting.  I think this philosophy makes Klipper great.  Based upon my testing I did see some speed improvements without sacrificing print quality.  For example a model I printed at 1500 acceleration on x/y took 1:42:10 (1hr, 42min) vs 2700/1500 which took 1:27:25, which is a somewhat significant ~15% improvement in print time.

I do think there are more elegantly-coded solutions than mine such as the module in #4661 or the one referenced by @JRHeaton .  

Perhaps a solution to the issues raised could be after some basic testing of code (whatever implementation is deemed the best) it can then be merged into mainline with the documentation clearly indicating the feature as experimental?  I think implementing a new module/printer which extends the Cartesian printer type would keep any issues isolated to those knowingly using an experimental module.  This would also allow for broader testing and data collection without impacting other users.

Honestly it’s not something that’s hugely important to me and my hacked up fork serves me well, however it does seem like something a decent amount of other people would like to see implemented.  I’m happy to help in any way, however closing this merge request since there are better implementations for merge candidates.

JRHeaton:
As was stated, this is definitely something that many cartestian users want implemented. I guess I am asking what specifically needs proven and in what way, @KevinOConnor, as to satisfy your requirements for it to be "valid?" Otherwise we are left hanging for the nth time and with disparate forks implementing varieties of this features with different levels of success. The community is willing to do the work, clearly, but we need you, the maintainer, to lay out the requirements.

dewi-ny-je:
@JRHeaton using Marlin pick 3 models (for example one simple box, one large cylinder, and a figurine), then print them at acceleration equal in all axes (use the lowest acceleration of the two axes), and provide print time and photos of the result.

Then keep one of the accelerations constant (Y I guess), and increase the other one (X probably). Provide print time and photos of the result.

Then we can see if there is an actual speed advantage. Be aware that even if there is one, someone has to write the code to implement it: the current developers work on whatever they find interesting (except for fixing bugs, obviously), not on call. If it's not too complex to code probably someone will be willing to implement it for you for a couple hundred Euro/dollars (which is less than what it would be the regular pay, but likely enough to get the interest of someone doing it for fun).

JRHeaton:
@dewi-ny-je @KevinOConnor You're missing the entire point of why people are continuously making this same feature PR over and over, and asking klipper to make it mainline. The physical mechanics of the machine are limiting. You have a bed that is up to say 10 times as much mass as your toolhead, and we're pretending there's no advantage to letting users **optionally** configure that Y axis to a lower accel while retaining the higher X accel? That is silly. We lose countless hours to having to limit our X accel to what our Y(bed) can handle. If we try to crank it, we get layer skips. 

This is not a PR chasing quality, **this is a feature/PR focused on optimizing utilization of the printer hardware on all axes**. What is the point of optimizations like input shaper and PA if we're hard limited by motion? Not everyone has a delta or CoreXY. Unfortunately, I see this dismissed recurrently and you guys ask the same questions, which are targetless. Again, this is an optional thing, nobody is forcing anyone to use differing accels, but the people who know how to use it will benefit greatly from it, especially in the speed printing arena.

PR #4768, opened 2021-10-06
External ADC via I2C Module

dawidmpunkt:
module: klippy/extras/i2c_adc.py
documentation: docs/i2c_adc.md

A module for external ADCs (ADS1013 to ADS1015 and MCP3421 to MCP3428) via I2C (and a rPi).
Voltage can be read via a virtual chip/pin.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3959,6 +3959,47 @@ serial:
 #   Auto cancel print when ping varation is above this threshold
 ```
 
+### [i2c_adc]
+
+Module to connect MCP34xx and ADS 1XXX analog-to-digital converters
+via I2C bus (one may define any number of sections with an "i2c_adc"
+prefix, for example [i2c_adc my_adc]).
+See the [i2c_adc guide](i2c_adc.md) for more information.
+
+The adc pin can be accessed by other modules via suffix.
+e.g. 'adc_pin: my_adc:'
+
+```
+[i2c_adc my_adc]
+i2c_address: 104
+#   The i2c address that the chip is using on the i2c bus.
+#   This parameter must be provided.
+#   Standard Adresses are for example:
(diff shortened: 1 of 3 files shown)


dawidmpunkt:
@KevinOConnor Thank you for your Feedback!

1. The primary function would be to connect sensors like load-cells, ambient sensors or to read the voltage off a filament width sensor module (the primary application I wrote this module for). Other options would be possible, but I have not thought of those yet.
2. I'll add it too the index files. I'll take a look at the header designations. I will put the info into i2c_adc.py.
3. Ok, I'll look into that.
4. The higher the resolution, the lower the sample rate. Some ADC modules are very slow at maximum resolution. The other way around, the faster the sample rate the less precise is the sensor. That would not be useful for e.g. load cells or a filament width sensor. I imagine, that different applications require different setting, so those parameters have default values and do not need to be set by the use, but they can be changed if necessary.
5. I don't quite understand the question. The aim of the module is to read voltage information from an external source (e.g. 1.75 volts) and to pass the value (1.75) it to other modules. E.g. the above mentioned filament width sensor (e.g. a hall based sensor) produces a voltage, that corresponds to the filament diameter (e.g. 1.75 volts). 
6. Right, I'll add the information.

KevinOConnor:
Thanks.  What is the state of this PR?

-Kevin

hesiod:
@KevinOConnor I'm considering picking up this PR since I'm designing a filament width sensor PCB with an I²C ADC at the moment. I'll only be able to work on this once the boards arrive, but before I'd like to know that the approach is feasible in principle.
Aside from the already mentioned issues (regarding `_sample_timer()` and the documentation), are there any major issues or reasons against supporting I²C ADCs from your side? Or improvements you'd like to see?

dawidmpunkt:
Please feel free to use this code. I don't have the time anymore to work on this PR.

PR #4769, opened 2021-10-06
Refactored tsl1401cl and hall_filament into filament_width_sensor

dawidmpunkt:
module: klippy/extras/filament_width_sensor.py
Documentation: docs/Filament_Width_Sensor.md

Refactored tsl1401cl and hall_filament_sensor into filament_width_sensor.
Included extruder.find_past_position() method.


KevinOConnor:
Thanks.  As high-level feedback, I agree that it would be good to refactor these two modules into a new module (as they share a lot of overlap).  However, it looks like this new module has many configuration quirks of the existing module (eg, `measurement_interval`, `measurement_delay`, `logging` options) and adds several new configuration options (eg, `extrude_factor_enable`).  As mentioned in the Contributing document ( https://www.klipper3d.org/CONTRIBUTING.html ) we really want the config options to be focused on providing a "high impact" benefit to users, and I fear all these config options will make the module difficult to utilize.

-Kevin

PR #4782, opened 2021-10-07
ds18b20: Fix for Python 3 compatibility

matthewlloyd:

### DIFF
--- a/klippy/extras/ds18b20.py
+++ b/klippy/extras/ds18b20.py
@@ -5,6 +5,7 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import logging
 import mcu
+import sys
 
 DS18_REPORT_TIME = 3.0
 # Temperature can be sampled at any time but conversion time is ~750ms, so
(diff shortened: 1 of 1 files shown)


matthewlloyd:
Assuming sensor_id is a str, it works in Python 2, but not Python 3:

```
pi@octopi: $ python2
Python 2.7.16 (default, Oct 10 2019, 22:02:15)
[GCC 8.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> class A(): pass
...
>>> self = A()
>>> self.sensor_id = 'asdf'
>>> "".join(["%02x" % (x,) for x in bytearray(self.sensor_id)])
'61736466'
>>>

pi@octopi: $ python3
Python 3.7.3 (default, Jan 22 2021, 20:04:44)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class A(): pass
...
>>> self = A()
>>> self.sensor_id = 'asdf'
>>> "".join(["%02x" % (x,) for x in bytearray(self.sensor_id)])
  File "<stdin>", line 1, in <module>
TypeError: string argument without an encoding
>>>
```

It works on both if you add a call to encode() though:

```
"".join(["%02x" % (x,) for x in bytearray(self.sensor_id.encode())])
```

KevinOConnor:
Thanks - commit 64cb8d98.

-Kevin

PR #4783, opened 2021-10-07
Add EXPAND_<macro> expand-only variant

sttts:
For a gcode macro `XYZ` this adds `EXPAND_XYZ` which just expands the Jinja2 template and prints it in the status.

TODO:

- [x] don't list this in the `HELP` command
- [ ] turn it into an independent command like `EXPAND XYZ`?

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -136,6 +136,7 @@ class GCodeMacro:
         else:
             self.gcode.register_command(self.alias, self.cmd,
                                         desc=self.cmd_desc)
+            self.gcode.register_command("EXPAND_" + self.alias, self.expand)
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Alas, I'm still missing what this change does.  I'm guessing it is some kind of macro debugging utility.  If so, my high-level feedback is that I'm not sure this is a good candidate for the Klipper master branch at this time.  In particular, I'm concerned that using g-code commands and the g-code console to debug g-code is not robust (in general, it's preferable to add new features via the klipper api server), that rendering the template can have side effects that would be confusing for users looking only to debug, and that reserving the EXPAND_ prefix on all G-Code commands would not be scalable.

Thanks.
-Kevin

PR #4787, opened 2021-10-07
spi_flash: fix compatibility with Python 3

Arksine:
This should fix spi flash to work both on Python 2 and Python 3.  I tested on both locally, however I was unable to test the workaround for bootloaders that don't rename the firmware file.  I believe it should work though.


Arksine:
No problem, I think your commit will take care of it.  :+1:

Issue #4054, opened 2021-03-15
Temperature sensor "NTC 100K beta 3950" is very inaccurate


ssotangkur:
The built in `sensor_type`  NTC 100K beta 3950 is too inaccurate to be used and shouldn't be provided unless fixed. The algorithm used, assumes the beta is a constant 3950 throughout the temp range, but in reality none of these types of thermistors spec their beta in the critical temp ranges needed for 3d printing. Instead, a linear interpolation of the resistance/temperature chart with several datapoints should be used. [Here](https://www.makeralot.com/download/Reprap-Hotend-Thermistor-NTC-3950-100K.pdf) is such a datasheet. 

Using a thermocouple equipped multi-meter I first confirmed that it matched the temp reading of the printer using its original `EPCOS 100K B57560G104F` and confirmed they were within 1C of each other over multiple temps throughout their range. To do so I placed the thermocouple into a slot next to the heater cartridge what was just wide enough to fully embed the probe inside. I then replaced the thermistor with an HT-NTC100K Beta 3950 sensor from Amazon. I repeated the temp check but this time noticed a very wide discrepancy (15-20C over-reporting). I tried various calibration methods but found that using the table provided to create a custom `[adc_temperature]` with the same resistance, temperature values, produced a highly accurate sensor. This brought the calculation back to within 1C of the thermocouple readings.

## The Ask
- remove line:  https://github.com/KevinOConnor/klipper/blob/3233ec08f4a7165d54413e7a24b0021053604053/klippy/extras/thermistor.py#L118
- Add it back as a "resistance sensor" someplace like here: https://github.com/KevinOConnor/klipper/blob/3233ec08f4a7165d54413e7a24b0021053604053/klippy/extras/adc_temperature.py#L281

KevinOConnor:
I'm not sure what you're reporting.  It is true that using beta isn't as accurate as steinhart-hart or a custom temperature table, but some sensors are defined as "beta 3950" - that is, the manufacturer produced the sensor to fit that curve as best as they can and the only temperature information they provide is that curve.  The "makeralot" link is just one thermistor - it isn't all "beta 3950" thermistors.

So, although "beta 3950" may not be very "accurate" it is more accurate than anything else we have.

-Kevin

ssotangkur:
What if literally every NTC 100k beta 3950 you can buy corresponds to the same datasheet? 

Here are the ones I could find:

[...]

PR #4789, opened 2021-10-08
docs: Note "NTC 100k beta 3950" as not preferred

crashmaxx:
docs: Note "NTC 100k beta 3950" as not preferred

As pointed out in issue #4054, the "NTC 100k beta 3950" sensor
type is not as accurate as "EPCOS 100K B57560G104F" for most
NTC 100k B3950 thermistors. The generic name and inclusion
next to other sensor types with more specific names, implies
that it would be a good default option.

To guide users away from this option, this PR removes it from
the [extruder] section, and lists it at the end of the options in
the "Common thermistors" section with "not preferred" in
parenthesis next to it.

Signed-off-by: Andrew Stowell <crashmaxx@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -696,9 +696,9 @@ heater_pin:
 #   periods) to the heater. The default is 1.0.
 sensor_type:
 #   Type of sensor - common thermistors are "EPCOS 100K B57560G104F",
-#   "ATC Semitec 104GT-2", "NTC 100K beta 3950", "Honeywell 100K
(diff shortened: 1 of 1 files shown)


grigi:
One minor comment, the docs doesn't make it clear, that if your thermistor says its a "beta 3950k" (like many cheap ones do) that you should probably use the EPCOS spec.

KevinOConnor:
I think this should be addressed with the merge of PR #4859 .

-Kevin

PR #4792, opened 2021-10-09
Fix diag pins for BTT Octopus

arne182:
PB10 is definitely the Heater pin.
with reference from https://github.com/bigtreetech/BIGTREETECH-OCTOPUS-V1.0/blob/master/Hardware/BIGTREETECH%20Octopus.pdf the Diag 0 1 2 pins have been added

Signed-off-by: Arne Schwarck <arneschwarck@gmail.com>

### DIFF
--- a/config/generic-bigtreetech-octopus.cfg
+++ b/config/generic-bigtreetech-octopus.cfg
@@ -208,7 +208,7 @@ max_z_accel: 100
 #[tmc2130 stepper_x]
 #cs_pin: PC4
 #spi_bus: spi1
-##diag1_pin: PB10
+##diag1_pin: PG6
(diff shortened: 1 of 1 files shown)


arne182:
> Thanks. @charlespick , @ShohninDmitriy do you have any comments on this?
> 
> Separately, commits need to have a signed-off-by line containing your full real name for me to commit directly - see: https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md
> 
> However, I can alter the config file using my signed-off-by line if you prefer.
> 
> -Kevin

Done
-Arne

KevinOConnor:
Thanks.

-Kevin

PR #4801, opened 2021-10-13
Fix typo

arne182:
As suggested in https://github.com/Klipper3d/klipper/pull/4792
Signed-off-by: Arne Schwarck arneschwarck@gmail.com


KevinOConnor:
Thanks.

-Kevin

PR #4810, opened 2021-10-16
controller_fan: Add min_speed feature

xsekax:
This feature allows to cooldown with a single fan not only the controller board, but other components, like RPi or\and a fanless PSU in all-in-one cases (voron legacy for example) when printer is idling. It helps reduce noise and still be able to ramp up when controller is turning on motors.


ReXT3D:
Thank you for this.  An extremely useful feature for case exhaust fans that should have the ability to provide continuous low flow-rate ventilation.

PR #4814, opened 2021-10-18
Add support for the BLTouch sensor on the Wanhao D9

mdshw5:
- Added signed-off commit message

This PR supersedes #4701 and is ready to merge.

### DIFF
--- a/config/printer-wanhao-duplicator-9-2018.cfg
+++ b/config/printer-wanhao-duplicator-9-2018.cfg
@@ -68,6 +68,23 @@ max_temp: 120
 [fan]
 pin: PE3
 
+# Uncomment for use on D9 MK2/3
+# [bltouch]
+# sensor_pin: PH3
+# control_pin: PH4
+# x_offset: 0
+# y_offset: -50
+# z_offset: 2.1
+
+# Uncomment for use on D9 MK2/3
+# [safe_z_home]
+# home_xy_position: 150,150 # Change coordinates to the center of your print bed
+# speed: 50
+# z_hop: 10
+# z_hop_speed: 5
+ 
+# Inductive probe config for use on D9 MK1
+
 [probe]
 pin: !PH3
 x_offset: 27
--- a/config/printer-wanhao-duplicator-9-2018.cfg
+++ b/config/printer-wanhao-duplicator-9-2018.cfg
@@ -69,19 +69,21 @@ max_temp: 120
 pin: PE3
 
 # Uncomment for use on D9 MK2/3
-# [bltouch]
-# sensor_pin: PH3
-# control_pin: PH4
-# x_offset: 25
-# y_offset: 0
-# z_offset: 0
+
+#[bltouch]
+#sensor_pin: PH3
+#control_pin: PH4
+#x_offset: 25
+#y_offset: 0
+#z_offset: 0
 
 # Uncomment for use on D9 MK2/3
-# [safe_z_home]
-# home_xy_position: 150,150 # Change coordinates to the center of your print bed
-# speed: 50
-# z_hop: 10
-# z_hop_speed: 5
+
+#[safe_z_home]
+#home_xy_position: 150,150 # Change coordinates to the center of your print bed
+#speed: 50
+#z_hop: 10
+#z_hop_speed: 5
 
 # Inductive probe config for use on D9 MK1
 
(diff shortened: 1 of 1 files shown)


MrPuffyNZ:
Hi all, i have just used the new config on my Wanhao D9 300 Mk2 and I am getting an error
Unknown pin chip name 'PE3
Recv: // [bltouch]
Recv: // sensor_pin'

can someone please help me identify what may be the issue.
my config is exactly the same as whats posted

mdshw5:
@MrPuffyNZ did you select `Atmega AVR` when you ran `make menuconfig`? I also just noticed that I didn't add the atmega2560 pin aliases to this config, which may likely fix your issue.

mdshw5:
@MrPuffyNZ I've added pin maps to the config. You can give it another go - I'd like to see if this works on original hardware since my machine was originally a MK1.

MrPuffyNZ:
Hi mdshw5,
thanks for the response, the fault seems to lie in the bltouch and safe home config as soon as thats uncommented i get the error as soon as I recomment it I get klipper status is ready, when I uncommented the bltouch and safe home config I commented out the probe config for the inductive probe i hope this was right and that is when i can recreate the error. if i reverse it and comment out the bltouch and safe home and uncomment the probe section I get klipper status: ready.

mdshw5:
@MrPuffyNZ my printer’s current config is here: https://gist.github.com/mdshw5/cefb2633c0575c47712aab1ef4a33cfd. Note that I’m running a modified effector and bed setup so many of my limits and positions will be different than your machine.

MrPuffyNZ:
Hi Matt,
thanks for that, my printer is heavily modified as well, running on linear rails for x and y with a static bed ie no adjustment via springs, so I completely relay on a probe. I am using a bondtech extruder so require my e value to be 415, also have you managed to make the original screen work with klipper. My printer is a 300 x 300 bed and i notice in your cfg file that you have some macro's at the end are they ok to use. sorry for so many questions and I appreciate your help.

mdshw5:
The macros should be fine. I think I saw your printer on the Facebook group - you’re the one that machined aluminum parts for the axis sleds. You should be able to get a working config going from this template, but maybe I’m missing something important. I’ve not been able to get the screen working as it runs a strange interface and firmware (DWIN DGUS). However there is a project that made some progress on a related screen: https://github.com/Desuuuu/DGUS-reloaded-Klipper.

MrPuffyNZ:
Hi Matt,
i now have movement etc, only problem now is the bltouch wont trigger when it touches the build plate - I get a "no trigger on z after full movement" any advice?

regards

Aaron

mdshw5:
Have you run through the initial tests to verify that your probe responds? https://www.klipper3d.org/BLTouch.html#initial-tests

MrPuffyNZ:
Hi Matt,
I have gone through the initial BLtouch Checks and the pin deploys on command and raises on command, when deployed it shows as open which is correct, but when you trigger it and run a query it still shows as open which is where the problem exists, where to from here, and once again thank you so much for your help.

regards

Aaron

mdshw5:
Very strange. I’ll post a photo of my wiring from the splitter board later tonight. Was your sensor working properly on the marlin firmware?

MrPuffyNZ:
HI Matt,
yes sensor was working fine on marlin firmware, this unit is a mk1 that had the mark2 upgrade done on it if that helps.

regards

Aaron

mdshw5:
FCA4667F-38AD-42F6-83AF-F4B1ED635B7B
F87A6858-5714-46E5-A9E9-AC2D4EA8F20A

Apologies for the mess but here is how my sensor is wired to the splitter board. If you’re not getting a closed circuit then I’d imagine you either have a flaky sensor or bad wiring.

mdshw5:
Strange. I have a MK1 that I did pretty much the same thing to. My splitter board looks like this (maybe you have a different revision with a different sense pin assignment?): 
image

MrPuffyNZ:
Hi Matt,
Ok so i have worked it out, the BLTouch I had was a revision 3.1 and it would not trigger, but I found an older genuine bltouch 2.0 and it works fine and reports everything back and stops on the z axis when probing the bed. The 3.1 version works fine under marlin which is interesting, so all good, now where to from here.

regards

Aaron

Gilabite:
My 3.1 works fine

KevinOConnor:
Thanks.  I'm not sure what the state of this PR is.  I'm fine with committing the commented out blocks for the bltouch, but this commit has whitespace errors (there should be blank line between probe comment and probe section, there should not be an extra space between `#` and commented out block), and there is an erroneous `board_pins` section added to the end of the config file.

-Kevin

mdshw5:
Thanks @KevinOConnor. I've made the necessary changes and this is ready to merge. I've fixed the whitespace issues, and admit that I didn't fully understand where the board pin aliases should be defined.

mdshw5:
This pull request is ready to be merged.

cron410:
I can confirm this works both with the V2 breakout board and with a modified V1 breakout board on two otherwise stock printers using a BLtouch 2.0. Might want to add a comment near the [probe] section that it needs to be removed if the [bltouch] section is used. This will help out some new users and cut some future issues.

PR #4821, opened 2021-10-21
Reduce code size of pru1 on beaglebone boards

KevinOConnor:
The PRU mcu code size has grown to the point where there is almost no free space available.  Currently it is using 8184 out of 8192 available bytes (as reported by the build test in the regression test cases).  This is making it difficult to deploy mcu updates (even on other boards) as an increase in code size causes build failures on the PRU.

This code change moves message encoding from pru1 to pru0 to free up space on the main pru1 core.  This frees up 412 bytes on pru1 (7772 of 8192 as reported by the build test).

I don't have a good way to test this change (other than simple test cases), so further testing is appreciated.  Please report your findings here (success or failure).

-Kevin

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -108,7 +108,7 @@ error:
 }
 
 // Encode a message
-uint_fast8_t
+static uint_fast8_t
 command_encodef(uint8_t *buf, const struct command_encoder *ce, va_list args)
 {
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  Verifying normal behavior continues to work would be great.

-Kevin

PR #4823, opened 2021-10-21
Update generic-bigtreetech-skr-2.cfg

kanawati975:
Hello!
I've used the SKR-2 Example file and built a fully functional Voron Switchwire running Klipper. 
I'd love to share the modifications and notes about the issues that may encounter anyone with this build. I'm also willing to share the whole config file with the notes, if that would help.

### DIFF
--- a/config/generic-bigtreetech-skr-2.cfg
+++ b/config/generic-bigtreetech-skr-2.cfg
@@ -2,6 +2,9 @@
 # To use this config, the firmware should be compiled for the
 # STM32F407 with a "32KiB bootloader".
 
+# If the mcu's id is not discoverable 'ls /dev/serial/by-id/*' repeat the 'make menuconfig' 
+# and choose 'Internal Clock' from the low level options
+
 # The "make flash" command does not work on the SKR 2. Instead,
 # after running "make", copy the generated "out/klipper.bin" file to a
 # file named "firmware.bin" on an SD card and then restart the SKR 2
--- a/config/generic-bigtreetech-skr-2.cfg
+++ b/config/generic-bigtreetech-skr-2.cfg
@@ -1,9 +1,6 @@
 # This file contains common pin mappings for the BigTreeTech SKR 2.
 # To use this config, the firmware should be compiled for the
-# STM32F407 with a "32KiB bootloader".
-
-# If the mcu's id is not discoverable 'ls /dev/serial/by-id/*' repeat the 'make menuconfig' 
-# and choose 'Internal Clock' from the low level options
+# STM32F407 with a "32KiB bootloader" and choose 'Internal Clock' from the low level options.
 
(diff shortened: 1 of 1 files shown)


JamesH1978:
I have not found this on the SKR2. I have had 3 now, and have helped quite a few others flash, and have never had to choose internal clock on USB

kanawati975:
@JamesH1978 I have this issue with the SKR-2 and was never able to find the mcu id. I tried almost every option in menuconfig and the only option that worked after flashing is internal clock. Do you think that I might have a faulty Board?

JamesH1978:
image
I have always found the default to work, tbh. As long as the SD used is fat/fat32 and of a sane size for fat32, say 128MB - 16GB. If the flash is not good, you won't get a by-id. I cannot speak for your board, but at least one of mine was faulty, but one of the drivers, not the usb

kanawati975:
@JamesH1978 Would you please elaborate about the driver port issue? 
Does it happen to be the Y driver socket/port?

JamesH1978:
In my case, the Z stepper would only go in 1 direction, no matter what the dir_pin was set with a ! or not. So STEPPER_BUZZ would push it always in an up direction, and not 1 step forwards and 1 step back as it should

kanawati975:
@JamesH1978 In my case, Sensorless Homing on the Y axis is completely random. There is no way to set the sensetivity on a fixed number and you can't home Y two times with the same result. Either crashing into the end, or too strong sensitivity that Y barely moves. That was also the case with Marlin, before migration to Klipper... Of course I tried another TMC2130 Driver to no avail. Finally I swapped Y with E1 and now I'm getting relatively okay (ish) Homing on the Y axis.

JamesH1978:
funny you should say that, i am having similar issues with Y and sensorless homing, but i put it down to a mechanical issue, as I have flat spots when it travels. which is why i assumed it was mechanical... I am going to rebuild my Y later and I will report back

kanawati975:
@JamesH1978 No, it's far from being mechanical (at least at my end) because I replaced everything, and nothing helped except swapping the socket. Previously there was fan issues but I thought it was a faulty MosFET. 
I'd glady help in whatever you may need, but I'm seriously considering to replace the Board with something smaller and simpler.

KevinOConnor:
Thanks.  I'm not sure of the state of this PR.  It definitely should not be necessary to disable the crystal on these boards.  I don't know why your board is having an issue.

-Kevin

kanawati975:
@KevinOConnor Just a quick update on the SKR2's Issues: I have deactivated sensorless homing and everything worked out prefectly normal (except the damaged fan mosFET). All the crazy behaviour was caused by the sensorless homing but I'm not sure if that's from the Board itself. What I do know for sure that using the Diag pin for Endstop is not a very good idea.

PR #4824, opened 2021-10-21
Update printer-eryone-er20-2021.cfg

cellopoly:
Update the ER-20 printer.cfg to add the extra config options required for Klipper.


mental405:
@cellopoly, I think the consensus is that this is fine to merge. I would just change the comment to say that users "Users may need to enable the serial interface and use the Eryone flash tool to update the firmware" as previously suggested and you will also need to provide a sign-off before it can be merged.

mental405:
Additionally  it looks like the changes  are longer than 80 characters. All text should be no longer than 80 characters.

Are you positive that the eryone tool is the only way to get this thing to flash appropriately?

PR #4828, opened 2021-10-23
config: Added cfg for Prusa i3 MK2/S

drew6017:
Parameters based off of those used in the original Prusa-firmware, and were adjusted for
the best results on a Prusa i3 MK2S running stock hardware.

Signed-off-by: Andrew Howard <ahoward@divisionind.com>

### DIFF
--- a/config/printer-prusa-i3-mk2s-2017.cfg
+++ b/config/printer-prusa-i3-mk2s-2017.cfg
@@ -1,12 +1,9 @@
-# Adapted for Prusa i3 MK2/S by Andrew Howard <ahoward@divisionind.com>. Derived from generic-mini-rambo.cfg.
+# Adapted for Prusa i3 MK2/S from generic-mini-rambo.cfg.
 # Firmware should be compiled for the AVR atmega2560.
 
 # See docs/Config_Reference.md for a description of parameters.
 # Note: Adjust [probe]->z_offset, run mesh bed leveling, and do
 #       skew correction as a MINIMUM before printing for the best experience.
-
-# remove if not running mainsail
-[include mainsail.cfg]
 
 [stepper_x]
 step_pin: PC0
@@ -79,7 +76,7 @@ step_pin: PC3
 dir_pin: !PL6
 enable_pin: !PA4
 microsteps: 16
-rotation_distance: 18.5572135
+rotation_distance: 18.56
 nozzle_diameter: 0.4
 filament_diameter: 1.75
 heater_pin: PE5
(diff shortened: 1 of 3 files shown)


drew6017:
Sorry for all the commits. Let me know if you want me to re-do the pull. I've been tweaking things as I go because I'm using this as my daily-driver printer. I have noticed that you get more consistent first layers doing the mesh bed level at the start of each print with 1 sample instead of how I have it now. 

People would have to add a line to the start gcode in their slicers to trigger it though because normal prusa preambles start mesh leveling unconventionally.

Sineos:
Thanks for the contribution.
Please check against https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md
On first glance following spec items:
* 2
* 3
* 5.II, 5.III
* 6.V, 6.VIII
* 8

drew6017:
I'll fix those as soon as I get a chance. 

As per the `square_corner_velocity` (SCV), I set it to 1 because corners were noticeably rounder on my test prints at anything over 2. Plus, the only down side to a lower SCV is increase print-time right? However, this change was not noticeable going from 5->1 in my prints, but the quality change was.

Sineos:
As per my understanding a SCV of 5 already is quite conservative. SCV of 1 means that the print-head nearly comes to a standstill on cornering moves.
Never tried such low values but I could imagine that it leads to unwanted extrusion issues.

drew6017:
Hey man I'm just using what works. In practice, the print head does not freeze at corners so I doubt it's a problem. I think SCV just means that it needs to "take the corner" at that velocity -- which would mean it would just deccel to 1mm/s at the corner then accel back to print speed. Idk tho, haven't looked at the code. I dont even know what classifies as a "corner" here, just 90deg? 89.9?

drew6017:
I think I got everything, but as for 6.v the native firmware uses steps per unit and the extruder is not geared. 

https://github.com/prusa3d/Prusa-Firmware/blob/6188870c2e86d281bae129067d1d7d6f32f2f1e2/Firmware/variants/1_75mm_MK2-RAMBo13a-E3Dv6full.h
```c
// Steps per unit {X,Y,Z,E}
#define DEFAULT_AXIS_STEPS_PER_UNIT   {100,100,3200/8,161.3}
```

`3200 (steps per revolution @ 16x micro) / 161.3 = 19.8 (rotation_distance)`

However, I used 18.5 because for some reason using 19.8 did not yield accurate extrusions.

KevinOConnor:
Thanks.  I'm not sure what the state of this PR is.  It doesn't look like it has completed a review.  (On a quick look, there is still a `max_extrude_cross_section` and an unusual `square_corner_velocity`).

-Kevin

PR #4829, opened 2021-10-23
Update generic MKS Robin Nano V1

laurienzu:
I updated the header to clarify wich usart to select when using UART or USB. Also the script after make was renamed to give the output bin a name that make possible to flash firmware when LCD is not attached.


KevinOConnor:
I'm not sure what the state of this PR is.  It looks like some of the comments are now outdated, and it would require a signed-off-by line ( https://www.klipper3d.org/CONTRIBUTING.html ).

-Kevin

PR #4832, opened 2021-10-24
Repeating mcu schedule insert position optimization

KevinOConnor:
This change adds a check to the low-level mcu scheduler to improve the performance of multiple timers that reschedule at the same period.  This type of rescheduling is common when multiple steppers control the same axis.  It is also common when making Z moves on a delta.

This is a result of the discussion at https://klipper.discourse.group/t/intermittent-tmc-driver-crash-unable-to-read-tmc-uart-registers/932 .

In particular, when homing a multi-stepper axis (or homing in general on a delta) the scheduler ends up rescheduling 6+ timers at the same time period.  The rescheduling of each of these timers results in a walk through the list of pending timers (eg, 6+) to find its new schedule location.  That results in an `O(n^2)` time complexity to reschedule for these types of repeating schedules.  This optimization avoids this rescheduling overhead (at least for these common multi-stepper setups).

The core change here involves storing the last "reschedule position" and checking if that position is a better starting point when rescheduling the next timer.  This makes the rescheduling process notably faster when many timers all reschedule at a similar interval.

This change actually results in a small pessimization of the benchmarks on most mcus.  However, I believe the change is worthwhile to make the scheduler more robust when dealing with repetitive timers.

-Kevin

### DIFF
--- a/src/sched.c
+++ b/src/sched.c
@@ -18,10 +18,10 @@
 static struct timer periodic_timer, sentinel_timer, deleted_timer;
 
 static struct {
-    struct timer *timer_list;
+    struct timer *timer_list, *last_insert;
     int8_t tasks_status;
     uint8_t shutdown_status, shutdown_reason;
-} SchedStatus = { .timer_list = &periodic_timer };
+} SchedStatus = {.timer_list = &periodic_timer, .last_insert = &periodic_timer};
 
 
 /****************************************************************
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Unfortunately, this PR had an error in it.  A fix for the error was implemented in commit 9ecddd1b.  All users are recommended to update - see https://klipper.discourse.group/t/flashing-of-micro-controller-code-needed/1263 for details.

-Kevin

PR #4835, opened 2021-10-25
Adds support for the STEVAL-3DP001V1 board from STmicroelectronics

h-oussama:
This PR supersedes https://github.com/Klipper3d/klipper/pull/4392 .

Signed-off-by: Oussama HOUIDAR houidar.ousama@gmail.com


config/generic-steval-3dp001v1.cfg
```diff
@@ -0,0 +1,327 @@
+# This file contains common pin mappings for STEVAL-3DP001V1 board.
+# to use this config, the firmware should be compiled for the STM32F4
```
JamesH1978, line 2: This will need to be expanded on, it has no mention of how you should set the make menuconfig, clocks, speeds, bootloader offsets etc. and then the method of flashing this file.

config/generic-steval-3dp001v1.cfg
```diff
@@ -0,0 +1,327 @@
...
+[fan]
+pin: PE8
+
+[heater_fan extruder]
+pin: PC4
+heater : extruder
+
+[mcu]
+serial: /dev/serial/by-id/usb-STMicroelectronics_STM32_STLink_0674FF514852897267245621-if02
+baud: 115200
```
JamesH1978, line 78: You have stated a custom speed, we generally do not set this unless their is a very good reason to do so, 250000 default works on the majority of known boards. Have you tested that it will not work at the default speed?

KevinOConnor:

-Kevin

KevinOConnor:
It seems this PR "slipped through the cracks" on my side.

Can you provide an update on the current status?  Is this something that should still be merged?

-Kevin

PR #4836, opened 2021-10-25
Added Config for Creality Ender 5 Pro 2020

SyrasX:
Idk if there is a reason why there was none. Would have had an easier time setting up Klipper with it.

Signed-off-by:  Maxim Schessler <maxim.schessler@gmail.com>


KevinOConnor:
I'm not sure what the state of this PR is.  At a minimum, the config would need to follow the guidelines at: https://www.klipper3d.org/Example_Configs.html .

-Kevin

SyrasX:
Removed the parts mentioned by @Fisheiyy. I added this config after getting my printer half a year ago and it works fine, Don't see anything not conforming with the guidlines after a short look. @KevinOConnor

Fisheiyy (approved):
Looks good to me

KevinOConnor:
Thanks.  It looks fine, but the file should be added to printers.test (point 4 at https://www.klipper3d.org/Example_Configs.html ).

-Kevin

SyrasX:
Done. @KevinOConnor

KevinOConnor:
Thanks.

-Kevin

PR #4839, opened 2021-10-25
Add option to PROBE command to report probe position instead of toolhead position.

Fabi0San:
PROBE command response implies that it is reporting the probe location:
`// probe at 0.908,-0.316 is z=0.002357
// Result is z=0.002357`
 but it is in fact reporting the toolhead location.

It may be a bug, but at this point I assume the current behavior of PROBE should stay stable as someone may depend on it.

So I created a new boolean optional parameter to PROBE for REPORT_PROBE_POSITION which takes into account the probe offset as configured.

Signed-off-by: Fabio Santos <fabiosan@live.com>

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -362,9 +362,12 @@ the [probe calibrate guide](Probe_Calibrate.md)):
 - `PROBE [PROBE_SPEED=<mm/s>] [LIFT_SPEED=<mm/s>] [SAMPLES=<count>]
   [SAMPLE_RETRACT_DIST=<mm>] [SAMPLES_TOLERANCE=<mm>]
   [SAMPLES_TOLERANCE_RETRIES=<count>]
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  I have the same comment as https://github.com/Klipper3d/klipper/pull/4820#issuecomment-1008488531 - it's not clear to me that there is a "high impact" benefit to adding a command parameter.  Note that it is already possible to obtain the last probe position using Klipper's macro system ( https://www.klipper3d.org/Status_Reference.html#probe ).

-Kevin

Fabi0San:
Thanks for the workaround suggestion @KevinOConnor.

But this behavior is very unique to Klipper and unlike any other motion control software from 3d printers to professional CNC mills and lathes, probing commands report probe position, not machine/turret/effector position, what would be the point of that? Proof is that on every place it is used inside Klipper it has to fetch offset and compute proper position.
Don't you think we should straighten that out?
Cheers,
   Fabio

PR #4843, opened 2021-10-26
gcode: Allow rename_existing also for register_mux_command

zellneralex:
﻿This PR allows now to use rename_existing at GCodes internally registered as register_mux_command
The changes where done by the discord user dalegaard.

At the tide integrated UIs like fluidd/mainsail you need to use the same GCode name as the base klipper. This was not possible for some of the GCodes. The user only sees an internal error with no explanation what the problem was. After the merge that problem would be gone.

I personally testet it with:
- SET_FILAMENT_SENSOR SENSOR
- SET_PIN
- SET_HEATER_TEMPERATURE
- ACTIVATE_EXTRUDER

Test Case looks like: 
```
[gcode_macro SET_PIN]
rename_existing: SET_PIN_BASE
gcode:
  ##### get params and prepare to send them to the base macro #####
  {% set get_params = [] %}
  {% for key in params %}
    {% set get_params = get_params.append(key + "=" + params[key])  %}
  {% endfor %}
  ##### end of definitions #####
  {action_respond_info(get_params|join(" "))}
  SET_PIN_BASE {get_params|join(" ")}
```
I also verified that rename_existing still works for GCodes registered as register_command. Used GCodes
- BED_MESH_CALIBRATE
- QUAD_GANTRY_LEVEL
- PAUSE
- RESUME

### DIFF
--- a/klippy/gcode.py
+++ b/klippy/gcode.py
@@ -126,7 +126,8 @@ class GCodeDispatch:
     def register_mux_command(self, cmd, key, value, func, desc=None):
         prev = self.mux_commands.get(cmd)
         if prev is None:
-            self.register_command(cmd, self._cmd_mux, desc=desc)
+            handler = lambda gcmd: self._cmd_mux(cmd, gcmd)
+            self.register_command(cmd, handler, desc=desc)
             self.mux_commands[cmd] = prev = (key, {})
         prev_key, prev_values = prev
         if prev_key != key:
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I agree this bug should be fixed.  However, I'm concerned about the performance impact of adding an additional python function to the command dispatch - calling a python function is slow (including lambdas).

FWIW, I was vaguely thinking we could add a new `rename_command()` method to the GCode class that gcode_macro could call - that new method could fixup the `self.mux_commands` dictionary.

-Kevin

dalegaard:
FWIW no mux commands are used in the common hot path, the only one that springs to mind as potentially affected by a performance degradation is `MANUAL_STEPPER`.

Changing the interface and introducing a new command to rename gcode commands feels like a premature optimization. Even knowing that function calls are slow, we are talking one extra function call outside the hot path of the vast majority of normal prints. Presumably the called command handler will take much longer to run than the single extra function call time.

A subtle case that would be broken by fixing up the dictionary is a module adding a mux command handler after `klippy:connect`. In that case the mux command dictionary key would have changed but the module would be trying to register to the original command name.

Just my two cents.

/Lasse

KevinOConnor:
Make sense.  I went ahead and committed this.  Thanks.

-Kevin

PR #4844, opened 2021-10-26
scripts: Leverage Docker Multistage Builds

mkuf:
Hi There, 

for a recent Project of mine called [prind](https://github.com/mkuf/prind), I build [Klipper Docker Images](https://hub.docker.com/r/mkuf/klipper) on a daily basis.  

These Images are decoupled from the actual Source and use git to fetch the code.  
Based on this work and my findings I reworked the existing Dockerfile to decrease image size and decouple the MCU buildprocess from the runtime Image.

See the following Comparison, where `master` refers to a Image built with the current Dockerfile.
`multistage-run` refers to the new runtime Image that executes klippy and `multistage-mcu` is used for building MCU code.

```
klipper         master            dbb0540ac413   46 minutes ago   1.79GB
klipper         multistage-run    87822351db88    2 minutes ago    443MB
klipper         multistage-mcu    8342679c52f9   23 seconds ago   1.65GB
```

While updating the Image itself I also added some documentation on common usecases.

If you deem this an Improvement this would certainly pave the way for an official klipper Docker Image that is universally applicable.

-Markus

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -1,44 +1,117 @@
-# This is an example Dockerfile showing how it's possible to install Klipper in Docker.
-# IMPORTANT: The docker build must be run from the root of the repo, either copy the
-# Dockerfile to the root, or run docker build with "-f", for example:
-#       docker build . -f scripts/Dockerfile -t klipper
-# Note that the host still needs to run Linux to connect the printers serial port to
-# the container.
-# When running, the serial port of your printer should be connected, including an
-# argument such as:
-#       --device /dev/ttyUSB0:/dev/ttyUSB0
-# It's also required that your control program (eg: OctoPrint) be included in the same
-# container as Docker does not allow sharing of the virtual serial port outside the
-# container.
-# The config should be in a file named "printer.cfg" in a directory mounted at:
-#       /home/klippy/.config/
-# For more Dockerfile examples with Klipper (and Octoprint) see:
-# https://github.com/sillyfrog/OctoPrint-Klipper-mjpg-Dockerfile
-FROM ubuntu:18.04
+# This is an example Dockerfile showing how it's possible to run Klipper in Docker.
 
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I don't have much experience with Docker, and I don't have much of a preferences on the contents of the example Dockerfile in the Klipper repo.  Maybe @sillyfrog can provide some comments.

-Kevin

speendo:
To overcome possible ressource conflicts, the method mentioned here might be useful:
https://github.com/mkuf/prind/issues/7

However, this is still untested.

sillyfrog:
@mkuf That's great having the smaller image, I have my setup so I can build the MCU inside the image as I have plenty of resources in my setup. I have included the `Dockerfile` and not make an image as I think there are going to be a _lot_ of different and competing requirements, making an official image fairly challenging. 
That said, it maybe worth including a comment in the current `Dockerfile` to link off to your project/Docker hub in case others are interested.
I like the concept of an image on Docker Hub, but would need to see what the general interest and uptake was. I don't have the bandwidth to maintain it, but happy to provide assistance if required from time to time.

mkuf:
Hi There,

when it comes to competing requirements I'd argue that the small Image is a common denominator for all Klipper users, as it contains the same things as a .deb or .rpm package that ships klipper. Anything else is up to the User to set up in their Environment. 

The Images themselves could be situated in the ghcr, without the requirement of an external docker registry.

I'd be happy to maintain the official Docker Image for klipper in the future.

-Markus

mkuf:
As I did not receive any further feedback, and the Docker file is no longer compatible with the current state of klipper, I'll close this PR. 

I'm happy to bring this to state where it is working again, if there is actually any intent on merging this.  
Anyone interested in this functionality may have a look at [prind](https://github.com/mkuf/prind), where I maintain ready to use klipper and moonraker images.

PR #4847, opened 2021-10-27
idle_timeout: exit the timer if the printer is shutdown

Arksine:
This resolves an issue where the idle_timeout attempts to transition from "Ready" to "Idle" when the printer is shutdown.  Currently this will result in a failed attempt to run the idle gcode, which will repeat every second until Klippy is restarted.

The following steps can be taken to reproduce the issue:
1) Set the idle timeout to a low value: `SET_IDLE_TIMEOUT TIMEOUT=2`
2) Run a command to put the printer in the `Printing` state: `G4 P1000`
3) Emergency Stop


KevinOConnor:
Thanks.

-Kevin

PR #4848, opened 2021-10-28
STM32H7 - Add UART4

adelyser:
STM32H7 SERIAL

Add a UART4 definition. This allows UART4 to be used on stm32h7 processors. This UART exists on the BTT-SKR-SE-BX board.


KevinOConnor:
Thanks.

-Kevin

PR #4849, opened 2021-10-28
STM32H7 - Add USBOTG support

adelyser:
STM32H7 USBOTG

Add USBOTG support for the STM32H7 processor. 2 sets of pins were defined as there are 2 USBOTG ports on the BTT-SKR-BX-SE board.


src/stm32/usbotg.c
```diff
@@ -382,30 +410,33 @@ OTG_FS_IRQHandler(void)
...
 
-DECL_CONSTANT_STR("RESERVE_PINS_USB", "PA11,PA12");
-
 // Initialize the usb controller
 void
 usb_init(void)
 {
     // Enable USB clock
+#if CONFIG_MACH_STM32H7
+    if (READ_BIT(PWR->CR3, PWR_CR3_USB33RDY) != (PWR_CR3_USB33RDY) ? 1 : 0) {
```
KevinOConnor, line 419: What does this `if` do?  I don't understand the `? 1 : 0` at the end.

-Kevin

adelyser: The end test is not needed, I guess I missed that on cleanup. I had gotten that form the stm32h7 library I was referencing for checking the USB 3.3V to be ready. I will remove the "? 1 : 0" from the if-statement and re-test.

adelyser: I have removed the extra logic from within the if-statement and tested the changes. Thanks for questioning that!

KevinOConnor:
Unfortunately, it looks like this PR isn't complete - it references a CONFIG_STM32_USB_PB14_PB15 but doesn't define it anywhere.  Similarly, it adds USB code for the stm32h7, but doesn't enable the USB code in the build for stm32h7.  What was the test process for this change?

-Kevin

adelyser:
Per comments in #4841, I broke up the big PR into separate topics. I have a separate branch for the STM32H743xx base changes,  which I didn't feel was ready until these PR's were added to master.

Currently, the CONFIG_STM32_USB_PB14_PB15 isn't defined, as that is part of the stm32h743xx branch I have. I can remove that piece and add it to that branch if you prefer.

The USB code isn't enabled currently for the STM32H7 processors. I have the STM32H743 changes to enable that in a separate branch as recommended. I felt that branch shouldn't be sent to a PR until the underlying fixes were in place to prevent bug reports on it not working, so that PR will follow all of my current PR's once they are approved.

For testing, I have a branch with ALL of my changes merged into it as one, and I use that for testing and building everything together. My stm32h743xx branch also has the test code added to build that model too in the runner upon commiting to the repo.

KevinOConnor:
Thanks.  We really want to break up the changes into a series of commits that are each independent and functional.  This is important because it facilitates tools like `git bisect` and `git revert`.  In particular, for the mcu code, most users lack the hardware to debug a failure - so having the ability to utilize `git bisect` when there is a regression is very useful.

So, if we're adding stm32h7 USB code we really want to add all the code for it in one commit; and mix no other features in that commit.  I know this is more upfront work, but this development approach has proven very effective.

What other changes would be necessary to support USB on the stm32h7?  If there are currently some bugs that prevent the stm32h7 code from running at all, is there a set of bugfixes that could be applied first, and then go on to add USB support?

-Kevin

adelyser:
I would argue that this PR is the stm32h7 usb bugfixes, and the stm32h743xx PR will enable the USB support, along with all the other pieces needed for the stm32h743 such as SPI, serial, bootloader offsets, etc....

This PR makes the USB functional on the stm32h7, while the stm32h743xx PR will enable them for use. Should I combine the general stm32h743xx changes in with these USBOTG changes?

KevinOConnor:
Okay, I'm confused.  I thought this PR was adding functionality (support for USB on stm32h7).  If this PR has a bug fix of some kind then we need to identify that and it must be in a commit separate from any feature enhancements.  That is, we must not mix adding new features with unrelated changes.  Otherwise, it denies us the ability to effectively use `git bisect` and `git revert`.

I do understand that reworking the commits into small functional changes is a lot of work.  However, my experience is that doing so dramatically improves our ability to later find and fix problems.  That is, the additional upfront work is ultimately a huge savings in work.  This development methodology is also used in other popular open source projects (eg, Linux and QEMU to name a few).

Cheers,
-Kevin

KevinOConnor:
Thanks.  I made some minor changes to usbotg.c and committed this change (commit 4eeb4620).

-Kevin

PR #4850, opened 2021-10-28
STM32H7 - Add SPI support

adelyser:
STM32H7 SPI

Add hardware SPI support for the STM32H7 processor. Tested with an ADXL345 on a BTT-SKR-BX-SE board.


KevinOConnor:
Thanks.  I'm confused by the changes to both spi.c and stm32h7_spi.c .  I'm guessing this is two logically different topics in one commit?  If so, I think the changes to the existing spi.c should be in a separate commit.

Also, I'm not sure using `__COUNTER__` is a good plan, as it would be difficult to verify that some system header (or other Klipper header) wont use it.  Since the `#if` declarations have to line up between `DECL_ENUMERATION` and `spi_bus`, I'm not sure of the gain.  Also, it's not necessary for stm32h7_spi.c to check for CONFIG_STM32_USB_PB14_PB15 nor CONFIG_MACH_STM32F1.

Thanks again,
-Kevin

adelyser:
Well, there are 2 topics, but they started as one. I originally added the stm32h7 code to spi.c, but felt there were too many stm32h7 specific changes, so I pulled those out into it's own file. When I started though, I really didn't like the nested definitions for all the spi bus configurations. I created what I thought was a "cleaner" method, and tested the output for all the different stm32 variants to ensure the definitions aligned with the array.

FYI, __COUNTER__ starts at 0 for each file. If __COUNTER__ is used in a different file, it won't affect this one as they are separate. I was also concerned with this and I did test it to make sure.

I have some new additions to this branch that I was thinking about adding. I created a spi.h to store the spi bus defintions in, shared between both spi.c and stm32h7_spi.c. Is this a better solution? I don't mind reverting the spi.c changes if you think I should keep them separeate, but they do share a lot of the common busses.

Lastly, the STM32H7 does need to check for CONFIG_STM32_USB_PB14_PB15. Those pins are reserved for another USB bus when selected, and they are used on the SKR-SE-BX board. The config PR will have those additions when it's ready.

KevinOConnor:
Thanks.  I agree that separating the stm32h7 code to its own stm32h7_spi.c is a good idea.  I suspect refactoring the spi code (eg, adding an spi.h file) will increase complexity without much gain.

If you wish to use `__COUNTER__` in stm32h7_spi.c then that's fine.  I don't think it adds much to spi.c - I fear it would increase code churn without much gain.  Changes to spi.c would need to be in a separate commit from this, but my suggestion would be to leave spi.c unchanged.

It is not necessary to test for `CONFIG_STM32_USB_PB14_PB15` in stm32h7_spi.c, as it only makes a declaration for the host to inspect.  The host would never utilize that definition because it will detect the conflict via the "reserved pins" declarations.  So, there's no harm in exporting it.  In general, it's my preference to reduce the number of #ifdefs in the code as I find that makes the code a bit more readable and improves compiler coverage.

-Kevin

adelyser:
Ok, so I reverted spi.c to it's former glory. I also removed the CONFIG_STM32_USB_PB14_PB15 as you are correct. I tested it and it does work fine.

KevinOConnor:
Thanks.

-Kevin

PR #4851, opened 2021-10-29
Support custom step pulse duration at run-time and optimize for tmc drivers

KevinOConnor:
The current mcu build defaults to a 2us step pulse duration.  Changing this duration requires recompiling of the micro-controller code today and any change applies to all steppers on the given mcu.  This PR changes the code so that the pulse duration is no longer a compile-time setting and is instead configured for each stepper in a new `step_pulse_duration` option in the printer.cfg file.

Along with this change, TMC drivers configured via UART or SPI mode will automatically enable new optimizations for generating step pulses.  In particular, the `DEDGE=1` driver setting is enabled so that steps are taken on both rising and falling edges of the step gpio pin.  These optimizations make it possible for the default mcu build to obtain dramatically higher step rates on 32-bit micro-controllers.  (Many of these chips will now be able to reach over a million steps per second.)

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,11 @@ All dates in this document are approximate.
 
 ## Changes
 
+20211104: The "step pulse duration" option in "make menuconfig" has
+been removed. A new `step_pulse_duration` setting in the
+[stepper config section](Config_Reference.md#stepper) should be set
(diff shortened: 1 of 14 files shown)


Sineos:
This is really cool: On my SKR1.4T I can set 256 microsteps with no interpolation for all steppers

PR #4852, opened 2021-10-30
SKR E3 Turbo builtin board thermistor

zomfg:
Minor update to `generic-bigtreetech-skr-e3-turbo.cfg`  to add the builtin 100K NTC

image


KevinOConnor:
I'm not sure what the state of this PR is.  At a minimum, commits need to have a signed-off-by line containing your real full name - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

Issue #4838, opened 2021-10-25
Fysetc Spider flash-sdcard.sh script fails


falo2k:
Details of issue and testing of fix outlined at this issue on fysetc github FYSETC/FYSETC-SPIDER#38

The profile for the Fysetc Spider flash-sdcard.sh script fails with an OSError.  The board_defs.py configuration appears to be incorrect, and should instead be:
```
'fysetc-spider': {
        'mcu': "stm32f446xx",
        'spi_bus': "swspi",
        "spi_pins": "PA6,PA7,PA5",
        "cs_pin": "PA4",
        "current_firmware_path": "OLD.BIN"
    }
```

PR #4856, opened 2021-10-30
Fysetc Spider SD Card Upload Fix

falo2k:
flash_sdcard / board_defs: Fix to allow Fysetc Spider to use flash_sdcard.sh correctly

Fix for issue #4838 - corrects the board defs for the Fysetc Spider.  Issue on Fysetc github has testimonials for fix working on both v1.0 and v1.1 of board.  As the S6 has not been tested, I have kept the original board definitions for the S6 split out as a separate entry.

Signed-off-by: Oli Wright (github@falo.co.uk)

### DIFF
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -50,9 +50,16 @@ BOARD_DEFS = {
         'spi_bus': "spi1",
         "cs_pin": "PA4"
     },
-    'fysetc-spider': {
+    'fysetc-s6': {
         'mcu': "stm32f446xx",
         'spi_bus': "spi1",
+        "cs_pin": "PA4",
+        "current_firmware_path": "OLD.BIN"
+    },
+    'fysetc-spider': {
+        'mcu': "stm32f446xx",
+        'spi_bus': "swspi",
+        "spi_pins": "PA6,PA7,PA5",
         "cs_pin": "PA4",
         "current_firmware_path": "OLD.BIN"
     }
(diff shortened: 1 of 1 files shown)


adelyser:
This doesn't seem like a proper fix. If the SPI bus isn't working correctly, it seems that the SPI bus should be debugged and fixed rather than just switching to software SPI.

falo2k:
"Proper" in what way?  The important thing is fixing functionality of updating the current version of the Spider, rather than some purity over what method to use.  If you have another way to do it, then do so.

adelyser:
It's not about "purity". The issue is the SPI bus isn't working. Switching it to use software SPI doesn't fix the fact that the SPI bus isn't working. Other SPI devices will also have issues. Rather than apply a band-aid, the underlying issue should be fixed. Software SPI was intended for using SPI on non-SPI pins.

falo2k:
While there may be a wider issue that might need looking at, the feature isn't "have a working SPI bus", it's "flash Spider over SD".  That feature doesn't work, this makes it work.

adelyser:
This isn't a feature change, you're fixing an issue (as stated in your title). Why not fix the root cause of the issue?

falo2k:
What a strange semantic argument.  It's a fix to an issue with a feature that allows flashing of spider boards.  I didn't say it was a new feature.  Switch the words around how you want - the goal is to be able to flash the spider board.  As I said, feel free to fix it in a different way and submit a PR.

Arksine:
I took a look at this and submitted a Pull Request that should resolve the underlying issue.  We have seen this happen before with STM32F1 bootloaders.  Some bootloaders fail to reset peripheral clocks and registers before jumping to the application, this can leave the peripheral inoperable should the application try to use it.  In this case the SPI1 peripheral is malfunctioning.

I have tested this with a Spider on a bench, however it would be a good idea for some Spider users to checkout PR #4885 and verify that it flashes correctly then functions as normal when printing.

falo2k:
Fixed by #4885

PR #4859, opened 2021-11-01
Deprecate "NTC 100K beta 3950" thermistor

KevinOConnor:
This PR moves the default thermistor definitions from thermistor.py to a new temperature_sensor.cfg file, introduces a new "Generic 3950" thermistor, and deprecates the "NTC 100K beta 3950" definition.

As discussed in #4054 , it seems most real-world "beta 3950" sensors don't actually follow a true "beta 3950" temperature curve, and instead follow a curve defined in the new "Generic 3950" table.  Deprecating the "NTC 100K beta 3950" definition seems like the best way to raise awareness and move printers over to a more accurate configuration.

-Kevin

Edit: Change name of new thermistor to "Generic 3950".

### DIFF
--- a/config/generic-bigtreetech-skr-2.cfg
+++ b/config/generic-bigtreetech-skr-2.cfg
@@ -73,7 +73,7 @@ max_temp: 250
 
 [heater_bed]
 heater_pin: PD7
-sensor_type: NTC 100K beta 3950
+sensor_type: Generic 3950
 sensor_pin: PA1
 control: watermark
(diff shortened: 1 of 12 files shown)


KevinOConnor:
After a little further thought, I think "Generic 3950" is probably a better name than "RepRap 3950".  I have updated this PR accordingly.

-Kevin

Jack77777777:
I am curious when this will be released?

PR #4869, opened 2021-11-03
duplicate pin number

kanawati975:
how come endstop_pin and diag1_pin share the same pin 1.27 ?


zellneralex:
That is how BTT made there schematic and for you to cut the diag pin in there manual if you want to use the endstop pin

kanawati975:
@zellneralex I know it's not a mistake in the config file. I cross referenced it with all available data from BTT. 
The question is: What if I want to use both? Just use another pin for endstop?

zellneralex:
Btw you did it only for one pin but there are  several connectors that share the uC pin between the pin used at the connector and a diag pin. On any of them you can not use both in parallel

KevinOConnor:
It's not clear to me what this PR is intended to do.

-Kevin

PR #4870, opened 2021-11-04
config: Add HermitCrab board

bigtreetech:

### DIFF
--- a/config/sample-bigtreetech-hermit-crab-canbus.cfg
+++ b/config/sample-bigtreetech-hermit-crab-canbus.cfg
@@ -1,0 +1,65 @@
+# This file contains common pin mappings for the BIGTREETECH HermitCrab
+# Canbus board. To use this config, the firmware should be compiled for the
+# STM32F072 with "8 MHz crystal" and "USB (on PA11/PA12)" or "CAN bus (on PB8/PB9)".
+# The "HermitCrab" micro-controller will be used to control the components on the nozzle.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[mcu HermitCrab]
+serial: /dev/serial/by-id/usb-Klipper_Klipper_firmware_12345-if00
+#canbus_uuid: 0e0d81e4210c
+
+[adxl345]
+cs_pin: HermitCrab: PB12
+spi_bus: spi2
+axes_map: y,z,-x
+
+[extruder]
+step_pin: HermitCrab: PA6
+dir_pin: !HermitCrab: PA7
+enable_pin: !HermitCrab: PA5
+microsteps: 16
+rotation_distance: 33.500
+nozzle_diameter: 0.400
+filament_diameter: 1.750
+heater_pin: HermitCrab: PA2
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: HermitCrab: PA1
+control: pid
+pid_Kp: 21.527
+pid_Ki: 1.063
+pid_Kd: 108.982
+min_temp: 0
+max_temp: 250
+
+[tmc2209 extruder]
+uart_pin: HermitCrab: PB0
+run_current: 0.650
+hold_current: 0.500
+stealthchop_threshold: 999999
+
+[fan]
+pin: HermitCrab: PA4
+
+[heater_fan hotend_fan]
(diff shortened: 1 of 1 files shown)


config/sample-bigtreetech-hermit-crab-canbus.cfg
```diff
@@ -0,0 +1,128 @@
...
+# STM32F072 with "8 MHz crystal" and USB (on PA11/PA12) or CAN bus (on PB8/PB9).
+# The "HermitCrab" micro-controller will be used to control the components on the nozzle.
+
+[mcu HermitCrab]
+#canbus_uuid: 0e0d81e4210c
+#serial: /dev/serial/by-id/usb-Klipper_Klipper_firmware_12345-if00
+
+[adxl345]
+cs_pin: HermitCrab: PB12
+spi_software_sclk_pin: HermitCrab: PB13
```
dmbutyugin, line 12: Any reason this is configured as a software SPI instead of a hardware SPI as
```
cs_pin: HermitCrab: PB12
spi_bus: spi2
```
?

config/sample-bigtreetech-hermit-crab-canbus.cfg
```diff
@@ -0,0 +1,128 @@
...
+accel_chip: adxl345
+probe_points:
+    100,100,20  # an example
+
+[extruder]
+step_pin: HermitCrab: PA6
+dir_pin: !HermitCrab: PA7
+enable_pin: !HermitCrab: PA5
+microsteps: 16
+rotation_distance: 3.45
```
dmbutyugin, line 35: The rotation_distance seems to be too small: it is typically an order of magnitude bigger. Not that it matters too much for the sample, and the users are expected to update it anyway, but maybe it would be great to specify a more realistic value from the start?

dmbutyugin (commented):
Thank you!

BTW, I'd think that a full name is required on a signed-off-by line:
https://www.klipper3d.org/CONTRIBUTING.html#submitting-a-pull-request

config/sample-bigtreetech-hermit-crab-canbus.cfg
bigtreetech: Modified to hardware SPI

config/sample-bigtreetech-hermit-crab-canbus.cfg
bigtreetech: Yeah, 3.45 is for H2 extruder.  It has been changed to 33.5 for Common extruders

dmbutyugin: Ah, I see, my bad. Then please feel free to revert the change and keep 3.45 as the rotation distance.

dmbutyugin:
Thank you!

Just some very minor things:
- I don't think we need 2-3 separate commits for this feature. It might be easier for you to pack everything into a single commit (e.g. using `git reset --soft "HEAD~3"` and then `git commit --amend` and `git push --force`). Otherwise, you would need to include signed-off-by line in every commit.
- If you could change the title of the commit to 'config: Add HermitCrab board', that would be great (this is the [preferred formatting](https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#submitting-a-pull-request) of the commit title).

KevinOConnor:
FYI, I can generally "squash and rebase" simple PRs like this, so rebasing isn't a requirement for merging (though it is always nice to see).

This PR should be changed to comply with item 6b at https://www.klipper3d.org/Example_Configs.html#guidelines though.

If using the "H2 extruder" for rotation_distance it would be nice to follow section 6e in the document above, but it is not a requirement for merging.

Thanks,
-Kevin

dmbutyugin:
Thanks! I think the PR looks good now.

@KevinOConnor it seems that 6b is now resolved. I somehow missed that point when reading the guidelines, so thank you for pointing that out. Rotation distance is now for a generic extruder (33.5) and matches the other examples.

dmbutyugin (approved):

bigtreetech:
Yeah I think it's ready to merge. If there is anything wrong, please inform me to modify it.

KevinOConnor:
Thanks.

-Kevin

PR #4876, opened 2021-11-04
Update generic-mellow-super-infinty-hv.cfg to use hardware SPI

tehmaze:
TMC2130/5160 can make use of `spi3`, which is operated on the software SPI pins in the current configuration. The supplied defaults for software SPI do not currently work.

Signed-Off-By: Wijnand Modderman-Lenstra <maze@maze.io>


KevinOConnor:
In general, PRs are on hold pending a new process for them (#4949).

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #4878, opened 2021-11-04
Workflow change - shutting down github issue tracker

KevinOConnor:
I'm planning to disable the Klipper "GitHub issue tracker".  I'm also looking at enabling the "GitHub Discussions" feature.  Going forward, I'm hoping there will be volunteers to help review PRs and identify confirmed software defects.

I'm finding that the GitHub issue tracker is becoming more of a burden than a help.  It seems many people have different views of what the issue tracker should be used for - some view it as a tool to track software defects, some as a todo list, some as a ticket tracker for requests, some as an area for general discussions, and many more views.  Unfortunately, this makes it difficult to manage - I feel it becomes a poor tool for tracking once the number of open tickets increases beyond a few dozen.  However, trying to limit what topics are "valid" on the issue tracker is not working well - it seems too many people have alternative views on what it should be used for - I fear attempts to manage it are consuming time and causing unnecessary conflict.  I also fear that not managing the issue tracker (no enforced guidelines ; no proactive closing of tickets) will give a poor image of the project - those people that view the github issue tracker as a tool for tracking defects may make a conclusion about the health of the project by the number of open "issues".

I'm hoping that using the more neutral "Discussions" feature of GitHub will work better.  In mechanics, the Discussions feature seems to be similar to the issue tracker, but it hopefully wont have this problem of differing expectations.  It'll be an area for discussions - not just discussions on "issues".

I'm also looking for long term contributors to Klipper to help volunteer with doing reviews.  Both reviews of GitHub Discussions and to help review GitHub Pull Requests.  By "long term contributor" I'm referring to people that have had regular activity helping users on Discourse, Discord, or contributing PRs to Klipper over at least six months.  It's not necessary to be a developer to be a "contributor" - many of the topics we need help with are reviews of bug reports, reviews of documentation changes, reviews of config files, reviews of PRs for copyrights/sign-offs, etc.  If interested, let me know.

The change to the Contact.md document in this PR removes the link to the Klipper Discourse server.  I don't have any plans to shutdown the Klipper Discourse.  However, there is quite a bit of overlap between the functionality of "GitHub Discussions" and "Discourse" - the proposal here is to see how well "GitHub Discussions" works as the primary "forum" for Klipper.  We may ultimately run both servers indefinitely, or choose just one of them.

Finally, I know this may seem as "yet another change" to the preferred Klipper forum (github issue tracker, mailing list, Discourse, GitHub Discussions, etc.).  It's certainly a concern of mine as well.  I can just say that I'm trying my best and hope we can come up with a solution that works well for the community.

-Kevin

### DIFF
--- a/docs/Contact.md
+++ b/docs/Contact.md
@@ -2,7 +2,7 @@
 
 This document provides contact information for Klipper.
 
-1. [Community Forum](#community-forum)
+1. [GitHub Discussions](#github-discussions)
 2. [Discord Chat](#discord-chat)
 3. [I have a question about Klipper](#i-have-a-question-about-klipper)
 4. [I have a feature request](#i-have-a-feature-request)
@@ -10,11 +10,11 @@ This document provides contact information for Klipper.
 6. [I have diagnosed a defect in the Klipper software](#i-have-diagnosed-a-defect-in-the-klipper-software)
 7. [I am making changes that I'd like to include in Klipper](#i-am-making-changes-that-id-like-to-include-in-klipper)
 
-## Community Forum
+## GitHub Discussions
 
 There is a
-[Klipper Community Discourse server](https://community.klipper3d.org)
-for discussions on Klipper.
+[Klipper Discussion area on GitHub](https://github.com/Klipper3d/klipper/discussions)
+for Klipper topics.
 
 ## Discord Chat
 
@@ -32,11 +32,11 @@ Many questions we receive are already answered in the
 documentation and follow the directions provided there.
 
 It is also possible to search for similar questions in the
-[Klipper Community Forum](#community-forum).
+[Klipper Discussion Forum](#github-discussions).
 
 If you are interested in sharing your knowledge and experience with
 other Klipper users then you can join the
-[Klipper Community Forum](#community-forum) or
+[Klipper Discussion Forum](#github-discussions) or
 [Klipper Discord Chat](#discord-chat). Both are communities where
 Klipper users can discuss Klipper with other users.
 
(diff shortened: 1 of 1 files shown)


Sineos:
Personally I have no strong emotions regarding Discourse or Github Discussions. Discourse seems to be a bit more flexible and "mature" but not substantially.

What I think is important:
* Users need a platform to address their problems
* Information (historical and current) need to be concentrated in one location
* Supporter resources need to be concentrated in one location

Switching to GH Discussions is fine. In this case Discourse should be shut down or set to read-only for a period of time and then shut down. Additionally it might make sense to copy some good information like contributed configs or howtos in in the KlipperScreen section over to GH.

prcutler:
That sounds more than fair.  It's hard enough maintaining a project, you should be able to choose the best way for you to manage some of the feedback.

One you merge the the PR, have you considered linking to it directly from the README to help direct people right away?

I've been lurking in Discord for the last few months, but I'll keep an eye on Discussions and see how I can help, especially on trying to reproduce bugs or I have had experience writing documentation for other open source projects if you want any help.

KevinOConnor:
@Sineos 
> What I think is important: [...]

I agree.

@prcutler 
> have you considered linking to it directly from the README to help direct people right away

It's possible to do that, but I think I'd rather encourage people to read the documentation before reaching out.

-Kevin

KevinOConnor:
After looking at this a little closer, I don't see a good way to disable the github issue tracker.  I can completely disable it, but that would remove access to its history.  As long as it is viewable, it seems that anyone can open a new ticket.

Considering that, I think this topic needs more thought before proceeding.  I do think the core idea of having more "reviewers" is needed, but I don't think we need to tie that to Issues tracker/Discussions/Discourse.  So, I'm withdrawing this proposal for now.

Thanks,
-Kevin

PR #4880, opened 2021-11-05
Update benchmarks after stepper and scheduler changes

KevinOConnor:
This PR updates the benchmarks after recent code changes.  There have been dramatic increases to the 32bit micro-controllers due to PR #4851 .

Unfortunately, I don't have access to the Duet2 Maestro nor the Duet2 Eth/wifi.  It would be great if someone with each of those boards could run through the steps at https://github.com/Klipper3d/klipper/blob/work-benchmarks-20211104/docs/Benchmarks.md and report the 1 stepper and 3 steppers benchmarks for each of those boards.

-Kevin

### DIFF
--- a/docs/Benchmarks.md
+++ b/docs/Benchmarks.md
@@ -67,10 +67,9 @@ in preparation for the next test:
 clear_shutdown
 ```
 
-To obtain the single stepper and dual stepper benchmarks, the same
-configuration sequence is used, but only the first block (for the
-single stepper case) or first two blocks (for the dual stepper case)
-of the above test is cut-and-paste into the console.py window.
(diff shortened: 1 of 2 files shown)


eamaclean:
Assuming I have not messed this up, the lowest ticks values I achieved on a Duet 2 Wifi were 48 for single stepper and 215 for three, or 2500K and 1674K. This was with gcc version 10.3.1 20210621 (release) (GNU Arm Embedded Toolchain 10.3-2021.07), if you want that for the record.

KevinOConnor:
Thanks!  I updated this PR with your results.

Hopefully someone will be able to test the Maestro as well.

-Kevin

PR #4530, opened 2021-07-27
Add SKR2 board support for SD flash

kmarty:
I tried it on my SKR2 board, it quite works except "full reset" of the board didn't happen (I have to reset manualy, by reset button or power cycle).
I can see this:
```
pi@fluiddpi:~/klipper $ ./scripts/flash-sdcard.sh /dev/serial/by-id/usb-Klipper_stm32f407xx_2B0023000247393439313338-if00 btt-skr-2
Flashing /home/pi/klipper/out/klipper.bin to /dev/serial/by-id/usb-Klipper_stm32f407xx_2B0023000247393439313338-if00
Checking FatFS CFFI Build...
Connecting to MCU...Connected
Checking Current MCU Configuration...Done
MCU needs restart: is_config=1, is_shutdown=0
Attempting MCU Reset...Done
Waiting for device to reconnect....Done
Connecting to MCU...Connected
Initializing SD Card and Mounting file system...

SD Card Information:
Version: 2.0
SDHC/SDXC: True
Write Protected: False
Sectors: 62333952
manufacturer_id: 3
oem_id: SD
product_name: SL32G
product_revision: 8.128
serial_number: ECAEE005
manufacturing_date: 1/2019
capacity: 29.7 GiB
fs_type: FAT32
volume_label: ENDER
volume_serial: 1081870871
Uploading Klipper Firmware to SD Card...Done
Validating Upload...Done
Firmware Upload Complete: firmware.bin, Size: 22328, Checksum (SHA1): D4516394B634E879F4A39CBA89E25C8CB548F34A
Attempting MCU Reset...Done
Waiting for device to reconnect....Done
Connecting to MCU...Connected
Verifying Flash...
SD Card Flash Error: Version Mismatch: Got 'v0.9.1-629-g0075b290-dirty-20210724_233606-fluiddpi...', expected 'v0.9.1-632-g7e88f922-dirty-20210727_144546-fluiddpi...'
Traceback (most recent call last):
  File "/home/pi/klipper/scripts/spi_flash/spi_flash.py", line 1159, in main
    spiflash.run()
  File "/home/pi/klipper/scripts/spi_flash/spi_flash.py", line 1113, in run
    self.run_reactor_task(self.run_verify)
  File "/home/pi/klipper/scripts/spi_flash/spi_flash.py", line 1095, in run_reactor_task
    k_reactor.run()
  File "/home/pi/klipper/klippy/reactor.py", line 269, in run
    g_next.switch()
[...]

PR #4881, opened 2021-11-05
Add support for BTT octopus to flash-sdcard.sh

2be2d:
Using swspi updates work for BTT octpus. Tested and works.

```
$ ./scripts/flash-sdcard.sh /dev/ttyAMA0  btt-octopus-v1
Flashing /home/pi/klipper/out/klipper.bin to /dev/ttyAMA0
Checking FatFS CFFI Build...
Connecting to MCU...Connected

SD Card Information:
Version: 2.0
SDHC/SDXC: True
Write Protected: False
Sectors: 7774208
manufacturer_id: 39
oem_id: PH
product_name: SD4GB
product_revision: 5.80
serial_number: DABB21FB
manufacturing_date: 7/2017
capacity: 3.7 GiB
fs_type: FAT32
volume_label: boot
volume_serial: 1575249500
Validating Upload...Done
Firmware Upload Complete: firmware.bin, Size: 20104, Checksum (SHA1): D7E8740FDB68C459B75EF651274AA2AFDA0C751E
Attempting MCU Reset...Done
Waiting for device to reconnect...Done
Connecting to MCU...Connected
Found and deleted firmware.bin after reset
Firmware Flash Successful
Current Firmware: v0.10.0-1-gb806d71e-dirty-20211105_051612-raspberrypi
Attempting MCU Reset...Done
SD Card Flash Complete
```

### DIFF
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -55,6 +55,12 @@ BOARD_DEFS = {
         'spi_bus': "spi1",
         "cs_pin": "PA4",
         "current_firmware_path": "OLD.BIN"
+    },
+    'btt-octopus': {
+        'mcu': "stm32f446xx",
+        'spi_bus': "swspi",
+        "spi_pins": "PC8,PD2,PC12",
+        "cs_pin": "PC11",
     }
 }
 
@@ -88,7 +94,10 @@ BOARD_ALIASES = {
     'mks-robin-e3d': BOARD_DEFS['mks-robin-e3'],
     'fysetc-spider-v1': BOARD_DEFS['fysetc-spider'],
     'fysetc-s6-v1.2': BOARD_DEFS['fysetc-spider'],
-    'fysetc-s6-v2': BOARD_DEFS['fysetc-spider']
+    'fysetc-s6-v2': BOARD_DEFS['fysetc-spider'],
+    'btt-octopus-v1': BOARD_DEFS['btt-octopus'],
+    'btt-octopus-v1.1': BOARD_DEFS['btt-octopus'],
+    
 }
 
 def list_boards():


Arksine:
Hi, you will need to add a signature to your commit as outlined in the [contributing documentation](https://www.klipper3d.org/CONTRIBUTING.html).  It would probably be a good idea to reword the commit message as well, it isn't necessary to put the debug output in there.  Something like the following would probably work:
```
flash-sdcard: add support for BTT Octopus

Signed-off-by:  Your Real Name <valid-email-address>
```

With regard to the change itself I have some questions.  My understanding is that the BTT Octopus uses the SDIO peripheral to connect to the sdcard.  Generally speaking this is not compatible with `flash-sdcard`, as you cannot switch between sdio and spi modes without power cycling the sdcard.  The output you provided shows it working, are you using the stock bootloader?   I think it would be a be a good idea if we could get multiple testers to confirm that this patch works as well in the event that BTT has different bootloaders out for the Octopus.  Thanks.

tobyfu:
I got what looked like a successful flash using this however it was only because I was attempting to flash the same version that was already installed, it seems the file is successfully copied to the SD card but does not flash, when the script compares the version numbers they match and so it assumes it was successful. However if I introduce a change to the MCU code, recompile and run the script again the version number doesn't match because it hasn't actually flashed and the script reports an error.

I remember the skr2 pull request #4530 and the MCU needed to actually be power cycled before it would flash so I went and turned the printer off and on again and then the version numbers matched and klipper was happy to connect.

I guess we would need BTT to give us a bootloader that will flash in SPI mode in order for this to work without any manual intervention.

Arksine:
Thanks for the feedback tobyfu, this sounds like the behavior I expected.  In your case, it is the bootloader that fails to init the SD Card because `flash-sdcard` initializes it in SPI mode.

As more and more boards use the SDIO peripheral it might be worth looking into adding SDIO support in Klipper.  If I get a chance it may be something I take a look at.

2be2d:
Interesting, i assumed the -dirty suffix would be the new modified version; maybe i messed that up somehow.

SDIO support would be most welcome, as this also affect the spider in at least one version of their bootloader.

miklschmidt:
@Arksine I would very much appreciate SDIO support as well! That would give us a way to flash the Octopus Pro 429 and the upcoming SKR 2 429.

KevinOConnor:
> As more and more boards use the SDIO peripheral it might be worth looking into adding SDIO support in Klipper. If I get a chance it may be something I take a look at.

Just out of curiosity, have you looked to see if it is possible to reset the sdcard from the sdio peripheral?  That is, it's not possible to reset the card using spi, but maybe the sdio registers can?  I've worked with sd cards in other projects and I know they can be tricky - but I vaguely recall the sdio interface having better control capabilities.

-Kevin

Arksine:
> Just out of curiosity, have you looked to see if it is possible to reset the sdcard from the sdio peripheral? That is, it's not possible to reset the card using spi, but maybe the sdio registers can? I've worked with sd cards in other projects and I know they can be tricky - but I vaguely recall the sdio interface having better control capabilities.

I haven't explored it, but I agree that its worth looking at.  From my recollection of the specification I think you are correct...the SPI mode contains a subset of the commands available through SDIO.

r0bb10:
@Arksine yeah SDIO support would be sweet, i bet that would be the common configuration for future boads also.

KevinOConnor:

-Kevin

rwijnhov:
Is this working now?

dowjames:
we need this to work

Budavaril:
we need this to work

miklschmidt:
FYI, you can flash the octopus via DFU if you make sure there's no SD card in the card slot. Just do the usual `make flash FLASH_DEVICE=/dev/whateverserialpath`. SDIO support would however be nice for those cards where DFU is broken such as the SKR 2 w/ STM32F429 and the Octopus Pro w/ SMT32F429.

PR #4883, opened 2021-11-06
Pass `commandline` as an additional macro param

whi-tw:
Although `params` is useful in most cases, M117, for instance, does not use the input as a list of parameters, but as a string.
For these cases, it is useful to have the full command line available.

An example macro would be:

```
[gcode_macro M117]
rename_existing: M117.1
gcode:
  {% set macro_name = "M117" %}
  {% macro log(message, level='info') %}
  RESPOND PREFIX="{ "\"[%s%s%s]\"" % (macro_name,":" if level else "", level) }" MSG="{message}"
  {% endmacro %}
  {% set message = "".join(commandline[5:]) %}
  {log(message, "")}
```

In this case, the input to M117 is sent both to the display as usual, but also to the console.

Although the obvious solution would be to use `{" ".join(params.keys()}`, this doesn't actually work as expected - the dict is reordered, and numbers seem just to be ignored. ie: `M117 hello world 123` is sent to the template as `M HELLO`

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -178,6 +178,7 @@ class GCodeMacro:
         kwparams = dict(self.variables)
         kwparams.update(self.template.create_template_context())
         kwparams['params'] = gcmd.get_command_parameters()
+        kwparams['commandline'] = gcmd.get_commandline()
         self.in_script = True
         try:
             self.template.run_gcode_from_command(kwparams)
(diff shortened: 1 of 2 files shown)


zellneralex:
It looks like both of you try to solve the same problem https://github.com/Klipper3d/klipper/pull/4727

KevinOConnor:
FYI, PR #4727 was merged.  It's fine if we want to also export a `{commandline}` pseudo-variable, but if so, we'd need to update the documentation for that.

-Kevin

PR #4884, opened 2021-11-06
Multiple steppers per delta tower

TypQxQ:
Changes and sample configuration to make it possible to use multiple steppers on each Delta tower.
Instead of using a stepper motor and a idler, with this one can use two stepper motors, each controlled by it's own stepper.

I have tried this out using two SKR 1.4 Turbo cards with TMC2209.
Tower A and B steppers and endstop on first micro-controller and tower C and it's endstop on the second micro-controller.

This is how it looks like in action: https://youtu.be/53Ut1UPKveQ

Signed-off-by: Andrei Ignat <andrei@ignat.se>


Hywelmartin:
nice.. this would minimize the springiness of the belt, in a perfect world by half

config/sample-delta-multi-mcu-multi-steppers.cfg
```diff
@@ -0,0 +1,178 @@
...
+step_pin: P2.2
+dir_pin: !P2.6
+enable_pin: !P2.1
+microsteps: 16
+full_steps_per_rotation: 400
+rotation_distance: 40
+endstop_pin: P1.29
+homing_speed: 50
+arm_length: 271.50
+position_endstop = 291.646495
```
meteyou, line 58: pls change the `=`to a `:`to be consistent

angeljrgsr:
Are you getting any delays across controllers? I'm about to build a big delta with an Octopus so I'm watching this VERY closely!

matti04:
Hello, I saw your YouTube video about your Detla printer and I quite liked it, but could you elaborate why you decided to use two drivers instead of putting both motors on one driver (just like the Z-plug does, for example)? 

If you "couple" them wiring-wise, you would never have to worry about the two drivers being slightly out of sync and producing opposing forces. I don't know how this problem affects real-life use though. 

The only reason I could think of for individual drivers are the current, some drivers can only supply a certain current and two steppers might need too much. 

Thank you for your reply, Matteo

TypQxQ:
Hello! 
Good question. 
I am not an expert on stepper motors but here is how I understand it:

To run a motor of a single stepper you can connect them in parallel or series. 

Running them in parallel would run them at half the voltage each.
To achieve high torque at high speed you need a higher voltage than the generated back-EMF.
That's why I believe  the Duet3 and BTT Octopus Pro have high voltage drivers.

Running them in series, doubling the current for the motors to draw from, would give you more inductance and resistance, thus even more generated back-EMF. I've read that it gives you 4 times more inductance than running in parallel but don't quote me on that. So basically the back-EMF makes you loose power even sooner than in parallel. 

I think most boards use series for dual Z. This is fine for Z where you don't run high speeds and need high holding torque, not for directional changes and acceleration.

Running them each from it's own stepper gives you the maximum speed because of the lower back EMF and higher torque because of them working together on the same belt.

I've used the RRF EMF Calculator to try different combinations here: https://www.reprapfirmware.org/emf.html

Thank you, Andrei

KevinOConnor:
Thanks.  As high-level feedback, I don't see an issue with enabling support for multiple steppers on a delta tower.  However, I'm not sure there's a large enough audience to justify an example config file just for that case.  (I'm unsure the audience for this feature would hit 100 users, which is what we typically use as a benchmark for adding new documents - https://www.klipper3d.org/CONTRIBUTING.html ).

-Kevin

TypQxQ:
I have now deleted the example file. I thought that it might be useful as I have seen comments about it being hard to understand how to use multiple steppers with the current Cartesian config.

Thought it might be nice to have a config file using two advanced options like multiple MCU and multiple steppers per tower. Showing how easy it is may broaden the adoption of this setups.

PR #4885, opened 2021-11-06
stm32f4: resolve bootloader issues

### LINKED ISSUE #4838: Fysetc Spider flash-sdcard.sh script fails

Arksine:
This PR resets the peripheral clocks on init and performs a factory reset on the peripheral registers as they are enabled in `enable_pclock()`.  The implementation mimicks how it is handled with the STM32F1, with the FLITF and SRAM clocks enabled.

As bit 8 of APB2RSTR resets all 3 ADCs I thought it wise to avoid it.  My assumption is that it would cause an unwanted reset if multiple ADCs are enabled.

This should resolve #4838 and #4856.

### DIFF
--- a/src/stm32/stm32f4.c
+++ b/src/stm32/stm32f4.c
@@ -260,6 +260,12 @@ armcm_main(void)
     SystemInit();
     SCB->VTOR = (uint32_t)VectorTable;
 
+    // Reset peripheral clocks (for some bootloaders that don't)
+    RCC->AHB1ENR = 0x38000;
+    RCC->AHB2ENR = 0;
+    RCC->APB1ENR = 0;
+    RCC->APB2ENR = 0;
+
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #4895, opened 2021-11-08
Add link to OctoPi-Klipper pre-installed image

guysoft:
Add a link to the new pre-installed image, and then change the section that uses OctoPi to explain how t per-existing image can be edited.
I keep seeing people in the discord getting stuck installing OctoPi, then Klipper, this is a much better solution.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -11,7 +11,12 @@ micro-controllers,
 [ARM based micro-controllers](Features.md#step-benchmarks), and
 [Beaglebone PRU](beaglebone.md) based printers.
 
-## Prepping an OS image
+## Use pre-installed ready to use image
+You can download a pre-installed image with OctoPi and Klipper with the correct default settings.
+
+[OctoPi-Klipper can be downloaded here](https://github.com/guysoft/OctoPi-Klipper).
+
+## Prepping an existing OS image
 
 Start by installing [OctoPi](https://github.com/guysoft/OctoPi) on the
 Raspberry Pi computer. Use OctoPi v0.17.0 or later - see the


docs/Installation.md
```diff
@@ -11,7 +11,12 @@ micro-controllers,
 [ARM based micro-controllers](Features.md#step-benchmarks), and
 [Beaglebone PRU](beaglebone.md) based printers.
 
-## Prepping an OS image
+## Use pre-installed ready to use image
+You can download a pre-installed image with OctoPi and Klipper with the correct default settings.
+
+[OctoPi-Klipper can be downloaded here](https://github.com/guysoft/OctoPi-Klipper).
```
yzheka, line 17: Then maybe add [Maisail](https://docs.mainsail.xyz/) and [fluidd](https://docs.fluidd.xyz/)?

guysoft: Possible,  for now I wanted to fix the out-of-date octoprint instructions. Which is what this pr does


Once its accepted adding others makes sense too.

KevinOConnor:
Thanks for working on this, and sorry for the delay in responding.

As high-level feedback, I'm reluctant to make a significant change to the main install instructions until we're ready to make a change that we're comfortable with long term.

I recognize there are several issues with the current installation instructions - we should convert to python3, we should launch moonraker by default, we should avoid using files in the /tmp directory, we should adopt a better location for the printer.cfg file, etc.  I'm also aware there are several other installation methods for Klipper that are very popular (eg, mainsailos, fluiddpi, kiauh, etc.).

So, I agree we need a change here, but I don't want to make many "incremental changes" that could lead to confusion, or could lead some users into reinstalling many times in a short period.  My current plan is to keep the current recommended installation instructions largely as they are until we're ready to have a new set of recommended installation instructions.

-Kevin

guysoft:
@KevinOConnor gihub bot reminded me.
I forgot to thank you for posting a detailed explanation of what you are planning for the future of Klipper. Much appreciated!

Also its worth mentioning that GIna and I added OctoPi-Klipper to raspsberrypi official rpi-imager OctoPi section. So that should make it a little more accessible in the meantime.

KevinOConnor:
Okay, thanks.  Do you have a link?

-Kevin

guysoft:
Its part of the official Raspberry Pi imaging tool, here is a screenshot from the selection.
Its under "under spesific purpose OS > OctoPi > OctoPi-Klipper"
Screenshot_20220201_140221

The imager is available in the main tab of Raspsberry Pi website:
https://www.raspberrypi.com/software/

PR #4903, opened 2021-11-09
spi_flash: Better fallback on MCU protocol change

jschuh:
Support a set of known past config responses. This is just a more graceful follow-up to 76bfc32, which will support forward and backward MCU compatibility (I manually tested rolling forward and rolling back).

@Arksine, I'd really appreciate your thoughts on this.

### DIFF
--- a/scripts/spi_flash/spi_flash.py
+++ b/scripts/spi_flash/spi_flash.py
@@ -99,7 +99,10 @@ SD_SPI_SPEED = 4000000
 # MCU Command Constants
 RESET_CMD = "reset"
 GET_CFG_CMD = "get_config"
-GET_CFG_RESPONSE = "config is_config=%c crc=%u is_shutdown=%c move_count=%hu"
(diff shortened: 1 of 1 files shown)


Arksine:
I think that this is probably the best short term approach to handling this issue.  I was trying to brainstorm a way of dealing with this should any command change, but I'm drawing a blank.  It seems unlikely that the SPI Commands will need to change anyway.

KevinOConnor:
I goofed on this.  I wanted to force a protocol mismatch for https://klipper.discourse.group/t/flashing-of-micro-controller-code-needed/1263 - I should have looked closer to see if I could do that without causing an error in spi_flash.

I agree best is probably to do a short-term workaround for now.

-Kevin

jschuh:
Yeah, I didn't have any better ideas. I do have some thoughts on how to add a test to catch breakage in the future, so I'll try to look into that.

PR #4904, opened 2021-11-09
Fixed the example for the two stepper expansion board

christophmuellerorg:
These pins need to be inverted, identically to the on board drivers.

### DIFF
--- a/config/generic-duet2-maestro.cfg
+++ b/config/generic-duet2-maestro.cfg
@@ -105,8 +105,8 @@ stealthchop_threshold: 999999
 #...
 
 # External steppers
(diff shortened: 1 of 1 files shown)


dmbutyugin (commented):
I do not own this board, but the change looks reasonable. Please note that all commits in the contributions should follow [this format](https://www.klipper3d.org/CONTRIBUTING.html#submitting-a-pull-request). Specifically, each commit should have 'Signed-off-by' line with a real name. And ideally the commit should start with a line of the following format "module: Capitalized, short (50 chars or less) summary". E.g. "config: Fixed Maestro2 example for the two stepper expansion board".

PR #4795, opened 2021-10-11
Round reported temperatures to 2 decimal places

Arksine:
This PR rounds temperatures reported by `get_status()` for the various sensors, as the additional precision is unnecessary.

Signed-off-by:  Eric Callahan <arksine.code@gmail.com>

PR #4909, opened 2021-11-10
virtual_sdcard: round reported progress to 2 decimal places

pedrolamas:
This PR rounds progress reported by get_status() for virtual_sdcard (which in turn is used by display_status), as the additional precision is unnecessary.

This follows up on #4795


KevinOConnor:
Thanks.  However, I think I'm missing the "big picture" - why do we want to round this value?  I could see this being an issue for API clients as it could reduce the precision of other file print time estimations.

-Kevin

pedrolamas:
Fair point... This made sense at the time as to avoid MQTT flooding but that has since been addressed in Moonraker so there is no need for this anymore.

PR #4911, opened 2021-11-11
change rotation distance at run time

ludovico26:
my modification for a personal prototype that has z axis modifyng rotation distance at run time

### DIFF
--- a/klippy/extras/modify_rotation.py
+++ b/klippy/extras/modify_rotation.py
@@ -8,7 +8,6 @@ class RotationDistanceModifier:
 		gcode.register_command("SET_STEPPER_ROTATION_DISTANCE", "STEPPER",
                                    self.name, self.cmd_SET_STEPPER_ROTATION_DISTANCE,
                                    desc=self.cmd_SET_STEPPER_DISTANCE_help)
-	   
 	cmd_SET_STEPPER_ROTATION_DISTANCE_help = "Change rot dist of individual stepper by name"
 	def cmd_SET_STEPPER_ROTATION_DISTANCE(self, gcmd):
 		toolhead = self.printer.lookup_object('toolhead')
--- a/klippy/extras/modify_rotation.py
+++ b/klippy/extras/modify_rotation.py
@@ -1,48 +1,37 @@
 
 import logging
-
 class RotationDistanceModifier:
-    def __init__(self, config):
-        self.printer = config.get_printer()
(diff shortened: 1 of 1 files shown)


netweaver1970:
maybe not the best place to ask but why/when would this functionality be used?

ludovico26:
it should be used only for the particular 3D printer we are trying to create that should "grow on its own profile".
Therefore we should be able to modify the steper rotation distance in order to permit it to go from a possible linear to an arc movement.

I know it is of no interest for normal klipper users and that it may destroy a normal printer in a standard operation but i didn't know how to make this change otherwise....

If you have any suggestions plese tell

netweaver1970:
he, I just got my V2 over the summer so who am I to suggest any better :) So this is to live compensate for the thermal expansion  while 'growing', instead of waiting for a thermal equilibrium and doing calculations based on enclosure/frame temp ?

ludovico26:
As it prints with PLA for the time being, it does not strictly require any enclosure or printed bed.
Moreover it may take some time before reching the newly extruded layer.
The reason for this was to move the 4 z axis motors differently, so to perform an arc or a linear trajectory in space and then to postprocess non planar slic3r to generate the g code needed to "print an arc".
It should be used for building purposes

netweaver1970:
ok. So the temp compensation I was blabbling about was just totally a red herring :) So this ultimately would allow easier non-planar printing by doing (live) bed (all 4 corners) non-linear adjustments, steered by the g-code postprocessing logic. Interesting to see it evolve. You have a link somewhere to follow it?

ludovico26:
It is a project made by the university of Bologna (Unibo) and i don't know when and how it may end or you cna find informations about it.
My goal was to configure all the electronics and to find the guidelines to properly calibrate the extruder, and to create the appropriate software.
As for the mechanics they need to recreate it so it will surely take more than an year time....

netweaver1970:
Ah ok, pity it's not public, not even some intermediate results.

good luck!

zellneralex:
Sorry to ask but why you want it then in main branch? A fork would be a much better place for your changes.

ludovico26:
When I forked Klipper in my github and i added the file it directly sent me here.
I am new to github so i may need som help about it

Creating a fork may be better and, if I don't have to wait for a maintainer it would be higly benefical.
Could you please tell me how to? 

Thanks in advance for all your help

PR #4917, opened 2021-11-12
temperature_sensors: add ATC Semitec 104NT-4-R025H42G thermistor

geoffrey-young:
module: extras/temperature_sensors.cfg

this thermistor is the 104NT-4 resold by Trianglelab (and others on Ali),
as well as the 300 degree resold by Slice Engineering, both of which use the
same RT table (TL even links to the ATC Semitec website)

see the 300 C tab on the Slice spreadsheet:
  https://docs.google.com/spreadsheets/d/1904x5JK-Sup-cX5DqHiiZWaFVTK6_PQBFxgi_6yXEJw/edit#gid=934228925
as well as the TL product page:
  https://www.aliexpress.com/item/32843785247.html
in addition to the RT table link provided in temperature_sensors.cfg

Signed-off-by: Geoffrey Young <geoffrey.young@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -701,10 +701,11 @@ heater_pin:
 #   periods) to the heater. The default is 1.0.
 sensor_type:
 #   Type of sensor - common thermistors are "EPCOS 100K B57560G104F",
-#   "ATC Semitec 104GT-2", "Generic 3950", "Honeywell 100K
-#   135-104LAG-J01", "NTC 100K MGB18-104F39050L32", "SliceEngineering
-#   450", and "TDK NTCG104LH104JT1". See the "Temperature sensors"
-#   section for other sensors. This parameter must be provided.
+#   "ATC Semitec 104GT-2", "ATC Semitec 104NT-4-R025H42G", "Generic
+#   3950","Honeywell 100K 135-104LAG-J01", "NTC 100K MGB18-104F39050L32",
+#   "SliceEngineering 450", and "TDK NTCG104LH104JT1". See the
+#   "Temperature sensors" section for other sensors. This parameter
+#   must be provided.
 sensor_pin:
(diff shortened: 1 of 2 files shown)


dmbutyugin:
Yes, indeed, it seems this sensor is similar to "ATC Semitec 104GT-2", but still a bit different. Though Trianglelab recommends to use 'ATC Semitec 104GT-2' as a sensor type in the absence of a better fit (e.g. in Marlin). But it probably makes sense to merge its definition to Klipper?

I've previously found this sensor in the table
http://www.semitec.co.jp/uploads/english/sites/2/2017/03/P18-NT-Thermistor.pdf
which matches your data source https://atcsemitec.co.uk/wp-content/uploads/2019/01/Semitec-NT-4-Glass-NTC-Thermistor.pdf and the values you specified in the sensor definition, so I can confirm that the values seem to be correct.

BTW, all commits in the PR need to have a Signed-off-by line: https://www.klipper3d.org/CONTRIBUTING.html#submitting-a-pull-request. And the title of the commit should ideally have a modified module name first (see the same link).

geoffrey-young:
thanks for the review, and for the contributing doc pointer - much appreciated.

I made the suggested updates, using what seems like the logical module name for the change.

dmbutyugin:
Sorry, AFAIU you actually need to update (e.g. via `git commit --amend`) the actual git commit message.

geoffrey-young:
yeah, that makes sense :)

I think it should be ok now...

dmbutyugin (approved):
Thanks! It looks good now, in my opinion.

dmbutyugin:
@KevinOConnor, I think the PR could be merged, unless you have reservations about having too many temperature sensors defined in Klipper.

KevinOConnor:
Thanks.

-Kevin

PR #4925, opened 2021-11-14
webhooks: create a deep copy for status queries

Arksine:
This resolves an issue where subscriptions will not update because the `last_query` contains a reference to an object returned by `get_status()`.  Currently this is known to affect `gcode_macro` variables.

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -171,7 +171,9 @@ class GCodeMacro:
             literal = ast.literal_eval(value)
         except ValueError as e:
             raise gcmd.error("Unable to parse '%s' as a literal" % (value,))
-        self.variables[variable] = literal
+        v = dict(self.variables)
+        v[variable] = literal
+        self.variables = v
     def cmd(self, gcmd):
         if self.in_script:
             raise gcmd.error("Macro %s called recursively" % (self.alias,))


KevinOConnor:
Thanks.  I agree this would be the "right" solution in general, however I'm concerned about the CPU overhead from doing this.

It seems a lot of API clients just subscribe to every possible Klipper variable at startup.  (Indeed, the "motan" client I wrote even does this.)  However, this means that Python has to compare every status dictionary every 250ms.  (And traverse every sub-item in every dictionary.)  This is actually very fast if the dictionaries are the same reference as Python will know the two dictionaries are the same if the reference is identical.  In particular, the `configfile` dictionaries are huge and never change at runtime.

So, my current approach has been to try to keep the `get_status()` methods aware of this quirk and to replace any dictionaries on a change.  This should probably be documented in Code_Overview.md .

I suspect the following would fix the issue in gcode_macros.py :
```diff
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -171,7 +171,9 @@ class GCodeMacro:
             literal = ast.literal_eval(value)
         except ValueError as e:
             raise gcmd.error("Unable to parse '%s' as a literal" % (value,))
-        self.variables[variable] = literal
+        v = dict(self.variables)
+        v[variable] = literal
+        self.variables = v
     def cmd(self, gcmd):
         if self.in_script:
             raise gcmd.error("Macro %s called recursively" % (self.alias,))
```

-Kevin

KevinOConnor:
FYI, I added a note to the docs in commit 779793c7.

-Kevin

Arksine:
Ah, makes sense.  I tested your suggestion and it indeed resolves the issue.  I went ahead and force pushed it, however if you would prefer to commit directly I can close this PR

KevinOConnor:
Thanks.

-Kevin

PR #4928, opened 2021-11-16
print_stats: Fix exception when canceling in/as first G-code

twelho:
When executing `CANCEL_PRINT` (directly or as part of a macro) as the first G-code instruction Klipper crashes with the following:
```
  File "/home/pi/klipper/klippy/klippy.py", line 201, in run
    self.reactor.run()
  File "/home/pi/klipper/klippy/reactor.py", line 269, in run
    g_next.switch()
  File "/home/pi/klipper/klippy/reactor.py", line 310, in _dispatch_loop
    timeout = self._check_timers(eventtime, busy)
  File "/home/pi/klipper/klippy/reactor.py", line 156, in _check_timers
    t.waketime = waketime = t.callback(eventtime)
  File "/home/pi/klipper/klippy/extras/virtual_sdcard.py", line 291, in work_handler
    self.print_stats.note_complete()
  File "/home/pi/klipper/klippy/extras/print_stats.py", line 45, in note_complete
    self._note_finish("complete")
  File "/home/pi/klipper/klippy/extras/print_stats.py", line 54, in _note_finish
    self.total_duration = eventtime - self.print_start_time
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
```

It seems that while a `None` check is present around other uses of `print_start_time`, one was missing in `_note_finish`. By adding this check the canceling now works correctly, and the printer no longer enters emergency mode. I also verified that `total_duration` correctly shows a (very small) non-zero value.

### Reproducing the issue

Print a file with only one `CANCEL_PRINT` instruction:

`cancel.gcode`:
```
CANCEL_PRINT
```

### DIFF
--- a/klippy/extras/print_stats.py
+++ b/klippy/extras/print_stats.py
@@ -48,6 +48,8 @@ class PrintStats:
     def note_cancel(self):
         self._note_finish("cancelled")
     def _note_finish(self, state, error_message = ""):
+        if self.print_start_time is None:
+            return
         self.state = state
         self.error_message = error_message
         eventtime = self.reactor.monotonic()


Arksine:
It would appear `note_cancel()` is called before `note_start()` is called.  I'm unsure as to how this is possible given that the virtual sdcard's `work_handler()` method calls `note_start()` before it reads the gcode file.

Its probably a good idea to add the NoneType check in `_note_finish` in the event that it does get called on a print that hasn't started, however I do think it would be useful to better understand why `self.print_start_time` is set to `None` in this case, as it shouldn't be.

Arksine:
Ok, I see what is happening.  The error is occurring because `note_complete()` is being called after `note_cancel()`.  I think we need to avoid that processing a `_note_finish()` multiple times. so I would suggest the following fix:
```python
 def _note_finish(self, state, error_message = ""):
    if self.print_start_time is None:
       return
    self.state = state
    self.error_message = error_message
    self.total_duration = eventtime - self.print_start_time
    if self.filament_used < 0.0000001:
        # No positive extusion detected during print
        self.init_duration = self.total_duration - \
            self.prev_pause_duration
    self.print_start_time = None
```

This also catches any attempt to call `note_finish()` before a `note_start()`.

twelho:
Yep, I do follow your logic and see where the issue is coming from now as well. Having the first complete/error/cancel signal determine the end state makes sense since it is the original cause of interruption. Any signals to finish a no-longer-ongoing print should be safe to ignore. I've refactored the changes to match your patch now @Arksine, looks clean to me.

Arksine:
Ok, the patch looks good to me.  You just need to sign off on the commit as noted in the [contributing document](https://www.klipper3d.org/CONTRIBUTING.html#submitting-a-pull-request).  I should have noticed that immediately to give you a heads up, sorry about that.

twelho:
TIL that GPG-signing a commit doesn't automatically add the signed-off-by line. I've added it now.

KevinOConnor:
Thanks.

-Kevin

PR #4934, opened 2021-11-17
Add FYSETC Cheetah v2.0 config

GerogeFu:
1. Add new generic config for FYSETC new released board: FYSETC Cheetah v2.0.
2. Fix Spider sdcard upload fail bug refer to: https://github.com/FYSETC/FYSETC-SPIDER/issues/38

### DIFF
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -52,7 +52,8 @@ BOARD_DEFS = {
     },
     'fysetc-spider': {
         'mcu': "stm32f446xx",
-        'spi_bus': "spi1",
+        'spi_bus': "swspi",
+        'spi_pins': "PA6,PA7,PA5",
         "cs_pin": "PA4",
         "current_firmware_path": "OLD.BIN"
     }
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -52,8 +52,7 @@ BOARD_DEFS = {
     },
     'fysetc-spider': {
         'mcu': "stm32f446xx",
-        'spi_bus': "swspi",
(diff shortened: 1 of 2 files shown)


scripts/spi_flash/board_defs.py
```diff
@@ -52,7 +52,8 @@
     },
     'fysetc-spider': {
         'mcu': "stm32f446xx",
-        'spi_bus': "spi1",
+        'spi_bus': "swspi",
```
adelyser, line 55: This shouldn't be changed, it should remain at spi1.

adelyser:
1. The config looks ok to me.
2. You will need to revert the board_defs.py changes and add an alias in there for the cheetah-v2.0 board.
3. Please make sure to follow https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md, especially the signed-of statement

webhdx:
@GerogeFu Any chance to move it forward?

KevinOConnor:
I'm not sure what the state of this PR is.  In general the config looks okay to me, however the EXP1_x and EXP2_x aliases need to use the pin ordering described in the sample_lcd.cfg file (eg, `EXP1_9=<GND>`).  Also, the config needs to be added to test/klippy/printers.test (as described in https://www.klipper3d.org/Example_Configs.html ).

-Kevin

BrienAllison:
I have tested this config on the Cheetah V2.0 board and verified the pin definitions except the endstop pins required inverting. Also It would also be nice to know the diag_pin definitions for the TMC2209 drivers on this board to allow sensorless homing. Not sure how else to get that information other than guessing. Otherwise I will just go ahead and install standard endstops.

Thanks.

PR #4935, opened 2021-11-17
slic3r_vars: Enable slic3r parameters to be read from gcode

whi-tw:
Slic3r (and prusaslicer / SuperSlicer) can output all their parameters into the gcode file it creates.

This allows these to be read within macros while printing:

`{printer.slic3r_vars.extruder_offset}` == `"0x0,20x0"` for example.


pedrolamas:
This is currently done on Moonraker side (it parses the GCode to retrieve certain features, such as the startup hotend and bed temperatures).

Having said that, I understand why you would want this on Klipper side for macros, but I don't think this is actually necessary for that purpose, as you can do this manually on any slicer just by adding parameters to your `START_PRINT` custom macro (just as you probably currently do for the startup temperatures!)

whi-tw:
For sure, fully understand that it's possible to do in other ways, but I find the start gcode really difficult to read and keep up to date.

You end up writing a load of supporting code in the macros, which just read these parameters into variables anyway - it's not even as if klipper automatically pulls certain settings in, you still have to do some kind of processing.

Now I've started using my fork, my start gcode in the slicer is just `START_PRINT`, and I do the processing of the config from within the macro. If I want to make any changes to my macros, I just read in the new variables, rather than having to find out what SuperSlicer refers to them as, and passing them in as extra parameters (which then have to be propagated throughout any sub-macros)

My macros are mirrored here: https://github.com/whi-tw/klipper-macros, and my `START_PRINT` is [here](https://github.com/whi-tw/klipper-macros/blob/98a7fd9b80259bb472de80563257262caa66f027/START_PRINT.cfg#L29-L52).

KevinOConnor:
Thanks.  It seems interesting, but as a high-level comment, I'm not really sure what this does or the benefit it provides to users.  I'm going to mark this as needing a reviewer that can weigh in on the benefit.

-Kevin

PR #4942, opened 2021-11-19
add flash details for STM32F429

JamesH1978:
Updates for generic-bigtreetech-skr-2.cfg to include comment on newly shipped board with STM32F429 instead of the original STM32F407

Signed-off-by: James Hartley james@hartleyns.com


JamesH1978 (commented):
Updates for generic-bigtreetech-skr-2.cfg to include comment on newly shipped board with STM32F429 instead of the original STM32F407

Signed-off-by: James Hartley james@hartleyns.com

KevinOConnor:
Thanks.

-Kevin

PR #4943, opened 2021-11-19
atsam: Fix I2C bitrate and minor fixes for UART and USB

eamaclean:
The UART and USB adjustments should have no functional impact but are cleanup in preperation for a port for the ATSAM E70 to support the Duet 3 6HC.

While checking on timing details for the E70 port I discovered an error in the I2C clock period calculation which results in twice the desired bitrate.


KevinOConnor:
Thanks.

-Kevin

PR #4948, opened 2021-11-20
Add STM32H743 Processor Configuration

adelyser:
STM32H743

Add the STM32H743 Processor configuration for use with the BTT-SKR-SE-BX board (Special board for the BIQU BX printer as of right now).

NOTE: USB and SPI PR's are not ready quite yet. so those features will be enabled once they are confirmed ready and have been tested.


KevinOConnor:
Thanks.  I broke this into two commits 7b69ef07 and b480734c.  I also made a few minor changes to the Kconfig declarations.

-Kevin

KevinOConnor:
I was looking at the stm32h7 spec and it looks like the "systick external clock reference" on the stm32h7 is the peripheral clock divided by 8.  (The systick external clock is used by not setting `SysTick_CTRL_CLKSOURCE_Msk` in `SysTick->CTRL`.)

So, if that feature is enable, `CONFIG_CLOCK_FREQ` is set to 15Mhz, FREQ_PERIPH is set to 120Mhz, and the cpu clock is set 480Mhz, then I think it might be a relatively simple way to enable the faster cpu clock.

-Kevin

KevinOConnor:
Or, maybe it's actually the system clock divided by 8 - so 60Mhz and 480Mhz..  It's hard to tell from the spec.

-Kevin

KevinOConnor:
Thinking about this further, the systick setting wouldn't alter the DWT cycle counter, so I guess that wouldn't work.  Sorry for the noise.

-Kevin

PR #4949, opened 2021-11-21
Minor improvements on documentation

dingyifei:
Renamed a few documents to follow the naming convention and fixed some minor issues


KevinOConnor:
Thanks.  I merged the first 6 of 7 commits.  I'm unsure about the 7th commit, as the Bootloaders.md document is a technical document aimed at developers - I think it may be preferable to explicitly list the URLs instead of trying to give names to them.

-Kevin

dingyifei:
> I think it may be preferable to explicitly list the URLs instead of trying to give names to them.

Agree, I'm going to amend this commit so only the out-of-date links are updated.  

Additionally, it seems like I don't have write permission for pushing into the klipper-translations workflow files. I can only pull from upstream (klipper) from my branch and pr it into Klipper-translations. It changed the committer of the commits and I think they will eventually break something.... Is there a workaround?

KevinOConnor:
> it seems like I don't have write permission for pushing into the klipper-translations workflow files.

I'm not sure why that would be.  Nothing has changed that I'm aware of (at least since we moved the repo to klipper3d/klipper-translations).  I changed your access on klipper-translations from "write" access to "maintain" access.  Maybe that will help?  Otherwise, maybe you could post the error message you're getting to see if that helps track the issue down?

-Kevin

dingyifei:
> it seems like I don't have write permission for pushing into the klipper-translations workflow files.

My fault, turns out to be my account setting issues, all of my personal access tokens on my new laptop lack a "workflow" permission, and GitHub Desktop behave differently on Linux.

KevinOConnor:
FYI, I get the impression this branch was changed, but github doesn't seem to have recognized that.  I guess the remaining changes will need a rebase onto the latest master branch.

-Kevin

dingyifei:
It seems like different committer results in different SHA, fetch upstream seems to do funny things... I think you can cherry-pick 43cccb9 from my branch into the master branch by adding my fork as another remote?
   
image

dingyifei:
@KevinOConnor do I need to do anything for my PR or I should wait for the new process?

KevinOConnor:
GitHub was doing something weird before, but it seems to have cleared.  So, I committed this change.  Thanks.

-Kevin

PR #4950, opened 2021-11-21
BIQU BX printer config

adelyser:
BIQU BX STM32H743

Add the BIQU BX printer config.

### DIFF
--- a/config/printer-biqu-bx-2021.cfg
+++ b/config/printer-biqu-bx-2021.cfg
@@ -1,0 +1,356 @@
+# This file contains common pin mappings for the BigTreeTech SKR SE BX.
+# To use this config, the firmware should be compiled for the
+# STM32H743 with a "128KiB bootloader". Additionally, GPIO pins PB5
+# and PE5 need to be set at microcontroller startup.
+
+# The "make flash" command does not work on the SKR SE BX. Instead,
+# after running "make", copy the generated "out/klipper.bin" file to a
+# file named "firmware.bin" on an SD card and then restart the SKR SE BX
+# with that SD card. After klipper has been flashed once to the board,
+# you can update klipper by leaving a microSD inserted and running the
+# scripts/flash-sd.sh script.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+########################################
+# Steppers
+########################################
+
+[stepper_x]
+step_pin: PG13
+dir_pin: !PG12
+enable_pin: !PG14
+microsteps: 16
+rotation_distance: 40
+full_steps_per_rotation: 400
+endstop_pin: tmc2209_stepper_x:virtual_endstop
+position_endstop: -13
+position_min: -13
+position_max: 250
+homing_speed: 30
+homing_retract_dist: 0
+
+[stepper_y]
+step_pin: PB3
+dir_pin: !PD3
+enable_pin: !PB4
+microsteps: 16
+rotation_distance: 40
+full_steps_per_rotation: 400
+endstop_pin: tmc2209_stepper_y:virtual_endstop
+position_endstop: -7
+position_min: -7
+position_max: 250
+homing_speed: 30
+homing_retract_dist: 0
+
+[stepper_z]
+step_pin: PD7
+dir_pin: PD6
+enable_pin: !PG9
+microsteps: 16
+rotation_distance: 8
+full_steps_per_rotation: 400
+endstop_pin: probe:z_virtual_endstop
+position_min: -2
+position_max: 250
+
+[stepper_z1]
+step_pin: PA8
+dir_pin: PC9
+enable_pin: !PD2
+microsteps: 16
+rotation_distance: 8
+full_steps_per_rotation: 400
+
+[extruder]
+step_pin: PC14
+dir_pin: !PC13
+enable_pin: !PC15
+microsteps: 16
+rotation_distance: 3.433
+full_steps_per_rotation: 200
+nozzle_diameter: 0.400
+filament_diameter: 1.750
+heater_pin: PC4
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PH4
+control: pid
+pid_Kp: 22.2
+pid_Ki: 1.08
+pid_Kd: 114
+min_temp: 0
+max_temp: 350
+pressure_advance: 0.123
+
+[safe_z_home]
+home_xy_position: 125,125
+speed: 200
+z_hop: 10
+z_hop_speed: 25
+
+########################################
+# TMC2209 configuration
+########################################
+
+[tmc2209 stepper_x]
+uart_pin: PG10
+diag_pin: ^PB11
+run_current: 0.800
+hold_current: 0.525
+sense_resistor: 0.150
+driver_SGTHRS: 127
+stealthchop_threshold: 999999
+
+[tmc2209 stepper_y]
+uart_pin: PD4
+diag_pin: ^PB12
+run_current: 0.800
+hold_current: 0.575
+sense_resistor: 0.150
+driver_SGTHRS: 137
+stealthchop_threshold: 999999
+
+[tmc2209 stepper_z]
+uart_pin: PD5
+run_current: 0.800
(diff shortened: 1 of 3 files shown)


config/printer-biqu-bx-2021.cfg
```diff
@@ -0,0 +1,390 @@
...
+
+########################################
+# Steppers
+########################################
+
+[stepper_x]
+step_pin: PG13
+dir_pin: !PG12
+enable_pin: !PG14
+microsteps: 16
```
Sineos, line 29: You do not use interpolate, which is good acc. to https://github.com/Klipper3d/klipper/pull/4977
Why this low `microsteps`?

config/printer-biqu-bx-2021.cfg
```diff
@@ -0,0 +1,390 @@
...
+
+########################################
+# TMC2209 configuration
+########################################
+
+[tmc2209 stepper_x]
+uart_pin: PG10
+diag_pin: ^PB11
+run_current: 0.800
+hold_current: 0.525
```
Sineos, line 106: You might want to refer to https://github.com/Klipper3d/klipper/pull/4977
Same for the rest of the `hold_current` references

config/printer-biqu-bx-2021.cfg
```diff
@@ -0,0 +1,390 @@
...
+[tmc2209 stepper_z1]
+uart_pin: PC8
+run_current: 0.800
+hold_current: 0.525
+sense_resistor: 0.150
+stealthchop_threshold: 999999
+
+[tmc2209 extruder]
+uart_pin: PI8
+interpolate: false
```
Sineos, line 136: Any reason to explicitly mention `interpolate` here?

config/printer-biqu-bx-2021.cfg
```diff
@@ -0,0 +1,390 @@
...
+    G0 Z{Z}
+    G90
+    M83
+    G0 X{X} Y{Y} F3000
+
+[gcode_macro RESUME]
+rename_existing: RESUME_BASE
+gcode:
+    RESTORE_GCODE_STATE NAME=PAUSE_state
+    RESUME_BASE
```
Sineos, line 390: Macros should only be defined if they are needed to support specific printer functions (https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md 5.iv)

config/printer-biqu-bx-2021.cfg
adelyser: This actually does use interpolate, the default is True.

config/printer-biqu-bx-2021.cfg
adelyser: I removed the hold_current

config/printer-biqu-bx-2021.cfg
adelyser: This disabled interpolate, as the default is true

config/printer-biqu-bx-2021.cfg
adelyser: The PRINT_START macro is definitely needed to support this printer. This is due to how the ABL is setup and the macro implemented "mimmics" the printers current Marlin behaviour for consistent probing of the bed.

I also setup all the other macros based on user feedback on the stock printer. Should I remove those anyway?

Thanks for the feedback!

config/printer-biqu-bx-2021.cfg
Sineos: Acc to #4977 it seems preferable to use higher `microsteps` and set `interpolate: false`
The BIQU BX SKR SE-BX should not even break a sweat with this

config/printer-biqu-bx-2021.cfg
Sineos: I don't know this printer, so I guess this is fine. Final call is on Kevin anyway

config/printer-biqu-bx-2021.cfg
adelyser: While the BIQU BX definitely doesn't have a problem with this config, testing shows that the printer is crazy loud using the spreadcycle mode. For default, I decided to leave it as stealthchop because most users will find it annoying and want to quiet it down anyways.

KevinOConnor:
Some feedback:
* The example config should not specify a pressure_advance setting.
* Specifying "interpolate=false" doesn't seem to make sense.
* I don't understand the comment at the top of the file regarding the display requiring the mcu to set the initial pins.  We typically only use the mcu initial pins setting to set pins that must be set in order to communicate with the host.  If there is some other necessary setting, it really should be in the printer.cfg file and not in "make menuconfig".
* The example macros look complex to me.  It's generally preferable to keep the example configs similar to ease long term maintenance.  I think the goal of the example configs should be to get an initial starting point for users - not to customize the printer for them.

-Kevin

KevinOConnor:
Also, the extruder rotation_distance looks very unusual.  It's generally preferable to specify a gear_ratio for extruders using a gearing mechanism.

-Kevin

adelyser:
@KevinOConnor Thanks for the feedback.

- I will remove the pressure advance setting, you are correct that shouldn't be in there.
- I will remove the interpolate setting
- So the GPIO pins need to be set so they are active when the printer boots, not when Klipper connects. The TFT Screen, specific to the BIQU BX printer, has an HDMI switcher and the backlight controlled from the mainboard LTDC driver in Marlin. These GPIOS are required to be set, or you will not see anything on the screen during bootup, or until Klipper connects.
- The only macro really needed is the PRINT_START macro. This printer comes with an ABL probe and there's a specific sequence needed to get it to work properly, which PRINT_START handles. I can remove all the other macros, or trim them out to be the basic macros needed.
-The weird thing about the rotational distance, I tried using the gear ratio method and my extrusion values were noticeably off. This was what I came up with to get a valid extrusion value that worked. I don't know what was going on with the gear ratio, but it was off.

adelyser:
@KevinOConnor Thanks for the response.

> I understand the gpios need to be set. I don't understand why they need to be set in "make menuconfig" instead of in "output_pin" settings in printer.cfg?

So these need to be set in the firmware because you want the display, while the RPI is booting, when the printer is turned on. The RPI is attached to the backside of the TFT and powered with the printer power. If these are put into the printer.cfg file, you get no display until klipper connects. If there is an error with klipper connecting, still no display, which is is a bad experience. It acts as if the printer isn't doing anything but running the fans. 

> You should be able to count the number of teeth on the extruder gearing mechanism and then apply that to your existing rotation_distance. So, for example, if it is 50:17 gear ratio, then you would replace rotation_distance: 3.433 with gear_ration: 50:17 and rotation_distance: 10.097 (3.433 * 50 / 17 == 10.097).

I totally understand what you're saying and I fully understand the calculations. At first, I did use a gear ratio of 7:1, with a rotation_distance of 24.03 (correct values for the H extruder, 932 esteps provided by BIQU). The extruder was over-extruding while printing. Changing that to 3.433, fixed the extrusion issue. I toggled it back and forth a couple times to confirm, and I don't know why it wasn't working correctly.

KevinOConnor:
> So these need to be set in the firmware because you want the display, while the RPI is booting, when the printer is turned on. 

This isn't what the comment in the example config file says.  If this is a "nice to have" then the comment should say that - for example, "Note: In order to update the TFT70-BX display when the micro-controller first starts, add PB5 and PE5 ...".

> Changing that to 3.433, fixed the extrusion issue.

That indicates something is very wrong and should be investigated.  The only thing Klipper does with the rotation_distance is multiply it with full_steps_per_rotation and gear_ratio to get an internal step_distance.  Updating the config with `gear_ratio: 7:1` and `rotation_distance: 24.031` results in an identical internal state.

-Kevin

KevinOConnor:
Thanks.

-Kevin

nicedevil007:
@adelyser I have a problem with your default config and hope you can help ? https://github.com/Klipper3d/klipper/issues/5236

small side question: did you figure out how to get the whole display to show the klipperscreen or are you using anything else?

PR #4953, opened 2021-11-22
screws_tilt_adjust: Add ADJUST_TO_Z parameter to SCREWS_TILT_CALCULATE

celtare21:
I found this quite useful because my bed level drifts away with just time. Having to always go back to the last good position from mind was a bit of a pain so I just introduced ADJUST_TO_Z to always level for a given base if wanted.

Output:
> $ SCREWS_TILT_CALCULATE ADJUST_TO_Z=1.5
> // probe at 32.000,219.000 is z=1.493227
> // probe at 32.000,219.000 is z=1.493422
> // probe at 32.000,219.000 is z=1.493539
> // probe at 32.000,219.000 is z=1.489906
>
> // probe at 200.000,219.000 is z=1.501625
> // probe at 200.000,219.000 is z=1.504672
> // probe at 200.000,219.000 is z=1.500570
> // probe at 200.000,219.000 is z=1.500805
>
> // probe at 32.000,63.000 is z=1.504984
> // probe at 32.000,63.000 is z=1.510414
> // probe at 32.000,63.000 is z=1.510219
> // probe at 32.000,63.000 is z=1.511859
>
> // probe at 200.000,63.000 is z=1.486742
> // probe at 200.000,63.000 is z=1.490258
> // probe at 200.000,63.000 is z=1.485648
> // probe at 200.000,63.000 is z=1.488852
>
> // 01:20 means 1 full turn and 20 minutes, CW=clockwise, CCW=counter-clockwise
>
> // front left screw : x=32.0, y=219.0, z=1.49332 : adjust CW 00:01

> // front right screw : x=200.0, y=219.0, z=1.50121 : adjust CCW 00:00
>
> // rear left screw : x=32.0, y=63.0, z=1.51032 : adjust CCW 00:01
>
> // rear right screw : x=200.0, y=63.0, z=1.48780 : adjust CW 00:01

Signed-off-by: Vlad Adumitroaie <celtare21@gmail.com>

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -485,14 +485,17 @@ The following commands are available when the
 [screws_tilt_adjust config section](Config_Reference.md#screws_tilt_adjust)
 is enabled (also see the
 [manual level guide](Manual_Level.md#adjusting-bed-leveling-screws-using-the-bed-probe)):
-- `SCREWS_TILT_CALCULATE [DIRECTION=CW|CCW] [<probe_parameter>=<value>]`:
+- `SCREWS_TILT_CALCULATE [ADJUST_TO_Z=<value>] [DIRECTION=CW|CCW] [<probe_parameter>=<value>]`:
   This command will invoke the bed screws adjustment tool. It will command the
   nozzle to different locations (as defined in the config file)
   probing the z height and calculate the number of knob turns to
-  adjust the bed level. If DIRECTION is specified, the knob turns will all
+  adjust the bed level. If ADJUST_TO_Z is specified, the z position
+  to raise or lower the screws will be set to that value of ADJUST_TO_Z.
+  If DIRECTION is specified, the knob turns will all
   be in the same direction, clockwise (CW) or counterclockwise (CCW).
   See the PROBE command for details on the optional probe parameters.
   IMPORTANT: You MUST always do a G28 before using this command.
+  IMPORTANT: ADJUST_TO_Z and DIRECTION can't be used together!
 
 ### Z Tilt
 
--- a/klippy/extras/screws_tilt_adjust.py
+++ b/klippy/extras/screws_tilt_adjust.py
@@ -13,6 +13,7 @@ class ScrewsTiltAdjust:
         self.printer = config.get_printer()
         self.screws = []
         self.max_diff = None
+        self.adjust_to_z = None
         # Read config
         for i in range(99):
             prefix = "screw%d" % (i + 1,)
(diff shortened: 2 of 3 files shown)


Sineos:
You would need your full name as "signed off" line as per https://www.klipper3d.org/CONTRIBUTING.html

Just my personal opinion: While I see this being useful, the terminus `FIXED_POSITION` seems misleading to me. Maybe `FIXED_Z` or `ADJUST_TO_Z`

celtare21:
Thank you @Sineos, adjusted as per recommendations

klippy/extras/screws_tilt_adjust.py
```diff
@@ -73,6 +75,11 @@ def probe_finalize(self, offsets, positions):
         else:
             # First screw is the base position used for comparison
             i_base, z_base = 0, positions[0][2]
+
+            if self.adjust_to_z is not None:
```
dmbutyugin, line 79: Note that this will only work if `DIRECTION` is not set. This may lead to confusing results: if a user specifies `ADJUST_TO_Z`, they would expect that the recommendation brings bed to the specified level. However, this won't happen if they accidentally or unknowingly also set `DIRECTION`. I'd suggest to do it like
```
  if self.adjust_to_z is not None:
    ...
  elif self.direction is not None:
    ...
  else:
    # First screw is the base position used for comparison
    i_base, z_base = 0, positions[0][2]
```
I'd suggest to mention this behavior in the documentation. And ideally, the code in `cmd_SCREWS_TILT_CALCULATE` should check that both options are not set together (in general, it is impossible to meet both requirements) and report an error.

celtare21: Yes it is intented to work with DIRECTION not set as that would not make sense. Sure i can check if both are set together and output something

dmbutyugin: Well, you should always expect errors in the user input unless validated :) Especially in this potentially non-obvious case.

docs/G-Codes.md
```diff
@@ -485,14 +485,17 @@ The following commands are available when the
...
   nozzle to different locations (as defined in the config file)
   probing the z height and calculate the number of knob turns to
-  adjust the bed level. If DIRECTION is specified, the knob turns will all
+  adjust the bed level. If ADJUST_TO_Z is specified, the z position
+  to raise or lower the screws will be set to that value of ADJUST_TO_Z.
+  If DIRECTION is specified, the knob turns will all
   be in the same direction, clockwise (CW) or counterclockwise (CCW).
   See the PROBE command for details on the optional probe parameters.
   IMPORTANT: You MUST always do a G28 before using this command.
+  IMPORTANT: ADJUST_TO_Z and DIRECTION can't be used together!
```
dmbutyugin, line 498: I would suggest not to use exclamation mark in the documentation (here and in Manual_Level.md). But otherwise looks good, thank you!

dmbutyugin (commented):
Thank you, now it looks good.

KevinOConnor:
Thanks.  I'm not sure what the state of this PR is.  I'm also not really sure what this PR does at a high-level.  I'm going to mark it as needing a reviewer that can help with that.

-Kevin

PR #4954, opened 2021-11-22
Klipper configuration for Anycubic i3 Mega with Trigorilla Pro motherboard

fex42:
I attach a config file for Anycubic i3 Mega S with the 32-bit Trigorilla mainboard (printer-anycubic-i3-mega-s-trigorilla-pro.cfg).

It's from user @wasgehtabjo in this forum who asked if someone can upload it to GitHub. Since I don't have this board myself I can't test it but I guess its a good idea to include the config since a lot of Anycubics have this board:

Here is the forum post where the config is from:

https://drucktipps3d.de/forum/topic/i3-mega-trigorilla-pro-und-klipper/?part=3#postid-183458


Sineos:
Thanks for the contribution. 
PR need to have a "signed off" line as per https://www.klipper3d.org/CONTRIBUTING.html

Please also check against https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md
On first glance:
* Header of the file should follow the standard of the other cfg files
* "STM32 Flash loader demonstrator" seems a Win only utility and needs log-in to download. A commonly available solution is preferable
* Extruder seems to be a geared one. It is preferable to use `gear_ratio`
* Filename of the cfg should contain the approximate year when the printer was released

fex42:
I don't have this board/printer myself (just wanted to help with the pull request) and the way it is flashed, the bootloader is lost. Pprobably we should not add this config to the "official" configuration collection.

Regarding the "gear_ratio", I found no printer config with that inside, so this would be the first printer cfg having it.

Flashing probably is possible with one of the described bootloaders and stm32flash / dfu-utils. But since I don't own the printer myself, I can't test it.

I close this PR for the time beeing. Maybe someone else owning this printer finds it and can test/describe flashing it without loosing the bootloader and use this one as a start.

PR #4956, opened 2021-11-23
gcode: add Gcode group feature

phil65:
This allows to assign a group to Gcode macros. This could be used by frontends to automatically group macros in the UI.

Theres quite a lot of stuff out there (ERCF, Klicky probe etc) adding a lot of macros to Klipper. These also get automatically exposed to the frontends.
Currently only Fluidd allows custom grouping of macros via UI settings. This is quite fiddly and that also has its drawbacks (f.e. renaming the Macro will lead to the macro losing its group)

This PR allows to assign a group directly to each macro, either by adding it to a macro in a .cfg file or via the corresponding methods in python. (ercf for example defines macros via both ways).

Implementation-wise this basically only mirrors how G-Code help strings work.

Signed-off-by: Philipp Temminghoff philipptemminghoff@gmail.com

### DIFF
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -230,6 +230,15 @@ might return:
 saved G-Code state", "PID_CALIBRATE": "Run PID calibration test",
 "QUERY_ADC": "Report the last value of an analog pin", ...}}`
 
+### gcode/groups
+
+This endpoint allows one to query the associated group for G-Code commands that have
+one defined. G-Codes delivered by Klipper itself do not have a group defined.
+Example:
+`{"id": 123, "method": "gcode/groups"}`
+might return:
+`{"id": 123, "result": {"MY_CUSTOM_GCODE": "Group1", "MY_CUSTOM_GCODE_2": "Group2", ...}}`
+
 ### gcode/script
 
 This endpoint allows one to run a series of G-Code commands. For example:
--- a/docs/Command_Templates.md
+++ b/docs/Command_Templates.md
@@ -48,6 +48,20 @@ gcode:
 This will be showing is you use the `HELP` command or use the autocomplete
 function.
 
+## Add a group to your macro
+
+You can also define a group for each G-Code macro. These can be used by front-end clients
+in order to group all your G-Codes automatically.
+Add `group:` with a concise group name in order to associate a macro with a group.
+For example:
+
+```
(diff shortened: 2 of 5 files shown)


th33xitus:
I think the actual macro management should be covered exclusively by the frontend and klipper should only provide supporting functions. 

With the idea in mind to make the fontend use the group name specified in the config for some kind of "auto-grouping" i see the following "issue": Adding, removing or renaming a macro group or assigning a macro to a group would require a firmware restart to parse the config file again. Current macro management in the frontend doesn't require that and you can re-organize all macros that exist at that moment without being forced to restart the firmware.

What about calling the "group" option "id" instead so a user can assign a unique id to each macro:
```yaml
[gcode_macro MACRO_NAME]
id: 8087dd6c-8733-4bce-8222-c9e6c34423d0
description: example description
gcode:
    M117 do something
```

Frontends then can use the id for assigning macros to a specific group which **they** manage if the id is available. If the id is not available, they still can fall back to use the macro name. I think that option gives more freedom to the frontend developer because he doesn't have to deal with group names given by a software which, on its own, doesn't even make use of macro groups.

So what am i basically saying?
Since you said you mirrored the description field, my idea basically boils down to suggesting a different name 😅

phil65:
Well, in the end, the frontend doesnt have to use the assigned groups or can allow to override those group names with its own system. They could also decide to just use the groups for their own grouping setup method (in Fluidd / Mainsail beta you choose all macros for your custom group from one very big list containing all macros, for many users this is >100 items long. This could also be improved by using this feature.)
I think not being able to assign group names "on-the-fly" (without a firmware restart) is not a big deal since you do that work only once.
Also, since macros can also be defined by python modules (f. e. ERCF), it´s sometimes really hard to know where some of the new-appearing macros come from. Groups could help with this. It would be pretty neat if you just get a whole new macro "category" with all the corresponding macros ater following the ERCF / klicky / Auto-Z-Calibration install procedure.

Anyways, if not wanted, I´ll close this again. No problem. :)

KevinOConnor:
Thanks.  It's not clear to me that there is consensus on this.  It might be best to raise this on Klipper Discourse to get further feedback.  I'm going to mark this as needing a reviewer if it's desired to go forward with this change.

-Kevin

PR #4957, opened 2021-11-23
Makefile: Allow overriding python version

ressu:
The build system still defaults to Python 2.x, but it's now possible to set PYTHON=python3 when building. This removes the need to install python2 on systems which don't have it by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ OBJCOPY=$(CROSS_PREFIX)objcopy
 OBJDUMP=$(CROSS_PREFIX)objdump
 STRIP=$(CROSS_PREFIX)strip
 CPP=cpp
-PYTHON=python2
+PYTHON?=python2
 
 # Source files
 src-y =


th33xitus:
@ressu Chances to get this PR merged is to update your commit message and add the required signed-off note to it. See: https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#format-of-commit-messages

After you updated this PR, hopefully this change can be merged :)

ressu:
Thanks for the heads up, I completely missed that guide. The commit message has been updated to match.

KevinOConnor:
Thanks.  But, I'm not sure why this patch is needed.  It's already possible to set the python version by running `make PYTHON=python3`.

-Kevin

ressu:
Let's use the following Makefile as example
```Makefile
FOO=Hello
BAR?=Makefile

all:
        @echo $(FOO) $(BAR)
```

If I invoke `make` with assignments from environment, I get the following.
```shell
$ FOO=Modify BAR=World make
Hello World
```

On the other hand, if I explicitly set the variables via arguments I get this:
```shell
$ make FOO=Modify BAR=World
Modify World
```

The latter will indeed work, but the pattern of taking the interpreter from environment is more consistent with how make generally works. Make will set a series of [compilers by default](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html) which are overridable from environment. To me this seems like a compiler selection more than anything else, which would support the use of environment variables too.

KevinOConnor:
Okay, I think I understand.  However, I'm not comfortable with Make pulling in settings from the environment - it increases the build complexity as an inadvertent name conflict could break the build in such a way that it is hard to diagnose.  The build currently supports `make PYTHON=python3` and I don't think we need to add additional mechanisms.

Cheers,
-Kevin

ressu:
Fair enough and I will gladly accept that from complexity side.

PR #4961, opened 2021-11-24
Replaces all SKR E3 default bed thermistor

streetgt:
Since most of people that uses SKR E3 v1, v1.2 and v2 are users that come have Ender 3 (E3 means Ender 3 right?) makes no sense having the default sensor wrong in the bed.

I had seen some people making the same question a few times in the Discord.

Signed-off-by: Tiago Cardoso <tiagocardosoweb@gmail.com>

### DIFF
--- a/config/generic-bigtreetech-skr-e3-dip.cfg
+++ b/config/generic-bigtreetech-skr-e3-dip.cfg
@@ -66,7 +66,7 @@ max_temp: 250
 
 [heater_bed]
 heater_pin: PC9
-sensor_type: ATC Semitec 104GT-2
+sensor_type: EPCOS 100K B57560G104F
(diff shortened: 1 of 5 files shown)


zomfg:
it's also in `generic-bigtreetech-skr-e3-dip.cfg` and `generic-bigtreetech-skr-e3-turbo.cfg`

streetgt:
> it's also in `generic-bigtreetech-skr-e3-dip.cfg` and `generic-bigtreetech-skr-e3-turbo.cfg`

Done. Thanks for informing.

KevinOConnor:
Thanks.  I don't see an issue with making this change.  Is it still worthwhile to merge?

-Kevin

PR #4962, opened 2021-11-25

Greendumb:
Update stock parameters


Sineos:
I cannot comment if the values are right, but they match to the Marlin definition of this printer.

In any case, your commit needs to have a "signed off" line as per https://www.klipper3d.org/CONTRIBUTING.html

PR #4964, opened 2021-11-25
statistics: Fix memory usage reading on Python 3.

vladimir-poleh:
Fixed work with `/proc/meminfo` to be compatible with both Python 2 and Python 3.

### DIFF
--- a/klippy/extras/statistics.py
+++ b/klippy/extras/statistics.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2018-2021  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import os, time, logging
+import os, time, logging, io
 
 class PrinterSysStats:
     def __init__(self, config):
@@ -13,7 +13,7 @@ class PrinterSysStats:
         self.last_mem_avail = 0
         self.mem_file = None
         try:
-            self.mem_file = open("/proc/meminfo", "rb")
+            self.mem_file = io.open("/proc/meminfo", "r")
         except:
             pass
         printer.register_event_handler("klippy:disconnect", self._disconnect)


zellneralex:
Works for me at bullseye Python3

adelyser:
A couple of comments:

- Your contribution needs a sign off line according to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md including your full real name.
- What is the specific issue you are fixing? Please add that to your PR description, either a link to the issue or the verbiage itself.
- Your title seems to indicate that there's a problem in python3. In python3, open() == io.open(), they are the same thing. I don't really understand how this fixes python3. If there's a reasoning as to why this fixes things, that should be outlined in your PR description.

zellneralex:
@adelyser may I can help since switched to python3 the printer.system_stats.memavail shows always zero. After applying that fix it shows the real value again.

vladimir-poleh:
> * Your contribution needs a sign off line according to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md including your full real name.

It is full real name. Just different language from the name in my account.

> * What is the specific issue you are fixing? Please add that to your PR description, either a link to the issue or the verbiage itself.

Issue: actual version of the Klipper running on the Python 3 is not providing host used memory information. In this case Fluidd frontend can't calculate memory usage and always showing 100%.

Will update PR.

> * Your title seems to indicate that there's a problem in python3. In python3, open() == io.open(), they are the same thing. I don't really understand how this fixes python3. If there's a reasoning as to why this fixes things, that should be outlined in your PR description.

Changing from `open` to `io.open` is recommended change for the code migration from Python 2 to Python 3. It's mostly for Python 2 as I can understand. Change of the file open mode is a real fix.

zellneralex:
@vladimir-poleh you need to provide a line like 

vladimir-poleh:
Commit message already contains this

`Signed-off-by: Uladzimir Palekh <visor.rti@gmail.com>`

vladimir-poleh:
@zellneralex, are there any additional changes required for this PR to match contribution rules?

I already have `Signed-off-by: Uladzimir Palekh <visor.rti@gmail.com>` in the commit message.

zellneralex:
@vladimir-poleh not to my knowledge. The problem is that the main developer is very busy atm. So it might take a while until it get merged.

KevinOConnor:
Thanks.  I implemented a slightly different change to fix this bug (commit 62cfc252).

-Kevin

PR #4968, opened 2021-11-28
Add support for russian Picaso 3D Designer 3d printer

white777777:
Printer have unique board based on 1284p and hd44780 like display connected through two i2s expanders (PCF8574A). I didn't make display pins and addresses configurable because I think that this is unique russian way of connecting such display.


KevinOConnor:
Thanks.  I'm not sure what the state of this PR is.  At a minimum it would need a signed-off-by line containing your full real name ( https://www.klipper3d.org/CONTRIBUTING.html ).  In addition, the hd44780 code would need to be refactored before making this change as I think having 3 copies of nearly the same code would be a maintenance burden.

-Kevin

PR #4972, opened 2021-11-28
Update config reference

jtrmal:
This has recently bit me -- I understand now that the reference was warning me, but I didn't understand all implications, so I thought this might help also someone else.


KevinOConnor:
I'm not sure what the state of this PR is.  It's not clear that we need to change the documentation for this.  At a minimum though, commits need to have a signed-off-by line containing your full real name ( https://www.klipper3d.org/CONTRIBUTING.html ).

-Kevin

jtrmal:
ok, let me close this -- sorry.

Issue #4915, opened 2021-11-12
Option 'step_pin' is not valid in section 'stepper_y1'


zrox86:
Hey does someone know what the this error is about"Option 'step_pin' is not valid in section 'stepper_y1'", 

have a mks robin nano v3,  on a corexz printer and two y step motors, with normal setup (also one y stepper) works great, if i put a second one it comes the error above, i have checked the pins on the board on github 

[https://github.com/makerbase-mks/MKS-Robin-Nano-V3.X/wiki/Wiring_Pinout_and_Size](url)

all 4 rev. have the same pin setup for this setup

[klippy.log](https://github.com/Klipper3d/klipper/files/7528825/klippy.log)

theophile:
Not sure if this is the reason for the error, but your [stepper_y1] config section is missing several required parameters. See here: https://github.com/Klipper3d/klipper/blob/master/docs/Config_Reference.md#stepper

zrox86:
Hi theophile, thx for the comment, but I have copied the parameters for the second stepper motor from the config file of voron 2.4 (as an orientation point, tweaked to my needs).

the step pin pd6 is somehow the problem, I am not sure is it bcs of the new update of klipper and the resulting mcu flashing, maybe the the pin map is not correct, I posted the problem to see if i have a problem in my logic of configuring the config file to see if sombody does have the same issue :D

zrox86:
I tried a different approach, used the configuration of pins for [stepper_y1] and reconfigured it as second extruder [extruder1], it seems klipper dont allow a second stepper for y axis also [stepper_y1] but a second extruder[extruder1], its logical but thought it would allow it. 

I found a solution for second stepper on y axis, configured the x (in klipper) as x (on board), z (in klipper) as y (on board) and  y (in klipper) as z (on board), bcs. on the board two z axis share the same driver.

It works :D, I will close the issue bcs found a workaround :D

Ed-k:
I had the same problem, but I want to use stepper_x1 and stepper_y1.  This patch allows multiple motors for any axis in corexy - https://github.com/Klipper3d/klipper/pull/4974

zrox86:
Thx for the ideas guys. My Idea with two motors on one driver (the workaround) have not worked properly.

 I used the code from here 
[https://github.com/Klipper3d/klipper/commit/bbd1ddadd189645fb5bcd263d991cdc7e99cf011](url)
and ported it to corexz.py like this

[...]

PR #4886, opened 2021-11-06
Multiple steppers per axis on CoreXY

TypQxQ:
Changes and sample configuration to make it possible to use multiple steppers on +/- axis on a CoreXY.
Using a stepper in each corner shortens the slack of the belt in a CoreXY. This is especially useful in larger setups.

This is tested in a test setup because my current quad stepper CoreXY is running RRF and I'm waiting for a new board to test Klipper on it but works well in a test setup.

I know this was earlier asked for by the HevORT group.

Signed-off-by: Andrei Ignat <andrei@ignat.se>

PR #4974, opened 2021-11-28
Support multiple motors for any axis in corexy

Ed-k:
The z axis supports multiple motors defined as stepper_z, stepper_z1, stepper_z2, ...
This allows the same thing for X and Y so stepper_x1, stepper_x2 ... and stepper_y1, ... will work.
This fixes issue https://github.com/Klipper3d/klipper/issues/4915


TypQxQ:
This is already being worked on in #4886. Please check that out.
Thank you for contributing!
/Andrei

KevinOConnor:
I think this is already available now with the merge of #4886 .

-Kevin

PR #4976, opened 2021-11-29
Add configuration item to enable / disable bilateral step pulse optimization

kluoyun:
After several days of testing, I found that if the motherboard with level conversion (to prevent short circuit from burning the main control) is updated to the new version of klipper (the newly added double-edge stepping pulse optimized version), there will be a stepper motor offset It’s not just me.
I tried to turn off the double edge stepping pulse optimization, yes, everything is normal after turning it off. Therefore I added a configuration item in'[stepper_x,y,z]', which is very useful for our users who use these motherboards .
thanks
Of course, it would be best if there is a better solution, thanks again.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -158,6 +158,11 @@ microsteps:
 #   following "unstep" signal edge. This is also used to set the
 #   minimum time between a step pulse and a direction change signal.
 #   The default is 0.000002 (which is 2us).
+#step_bothedges_optimized: True
+#   Enable the optimization of double side stepping pulse of stepping
+#   driver. This function may appear offset on some boards (UART or SPI
+#   mode). If so, please turn off this function.The default is True (Is
+#   enabled)
 endstop_pin:
 #   Endstop switch detection pin. If this endstop pin is on a
 #   different mcu than the stepper motor then it enables "multi-mcu
--- a/klippy/stepper.py
+++ b/klippy/stepper.py
@@ -17,10 +17,12 @@ class error(Exception):
 # Interface to low-level mcu and chelper code
 class MCU_stepper:
     def __init__(self, name, step_pin_params, dir_pin_params, step_dist,
-                 step_pulse_duration=None, units_in_radians=False):
+                 step_pulse_duration=None, step_bothedges_optimized=True,
+                 units_in_radians=False):
         self._name = name
         self._step_dist = step_dist
         self._step_pulse_duration = step_pulse_duration
+        self._step_bothedges_optimized = step_bothedges_optimized
         self._units_in_radians = units_in_radians
         self._mcu = step_pin_params['chip']
         self._oid = oid = self._mcu.create_oid()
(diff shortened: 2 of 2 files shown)


dalegaard:
Hi there!

I don't have any comments on functionality, but please review the error log in the build checks and make sure to sign-off your commits. See https://www.klipper3d.org/CONTRIBUTING.html for details on how and why :-)

It would also be useful to squash the two commits together and add a comment in the commit as to why it's required. Finally, please follow the commit first line format on the CONTRIBUTING document.

This will help speed up action on this! Thanks!

- Lasse

kluoyun:
> 

Ok, thank you very much for your reply

kluoyun:
Hi there!

I don't have any comments on functionality, but please review the error log in the build checks and make sure to sign-off your commits. See https://www.klipper3d.org/CONTRIBUTING.html for details on how and why :-)

It would also be useful to squash the two commits together and add a comment in the commit as to why it's required. Finally, please follow the commit first line format on the CONTRIBUTING document.

This will help speed up action on this! Thanks!

Lasse

kluoyun:
module: New configuration items for enabling and disabling double-edge step pulse optimization

Enabling or disabling double-edge stepping pulse optimization in the configuration file can solve the problem of motor drift on some motherboards.

If you have a better solution, please let me know, thank you

Signed-off-by: XiaoK <xiaok@zxkxz.cn>

kluoyun:
> Hi there!
> 
> I don't have any comments on functionality, but please review the error log in the build checks and make sure to sign-off your commits. See https://www.klipper3d.org/CONTRIBUTING.html for details on how and why :-)
> 
> It would also be useful to squash the two commits together and add a comment in the commit as to why it's required. Finally, please follow the commit first line format on the CONTRIBUTING document.
> 
> This will help speed up action on this! Thanks!
> 
> * Lasse

Hello, should I close this PR and start over

KevinOConnor:
FYI, this issue was also raised informally on the Klipper Discord.  As high-level feedback, I would want to see a diagnosis of the root cause of the issue before making any code changes.  That is, I'm reluctant to "fix a symptom".

Separately, as mentioned, PRs would need to follow the guidelines at https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

KevinOConnor:
FYI, I realized that commit 689231df made a change in behavior that should have been documented.  (Specifically, the default step pulse duration went from 2us to as little as ~300ns for TMC drivers configured in UART/SPI mode.) I just committed a change to make sure the documentation properly reflects this change in behavior (commit 520273e5).  The change also makes it possible to configure a custom `step_pulse_duration` even for TMC drivers configured in UART/SPI mode.

It's unclear to me if this fixes the underlying issue reported here - but if you do have a level shifter that requires a larger switching time than the TMC drivers require then you should grab the latest code and configure an appropriate time in the `step_pulse_duration` setting.

-Kevin

kluoyun:
OK, thank you very much for your attention. I'll test it today

PR #4979, opened 2021-11-30
Update anycubic vyper config 

cryd-s:
Signed-off-by: Manuel Fichtner manuel.fichtner@me.com

### DIFF
--- a/config/printer-anycubic-vyper-2021.cfg
+++ b/config/printer-anycubic-vyper-2021.cfg
@@ -12,8 +12,8 @@
 #     - Communication interface: Serial (on USART1 PA10/PA9)
 #
 # To install the Firmware rename the klipper bin to `main_board_20xxxxxx.bin`
-# and copy it to an SD Card.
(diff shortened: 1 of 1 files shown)


daduke:
does the nozzle LED in this config work for anyone? I tried `SET_PIN PIN=LED VALUE=1`, it says OK but doesn't turn on the LED

Crykor90:
> does the nozzle LED in this config work for anyone? I tried `SET_PIN PIN=LED VALUE=1`, it says OK but doesn't turn on the LED. I can't seem to find the `pin: mcu:` config syntax mentioned anywhere either, is this correct? The same seems to be true for the `temperature_fan Mainboard` which doesn't turn on for me.

You have to disable SWD in the makemenu config ;)

KevinOConnor:
I'm not sure what the state of this PR is.  I'm fine with merging this change, but I think the comments should be wrapped to 80 columns and the spelling/grammar should be checked.

-Kevin

PR #4986, opened 2021-12-02
Add configuration for Modix Big60 printer

gjsmo:
config: Add configuration for Modix Big60 printer

Add example configuration for the Modix Big60 printer with the following configuration:

* Duet 2 Wifi
* Dual Volcano hotends
* PT100 RTDs
* BLtouch

Signed-off-by: Gabriel Smolnycki <gabjsmo0@gmail.com>


Sineos:
Thanks for your contribution.

Some comments:
* Please cross check with https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md --> e.g. the filename should contain the year of the printer release etc
* Only the original printer configuration will be added. There are too many custom modifications out there and the cfg's should support the printer as delivered by the manufacturer
* For the TMC settings please cross check with https://github.com/Klipper3d/klipper/pull/4977

gjsmo:
@Sineos 

It is unlikely any config for this printer will match exactly with anyone else's, as it is fairly configurable from the manufacturer. The hotends in particular have a number of different stock options. Nonetheless, I think this is as close as it will get to "standard".

For whatever reason they used a separate PID controller for the bed heater, so I did build a heater mod for it. Is there anywhere to include that section of the configuration? It seems a shame to not share it with others.

As far as the TMC driver options, these match the manufacturer's configuration for RRF. Both options as configured make the noise considerably more tolerable. I wouldn't consider them optional in this case.

With that being said, I believe I've made all of the requested updates to match the docs/Example_Configs.md recommendations.

Sineos:
Thanks for the changes

> It is unlikely any config for this printer will match exactly with anyone else's, as it is fairly configurable from the manufacturer. The hotends in particular have a number of different stock options. Nonetheless, I think this is as close as it will get to "standard".

Agreed

> For whatever reason they used a separate PID controller for the bed heater, so I did build a heater mod for it. Is there anywhere to include that section of the configuration? It seems a shame to not share it with others.

While it is a cool solution, the goal is to support the printer as it comes from the manufacturer to allow (Klipper) beginners a seamless start.

> As far as the TMC driver options, these match the manufacturer's configuration for RRF. Both options as configured make the noise considerably more tolerable. I wouldn't consider them optional in this case.

- A powerful 32bit board like the Duet can easily achieve 128 (or more) native microsteps in current Klipper versions. No need to pay for a potential impact on accuracy by using interpolate (on my 120MHz board I can run 256 microsteps w/o issue)
- The current switching from `run_current` to `hold_current` has been found to also incur potential impact on accuracy at little or no gain in terms of temperature 
- Are you sure about the z-stepper `run_current: 2.400`? This would give a max current of ~3.4A for this stepper

You have some stray blank lines in the config, otherwise looks good for me 👍

gjsmo:
@Sineos

Regarding the driver config, I did spend quite a bit of time tuning to get as close to the RRF performance as possible. Notably, some of the defaults (particularly interpolate and hold_current) make the steppers _incredibly_ loud. They are significantly larger than a typical printer. I personally cannot stand being in the same room with the defaults.

The currents are definitely correct. There are four Z steppers and two X steppers, this certainly contributes to the noise. I found that the hold_current option had a large effect, I wouldn't be surprised if that's mostly on the Z axis.

I can try turning microsteps up to 256 and turning off interpolation and see if it's the same in terms of noise - in theory it should be. I can also test turning off hold_current, it may have acceptable results on X and Y. I'm not concerned about motor heating, it really is all about the noise. I'll try that sometime this weekend.

Gilabite:
@gjsmo I have 42-60 motors on my big corexy and turning off interpolation and turning microstepping to 256 makes a huge difference in terms of noise

Sineos:
> Notably, some of the defaults (particularly interpolate and hold_current) make the steppers _incredibly_ loud.

Following your input, I tested `hold_current` on my printer and in terms of noise it makes virtually no difference at all. The above linked pull-request will remove all `hold_current` from all the cfg.

> The currents are definitely correct.

According to https://www.modix3d.com/tech-spec/, the used steppers are MT-1705HS200A, with a rated current of 2A (most likely peak current). As such a peak current of ~3.4A does not seem correct

gjsmo:
@Sineos

I had a chance to test and push these changes today. Going to 256 microsteps and turning off interpolate had no effect, which I suppose make sense. Without idle_current_percent, I detect an increase in high frequency noise (the "hiss") but it's not as bad as I remember. I believe I originally converted this printer to run Klipper at v0.8.0, so it's definitely possible some changes have been made in the meantime.

Yes, there are four Z axis steppers. They are all in parallel, which is most common configuration in my experience. 2.4A is the correct current, they will not move if it's lowered too much more - the bed alone is around 15lb. This undoubtedly contributes to the noise, considering that there's a total of 9 motors. I'll be keeping idle_current_percent in place on my machine, for my own sanity.

Sineos:
Looks good to me

KevinOConnor:
Thanks.  I noticed a few minor things:
1. Prefer `name: value` syntax (for bltouch/z_offset setting).
2. Can you explain how the Z rotation_distance of 1.6 was obtained?
3. It really helps if you can add this new printer config file to tests/klippy/printers.test .

-Kevin

gjsmo:
@KevinOConnor 

I added the new config to test/klipper/printers.test, looks like I overlooked that part of the example config recommendations.

I fixed the bltouch/z_offset line. It's worth noting that this is copied from the SAVE_CONFIG section, which appears to use `=` instead of `:` - worth opening an issue for that?

The Z `rotation_distance` of 1.6 is a direct conversion from the RRF config. The steppers have a belt reduction and the ballscrews are also fairly fine pitch. I can confirm that parts printed with this config are dimensionally accurate, even up to around 500mm tall.

KevinOConnor:

If the printer uses common T8 leadscrews for Z with a 5:1 gear reduction then it's preferable to use `gear_ratio: 5:1` and `rotation_distance: 8` for the Z (similar to the recommendation for extruders).  Otherwise, do you know how the printer actually gets to a rotation_distance of 1.6?

-Kevin

robthide37:
Alot of these machines use an expansion board and then have separate drivers for the x motors since they run cartesian with a y rail floating vs x rail . the question is the rrf setup like this runs an endstop on each y motor and I could not find a way to configure this on klipper ?

robthide37:
nevermind i figured it out

PR #4988, opened 2021-12-02
add atmega328/p internal temperature sensor adc pin support

eikeime:
Add PE4 as atmega328/P internal temperature sensor pin.
Extend ADMUX adc pin register to mux3.

in printer.cfg user could add this sensor type to use this sensor pin:
`[adc_temperature M328]`
`temperature1: -40`
`voltage1: 0.257`
`temperature2: 125`
`voltage2: 0.516`

Signed-off-by: Hu Guangzheng <eikeime@gmail.com>

### DIFF
--- a/src/avr/adc.c
+++ b/src/avr/adc.c
@@ -11,10 +11,14 @@
 #include "pgm.h" // PROGMEM
 #include "sched.h" // sched_shutdown
 
+#if CONFIG_MACH_atmega328p
+    #define ADC_TEMPERATURE_PIN 0x04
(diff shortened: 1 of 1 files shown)


Sineos:
Many thanks. Without judging code or functionality, some remarks:

* Your contribution needs a sign off line according to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md including your full real name.
* The change has to be documented. Please see https://www.klipper3d.org/Config_Reference.html?#builtin-micro-controller-temperature-sensor

PR #4993, opened 2021-12-05
github: adds GitHub issue templates

pedrolamas:

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,42 @@
+name: Bug report
+description: Create a report to help us improve
+labels: ["bug", "triage"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        This issue form is for reporting bugs only!
+        If you have a feature request, please use [feature_request](/new?template=feature_request.yml)
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened
+      description: >-
+        A clear and concise description of what the bug is.
+    validations:
+      required: true
+  - type: textarea
+    id: expected-behavior
+    attributes:
+      label: What did you expect to happen
+      description: >-
+        A clear and concise description of what you expected to happen.
+    validations:
+      required: true
+  - type: textarea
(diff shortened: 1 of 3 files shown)


pedrolamas:
@KevinOConnor not sure if this is something that interests you, but I thought it would be a good thing anyway!

This was based on my own, as shown here:
https://github.com/pedrolamas/pedrolamas.com/issues/new/choose

Also in use now in Moonraker here:
https://github.com/Arksine/moonraker/issues/new/choose

KevinOConnor:
Thanks.  A change like this would need to also come with an associated change to Contact.md as we currently consider feature requests and general requests for help to be "not in scope" for the github issue tracker.  The only thing the documentation currently considers "in scope" for the issue tracker is "sharing the results of work done to improve Klipper".

That said, I've "given up" on maintaining the github issue tracker.  If there are some regular Klipper contributors that are interested in maintaining it going forward then I think they should have the most say in how it is run.  If so, lets start with a discussion on what the workflow should be.

Cheers,
-Kevin

pedrolamas:
Thanks Kevin, that's quite fair!

I'll leave this open for a while more, just in case anyone else wishes to make any further comments.

dewi-ny-je:
"This issue form is for reporting bugs only!"

I would add an extra line:
"Requests related to help for configuration or usage will be closed without any further delay. They should be asked in the Discord chat and Discourse web forum."

The form for feature requests should also state that:
"It is unlikely that any feature request will be taken by anyone, since currently the workload on the developers is far higher than what they can handle. If you still want to file a feature request, be aware that it will be mostly for keeping track of it, in case in the future someone thinks about the same feature."

Kevin decided that the issue tracker should be used only for things on which someone is working on, which I strongly disagree since if there is REALLY a bug, it should be opened even by people who are unable to code... and (if confirmed) it should not be closed as long as it is present.

Once these new templates are in place I can take care of checking new issues and close the requests for support and take care of asking about more information/logs when they appear to be related to real bugs. I cannot code but I'm already checking regularly issues and pulls out of curiosity to see if there are interesting discussions and features... Cleaning up issues would not be a big job on top of it.

PR #4994, opened 2021-12-05
Add files to support AR100

eliasbakken:
SUPPORT FOR AR100 CPU

Some Allwinner SoCs have a coprocessor that 
can be used to add real time functionality to 
a Linux based OS. The AR100 present in 
A64, H3, H5 and H6 are examples of this. 
This PR adds support for the AR100 in the A64
which is in use on Recore controller board. 

Signed-off-by: Elias Bakken <elias@iagent.no>


src/ar100/gpio.h
```diff
@@ -0,0 +1,60 @@
...
+  PIO_ALT4,
+  PIO_ALT5,
+  PIO_DISABLE
+};
+
+struct gpio_mux {
+  uint32_t pin;
+  uint8_t bank;
+  uint32_t reg;
+};
```
KevinOConnor, line 23: It would be preferable to not define ar100 specific definitions in gpio.h.  Instead it's typical to create an "internal.h" header for that purpose.  The reason this is preferable is it can help avoid possible future naming conflicts with the generic code.

src/ar100/gpio.h
```diff
@@ -0,0 +1,60 @@
...
+gpio_out_toggle_noirq(struct gpio_out pin){
+  data_regs[pin.bank] ^= (1<<pin.pin);
+  *((volatile uint32_t *)(pin.reg)) = data_regs[pin.bank];
+}
+
+static inline __attribute__((always_inline)) void
+gpio_out_toggle(struct gpio_out pin){
+  gpio_out_toggle_noirq(pin);
+}
+
```
KevinOConnor, line 55: It would be preferable to define these functions in the c code.  Since we use `-flto` during the compilation, it is still possible to ensure code is inlined even when defined in the c files.

src/ar100/main.c
```diff
@@ -0,0 +1,176 @@
...
+    if (ret) {
+        len -= pop_count;
+        if (len) {
+            memcpy(receive_buf, &receive_buf[pop_count], len);
+            sched_wake_task(&console_wake);
+        }
+    }
+    receive_pos = len;
+}
+DECL_TASK(console_task);
```
KevinOConnor, line 112: Can this uart code utilize the helpers in generic/serial_irq.c ?

src/ar100/main.c
```diff
@@ -0,0 +1,176 @@
...
+// Handle shutdown request from ar100
+static void
+shutdown_handler(uint32_t *args)
+{
+    shutdown("Request from ar100");
+}
+
+const struct command_parser shutdown_request = {
+    .func = shutdown_handler,
+};
```
KevinOConnor, line 135: Dead code?

src/ar100/main.c
```diff
@@ -0,0 +1,176 @@
...
+    goto *reset;
+}
+DECL_COMMAND_FLAGS(command_reset, HF_IN_SHUTDOWN, "reset");
+
+void
+save_data(void)
+{
+    extern char __data_start, __data_end, __copy_start;
+    memcpy (&__copy_start, &__data_start, &__data_end - &__data_start);
+}
```
KevinOConnor, line 158: What does this save/restore code do?

src/ar100/main.c
```diff
@@ -0,0 +1,176 @@
...
+{
+    save_data();
+    r_uart_init();
+    uart_puts("** AR100 v0.1.4 ");
+    if(exception == 0){
+      uart_puts("start ** \n");
+    }
+    if(exception == 1){
+      uart_puts("reset ** \n");
+    }
```
KevinOConnor, line 173: In general, we don't want to add debugging code to the main Klipper repository.  That is, the main code already generates status messages over the main "console" channel and supports a debugging feature via the `output()` function.  I've found that trying to maintain a secondary debugging channel can become a maintenance burden.

src/ar100/spicmds.c
```diff
@@ -0,0 +1,141 @@
+// Commands for sending messages on an SPI bus
+//
```
KevinOConnor, line 2: Why is this file copied from src/spicmds.c?  What's different about it and why can't the main spicmds.c be used?

src/ar100/timer.c
```diff
@@ -0,0 +1,52 @@
...
+}
+
+void timer_reset(void){
+  mtspr(SPR_TICK_TTCR_ADDR, 0);
+}
+
+// Activate timer dispatch as soon as possible
+void timer_kick(void){
+    timer_set(timer_read_time() + 50);
+}
```
KevinOConnor, line 45: I'm not particularly concerned about indentation style, but I do think the src/ar100/ code should follow one style.  So, 2 space, 3 space, or 4 space indents, but not a mix.  If you have no preference, matching the other mcu code with 4 spaces would be preferred.

src/ar100/timer.h
```diff
@@ -0,0 +1,21 @@
...
+    unsigned long ret;
+    __asm__ __volatile__ ("l.mfspr %0,r0,%1" : "=r" (ret) : "K" (add));
+    return ret;
+}
+
+static inline void mtspr(unsigned long add, unsigned long val){
+    __asm__ __volatile__ ("l.mtspr r0,%1,%0" : : "K" (add), "r" (val));
+}
+
+uint8_t timer_interrupt_pending(void);
```
KevinOConnor, line 15: Same comment as gpio.h - since timer.h is used by the generic code it would be preferable to move local definitions to a local header (eg, internal.h).

KevinOConnor (commented):
Thanks.  At a high-level the changes look good to me.  I do have some minor code comments and code questions below.  I also have some high-level questions that I'll send in a separate comment.

-Kevin

KevinOConnor:
I have a couple of high-level questions:

Typically, we try to target Klipper enhancements to a user base of at least 100 people.  It's not clear to me that this support would currently reach that threshold.  I think it would be okay to commit this in preparation for an upcoming user base of sufficient size.  Can you provide some info on expected rollout of boards with this chip?  Do you know which readily available chips have the ar100?  Should we commit this change on an expected user base that does not materialize, and if we find it impedes future development, then be aware that we might need to remove support.

Are there significant code size limitations to the ar100?  I noticed there is a comment in the Makefile about code size.  The PRU does have a notable code limitation (no more than 2048 instructions total on a PRU instance).  I'm not sure if the ar100 has a similar type of limitation.

How are interrupts handled on the ar100?  I noticed this code uses polling (as the ar100 and "linux" mcu does).  I'm not sure if that was done to simplify the code or if the ar100 is not capable of generating traditional "interrupts".

Thanks,
-Kevin

KevinOConnor:
Also, what other code changes are needed to get the Recore board working after merging this PR?  Are there host code changes needed?

-Kevin

eliasbakken:
I'll fix the issues you have commented on and resubmit. 
As for the high level questions, let me try and answer those now.
About 30 have been made and sold of rev A5, A6 is in production with a starting batch of 100. We can wait with the merge until those have reached customers if you want. 
The code is run from SRAM. It is 64 KB in total that can be used, but as far as I understand, that memory is shared with ATF (arm trusted firmware) needed for boot and shut down. I don't this it is needed during regular runtime, but I'm not sure. I've kept code base small to not interfere with ATF. 
I've not looked much into interrupts for ar100, but there is some support for gpio interrupts at least, depending on the gpio bank used. I'm thinking since nothing else is running on the cpu, then polling is a good strategy. The code base is not optimal, but it prints fine, and a decent starting point. Happy to discuss options if someone wants to optimize for speed or temperature.
 
There are some other minor changes that can be submitted as a separate PR to get full support for Recore. One thing is "built in" thermocouple support with an on board op-amp and cold junction compensation. The other is support for over current alarm using INA831. The rest i think can be mostly handled with existing code base. At least some of these features can be useful as general purpose code for other boards as well.

KevinOConnor:
Okay, thanks.  If deployment is imminent then I don't think we need to wait for it.

Some additional questions - what is the high-level plan for "watchdog" and emergency heater stopping on the ar100?  Will the recore boards always run a "linux mcu" (similar to the way the beaglebone/pru code does)?

-Kevin

eliasbakken:
Great! 

The ar100 only controls the stepper motors and end stops on Recore. There is a separate (low cost) STM32 for heaters and analog stuff, so that should be taken care of I think. Right now there is no "Linux mcu" similar to the Beaglebone implementation. 

I'm not sure if that answers the question? It should be more similar to a RPI setup in that regard.

KevinOConnor:
As another question - I didn't see `DECL_ENUMERATION_RANGE("pin", ...)` declarations in the code - what is the common pin naming scheme for these micro-controllers?  I don't think we want to export raw integers for the pins - as no other board does that today.

-Kevin

eliasbakken:
Ah, I think most of the code was developed before that was done on a micro controller level or something. There used to be a mapping in the pins.py file, but it was removed. I'll take a look at the other platforms and see if something similar can be done.

src/ar100/Makefile
```diff
@@ -0,0 +1,58 @@
...
+	$(Q) cp $(OUT)ar100.bin /opt/firmware
+
+flash: $(OUT)ar100.bin
+	$(Q) ./scripts/flash-ar100.py $(OUT)ar100.bin
+
+dump:
+	$(CROSS_PREFIX)objdump -d out/klipper.elf  > out/klipper.dump
+
+size:
+	$(CROSS_PREFIX)size out/klipper.elf
```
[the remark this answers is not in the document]
eliasbakken: These are really just helper targets for development, not needed upstream.

src/ar100/main.c
eliasbakken: This looks a bit strange, I agree. The problem is that since AR100 code is run from SRAM, a "reset" will not restore any variables from the ".data" segment. Those variables are only written during the initial loading of the program (the main CPU loads the binary into SRAM via systemd) and then the data is lost after execution begins. Therefore, the data needs to be stored somewhere. The code above here makes a copy of the ".data" segment and restores it upon reset. I think a better solution might be to identify the variables in question (I think it's only 4-5 variables) and set the value explicitly. It will require some development though. I'm open to other ideas as well.

src/ar100/main.c
eliasbakken: The debug string coming from this code is output directly to UART0, the same port normally used for kernel messages etc. But it's OK to remove it now, it was for debugging and at the current maturity level, it is not needed.

src/ar100/main.c
eliasbakken: Yes, I think some of this code can be removed in favor of what is in generic/serial_irq.c. Regarding your question about interrupts, there is support for UART interrupt, which would match better with the existing code base. It is a bit of work though, so if it's possible to do it in a separate PR, that would be great.

eliasbakken:
I've addressed some of the issues, but I've not gotten around to retesting the benchmarks that look weird. I'll take another look at this soon.

brotherdust:
@eliasbakken, was this ever merged? I'd like to get my Recore board up and running on the latest Klipper fw. Thanks!

eliasbakken:
@brotherdust there were a couple of minor things that I needed to fix for the merge to go through. I will resubmit once I get some time to look at this again. In the meantime, I (or you) can sync with the upstream branch to get the latest changes. There is a need for one more PR once this one has been merged, so it might be a while longer.

PR #4997, opened 2021-12-06
manual_stepper: add `get_status` method

ayufan:
This exposes two additional informations on `manual_steppers`:

- `enabled`
- `position`
- `mcu_position`

This is status returned by Moonraker `/printer/objects/query?manual_stepper idler_stepper`:

```json
{
  "result": {
    "eventtime": 1628.386874847,
    "status": {
      "manual_stepper idler_stepper": {
        "enabled": false,
        "position": 0.0,
        "mcu_position": -231
      }
    }
  }
}
```

### DIFF
--- a/klippy/extras/manual_stepper.py
+++ b/klippy/extras/manual_stepper.py
@@ -103,6 +103,13 @@ class ManualStepper:
             self.do_move(movepos, speed, accel, sync)
         elif gcmd.get_int('SYNC', 0):
             self.sync_print_time()
+    def get_status(self, eventtime):
+        stepper_enable = self.printer.lookup_object('stepper_enable')
+        se = stepper_enable.lookup_enable(self.rail.get_name())
+        return {
+            'enabled': se.is_motor_enabled(),
+            'position': self.rail.get_commanded_position()
+        }
     # Toolhead wrappers to support homing
     def flush_step_generation(self):
         self.sync_print_time()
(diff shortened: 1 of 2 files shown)


klippy/extras/manual_stepper.py
```diff
@@ -103,6 +103,13 @@ def cmd_MANUAL_STEPPER(self, gcmd):
             self.do_move(movepos, speed, accel, sync)
         elif gcmd.get_int('SYNC', 0):
             self.sync_print_time()
+    def get_status(self, eventtime):
```
ayufan, line 106: I was unsure which one is better. This alternative variant:

```python

        if not self.steppers:
            return {}
        stepper_enable = self.printer.lookup_object('stepper_enable')
        s = self.steppers[0]
        se = stepper_enable.lookup_enable(s.get_name())
        return {
            'enabled': se.is_motor_enabled(),
            'position': s.rail.get_commanded_position()
        }
```

KevinOConnor:

In practice, the `get_status()` function gets called a lot - many clients just subscribe to everything and as a result every `get_status()` method gets called several times a second.

As a result, I'm leery of adding any notable function calls to a `get_status()` function.  It's fine to return local variables (eg, the current position), but I don't think we want to provide low-level values that the user is unlikely to be able to use (eg, mcu_position).

Also, is this info now available via the `motion_report` module?

-Kevin

ayufan:
> Also, is this info now available via the motion_report module?

I will have to check. It was quite some time since I authored this :)

Issue #1787, opened 2019-07-05
[FIRMWARE RETRACTION] Z-Hop not taken into acount during z moves


Nitrock28:
There seems to be an issue when using firmware retraction with a non-zero Z-hop, where if a move containing a Z coordinate is issued after retraction, the Z hop offset is not taken into account. 

I am using the prusa-slicer on my delta printer and the Gcode looks like this at the beginning : 

```
****Heaters startup*****
G92 E0
G10 ; retract
G92 E0
G1 Z0.300 F9000.000
G1 X-0.004 Y-37.308
G11 ; unretract
G92 E0
****First move****
```
 I have activated the "retraction at layer change" feature, so the same problem happens on every layer. The retraction happens before the new Z value is set and the unrectract after.

What currently happens is that the toolhead actually moves at the requested Z position (0.3mm from the bed) and when the unretract is issued, the nozzle crashes in the bed as the Z-Hop is undone.

Shouldn't the Zhop be treated the same way as a Gcode offset? I am pretty sure Marlin treats it that way.

[klippy.zip](https://github.com/KevinOConnor/klipper/files/3362885/klippy.zip)

KevinOConnor:
This is a known issue that was discussed at the end of PR #1617.  Unfortunately, there does not seem to be an easy fix for this.  It may be necessary to remove the z_hop capability from the firmware_retraction module.

-Kevin

KevinOConnor:
FYI, I removed support for z_hop (commit 0343d0c7) until we can get this sorted out.

-Kevin

Nitrock28:
Ok it makes sense,
I temporarly solved my issue using gcode macros with a variable for the retracted state, and using the gcode_offset to do the Zhop as suggested in PR #1617.
For the most part it works, but I had to take make a "safe" set_gcode_offset alias macro to take the curent value of Z_Hop into account, otherwise tuning the first layer can be dangerous if the offset is applied during a retraction. I don't have the macros on hand at the moment but I can post them later if it's of any interest.

KevinOConnor:
FYI, in a separate conversation I had with @Arksine, we came up with two possible solutions:

[...]

PR #4998, opened 2021-12-06
firmware_retraction: Reintroduce support for lifting Z on retract

wlhlm:
This reintroduces lifting Z on G10. The initial implementation of `firmware_retraction` had similar functionality, but didn't account for unrelated Z movements of the toolhead after G10, but before G11 (#1787) and was quickly [removed](https://github.com/Klipper3d/klipper/commit/0343d0c73aed236232592e9c46874de7ab761139). This pull request follows one of the strategies outlined by Kevin in https://github.com/Klipper3d/klipper/issues/1787#issuecomment-510531375 to deal with the issue of the original code.

## Implementation details

Kevin [outlined](https://github.com/Klipper3d/klipper/issues/1787#issuecomment-510531375) two approaches for implementing Lift Z in a way that deals with the issue of moving the toolhead in Z after G10. I think the method of overwriting G0/G1 with special procedures that check for Z moves on G10 (and subsequent undo on G11) is the most straightforward approach and easiest to reason about, implementation-wise.

On top I've added some testcases that cover both basic `firmware_retraction` functionality and the new lift Z feature to increase confidence that bugs like #1787 don't happen again.

## Performance

I have done some performance testing using batch mode and wasn't able to find significant performance regressions using Lift Z. Since `firmware_retraction`, as it is currently implemented in Klipper, is pretty much just a glorified macro, I've also reimplemented G10/G11+lift Z purely as macro and there I saw a significant performance hit (to the tune of 60% longer processing times in batch mode). The issue with the pure macro implementation is that in order to detect Z movements, G0/G1 need to be permanently overwritten since macros can't be defined at runtime, which means every G0/G1 incurs a jinja conditional check leading to bad performance.

Based on this I think adding this as a Python implementation is the right approach and it actually gives `firmware_retraction` more legitimacy as a Python module rather than just a macro.


klippy/extras/firmware_retraction.py
```diff
@@ -14,22 +14,46 @@ def __init__(self, config):
...
+            self.prev_G0 = self.gcode.register_command('G0', None)
+            self.gcode.register_command('G0', self.cmd_G0)
+            self.prev_G1 = self.gcode.register_command('G1', None)
+            self.gcode.register_command('G1', self.cmd_G1)
+        elif self.prev_G0 and self.prev_G1:
+            self.gcode.register_command('G0', None)
+            self.gcode.register_command('G0', self.prev_G0)
+            self.gcode.register_command('G1', None)
+            self.gcode.register_command('G1', self.prev_G1)
+            self.prev_G0 = self.prev_G1 = None
```
wlhlm, line 49: This code looks a bit 'ugly' from far away, but I'm unable to come up with a nicer implementation. I'm interested in feedback. Also are people fine with the `func(undo=True/False)` pattern rather than `func()/func_reset()` for example?

klippy/extras/firmware_retraction.py
```diff
@@ -42,33 +66,53 @@ def cmd_SET_RETRACTION(self, gcmd):
...
-                             self.unretract_extra_length, self.unretract_speed))
+                             self.unretract_extra_length, self.unretract_speed,
+                             self.lift_z))
+
+    def cmd_G0(self, gcmd):
+        self.prev_G0(gcmd)
+        self._check_z_lift(gcmd)
+    def cmd_G1(self, gcmd):
+        self.prev_G1(gcmd)
+        self._check_z_lift(gcmd)
```
wlhlm, line 87: I'm explicitly overwriting G0 and G1 separately here even though their basic klipper implementation is identical since users might redefine them separately using macros.

kamilkrzywda (approved):

KevinOConnor:
Thanks.  As a high-level comment though, I'm not sure I'm going to be able to provide a sufficient review of this.  I know from past experience there were several "unusual corner-cases" with the z-lift functionality that could cause nozzle crashes if not handled correctly.  That is why the original z-lift functionality was removed.  Maybe @FHeilmann could help with a review of this?

-Kevin

FHeilmann:
> Thanks. As a high-level comment though, I'm not sure I'm going to be able to provide a sufficient review of this. I know from past experience there were several "unusual corner-cases" with the z-lift functionality that could cause nozzle crashes if not handled correctly. That is why the original z-lift functionality was removed. Maybe @FHeilmann could help with a review of this?
> 
> -Kevin

I'm not sure if redefining G0 and G1 is a good idea here, wouldn't it be sufficient to record and check the Z-positions at retract and unretract, respectively?

If G0 and G1 are redefined, this will silently break some of the bigger configs out there that redefine G0 and G1 themselves, creating further headaches, or rather making those config packs incompatible with this change.

Sonnenwerk:
Hi there, 
sadly i am not able to review this topic, but i would like to point out, that i am still interested in this feature (firmware retract incl. zhop). It would really help to reduce my slicing/uploading effort when i change my printers toolhead(s).

tfjield:
I hope this can be implemented.  I've been using the macro implementation, and if there is a performance hit it would be great to see it incorporated via Python.

encodedLife:
Hello @wlhlm,

first of all thank you for your contribution. This is one of the features, I really missed in Klipper Firmware.
I`m using your retraction script since one week and I observed one thing which is suspicious for me. In your script in G11 command, you extrude first and then you are going down with the z-Axis. 
I think the order of actions should be first going down with the z-Axis and then unretract.

What do you think about that?

PR #5003, opened 2021-12-07
klippy.configfile: Add conditional include

pkucmus:
This is only a proof of concept for a feature @miklschmidt requested.
The feature builds a separate buffer for the `[include_conditionals]` section and parses that out to a conditional config.
The conditional config is later used to determine if all the conditions of the `[include_if]` block are met (in a ALL / AND fashion).

Signed-off-by: Paweł Kucmus <pkucmus@gmail.com>

Not being sure if this is the right place for it I'd like to at least start a conversation. @miklschmidt briefed me on his discussion with @KevinOConnor and I implemented what was proposed to me but I have a few alternative solutions proposals which I'll describe right after the proposed solution:

### `include_if` instruction

This would require a extra `[include_conditionals]` section that would hold the variables understood in the `include_if` headers

```ini
[include_conditionals]
hotend: v6
test: True

[include_if hotend==v6 test==True something.cfg]
```

In the proposed solution the `include_conditionals` would only be defined in the entrypoint .cfg file (i.e. printer.cfg) and would need to be above any `[include_if]` section.

This can work as presented in the POC but it's introducing this very limited `if` logic - which means we now have conditional logic in Klipper .cfg files. 

### `if` instruction

`include_if` that is being introduced is rubbing me the wrong way in general here as it's introducing "programming logic" into INI/.cfg files but if that is the way to go then maybe we should avoid a future where there will be a need for more `{instruction}_if` blocks and go straight into an `[if]` block?
The proposed solution would look like this:

```ini
[if hotend==v6 test==True include something.cfg]
```

then we could go into other conditionals like:

```ini
[if test==True temperature_sensor raspberry_pi]
sensor_type: temperature_host
```

I like the syntax more since it looks more like a normal programing language but this might be going too far.

### pre rendering

Another idea would be to not invent anything new but rather create a pre_render plugin system which Klippy could call to retrieve cfg files. In this solution we could create i.e. a Jinja2PreRenderer which would have the ability to go crazy with `if`, `for`, etc. Then output a cfg file (or files) which would be consumed by Klippy as usual. 
This would isolate the INI "standard" from any "advanced" logic while allowing - let's call them vendor needs - to be satisfied with Jinja2 (or other Python pre renderer plugin one may need).

I am a fan of isolation and this solution is producing the least impact on Klippy while opening a world of possibilities like calling an API to get values to fill a template (just an abstract example, not sure how much use it has).

Please let me know what you think.

### DIFF
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -237,6 +237,8 @@ class PrinterConfig:
         # Buffer lines between includes and parse as a unit so that overrides
         # in includes apply linearly as they do within a single file
         buffer = []
+        conditional_buffer = []
+        gathering_conditionals = False
         for line in lines:
             # Strip trailing comment
             pos = line.find('#')
@@ -245,6 +247,23 @@ class PrinterConfig:
             # Process include or buffer line
             mo = configparser.RawConfigParser.SECTCRE.match(line)
             header = mo and mo.group('header')
+            if header and header.startswith('include_conditionals'):
+                gathering_conditionals = True
+                conditional_buffer.append(line)
+                continue
+            elif gathering_conditionals:
+                if not line or header:
+                    gathering_conditionals = False
+                    self._parse_config_buffer(conditional_buffer, filename, fileconfig)
+                    conditional_config = ConfigWrapper(self.printer, fileconfig, {}, 'include_conditionals')
+                    continue
+                conditional_buffer.append(line)
+
+            if header and header.startswith('include_if '):
+                conditions, include = self._parse_conditional(header)
+                if all([conditional_config.get(variable) == value for variable, value in conditions]):
+                    header = "include {}".format(include)
+
             if header and header.startswith('include '):
                 self._parse_config_buffer(buffer, filename, fileconfig)
                 include_spec = header[8:].strip()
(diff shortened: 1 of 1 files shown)


pkucmus:
@miklschmidt maybe you could provide the rationale / use case that you are after here (for historical purposes)

miklschmidt:
@pkucmus Yessir! The following is the rationale i discussed with Kevin on the klipper discord. We agreed that a "mega driver" section wasn't a good idea, so i went ahead and discussed the conditional include solution with @pkucmus who took it upon himself to do the work 😍 (thanks man).

**Rationale**
RatOS/V-CoreOS comes with a modular config that allows people to set up their printer super quickly by commenting/uncommenting includes and a couple of overrides (such as endstop positions). This way control boards, extruders, hotends, endstops, probes, steppers etc etc are all completely modular and can be swapped and combined easily for a quick working config.

**The problem**
Currently stepper drivers are defined in sections named after the type of driver. That means making stepper drivers modular via different includes is not possible since old definitions cannot be overriden. Many configs need to override driver specific settings such as run_current and hold_current (in case of different steppers, extruders etc). We also have a "performance" include that disables stealthchop, maximizes speed and acceleration limits and adjusts stepper current.

**Solution 1: Conditional includes**
This would be an extremely powerful feature. I'm imagining attaching a condition with jinja to a [include file.cfg] section. Something like this:
```properties
[gcode_macro globalconfig]
variable_tmc2209: True
...

[include tmc2209.cfg condition={printer["gcode_macro globalconfig"].tmc2209|lower == true}]
```

**Solution 2: Unified driver section**
This is probably less clean as it would require a new section type which includes all possible driver parameters in one.
```properties
[driver stepper_x]
type: tmc2209
...

[driver stepper_y]
type: tmc5160
...
```

Flapstaart:
As an Xchange owner I am looking for a system that would allow me to use non-Gcode input to determine what tool plate I attach to my x-carriage and select/modify configuration accordingly. I can't rely on Gcode because it requires the re-use of all pins and switching to different a type of thermistor can throw the MCU into a panic because the temperature would be far out of range. When the Klipper host detects a change in tool plate ID providing a valid ID different from the config in use, it would then apply the configuration file associated with said tool plate and restart firmware to apply changes.

All kinds of input could be used here like a hardware rotary selector, color sensor, perhaps HAL effect, but ideally RFID.

KevinOConnor:
Interesting, thanks.  I'm not really sure what the state of this PR is.  In general, I can see the utility of some conditional support for config file including.  However, it's not clear to me from just this draft code snippet what the pros/cons of this implementation are.  It might be worthwhile to move this to a discussion on the Klipper Discourse server.

-Kevin

PR #5022, opened 2021-12-14
config: Add BigTreeTech SKR Pico V1.0 board

bigtreetech:


KevinOConnor:
Thanks.  I committed this change with some minor changes on top (commit 634ae08d).

-Kevin

PR #5038, opened 2021-12-18
lib: Added cmsis libraries for stm32g0

bigtreetech:
This library file is copied from STM32 Official library without any modification.


KevinOConnor:
Thanks.  This should be resolved with the merge of #5049.

-Kevin

PR #5039, opened 2021-12-18
stm32: Add initial support for stm32g0

bigtreetech:
Initialization support for new MCU `STM32G0B1` (USBFS support is not included because the USB register address of STM32G0 is 16 bits wide, but can only be read or written 32 bits.  It can only be assigned by `a &= b << 16`, can not by force conversion bits wide `(uint16_t *) a = b`. Therefore, it is quite different from the existing USBFS source code, which needs to be further sorted and submitted to PR separately)

### DIFF
--- a/scripts/flash_usb.py
+++ b/scripts/flash_usb.py
@@ -244,7 +244,7 @@ MCUTYPES = {
     'sam3': flash_atsam3, 'sam4': flash_atsam4, 'samd': flash_atsamd,
     'lpc176': flash_lpc176x, 'stm32f103': flash_stm32f1,
     'stm32f4': flash_stm32f4, 'stm32f042': flash_stm32f4,
-    'stm32f072': flash_stm32f4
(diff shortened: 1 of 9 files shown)


bigtreetech:
* The USB DFU feature can be updated normally. However, an error of `"Error running dfu-util"` will be reported after the update is completed. Further debugging is required.
* `flash-sdcard` feature has also been tested and is normal

KevinOConnor:
Thanks.  I have also been working on stm32g0 support.  I've uploaded my changes to PR #5049.  I was able to refactor the usbfs.c code so that it supports stm32f0, stm32f103, and stm32g0.

Let me know if you have any comments on the implementation in that PR, or if you see any problems with it.

Thanks again,
-Kevin

KevinOConnor:
Thanks. This should be resolved with the merge of #5049.

-Kevin

PR #5040, opened 2021-12-18
config: Add BigTreeTech SKR mini E3 V3.0 board

bigtreetech:


KevinOConnor:
Thanks.  I committed this change (along with 9f31a35e on top).

-Kevin

PR #5041, opened 2021-12-18
stm32: Add USBFS support for stm32g0

bigtreetech:


KevinOConnor:
Thanks. This should be resolved with the merge of #5049.

-Kevin

PR #5049, opened 2021-12-20
Add support for stm32g0 boards

KevinOConnor:
This adds support for the stm32g0b1 chip (as found on the BigTreeTech skr-mini-e3-v3).

In addition to the raw stm32g0 support, there was some refactoring of the stm32 code.

Performance on the new chip looks pretty good.  It's almost as fast as the stm32f103 chips.  (Even though it's clocked at 64Mhz instead of the f103's 72Mhz).

-Kevin


src/stm32/stm32f4.c
```diff
@@ -252,16 +209,67 @@ clock_setup(void)
...
+    uint32_t *sysbase = (uint32_t*)0x1fff0000;
+    asm volatile("mov sp, %0\n bx %1"
+                 : : "r"(sysbase[0]), "r"(sysbase[1]));
+}
+
+// Handle USB reboot requests
+void
+usb_request_bootloader(void)
+{
+    if (CONFIG_STM32_FLASH_START_4000)
```
wlhlm, line 258: I think this kind of build configuration (selection of bootloader type) makes more sense in Kconfig.

KevinOConnor: Thanks.  The code changes here are just code movement - there should be no change in behavior.  The intent was to group the existing bootloader support code into one area of the file and make it's current behavior more clear.

-Kevin

wlhlm: Oh, I've missed that. I see now that this behavior was already there. Sorry for the noise.

src/stm32/spi.c
```diff
@@ -21,7 +21,7 @@ DECL_ENUMERATION("spi_bus", "spi1", 1);
 DECL_CONSTANT_STR("BUS_PINS_spi1", "PA6,PA7,PA5");
 DECL_ENUMERATION("spi_bus", "spi1a", 2);
 DECL_CONSTANT_STR("BUS_PINS_spi1a", "PB4,PB5,PB3");
-#if CONFIG_MACH_STM32F0 || CONFIG_MACH_STM32F2 || CONFIG_MACH_STM32F4
+#if !CONFIG_MACH_STM32F1
```
adelyser, line 24: Couldn't you cover all processors, new and current with this instead:
#ifdef SPI2

KevinOConnor: Alas, no.  The stm32f1 has SPI2, it just doesn't have ability to route it to pins PC2,PC3,PB10.

-Kevin

adelyser: Ah, you're right.  Although, I have to wonder if it really needs to be switched out. In the same manner that another grouping of SPI pins may not map out on a certain board yet are enabled, if someone setting up a STM32F1 board never enables them because they aren't mapped in the schematic or can't be used, would it be considered the same thing?

KevinOConnor: The stm32f1 has a goofy alternative mapping system, so I don't really know what would happen if a user tried to enable spi2a on it.  (In contrast, on the other stm32 chips, if a user enabled a bus that didn't exist, it just wouldn't work - but should not cause internal conflicts.)  I don't think the #ifdef is super critical, but I don't see a harm in it either.

-Kevin

PR #5053, opened 2021-12-21
toolhead: Fix toolhead stop on SET_VELOCITY_LIMIT

RealDeuce:
This fixes issue #5050

Signed-off-by: Stephen Hurd <deuce@synchro.net>

### DIFF
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -552,7 +552,6 @@ class ToolHead:
         self.wait_moves()
     cmd_SET_VELOCITY_LIMIT_help = "Set printer velocity limits"
     def cmd_SET_VELOCITY_LIMIT(self, gcmd):
-        print_time = self.get_last_move_time()
         max_velocity = gcmd.get_float('VELOCITY', None, above=0.)
         max_accel = gcmd.get_float('ACCEL', None, above=0.)
         square_corner_velocity = gcmd.get_float(
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -567,18 +567,17 @@ class ToolHead:
         if requested_accel_to_decel is not None:
             self.requested_accel_to_decel = requested_accel_to_decel
         self._calc_junction_deviation()
-        msg = ("max_velocity: %.6f\n"
-               "max_accel: %.6f\n"
-               "max_accel_to_decel: %.6f\n"
-               "square_corner_velocity: %.6f" % (
-                   self.max_velocity, self.max_accel,
-                   self.requested_accel_to_decel,
-                   self.square_corner_velocity))
-        self.printer.set_rollover_info("toolhead", "toolhead: %s" % (msg,))
         if (max_velocity is None and
             max_accel is None and
             square_corner_velocity is None and
             requested_accel_to_decel is None):
+            msg = ("max_velocity: %.6f\n"
+                   "max_accel: %.6f\n"
+                   "max_accel_to_decel: %.6f\n"
+                   "square_corner_velocity: %.6f" % (
+                       self.max_velocity, self.max_accel,
+                       self.requested_accel_to_decel,
+                       self.square_corner_velocity))
             gcmd.respond_info(msg, log=False)
     def cmd_M204(self, gcmd):
         # Use S for accel
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -567,17 +567,18 @@ class ToolHead:
         if requested_accel_to_decel is not None:
             self.requested_accel_to_decel = requested_accel_to_decel
         self._calc_junction_deviation()
+        msg = ("max_velocity: %.6f\n"
+               "max_accel: %.6f\n"
+               "max_accel_to_decel: %.6f\n"
+               "square_corner_velocity: %.6f" % (
+                   self.max_velocity, self.max_accel,
+                   self.requested_accel_to_decel,
+                   self.square_corner_velocity))
+        self.printer.set_rollover_info("toolhead", "toolhead: %s" % (msg,))
         if (max_velocity is None and
             max_accel is None and
             square_corner_velocity is None and
             requested_accel_to_decel is None):
-            msg = ("max_velocity: %.6f\n"
-                   "max_accel: %.6f\n"
-                   "max_accel_to_decel: %.6f\n"
-                   "square_corner_velocity: %.6f" % (
-                       self.max_velocity, self.max_accel,
-                       self.requested_accel_to_decel,
-                       self.square_corner_velocity))
             gcmd.respond_info(msg, log=False)
     def cmd_M204(self, gcmd):
         # Use S for accel


dmbutyugin (approved):
I agree that calling `get_last_move_time()` appears to be unnecessary in `SET_VELOCITY_LIMIT`. Besides, `M204` does not use that either.

RealDeuce:
When printing a fairly complex set of parts with this patch, I'm getting shutdowns due to timer too close (log attached).  I notice there's a *large* number of log entries regarding the new velocity leading up to the shutdown and am wondering if that's causing the SD card to overload and triggering the shutdown.  Regardless, I don't see why the log needs to be filled with these messages when dynamic acceleration is being used.

[klippy.log.gz](https://github.com/Klipper3d/klipper/files/7757227/klippy.log.gz)

dmbutyugin:
> When printing a fairly complex set of parts with this patch, I'm getting shutdowns due to timer too close (log attached). I notice there's a _large_ number of log entries regarding the new velocity leading up to the shutdown and am wondering if that's causing the SD card to overload and triggering the shutdown. Regardless, I don't see why the log needs to be filled with these messages when dynamic acceleration is being used.
> 
> [klippy.log.gz](https://github.com/Klipper3d/klipper/files/7757227/klippy.log.gz)
> 
> Edit: Host is an RPi 3A+ running FluiddOS

The writes are buffered, so I'm not sure that's what's causing the issues. If you look at your own log, you see that the output from this command is interleaved with other logging, which happens at approximately the same rate. In any case, you could test the second commit on your printer and see if it resolves the timer too close issue? Otherwise, it may be something else that overloads either the host, or the mcu board:
klippy

RealDeuce:
> The writes are buffered, so I'm not sure that's what's causing the issues. If you look at your own log, you see that the output from this command is interleaved with other logging, which happens at approximately the same rate.

Yes, generally the stats were at approximately the same rate, but on some occasions (such as just before the stop), there are a lot more accel changes than stats.

> In any case, you could test the second commit on your printer and see if it resolves the timer too close issue? Otherwise, it may be something else that overloads either the host, or the mcu board: klippy

Yes, I'm printing the same gcode now and will update the PR when it's done (one way or the other).  I think that regardless of the result it shouldn't go into the debug log every time though... I suspect the logging to the debug log is a holdover from the initial development of the command, not something that's especially interesting while debugging (or I would expect M204 to also log).

RealDeuce:
Aaaand, I forgot to put a Signed-off-by on the second commit.  I'll redo the PR when the print completes, sorry.

RealDeuce:
With the second patch, I get this: 
after

Log: 
[klippy-after.log.gz](https://github.com/Klipper3d/klipper/files/7761228/klippy-after.log.gz)

And the print completes successfully.

RealDeuce:
And the graph showing the start of both prints...
start-s

RealDeuce:
Did an amend and force-push instead of a new PR.

dmbutyugin:
OK, thanks! I do not have objections myself. But I will defer the decision to Kevin whether to the commit the change to reduce logging.

dmbutyugin (approved):

KevinOConnor:
Unfortunately, this change (at least as it is currently implemented) could result in a regression.  The call to `self.get_last_move_time()` is intentional so that the lookahead queue is flushed.  Some variables of the toolhead (eg, `toolhead.junction_deviation`) are inspected by the lookahead code and if a change is made to these variables without flushing the lookahead code then the change could be made out-of-order with the requested command and may lead to an inconsistent internal state.

Also, I think we do want to update the log on any fundamental change to the motion kinematics as that would be important diagnostic information to have should one need to perform an investigation.

Separately, FWIW, as mentioned in #4477, it's not clear to me what the benefit of changing acceleration is mid-print.  (Yes, I know slicers can be configured to do it and I know lots of people configure the slicers this way, but I haven't seen evidence showing an actual quality improvement resulting from it.)  So, just to be aware, making it "easier to do something goofy" isn't really a priority for me.  If someone has evidence to the contrary, I'd certainly like to see it.

Cheers,
-Kevin

KevinOConnor:
On a further look, I was incorrect and there is not a race with `toolhead.junction_deviation` and I don't see that it is necessary to flush the lookahead queue.  (And the `M204` command does not flush the lookahead queue.)  I suspect the flush was a leftover from some previous version of the code that did need flushing.

That said, I would not wish to eliminate the logging (at least at this time).

-Kevin

dmbutyugin:
My brief inspection also showed that the parameters are captured within each move upon adding that move into the queue, so flushing _appears_ to be unnecessary. Looking at the git blame, it looks like that call was introduced from pretty much in the first version of `SET_VELOCITY_LIMIT` command (bdd26e421b8d3a8fb902f14f2ab4f787f248c4a5), but perhaps it served a purpose back then.

As for changing the acceleration mid-print, I can imagine that the acceleration could be increased for non-printing moves, in the attempt to reduce stringing (when the tip of the nozzle quickly moves after retract, it can 'break' the string), or to reduce the print times (say, because on non-print moves, smoothing in input shaping does not matter too much, as well as any remaining ringing). How much benefit it actually brings - I do not know, I personally use the same acceleration settings.

RealDeuce:
Log elimination reverted.

RealDeuce:
> As for changing the acceleration mid-print, I can imagine that the acceleration could be increased for non-printing moves, in the attempt to reduce stringing (when the tip of the nozzle quickly moves after retract, it can 'break' the string), or to reduce the print times (say, because on non-print moves, smoothing in input shaping does not matter too much, as well as any remaining ringing). How much benefit it actually brings - I do not know, I personally use the same acceleration settings.

It makes a pretty big difference to use a higher acceleration for travel and infill, then lower the acceleration to prevent ringing in perimeters.  On a full plate with many small objects with line infill, it shaved about 20% off the print time.  It's important that the acceleration be changed during the travel move though or the transition to extruding gets performed at the travel acceleration rate, which brings back the ringing.

KevinOConnor:
Thanks.  I committed this (reduced) change.

-Kevin

KevinOConnor:
> It makes a pretty big difference to use a higher acceleration for travel and infill, then lower the acceleration to prevent ringing in perimeters. On a full plate with many small objects with line infill, it shaved about 20% off the print time.

Thanks, that certainly sounds like a useful test and a measurable quality improvement.  Would you be able to provide some before and after pictures showing the difference?  (Along with the g-code and klipper logs.)

-Kevin

dmbutyugin:
> It makes a pretty big difference to use a higher acceleration for travel and infill, then lower the acceleration to prevent ringing in perimeters. On a full plate with many small objects with line infill, it shaved about 20% off the print time.

YMMV, but large acceleration values on infill may either make the infill appear through walls (if there is insufficient number of perimeters) or conversely, increase smoothing in infill with input shaping and create gaps between infill and perimeters, where infill would normally touch the perimeters.

So, if you perform some tests as Kevin suggested, it would be really great if you could look into this too and test and show how the infill looks like near the perimeters.

> It's important that the acceleration be changed during the travel move though or the transition to extruding gets performed at the travel acceleration rate, which brings back the ringing.

If a travel is a single move, I'm not sure how it can be achieved. Not unless the slicer splits that move further. On the other hand, a travel is oftentimes followed by unretraction, which gives a bit of time to settle the ringing.

RealDeuce:
> If a travel is a single move, I'm not sure how it can be achieved. Not unless the slicer splits that move further. On the other hand, a travel is oftentimes followed by unretraction, which gives a bit of time to settle the ringing.

SuperSlicer splits the move and does the acceleration change in the middle of the travel... and I translate M204 to SET_VELOCITY_LIMIT with a macro so I can keep ACCEL_TO_DECEL at half, which is actually why I started looking into this.

> YMMV, but large acceleration values on infill may either make the infill appear through walls (if there is insufficient number of perimeters) or conversely, increase smoothing in infill with input shaping and create gaps between infill and perimeters, where infill would normally touch the perimeters.

I haven't noticed an issue, but I've only used the feature on prints I want to be very fast (2 perimeters, very low line infill) and don't require strength.  I suspect line infill will be more resistant to the issue than more complex infill patterns, and since the infill is only being used as supports for later layers, as long as they're in place long enough to act as supports, they're good enough for my needs.

I'll try to find time to do the test in the next couple week or two, the printer I use this on is currently apart more often than it's printing right now due to testing some new parts.

PR #5056, opened 2021-12-22
Corrected index document

comradekingu:


dingyifei:
> @dingyifei How could it not be? Getting the description of the software wrong is not a good look. [Klipper3d/klipper#5056](https://github.com/Klipper3d/klipper/pull/5056)

Since the other PR was closed, I'm going to comment here. Ultimately, @KevinOConnor make the call. I'm not a native English speaker, so I might get something wrong here.

---------------

**Using a different dialect isn't wrong.** There are various English dialects, so I don't see the need for dialect-only PR. For example, I'm using a different dialect, and my opinion can be right or wrong regardless of my dialect.

For example, a PR with `Klipper is a 3d-Printer firmware` -> `Klipper is a 3D-printer firmware` don't serve a purpose. Some articles use `3d Printer Firmware`, `3D Printer Firmware`, etc. They are all reasonable ways to address the same thing. This kind of PRs only changes the dialect.

Alternatively, terms like `LaTeX` `GitHub` are probably worth correcting cuz they address specific subjects. 

More active voice and more unambiguous wording are certainly improvements; hence if your PR is a rewrite with these features, it's probably worth accepting.

README.md
```diff
@@ -4,13 +4,11 @@ Welcome to the Klipper project!
 
 https://www.klipper3d.org/
 
-Klipper is a 3d-Printer firmware. It combines the power of a general
-purpose computer with one or more micro-controllers. See the
-[features document](https://www.klipper3d.org/Features.html) for more
-information on why you should use Klipper.
+Klipper is a 3D-printer firmware combining the power of a general \
```
comradekingu, line 7:
```suggestion
Klipper is 3D-printer firmware combining the power of a general \
```

docs/index.md
```diff
@@ -6,12 +6,11 @@ title: Welcome
 
 ![](img/klipper-logo.png){ .center-image }
 
-Klipper is a 3d-Printer firmware. It combines the power of a general
-purpose computer with one or more micro-controllers. See the
-[features](Features.md) document for more information on why you
-should use Klipper.
+Klipper is a 3D-printer firmware combining the power of a general
```
comradekingu, line 9:
```suggestion
Klipper is 3D-printer firmware combining the power of a general
```

comradekingu:
I am just someone who showed up to help you pro bono.
If the desired style is undeniable errors, I have no problems also calling it a dialect.
That is almost _dialectic_, so why not?

https://en.wikipedia.org/wiki/3D_printing
https://www.dictionary.com/browse/3d
https://www.merriam-webster.com/dictionary/3D
https://dictionary.cambridge.org/dictionary/english/3-d-printer?q=3d https://www.thefreedictionary.com/3d

The relative ease with which all dictionaries discard alternative non-objective realities is only a problem if we also assume people read the same way.
In keeping with two different sets of mutually exclusive logic, one champions consistency and correctness, and like in your example, it is all a matter of taste.
Perceptions belong to those ignorant of the non-facts.

Short of pondering the irony of someone published in "Cultural Studies of Iowa" electing to omit the old norse "þeir" to make the (indistinguishable from satire) "Should Writers Use They Own English?" blogpost, I became convinced this handiwork was penned by someone with the career path of a diversity consultant. Whether I am _absolutely_ correct in that assumption holds no precedence over what fantastic insights language can convey.

Maybe the debt-trap of American "academia" will blossom and produce any result, any day now?
Maybe the other "3d printer firmware" project will change what some call "the English language" itself?
It looks like malware, and it is in the linguistic _style_ of malware. 
Delightfully "3D printer" is also featured on that other website, so no conclusions are possible.
The only other search-result is SourceForge, which was in the business of shipping _only_ malware at one point, and then Klipper/[klipper](https://hosted.weblate.org/translate/klipper/rpi_microcontroller/en/?checksum=595a689885382d92) itself.
Why make all this effort to not look like malware? Malware is maligned ware, we must rectify this.
Perhaps users will find solace in the tremendous integrity it takes to opt to do so on both the "y-axis" and "Y-axis" in the same corpus, for a concern over style.

dingyifei:
@comradekingu I'm sorry for referring to an article that is potentially offending and controversial. I referred to the article simply want to support my point that your change improves clarity and is not a just repeat things in different dialect.

The "3D Printing" thing I was referring to might've been a bad example. I'm just saying there is no need to make a PR to only a word from one to another when they refer to the same thing and only being used once. My wording can be better.

You need to modify your commit a bit to make it match the requirements. I think you can just squash all these commits and modify the commit message.

KevinOConnor:
I'm not really sure what this PR does.  As a high-level comment, it "looks different" but not necessary "better".  Separately, commits need to have a signed-off-by line containing your full real name ( https://www.klipper3d.org/CONTRIBUTING.html ).

-Kevin

README.md
```diff
@@ -4,13 +4,11 @@ Welcome to the Klipper project!
...
 
-Klipper is a 3d-Printer firmware. It combines the power of a general
-purpose computer with one or more micro-controllers. See the
-[features document](https://www.klipper3d.org/Features.html) for more
-information on why you should use Klipper.
+Klipper is 3D-printer firmware combining the power of a general \
+purpose computer with one or more micro-controllers. \
+Check out the [features](https://www.klipper3d.org/Features.html) document.
 
-To begin using Klipper start by
```
comradekingu, line 12: Pleonasm, "begin", "start"

README.md
```diff
@@ -4,13 +4,11 @@ Welcome to the Klipper project!
 
 https://www.klipper3d.org/
 
-Klipper is a 3d-Printer firmware. It combines the power of a general
```
comradekingu, line 7: 3D as per https://en.wikipedia.org/wiki/3D_printing
"a firmware" is wrong, but it could be a firmware _solution_ etc.

docs/index.md
```diff
@@ -6,12 +6,11 @@ title: Welcome
...
+Klipper is 3D-printer firmware combining the power of a general
+purpose computer with one or more micro-controllers.
+Check out the [features](Features.md) document.
 
-To begin using Klipper start by [installing](Installation.md) it.
+Begin using Klipper by [installing](Installation.md) it.
 
 Klipper is Free Software. Read the [documentation](Overview.md) or
-view [the Klipper code on github](https://github.com/Klipper3d/klipper).
+view [the Klipper code on GitHub](https://github.com/Klipper3d/klipper).
```
comradekingu, line 16: As per the spelling on this site

comradekingu:
@KevinOConnor Annotations added. Feel free to sign off on it, I just did it with something obvious to dip my feet in.

PR #5059, opened 2021-12-22
Update Resonance Docs

mneuhaus:
I had quite some issues trying to get my ADXL345 running.
I checked and double checked everything, got 1 reading but then nothing.
Turns out, the communication is quite susceptible to signal degradation and interference.  After using an old Network cable which is shielded and twisted it worked without a problem. Since quite some people seem to be having similar issues i thought it might make sense to point this out in the docs.

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -37,9 +37,15 @@ Fritzing wiring diagrams for some of the ADXL345 boards:
 
 
 Double-check your wiring before powering up the Raspberry Pi to prevent
-damaging it or the accelerometer. Try to keep the cables as short as possible, 
-or use something shielded/twisted like a network cable to prevent communication
+damaging it or the accelerometer. Try to keep the cables as short as possible
+and/or use a **shielded** ethernet wire to prevent communication
 issues because of signal interference an degradation.
+For best singnal transmission we recommend the following pairing of wires:
+
+* GND on Shield on the mainboard side
+* One pair for GND+MISO (SDO)
+* One pair for 3.3V+MOSI (SDA)
+* One pair for SCLK (SCL)+CS
 
 ### Mounting the accelerometer
 
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -37,7 +37,9 @@ Fritzing wiring diagrams for some of the ADXL345 boards:
 
 
 Double-check your wiring before powering up the Raspberry Pi to prevent
-damaging it or the accelerometer.
+damaging it or the accelerometer. Try to keep the cables as short as possible, 
+or use something shielded/twisted like a network cable to prevent communication
+issues because of signal interference an degradation.
 
 ### Mounting the accelerometer
 


docs/Measuring_Resonances.md
```diff
@@ -37,7 +37,9 @@ Fritzing wiring diagrams for some of the ADXL345 boards:
 
 
 Double-check your wiring before powering up the Raspberry Pi to prevent
-damaging it or the accelerometer.
+damaging it or the accelerometer. Try to keep the cables as short as possible, 
+or use something shielded/twisted like a network cable to prevent communication
+issues because of signal interference an degradation.
```
Sineos, line 42: > because of signal interference an degradation.

Should be **and** or **and / or**

Sineos:
I made the same experience. Above 20cm regular wires and it stopped working. Also used a salvaged Cat 5E cable.

Please make sure to follow https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md, especially the signed-of statement

dmbutyugin:
Thank you! Additionally, I think twisted cable does not do much - only accidentally it may fix things (SPI protocol does not make use of a twisted pair), but a shielded cable could indeed make a difference.
Also, to add to what Sineos has said about the signed-off-by line, it seems that the Build test has failed, you can check the error message - it seems that you need to remove the trailing whitespace on the line.

Sineos:
> I think twisted cable does not do much

Not sure about this as well, but a shield without GND is not worth much as well (and I did not ground mine)
What could be mentioned as well is that twisting should always happen against a "stable" counterpart. I used:
- VCC / MISO
- GND / CLK
- CS / MOSI

dmbutyugin:
> > I think twisted cable does not do much
> 
> Not sure about this as well, but a shield without GND is not worth much as well (and I did not ground mine).

I'm not sure about that - it may still deflect some EMF interference. But I'm not an expert in these kind of things.

> What could be mentioned as well is that twisting should always happen against a "stable" counterpart. I used:
> 
>     * VCC / MISO
> 
>     * GND / CLK
> 
>     * CS / MOSI

That's an interesting idea. Note that CS is not really stable: it goes down in the beginning of communication, and errors in CS signal transmission will result in the wrong reads/writes. Advancing this idea, I could even suggest to use GND as a counterpart for all wires (you do need more wires that way though). This is because all signals are measured against GND level, so if GND is passed along, any external interference will, hopefully, affect both wires in the same manner, reducing or eliminating the noise. I haven't tried that myself though.

Sineos:
> That's an interesting idea. Note that CS is not really stable: it goes down in the beginning of communication, and errors in CS signal transmission will result in the wrong reads/writes. Advancing this idea, I could even suggest to use GND as a counterpart for all wires

This is how I learned it a very long time ago. Either you have symmetrical lines then twist these or twist against a stable line or use coaxial cables with GND as return line. Of course the CS / MOSI combination is kind of a band-aid but even with the 4 pairs in a patch cable you could not twist CS and MOSI with GND.
From the same lesson comes the information that a shielding without ground is basically useless as the energy cannot drain, whereas shielding against electrical fields requires one sided ground and against magnetic fields even both sides (leading to unwanted effects like ground loops). The hell of EMI 🤷

susisstrolch:
Imho the critical signals are CLK and CS. 
MISO/MOSI are half duplex, so no need to give them a GND counterpart.
So with a CAT5|6 cable I'd suppose

- GND / Vcc
- MISO/MOSI
- GND/CLK
- GND/CS
and the shield connected to GND on the power side (MCU).

mneuhaus:
I'm far from an expert on cables, i can update the wording however you guys like.
It was just my experience and from a search on discord, etc. plenty of other people have similar issues and going for a shielded/twisted/better cable helped instantly.

Sineos:
> Imho the critical signals are CLK and CS.

@susisstrolch: Can you elaborate on this? CLK, I agree, but CS is pulled to GND at start and stays there. Being on GND potential anyway, this should be one of the most resilient data lines.

susisstrolch:
> Being on GND potential anyway, this should be one of the most resilient data lines.
That's true for the driving side, but not for the input side, which has a higher impedance and also a complex resistance (the wire) between.
And - from this point of view - CS is as critical as CLK.
CLK may lead to phantom bits, means reading wrong registers. CS would invalidate the current SPI session.
So, imho, both signals should be wired against GND.

Sineos:
> So, imho, both signals should be twisted against GND.

Ideally, yes. Personally I do not like the MISO/MOSI pair as these are non symmetrical lines and effectively my wanted data at 3.2 kHz (at least MISO).
Anyway, I guess this is kind of a theoretical discussion. @mneuhaus cable is working without thinking too much about the pairing, my cable works as well. So anything shielded, twisted and good copper is better than these tinned cheapo (jumper) cables

Drachenkaetzchen:
We give out the following recommendations on the Voron discord on how to best wire the shielded(!) ethernet wire:

- GND on Shield on the mainboard side
- One pair for GND+MISO (SDO)
- One pair for 3.3V+MOSI (SDA)
- One pair for SCLK (SCL)+CS

Note that GND+SCLK / GND+CS on two separate pairs would work as well, but harder to wire on the ADXL side, that's why we stick with SCLK+CS. Wire lengths are usually in the order of 1-1.5m.

Up until now it worked every time, given that the sensor was previously verified working using a short cable. In my experience, the most important part is actually putting GND on the shield, but YMMV.

To do it properly™️ we would need to go down the whole rabbit hole, which includes line impedance, signal reflections and transmission line theory, which is probably not needed.

Sineos:
> * GND on Shield on the mainboard side
> * One pair for GND+MISO (SDO)
> * One pair for 3.3V+MOSI (SDA)
> * One pair for SCLK (SCL)+CS

Sounds good for me as well and is basically just the same as my reccomendation:
* GND, VCC and CS considered as "stable"
* SCLK, MISO, MOSI paired with the stable ones

In which order you do this, probably makes no difference

KevinOConnor:
I'm not sure what the state of this PR is.  At a minimum it would need a signed-off-by line ( https://www.klipper3d.org/CONTRIBUTING.html ) and the additional docs need a spell-check.  Otherwise, I think it is up to @dmbutyugin to decide.

-Kevin

PR #5060, opened 2021-12-23
Create qwiic_dual_ssr_relay.py

parsko41:
Hi, I am very much new to this.  I have this working library I want to contribute.  Please take a look.  It works.  I want to also provide a config file for example settings.  I am worried that this might step on Sparkfun's toes, and I have contributed it to their repository too.   Please forgive me for this, and I hope to resolve it as time moves on.  Please have a look and let me know if this makes any sense.
-Parsko

### DIFF
--- a/qwiic_dual_ssr_relay.py
+++ b/qwiic_dual_ssr_relay.py
@@ -1,0 +1,197 @@
+# qwiic_dual_ssr_relay i2c based two Solid State Relay
+#
+# Copyright (C) 2021 Luke Parker <parsko@gmail.com>
+#
+#   Author(s): Gaston Williams
+#   Luke stole much code from Gaston!  He get's all credit.
+#   Please see:
+#   https://www.sparkfun.com/products/15093
+#   and
+#   https://github.com/adafruit/circuitpython/releases
+#   and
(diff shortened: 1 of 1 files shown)


Sineos:
Wow, 150 USD for a SSR that even exposes mains voltage due to lack of housing 

Anyway, thanks for your contribution. With your different contributions to Sparkfun and here, I do not think it is an issue as long as licenses match and proper copyright is given.

Please also follow https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md, especially the signed-off line and the `Format of commit messages` section

Sineos:
@parsko41 without trying to be a PITA, please check again above linked document:
- Commit messages should provide a short description of the change
- Comments like you have in your commit message should be posted like my comment here. Please see https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#format-of-commit-messages for an example. Especially mind the signed-off line
- The new functions need documentation. Please see https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review Point 4
- Your config examples will not be found by any user as they are not looking into the module code. Should be removed there and usage explained in the documentation.

PR #5061, opened 2021-12-23
CFG Elegoo Neptune 2S V1.3 with BLTouch and Mesh

oinosme:
This adds the configuration file for the Elegoo Neptune 2S on V1.3 with BLTouch and mesh enabled.

### DIFF
--- a/config/printer-elegoo-neptune2s-v1.3-bltouch.cfg
+++ b/config/printer-elegoo-neptune2s-v1.3-bltouch.cfg
@@ -1,0 +1,164 @@
+# This configuration file seems to work for the Elegoo Neptune 2S
+# with ZNP Robin Nano V1.3.
+# To configure select "extra low-level configuration setup"
+# STM32F407 32 KiB Bootloader and serial on # USART3
+# PB11/PB10 communication.
+# After running make, copy the klipper.bin file and rename as
+# elegoo.bin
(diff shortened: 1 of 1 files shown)


just-trey:
@oinosme you may be interested in my pr. It covers both the 1.2 and 1.3 board. https://github.com/Klipper3d/klipper/pull/5213

oinosme:
@just-trey thanks!

KevinOConnor:
I'm not sure what the state of this PR is - at a minimum it would need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html and https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

Fisheiyy (changes requested):
I understand the point of this PR is to add a BLTouch config for the Neptune 2S V1.3 but unless it comes stock on the printer there is no point in adding a whole config for the printer, and this config is also very messy with lots of comments and the whole auto generated stuff at the bottom of the config which could mess up other peoples printers if they didnt know about it since it contains a bed mesh that isnt theirs and a z offset that isnt theirs.

just-trey:
@Fisheiyy  @KevinOConnor  I appreciate the effort here but I believe the pr I submitted that was approved #5213 is sufficient and this config is not needed as I have comments that allow configurations for both board versions. @oinosme would you agree to close this PR?

Issue #5062, opened 2021-12-23
Typo in "Pressure_Advance.md"


tispokes:
is: `pressure_advance = <calculated_value>`
should: `pressure_advance: <calculated_value>`

Right before the "Important Notes" section starts

Sineos:
Both notations are valid, but you are right, the preferred one is `key: value`
I suggest you open a PR as correction.

tispokes:
Again what learned.

PR #5063, opened 2021-12-23
Updated Pressure_Advance.md to preferred style

tispokes:
PR as @Sineos said on Issue: https://github.com/Klipper3d/klipper/issues/5062


KevinOConnor:
Thanks.  I'm not sure this change adds much though - I don't think the existing text is confusing.

-Kevin

PR #5065, opened 2021-12-25
Unify enable_pclock() code on stm32

KevinOConnor:
Unify the handling of the enable_pclock() and is_enabled_pclock() code
across all stm32 chips.  All chips will now perform a peripheral reset
on enable_pclock() (this is a change for stm32f0 and stm32h7).  The
enable_pclock() code will now also disable irqs during the enable.

It is hoped this will make the code a little simpler and a little easier to maintain.

-Kevin


Arksine:
Nice, looks good.  Should resolve any potential bootloader issues across all stm32 devices.

KevinOConnor:
Thanks for reviewing.  I committed this change.

-Kevin

PR #5066, opened 2021-12-25
sr595: Add support for shift register

hmortensen83:
sr595: Add support for shift register

Signed-off-by: Heino Mortensen <heino@no-mercy.dk>


KevinOConnor:
Thanks.  I'm not sure what the state of this PR is.  It currently has a merge conflict.

I'm not sure I understand the main goal of the implementation.  It seems the code implements both a g-code command layer and a "pin abstraction" layer.  However, the "pin abstraction" layer seems incomplete - as many of the main functions aren't implemented.  I'd guess a "shift register" implementation would want to focus on a "pin abstraction".

-Kevin

PR #5069, opened 2021-12-25
Update generic-mks-robin-nano-v1.cfg

karlla1220:
Since the config name is containing v1, many of users are using this config for MKS robin Nano v1.3 as well. However, Nano V1.3 board requires different compile options and python scripts shouldn't be used. Since V1.3 board still uses same pinout (maybe), it may be reasonable to just add description for V1.3 rather than add another config file.

### DIFF
--- a/config/generic-mks-robin-nano-v1.cfg
+++ b/config/generic-mks-robin-nano-v1.cfg
@@ -4,11 +4,20 @@
 # configuration setup", select the 28KiB bootloader, and serial (on
 # USART3 PB11/PB10) communication.
 
+# For MKS Robin Nano V1.3 and MKS Robin Nano-S V1.3 boards, the
+# firmware should be compiled for the STM32F1407. When running 
+# "make menuconfig", enable "extra low-level configuration setup", 
+# select the 32KiB bootloader, and serial (on USART3 PB11/PB10) 
+# communication.
+
 # Note that the "make flash" command does not work with MKS Robin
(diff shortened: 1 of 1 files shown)


config/generic-mks-robin-nano-v1.cfg
```diff
@@ -4,11 +4,20 @@
 # configuration setup", select the 28KiB bootloader, and serial (on
 # USART3 PB11/PB10) communication.
 
+# For MKS Robin Nano V1.3 and MKS Robin Nano-S V1.3 boards, the
+# firmware should be compiled for the STM32F1407. When running
```
adelyser, line 8: Typo: perhaps you mean STM32F407

adelyser:
Several comments:

1. You will have to fix the pipeline failures.
2. I would suggest creating a generic-mks-robin-nano-v1.3.cfg to clearly define the differences in building and flashing methods for that specific board.
3. Your commit message needs to be changed to meet the required format as outlined in the [contributing ](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)documentation.

config/generic-mks-robin-nano-v1.cfg
karlla1220: Thanks for the comment. It supposed to be STM32F407. I will re-request.

karlla1220:
Thanks for the comment! I will change my commit based on given comments!

PR #5070, opened 2021-12-26
docs: Address Names In Status Reference with \< And \>

dingyifei:
Signed-off-by: Yifei Ding <yifeiding@protonmail.com>

There are many ways to address the name to reference an object like "some name", "config name", "stepper name", "led name" throughout the documentations. The second commit change all of them to config_name.

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -62,8 +62,8 @@ The following information is available in the
 
 The following information is available in
 [fan](Config_Reference.md#fan),
-[heater_fan some_name](Config_Reference.md#heater_fan) and
-[controller_fan some_name](Config_Reference.md#controller_fan)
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, I think this may introduce some subtle confusion to the reference documentation.

If there is a config section such as `[temperature_sensor some_name]` then sometimes we need to access it using its "config name" (`temperature_sensor some_name`) and sometimes we need to access it using it's "name" (`some_name`).  The documentation is making a distinction because some times there is a distinction.

I'm not sure about using the "< angle brackets >" within a literal description because it may lead to confusion.  For example, is it actually `[dotstart foo]` or is actually `[dotstart <foo>]`.

-Kevin

Issue #4963, opened 2021-11-25
Issue with the Documentation Search Function on https://www.klipper3d.org/


Sineos:
Browser:
* Firefox 94.0.1
* Chrome  95.0.4638.69

https://www.klipper3d.org/Rotation_Distance.html

![grafik](https://user-images.githubusercontent.com/4459312/143442320-c06f4747-908b-4192-8517-747d75556112.png)

When going to the same Chapter via the search function and highlighting:

https://www.klipper3d.org/Rotation_Distance.html?h=rotat

![grafik](https://user-images.githubusercontent.com/4459312/143444026-55e4b208-fcab-438d-a96c-277820c2681c.png)

KevinOConnor:
I'd guess this is a defect in mkdocs.  I'm not sure how we'd be able to fix this in the Klipper build itself.

-Kevin

Sineos:
https://github.com/squidfunk/mkdocs-material/issues/2068

Seems there is a fix available. Maybe @dingyifei could cross check

KevinOConnor:
Okay.  It's possible to update the mkdocs-material version (it is specified in docs/_klipper3d/mkdocs-requirements.txt ).  However, it is unclear to me if this issue was addressed in the latest publicly released version.

-Kevin

KevinOConnor:
It does seem that an upgrade to mkdocs-material v8.1.3 fixes this issue (commit 608cd38d).

Thanks.
-Kevin

Sineos:
Fix confirmed 👍

PR #5072, opened 2021-12-27
Mkdocs i18n support

dingyifei:
TODO:
- [x] Add static i18n plugin
- [x] Modify the pre-mkdocs script when translation is ready
- [x] Verify everything works together


dingyifei:
I verified the script working by running the commands below:

```bash
./scripts/pre-mkdocs.sh
mkdocs serve --config-file docs/_klipper3d/mkdocs.yml
```

The search index is correctly separated between languages and the result looks decent：

image

image

@KevinOConnor

KevinOConnor:
Thanks.  At a high-level, I agree with the change.  I think it would be good to get translations on the main klipper3d.org site.

Can you give a high-level overview of how this will work?  It looks to me like the script will clone the klipper-translations repo and copy some `.md` files?  If so, why the hardcoding for zh in the script (both in the `if` and in the `mv`)?

-Kevin

dingyifei:
> Can you give a high-level overview of how this will work? It looks to me like the script will clone the klipper-translations repo and copy some .md files? If so, why the hardcoding for zh in the script (both in the if and in the mv)?

The weird fixes I added are mainly to address the quirks of Static I18n plugin for MkDocs.

The steps to add a Language:

1. Change i18n plugin settings by modifying mkdocs yml
2. Modify the pre-mkdocs script so the language files are copied.

Zh-hans currently need a hack instead of *.$lang.md because i18n plugin doesn't recognize zh-hans/zh_hans. The workaround here is to replace zh-hans with zh. It is a problem with language code standards, I suspect replacing zh-hans with zh_CN might work, but I haven't tested it.

I didn't script it to move all the languages because it takes more time to run. In the previous commit I made, works for all languages.

dingyifei:
Basically when the workflow is triggered, the following will happen:

1. Pre-mkdocs script clone the translations repo and copy all the md files for languages we want to have with the file names modified to include language code.
2. Regular mkdocs deploy where i18n plug-in detects the language code and use those files for the language the language code represents. If a file is missing, the original md file is used.
3. I18n plug-in change the index so search results are not mixed together.

dingyifei:
Confirmed fix for #4963 works for this PR as well

It seems like mdpo created an incompatible index.zh.md. The pre-mkdocs script probably needs to address this issue in some way.

Original: 

```
---
hide:
  - toc
title: Welcome
---
```

mdpo:
```
***

hide:

- toc title: Welcome

***
```

dingyifei:
I think I can modify the script so it takes string pairs for renaming files. It is possible more languages will require such fix.

dingyifei:
I think it's ready now. The workflow will run every day to update the deployed page. zh-hans has 75% translated

The non-english index page is still broken. It will need a fix in the Klipper-translations PO to MD in the future.

KevinOConnor:
Thanks.  In general it looks fine to me.  Could me make a few changes though?

1. Name the script `fetch-translations.sh` instead of `pre-mkdocs.sh`.
2. Don't use `git submodule`.  Just clone the repo using something like `git clone --depth 1`.
3. How about putting the translation list in the klipper-translations repo and using something like the following in `fetch-translations.sh`:
```bash
...
cat /path/to/klipper-translations/active_translations | while read dirname langname ; do
    ...
    mv "$file" "./docs/${mdfilename//.md/.${langname}.md}"
done
```
The above would assume the klipper-translations repo had a file "active_translations" file with something like:
```
zh-hans zh
```

-Kevin

dingyifei:
> Thanks. In general it looks fine to me. Could me make a few changes though?
> 
>     1. Name the script `fetch-translations.sh` instead of `pre-mkdocs.sh`.
> 
>     2. Don't use `git submodule`.  Just clone the repo using something like `git clone --depth 1`.
> 
>     3. How about putting the translation list in the klipper-translations repo and using something like the following in `fetch-translations.sh`:
> 
> 
> ```shell
> ...
> cat /path/to/klipper-translations/active_translations | while read dirname langname ; do
>     ...
>     mv "$file" "./docs/${mdfilename//.md/.${langname}.md}"
> done
> ```
> 
> The above would assume the klipper-translations repo had a file "active_translations" file with something like:
> 
> ```
> zh-hans zh
> ```
> 
> -Kevin

Sure, it's a better solution than adding additional lines in the deploy workflow.

dingyifei:
One issue with this implementation is the mkdocs.yml would require modification for every language added to the active_translations, otherwise, they won't be displayed.

docs/CONTRIBUTING.md
```diff
@@ -301,15 +301,13 @@ contributions) and contain a current email address.
 [Klipper-translations Project](https://github.com/Klipper3d/klipper-translations)
 is a project dedicated to translating Klipper to different languages.
 [Weblate](https://hosted.weblate.org/projects/klipper/) hosts all the
-Gettext strings for translating and reviewing. Locales can merge into
-the Klipper project once they satisfy the following requirements:
+Gettext strings for translating and reviewing. Locales can be displayed on
+<klipper3d.org> once they satisfy the following requirements:
```
KevinOConnor, line 305: Using angle brackets in markdown (outside of backticks) causes havoc.  Use backticks, bold (via *), or underline (via _) instead.

docs/CONTRIBUTING.md
```diff
@@ -321,15 +319,20 @@ If a translation already exists in the Klipper repository and no
...
+are met, you need to:
 
-```yml
-nav:
-  - existing hierachy
-  - <language>:
-    - locales/<language code>/md file
-```
+1. update klipper repository <mkdocs.yml>
+2. update klipper-tranlations repository <active_translations>
```
KevinOConnor, line 326: Same comment here.

scripts/fetch-translations.sh
```diff
@@ -0,0 +1,16 @@
...
+
+git clone --depth 1 https://github.com/Klipper3d/klipper-translations
+
+cat ./klipper-translations/active_translations | while read dirname langname ; do
+  local_dir=klipper-translations/docs/locales/$dirname
+  echo "Moving $dirname to $langname"
+  for file in $local_dir/*.md; do
+    mdfilename=${file/$local_dir\//}
+    mv $file "./docs/${mdfilename//.md/.${langname}.md}"
+  done
```
KevinOConnor, line 15: Because we're reading `dirname` and `langname` from an external source, we should be careful to wrap all expansions in double quotes to avoid weird things happening if a shell character is in the input.

For example:
```bash
  local_dir="klipper-translations/docs/locales/${dirname}"
  echo "Moving $dirname to $langname"
  for file in "${local_dir}"/*.md; do
    mdfilename="${file/${local_dir}\//}"
    mv "${file}" "./docs/${mdfilename//.md/.${langname}.md}"
  done
```

KevinOConnor (commented):
Thanks.  I have a few minor comments.

As for updating mkdocs.yml - I agree it is not optimal, but I don't know enough about the i18n extension to suggest an alternative.  I'm okay with updating mkdocs.yml on each new language.

-Kevin

dingyifei:
I have resolved the issues and added a new feature to the script so it can replace index.lang.md with manual-index.lang.md since mdpo generated index.md that can't be correctly rendered in MkDocs.

KevinOConnor:
Thanks.  It looks fine to me.  Please confirm when you are ready to make the transition and I will commit.

-Kevin

KevinOConnor:
Separately, we could automate the update of mkdocs.yml with something like the following:
```bash

MKDOCS_CONFIG=./docs/_klipper3d/mkdocs.yml
MKDOCS_MATCH="        # Additional languages get automatically added here."

cat ./klipper-translations/active_translations | while read dirname langname langdesc ; do
    ...
    sed -i "s/^\(${MKDOCS_MATCH}\)$/        ${langname}: ${langdesc}\n\1/" ${MKDOCS_CONFIG}
done
```

And then add the following to mkdocs.yml:
```diff
--- a/docs/_klipper3d/mkdocs.yml
+++ b/docs/_klipper3d/mkdocs.yml
@@ -26,6 +26,13 @@ plugins:
   - exclude:
       glob:
         - README.md
+  - i18n:
+      default_language: en
+      languages:
+        en: English
+        # Additional languages get automatically added here.
 
 # Website layout configuration (using mkdocs-material theme)
 theme:
```

The above would require a third field (the local language name) in the `active_translations` file.

Also, if you do update this patch, the comment at the top of fetch-translations.sh should be updated.  Also, I think it may be better to put this script in docs/_klipper3d/ instead of in scripts/ .

-Kevin

dingyifei:
I just found an alternative solution:

add this line to mkdocs.yml and modify the plugin syntax:
```
INHERIT: translations.yml

plugins:
  search: {}
  mkdocs-simple-hooks:
    hooks:
      on_page_markdown: "docs._klipper3d.mkdocs_hooks:transform"
  exclude:
    glob:
      - README.md
```

add translations.yml to klipper- translations:

```
plugins:
  i18n:
    default_language: en
    languages:
      en: English
      zh: 简体中文
```

add a line to fetch-translations.sh:
```
cp "./klipper-translations/translations.yml" "./docs/_klipper3d"
```

KevinOConnor:
That does simplify things.  Though, I'd prefer not have the mkdocs configuration spread between two repos.  Maybe we could keep a translations.yml file here with:
```
plugins:
  i18n:
    default_language: en
    languages:
      en: English
```

And extend the fetch-translations.sh script to do something like:
```
cat ./klipper-translations/active_translations | while read dirname langname langdesc ; do
    ...
    echo "      ${langname}: ${langdesc}" >> ./docs/_klipper3d/translations.yml
done
```

-Kevin

KevinOConnor:
Oh, another thing we might want to do is change the `cat ... | read` to something like:

```
egrep -v '^ *(#|$)' ./klipper-translations/active_translations | while read dirname langname langdesc ; do
    ...
done
```

That way comment and blank lines are ignored in the active-translations file.

-Kevin

dingyifei:
> Oh, another thing we might want to do is change the `cat ... | read` to something like:
> 
> ```
> egrep -v '^ *(#|$)' ./klipper-translations/active_translations | while read dirname langname langdesc ; do
>     ...
> done
> ```
> 
> That way comment and blank lines are ignored in the active-translations file.
> 
> -Kevin

Done! I changed the active-translations file to CSV so it can contain spaces in the displayed language names.

I think the changes are ready to merge

KevinOConnor:
Thanks.  I committed the change, but it doesn't look like it works correctly.  I fixed up the mkdocs-requirements.txt file (commit c78676ff).  However, it reports an error (`grep: ./klipper-translations/active_translations: No such file or directory`).  Perhaps the `git clone` line wasn't supposed to be commented out?

Also, can we add a '#' to the start of the comment line in the active-translations.csv file in the klipper-translations repo?  Currently the file looks like:
```csv
Directory Name,Language Name,Displayed Language Name,Comments
zh-hans,zh,简体中文,
```
And I think it would be more robust if the file looked like:
```csv
#Directory Name,Language Name,Displayed Language Name,Comments
zh-hans,zh,简体中文,
```

That way, we don't need the `tail` in the fetch-translations.sh script and other unix tools are more likely to recognize the first line as a comment.

-Kevin

dingyifei:
Sorry I didn't realize the # that comment out the git clone was committed when I was testing the changes. 

I'll modify the script and active_translations. 

Should I make another PR to address these changes?

KevinOConnor:
A new PR is best.  I don't mind making the change, but if you submit the PR it helps ensure we have the correct timing between committing changes here and to klipper-translations.

Thanks,
-Kevin

dingyifei:
klipper-translations is ready

https://github.com/Klipper3d/klipper/pull/5263

PR #5073, opened 2021-12-27
Fix markforged kinematics name in config reference

controlflow:
I was getting unrecognized kinematics error until replaced `-` with `_`.


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit directly - see https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md .  However, I did update the docs (commit 0db696b0).

-Kevin

PR #5076, opened 2021-12-29
Update Config_Reference.md

LighttheLastFlare:
Added info on neopixel not being supported on the "linux mcu" implementation of Klipper


KevinOConnor:
Thanks.  The change looks fine, but commits need to have a signed-off-by line containing your full real name - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

KevinOConnor:
A comment was added to the docs in commit 35ff49e5.

-Kevin

PR #5077, opened 2021-12-29
BUGFIX: Fix the GPIO register for stm32h7

adelyser:
https://github.com/Klipper3d/klipper/commit/dc3ac2b424dc2f915f825227ac941ab1b982ed21 breaks the STM32H7 processor GPIO because it's using the incorrect register.

This PR fixes the issue.


KevinOConnor:
Thanks.

-Kevin

PR #5079, opened 2021-12-29
Use hardware SPI for the SKR-PRO

adelyser:
Switch the SKR PRO to use the hardware SPI bus. 

### DIFF
--- a/docs/SDCard_Updates.md
+++ b/docs/SDCard_Updates.md
@@ -127,8 +127,7 @@ and the following additional field should be specified:
   the SD Card in the format of `miso,mosi,sclk`.
 
 It should be exceedingly rare that Software SPI is necessary, typically only
-boards with design errors will require it. The `btt-skr-pro` board definition
-provides an example.
+boards with design errors will require it.
 
 Prior to creating a new board definition one should check to see if an
 existing board definition meets the criteria necessary for the new board.
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -46,8 +46,7 @@ BOARD_DEFS = {
     },
     'btt-skr-pro': {
         'mcu': "stm32f407xx",
-        'spi_bus': "swspi",
-        'spi_pins': "PA6,PB5,PA5",
+        'spi_bus': "spi1b",
         "cs_pin": "PA4"
     },
     'btt-gtr': {
--- a/src/stm32/spi.c
+++ b/src/stm32/spi.c
@@ -21,21 +21,23 @@ DECL_ENUMERATION("spi_bus", "spi1", 1);
 DECL_CONSTANT_STR("BUS_PINS_spi1", "PA6,PA7,PA5");
 DECL_ENUMERATION("spi_bus", "spi1a", 2);
 DECL_CONSTANT_STR("BUS_PINS_spi1a", "PB4,PB5,PB3");
+DECL_ENUMERATION("spi_bus", "spi1b", 3);
+DECL_CONSTANT_STR("BUS_PINS_spi1b", "PA6,PB5,PA5");
 #if !CONFIG_MACH_STM32F1
- DECL_ENUMERATION("spi_bus", "spi2a", 3);
+ DECL_ENUMERATION("spi_bus", "spi2a", 4);
  DECL_CONSTANT_STR("BUS_PINS_spi2a", "PC2,PC3,PB10");
 #endif
 #ifdef SPI3
- DECL_ENUMERATION("spi_bus", "spi3", 4);
+ DECL_ENUMERATION("spi_bus", "spi3", 5);
  DECL_CONSTANT_STR("BUS_PINS_spi3", "PB4,PB5,PB3");
  #if CONFIG_MACH_STM32F4
-  DECL_ENUMERATION("spi_bus", "spi3a", 5);
+  DECL_ENUMERATION("spi_bus", "spi3a", 6);
   DECL_CONSTANT_STR("BUS_PINS_spi3a", "PC11,PC12,PC10");
   #ifdef SPI4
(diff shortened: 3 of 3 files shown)


Arksine:
The fix itself looks good to me, however its probably a good idea to rebase against master (avoid the merge commit) and break this down into 3 commits in the documented format, for example:

```
stm32f4: add support for spi1b alternate

Signed-off-by: ...
```
```
spi_flash: use hardware spi for skr-pro boards

Signed-off-by: ...
```
```
docs: update spi_flash documentation

Signed-off-by: ...
```

adelyser:
> The fix itself looks good to me, however its probably a good idea to rebase against master (avoid the merge commit) and break this down into 3 commits in the documented format, for example:
> 
> ```
> stm32f4: add support for spi1b alternate
> 
> Signed-off-by: ...
> ```
> 
> ```
> spi_flash: use hardware spi for skr-pro boards
> 
> Signed-off-by: ...
> ```
> 
> ```
> docs: update spi_flash documentation
> 
> Signed-off-by: ...
> ```

Thanks for the feedback. Are you suggesting I have 3 commits in the 1 PR, or 3 separate PR's?

Arksine:
Everything looks good, but there is one problem that I have discovered.  When we change the board definition for the skr pro we will break `spi_flash` for users that have not yet upgraded.  I think perhaps the correct approach would be to commit the changes to the stm32 source, then wait a bit before make a change to the board definiton in spi flash.  Software SPI will continue to work.

KevinOConnor:
Thanks.  I think we'll also need to use a different naming convention here as it's not a different set of pins for the spi1 peripheral - it's a mix of pins from the existing "spi1" and "spi1a".

What's the high-level gain of making this change?  Is the bus needed for other purposes or just for spi_flash?  Is there an issue with using software spi with spi_flash?

-Kevin

adelyser:
@Arksine I did think that could be an issue and wasn't sure if it would be better to let it "soak" first. That's probably the best option if we move forward.

@KevinOConnor I'm not sure what the concern is about the naming convention, they are all pins from the SPI1 hardware on the stm32. Depending on the manufacturers layout of the board, they could potentially use any combination of the SPI1 pins in their design. I don't really see it as any different than someone trying to use "spi2" on the stm32h7 if they have "USB (on PB14/PB15)" selected, causing a conflict. 

The big reason I did this was mainly because the docs say to use hardware spi when available, and software spi use should be rare, and it lists the SKR PRO as one of the exceptions. This change mainly shows hardware spi does work on the SKR PRO now, which I discovered while trying some other things during other development. There really isn't an issue with the software spi working right now.

Maybe this sparks a bigger discussion on how the spi busses are defined. There could be a large number of combinations of each SPI bus that could potentially be used in up-coming board designs. I thought there might be a "better way" than hard-defined busses in the firmware that require recompilation. One idea I had is maybe all SPI definitions should be in printer.cfg, just like the Software SPI. Instead, a SPI bus might look something like this:
```
spi_bus: spi1 #could also be swspi when needed
spi_miso_pin: PA6
spi_mosi_pin: PB5
spi_sclk_pin: PA5
spi_function: 5 #This could be defaulted to the normal fuction based on processor, usually 5
```

KevinOConnor:
> Depending on the manufacturers layout of the board, they could potentially use any combination of the SPI1 pins in their design.

True, but in practice the STM docs tend to group the pins together and most boards use them as the groups generally indicated in the docs.

> Maybe this sparks a bigger discussion on how the spi busses are defined. There could be a large number of combinations of each SPI bus that could potentially be used in up-coming board designs.

I agree.  FYI, the samd mcu uses a different approach.  (On the samd boards, there really are no pre-defined groups and boards layout the pins in many different ways.)  See https://www.klipper3d.org/Config_Reference.html#samd_sercom .

However, the samd setup is a bit cumbersome for users.  They have to define a `samd_sercom` config section for every hardware bus they use.  For example:

```
[samd_sercom sercom4]
sercom: sercom4
rx_pin: PA12
tx_pin: PB10
clk_pin: PB11

[somesection]
spi_bus: sercom4
...

```

I was vaguely thinking that an alternative might be to enable something like the following in the config:
```
[somesection]
spi_bus: mymcu:spi1(PA6,PA7,PA5)
```
Reworking all the code (and configs) to a new system would be a lot of work though.  (It's also not clear to me if the above is a good plan, how it would interact with the `cs_pin`, how it would interact with software spi, etc. .)

-Kevin

KevinOConnor:
You do have a good point though, in that we could just rename `spi_software_x_pin` to `spi_x_pin` and use that to define the pins.

FYI, it isn't necessary to define a new `spi_function` config option as that can be determined automatically from the spi_bus name.

-Kevin

KevinOConnor:
Another option would be to go the samd_sercom route, but make it more generic.  Something like:

```
[spi_bus mybus]
device: spi1
miso_pin: mymcu:PA6
mosi_pin: mymcu:PA7
sclk_pin: mymcu:PA5

[somesection]
spi_bus: mybus
```

That might be more convenient for tmc driver spi buses where there are a number of devices using the same spi bus (and thus it would require less cut-and-paste for all the pins).

-Kevin

adelyser:
@KevinOConnor Funny thing, I was just thinking of something very similar to that last one. It might be good to put a sample_stm32_spi.cfg file too and list out all the possible pins for each stm processor that can be used, or alias popular groupings that could be included.

I can take on trying to rework the stm32 SPI stuff and see what pans out as easy to use yet more flexible.

Regarding this MR though, I'm fine either way if you want to merge it or I can close it if iit's deemed not needed.

KevinOConnor:
Thanks.  My feedback is that this PR could introduce a regression and it's probably not worth risking at the current time.

-Kevin

adelyser:
Thanks, closing the PR

PR #5104, opened 2022-01-03
rp2040: add make flash support

dalegaard:
This adds `make flash` support for the rp2040 target. Flashing is performed using `picotool` and doesn't require root.

The user specifies the serial of the rp2040 they wish to flash as the device. This device is looked for and reset into bootsel mode, after which picotool is invoked on the first rp2040 found. Sadly, an rp2040 in bootsel mode doesn't appear to expose its flash serial number in any way, which means we can't be more specific in which target we flash.

### DIFF
--- a/lib/.gitignore
+++ b/lib/.gitignore
@@ -3,3 +3,4 @@ bossac/bin/
 bossac/obj/
 hidflash/hid-flash
 hub-ctrl/hub-ctrl
+rp2040_flash/rp2040_flash
--- a/lib/README
+++ b/lib/README
@@ -74,6 +74,11 @@ version 1.2.0 (bfcbefafc5d2a210551a4d9d80b4303d4ae0adf7). It has been
 modified so that it can build outside of the pico sdk. See
 rp2040.patch for the modifications.
 
+The rp2040_flash directory contains a light-weight bootsel flash tool.
+It uses C part of the the `picoboot_connection` directory found in:
+  https://github.com/raspberrypi/picotool.git
+version v1.1.0 (55fd880c3dc029b961fc1a0967a6cfdc0af02721).
+
 The hub-ctrl directory contains code from:
   https://github.com/codazoda/hub-ctrl.c/
 revision 42095e522859059e8a5f4ec05c1e3def01a870a9.
--- a/lib/rp2040/boot/picoboot.h
+++ b/lib/rp2040/boot/picoboot.h
@@ -1,0 +1,124 @@
+/*
+ * Copyright (c) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _BOOT_PICOBOOT_H
+#define _BOOT_PICOBOOT_H
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <assert.h>
+
+#ifndef NO_PICO_PLATFORM
+#include "pico/platform.h"
+#endif
+
+/** \file picoboot.h
+*  \defgroup boot_picoboot boot_picoboot
+*
+* Header file for the PICOBOOT USB interface exposed by an RP2040 in BOOTSEL mode.
+*/
+
+#define PICOBOOT_MAGIC 0x431fd10bu
+
+// --------------------------------------------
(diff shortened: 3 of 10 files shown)


KevinOConnor:
Thanks.  In general, it looks good to me.  Some high-level questions:

1. What system package has the picotool binary?  Shouldn't we update the install dependencies for it?
2. Can you expand on how the device/bus address is found for picotool?  The hidtool code (`flash_dfuutil()`) converts the given serial path to a "devpath" and "buspath", requests bootloader via the "devpath", waits for the device to reconnect, and then flashes the device using the "buspath".  Could this mechanism be reused for picotool?  Why can't we flash the appropriate mcu by using the usb path?

Cheers,
-Kevin

dalegaard:
Hi Kevin,

I'll check what package contains picotool, that's a good point.

I'm not super familiar with USB so this is my lack of familiarity showing here. I was unaware that the MCU would come back on the same USB path. This makes things much nicer. I've force pushed a new patch that uses the `translate_serial_to_usb_path` function to find the device and follows the mechanism you highlighted.

If the MCU is already in bootloader mode, like for the first time flashing Klipper, there's no good way to target it. For that I've added the ability to specify `FLASH_DEVICE=first` which will invoke `picotool load` with no bus and address arguments, resulting in it flashing the first available rp2040. As the chip ID is not exposed in bootsel mode, that seems like the best we can do.

Best regards,
Lasse

KevinOConnor:
Ah, okay.  I don't think this support will be that useful if picotool isn't automatically provided.  Said another way, if a user can find, obtain, and build picotool on their own, then I think they would be able to run picotool on their own as well.
Typically we would place tools like this in the lib/ directory of the Klipper repo (like the existing hidtool and bossac).  Last I checked, though, picotool required the C++ toolchain to be installed and I was concerned that would be too many dependencies for the Klipper install.

Thanks again,
-Kevin

dalegaard:
Hi Kevin,

You are right, picotool requires a C++ toolchain.

Seeing as the tool is open source, the protocol is documented, and from I can tell we have libusb available, should I maybe take a stab at adding a minimal C based port of the minimum functionality we require? We could then include that in the lib/ folder.

Best regards,
Lasse

KevinOConnor:
Yeah - I noticed picoboot_connection.c as well.  A small C flashing program would certainly simplify deployment.

-Kevin

dalegaard:
Hi Kevin,

I've pushed a new version with a simple C program that performs the flashing. As I'm not super at home in C, I'd appreciate a review of it. It appears to work very well in testing.

Best regards,
Lasse

KevinOConnor:
Thanks!  In general it looks fine to me.  One thing I noticed is that we'll need to add a comment to lib/README with information on the new rp2040_flash directory.  Were the files added to rp2040/ and rp2040_flash/ taken unmodified from upstream?  (If not, we'll also need to document any changes.)

-Kevin

dalegaard:
Hi Kevin,

I've updated the `README`.

The `rp2040/` files and the `picoboot_connection.{c,h}` are unmodified copies.

Best regards,
Lasse

wlhlm:
> Ah, okay. I don't think this support will be that useful if picotool isn't automatically provided. Said another way, if a user can find, obtain, and build picotool on their own, then I think they would be able to run picotool on their own as well.

dfu-util for example is also not bundled with Klipper, but used for `make flash`. Similar thing for stm32flash.

> Last I checked, though, picotool required the C++ toolchain to be installed and I was concerned that would be too many dependencies for the Klipper install.

bossa is a C++ applications too. Moreover, the Klipper install scripts already install a C++ toolchain (via build-essential for example).

Sineos:
> bossa is a C++ application too. Moreover, the Klipper install scripts already install a C++ toolchain (via build-essential for example).

If I understood correctly, you also need the pico SDK --> https://github.com/raspberrypi/pico-sdk
This again brings in dependencies and needs to be cloned from git separately

KevinOConnor:
Thanks.

-Kevin

timmit99:
Does this work over UART? Just tested it today with my Ras Pi V1.2 and Pi Pico running via UART (the Pico has ALREADY been flashed via the USB mode and its known working in klipper) and it doesn't seem to like `FLASH_DEVICE=/dev/ttyAMA0`

PR #5113, opened 2022-01-06
Update pin description for neopixel section

schnello:
Add the information the Config_Reference.md, Section [neopixel]:
linux mcu pins are not supported.

Signed-off-by: Christian Schnellrieder <schnellrieder.cs@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2553,7 +2553,7 @@ sections with a "neopixel" prefix). One may set the LED color via
 [neopixel my_neopixel]
 pin:
 #   The pin connected to the neopixel. This parameter must be
-#   provided.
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  We can certainly add a comment to the documentation about linux/neopixel.  As in #5076 .  However, the sentence here doesn't make sense to me - I'm not sure I'd understand it if I didn't already know the issue.

-Kevin

schnello:
Thanks for the comment:
Would this line make more sense from your view:
"pins provided by an linux mcu (aka Raspberry Pi MCU) are not supported"

KevinOConnor:
FWIW, I'd add the warning to the neopixel description; not the pin description.  As in #5076 .

-Kevin

PR #5115, opened 2022-01-06
Update the SD Update title to be more accurate and less confusing/Update TMC 2208 and 2209 registers to not conflict

camerony:
Quickly looking at the Overview or the SDCard page implies the information is for regular SD updates rather than updating via SSH.  This should help people find this topic more readily.


camerony:
Update TMC default registers for 2208 and 2209 to set "multistep_filt" to false, since it's not compatible with "dedge".

ld3300:
hi @camerony 
I can't comment on the code, but make sure you add the Signed-off-by: line as stated in the [contributing docs](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)

camerony:
TMC: Update TMC 2208/2209 Default Registers

TMC documentations states that
"1: Enable step impulse at each step edge to reduce step
frequency requirement. This mode is not compatible
with the step filtering function (multistep_filt)"
Page 33
https://www.trinamic.com/fileadmin/assets/Products/ICs_Documents/TMC2209_Datasheet_V103.pdf 

Signed-off-by: Cameron River <camerony@gmail.com>

dewi-ny-je:
My  understanding is that commits should focus on an issue a time. You are committing together a change of the docs about SD card and a code change related to TMC drivers. I think they belong to separate commits.

camerony:
> My understanding is that commits should focus on an issue a time. You are committing together a change of the docs about SD card and a code change related to TMC drivers. I think they belong to separate commits.

Technically it's two commits but 1 pull request.  I wonder if I can revert the Pull Request or cancel it since it hasn't been approved for months

PR #5119, opened 2022-01-07
Add dedicated hardware pins on generic-bigtreetech-skr-mini-e3-v2.cfg

strayr:
Added display, runout, neopixel, beeper and bltouch dedicated ports to config.

These definitions are in the pinout availabe at https://github.com/bigtreetech/BIGTREETECH-SKR-mini-E3/blob/master/hardware/BTT%20SKR%20MINI%20E3%20V2.0/Hardware/BTT%20SKR%20MINI%20E3%20V2.0-PIN.pdf

I'm fielding a lot of requests for help on r/klippers related to pins that aren't mentioned in the sample configs, i'd like to update these for the boards I'm familiar with and I'd like feedback on how i've done it, and the devs preferred style for this. 

I don't think it's reasonable to ask the current influx of new users with badly documented boards to find the correct pinout diagram or read through marlin sourcecode in order to use off-the-shelf accessory kits on boards with **clearly labelled sockets that are intended for this additinal hardware**. It's quite common to start with a cheap creality, throw in a bltouch, filament sensor, this board that is sold as a drop-in upgrade and then try klipper.

I've copied the style of the creality 4.2.10 board and all my additions are comments so will not cause issues if the sample config is used as-is.

I intend to replicate this for the other boards i am familiar with if appropriate.

I'm not sure how to document pins intended for functionality that isn't supported by klipper like power loss detection.

Should there be a link in the configs to manufacturer's pinouts where they exist?

Signed-off-by: Rowland Straylight rowlandstraylight@gmail.com

### DIFF
--- a/config/generic-bigtreetech-skr-mini-e3-v2.0.cfg
+++ b/config/generic-bigtreetech-skr-mini-e3-v2.0.cfg
@@ -126,3 +126,32 @@ aliases:
     EXP1_2=PA15, EXP1_4=<RST>, EXP1_6=PB9,  EXP1_8=PB15, EXP1_10=<5V>
 
 # See the sample-lcd.cfg file for definitions of common LCD displays.
+
+# This is the pin config for a 128x64 display as found on the ender3(pro)
+#[display]
+#lcd_type: st7920
+#cs_pin: PB8
+#sclk_pin: PB9
+#sid_pin: PB15
+#encoder_pins: ^PA10, ^PA9
+#click_pin: ^!PA15
+
+# [filament_switch_sensor spool]
+# pause_on_runout: True
+# switch_pin: ^!PC15
+
+# Neopixel LED support
+# [neopixel led_neopixel]
+# pin: PA8
+
+# BL-touch
+#[bltouch]
+#sensor_pin: ^PC14
+#control_pin: PA1
+
+# [output_pin beeper]
+# pin: PB5
+
+# Additional pins available
+# PC13  // Power Supply Switch
+# PC12  // Power Loss Sense
--- a/config/generic-bigtreetech-skr-mini-e3-v2.0.cfg
+++ b/config/generic-bigtreetech-skr-mini-e3-v2.0.cfg
@@ -119,6 +119,10 @@ max_z_accel: 100
 [static_digital_output usb_pullup_enable]
 pins: !PA14
 
+########################################
+# EXP1 / EXP2 (display) pins
+########################################
+
 [board_pins]
 aliases:
     # EXP1 header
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I think it would be good to add additional pin aliases and "commented out" sections for hardware pin headers to the example config files.

It looks like you've copied the layout from the `generic-creality-v4.10.0.cfg` file.  That file is a bit of an outlier, and I'm not sure it's a good starting point.  The file layout is a bit haphazard, the comments are sometimes free form text and sometimes "commented out" config sections with no visual distinction between them, and it does not utilize the `board_pins` feature.

I'd say `generic-bigtreetech-skr-v1.3.cfg` is probably a better starting point.  (At least, in the sense of being more similar to the other configs.)

As to your high-level question about best practices for the configs going forward - my feedback is that I think we should try to keep all the configs concise and similar.  Similar layout, similar style comments, similar config section order, similar base hardware definitions.  This is important because we have a large number of example configs in the repository today (over 150), and although one may not like the quality of the current configs, I'm positive it will be dramatically worse if we have 150+ radically different example configs.  Most printers are similar and we should emphasize that similarity.  Along the same lines, in keeping with being "concise", we don't want to copy-and-paste comments between configs as then we'll have 100s of places to update should we need to change something.

We can also change the general layout of the configs going forward (as in https://klipper.discourse.group/t/reworking-example-config-files-to-have-pin-aliases-at-top-of-file/1689 ).  I think to do that will require a consensus and volunteers willing to submit patches updating existing configs, submit documentation updates to the config guidelines ( https://www.klipper3d.org/Example_Configs.html ), review config patches ( https://www.klipper3d.org/CONTRIBUTING.html ), and generally keep the configs similar by virtue of coordinating all the changes necessary to move all/most configs to a new layout.  This is, understandably, a lot of work.

Cheers,
-Kevin

strayr:
Can I take this as consensus?
>  If the board has a specific hardware port to facilitate an optional peripheral (eg, a bltouch port) then one can add a "commented out" config section for the given device. (https://www.klipper3d.org/Example_Configs.html section 5a)
 
Are we looking to see a section like this at the bottom of the config with both [board_pins] and config stubs? Should the config stubs use the aliases? Should there be one [board_pins] definition, or split out so display pins are next to display etc.? I've put the definitions in alphabetical order rather than try to reflect physical layour or popularity.

```
########################################
# Optional Peripherals
########################################c

[board_pins]
aliases:
    # Neopixel Support
    NEOPIXEL_1=<GND>, NEOPIXEL_2=PA8, NEOPIXEL_3=<5V>,
     # Runout sensor
    E_STOP_1=PC15, E0_STOP_2=<GND>, E0_STOP_3=<5V>,
    # Z-Probe (BL-Touch)
    Z_PROBE_1=<GND>, Z_PROBE_2=<5V>, Z_PROBE_3=PA1, Z_PROBE_4=<GND>, Z_PROBE_5=PC14

# BL-touch
#[bltouch]
#sensor_pin: ^Z_PROBE_5
#control_pin: Z_PROBE_3

# [filament_switch_sensor spool]
# pause_on_runout: True
# switch_pin: ^E0_STOP_2

# Neopixel LED support
# [neopixel led_neopixel]
# pin: NEOPIXEL_2

```


Changing the overall layout of the configs to put [board_pins] first is a separate argument and requires more systemic change and more work than just adding the definitions required for optional peripherals at the bottom of configs. I think it's important to get the pinouts for popular and expecially badly documented bords in the configs.

Sineos:
I like this approach 👍

strayr:
Another approach that is more consise and avoids config stubs is to add aliases like this

Option 1: combined physical and logical naming. I dislike this
```
[board_pins]
aliases:
    ...
    # Z-Probe (BL-Touch)
    Z_PROBE_1=<GND>, Z_PROBE_2=<5V>, Z_PROBE_SERVO=PA1, Z_PROBE_4=<GND>, Z_PROBE_SENSE=PC14
...
```

Or option 2: both physical and logical naming. This is the most help, but the most effort
```
[board_pins]
aliases:
    ...
    # Z-Probe (BL-Touch)
    Z_PROBE_1=<GND>, Z_PROBE_2=<5V>, Z_PROBE_3=PA1, Z_PROBE_4=<GND>, Z_PROBE_5=PC14 #physical layout
    Z_PROBE_SERVO=PA1, Z_PROBE_SENSE=PC14 #logical layout
...
```

Or option 3: logical names only, this is least work to do, but doent fully describe the physical layout,
```
[board_pins]
aliases:
    ...
    # Z-Probe (BL-Touch)
    Z_PROBE_SERVO=PA1,  #Z_PROBE_3
    Z_PROBE_SENSE=PC14 #Z_PROBE_5
...
```

Is there a schema somwhere for logical pin names? Should I crib off marlin and use SERVO0_PIN and use Z_MIN_PROBE_PIN in the example above?

Sineos:
As stated in https://klipper.discourse.group/t/reworking-example-config-files-to-have-pin-aliases-at-top-of-file/1689 this would need a common agreement on the approach to take.
It makes only sense when the boards are updated step by step in the same manner and new contributions use the same method from the beginning.

strayr:
I think it got side tracked into a discussion about what to do regarding
pin aliases? I'm not sure what the outcome was as it kind of fizzled out? I
may have missed a nuance there.

I'd like to have the available pins and their intended use marked,
particularly for touch probes, and I think dedicated neopixel is becoming
really relevant with what voron are doing.

If you're still interested I can update this, but please confirm the
direction you want to go with aliases.

PR #5127, opened 2022-01-08
config: update mks rumba32 example config

DanielBergsch:
Makerbase Rumba32 needs a bootloader offset of 0kb, added information to config example 

Signed-off-by: Daniel Bergsch <Danielbergsch@hotmail.de>


KevinOConnor:
Thanks.  It looks fine to me (though confirmation from @popshansen would be good).  Is it still worthwhile to merge this?

-Kevin

Issue #3373, opened 2020-09-26
SPI DAC DAC084S085 implementation


Blaster1920:
LOG FILE: [klippy.log](https://github.com/KevinOConnor/klipper/files/5286653/klippy.log)

Hi, I'm trying to communicate to my SPI DAC that controls the motor drivers's VRef on my Alligator rev.2 board.

The DAC is a DAC084S085 and here's the implementation for RepRapFirmware (RRF):
[DAC084S085 RRF.zip](https://github.com/KevinOConnor/klipper/files/5286657/DAC084S085.RRF.zip)

So far with the help of @MarcoAntonini we managed to generate the values that needs to be used to set the DAC channel and the DAC value (then translated into vref).
[dac_alligator py.txt](https://github.com/KevinOConnor/klipper/files/5286664/dac_alligator.py.txt)

Unfortunately as you can see from the .log it looks like the custom subsection shouldn't be in mcu(?)
What's would be the correct implementation for this?

Thanks in advance and congratulation for the immense work you've done so far.

KevinOConnor:
The `custom` option was removed some time back (it was too cryptic to use or maintain).

I recommend you take a look at klippy/extras/ad5206.py - you'll want to copy that to klippy/extras/dac084S085.py and then make the changes necessary to set the 4 channels on your particular device.

-Kevin

Blaster1920:
> The `custom` option was removed some time back (it was too cryptic to use or maintain).
> 
> I recommend you take a look at klippy/extras/ad5206.py - you'll want to copy that to klippy/extras/dac084S085.py and then make the changes necessary to set the 4 channels on your particular device.
> 
> -Kevin

Hi,
I tried to do as suggested but the SPI bus is not being set. I used a signal spoofer to check the busses but they're empty.
This is the script I've written:
[dac084S085.txt](https://github.com/KevinOConnor/klipper/files/5302091/dac084S085.txt)

And this is how it's called from the config file:
`[dac084S085 my_digipot]
enable_pin: PB14
vrefX: 125
vrefY: 125
vrefZ: 125
vrefE: 125
`
Any suggestion? With RRF I can confirm the board is functional.

KevinOConnor:
[...]

PR #3840, opened 2021-01-25
Alligator Board rev2 support

lorenzing:
This is the integration of the SPI DAC DAC084S085 that controls the motor drivers's VRef on my Alligator rev.2 board made by Jacopo Franco and Marco Antonini. 
Thanks in advance and congratulation for the work you've done.

PR #5134, opened 2022-01-09
SPI DAC DAC084S085 implementation

lorenzing:
SPI DAC DAC084S085 integration to support Alligator Board

Fixes: #3373
Signed-off-by: Lorenzo Franco lorenzo.franco@lorenzing.com

@KevinOConnor, I totally agree with the licensing terms described at https://github.com/KevinOConnor/klipper/blob/master/docs/developer-certificate-of-origin

This is the integration of the SPI DAC DAC084S085 that controls the motor drivers's VRef on my Alligator rev.2 board made by Jacopo Franco and Marco Antonini #3373.
New version tested of PR #3840.
I am attaching the klippy log (for now it is being tested on a Raspberry Pi B so with very few resources).
[klippy.log.txt](https://github.com/Klipper3d/klipper/files/7835820/klippy.log.txt)


lorenzing:
I am attaching the new klippy log (with commit files).
[klippy.log.txt](https://github.com/Klipper3d/klipper/files/7836123/klippy.log.txt)
All working good.

lorenzing:
@KevinOConnor if you need other things for the pull I'm at your disposal. 
The printer is working fine, even the expansion piggy card (3 extra driver/hotend) installed on the motherboard.
Hope the example file is clean enough, I preferred to show the connections for display and bltouch (even if commented) because otherwise they could be complex to do for a new user, many users of this board come from MK4duo (died) which is very different and uses Arduino pinouts instead of SAM3X Pin Name.

giyowiy:
It's great, I gave my alligator new life!
Everything seems to work, I hope it will be inserted in the official code soon.

Thanks Gyoway

lorenzing:
Thanks for your feedback @giyowiy

Giampi79:
Wow good job @lorenzing! I read about your work on the reprap forum, hoping the official support for this board is added @KevinOConnor I have an Alligator and I am confident to use it soon with klipper

lorenzing:
Thanks @Giampi79!
Hi @Sineos, I saw you have a lot of experience in config files, when you get it could you check that generic-alligator-r2.cfg in this request?
Thank you so much!

lorenzing:
I am attaching the log of a configuration tested on raspberry pi zero 2 w using repetier server. Tested with today's latest commits, it works fine.
[klippy_BLV_Max_Klipper.log.txt](https://github.com/Klipper3d/klipper/files/7982393/klippy_BLV_Max_Klipper.log.txt)

config/generic-alligator-r2.cfg
```diff
@@ -0,0 +1,162 @@
...
+[static_digital_output DRV8825_microstepping]
+pins:PC10
+pins:PC29
+pins:PC19
+pins:PC18
+
+[dac084S085 my_digipot]
+enable_pin: PB14
+spi_bus : spi0
+motor1 : 100 # define motor current from 0 to 255 where 255 is 2.5A
```
bevanweiss, line 22: I think labelling these 'channel1' etc would be better (here and in Python module).
On the alligator they're motor1-4, but if used elsewhere the channel assignment is different (i.e. if channel1 is actually for motors1-4 [x,y,z1,z2] and channel2 for extruders [e1,e2]) then calling them motor1-4 would be very confusing.

lorenzing: Thanks a lot for your suggestion! This is my first code on klipper and I really appreciate every comment. I hope the code is clearer that way

bevanweiss: It's more to maintain a separation between the device that your driver is referring to, and the one particular implementation of the device that is in your specific board.
So the DAC084S085 doesn't drive 'motors' right... it just drives a voltage output channel.  Your board just happens to use the first channel of voltage output for motor 1.. but another board might use the first channel for a laser power setpoint, or for a control valve % open command.
So it's important to remove the naming for the device away from the one implementation.

Good work, nicely done :)

lorenzing: I understand what you mean, but the DAC on Alligator boards is only connected to the drivers. It cannot be used for anything else.

Thanks, Lorenzo

bevanweiss: But the DAC does exist in other places than the Alligator board ;)
Your driver is (supposed to be) for the DAC.. not **just** the DAC on the Alligator board.

lorenzing: Of course, you are completely right, now using channel name seems more correct to me too. Thanks again for your feedback

lorenzing:
I have also added support to the Alligator rev.3 board hardware, adding the relative sample file.

lorenzing (commented):
Tested

bevanweiss:
> The job has been completed, tested and is ready for the pull.

Have you addressed the concerns raised by Kevin
https://github.com/Klipper3d/klipper/issues/3373#issuecomment-705239905

> eg, add copyright, time.sleep() isn't valid, add scale config parameter, add docs to config/example-extras.cfg

I think you've got the copyright done.
Have you fixed up the time.sleep()?
Unsure about adding scale config parameter... perhaps Kevin wanted like a channel1_minval, channel2_maxval type of thing, where it would then scale the channel1_value to output a proportional value between minval/maxval.  Or perhaps he meant something else.  It might be worth looking at things like the PWM driver to see for an example of scaling.

You also probably want to raise a new Issue to tag this PR against.  Which details what is currently missing from Klipper, and what should be added.  Then you change the PR comment to reflect that it would fix the Open issue created.

lorenzing:
Excuse me I was thrilled that everything works.
> I think you've got the copyright done. Have you fixed up the time.sleep()? 
>
I removed the delay, on the weekend I try the code without the time.sleep ()
>Unsure about adding scale config parameter... perhaps Kevin wanted like a channel1_minval, channel2_maxval type of thing, where it would then scale the channel1_value to output a proportional value between minval/maxval. Or perhaps he meant something else. It might be worth looking at things like the PWM driver to see for an example of scaling.
> 
Could you help me about adding scale config parameter? I don't think I understand how to add it.

Thank you so much
Lorenzo

bevanweiss:
For the scaling you should look to here
https://github.com/Klipper3d/klipper/blob/0b918b357cb0c282d53cbdf59e1931a2054cd60a/klippy/extras/ad5206.py

lorenzing:
Hi @bevanweiss, I've implemented scaled values like @KevinOConnor did on the ad5206.py. What do you think?
I just have a doubt about the conversion from float to int in python, whether it is done before or after the computation (not being able to measure the value on the board).

KevinOConnor:
Thanks.  At a high-level, it looks fine, but there are a few minor issues:
1. The two new config files have several instances of `option = value` syntax instead of `option: value` syntax. See: https://www.klipper3d.org/Example_Configs.html 
2. Both new config files need to be added to the test/klippy/printers.test file (see link above).
3. The generic-alligator-r2.cfg file should not have a commented out display section (see link above).
4. The generic-alligator-r2.cfg file should not have a commented out bltouch section.  If the board has an explicit bltouch port that you want to document, then provide a commented out bltouch section just describing the pins (and not anything about the bltouch hardware).
5. The dac084S085.py code makes a reference to channels 0 - 5, but the chip specification ( https://www.ti.com/lit/ds/symlink/dac084s085.pdf ) uses channels A, B, C, D.  So, the config and code should use channel_A, channel_B, channel_C, and channel_D.
6. I don't see how the daco84s085.py code could work because it attempts to transmit a 16-bit value where an 8-bit value is expected.  Was the code tested?

For the code, you could try something like the following snippet:
```python
class dac084S085:
    def __init__(self, config):
        self.spi = bus.MCU_SPI_from_config(
            config, 1, pin_option="enable_pin", default_speed=10000000)
        scale = config.getfloat('scale', 1., above=0.)
        for chan, name in enumerate("ABCD"):
            vref = config.getfloat('channel_%s' % (name,), None,
                                   minval=0., maxval=scale)
            if vref is not None:
                self.set_register(chan, int(val * 255. / scale))
    def set_register(self, chan, value):
        b1 = (chan << 6) | (1 << 4) | ((value >> 4) & 0x0f)
        b2 = (value << 4) & 0xf0
        self.spi.spi_send([b1, b2])
```

-Kevin

lorenzing:
Thank you so much @KevinOConnor you are the best!
I just tried the code you sent me and made a small change on a variable but it works very well:
> 
> ```python
> class dac084S085:
>     def __init__(self, config):
>         self.spi = bus.MCU_SPI_from_config(
>             config, 1, pin_option="enable_pin", default_speed=10000000)
>         scale = config.getfloat('scale', 1., above=0.)
>         for chan, name in enumerate("ABCD"):
>             val = config.getfloat('channel_%s' % (name,), None,
>                                    minval=0., maxval=scale)
>             if val is not None:
>                 self.set_register(chan, int(val * 255. / scale))
>     def set_register(self, chan, value):
>         b1 = (chan << 6) | (1 << 4) | ((value >> 4) & 0x0f)
>         b2 = (value << 4) & 0xf0
>         self.spi.spi_send([b1, b2])
> ```
> 

I wanted to ask if int (val * 255. / scale)) in the scale division takes the value 2.5 or if it only takes the integer value 2.
For everything else I study better and arrange the example configurations.

Thank you again
Lorenzo

bevanweiss:
It might be worth considering a per-channel scaling (vs one common scaling for all channels).

As for the float->int conversion, the int(x) takes care of that. One consideration there however is that int(x) performs truncation.
So you may actually want to do int(round(X, 1)) or similar to prevent loss of range (especially when converting to only 8bits).
Otherwise you'd end up with int(0.999999999999*255) = 254

lorenzing:
> It might be worth considering a per-channel scaling (vs one common scaling for all channels).
> 
> As for the float->int conversion, the int(x) takes care of that. One consideration there however is that int(x) performs truncation. So you may actually want to do int(round(X, 1)) or similar to prevent loss of range (especially when converting to only 8bits). Otherwise you'd end up with int(0.999999999999*255) = 254

If I'm not misunderstood, do you mean something like that?
> 
> ```python
>             if val is not None:
>                 digipot = int(round(val, 1))
>                 self.set_register(chan, int(digipot* 255. / scale))
>
> ```
> 
My biggest concern however is with the scale value which must be 2.5. I don't want it rounded to 2 before performing the division (for example 255 / 2.5).
I don't know if for the same reason I saw @kevin doing this in the ad5206.py
> 
> ```python
>             if val is not None:
>                 self.set_register(i, int(val * 256. / scale + .5))
>
> ```
> 
What I don't know is if the division is done before or after rounding the scale value. If the result of the division is rounded and not the scale value's rounded, that's okay for me.
Sorry for the trouble again.

bevanweiss:
The function call (int or round etc) applies to the passed parameter, which if you pass an expression is the evaluated outcome of the expression.
So it will calculate the full input*255/scale first, and then would truncate to an integer type (or round to an integer values float, and then truncate to an integer type).

It's only the final value that should be converted to 8bit int ideally. Not any intermediate values.

I see that the reference code only had a single scale, you could always support both.. get 'scale' first, then use that as default value when getting channel_A_scale.. so both would work.

lorenzing:
> The function call (int or round etc) applies to the passed parameter, which if you pass an expression is the evaluated outcome of the expression. So it will calculate the full input*255/scale first, and then would truncate to an integer type (or round to an integer values float, and then truncate to an integer type).
> 
> It's only the final value that should be converted to 8bit int ideally. Not any intermediate values.
> 
This is great news, if the scale value is not rounded, it's not very important to round the result of the division. Even if it's truncated, the difference in current is irrelevant.

lorenzing (commented):
Tested, I attach klippy log.
[klippy_BLV_Max_Klipper.log.txt](https://github.com/Klipper3d/klipper/files/8115799/klippy_BLV_Max_Klipper.log.txt)

PR #5139, opened 2022-01-10
module: filament_switch_sensor

geoffrey-young:
reset min_event_systime on sensor enable/disable

this significantly helps motion encoder sensors, as when prints
start or resume a sensor enable starts fresh (rather than in
the middle of a cycle)

Signed-off-by: Geoffrey Young <geoffrey.young@gmail.com>

### DIFF
--- a/klippy/extras/filament_switch_sensor.py
+++ b/klippy/extras/filament_switch_sensor.py
@@ -101,6 +101,7 @@ class RunoutHelper:
         gcmd.respond_info(msg)
     cmd_SET_FILAMENT_SENSOR_help = "Sets the filament sensor on/off"
     def cmd_SET_FILAMENT_SENSOR(self, gcmd):
+        self.min_event_systime = self.reactor.NEVER
         self.sensor_enabled = gcmd.get_int("ENABLE", 1)
 
 class SwitchSensor:


geoffrey-young:
hi, and thanks for the review :)

just so we're all on the same page, the use case here is starting with a
disabled sensor then calling:

  SET_FILAMENT_SENSOR sensor ENABLE=1

at the start of a print.  as the print progresses, runout/pause routines
call:

  SET_FILAMENT_SENSOR N ENABLE=0

so we can swap filament, etc. then the resume routine calls

  SET_FILAMENT_SENSOR sensor ENABLE=1

just prior to starting the print back up again.

assuming this is an appropriate model, the issue becomes that the resume
routine happens to start with the last encoder event time rather than anew,
leading to some odd results (for me, at least).

after debugging all weekend, this code was the simplest approach I could
find to remedy the issue.  however, I also scratched my head quite a bit,
since I'm clearly not the first person to use an encoder (BTT in my case)
and so the community certainly would have spotted this by now if it were
larger than just me.

anyway, if it is an actual issue and not user error, I can make the code
more explicit and only reset on ENABLE, but at the cost of adding more code
where the outcome doesn't seem to matter.

geoffrey-young:
ugh, never mind - you were right :/

closing.

PR #5142, opened 2022-01-11
Correct Fan Pin for the CR-6 SE 2021 Example Config

filiptrplan:
The fan pin PA0 is not correct for the 2021 revision of the CR-6 SE using the
4.5.3 revision of the Creality board. Changed the pin to PB15 in order to
get it working

Signed-off-by: Filip Trplan <filip@trplan.si>

### DIFF
--- a/config/printer-creality-cr6se-2021.cfg
+++ b/config/printer-creality-cr6se-2021.cfg
@@ -80,7 +80,7 @@ min_temp: 0
 max_temp: 120
 
 [fan]
-pin: PA0
+pin: PB15
 kick_start_time: 0.5
 
 [mcu]


Sineos:
https://github.com/Klipper3d/klipper/commit/77c486029a563faf5b2e08412fc22e6cd097b907#diff-392673d70021f5a57c1a16c9d6bb6c6a98cd34475038fedc9a0cdfe708deaca5

@cbluoss Can you confirm or are we talking about different hardware revisions?

filiptrplan:
> @cbluoss Can you confirm or are we talking about different hardware revisions?

The Mainboard version is the exact same and there shouldn't be any alternative revisions.

https://github.com/CR6Community/Marlin

According to the community firmware there are only the 4.5.2 and the 4.5.3 revisions of the mainboard in this printer. Just waiting for the confirmation now.

Sineos:
According to:
https://github.com/MarlinFirmware/Marlin/blob/2.0.x/Marlin/src/pins/stm32f1/pins_CREALITY_V452.h
https://github.com/MarlinFirmware/Marlin/blob/2.0.x/Marlin/src/pins/stm32f1/pins_CREALITY_V453.h

And also the slightly reworked:
https://github.com/CR6Community/Marlin/blob/extui/Marlin/src/pins/stm32f1/pins_CREALITY_V452.h
https://github.com/CR6Community/Marlin/blob/extui/Marlin/src/pins/stm32f1/pins_CREALITY_V453.h

is this change correct. I recommend merging.
Fan pin `PA0` belongs to v4.5.2 and `PB15` to v4.5.3

These Creality boards are a pin in the back. Maybe it makes sense to add the board version to the config file name for these boards. @KevinOConnor what do you think?

KevinOConnor:
If you're asking me to merge - that's fine, I'll merge if you think it is ready.

If you're asking me if we should change the filename, I'm not sure we need to do that at this point.  Adding a few comments in the file should be sufficient for now.

Thanks.
-Kevin

Sineos:
Fine for merging from my side.

Question related to the file naming.

KevinOConnor:
Thanks.

-Kevin

PR #5143, opened 2022-01-11
Refactor extruder steppers and extruder_stepper

KevinOConnor:
This PR reworks the extruder stepper logic to make it more flexible.  In particular, it is now possible to use the SYNC_STEPPER_TO_EXTRUDER command on either an `extruder` object or an `extruder_stepper` object.  It is also now possible to call SET_PRESSURE_ADVANCE and SET_EXTRUDER_STEP_DISTANCE on `extruder_stepper` objects.

This may facilitate synchronizing of multiple extruders on idex printers (such that one can print multiple objects simultaneously).  As in #4489.

This may facilitate support for mixing extruders (as in #3920, #4082, #4566).

This PR also deprecates the `shared_heater` support in extruder config sections.  It is thought that one can now obtain the same support by using `extruder_stepper` config sections.

I do not have a good way to test this code.  Feedback and testers would be appreciated.

-Kevin

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1835,8 +1835,9 @@ more information.
 ```
 [extruder_stepper my_extra_stepper]
 #extruder: extruder
-#   The extruder this stepper is synchronized to. The default is
-#   "extruder".
+#   The extruder this stepper is synchronized to. If this is set to an
+#   empty string then the stepper will not be synchronized to an
+#   extruder. The default is "extruder".
 #step_pin:
 #dir_pin:
 #enable_pin:
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -182,6 +182,11 @@ The following standard commands are supported:
   result in excessive pressure between extruder and hot end. Do proper
   calibration steps with filament before use. If 'DISTANCE' value is
   not included command will return current step distance.
+- `SYNC_STEPPER_TO_EXTRUDER STEPPER=<name> [EXTRUDER=<name>]`: This
+  command will cause the given extruder STEPPER (as specified in an
+  [extruder](Config_Reference#extruder) or
+  [extruder stepper](Config_Reference#extruder_stepper) config
+  section) to become synchronized to the given EXTRUDER.
 - `SET_STEPPER_ENABLE STEPPER=<config_name> ENABLE=[0|1]`: Enable or
   disable only the given stepper. This is a diagnostic and debugging
   tool and must be used with care. Disabling an axis motor does not
@@ -343,16 +348,6 @@ enabled:
   move completes, however if a manual stepper move uses SYNC=0 then
   future G-Code movement commands may run in parallel with the stepper
   movement.
-
-### Extruder stepper Commands
-
-The following command is available when an
-[extruder_stepper config section](Config_Reference.md#extruder_stepper)
-is enabled:
-- `SYNC_STEPPER_TO_EXTRUDER STEPPER=<extruder_stepper config_name>
-  [EXTRUDER=<extruder config_name>]`: This command will cause the given
-  STEPPER to become synchronized to the given EXTRUDER, overriding
-  the extruder defined in the "extruder_stepper" config section.
 
 ### Probe
 
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -186,7 +186,9 @@ The following standard commands are supported:
   command will cause the given extruder STEPPER (as specified in an
   [extruder](Config_Reference#extruder) or
   [extruder stepper](Config_Reference#extruder_stepper) config
-  section) to become synchronized to the given EXTRUDER.
+  section) to become synchronized to the given EXTRUDER. If EXTRUDER
+  is an empty string then the stepper will not be synchronized to an
+  extruder.
 - `SET_STEPPER_ENABLE STEPPER=<config_name> ENABLE=[0|1]`: Enable or
   disable only the given stepper. This is a diagnostic and debugging
   tool and must be used with care. Disabling an axis motor does not
--- a/klippy/chelper/__init__.py
+++ b/klippy/chelper/__init__.py
@@ -133,8 +133,8 @@ defs_kin_winch = """
 
 defs_kin_extruder = """
     struct stepper_kinematics *extruder_stepper_alloc(void);
-    void extruder_set_smooth_time(struct stepper_kinematics *sk
-        , double smooth_time);
+    void extruder_set_pressure_advance(struct stepper_kinematics *sk
+        , double pressure_advance, double smooth_time);
 """
 
 defs_kin_shaper = """
--- a/klippy/chelper/kin_extruder.c
+++ b/klippy/chelper/kin_extruder.c
@@ -52,15 +52,17 @@ extruder_integrate_time(double base, double start_v, double half_accel
 
 // Calculate the definitive integral of extruder for a given move
 static double
-pa_move_integrate(struct move *m, double base, double start, double end,
-                  double time_offset)
+pa_move_integrate(struct move *m, double pressure_advance
+                  , double base, double start, double end, double time_offset)
 {
     if (start < 0.)
         start = 0.;
     if (end > m->move_t)
         end = m->move_t;
     // Calculate base position and velocity with pressure advance
-    double pressure_advance = m->axes_r.y;
+    int can_pressure_advance = m->axes_r.y != 0.;
+    if (!can_pressure_advance)
+        pressure_advance = 0.;
     base += pressure_advance * m->start_v;
     double start_v = m->start_v + pressure_advance * 2. * m->half_accel;
     // Calculate definitive integral
@@ -72,34 +74,36 @@ pa_move_integrate(struct move *m, double base, double start, double end,
 
 // Calculate the definitive integral of the extruder over a range of moves
 static double
-pa_range_integrate(struct move *m, double move_time, double hst)
+pa_range_integrate(struct move *m, double move_time
+                   , double pressure_advance, double hst)
 {
     // Calculate integral for the current move
     double res = 0., start = move_time - hst, end = move_time + hst;
     double start_base = m->start_pos.x;
-    res += pa_move_integrate(m, 0., start, move_time, start);
-    res -= pa_move_integrate(m, 0., move_time, end, end);
+    res += pa_move_integrate(m, pressure_advance, 0., start, move_time, start);
+    res -= pa_move_integrate(m, pressure_advance, 0., move_time, end, end);
     // Integrate over previous moves
     struct move *prev = m;
     while (unlikely(start < 0.)) {
         prev = list_prev_entry(prev, node);
         start += prev->move_t;
         double base = prev->start_pos.x - start_base;
-        res += pa_move_integrate(prev, base, start, prev->move_t, start);
+        res += pa_move_integrate(prev, pressure_advance, base, start
+                                 , prev->move_t, start);
     }
     // Integrate over future moves
     while (unlikely(end > m->move_t)) {
         end -= m->move_t;
         m = list_next_entry(m, node);
         double base = m->start_pos.x - start_base;
-        res -= pa_move_integrate(m, base, 0., end, end);
+        res -= pa_move_integrate(m, pressure_advance, base, 0., end, end);
     }
     return res;
 }
 
 struct extruder_stepper {
     struct stepper_kinematics sk;
-    double half_smooth_time, inv_half_smooth_time2;
+    double pressure_advance, half_smooth_time, inv_half_smooth_time2;
 };
 
 static double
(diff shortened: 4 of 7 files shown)


Black6spdZ:
very interested in this and willing to test

kodonnell:
>  Feedback and testers would be appreciated.

I've got an IDEX and have been following e.g. #4464 . As a newbie to klipper (and 3d printers), this might be a really silly question/suggestion ... but I wonder if a single `SYNC_MOTORS` command is more generic, more obvious, and more flexible in future. In this case (with IDEX in mind) you could sync the steppers on the extruders so that `extruder` is the master, and `extruder1` is the slave and is just sent the exact same pulses etc. There are possibly other cases where you'd want to synchronise other motors (e.g. the two X motors on duplicate mode for IDEX, or synced-but-reversed for mirror mode). Unless I'm missing something, one then doesn't need to worry about adding support for pressure advance etc. - the slave motor just duplicates whatever the master does, and if the master is doing pressure advance, then so will the slave, etc.

Again, apologies if I've missed something obvious.

Black6spdZ:
> > Feedback and testers would be appreciated.
> 
> I've got an IDEX and have been following e.g. #4464 . As a newbie to klipper (and 3d printers), this might be a really silly question/suggestion ... but I wonder if a single `SYNC_MOTORS` command is more generic, more obvious, and more flexible in future. In this case (with IDEX in mind) you could sync the steppers on the extruders so that `extruder` is the master, and `extruder1` is the slave and is just sent the exact same pulses etc. There are possibly other cases where you'd want to synchronise other motors (e.g. the two X motors on duplicate mode for IDEX, or synced-but-reversed for mirror mode). Unless I'm missing something, one then doesn't need to worry about adding support for pressure advance etc. - the slave motor just duplicates whatever the master does, and if the master is doing pressure advance, then so will the slave, etc.
> 
> Again, apologies if I've missed something obvious.

this? https://www.klipper3d.org/Config_Reference.html?h=sync#extruder_stepper

jalanjarosz:
Kevin, 
I haven't looked into the code changes of this PR.  I'm not adept to Python yet.

Suggestion:
Klipper "Extruder" is a combination of elements: Extruder, Heater, Temp Sensor, and Fan (keeping things simple).

Why not remove the Heater and Fan sub-elements from the Extruder and reference the generic "stand-alone" modules that are in Klipper? ie: add a "Tool_x" module that calls out a Heater, Temp Sensor, Cooling Fan, Tool Fan, and Extruder modules that reference/call the generic objects or list of generic objects?

You could possibly extend the "Tool_x" module to include things like spindle and coolant pumps expanding the use of Klipper into milling or other maker hobbies.

theophile:
As currently set up, my printer has six bowden "extruders" that feed into a single direct-drive "extruder_stepper" at the hotend. Below is a quick diagram, where "Ex_" represents an Extruder printer object and "ES" represents an Extruder_Stepper printer object under the current code: 
```
 _____   _____   _____   _____   _____   _____
|     | |     | |     | |     | |     | |     |
| Ex0 | | Ex1 | | Ex2 | | Ex3 | | Ex4 | | Ex5 |
|_____| |_____| |_____| |_____| |_____| |_____|
   |       |       |       |       |       |
   |       |       |       |       |       |
   |       |       |       |       |       |
   |_______|_______|_______|_______|_______|
                       |
                       |
                       |
                     __|__
                    |     |
                    | ES  |
                    |_____|
                       V
```
Currently, each of my Extruders has a shared heater because I only have a single hotend. Also, the Extruder_Stepper always has to be synced to the currently active Extruder because they work in tandem in a sort of push-pull arrangement. This works fine for my purposes.

If I'm understanding correctly, if this PR is merged, then I would need to reconfigure my setup so that I have only one Extruder, and six Extruder_Steppers, like so:
```
 _____   _____   _____   _____   _____   _____
|     | |     | |     | |     | |     | |     |
| ES0 | | ES1 | | ES2 | | ES3 | | ES4 | | ES5 |
|_____| |_____| |_____| |_____| |_____| |_____|
   |       |       |       |       |       |
   |       |       |       |       |       |
   |       |       |       |       |       |
   |_______|_______|_______|_______|_______|
                       |
                       |
                       |
                     __|__
                    |     |
                    | Ex0 |
                    |_____|
                       V
```
Then I would not need a shared heater because I only have one Extruder. And instead of switching between six Extruders and then making sure the single Extruder_Stepper is synced to the appropriate active Extruder, I would only ever have one active Extruder and would simply change which of the six Extruder_Steppers is synced to it.

Is this correct? If so, I can foresee problems arising from the fact that an Extruder_Stepper does not move on its own, but only in sync with an Extruder:

- In my current setup, during filament changes, I disable the direct-drive Extruder_Stepper at the hotend once the filament is retracted from it, and I don't re-enable it again until the new filament is ready to be loaded into it. This prevents it from grinding the gears needlessly when it's "empty." But under the new system, I would not be able to do this, because that stepper would now be the only "Extruder," and it would need to be moving in order for the upstream Extruder_Steppers to be moving.
- Similarly, a more common setup with a 2-in-1-out system is to have two extruders that feed independently into a single hotend. I don't think it's suitable for Extruder_Steppers to replace the shared heater, because in this use case, the Extruder_Stepper would not be able to move independently of the Extruder.

KevinOConnor:
> Extruder, I would only ever have one active Extruder and would simply change which of the six Extruder_Steppers is synced to it.
Is this correct?

Yes.

The proposal in this branch is to consider an `extruder` object as a combination of a temperature controlled hotend heater, a motion queue, and an extruder_stepper.

> In my current setup, during filament changes, I disable the direct-drive Extruder_Stepper at the hotend once the filament is retracted from it, and I don't re-enable it again until the new filament is ready to be loaded into it. This prevents it from grinding the gears needlessly when it's "empty." But under the new system, I would not be able to do this, because that stepper would now be the only "Extruder," and it would need to be moving in order for the upstream Extruder_Steppers to be moving.

It should be possible to disassociate the extruder stepper from the extruder motion queue:
```
SYNC_STEPPER_TO_EXTRUDER STEPPER=extruder EXTRUDER=
```

> Similarly, a more common setup with a 2-in-1-out system is to have two extruders that feed independently into a single hotend. I don't think it's suitable for Extruder_Steppers to replace the shared heater, because in this use case, the Extruder_Stepper would not be able to move independently of the Extruder.

I'm not sure what you are reporting.  I don't see why it wouldn't work.  I'd envision the user would supply macros like the following:
```
[gcode_macro T0]
gcode:
    # Deactivate stepper in extruder
    SYNC_STEPPER_TO_EXTRUDER STEPPER=extruder EXTRUDER=
    # Activate stepper in my_extruder_stepper
    SYNC_STEPPER_TO_EXTRUDER STEPPER=my_extruder_stepper EXTRUDER=extruder

[gcode_macro T1]
gcode:
    # Deactivate stepper in my_extruder_stepper
    SYNC_STEPPER_TO_EXTRUDER STEPPER=my_extruder_stepper EXTRUDER=
    # Activate stepper in extruder
    SYNC_STEPPER_TO_EXTRUDER STEPPER=extruder EXTRUDER=extruder
```

Am I missing something?

Black6spdZ:
wouldn't this break the opportunity to run separate extruders in a mixing config? or maybe constantly synced but with the capability to alter the proportion "steps/mm" of the two?

theophile:
> I'm not sure what you are reporting. I don't see why it wouldn't work.

As I understand it, an Extruder_Stepper cannot move alone. It can only move in sync with an Extruder. So in your example macros, if the user issues `T1`, then only "extruder" will be executing the moves while "my_extruder_stepper" sits idle. If the user issues `T0`, then _both_ "extruder" _and_ "my_extruder_stepper" will be executing the moves in sync. There is no way to get "my_extruder_stepper" to execute the moves by itself, while "extruder" sits idle.

In a basic 2-in-1-out configuration, the user needs two steppers to be able to move independently of one another. And with a mixing hotend like @Black6spdZ is talking about, the user needs multiple steppers to move independently of one another simultaneously.

KevinOConnor:
> If the user issues T0, then both "extruder" and "my_extruder_stepper" will be executing the moves in sync. 

No - this PR explicitly adds the ability to disassociate an extruder stepper motor from the extruder motion queue.  The example `T0` macro above invokes the appropriate command and thus the extruder stepper motor will no longer move with extruder movement.  (Pending test results that everything works as expected, of course.)

-Kevin

theophile:
Ah, that's helpful. I may have been making incorrect assumptions based on the current implementation. If I'm following, then under this PR, an Extruder will simply be a default association of a heater, a motion queue, and an extruder_stepper. Setting the active "Extruder" simply activates these three associated objects, one of which is a defined Extruder_Stepper.

From there, the user can change these independent objects. With respect to the Extruder_Stepper, the user could "unsync" the current one and "sync" a different one instead (so as to use a 2-in-1-out hotend). Or sync an additional Extruder_Stepper to the already active one (as in my push-pull configuration, or an IDEX printer).

If that's right, I like it! One thought though: Wouldn't it make more sense to sync the Extruder_Stepper to a movement queue object, rather than an Extruder object? I think that's what had me confused. In your example macro, the line `SYNC_STEPPER_TO_EXTRUDER STEPPER=my_extruder_stepper EXTRUDER=extruder` makes me think that the stepper motor defined as "my_extruder_stepper" will mirror the movements performed by the stepper motor defined in "extruder."

Maybe the command JOIN_STEPPER_TO_QUEUE or something along those lines could make the intent/usage clearer?

KevinOConnor:
I agree the SYNC_STEPPER_TO_EXTRUDER command is a bit confusing.  I reused the existing command so as not to break any existing users.

However, it does seem that it's probably better to remove SYNC_STEPPER_TO_EXTRUDER and introduce a new command (eg, `SYNC_EXTRUDER_MOTION EXTRUDER=xxx MOTION_QUEUE=yyy`).  I'll look to update this PR in the next few days.

-Kevin

Black6spdZ:
> `SYNC_EXTRUDER_MOTION EXTRUDER=xxx MOTION_QUEUE=yyy`)

hmm, this looks promising.. but I am still at a loss how to alter the extruder's motion as a set percentage of it's original "rotation_distance" calc. There is legecy set_extruder_step_distance but it's value is inversely proportional and would be exponentially large to attempts say a 1% extrusion rate

Marlin-bert:
Hi @KevinOConnor,

I was testing this PR and PR#4464 together with @Tircown on a Cartesian idex printer. We observed the following strange behavior: If PA is active on one extruder, klipper will go into shutdown, as soon as a extrude command (G1 E…) is issued for the given extruder. In other words: 
Everything is working as long as PA is deactivated for every extruder, that gets used.
This behavior was also observed by @DrumClock.
Following tests have been made:

1. EX0:PA EX1:No PA
	1.1 Dual Color: Internal error on command:"G1" (shutdown after homing)
	1.2 Duplication/Mirror: Internal error on command:"G1" (shutdown after homing)

2. EX0:No PA EX1:PA
	2.1 Dual Color: Internal error on command:"G1" (shutdown when T1 activated)
	2.2 Duplication/Mirror: Internal error on command:"G1" (shutdown after homing)

3. EX0:PA = EX1:PA
	3.1 Dual Color: Internal error on command:"M109" (shutdown after homing)
	3.2 Duplication/Mirror: Internal error on command:"G1" (shutdown after homing)

4. EX0:PA EX1:PA
	4.1 Dual Color: Internal error on command:"M109" (shutdown after homing)
	4.2 Duplication/Mirror: Internal error on command:"G1" (shutdown after homing)

KevinOConnor:
Thanks for testing.  Unfortunately there was a typo in the pressure advance code.  I'm not sure how it managed to pass the regression test cases.  I've corrected it (hopefully) and rebased this branch.

-Kevin

KevinOConnor:
> hmm, this looks promising.. but I am still at a loss how to alter the extruder's motion as a set percentage of it's original "rotation_distance" calc. There is legecy set_extruder_step_distance but it's value is inversely proportional and would be exponentially large to attempts say a 1% extrusion rate

I'm not really sure what you're reporting.  If you've got a printer you'd like to test, can you provide a description of it, your current config, and your test attempts with this branch?

It may be possible to configure a mixing extruder with this branch using the SET_EXTRUDER_STEP_DISTANCE command on extruder_stepper and extruder objects.  I don't have mixing extruder hardware though, so can't test it.

-Kevin

DrumClock:
Hi @KevinOConnor 
YES it works ..... 

- extruder with value pressure_advance: 0.8 
- extruder1 where bya pressure_advance: 0.0

the "plucking" of the extruder was seen as the X Y changes
https://youtu.be/PEEIpGroXLM

KevinOConnor:
Thanks.

I went ahead and merged this PR, but without the deprecation of the shared_heater.  I'll plan to make a separate PR that deprecates shared_heater and SYNC_STEPPER_TO_EXTRUDER.

-Kevin

KevinOConnor:
FYI, PR #5210 has the updates to deprecate `shared_heater`, along with other changes.

-Kevin

PR #5150, opened 2022-01-13
Rework tmc run_current selection to prefer vsense=1

KevinOConnor:
It was possible for tmc2130/tmc2208/tmc2209 drivers to be programmed with an irun=16 or irun=17 when it was also possible to configure the driver with irun=31
and vsense=1 instead.  This would occur on tmc2130/tmc2208/tmc2209 drivers for run_current values around 0.900 to 1.000 amps (when using a typical sense_resistor settings of 0.110 Ohms).

Programming the driver with irun=16 or 17 is not ideal as it could result in reduced microstep precision.

A similar issue could impact tmc2660 drivers.

This change updates the code to improve the run_current programming to avoid the above issue.

If one was using one of the impacted run_current settings then it is possible this change may invalidate any previous sensorless homing or spreadcycle tuning.

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,12 @@ All dates in this document are approximate.
 
 ## Changes
 
+20220116: The tmc2130, tmc2208, tmc2209, and tmc2660 `run_current`
+calculation code has changed. For some `run_current` settings the
+drivers may now be configured differently. This new configuration
+should be more accurate, but it may invalidate previous tmc driver
+tuning.
+
 20211230: Scripts to tune input shaper (`scripts/calibrate_shaper.py`
 and `scripts/graph_accelerometer.py`) were migrated to use Python3
 by default. As a result, users must install Python3 versions of certain
--- a/klippy/extras/tmc2130.py
+++ b/klippy/extras/tmc2130.py
@@ -116,30 +116,33 @@ class TMCCurrentHelper:
         vref = 0.32
         if vsense:
             vref = 0.18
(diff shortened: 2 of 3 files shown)


JNP-1:
With the current method, irun and ihold can reach a maximum value of 28. But the value should be able to go up to 31 (CurrentScaler off).
If I am not completely wrong I should get an irun=31 with vsense=1 at a run_current=0.25 and a sense_resitor=0.500. But in fact irun=17 with vsense=0 comes out.

KevinOConnor:
> If I am not completely wrong I should get an irun=31 with vsense=1 at a run_current=0.25 and a sense_resitor=0.500. But in fact irun=17 with vsense=0 comes out.

Make sure you're on the code from this development branch.  Otherwise, I'd need to see the full klipper log from the event to diagnose what is occurring.

-Kevin

KevinOConnor:
Oh, just to be clear, if you have a sense resistor of 0.500 ohms and you specified `sense_resistor = 0.500` in the config then note that irun=17 is the correct outcome on tmc2130/tmc2208/tmc2209.  The effective sense resistor is the actual sense resistor plus 0.020 as specified in the tmc docs.

-Kevin

KevinOConnor:
Ah, nevermind.  With `sense_resistor=0.500` and `run_current=0.250` you should get `irun=31` and I do in my test.  So, if you're not, make sure you're on this code and provide the full klipper log.

-Kevin

JNP-1:
I did my test with v0.10.0-227-gbabb067b just to see what difference it makes ;-)

PR #5156, opened 2022-01-15
pause_resume: Do not klobber default state name by accident.

individ-divided:
Previously accidentally used 'default' state name instead of the
stated one ('PAUSE_STATE'). This might collide with user usage of the
SAVE_GCODE_STATE macro.


Arksine:
Nice catch, looks good to me.  This very well could have led to potentially issues resuming if the "default" state was overwritten during a pause.

KevinOConnor:
Thanks.

-Kevin

PR #5162, opened 2022-01-15
BTT Octopus generic config: Fixing Driver3 enable pin

denis-itskovich:
The Driver3's enable pin should be inverted, in same manner as other drivers. For some reason (typo?) in the example config it was not inverted


legend069:
i can confirm mine is inverted for my config.

KevinOConnor:
Thanks.  The change looks fine to me, but commits need to have a signed-off-by line containing your real full name - see: https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

PR #5167, opened 2022-01-16
ignore the intimediary files folder

kpishere:
A housekeeping change. chelper looks like compiled intermediary files and just want to cut down the noise of any changes vs non-changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ out
 .config
 .config.old
 klippy/.version
+klippy/chelper


dmbutyugin:
It is not though: it is a directory with C implementation of certain core functionality:
https://github.com/Klipper3d/klipper/tree/master/klippy/chelper

kpishere:
@dmbutyugin Ah, well duh, I look dumb.  But when I revisit why I was thinking that, I see this path locally ...

klippy/chelper/c_helper.so.dSYM/

I see an *.so file in there, could this possibly be just a Mac artifact?

dmbutyugin:
Yes, it is likely some Mac-specific artifact.

PR #5181, opened 2022-01-21
filament_motion_sensor: Extended information

Cosik:
Added support for semiautomatic calibration for motion filament sensor

Hi I have more advanced filament motion sensor ([my design](https://www.prusaprinters.org/prints/112108-optical-filament-sensor-v21-with-drive-gear-for-17)) which has much higher resolution (0.239mm per state change) that other sensors on the market. My sensor is detecting also jamming and slipping filament which could be used to checking flow without additional equipment. Problem was that there was no implementation for calibration. My change could be used to do that.

I also changed CHECK_RUNOUT_TIMEOUT which is shouldn't be done but, it was to high and introduce big fluctuation of measurements.
Examples for 0.250:
> measured sensitivity 1.4366624980506812 over 35.87850745479
>  G1 E5.9797 F120
> measured sensitivity 1.4110518778766523 over 29.898168435135
> G1 E5.9797 F120
> G1 E5.9797 F120
> measured sensitivity 1.4786398526478919 over 23.918711081265

for 0.05
> measured sensitivity 1.2339777544849444 over 59.79898186762499
> G1 E5.9797 F120
> measured sensitivity 1.26369648992579 over 53.819524513755
> G1 E5.9797 F120
> measured sensitivity 1.2483245301150614 over 47.840067159884995
> G1 E23.92 F120
> measured sensitivity 1.2506111232890826 over 23.919592747049997

My proposal is to make it configurable in section for filament motion sensor.

When we will get conclusion of proper solutions I will also update documentation for those changes in this PR.

I also want to call @TheJoshW to board ;)

Signed-off-by: Michal Sentyrz <sentyrz.m@gmail.com>

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -624,6 +624,10 @@ is enabled.
 - `QUERY_FILAMENT_SENSOR SENSOR=<sensor_name>`: Queries the current
   status of the filament sensor. The data displayed on the terminal
   will depend on the sensor type defined in the confguration.
+  In a case when filament montion sensor is set query will retrun
+  information about measured sensitivity. This represent value
+  measured by firmware and could be used to calibration/tune of value
+  set in config file. This should prevent false positive triggering.
 - `SET_FILAMENT_SENSOR SENSOR=<sensor_name> ENABLE=[0|1]`: Sets the
   filament sensor on/off. If ENABLE is set to 0, the filament sensor
   will be disabled, if set to 1 it is enabled.
--- a/klippy/extras/filament_motion_sensor.py
+++ b/klippy/extras/filament_motion_sensor.py
@@ -75,8 +75,7 @@ class EncoderSensor:
     def encoder_event(self, eventtime, state):
         if self.extruder is not None:
             self._update_filament_runout_pos(eventtime)
-            # Check for filament insertion
-            # Filament is always assumed to be present on an encoder event
+            # Measured sensor sensitivity based on mean value
             self.measured_sensitivity += (self.filament_runout_pos -
                                           self.extruder_pos)
             self.measured_sensitivity /= 2.0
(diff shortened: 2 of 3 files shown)


KevinOConnor:

I'm not entirely sure what this PR does.  The comments indicate that this does some form of "auto calibration", but the code seems mostly about moving the location of the QUERY_FILAMENT_SENSOR command.  I'm a little confused.

-Kevin

Cosik:
Hi,

This PR extend functionality of existing implementation of filament motion sensor. It support:
1. Easier calibration for custom sensors (documentation will be added)
2. Extended displayed information form sensor like it is in RRF `M591 D0
Duet3D magnetic filament monitor on input 3, enabled, sensitivity 24.80mm/rev, allow 70% to 130%, check every 3.0mm, version 3, mag 126 agc 128, measured sensitivity 24.78mm/rev, min 97% max 102% over 109.0mm`
3. Add possibility to change sample time, for example in my sensor 250ms is to big value.

QUERY_FILAMENT_SENSOR was move and rewritten because it used so far that one from generic filament sensor with simple switch.

Cosik:
@KevinOConnor I just retrun from vacation but really I don't usterstand why it should be separate. This commit extend functionality for motion sensor, one of it is display measured values.

To good understand is to split it to separate pull requests or commit all changes related to one functionality (example implementation and doc) in separate commits in one PR?

KevinOConnor:
Alas, I don't really understand what this PR does.  I'd need to understand its high-level usage to be able to review it.  Splitting it up may help me understand it.

Looking a little closer this seems to just add some text to the QUERY_FILAMENT_SENSOR command.  What is "measurement sensitivity" and how does a user utilize it?

-Kevin

Cosik:
@KevinOConnor is added documentation sufficient? Or still should be clarified?

PR #5184, opened 2022-01-22
extras/tmc.py: Fix number of arguments in phase endstop

D4SK:
Found this error.

Also i tested a tmc261 chip with the tmc2660 code, it is fully compatible except for the lower current limit of 1.4A. Maybe this can be noted somewhere. The tmc261 is interesting because it supports up to 60v.


KevinOConnor:
Thanks.

As to tmc261, someone would need to submit a PR updating the docs and a reviewer would need to review it.  Due to time constraints on my side, it's unlikely I'll get to something like that in the near term.

-Kevin

PR #5188, opened 2022-01-23
docs: Updated simulavr building instructions

dmbutyugin:


dmbutyugin:
The current instructions are for some older version of simulavr and do not work with the latest simulavr release 1.1.0 from Dec 2019.

KevinOConnor:
Thanks.

-Kevin

PR #5191, opened 2022-01-24
Updated filament_motion_sensor.py

straccio:
added option: consecutive_count (default 3)
added gcode: FILAMENT_MOTION_SENSOR VERBOSE=[1/0] in order to enable the verbose
Changed the check _extruder_pos_update_event in order set filament not present when the filament difference beetwen the runnout sensor and the extruder position is greater than the detection_length for consecutive_count times.

In my test with this mode i can use the detection_lenght to 4mm (my sensor have 3mm + some margin seen with verbose) and trigger the runout early in case of clogs

### DIFF
--- a/klippy/extras/filament_motion_sensor.py
+++ b/klippy/extras/filament_motion_sensor.py
@@ -45,8 +45,9 @@ class EncoderSensor:
                 self._handle_not_printing)
         # Register GCODE commands
         self.gcode = self.printer.lookup_object('gcode')
-        self.gcode.register_command(
-            'SET_FILAMENT_MOTION_SENSOR', self.cmd_SET_FILAMENT_MOTION_SENSOR,
+        self.gcode.register_mux_command(
+            'SET_FILAMENT_MOTION_SENSOR', 'SENSOR',
+            self.runout_helper.name, self.cmd_SET_FILAMENT_MOTION_SENSOR,
             desc=self.cmd_SET_FILAMENT_MOTION_SENSOR_help)
     cmd_SET_FILAMENT_MOTION_SENSOR_help = \
         "Enable debug logging for the encoder event"
--- a/klippy/extras/filament_motion_sensor.py
+++ b/klippy/extras/filament_motion_sensor.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2021 Joshua Wherrett <thejoshw.code@gmail.com>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import logging
+import logging, collections
 from . import filament_switch_sensor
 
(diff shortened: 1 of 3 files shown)


Cosik:
@straccio did you look at #5181  ?

klippy/extras/filament_motion_sensor.py
```diff
@@ -36,12 +40,31 @@ def __init__(self, config):
...
         self.printer.register_event_handler('idle_timeout:idle',
                 self._handle_not_printing)
+        # Register GCODE commands
+        self.gcode = self.printer.lookup_object('gcode')
+        self.gcode.register_command(
+            'FILAMENT_MOTION_SENSOR', self.cmd_FILAMENT_MOTION_SENSOR,
+            desc=self.cmd_FILAMENT_MOTION_SENSOR_help)
+    cmd_FILAMENT_MOTION_SENSOR_help = "Enable debug logging for the encoder event"
+    def cmd_FILAMENT_MOTION_SENSOR(self, gcmd):
+        options = collections.OrderedDict({
```
Cosik, line 51: Also are you sure that using dictionary to enable verbose/debug is necessary?

Cosik (commented):
General question/suggestion. If you read PR for first filament motion sensor implementation you should get information that instead of using additional counter you should multiply detection value. In my opinion there is no need to introduce additional counter, instead documentation should include this information.

klippy/extras/filament_motion_sensor.py
straccio: It's not the same thing multiply the detection_lenght.
I currently use the BTT Smart Sensor V2, that not need to be calibrated, the pin goes up every 3mm forward o backward(retraction).
I suppose to check if the extruder try to push forward the filament and that can't go out for a clog.
On the first check i can determine that the filament was pushed for more than 3mm and nothing happen, no trigger from the sensor. This can happen because there's a partial clog, the extruder does not have enough force to resolve the clog and loose steps, the retraction does not trigger the sensor or the past calculated position can't be precise.
 
On the next "time" checks, i need to know if the distance is again greater than 3mm not 6,9,12 mm.

I think the calibration of this kind of sensor is useful if you don't know the correct detection_lenght to play with it. 
The real work can be the accuracy of the past position, but it's too difficult and not so useful; or the consider the trigger when the filament goes forward that is different when the trigger goes backward for a retraction.

straccio: For the verbose/debug it's not important. I have used it to understand what happen in print, how that difference change and understand how to perform better, if it perform better! :P

KevinOConnor:
Thanks.  Alas, I'm not sure what this PR does or its current state.  I think it will need a volunteer to review it.

-Kevin

PR #5197, opened 2022-01-26

laurienzu:
Rotation distance is wrong, I corrected them please accept changes... a lot of users are having problems!

### DIFF
--- a/config/printer-twotrees-sapphire-pro-2020.cfg
+++ b/config/printer-twotrees-sapphire-pro-2020.cfg
@@ -1,12 +1,13 @@
 # This file contains common pin mappings for the Two Trees Sapphire
-# Pro printer from 2020. To use this config, the firmware should be
+# Pro printer from 2020 (Robin Nano 1.2 and 2208/A4988 drivers). 
+# To use this config, the firmware should be
 # compiled for the STM32F103. When running "make menuconfig", enable
 # "extra low-level configuration setup", select the 28KiB bootloader,
 # serial (on USART3 PB11/PB10) communication, and set "GPIO pins to
 # set at micro-controller startup" to "!PC6,!PD13".
 
-# Note that the "make flash" command does not work with the Sapphire
-# Pro. After running "make", run the following command:
+# Note that the "make flash" command does not work with the Robin Nano. 
+# After running "make", run the following command from the klipper folder:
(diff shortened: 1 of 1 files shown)


th0m4sek:
As owner of Sapphire Pro with MKS Robin Nano v1.2 I'm confirming changes.

laurienzu:
I can't see the error

ld3300:
Hi @laurienzu,
You need to include a Signed-off-by: statement in your commit comment as described [here](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)

laurienzu:
> Hi @laurienzu, You need to include a Signed-off-by: statement in your commit comment as described [here](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)

I'm sorry but I'm so bad at Github, is like this okay?

ld3300:
@laurienzu, That's the right track, make sure you add an email address as well.
-David

ld3300:
That looks good to me, though I can't verify anything personally.

KevinOConnor:
This PR updates one file and adds another file.  I'm not sure if that is intentional.  The file added does not have the correct file extension (and should not have parenthesis in the file name), and was not added to the regression test case list.  The updated file does not wrap the text to 80 columns.  It's also unclear why the rotation distance has changed - has this been confirmed with @thucar ?

-Kevin

thucar:
I am not aware of any rotation distance changes being needed. @laurienzu what caused the need for that change? Is your Sapphire Pro using stock motor drivers and steppers?

laurienzu:
> I am not aware of any rotation distance changes being needed. @laurienzu what caused the need for that change? Is your Sapphire Pro using stock motor drivers and steppers?

Yes I can confirm the latest versions of Sapphires bot Pro and Plus have 20T pinions so rotation distance is 40.... for Z the pro uses 2mm lead screw and the plus 8mm lead so rotation distances are changed accordingly. 

I'm sorry if I made a mess but I'm trying to learn GitHub and how to do it in the correct way, I also wanted to clarify instructions at the beginning of the config for those who are not really familiar with klipper and especially with Robin nano required script for correct flashing.

thucar:
@laurienzu no harm done :) If there is a new version of Sapphire Pro out there, then it might be better to create a separate template for it and add its release year into the file name. Like this file has 2020 in its title.

PR #3278, opened 2020-09-03
Python3 support

KevinOConnor:
This branch tracks changes to the Klipper host code so that it can run on Python3.  It is intended for developers that wish to test and develop Klipper on Python3.

There are no immediate plans to switch Klipper from Python2 to Python3.

For those that do wish to test, it will require pulling this branch, installing a python3 virtualenv, and running the new code.  For example:
```
sudo service klipper stop
cd ~/klipper ; git fetch ; git checkout origin/work-python3-20200612
virtualenv -p /usr/bin/python3 ~/python3-env
~/python3-env/bin/pip install -r ~/klipper/scripts/klippy-requirements.txt

~/python3-env/bin/python ~/klipper/klippy/klippy.py ~/printer.cfg -l /tmp/klippy.log
```

If during testing you find an issue, please report as a comment on this PR.  Code fixes for Python3 should be submitted as PRs against this branch (work-python3-20200620).

-Kevin

P.S. Please don't comment here requesting we convert to Python3.  Please don't comment here to inform us of the pending doom of Python2.  We've known about the state of Python2 for over a decade.  We'll switch to Python3 if and when we feel that is appropriate - it will not be based on the schedules of any other projects.

Issue #14, opened 2017-06-09
Python 3


jeremyherbert:
Hi, 

This looks very interesting! Is python3 support planned?

Thanks,
Jeremy

KevinOConnor:
On Fri, Jun 09, 2017 at 03:52:07AM -0700, Jeremy Herbert wrote:
> This looks very interesting! Is python3 support planned?

Thanks.

In earlier revisions, I made the code Python3 compatible and ran some
benchmarks.  Unfortunately, those benchmarks showed that Python3 was
measurably slower than Python2.

Some additional Python2 only code has likely been introduced since
that time, but it should not be difficult to make the code support
Python3 again.  However, there does not appear to be a compelling
reason to use the slower Python3.

-Kevin

jeremyherbert:
Would you be happy for me to look into this further? Perhaps you tripped on a bug in python itself, because one would think that python3 should be better (not that I don't believe you of course). I find this host/real time slave paradigm very interesting; I've tried to do something like this in a custom CNC application before, but I unfortunately ran out of time. I'd like to use python3 due to some of the newer features (asyncio, etc) which I would like to use elsewhere, and of course because python2 only has bare minimum bugfix releases. 

It might also be useful to set up Travis to protect against any regressions, I can try to give this a go too.

KevinOConnor:
On Fri, Jun 09, 2017 at 07:12:48PM -0700, Jeremy Herbert wrote:
> Would you be happy for me to look into this further? Perhaps you
> tripped on a bug in python itself, because one would think that
> python3 should be better (not that I don't believe you of course). I
> find this host/real time slave paradigm very interesting; I've tried
> to do something like this in a custom CNC application before, but I
> unfortunately ran out of time. I'd like to use python3 due to some
> of the newer features (asyncio, etc) which I would like to use
> elsewhere, and of course because python2 only has bare minimum
> bugfix releases.

It's not difficult to get Klipper running on python3 - I just did it
as a quick test.  (At least enough porting to make batch processing of
gcode files work.)  If you're curious, I pushed the changes to a new
work-python3-20170610 branch.

Tests with python3 continue to show around a 10% slow down relative to
python2.  That's not a huge deal if there was something in python3
that was important to have.

-Kevin

jed-frey:
Could something like Numba be used do JIT compiling of the code? You could then really speed up things offloading to a GPU.

https://numba.pydata.org/

> That's not a huge deal if there was something in python3 that was important to have.

[...]

PR #5200, opened 2022-01-27
Update to run with python3

kpishere:
Added a header file (from linux) to aide in building of tests on mac.  Added pre-compiler switches to use said file with its adjustments for mac only on __APPLE__ builds.

Made the minimal changes to allow Klipper to run on Python3.  It passes all tests provided in repo thus far.

Note mac users:  Your environment must be adjusted such that /usr/bin/gcc and /usr/bin/g++ link to /usr/local/bin/gcc-11 and /usr/local/bin/g++-11 for this repo to build cleanly with compiler installed from homebrew. (SIP must be disabled so you can modify 'system' file system too ... or, adjust your path variables, possibly).

### DIFF
--- a/klippy/chelper/__init__.py
+++ b/klippy/chelper/__init__.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2016-2021  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
+from __future__ import absolute_import
 import os, logging
 import cffi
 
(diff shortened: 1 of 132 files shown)


kpishere:
Yeah, this isn't gonna pass your automated pipeline until it is configured to use python3.  Did notice the work of others in completed PRs for prepatory changes to python3.  Explains why this seemed so simple in this PR.

The change of course, needs to be planned for all.  In your hands.  Let me know if there is something I can do.  I'll personally be continuing to use this branch in my further testing.

dmbutyugin:
FYI: Klipper should already be able to run with Python3 [#3278]. You can see further references [here](https://github.com/Klipper3d/klipper/issues/14#issuecomment-936854879). It might still be the case that some regressions are present, but it is not very likely they are at that scale.

As for the mac build - the changes should better be sent as a separate PR, so that they can be evaluated and reviewed separately.

kpishere:
this PR can be abandoned.  I've made a new one for the mac.  I'm not gonna worry about this python3 thing.  I'd rather put my energy into Rust rather than Python anyhow.

PR #5206, opened 2022-01-29
Update Resonance_Compensation.md to recommend SET_VELOCITY_LIMIT

KevinOConnor:
I noticed that Resonance_Compensation.md was still recommending updating the printer.cfg file to set max_accel=7000.  It's simpler to make all those adjustments by issuing a `SET_VELOCITY_LIMIT` command.  This PR updates the document.

@dmbutyugin - FYI.

-Kevin

### DIFF
--- a/docs/Resonance_Compensation.md
+++ b/docs/Resonance_Compensation.md
@@ -23,9 +23,8 @@ of the stealthChop mode of Trinamic stepper drivers.
 
 ## Tuning
 
-Basic tuning requires measuring the ringing frequencies of the printer and
-adding a few parameters to `printer.cfg` file.
-
+Basic tuning requires measuring the ringing frequencies of the printer
+by printing a test model.
 
 Slice the ringing test model, which can be found in
 [docs/prints/ringing_tower.stl](prints/ringing_tower.stl), in the slicer:
@@ -45,31 +44,28 @@ Slice the ringing test model, which can be found in
 
 First, measure the **ringing frequency**.
 
-1. Increase `max_accel` and `max_accel_to_decel` parameters in your
-   `printer.cfg` to 7000. Note that this is only needed for tuning, and more
-   proper value will be selected in the corresponding
-   [section](#selecting-max_accel).
-2. If `square_corner_velocity` parameter was changed, revert it back to 5.0.
-   It is not advised to increase it when using the input shaper because it can
-   cause more smoothing in parts - it is better to use higher acceleration
-   value instead.
-3. Restart the firmware: `RESTART`.
-4. Disable Pressure Advance: `SET_PRESSURE_ADVANCE ADVANCE=0`.
-5. If you have already added `[input_shaper]` section to the printer.cfg,
+1. If `square_corner_velocity` parameter was changed, revert it back
+   to 5.0. It is not advised to increase it when using input shaper
+   because it can cause more smoothing in parts - it is better to use
+   higher acceleration value instead.
+2. Increase `max_accel_to_decel` by issuing the following command:
+   `SET_VELOCITY_LIMIT ACCEL_TO_DECEL=7000`
+3. Disable Pressure Advance: `SET_PRESSURE_ADVANCE ADVANCE=0`
+4. If you have already added `[input_shaper]` section to the printer.cfg,
    execute `SET_INPUT_SHAPER SHAPER_FREQ_X=0 SHAPER_FREQ_Y=0` command. If you
    get "Unknown command" error, you can safely ignore it at this point and
    continue with the measurements.
-6. Execute the command
-   `TUNING_TOWER COMMAND=SET_VELOCITY_LIMIT PARAMETER=ACCEL START=1250 FACTOR=100 BAND=5`.
+5. Execute the command:
+   `TUNING_TOWER COMMAND=SET_VELOCITY_LIMIT PARAMETER=ACCEL START=1500 STEP_DELTA=500 STEP_HEIGHT=5`
    Basically, we try to make ringing more pronounced by setting different large
    values for acceleration. This command will increase the acceleration every
    5 mm starting from 1500 mm/sec^2: 1500 mm/sec^2, 2000 mm/sec^2, 2500 mm/sec^2
    and so forth up until 7000 mm/sec^2 at the last band.
-7. Print the test model sliced with the suggested parameters.
-8. You can stop the print earlier if the ringing is clearly visible and you see
+6. Print the test model sliced with the suggested parameters.
+7. You can stop the print earlier if the ringing is clearly visible and you see
    that acceleration gets too high for your printer (e.g. printer shakes too
    much or starts skipping steps).
-9. Use X and Y marks at the back of the model for reference. The measurements
+8. Use X and Y marks at the back of the model for reference. The measurements
    from the side with X mark should be used for X axis *configuration*, and
    Y mark - for Y axis configuration. Measure the distance *D* (in mm) between
    several oscillations on the part with X mark, near the notches, preferably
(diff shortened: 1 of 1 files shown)


docs/Resonance_Compensation.md
```diff
@@ -79,14 +72,14 @@ First, measure the **ringing frequency**.
...
-    velocity, so the frequency is 100 * 6 / 12.14 ≈ 49.4 Hz.
-12. Do (9) - (11) for Y mark as well.
+8. Count how many oscillations *N* the measured distance *D* corresponds to.
+   If you are unsure how to count the oscillations, refer to the picture
+   above, which shows *N* = 6 oscillations.
+9. Compute the ringing frequency of X axis as *V* &middot; *N* / *D* (Hz),
+   where *V* is the velocity for outer perimeters (mm/sec). For the example
+   above, we marked 6 oscillations, and the test was printed at 100 mm/sec
+   velocity, so the frequency is 100 * 6 / 12.14 ≈ 49.4 Hz.
+10. Do (9) - (11) for Y mark as well.
```
dmbutyugin, line 82: This should read 'Do (7) - (9) for Y mark as well.' now (or different numbers, if the square_corner_velocity point is kept).

docs/Resonance_Compensation.md
```diff
@@ -45,31 +44,25 @@ Slice the ringing test model, which can be found in
...
+   `SET_VELOCITY_LIMIT ACCEL_TO_DECEL=7000 SQUARE_CORNER_VELOCITY=5`
+2. Disable Pressure Advance: `SET_PRESSURE_ADVANCE ADVANCE=0`
+3. If you have already added `[input_shaper]` section to the printer.cfg,
    execute `SET_INPUT_SHAPER SHAPER_FREQ_X=0 SHAPER_FREQ_Y=0` command. If you
    get "Unknown command" error, you can safely ignore it at this point and
    continue with the measurements.
-6. Execute the command
-   `TUNING_TOWER COMMAND=SET_VELOCITY_LIMIT PARAMETER=ACCEL START=1250 FACTOR=100 BAND=5`.
+4. Execute the command:
+   `TUNING_TOWER COMMAND=SET_VELOCITY_LIMIT PARAMETER=ACCEL START=1250 FACTOR=100 BAND=5`
```
dmbutyugin, line 56: I see that TUNING_TOWER has now the STEP_DELTA and STEP_HEIGHT parameters. Would it make sense to update this command here and below in other instances to use them instead? I think we get countless questions on discord where people are confused and ask if the command is wrong. It will read

`TUNING_TOWER COMMAND=SET_VELOCITY_LIMIT PARAMETER=ACCEL START=1500 STEP_DELTA=100 STEP_HEIGHT=5`

if I am not mistaken

docs/Resonance_Compensation.md
```diff
@@ -45,31 +44,25 @@ Slice the ringing test model, which can be found in
 
 First, measure the **ringing frequency**.
 
-1. Increase `max_accel` and `max_accel_to_decel` parameters in your
-   `printer.cfg` to 7000. Note that this is only needed for tuning, and more
-   proper value will be selected in the corresponding
-   [section](#selecting-max_accel).
-2. If `square_corner_velocity` parameter was changed, revert it back to 5.0.
```
dmbutyugin, line 52: Note that the way instructions go, it was suggested to set square_corner_velocity back to 5.0 and not change it back after tuning. If we remove this sentence, and instead suggest to set it via `SET_VELOCITY_LIMIT SQUARE_CORNER_VELOCITY=5`, then the testing will be done with one value, but the real printing will be done with a different value. This may be undesirable. So, maybe it makes sense to keep this point? Or mention it earlier, in the non-numbered part of the section. Alternatively, we should probably not suggest to reset SQUARE_CORNER_VELOCITY. However my estimations (via calibrate_shaper.py) indicate that having higher SQUARE_CORNER_VELOCITY very quickly increases the smoothing, such that SQUARE_CORNER_VELOCITY==10 will already create too much smoothing even with rather high resonance frequencies and low accelerations, this is why the suggestion was to keep SQUARE_CORNER_VELOCITY low.

docs/Resonance_Compensation.md
```diff
@@ -150,16 +143,16 @@ For most of the printers, either MZV or EI shapers can be recommended. This
...
 
-1. Restart the firmware: `RESTART`.
-2. Disable Pressure Advance: `SET_PRESSURE_ADVANCE ADVANCE=0`.
-3. Execute `SET_INPUT_SHAPER SHAPER_TYPE=MZV`.
-4. Execute the command
-   `TUNING_TOWER COMMAND=SET_VELOCITY_LIMIT PARAMETER=ACCEL START=1250 FACTOR=100 BAND=5`.
-5. Print the test model sliced with the suggested parameters.
+1. Restart the firmware: `RESTART`
+2. Prepare for test:
+   `SET_VELOCITY_LIMIT ACCEL_TO_DECEL=7000 SQUARE_CORNER_VELOCITY=5`
```
dmbutyugin, line 150: Following the doc, this printed model will also be used later to choose max_accel. Therefore, I would suggest not to override SQUARE_CORNER_VELOCITY at least for this model (and maybe, not anywhere in the testing via this command, see my other comment above), as the results will be different from the real-life printing.

docs/Resonance_Compensation.md
```diff
@@ -335,26 +324,26 @@ accelerometer and measure the resonances with it (refer to the
...
+For tuning, add empty `[input_shaper]` section to your
+`printer.cfg`. Then, assuming that you have sliced the ringing model
+with suggested parameters, print the test model 3 times as
+follows. First time, prior to printing, run
 
 1. `RESTART`
-2. `SET_PRESSURE_ADVANCE ADVANCE=0`.
-3. `SET_INPUT_SHAPER SHAPER_TYPE=2HUMP_EI SHAPER_FREQ_X=60 SHAPER_FREQ_Y=60`.
-4. `TUNING_TOWER COMMAND=SET_VELOCITY_LIMIT PARAMETER=ACCEL START=1250 FACTOR=100 BAND=5`.
+2. `SET_VELOCITY_LIMIT ACCEL_TO_DECEL=7000 SQUARE_CORNER_VELOCITY=5`
```
dmbutyugin, line 333: Actually, here too: it is better to not override SQUARE_CORNER_VELOCITY here.

dmbutyugin (commented):
Thank you for doing this, Kevin! Now indeed there is no need to change acceleration parameters via the config. I only had a few comments about the new tuning process.

KevinOConnor:
Thanks.

I reverted the square_corner_velocity changes.  I left in the original suggestion to not set it above 5.0 and the various tests no longer reference it.

I updated the TUNING_TOWER commands to use STEP_HEIGHT/STEP_DELTA.  BTW, I think it should be `STEP_DELTA=500`.  There is still one TUNING_TOWER command using FACTOR - I wasn't quite sure how that one is supposed to be set.

I found a few places using list numbering and attempted to fix them up.

Cheers,
-Kevin

dmbutyugin:
Thank you, Kevin!

> I updated the TUNING_TOWER commands to use STEP_HEIGHT/STEP_DELTA. BTW, I think it should be `STEP_DELTA=500`.

Yes, right, my bad.

> There is still one TUNING_TOWER command using FACTOR - I wasn't quite sure how that one is supposed to be set.

Yes, that one is in 'Fine-tuning resonance frequencies' section. There are explicit formulas to calculate the parameters there and how to interpret the results, so I think that one is fine. And TBH, that section was written before adxl345 support for resonance testing was added. Nowadays, I doubt that method is useful - it feels more like a waste of time. Perhaps we should delete that section eventually.

> I found a few places using list numbering and attempted to fix them up.

Yep, I didn't notice any other numbering problems either.

dmbutyugin (approved):

KevinOConnor:
Thanks for reviewing.  I committed this change.

> Perhaps we should delete that section eventually.

Yeah - the adxl345 mechanism is generally easier and more robust.  At some point it might make sense to simplify Resonance_Compensation.md to only have a simple calibration guide and point users to the adxl345.

-Kevin

PR #5207, opened 2022-01-29
Initial Support Ultimaker 2+

drphil3d:
Tested config profile for Ultimaker 2+ with ultipanel lcd based on generic-ultimaker-ultimainboard-v2.cfg


drphil3d:
Does this look good?
@KevinOConnor

dingyifei:
You need to modify the commit title and sign off line based on the documentation

PR #5210, opened 2022-01-30
Support reversing extruder direction and more

KevinOConnor:
This PR adds a new `SET_EXTRUDER_ROTATION_DISTANCE` command.  It supports specifying a "negative rotation distance" which will have the effect of reversing the extruder stepper direction.  This may be useful for printers with a single extruder motor that can switch between two hotends via a servo.

I do not have a good way to test this PR.  Feedback would be appreciated.

This PR is a continuation of PR #5143.  This PR also deprecates the `shared_heater` config option, deprecates the `SYNC_STEPPER_TO_EXTRUDER` command, and deprecates the `SET_EXTRUDER_STEP_DISTANCE` command.  There are new replacements for all of these features.

This PR was built on top of PR #5209.

@drumclock , @tircown - FYI.

-Kevin


DrumClock:
Hi, @KevinOConnor 
**I will try to provide feedback as soon as possible.**

DrumClock:
Hi @KevinOConnor 

### **TEST: SYNC_EXTRUDER_MOTION**

1/ `ACTIVATE_EXTRUDER EXTRUDER = extruder`
2/` SYNC_EXTRUDER_MOTION EXTRUDER = extruder1 MOTION_QUEUE = extruder`
3/ `G1 E10` - both extruders work together

4/ `ACTIVATE_EXTRUDER EXTRUDER = extruder1`
5/` G1 E10` - no extruder works

6/ `SYNC_EXTRUDER_MOTION EXTRUDER = extruder 1 MOTION_QUEUE = extruder1`
7/ `G1 E10` - extruder 1 works independently
8/ `ACTIVATE_EXTRUDER EXTRUDER = extruder`
9/ `G1 E10 `- extruder works independently

⚠️   **ATTENTION**  ⚠️ 
if in `SYNC_EXTRUDER_MOTION mode EXTRUDER=extruder1 MOTION_QUEUE=extruder `
I use the command `SET_EXTRUDER_ROTATION_DISTANCE` I can change the direction of rotation or the step ratio but always only for the given extruder. So the result is that both extruders work together but each in a different direction and at a different speed (stepping) 

⚠️   **ATTENTION**  ⚠️
`SYNC_EXTRUDER_MOTION EXTRUDER=extruder1 MOTION_QUEUE= `
`extruder1`  is desynchronized but for  `ACTIVATE_EXTRUDER EXTRUDER = extruder1`
does not respond to `G1 E10 `    -------  It is a mistake ? or so it should be.... 😱


### **TEST: SET_EXTRUDER_ROTATION_DISTANCE**

1 /` SET_EXTRUDER_ROTATION_DISTANCE EXTRUDER = extruder DISTANCE = 33.500`
... printer.configfile.settings ['extruder']. rotation_distance = 33.5
... printer.configfile.config ['extruder']. rotation_distance = 33.500

2 / `SET_EXTRUDER_ROTATION_DISTANCE EXTRUDER = extruder DISTANCE = -33.500`
... printer.configfile.settings ['extruder']. rotation_distance = 33.5
... printer.configfile.config ['extruder']. rotation_distance = 33.500

When I change the `rotation_distance`, **the new value is not saved anywhere**, so it is not possible to find out the currently set `rotation_distance` for the given `extruder` 
This could be a problem when creating a custom` [gcode_macro]` without using `variable_ `to determine the current value.

### Is there anything else I need to test?

### **NOTE:**
So far, I haven't figured out how to use **the same stepper motor** for `extruder `and `extruder1`

KevinOConnor:
Thanks.

> if in SYNC_EXTRUDER_MOTION mode EXTRUDER=extruder1 MOTION_QUEUE=extruder
I use the command SET_EXTRUDER_ROTATION_DISTANCE I can change the direction of rotation or the step ratio but always only for the given extruder. So the result is that both extruders work together but each in a different direction and at a different speed (stepping)

This was the intended behavior.  The SET_EXTRUDER_ROTATION_DISTANCE and SYNC_EXTRUDER_MOTION work on the extruder stepper motor specified in the command.

> SYNC_EXTRUDER_MOTION EXTRUDER=extruder1 MOTION_QUEUE=
extruder1 is desynchronized but for ACTIVATE_EXTRUDER EXTRUDER = extruder1
does not respond to G1 E10 ------- It is a mistake ? or so it should be...

This was the intended behavior.  Each extruder object has a "stepper motor" and a "motion queue".  The SYNC_EXTRUDER_MOTION command alters the behavior of the stepper motor, while ACTIVATE_EXTRUDER command activates a particular motion queue.  So, if you have a motion queue with no steppers associated with it, then no movement will be observed.

> When I change the rotation_distance, the new value is not saved anywhere

This info can be exported if desired.

> So far, I haven't figured out how to use the same stepper motor for extruder and extruder1

I'm not sure what your hardware is, so it's hard for me to say.  If you've got a single extruder stepper motor controlling two filament drives, then I'd guess you would define only one `[extruder]` object.  Then define `T0` and `T1` macros - on a toolhead switch alter the servo setting and issue a SET_EXTRUDER_ROTATION_DISTANCE to reverse the direction of motion.

-Kevin

DrumClock:
Thanks,
I'm testing all the combinations that occurred to me, so I described the behavior.  It's up to you to evaluate it
 ..... I'm just a beta tester :)

> I'm not sure what your hardware is, so it's hard for me to say. If you've got a single extruder stepper motor controlling two filament drives, then I'd guess you would define only one `[extruder]` object. Then define `T0` and `T1` macros - on a toolhead switch alter the servo setting and issue a SET_EXTRUDER_ROTATION_DISTANCE to reverse the direction of motion.

Yes, now I use the definition` [extruder]` for HotEnd T0 and `[heater_generic]` for HotEnd T1 .... unfortunately it is not a full-fledged solution for T1 because for T1 it is not possible to use eg these parameters:
`min_extrude_temp`  `pressure_advance` `max_extrude_cross_section`

the `servo `simultaneously switches the position of the HotEnd and the fiber pressure on the stepper extruder 

Video of how SWITCHING EXTRUDER works:  https://youtu.be/BOaIMiyzaks

MFG7

KevinOConnor:
> Yes, now I use the definition [extruder] for HotEnd T0 and [heater_generic] for HotEnd T1 .... unfortunately it is not a full-fledged solution for T1 because for T1 it is not possible to use eg these parameters:
min_extrude_temp pressure_advance max_extrude_cross_section

I don't understand the issue.  There isn't a `min_extrude_temp` safety check for T1, but it should be okay if you're careful.  For `pressure_advance` the T1 macro can call `SET_PRESSURE_ADVANCE` to specify what is needed when switching filament.  The `max_extrude_cross_section` is just a safety check - set it to the maximum value that you need for both filaments.

That said, this PR just adds the ability to reverse the extruder stepper motor at runtime via `SET_EXTRUDER_ROTATION_DISTANCE`.  It doesn't add any particular features beyond that.

-Kevin

DrumClock:
Yes this PR is not about my configuration ..... Forget about my problem ...

KevinOConnor:
Okay, thanks for testing and providing feedback.

I don't own one of these "multi-filament extruders", so I don't really know what is useful or not useful.  I added support for reversing the direction of the extruder motor at runtime because I thought it may be useful.  If it isn't useful though, I'll rethink that strategy.

Thanks again,
-Kevin

DrumClock:
**I'm very happy to help with testing.**

Reversing the direction of the extruder motor at runtime is definitely a useful thing as well as synchronizing the extruder.

I can imagine the same for any stepper (X, Y, Z etc).
I'm already thinking of controls for dual_x and dual_y os when there will be only one extruder (or more) on each ..... but these are just my visions.

PR #5213, opened 2022-01-30
Elegoo Neptune 2 config update

just-trey:
This pin mapping on the 2 boards available for this machine are identical so I made some adjustments. 

- Renamed file to remove version-specific information
- Updated the comment section to explain `make` differences depending on the version for the board
- Added comment on `rotation_distance` to provide the setting for the 2s version (dual extruder)
- Added `bed_screws` section to align the score locations on the bed

Signed-off-by: Trey Wentz <trey.wentz@gmail.com>


ld3300 (commented):
Hi @just-trey,
Can you clarify why max_extrude_only_distance has been increased to 200?
Best,
-David

tuckahoe-little-league:
> 

Sure, simple answer. I felt  100 was too short. I'm actually not that emphatic about it and will actually change it back.  I see the enders are set to 100 so 100 makes sense

just-trey:
@ld3300 

Hi David, no real good reason other then I wanted to be able to extrude more when loading filament. So personal preference. I went ahead and put it back though after I saw the enders had 100. That made sense. Thanks for the question and making me think it through.

KevinOConnor:
Thanks.

-Kevin

PR #5214, opened 2022-01-31
flash_usb: use sudo for rp2040 flashing

dalegaard:
The rp2040 can be flashed without sudo when using udev rules to give the user permission, but in a standard configuration sudo is required.

Here we make it possible for flash_usb to use sudo for the rp2040 target, and make it the default when using `make flash` for the rp2040. As for other targets, one can set `NOSUDO=1` to not call through sudo.

When I did the `make flash` support I was using a machine with udev rules, and thus never noticed that root is required in the common case.


KevinOConnor:
Thanks.

-Kevin

PR #5217, opened 2022-01-31
Multiple steppers on delta tower

TypQxQ:
Changes and sample configuration to make it possible to use multiple steppers on each Delta tower.
Instead of using a stepper motor and a idler, with this one can use two stepper motors, each controlled by it's own stepper.

I have tried this out using two SKR 1.4 Turbo cards with TMC2209.
Tower A and B steppers and endstop on first micro-controller and tower C and it's endstop on the second micro-controller.

This is how it looks like in action: https://youtu.be/53Ut1UPKveQ

**Benefits** of using a stepper instead of a pulley is shortening the stretchable belt length.

**Benefits** of using a stepper driver for each stepper to serial or parallel gives you the maximum speed because of the lower back EMF and higher torque because of them working together on the same belt.

To run a motor of a single stepper you can connect them in parallel or series and is not desirable:

Running them in parallel would run them at half the voltage each.
To achieve high torque at high speed you need a higher voltage than the generated back-EMF.
That's why I believe the Duet3 and BTT Octopus Pro have high voltage drivers.

Running them in series, doubling the current for the motors to draw from, would give you more inductance and resistance, thus even more generated back-EMF. I've read that it gives you 4 times more inductance than running in parallel but don't quote me on that. So basically the back-EMF makes you loose power even sooner than in parallel.

I think most boards use series for dual Z. This is fine for Z where you don't run high speeds and need high holding torque, not for directional changes and acceleration.

I've used the RRF EMF Calculator to try different combinations here: https://www.reprapfirmware.org/emf.html

**This is a duplicate of #4884** because that timed out and was automatically closed.

Signed-off-by: Andrei Ignat andrei@ignat.se


lorenzing:
Soon I want to try it. Thank you for this feature

KevinOConnor:
Thanks.  At a high-level this change looks fine.  However, there are unrelated changes to delta.py that I'm unsure why were made.  As near as I can tell the only change should have been converting the calls to `PrinterRail()` to `LookupMultiRail()` in that file. (The changes to stepper.py look fine.)

-Kevin

TypQxQ:
@KevinOConnor Thankyou. I have now removed the mentioned changes.

KevinOConnor:
Thanks.

-Kevin

PR #5221, opened 2022-02-01
Update and rename printer-twotrees-sapphire-plus-2020.cfg to printer-…

laurienzu:
…twotrees-sapphire-plus-V1.1.cfg

Updated for latest printer models sold, please accept changes since this is the most popular model at the moment


laurienzu:
> Hey @laurienzu, are the config changes for a new model, or was the 2020 version of the printer incorrect? Should this be added as a new config example for the new machine version?
> 
> Please be sure to also add the Signed-off-by: line as [documented](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages) Best, -David

There are a lot of models and yes this is a new model (the latest one) so maybe is better to add it as a new config, thanks :)

Issue #5189, opened 2022-01-23
BUG?: Chamber Heater with DS18B20 Temperature Sensor generates "MCU 'mcu' shutdown: Scheduled digital out event will exceed max_duration" failure


Markus-1984:
Hi,

I´m not able to get the Chamber Heater working with the DS18B20 Sensor.
If I set the DS18B20 only as a temperature sensor it works well:

[temperature_sensor chamber]

gcode_id: CHAMBER
sensor_type: DS18B20
sensor_mcu: rpi
serial_no: 28-3c01f095187a
ds18_report_time: 2

If I set it in the generic heater section like this:

[heater_generic chamber]
heater_pin: PA3
sensor_type: DS18B20
serial_no: 28-3c01f095187a
ds18_report_time: 2
sensor_mcu: rpi
min_temp: 0
max_temp: 100
#control: watermark
#control: pid
#pid_kp: 100.0
#pid_ki: 1.0
#pid_kd: 1.0
gcode_id: CHAMBER

 It will generate a "MCU 'mcu' shutdown: Scheduled digital out event will exceed max_duration" error

As I started with the DS18B20 sensor, the default value for the ds18_report_time: 1 was to low, and generates the same error.
So I´ve increased it to 2 and so it works well.

Referring to https://github.com/Klipper3d/klipper/issues/1641#issuecomment-494381609 you wrote that the temp sensors are sampled every 0.3 seconds. Is it possible that klipper ignores the ds18_report_time: 2 in the generic_heater section and this would be the reason for this error?

You can see it in the attached LOG that the DS18B20 is the reason for the error, and it´s the same error if I set the ds18_report_time to 1.

[klippy.log](https://github.com/Klipper3d/klipper/files/7921272/klippy.log)

Used Hardware:
RPi 3B
BTT Octopus V1.0

Sineos:
Default for `ds18_report_time` is but the allowed range is 1 to 3.
Otherwise maybe @theopensourcerer or @akshaal could provide some insight

Markus-1984:
Good to know, but for me 1 was not working. Only after I have increased it to 2, the sensor works as a temperature sensor. But not as a sensor for a heater. That's the reason why I think the generic_heater will ignore the statement for the longer sample time.

charredchar:
[...]

Issue #4874, opened 2021-11-04
BME680 gas sensor i2c_read_response error handling


falo2k:
Just had a print crash out at 90% with a complaint about reading i2c data.  Log file attached, with snippet below of unhandled exception - seems to be related to my BME680 gas sensor.  I would imagine this could be handled gracefully without impacting an ongoing print.  It looks like there was a previous issue on i2c exceptions that fizzled #4708.

```
Unhandled exception during run
Traceback (most recent call last):
  File "/home/pi/klipper/klippy/klippy.py", line 201, in run
    self.reactor.run()
  File "/home/pi/klipper/klippy/reactor.py", line 269, in run
    g_next.switch()
  File "/home/pi/klipper/klippy/reactor.py", line 310, in _dispatch_loop
    timeout = self._check_timers(eventtime, busy)
  File "/home/pi/klipper/klippy/reactor.py", line 156, in _check_timers
    t.waketime = waketime = t.callback(eventtime)
  File "/home/pi/klipper/klippy/extras/bme280.py", line 281, in _sample_bme680
    res_heat_0 = self._calculate_gas_heater_resistance(self.gas_heat_temp)
  File "/home/pi/klipper/klippy/extras/bme280.py", line 421, in _calculate_gas_heater_resistance
    heater_data = self.read_register('RES_HEAT_VAL', 3)
  File "/home/pi/klipper/klippy/extras/bme280.py", line 455, in read_register
    params = self.i2c.i2c_read(regs, read_len)
  File "/home/pi/klipper/klippy/extras/bus.py", line 182, in i2c_read
    return self.i2c_read_cmd.send([self.oid, write, read_len])
  File "/home/pi/klipper/klippy/mcu.py", line 501, in send
    return self._do_send([self._cmd.encode(data)], minclock, reqclock)
  File "/home/pi/klipper/klippy/mcu.py", line 499, in _do_send
    raise self._error(str(e))
CommandError: Unable to obtain 'i2c_read_response' response

```

[klippy.log.i2c_read.zip](https://github.com/Klipper3d/klipper/files/7478189/klippy.log.i2c_read.zip)

simonkuehling:
In most cases, communication faults on the i2c bus are a hardware related issue - like noise on the lines, too long cables or otherwise poor signal quality.
Can you tell a little about your electronics setup and wiring of the sensor?

falo2k:
Sure, happy to share.  Details of the setup are below.

[...]

PR #5233, opened 2022-02-04
DS18B20: Optionally allow read errors

functionpointer:
My Voron would fail because my DS18B20 chamber temp sensor has intermittent failures, even when powered by 5V.
So i made klipper tolerate the error.

This PR adds a configuration option that allows read errors from DS18B20.
When activated, read errors won't emergency-stop the printer.
Instead it will just report 0C.

The PR also improves error reporting of the sensor a little bit.

### DIFF
--- a/klippy/extras/ds18b20.py
+++ b/klippy/extras/ds18b20.py
@@ -10,6 +10,7 @@ DS18_REPORT_TIME = 3.0
 # Temperature can be sampled at any time but conversion time is ~750ms, so
 # setting the time too low will not make the reports come faster.
 DS18_MIN_REPORT_TIME = 1.0
+DS18_MAX_CONSECUTIVE_ERRORS = 4
 
 
 class DS18B20:
@@ -31,7 +32,10 @@ class DS18B20:
 
     def _build_config(self):
         sid = "".join(["%02x" % (x,) for x in self.sensor_id])
-        self._mcu.add_config_cmd("config_ds18b20 oid=%d serial=%s" % (self.oid, sid))
+        self._mcu.add_config_cmd(
+            "config_ds18b20 oid=%d serial=%s"
+            " max_error_count=%d" % (self.oid, sid, DS18_MAX_CONSECUTIVE_ERRORS)
+        )
 
         clock = self._mcu.get_query_slot(self.oid)
         self._report_clock = self._mcu.seconds_to_clock(self.report_time)
@@ -51,7 +55,9 @@ class DS18B20:
     def _handle_ds18b20_response(self, params):
         temp = params["value"] / 1000.0
 
-        if temp < self.min_temp or temp > self.max_temp:
+        if params["fault"] != 0:
+            temp = 0  # read error! report 0C and don't check temp range
+        elif temp < self.min_temp or temp > self.max_temp:
             self.printer.invoke_shutdown(
                 "DS18B20 temperature %0.1f outside range of %0.1f:%.01f"
                 % (temp, self.min_temp, self.max_temp)
--- a/klippy/extras/ds18b20.py
+++ b/klippy/extras/ds18b20.py
@@ -11,6 +11,7 @@ DS18_REPORT_TIME = 3.0
 # setting the time too low will not make the reports come faster.
 DS18_MIN_REPORT_TIME = 1.0
 
+
 class DS18B20:
     def __init__(self, config):
         self.printer = config.get_printer()
@@ -19,39 +20,46 @@ class DS18B20:
         self.temp = self.min_temp = self.max_temp = 0.0
         self._report_clock = 0
         self.report_time = config.getfloat(
-            'ds18_report_time',
-            DS18_REPORT_TIME,
-            minval=DS18_MIN_REPORT_TIME
+            "ds18_report_time", DS18_REPORT_TIME, minval=DS18_MIN_REPORT_TIME
         )
-        self._mcu = mcu.get_printer_mcu(self.printer, config.get('sensor_mcu'))
+        self._mcu = mcu.get_printer_mcu(self.printer, config.get("sensor_mcu"))
         self.oid = self._mcu.create_oid()
-        self._mcu.register_response(self._handle_ds18b20_response,
-            "ds18b20_result", self.oid)
+        self._mcu.register_response(
+            self._handle_ds18b20_response, "ds18b20_result", self.oid
+        )
         self._mcu.register_config_callback(self._build_config)
 
     def _build_config(self):
         sid = "".join(["%02x" % (x,) for x in self.sensor_id])
-        self._mcu.add_config_cmd("config_ds18b20 oid=%d serial=%s"
-                                 % (self.oid, sid))
+        self._mcu.add_config_cmd("config_ds18b20 oid=%d serial=%s" % (self.oid, sid))
 
         clock = self._mcu.get_query_slot(self.oid)
         self._report_clock = self._mcu.seconds_to_clock(self.report_time)
-        self._mcu.add_config_cmd("query_ds18b20 oid=%d clock=%u rest_ticks=%u"
-            " min_value=%d max_value=%d" % (
-                self.oid, clock, self._report_clock,
-                self.min_temp * 1000, self.max_temp * 1000), is_init=True)
+        self._mcu.add_config_cmd(
+            "query_ds18b20 oid=%d clock=%u rest_ticks=%u"
+            " min_value=%d max_value=%d"
+            % (
+                self.oid,
+                clock,
+                self._report_clock,
+                self.min_temp * 1000,
+                self.max_temp * 1000,
+            ),
+            is_init=True,
+        )
 
     def _handle_ds18b20_response(self, params):
-        temp = params['value'] / 1000.0
+        temp = params["value"] / 1000.0
 
         if temp < self.min_temp or temp > self.max_temp:
             self.printer.invoke_shutdown(
                 "DS18B20 temperature %0.1f outside range of %0.1f:%.01f"
-                % (temp, self.min_temp, self.max_temp))
+                % (temp, self.min_temp, self.max_temp)
+            )
 
-        next_clock      = self._mcu.clock32_to_clock64(params['next_clock'])
+        next_clock = self._mcu.clock32_to_clock64(params["next_clock"])
         last_read_clock = next_clock - self._report_clock
-        last_read_time  = self._mcu.clock_to_print_time(last_read_clock)
+        last_read_time = self._mcu.clock_to_print_time(last_read_clock)
         self._callback(last_read_time, temp)
 
     def setup_minmax(self, min_temp, max_temp):
@@ -69,8 +77,9 @@ class DS18B20:
 
     def get_status(self, eventtime):
         return {
-            'temperature': round(self.temp, 2),
+            "temperature": round(self.temp, 2),
         }
+
 
 def load_config(config):
     # Register sensor
--- a/src/linux/sensor_ds18b20.c
+++ b/src/linux/sensor_ds18b20.c
@@ -181,7 +181,8 @@ fail4:
 fail5:
     shutdown("Could not start DS18B20 reader thread");
 }
-DECL_COMMAND(command_config_ds18b20, "config_ds18b20 oid=%c serial=%*s");
+DECL_COMMAND(command_config_ds18b20,
+             "config_ds18b20 oid=%c serial=%*s");
 
 void
 command_query_ds18b20(uint32_t *args)
--- a/src/linux/sensor_ds18b20.c
+++ b/src/linux/sensor_ds18b20.c
@@ -13,7 +13,6 @@
 #include <time.h> // clock_gettime
 #include "basecmd.h" // oid_alloc
 #include "board/irq.h" // irq_disable
-#include "board/misc.h" // output
 #include "command.h" // DECL_COMMAND
 #include "internal.h" // report_errno
 #include "sched.h" // DECL_SHUTDOWN
@@ -51,17 +50,23 @@ struct ds18_s {
     int temperature;
     struct timespec request_time;
     uint8_t status;
-    const char* error;
+    uint8_t error_count;
+    uint8_t max_error_count;
 };
 
-// Lock ds18_s mutex, set error status and message, unlock mutex.
+// Lock ds18_s mutex, set error status, unlock mutex.
 static void
-locking_set_read_error(struct ds18_s *d, const char *error)
+locking_handle_read_error(struct ds18_s *d)
 {
     pthread_mutex_lock(&d->lock);
-    d->error = error;
     d->status = W1_ERROR;
-    pthread_mutex_unlock(&d->lock);
+    d->error_count++;
+    if (d->error_count <= d->max_error_count) {
+        pthread_mutex_unlock(&d->lock);
+    } else {
+        pthread_mutex_unlock(&d->lock);
+        pthread_exit(NULL);
+    }
 }
 
 // The kernel interface to DS18B20 sensors is a sysfs entry that blocks for
@@ -89,15 +94,14 @@ reader_start_routine(void *param) {
         int ret = read(d->fd, data, sizeof(data)-1);
         if (ret < 0) {
             report_errno("read DS18B20", ret);
-            locking_set_read_error(d, "Unable to read DS18B20");
-            pthread_exit(NULL);
+            locking_handle_read_error(d);
+            continue;
         }
         data[ret] = '\0';
         char *temp_string = strstr(data, "t=");
         if (temp_string == NULL || temp_string[2] == '\0') {
-            locking_set_read_error(d,
-            "Unable to find temperature value in DS18B20 report");
-            pthread_exit(NULL);
+            locking_handle_read_error(d);
+            continue;
         }
         // Don't pass 't' and '=' to atoi
         temp_string += 2;
@@ -113,8 +117,8 @@ reader_start_routine(void *param) {
         ret = lseek(d->fd, 0, SEEK_SET);
         if (ret < 0) {
             report_errno("seek DS18B20", ret);
-            locking_set_read_error(d, "Unable to seek DS18B20");
-            pthread_exit(NULL);
+            locking_handle_read_error(d);
+            continue;
         }
     }
     pthread_exit(NULL);
@@ -151,6 +155,8 @@ command_config_ds18b20(uint32_t *args)
     }
 
     struct ds18_s *d = oid_alloc(args[0], command_config_ds18b20, sizeof(*d));
+    d->max_error_count = args[3];
+    d->error_count = 0;
     d->timer.func = ds18_event;
     d->fd = fd;
     d->status = W1_IDLE;
@@ -182,7 +188,7 @@ fail5:
     shutdown("Could not start DS18B20 reader thread");
 }
 DECL_COMMAND(command_config_ds18b20,
-             "config_ds18b20 oid=%c serial=%*s");
+             "config_ds18b20 oid=%c serial=%*s max_error_count=%c");
 
 void
 command_query_ds18b20(uint32_t *args)
@@ -216,12 +222,15 @@ ds18_send_and_request(struct ds18_s *d, uint32_t next_begin_time, uint8_t oid)
 
     pthread_mutex_lock(&d->lock);
     if (d->status == W1_ERROR) {
-        // try_shutdown expects a static string. Output the specific error,
-        // then shut down with a generic error.
-        output("Error: %s", d->error);
-        pthread_mutex_unlock(&d->lock);
-        try_shutdown("Error reading DS18B20 sensor");
-        return;
+        if (d->error_count > d->max_error_count) {
+            try_shutdown("Error reading DS18B20 sensor");
+            pthread_mutex_unlock(&d->lock);
+            return;
+        } else {
+          sendf("ds18b20_result oid=%c next_clock=%u value=%i fault=%u"
+                  , oid, next_begin_time, d->temperature, d->error_count);
+          d->status = W1_READ_REQUESTED;
+        }
     } else if (d->status == W1_IDLE) {
         // This happens the first time requesting a temperature.
         // Nothing to report yet.
@@ -229,8 +238,8 @@ ds18_send_and_request(struct ds18_s *d, uint32_t next_begin_time, uint8_t oid)
         d->status = W1_READ_REQUESTED;
     } else if (d->status == W1_READY) {
         // Report the previous temperature and request a new one.
-        sendf("ds18b20_result oid=%c next_clock=%u value=%i"
-              , oid, next_begin_time, d->temperature);
+        sendf("ds18b20_result oid=%c next_clock=%u value=%i fault=%u"
+              , oid, next_begin_time, d->temperature, 0);
         if (d->temperature < d->min_value || d->temperature > d->max_value) {
             pthread_mutex_unlock(&d->lock);
             try_shutdown("DS18B20 out of range");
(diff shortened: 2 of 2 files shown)


functionpointer:
How would I do this in a macro?
The Linux kernel does not return a sensor reading and klipper is hardcoded to stop the printer

ld3300:
Does it still error out if it is set up as a generic sensor? (I'm guessing it probably does, but I can't check it) 
This is just my opinion, but allowing a firmware bypass for a safety feature because of hardware issues is not be an ideal commit to the general repo.  If the process is valuable for your personal use case it may be best to keep it for your running machine even if it doesn't get merged here.

charredchar:
> This is just my opinion, but allowing a firmware bypass for a safety feature because of hardware issues is not be an ideal

While I agree it would not be great to bypass safety if the sensor was used for something like a heater, it might be worth it to allow it on something like a generic temp sensor that'll be used just to keep an eye on things. Currently, even as a generic sensor it'll error and shutdown. This basically makes the DS18B20 useless as it can miss reports randomly.

ld3300:
How large of a user base do you suspect that this effects?  I'm not sure what the maintainers' views would be, but if it's usage is rare it might be a hard sell.

Make sure you add the "[Signed-off-by:](https://www.klipper3d.org/CONTRIBUTING.html?h=signed#format-of-commit-messages)" line to your original commit message.  It is required to have a commit accepted.

theopensourcerer:
@charredchar I was one of the peeps who put together the code to support these sensors in the first place (I did some of the python, not the linux/threaded side). I've been using one on my HEVO for ages and never had an issue with it at all. There is another ticket open, #5189, that might indicate the problem could be to do with it being used as a heater sensor, rather than a fan sensor?

It might also be worth seeing if the 1-wire kernel module has changed recently, as the DS18B20 is dependent on that?

functionpointer:
> Does it still error out if it is set up as a generic sensor? (I'm guessing it probably does, but I can't check it) This is just my opinion, but allowing a firmware bypass for a safety feature because of hardware issues is not be an ideal commit to the general repo. If the process is valuable for your personal use case it may be best to keep it for your running machine even if it doesn't get merged here.

Yes it does, i have it configured as follows (including my proposed feature):
```
[temperature_sensor enclosure]
sensor_type: DS18B20
serial_no: 28-3c01d6070703
sensor_mcu: host_mcu
ds18_report_time: 3
allow_fail: true
```

functionpointer:
Maybe another more safe suggestion (including use for heaters, etc) would be a retry counter.
For example we could allow up to three consequtive failed reads. If a fourth one occurs we stop the printer.

The number of retries should of course be made configurable.
This solution would shut down the printer in case something is actually wrong (broken wire or something) while tolerating random read errors.

What do you think? Would this be acceptable for the main repo?

functionpointer:
> Maybe another more safe suggestion (including use for heaters, etc) would be a retry counter. For example we could allow up to three consequtive failed reads. If a fourth one occurs we stop the printer.
> 
> The number of retries should of course be made configurable. This solution would shut down the printer in case something is actually wrong (broken wire or something) while tolerating random read errors.
> 
> What do you think? Would this be acceptable for the main repo?

Bump. Maybe @ld3300 has an opinion about this? I would be happy to implement this but only if it has a chance of getting accepted.

ld3300:
I'm not one of the developers, so I don't really have a say on accepting it.  However, I do think your idea may be a good middle ground that still allows for safe operation, while accounting for occasional transmission errors.

KevinOConnor:
Thanks.  However, as indicated, I don't think it makes sense to add a config option to disable a safety feature.  That said, it's certainly possible to make the code more resilient to transient errors.  If looking for feedback, might be best to start a discussion on Klipper Discourse.

-Kevin

functionpointer:
I have now implemented my suggestion. With my changes, Klipper will ignore a limited number of errors before stopping the printer. The safety check is thus no longer disabled but only slowed down a little.

I have tested the code on my machine. I can literally unplug the sensor and re-plug it. If i do it quickly enough the printer continues, otherwise it stops.

The current behaviour still sends 0°C for the tolerated errors. This shows the errors nicely in mainsail's temperature graph. Is this okay or would a different behaviour be better?

Sineos:
Would it make sense to copy over this approach to other "non critical" sensors?
Ref https://github.com/Klipper3d/klipper/issues/4874

functionpointer:
> Would it make sense to copy over this approach to other "non critical" sensors? Ref #4874

Hmm maybe, but maybe a more general solution would be better instead of copying the same logic several times.

Sineos:
If there is a solution implemented, IMO, it make sense to have one that works for other sensor devices as well that show the same behavior.
How generic this could possible be is over my pay grade

ReXT3D:
> Would it make sense to copy over this approach to other "non critical" sensors?
> Ref #4874

FWIW, I think that all remote "smart" sensors should have fault tolerance, even if they are "critical".  In principle, the amount of tolerance or acceptable duration of data loss or erroneous data should be driven by the applicable control loop time constants and the criticality or hazard associated with the interface.

I shared similar thoughts on [Pull 4164](https://github.com/Klipper3d/klipper/pull/4164) in my (perhaps somewhat academic) [comment](https://github.com/Klipper3d/klipper/pull/4164#issuecomment-862492779).  It would be really nice to have some form of standard fault tolerance mechanism built into Klipper, realizing that different sensors will require special treatment in the code.

Sineos:
> In principle, the amount of tolerance or acceptable duration of data loss or erroneous data should be driven by the applicable control loop time constants and the criticality or hazard associated with the interface.

Basically I would agree but I'm seeing too many "creative" configurations and I strongly think that the firmware should take some responsibility to protect the user from himself. If something is not working like intended, the first reflex is to tune down the safety / sanity checks and measures. 
So in reality I do not think the firmware should allow too much flexibility with respect to "critical" sensors.

functionpointer:
Maybe this big picture stuff can be reached incrementally. In the end, every sensor will need to implement the actual error handling for itself. Reporting those errors and reacting to them (shut down when enough have happened) could be generalized.

With this in mind, can we get this PR merged in its current state?

KevinOConnor:
I agree that the temperature sensor code should be more resilient to intermittent noise.  I don't think we should expose a new user-facing config option for this - a host default of 4 samples should be sufficient (for example, see `RANGE_CHECK_COUNT` in klippy/extras/adc_temperature.py).

As for this PR, I'd like to see feedback from @theopensourcerer and @joshhead .  Also, please follow the code style of the existing code (use a space between `if` and `(`), and do not add `output()` calls as they are only intended for debugging.

-Kevin

theopensourcerer:
Caveat - I've not even turned my printer on for nearly a year now so I'm probably not the best person to provide feedback.

The first comment I have, is that when we developed this module and when I was using it post-development for quite a while, I never had a print failure due to the DS18B20 not responding. So my first thought would be to try and understand why these errors are happening and try to prevent them in the first place, rather than working around something which, perhaps, shouldn't be worked around. Having said that, it seems there is more than one user who have had issues so maybe I was just lucky ;-)

The python code changes look fine to me, but the C stuff was done by @joshhead and not something I really understand that well.

src/linux/sensor_ds18b20.c
```diff
@@ -51,17 +50,23 @@ struct ds18_s {
     int temperature;
     struct timespec request_time;
     uint8_t status;
-    const char* error;
+    uint8_t error_count;
+    uint8_t max_error_count;
```
joshhead, line 54: Suggest moving to a constant along with W1_READ_TIMEOUT_SEC above.

maybe `#define W1_MAX_CONSECUTIVE_ERRORS 4`

src/linux/sensor_ds18b20.c
```diff
@@ -217,10 +224,15 @@ ds18_send_and_request(struct ds18_s *d, uint32_t next_begin_time, uint8_t oid)
     if (d->status == W1_ERROR) {
         // try_shutdown expects a static string. Output the specific error,
         // then shut down with a generic error.
-        output("Error: %s", d->error);
         pthread_mutex_unlock(&d->lock);
```
joshhead, line 227: The mutex_unlock needs to go inside the conditional. It should stay locked until after the pthread_cond_signal call below, if not shutting down here.

src/linux/sensor_ds18b20.c
```diff
@@ -217,10 +224,15 @@ ds18_send_and_request(struct ds18_s *d, uint32_t next_begin_time, uint8_t oid)
...
-        try_shutdown("Error reading DS18B20 sensor");
-        return;
+        if (d->error_count > d->max_error_count) {
+            try_shutdown("Error reading DS18B20 sensor");
+            return;
+        } else {
+          sendf("ds18b20_result oid=%c next_clock=%u value=%i"
+                  , oid, next_begin_time, 0);
```
joshhead, line 233: What if 0 is configured to be out of range? We could check the range here, or base the value on the configured temperature range, but I think it would be simpler to wait and not send anything until after a successful read.

src/linux/sensor_ds18b20.c
```diff
@@ -217,10 +224,15 @@ ds18_send_and_request(struct ds18_s *d, uint32_t next_begin_time, uint8_t oid)
...
+          d->status = W1_IDLE;
```
joshhead, line 234: IMO it would make sense to go directly back to W1_READ_REQUESTED.

```suggestion
        d->request_time = request_time;
        d->status = W1_READ_REQUESTED;
```

src/linux/sensor_ds18b20.c
```diff
@@ -247,8 +260,15 @@ ds18_send_and_request(struct ds18_s *d, uint32_t next_begin_time, uint8_t oid)
         if (request_time.tv_sec - d->request_time.tv_sec > W1_READ_TIMEOUT_SEC)
         {
             pthread_mutex_unlock(&d->lock);
-            try_shutdown("DS18B20 sensor didn't respond in time");
-            return;
+            //no response yet
+            d->error_count++;
```
joshhead, line 264: It might make sense to leave this part out and always fail on read timeouts. Errors in the read thread set a state of W1_ERROR which in turn sets the state back to W1_IDLE then W1_READ_REQUESTED (or as I suggested above skip W1_IDLE), and this also resets the read timeout time. So to get to this point we would have been waiting for a response for 5 seconds without encountering an error in the read thread. It seems likely to be a different kind of error than the transient ones that the error counter is addressing.

If keeping this behavior, the pthread_mutex_unlock shouldn't come before the read and writes to d->error_count.

joshhead (commented):
This seems like a good idea. Transient errors largely have not been an issue in my setup but I think I did encounter one once.

klippy/extras/ds18b20.py
```diff
@@ -10,6 +10,8 @@
 # Temperature can be sampled at any time but conversion time is ~750ms, so
 # setting the time too low will not make the reports come faster.
 DS18_MIN_REPORT_TIME = 1.0
+DS18_MAX_CONSECUTIVE_ERRORS = 4
```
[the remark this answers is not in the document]
KevinOConnor: Thanks for reviewing.

> I'm not that familiar with Python coding style in this project or generally so I can't say if they're worth keeping.

It can be convenient to put code constants in the Python code, as it's faster to test and debug changes that way.  (One only need change the python file and restart, instead of flashing a micro-controller.)  I don't have a particular preference for this code constant.

> I think the rest of the changes to this file are all formatting changes.

I agree that functional code changes and formatting/whitespace code changes should not be mixed in the same commit.  There's a note on this at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review (point 6).

-Kevin

src/linux/sensor_ds18b20.c
functionpointer: Not sending a report is not ideal IMO because it completely hides the error. The current 0C solution at least shows it in mainsail. Of course, out of range is a problem.

What if we moved the error handling into python? The firmware could simply notify the host about the error without actually handling it. The notification would be easiest with a special temperature (like `0xFFFF`) or with a message type (`ds18b20_error`?).

The python code could then handle the error including out of range. This could be repeated for other sensors and the error handling could be more centralized in the future. Aggregate reports or nice visualizations would be possible.

src/linux/sensor_ds18b20.c
joshhead: @functionpointer I notice you changed the line from W1_IDLE to W1_READ_REQUESTED, but didn't update the request time. I'm not sure if that was intentional or if you missed that line in my suggested change.

I was assuming that we would want to reset the 5 second timeout between error retries, but it could also make sense to time out after 5 seconds cumulatively. In fact if we used a cumulative timeout we could leave out the error count entirely.

src/linux/sensor_ds18b20.c
joshhead: It does seem like reporting the error somehow would be useful even if the sensor recovers. Sending 0C so that the anomaly shows up in the temperature graph seems like an easy solution but maybe not the best case in the long term.

> This could be repeated for other sensors and the error handling could be more centralized in the future. Aggregate reports or nice visualizations would be possible.

I like that idea, but it sounds more ambitious than this fix. Are there other modules that report non-fatal errors now (and what do they do)?

functionpointer: No other sensors yet, but plenty of discussion above. It seems that i2c sensors are similarly not 100% reliable on every read.

I agree 0°C is just a stopgap.

Sineos: > I like that idea, but it sounds more ambitious than this fix. Are there other modules that report non-fatal errors now (and what do they do)?

@joshhead See https://github.com/Klipper3d/klipper/issues/4874. A common fix would really make sense IMO

KevinOConnor: FWIW, might be better to add a new field to the ds18b20_result message - for example, "ds18b20_result oid=%c next_clock=%u value=%i fault=%u".  See `thermocouple_response()` in src/thermocouple.c as an example.

-Kevin

samwiseg0:
> > Would it make sense to copy over this approach to other "non critical" sensors? Ref #4874
> 
> Hmm maybe, but maybe a more general solution would be better instead of copying the same logic several times.

I agree with this. As of now the i2c MCU code is hard coded to cause an MACU shutdown on any i2c errors. I would like to see this solution extended to those sensors as well.

pmaersk:
I apologize if this comment is inappropriate here, but is it possible to ask when this is merged back to main branch?
And is it correct understood that with this change, an error is returned to the python script upon read error and Klipper will no longer terminate print? I have a ds18B20 sensor on each stepper and as you can imagine, with old code, I never get to finish a print.

functionpointer:
> I apologize if this comment is inappropriate here, but is it possible to ask when this is merged back to main branch? And is it correct understood that with this change, an error is returned to the python script upon read error and Klipper will no longer terminate print? I have a ds18B20 sensor on each stepper and as you can imagine, with old code, I never get to finish a print.

That is correct. If you don't want to wait you can checkout this code already and try it on your printer.

I would also like to know when this code will be reviewed. I have just rebased it and hopefully it now complies to the commit format.

pmaersk:
> That is correct. If you don't want to wait you can checkout this code already and try it on your printer.

True, however I hope we are close to merger and I automatically download new main branch. I would need to disable that. Possible, but undesirable, if we are close to merge your contrib ution. BTW thanks for the contribution :-)
> 
> I would also like to know when this code will be reviewed. I have just rebased it and hopefully it now complies to the commit format.




I tried to search around on the website, but it isn't made clear, how long time it will take to be merged after sort of consensus has been reached.

Any way, you write it will fail after 5 sensor read errors in a row. Is that 5 read errors per DS18B20? Imagine I have 10 sensors, it will only fail if at least one of the sensors fails 5 time in a row? It is not 5 fails for the DS18B20 class of sensors?

In my case I will have 16-20 sensors on motors and motor controllers etc somewhat increasing chances for reaching 5 errors in a row, especially if they are all daisy chained (increases chances of transmission interference). And the larger number of sensors does increase the time it takes to get to read them all, perhaps between a second or two.
The printer will stop when the senso

functionpointer:
Errors are tracked for each sensor individually

Fjbarchesky:
It is unclear to me how to setup my config for a ds18b20 after pulling [this ](https://github.com/Klipper3d/klipper/pull/5233/commits/5907c2dfbf794c48c6d6c3656aa286fb6a8e0726)update into my rpi/avr setup. Currently receive: 

```
MCU Protocol error

This is frequently caused by running an older version of the
firmware on the MCU(s). Fix by recompiling and flashing the
firmware.

Your Klipper version is: v0.10.0-419-g4490a584-dirty
MCU(s) which should be updated:
mcu: Current version v0.10.0-419-g4490a584-dirty-20220522_155441-mainsailos
rpi: Current version v0.10.0-419-g4490a584-dirty-20220522_155341-mainsailos
Up-to-date MCU(s):
<none>

mcu 'rpi': Unable to encode: config_ds18b20
```
Please advise if there are additional fields to define in a ds18b20 config within the printer.cfg file. I currently have: 

```
[temperature_sensor ds18b20-chamber]
sensor_type: DS18B20
serial_no:28-3c66f648a996
sensor_mcu: rpi
ds18_report_time: 1
```

Above includes an `allow_fail: true` however when attempting to restart klipper it fails with: 
`Option 'allow_fail' is not valid in section 'temperature_sensor ds18b20-chamber'
`

Not sure what is going on here but this is my log file: 
[klippy.log](https://github.com/Klipper3d/klipper/files/8750239/klippy.log)

functionpointer:
`allow_fail` has been removed after feedback from the community. Your config looks good.

Have you "reflashed" the mcu firmware for your RPI?

Remember that you have the RPI set-up as a second mcu so it is also running a firmware as if it was another avr board. You have programs running in parallel: the main klippy program, the firmware on the avr and a second firmware on the RPI.

Fjbarchesky:
I reflashed my avr, now I receive: 
```
MCU 'rpi' shutdown: Missed scheduling of next digital out event
This is generally indicative of an intermittent
communication failure between micro-controller and host.
```
The above error occurs when I attempt to use a ds18b20 in a generic heater (within printer.cfg): 

```
[heater_generic chamber_heater]
heater_pin: rpi:gpio27
sensor_type: DS18B20
serial_no:28-3c66f648a996
sensor_mcu: rpi
control: watermark
pwm_cycle_time: 0.8
max_delta: 5
min_temp: 0
max_temp: 50
max_power: 1.0
gcode_id: C

[verify_heater chamber_heater]
max_error: 120
check_gain_time: 60
hysteresis: 5
heating_gain: 2

[heater_fan chamber_fan]
pin: rpi:gpio22
heater: chamber_heater
heater_temp: 10.0
fan_speed: 1.0
```

Any advice?

functionpointer:
Does it work when configured as a generic temperature sensor?

Fjbarchesky:
Yes! the sensor talks:

image

functionpointer:
Yay!

Not sure why the heater configuration doesn't work. I will take a look tomorrow.

Until then you can try unplugging and replugging the sensor to verify that the errors are indeed being tolerated for a few seconds.

functionpointer:
Your config works fine on my machine (with some tweaks like DS18B20 id and mcu name).

Is your cpu usage ok? What RPI are you running?

Fjbarchesky:
Thank you for the assistance here. I believe my RPI is 3B+. I think my loads are fine: 
system_loads

Fjbarchesky:
I was able to check that the sensor can be disconnected, the temperature reading goes to 0C , and the system does not reboot.

Fjbarchesky:
I believe I have discovered the issue: the generic heater should not utilize rpi gpio pins, I found a comment on the discord #configuration server with respect to generic heater function and usage with a raspberry pi. The message states "ok, so a heater pin is not a gpio pin, it expects a mosfet pwm pin, which pi's do not have". I appreciate your assistance @functionpointer! thanks again. Now I need to figure out how to write a macro that utilizes the output_pin / temperature reading.

functionpointer:
> I believe I have discovered the issue: the generic heater should not utilize rpi gpio pins, I found a comment on the discord #configuration server with respect to generic heater function and usage with a raspberry pi. The message states "ok, so a heater pin is not a gpio pin, it expects a mosfet pwm pin, which pi's do not have". I appreciate your assistance @functionpointer! thanks again. Now I need to figure out how to write a macro that utilizes the output_pin / temperature reading.

I have a RPI 4 and I used gpio19 for the heater and gpio13 for the fan. I didn't actually have anything connected so I don't know if the pins actually did anything, but Klipper was running just fine even after I turned the heater on.

Fjbarchesky:
Hmm very odd. Not sure what to think now.

KevinOConnor:
What's the status of this PR?  I'm fine with committing a change to the ds18b20 code to make it more resilient to sporadic errors.  Looking at this PR though, it seems it has a bunch of unrelated whitespace changes which complicates review and leads to problems should we need to revert a change (this is mentioned briefly as point 6 at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review ).

Thanks.
-Kevin

src/linux/sensor_ds18b20.c
functionpointer: I have added the `fault` parameter.

klippy/extras/ds18b20.py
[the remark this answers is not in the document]
functionpointer: I have force-pushed new commits which separate the formatting changes.

KevinOConnor:
@joshhead - did you have any final comments on the latest iteration of this PR?

-Kevin

joshhead:
The latest changes look good to me!

dmaxwellNEU:
@Fjbarchesky did you ever figure out how to make this work with a generic heater and DS18B20 sensor? You said you found the clue about pwm required pin on discord?

Fjbarchesky:
I haven’t used my enclosure heater in awhile, however I believe I wrote a few macros and used ‘output_pin’ to toggle a discrete based on the temperature readings. 

The reason why I think I could not get the generic heater to work was because: I only have spare general discrete(s) to toggle a separate relay connected to my pi or main board. A generic heater requires a heater circuit not a discrete.

Fjbarchesky:
> @Fjbarchesky did you ever figure out how to make this work with a generic heater and DS18B20 sensor? You said you found the clue about pwm required pin on discord?

I haven’t used my enclosure heater in awhile, however I believe I wrote a few macros and used ‘output_pin’ to toggle a discrete based on the temperature readings.

The reason why I think I could not get the generic heater to work was because: I only have spare general discrete(s) to toggle a separate relay connected to my pi or main board. A generic heater requires a heater circuit not a discrete.

dmaxwellNEU:
> > @Fjbarchesky did you ever figure out how to make this work with a generic heater and DS18B20 sensor? You said you found the clue about pwm required pin on discord?
> 
> I haven’t used my enclosure heater in awhile, however I believe I wrote a few macros and used ‘output_pin’ to toggle a discrete based on the temperature readings.
> 
> The reason why I think I could not get the generic heater to work was because: I only have spare general discrete(s) to toggle a separate relay connected to my pi or main board. A generic heater requires a heater circuit not a discrete.

Ok that makes sense. It's a shame that the code for klipper couldn't have easily been edited to include regular gpio pins as an option. On/off state is fine for my heater no need for pwm control. If that really is the issue, then the only way is like you said to move to discrete gpio pin and just use it as an output_pin with temperature reading based control. 

Admins aren't willing to entertain such a change?

dmaxwellNEU:
> > I believe I have discovered the issue: the generic heater should not utilize rpi gpio pins, I found a comment on the discord #configuration server with respect to generic heater function and usage with a raspberry pi. The message states "ok, so a heater pin is not a gpio pin, it expects a mosfet pwm pin, which pi's do not have". I appreciate your assistance @functionpointer! thanks again. Now I need to figure out how to write a macro that utilizes the output_pin / temperature reading.
> 
> I have a RPI 4 and I used gpio19 for the heater and gpio13 for the fan. I didn't actually have anything connected so I don't know if the pins actually did anything, but Klipper was running just fine even after I turned the heater on.

@functionpointer do you think you could please elaborate? If the pin assignment in a generic_heater is really not required to be pwm, then why am I also having trouble? It would be nice to be able to use the generic_heater with the DS18B20 I already have working for my cooling fan. I've looked everywhere online, and this thread is the closest I've seen to finding an answer. Being able to control the chamber heater, fan, temp monitoring all seem like features that should be easier and clearer to setup. Comparing to Duet reprap and octoprint enclosure plugin, this seems like its 97% the way there ... Is there anyway this could be readdressed? Thanks!

PR #5235, opened 2022-02-05
Add hardware emergency stop

chibisuke:
While trying to implement my emergency stop button I noticed that gcode_button doesn't respond instantly even to printer.action_emergency_stop() calls inside the gcode, so I decided to implement a dedicated emergency stop button config option that bypasses the gcode processor und performs the emergency stop nearly instantenious.


Gilabite:
thats weird because I have had a hardware e stop button on all my printers for several years and it always stops instantly this is how mine are

[gcode_button kill_switch]
pin = ^!PE4
press_gcode =
 { action_emergency_stop() }

chibisuke:
Thats what I tried, but that doesn't work during homing or quad gantry levelng (and propably other long running stuft)

Gilabite:
works fine for me on everything

chibisuke:
I already discussed this in discord before creating the patch, because I thought I might be missing something, but I got told there that due to the gcode-lock its impossible for a gcode button to respond immediately.

Gilabite:
Yeah Im still confused as to why mine works fine then. I have never had an issue with it not stopping instantly during any kind of use and have used it during every process a printer can have over the time Ive had them on my printers. Ive even used it to cancel a print when I didnt want to walk across the room to my PC to stop it in mainsail. I would have done something about it a long time ago if it didnt work instantly like it should. Did you use a "^" on the pin in your config? What kind of button are you using? Are you connecting it to the MCU or Pi? Did you try running other gcode before or after the action_emergency_stop() command? Im just trying to figure out why mine works and yours doesnt not trying to bash your work or anything.

chibisuke:
Nope, I'm not using any ^ since the port on my btt octopus already has a pull-up resistor onboard, so enableing it in the MCU would possibly cause issues. The button is a normal NC pushbutton and its connected to the MCU. I'm not running any other gcode on that button.

dalegaard:
Hmpf, after looking over the code more carefully, the gcode expansion should be happening before the lock is taken, and thus it should work. Maybe something else is going on in your case @chibisuke :thinking: 

Best regards,
Lasse

chibisuke:
yea I've seen that aswell in the code, and I don't get why this is not working as its surposed to be, but to be honest I just needed a solution that works and didn't feel like wasting hours and hours of try and error. 

if you have any ideas what to try, I'm glad to test it but appeart from that I'm not going to waste any more time on this topic.

bhelm:
I have tested it and i can confirm that action_emergency_stop() works instantly on homing.

However, at first i used M112 instead which indeed executes after homing is finished. I was looking at the gcode docs, saw M112 emergency shutdown and thought i was good to go. I feel that an proper kill switch example should be added to the docs, as it is a common feature and there is room to do it wrong.

PR #5237, opened 2022-02-06
fix integer handling for python3 in bme280.py

cbc02009:
Incorrect math in bme280.py was causing issues when using Python 3. This fix should work with both python 3 and python 2.

Singed-off-by: Christopher Conroy <cbc02009@gmail.com>


KevinOConnor:
Thanks.

-Kevin

PR #5239, opened 2022-02-06
BUG: Fix STM32H7 ADC reset

### LINKED ISSUE #5236: BiquBX was addded to Klipper but temp not working
Hey guys, I was happy to see that the BiquBX's mainboard was added to Klipper.
We started today on configuring everything with the example cfg provided by https://github.com/Klipper3d/klipper/commit/c15d38b973a56978535f36c0fe61f5f8de468248

Unfortunately the bed temperature isn't working at all, so maybe there seems to be still an issue with the board? It was working with marlin before.

image

adelyser:
module: STM32 ADC
Don't reset the ADC peripheral if the clock is already enabled.

Fixes #5236

### DIFF
--- a/src/stm32/stm32h7_adc.c
+++ b/src/stm32/stm32h7_adc.c
@@ -114,7 +114,9 @@ gpio_adc_setup(uint32_t pin)
     ADC_TypeDef *adc;
     if (chan >= 40){
         adc = ADC3;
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #5243, opened 2022-02-08
Config for Monoprice Select Mini v1

j4c3:
module: Config, MPSM v1 file.

There are a few differences between v1 and v2.  I have verified that these options when flashed will function and connect via usb for both the original and revised versions of the v1 board.  A few other printers have version specific config files, but if it makes sense to combine the v1 and v2 here, I can write that up instead.  (I believe there is a v3 of this printer, but I don't know if the v2 config works with that printer or not.)

Signed-off-by: Jace Wiseman <jwiseman@gmail.com>


ld3300:
You need to include the Signed-off-by: as described in the [Contribution](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages) docs.

I recommend also replacing all the instances of "you" and "your" with more generic references like: "the printer" instead of "your printer".

KevinOConnor:
Thanks.  In general it looks fine to me.  Is the filename intended to mean "pre-2018", or is "pre" part of the version information?  If the former, just use a year that is roughly accurate (eg, 2017) as I think the term "pre" could be confusing.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #5245, opened 2022-02-08
displaystatus: Allow M73 to unset build percentage

JeremyRuhland:
Running M73 without additional parameters will now null the progress variable, mirroring the behavior of M117. This permits control of the progress variable to be returned to virtual_sdcard after it has been modified by an `M73 P<percent>` command.

Previously there was no way to release control of the progress percentage variable back to virtual_sdcard once an `M73 P<percent>` command was issued.

An example use case: A print_start macro may update the progress percentage using `M73 P<percent>` to indicate how close the homing, heating and priming process is to completion. At the end of the macro an `M73` command can be issued allowing virtual_sdcard to automatically update the progress bar during the print to reflect how close it is to completion.

Code has been tested on my voron 2.4.

Signed-off-by: Jeremy Ruhland <jeremy-klipper@goopypanther.org>

### DIFF
--- a/klippy/extras/display_status.py
+++ b/klippy/extras/display_status.py
@@ -1,5 +1,6 @@
 # Module to handle M73 and M117 display status commands
 #
+# Copyright (C) 2022  Jeremy Ruhland <jeremy-klipper@goopypanther.org>
 # Copyright (C) 2018-2020  Kevin O'Connor <kevin@koconnor.net>
 # Copyright (C) 2018  Eric Callahan <arksine.code@gmail.com>
 #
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  As high-level feedback though, we don't add "new features to old g-code".  You might want to look at overriding the display settings or overriding the M73 command with a gcode_macro if you need to customize the behaviour.

-Kevin

JeremyRuhland:
The alternate usage in the reprap firmware is a good point, this probably isn't the best way to do this.
I don't think I can fix it by overriding display settings because the behavior originates here:

https://github.com/Klipper3d/klipper/blob/38b992532abb45af0a11522b252f5207fdb931f5/klippy/extras/display_status.py#L26-L31

Even if I switched to a different display configuration that pulled the progress directly from virtual_sdcard it wouldn't be able to affect the web GUI.

Maybe the best solution would be to register a new command, something like `RESET_DISPLAY_PROGRESS` that would perform this action while leading the `M73` behavior alone.

pedrolamas:
> Maybe the best solution would be to register a new command, something like `RESET_DISPLAY_PROGRESS` that would perform this action while leaving the `M73` behavior alone.

I believe that should be possible by adding a new custom module (like [this one](https://github.com/th33xitus/kiauh/blob/master/resources/gcode_shell_command.py))

Arksine:
I think that part of the question here is whether or not its appropriate to use M73 for anything but build progress.  G-Code macros and variables could be used to track the progress of procedures outside of the build without changing the behavior of `display_status`.

pedrolamas:
@JeremyRuhland In terms of using the M73, I think we all agree that it is not the best way to improve the situation you described, so I don't think this PR (as it currently stands right now) should be merged.

PR #5249, opened 2022-02-08
Update printer-tronxy-x5sa-v6-2019.cfg

Polar-Ted:
Lines 94 and 98..  
[heater_fan] and [fan] max_power was set to 0.5
This should be 1 for both.

Signed-off-by: Scott Schering sschering@gmail.com

### DIFF
--- a/config/printer-tronxy-x5sa-v6-2019.cfg
+++ b/config/printer-tronxy-x5sa-v6-2019.cfg
@@ -91,11 +91,11 @@ pid_Kd: 898.279
 
 [heater_fan hotend_fan]
 pin: PG14
-fan_speed: 0.5
+fan_speed: 1
 
 [fan]
 pin: PG13
-max_power: 0.5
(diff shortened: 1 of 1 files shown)


ld3300:
Please add Signed-off-by: as required in the [Format of a commit message](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)

Polar-Ted:
> Please add Signed-off-by: as required in the [Format of a commit message](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)

done, Thank you

ld3300:
I think your change is a good idea.  I can't approve the commit, but since a value of 1.0 is the default for both of those parameters it would be best to remove the fan_speed and max_power lines from the config file instead of updating them.  That would make your example config follow the example configuration guide note to "Avoid defining field values that are set to their default value."

Issue #5198, opened 2022-01-26
MAX31865 and TMC5160 on shared SPI bus doesn't work (stm32f429)


Serbitor:
I have an stm32f429 board (octopus pro).
I have two TMC5160 stepper modules installed and an on-board MAX31865, these are all wired (by the board) to SPI1.

When both TMC5160 and MAX31865 are configured using hardware SPI no temperature readings are possible and the printer goes immediately into shutdown.

After removing the configuration for TMC5160 drivers entirely and restarting firmware, the MAX31865 readings start working and provide the correct values.

If I convert all three devices (TMC5160 *2 and MAX31865) to use software SPI all three devices work correctly. 

The relevent pins are these:
spi1_sclk: PA5
spi1_miso: PA6
spi1_mosi: PA7
TMC5160 stepper_x cs:  PD11
TMC5160 stepper_x cs:  PC4
MAX31865 cs: PF8

Broken config sections with hardware SPI:

```
[tmc5160 stepper_x]
cs_pin: PD11
spi_bus: spi1
run_current:  2.500

[tmc5160 stepper_y]
cs_pin: PC4
run_current:  2.500
hold_current: 2.000
spi_bus: spi1

[extruder]
step_pin: PE6
dir_pin: PA14
enable_pin: !PE0
microsteps: 16
rotation_distance: 4.54
nozzle_diameter: 0.330
filament_diameter: 1.750
heater_pin: PA2
control: pid
[klippy.log](https://github.com/Klipper3d/klipper/files/7945183/klippy.log)

min_temp: 0
max_temp: 300
pid_Kp: 15.247
pid_Ki: 0.664
pid_Kd: 87.479
pressure_advance: 0.06
sensor_type: MAX31865
sensor_pin: PF8
spi_bus: spi1
rtd_nominal_r: 1000
rtd_reference_r: 4300
```

Working config sections with software SPI:

```
[tmc5160 stepper_x]
cs_pin: PD11
spi_software_sclk_pin: PA5
spi_software_miso_pin: PA6
spi_software_mosi_pin: PA7
run_current:  2.500
hold_current: 2.000

[tmc5160 stepper_y]
cs_pin: PC4
run_current:  2.500
hold_current: 2.000
spi_software_sclk_pin: PA5
spi_software_miso_pin: PA6
spi_software_mosi_pin: PA7

[extruder]
step_pin: PE6
dir_pin: PA14
enable_pin: !PE0
microsteps: 16
[...]

PR #5250, opened 2022-02-09
BUG: Fix STM32 SPI CR1

bigtreetech:
STM32 SPI_CR1 should not be changed when communication is ongoing.

fix https://github.com/Klipper3d/klipper/issues/5198

### DIFF
--- a/src/stm32/spi.c
+++ b/src/stm32/spi.c
@@ -100,6 +100,7 @@ void
 spi_prepare(struct spi_config config)
 {
     SPI_TypeDef *spi = config.spi;
+    spi->CR1 &= ~((uint16_t)SPI_CR1_SPE);
     spi->CR1 = config.spi_cr1;
 }
 


cbc02009:
This does not work for me. With the patch and newly flashed firmware I still get the same `ADC out of range` error I was getting last time I tried using hardware SPI.

cbc02009:
what's interesting is using `spi->CR1 - 0x00;` works for me, so there must be some other bits in the register that aren't being handled properly.

KevinOConnor:
Thanks.  I think there are two separate issues occurring.
1. The `spi_transfer()` function should not be exiting until after all SCLK pin updates are complete.  Otherwise, the CS pin may rise before the final SCLK pin updates, and subsequent writes to CR1 might occur while a transmission is still in progress (in spi mode 0 and 1).
2. It looks like the specs do state that the SPE bit needs to be cleared before changing CPOL or CPHA bits.

Based on this, something like the following should fix it:

```diff
--- a/src/stm32/spi.c
+++ b/src/stm32/spi.c
@@ -100,6 +100,11 @@ void
 spi_prepare(struct spi_config config)
 {
     SPI_TypeDef *spi = config.spi;
+    uint32_t cr1 = spi->CR1;
+    if (cr1 == config.spi_cr1)
+        return;
+    // The SPE bit must be disabled before changing CPOL/CPHA bits
+    spi->CR1 = cr1 & ~SPI_CR1_SPE;
     spi->CR1 = config.spi_cr1;
 }
 
@@ -117,4 +122,7 @@ spi_transfer(struct spi_config config, uint8_t receive_data,
             *data = rdata;
         data++;
     }
+    // Wait for any remaining SCLK updates before returning
+    while ((spi->SR & (SPI_SR_TXE|SPI_SR_BSY)) != SPI_SR_TXE)
+        ;
 }
```

-Kevin

cbc02009:
```
--- a/src/stm32/spi.c
+++ b/src/stm32/spi.c
@@ -100,6 +100,12 @@ void
 spi_prepare(struct spi_config config)
 {
     SPI_TypeDef *spi = config.spi;
+    uint32_t cr1 = spi->CR1;
+    if (cr1 == config.spi_cr1)
+        return;
+    // The SPE bit must be disabled before changing CPOL/CPHA bits
+    spi->CR1 = cr1 & ~SPI_CR1_SPE;
+    spi->CR1; // Force flush of previous write
     spi->CR1 = config.spi_cr1;
 }
 
@@ -117,4 +123,7 @@ spi_transfer(struct spi_config config, uint8_t receive_data,
             *data = rdata;
         data++;
     }
+    // Wait for any remaining SCLK updates before returning
+    while ((spi->SR & (SPI_SR_TXE|SPI_SR_BSY)) != SPI_SR_TXE)
+        ;
 }

``` appears to work for me

KevinOConnor:
Okay, thanks.  I went ahead and made these changes (commit 99d55185 and e3cbe7ea).

-Kevin

PR #5251, opened 2022-02-09
Update to Tronxy printer config files

64bittuning:
The Tronxy config files that are currently available, while good, are somewhat incomplete. Over the last few years, Tronxy has produced the X5SA and XY2 Pro with a few different mainboards. The MOST COMMON are the CXY-V6-191017 and CXY-V5-180409. The pin-mapping of these two boards is almost identical, but the changes in direction of stepper motors, and various extruder options have caused the waters to be muddied when it comes to getting the right file.
There has been a large influx of interest in Klipper amongst the Tronxy community and these files should help solve many community issues.

The two new files should cover both mainboards across both printer models, and the older three config files should be considered obsolete. Most of their content has been repackaged here for ease, along with some additional notations to help get the machines operational.

All original authors have been tagged for their work.

Co-Authored-By: Paul McGowan <17770733+mental405@users.noreply.github.com>
Co-Authored-By: TheBugLebowsky <morgan@morganino.it>
Co-Authored-By: Vladimir Serov <me@cab404.ru>
Co-Authored-By: Raabi91 <38133278+Raabi91@users.noreply.github.com>
Co-Authored-By: gcastle2007 <77104012+gcastle2007@users.noreply.github.com>


64bittuning:
Apparently the built in tests here, do not like the fact that the old configs were removed. Unfortunately, they should be. Please consider replacing 
/printer-tronxy-x5sa-pro-2020.cfg
/printer-tronxy-x5sa-v6-2019.cfg
and
/printer-tronxy-xy-2-Pro-2020.cfg

with
/printer-tronxy-X5SA-and-X5SAPro.cfg
and
/printer-tronxy-XY2Pro.cfg

config/printer-tronxy-X5SA-and-X5SAPro.cfg
```diff
@@ -0,0 +1,228 @@
...
+pin: PD6
+
+[filament_switch_sensor sentinel]
+pause_on_runout: True
+runout_gcode:
+  M25
+switch_pin: !PA15
+
+[output_pin beeper]
+pin: PB0
```
bevanweiss, line 166: For my printer with the V6 Chitu board having this as a pure digital output pin doesn't work, as the board has a piezo buzzer which needs some PWM to get it to make an appropriate audible sound.
Unsure of how to do PWM but still have it as an On/Off type of signal (monotone).

I currently have the following in my config (setting the value to something like 1600 sounds 'good enough' for me, 0 to turn it off again).
[output_pin beeper]
pin: PB0
pwm: True ; Chitu V6 uses piezo buzzer which requires PWM
hardware_pwm: True
cycle_time: 0.001 ; PWM frequency : 0.001 = 1ms will give a base tone of 1kHz
scale: 4000 ; PWM parameter will be in the range of (0-4000 Hz).

64bittuning: I never really cared about having the beeper active, so I didnt even consider this. I'll test it on the 191017 board I have here.

KevinOConnor:
Thanks.  I'm not sure what the state of this PR is.  I don't have a preference on the tronxy example configs.  However, the new configs would need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html (at least points 4, 7, 7g).

If you're replacing existing config files then you should ping the submitters of those files to get their feedback (run `git log` on the existing configs to see who submitted them).

Also, commits need to have a signed-off-by line containing the full real name of contributors - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

PR #5252, opened 2022-02-09
Adding bed_screws to Ender 3

hatkyinc2:
Adding the bed_screws to Ender 3 by what I could measure mine.
This enables `BED_SCREWS_ADJUST`.

https://github.com/Klipper3d/klipper/blob/master/docs/Manual_Level.md#adjusting-bed-leveling-screws

Signed-off-by: Ofir Petrushka <ofir.petrushka@gmail.com>

### DIFF
--- a/config/printer-creality-ender3-2018.cfg
+++ b/config/printer-creality-ender3-2018.cfg
@@ -88,6 +88,12 @@ max_accel: 3000
 max_z_velocity: 5
 max_z_accel: 100
 
+[bed_screws]
+screw1: 30.5, 37
+screw2: 30.5, 207
+screw3: 204.5, 207
+screw4: 204.5, 37
+
 [display]
 lcd_type: st7920
 cs_pin: PA3


ld3300:
Please add the Signed-off-by: statement as require in the [Contributing Guidlines](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)

hatkyinc2:
@ld3300 Thanks for having a look, is this ok now?

ld3300:
Edit your original post comment and add the signed off by statement there.

hatkyinc2:
Added

qwewer0:
I'm not too sure about these values.
On my ender 3, the screws are 32.5mm inwards from the edge of the bed.

hatkyinc2:
I measured it multiple times originally, probably could have taken pics, I don't know if they would always be consistent or this changes at some point (date in time where something in manufacturing changed, "versions" of the bed...) Also note this is original Ender 3 and not derivatives like pro, v2...

Honestly I don't mind if it's changed a few mm to either directions, it's only used for manual leveling and close is better than not having anything set IMHO. This is a template and might need adjusting, I think just having it makes it easier for people starting.

rpodgorny:
is this in correct section? my ender3 config has the screw positions in [screws_tilt_adjust] :-O

hatkyinc2:
As you can read in the referenced doc BED_SCREWS_ADJUST is used with manual leveling

screws_tilt_adjust is for ABL probes

KevinOConnor:
Thanks.

-Kevin

PR #5256, opened 2022-02-11
Use probe_offset in Z_TILT_ADJUST and SCREWS_TILT_CALCULATE

ClockeNessMnstr:
Homing and BED_MESH both use the probe offset. 

Z_TILT_ADJUST and SCREWS_TILT_CALCULATE did not include the offset.

I can't quite tell if the adjustment calculation for Z_TILT_ADJUST already compensates for the offset and if the calculation would be the same or would be degraded / improved

### DIFF
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -166,7 +166,9 @@ class PrinterProbe:
         while retries < samples_retries and retries > last_retries:
             sample_num = 0
             while sample_num < sample_count:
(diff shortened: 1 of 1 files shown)


ld3300:
Hi, can you describe why you feel including the probe offset for those functions would be beneficial, since both of those functions only need to know the difference in where the probe is triggered at different places on the bed.  Is there a use case where the nozzle offset from the probe makes a difference to these measurements?

Please also include "Signed-off-by:" statement in your commit comment as stated in the [format section](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)

ClockeNessMnstr:
Wrong branch

PR #5257, opened 2022-02-12
Use probe_offset in Z_TILT_ADJUST and SCREWS_TILT_CALCULATE

ClockeNessMnstr:
Homing and BED_MESH both use the probe offset.

Z_TILT_ADJUST and SCREWS_TILT_CALCULATE did not include the offset.

~~I can't quite tell if~~ the adjustment calculation for Z_TILT_ADJUST already (appears to) compensate for the offset

### DIFF
--- a/klippy/extras/screws_tilt_adjust.py
+++ b/klippy/extras/screws_tilt_adjust.py
@@ -34,6 +34,7 @@ class ScrewsTiltAdjust:
         self.probe_helper = probe.ProbePointsHelper(self.config,
                                                     self.probe_finalize,
                                                     default_points=points)
+        self.probe_helper.use_xy_offsets(True)
         self.probe_helper.minimum_points(3)
         # Register command
         self.gcode = self.printer.lookup_object('gcode')
--- a/klippy/extras/z_tilt.py
+++ b/klippy/extras/z_tilt.py
@@ -131,6 +131,7 @@ class ZTilt:
                                            parser=float, count=2)
         self.retry_helper = RetryHelper(config)
         self.probe_helper = probe.ProbePointsHelper(config, self.probe_finalize)
+        self.probe_helper.use_xy_offsets(True)
         self.probe_helper.minimum_points(2)
         self.z_status = ZAdjustStatus(self.printer)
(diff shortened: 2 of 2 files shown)


ClockeNessMnstr:
@ld3300 

Benefit is setting the screw positions in the config would probe there. It was unexpected having the probe not go to the point in the config. 

I see the bed_tilt and bed_screws code now though. that would also need adjusting.

ld3300:
@ClockeNessMnstr,
My bad, for some reason I was only thinking of the z offset, but you are right that you would need the probe to hit the positions in the X and Y

kusold:
Does this account for setups that don't have enough space to get the probe over the screw? Will it instead just probe at the max position?

KevinOConnor:
Alas, there isn't consensus on this change.  Some people think in terms of nozzle coordinates and some people think in terms of bed coordinates - we can't seem to win.  I suspect moving this to Klipper Discourse may be better.

-Kevin

laurienzu:
I think too that there is an error in the screws_tilt_adjust function. The config reference says to specify the screws location but it doesn't take into account the probe offset. So, if I set the screws position correctly I will have my probe outside the bed resulting into a crash, if I consider the offset and add it manually to screws position then the calculation will be totally wrong....

In my opinion in this function is better to specify the probing positions for the screws and then the algorithm should calculate the correct adjusting amount for screws.

ClockeNessMnstr:
> I think too that there is an error in the screws_tilt_adjust function. The config reference says to specify the screws location but it doesn't take into account the probe offset. So, if I set the screws position correctly I will have my probe outside the bed resulting into a crash, if I consider the offset and add it manually to screws position then the calculation will be totally wrong....
> 
> In my opinion in this function is better to specify the probing positions for the screws and then the algorithm should calculate the correct adjusting amount for screws.

From memory there was two functions that looked like they were wrong but there's a display accessible function to adjust these manually I believe where it makes sense not to have the offset. 
I think all four functions could have been modified but I do not have a screen so I didn't look at the "fix"

It should be adjusting correctly I believe but differs in behavior from other probing procedure for sure

PR #5258, opened 2022-02-12
make PT100 available in configuration

mngr0:
poor PT100, defined but unusable

### DIFF
--- a/klippy/extras/adc_temperature.py
+++ b/klippy/extras/adc_temperature.py
@@ -277,7 +277,7 @@ DefaultVoltageSensors = [
 ]
 
 DefaultResistanceSensors = [
-    ("PT1000", calc_pt100(1000.),
+    ("PT1000", calc_pt100(1000.)),
     ("PT100", calc_pt100(100.))
 ]
 
--- a/klippy/extras/adc_temperature.py
+++ b/klippy/extras/adc_temperature.py
@@ -277,7 +277,8 @@ DefaultVoltageSensors = [
 ]
 
 DefaultResistanceSensors = [
-    ("PT1000", calc_pt100(1000.))
+    ("PT1000", calc_pt100(1000.),
+    ("PT100", calc_pt100(100.))
 ]
 
 def load_config(config):


ld3300:
I can't help with verifying the code, but make sure you add the Signed-off-by: statement to your commit as described [here](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages)
Best,
David

Sineos:
I would agree to connect a PT1000 without an amplifier board, but IMO doing this for a PT100 makes absolutely no sense

jakep82:
This is not a valid way to connect a PT100. They require an amplifier.

mngr0:
May I ask you why a PT100 requires an amplifier? Is it necessary in 2 wire connection too?

dewi-ny-je:
> May I ask you why a PT100 requires an amplifier? Is it necessary in 2 wire connection too?

A PT100 with a standard 4700 ohm pullup resistor has a resolution of about 3.5 °C/step of the ADC converter. It's way way too rough for any decent temperature regulation.

If you change the pullup to 1000 ohm you have about 1 °C/step of the ADC, which is maybe acceptable but far from decent.

mngr0:
> A PT100 with a standard 4700 ohm pullup resistor has a resolution of about 3.5 °C/step of the ADC converter. It's way way too rough for any decent temperature regulation.

If the device I am using has a 12 bit or 16 bit ADC resolution? would it still be a problem? (maybe I would need to tell klipper how to configure the adc peripheral in order to use 16bit, but it does not seem this hard to do)

Sineos:
> If you change the pullup to 1000 ohm you have about 1 °C/step of the ADC, which is maybe acceptable but far from decent.

Are you sure with this calculation? A PT100 changes its resistance between 0°C (100 Ohm) and 260°C (197.71 Ohm) by a mere 97 Ohm, which dictates the voltage drop and needs to resolved by ADC. IMO this does not depend on the pull-up and is simply too low for the typical 12bit ADC

A PT1000 changes its resistance between 0°C (1000 Ohm) and 260°C (1977.12 Ohm) by 977 Ohm, which is more than enough to get a decent resolution.

dewi-ny-je:
> > If you change the pullup to 1000 ohm you have about 1 °C/step of the ADC, which is maybe acceptable but far from decent.
> 
> Are you sure with this calculation? A PT100 changes its resistance between 0°C (100 Ohm) and 260°C (197.71 Ohm) by a mere 97 Ohm, which dictates the voltage drop and needs to be resolved by ADC. IMO this does not depend on the pull-up and is simply too low for the typical 12bit ADC
> 
> A PT1000 changes its resistance between 0°C (1000 Ohm) and 260°C (1977.12 Ohm) by 977 Ohm, which is more than enough to get a decent resolution.

I prepared an XLS calculating the exact values, but to be honest it was meant for PT1000. For PT100 I simply divided by 10, assuming the other coefficients are the same. According to https://www.kongsberg.com/globalassets/maritime/km-products/product-documents/tsiec751_ce.pdf (footnote) it should be correct.

Check the XLS.
[PT1000 resolution.xlsx](https://github.com/Klipper3d/klipper/files/8116075/PT1000.resolution.xlsx)

dewi-ny-je:
> > A PT100 with a standard 4700 ohm pullup resistor has a resolution of about 3.5 °C/step of the ADC converter. It's way way too rough for any decent temperature regulation.
> 
> If the device I am using has a 12 bit or 16 bit ADC resolution? would it still be a problem? (maybe I would need to tell klipper how to configure the adc peripheral in order to use 16bit, but it does not seem this hard to do)

If you use 1000 ohm pullup (and you get a current higher than what it's recommended!) you would get 0.25 °C/step, which is in principle ok, but what about noise? we are talking about very tiny voltage increment here. Have you actually tried on a test board to see if it works?

Also, which boards offer 16 bit ADC?

Basically your proposed change requires an uncommon 16 bit ADC and requires the use of a current much higher than what it's recommended. It's more like asking for troubles and it helping basically no one (the only one may be you). Changes which impact an extremely small number of users are usually not accepted. If you really want to have it, you can clone klipper to your own repo, make the change, and tell Moonraker (if you sue it) to update Klipper from your repository.

dewi-ny-je:
XLS with calculation for @Sineos 
[PT1000 resolution.xlsx](https://github.com/Klipper3d/klipper/files/8118507/PT1000.resolution.xlsx)

Sineos:
I'm not an expert in this field and I happily stand corrected but it seems you are right @dewi-ny-je 
https://docs.google.com/spreadsheets/d/1nXyBU2pkLM6LxLrdwj021RS8CE3KMnNbjorMpSoazds/edit?usp=sharing

According to my calculation we get ~0.5°C resolution for a PT1000 and ~3.2°C for a PT100 (with a 4.7 kOhm pullup) and ignoring all other issues like driving a PT100 with a way to high current.

> I don't want to pile on, but this is correct and it will create extra burden for people like me who are helpers on Discord and Discourse. Users will see PT100 listed as an option, connect one directly to the thermistor port of a standard board with 12bit ADC and 4.7k pullup, and then they'll complain and seek help when it doesn't work.

Fully agree with this. At the current state 16bit or more ADC are not common and in no way warrant to allow such a combination where a 10 USD MAX31865 board allows to connect such a probe correctly.
And even if such 16bit ADC would be more common it would still be the wrong way to connect, as typically PT100 are only supplied with few hundred micro-ampere current to avoid warming. This is also not given with driving it over a voltage divider.

mngr0:
Thanks everyone for the replies, I appreciate the feedback.
I actually made my own board with an ATSAME, the whole family of controllers have ADCs capable of 16bit resolution, I thought it was something that can be found in recent microcontrollers.
I was wrong, I am sorry to have wasted the time of contributors.

PR #5259, opened 2022-02-13
docs: Drop duplicated firmware_retraction G-Codes section

wlhlm:
Diff may look a bit unintuitive, but previously there were two `[firmware_retraction]` sections in the G-Codes docs.


wlhlm:
Doh!

Fixed.

PR #5260, opened 2022-02-13
docs: Specify the meaning of minutes for bed tramming

kubik369:
The documentation does not specify whether the "minutes" in the output of `SCREWS_TILT_CALCULATE` mean minutes of a degree (1/60th of a degree) or minutes of a clock-face (6 degrees). The commit in this PR notes this distinction.

Personally, I just encountered this and was wondering how the hell am I supposed to turn the wheel by 7', as that is an incredibly small angle :sweat_smile: I was told that it is obvious that the output of the command means minutes of a clock-face (e.g. 1 minute = 6 degrees), but I don't agree and hence this PR.

### DIFF
--- a/docs/Manual_Level.md
+++ b/docs/Manual_Level.md
@@ -179,6 +179,9 @@ Recv: // rear right screw : y=155.0, y=190.0, z=2.71500 : adjust CCW 00:50
 Recv: // read left screw : x=-5.0, y=190.0, z=2.47250 : adjust CW 00:02
 Recv: ok
 ```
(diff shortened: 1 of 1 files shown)


bevanweiss:
It seems strange to even use minutes, I would have thought 1.1 turns, or 1.2 turns would be less confusing.
Klipper does seem to match Marlin on this front.  But I think they're both wrong :P

KevinOConnor:
I updated the documentation with a note (commit 92de6e91).

-Kevin

PR #5262, opened 2022-02-14
Update printer-tronxy-x5sa-v6-2019.cfg

Polar-Ted:
Lines 94 and 98. Fan speed and max power should be 1 and not 0.5
removed lines to use default settings. 

[heater_fan hotend_fan]
pin: PG14

[fan]
pin: PG13

Signed off by: Scott Schering sschering@gmail.com


ld3300 (approved):
Looks good to me.
-David

PR #5263, opened 2022-02-14
docs: Fix Translation Support Script

dingyifei:

### DIFF
--- a/docs/_klipper3d/fetch-translations.sh
+++ b/docs/_klipper3d/fetch-translations.sh
@@ -3,7 +3,7 @@
 # This is a make shift script before the current structure of
 # Klipper-translations can be directly utilized by mkdocs
 
-#git clone --depth 1 https://github.com/Klipper3d/klipper-translations
+git clone --depth 1 https://github.com/Klipper3d/klipper-translations
 
 while IFS="," read dirname langname langdesc note; do
   # move and rename markdown files
@@ -26,4 +26,4 @@ while IFS="," read dirname langname langdesc note; do
 
   # add to translations.yml
   echo "      ${langname}: ${langdesc}" >> ./docs/_klipper3d/translations.yml
-done <  <(egrep -v '^ *(#|$)' ./klipper-translations/active_translations | tail -n +2)
+done <  <(egrep -v '^ *(#|$)' ./klipper-translations/active_translations)


KevinOConnor:
Okay, thanks.  I committed this change.

However, the main website shows an icon I don't recognize for selecting the language (at least on firefox):

klipper-cap

I'm referring to the icon in the top right, between the lightbulb and the magnifying glass.

Do you know what that icon means, and how it can be changed?  I think it should be changed.

-Kevin

dingyifei:
image

I think this icon is default to the theme, I'm guessing the 文/A just means "language", since 文 is just "word" in Chinese. I don't think the icon change when I add language. I'll take a look and see if I can change it to a more intuitive one.

KevinOConnor:
Okay, thanks.  I guess I've never seen that icon before (perhaps due to my not knowing very many languages).

I think the "material language icon" (eg, https://www.w3resource.com/icon/material-icons/action-icons/language.php or https://fontawesomeicons.com/materialdesign/icons/language#! ) is a little more intuitive (a wire outline of a globe).

I'm not sure how to change that in mkdocs though.

-Kevin

dingyifei:
I found the globe icon in material theme, but I can't figure out how to use it.
image

I'm not very familiar with the material theme and MkDocs either. 

Tried to modify mkdocs.yml but it seems like I can only add icons but not replace icons.

@damlobster could you take a look and see if there is a way to replace the language selector icon?

https://squidfunk.github.io/mkdocs-material/setup/changing-the-language

KevinOConnor:
I was able to change the icon (commit 131cca2b).  I'm not sure the new "deathstar glyph" is a great choice, but I found the existing icon to be confusing.  Another option I looked at was setting the icon to "fontawesome/solid/language", but that icon rendered really tiny for some reason.

-Kevin

Sineos:
The new language feature has a side impact on the search function. Mixing up results from the different langs:

grafik

dingyifei:
This is strange, the i18n-static is supposed to separate the search index between languages, I'll look into it

dingyifei:
It seems like this line has something to do with the issue

```
WARNING  -  Language 'zh' is not supported by lunr.js, not setting it in the
            'plugins.search.lang' option
```

KevinOConnor:
Any further thoughts on the search functionality?  It seems to me that, effectively, the website search functionality no longer works (too many false positives due to matches in other languages).  If there isn't  a short term fix, I'd be inclined to disable i18n support until we can figure out a solution.

-Kevin

dingyifei:
> Any further thoughts on the search functionality? It seems to me that, effectively, the website search functionality no longer works (too many false positives due to matches in other languages). If there isn't a short term fix, I'd be inclined to disable i18n support until we can figure out a solution.

I did some research and it seems like if we want to automatically generate multiple sites, we need to add a big chunk of stuff in MkDocs configuration and the fetch translations script. https://github.com/squidfunk/mkdocs-material/discussions/2346. I think a short-term fix is to just remove i18n support and temporarily host i18n stuff on Klipper translations gh page?

dingyifei:
#5322

PR #5270, opened 2022-02-16
config: Expose options awaiting to be saved

Misterke:
Adds a save_config_pending_items to the status reported by
configfile reflecting the items and values that a future
SAVE_CONFIG would actually persist.

Signed-off-by: Kurt Haenen <kurt.haenen@gmail.com>

### DIFF
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -341,7 +341,7 @@ class PrinterConfig:
         svalue = str(value)
         self.autosave.fileconfig.set(section, option, svalue)
         pending = dict(self.status_save_pending)
-        if not section in pending:
+        if not section in pending or pending[section] is None:
             pending[section] = {}
         else:
             pending[section] = dict(pending[section])
@@ -350,12 +350,18 @@ class PrinterConfig:
         self.save_config_pending = True
         logging.info("save_config: set [%s] %s = %s", section, option, svalue)
     def remove_section(self, section):
-        self.autosave.fileconfig.remove_section(section)
-        if section in self.status_save_pending:
+        if self.autosave.fileconfig.has_section(section):
+            self.autosave.fileconfig.remove_section(section)
+            pending = dict(self.status_save_pending)
+            pending[section] = None
+            self.status_save_pending = pending
+            self.save_config_pending = True
+        elif (section in self.status_save_pending and
+              self.status_save_pending[section] is not None):
             pending = dict(self.status_save_pending)
             del pending[section]
             self.status_save_pending = pending
-        self.save_config_pending = True
+            self.save_config_pending = True
     def _disallow_include_conflicts(self, regular_data, cfgname, gcode):
         config = self._build_config_wrapper(regular_data, cfgname)
         for section in self.autosave.fileconfig.sections():
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -140,6 +140,7 @@ class PrinterConfig:
         self.autosave = None
         self.deprecated = {}
         self.status_raw_config = {}
+        self.status_save_pending = {}
         self.status_settings = {}
         self.status_warnings = []
         self.save_config_pending = False
@@ -331,17 +332,29 @@ class PrinterConfig:
         return {'config': self.status_raw_config,
                 'settings': self.status_settings,
                 'warnings': self.status_warnings,
-                'save_config_pending': self.save_config_pending}
+                'save_config_pending': self.save_config_pending,
+                'save_config_pending_items': self.status_save_pending}
     # Autosave functions
     def set(self, section, option, value):
         if not self.autosave.fileconfig.has_section(section):
             self.autosave.fileconfig.add_section(section)
         svalue = str(value)
         self.autosave.fileconfig.set(section, option, svalue)
+        pending = dict(self.status_save_pending)
+        if not section in pending:
+            pending[section] = {}
+        else:
+            pending[section] = dict(pending[section])
+        pending[section][option] = svalue
+        self.status_save_pending = pending
         self.save_config_pending = True
         logging.info("save_config: set [%s] %s = %s", section, option, svalue)
     def remove_section(self, section):
         self.autosave.fileconfig.remove_section(section)
+        if section in self.status_save_pending:
+            pending = dict(self.status_save_pending)
+            del pending[section]
+            self.status_save_pending = pending
         self.save_config_pending = True
     def _disallow_include_conflicts(self, regular_data, cfgname, gcode):
         config = self._build_config_wrapper(regular_data, cfgname)
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -41,6 +41,8 @@ The following information is available in the `configfile` object
   here.) All values are returned as strings.
 - `save_config_pending`: Returns true if there are updates that a
   `SAVE_CONFIG` command may persist to disk.
+- `save_config_pending_items`: Contains the sections and options that
+  were changed and would be persisted by a `SAVE_CONFIG`.
 - `warnings`: A list of warnings about config options. Each entry in
   the list will be a dictionary containing a `type` and `message`
   field (both strings). Additional fields may be available depending


Misterke:
Main reason for this would be to allow my macros (and those of other people) to be able to see what the current "state" would be that would get persisted by a SAVE_CONFIG.  Currently if you see in a GUI that a config save is pending, there's always the risk that you might be saving something that you did not really want.  With this change at least macros (what I'm using it for) or other solutions (GUIs) can be built to indicate what you are actually going to save.

dewi-ny-je:
It would be good then to also have a selective "SAVE_CONFIG", to save only what is desired. Options like "MESH", "ENDSTOPS", ...

Misterke:
> It would be good then to also have a selective "SAVE_CONFIG", to save only what is desired. Options like "MESH", "ENDSTOPS", ...

I personally would agree, but I think this would be opening Pandora's box.  Those values are being stored/changed by the various modules in Klipper (and added as extension to Klipper).  There might be some relationships between such values which could be disturbed if you would save one value, but not some other, which could then cause the reset after SAVE_CONFIG to have a broken config.

With the currently proposed PR tools or GUIs can at least check the values being stored and compare each with the matching static config item.  For items the tool/GUI really knows, it could then reset things to the original config using the appropriate gcode command for that item.  This would have to be programmed specifically for each (set of) item(s), but at least would ensure that any relationship between the items would be retained.

dewi-ny-je:
@Misterke  true.
Resetting what should not be saved is also an option, but I was specifically thinking about those two use cases: "SAVE_CONFIG MESHES" and "SAVE_CONFIG LIMITS" (endstop, z offset).
Those two should be easy to isolate.

Both alternatives are fine, as long as selective resetting before saving is possible (I was not aware of it).

Misterke:
> @Misterke true. Resetting what should not be saved is also an option, but I was specifically thinking about those two use cases: "SAVE_CONFIG MESHES" and "SAVE_CONFIG LIMITS" (endstop, z offset). Those two should be easy to isolate.

I don't feel this belongs in this pull-request as I don't think the configfile.py code should contain any knowledge on what other modules do with the stored data.  The current PR delivers a mechanism to see what config items are being saved while remaining agnostic on what those might actually mean.  Similar to what is provided for the stored config.  Also there configfily.py is just giving access to it while remaining agnostic on what each of the items might mean.

Knowledge on how to reset something you changed, but haven't saved with SAVE_CONFIG yet would then either belong in the modules that added those config items (so, for meshes bed_mesh.py and for limits probably stuff like probe.py) which in fact could be added without the change proposed here *OR* it could be added at a higher level, ex. in a GUI or in Macros by making those have knowledge on what the various config items mean - which is often already the case - and how to change the associated values (like the Heightmap in Mainsail or the items to Z baby-step).  For that latter category, this change is important as it gives insight in the currently active value of each of the config values rather than only seeing the value the system was started with.

Misterke:
Still no reviewer picking this up?  The turn-around on pull-requests on Klipper is just ridiculously slow.  I fear this will be the nail in the coffin for Klipper ...  I'm not even entering pull-requests anymore for my further changes as there's anyway hardly a chance of it getting picked up in due time.

Misterke:
Still no reviewer on this pull-request.  I honestly have given up on attempting to get any changes pushed upstream for this project.  I'll just maintain my own fork and regularly rebase them on top of what's happening on the main klipper repo.  I'll still leave this pull-request open, but in the meanwhile don't expect much anymore ...

dewi-ny-je:
@Misterke have you tried asking in the Discord and Discourse from the project? Maybe if you advertise it a bit someone will do it.
It would be a pity not to have it merged

Misterke:
> @Misterke have you tried asking in the Discord and Discourse from the project? Maybe if you advertise it a bit someone will do it. It would be a pity not to have it merged

I have tried to raise interest in the past on Discord, both on the configuration and dev channels, but I do not get the feeling the few "reviewers" for the Klipper project are interested in progressing the project with ideas other than their own interests.  I can understand attempts to control what goes into the code-base as also on Marlin some things I tried to get in, failed to be accepted, *BUT* the Marlin PR was at least picked up and comments were provided on why the reviewer / project was not in favor of my proposed change.  That I find good (though still frustrating for me) behavior, but on Klipper PRs just sit there for ages with no input whatsoever from the repo owners, which I find very disappointing, but will not spend more of my time on.

klippy/configfile.py
```diff
@@ -331,17 +332,29 @@ def get_status(self, eventtime):
...
+            pending[section] = {}
+        else:
+            pending[section] = dict(pending[section])
+        pending[section][option] = svalue
+        self.status_save_pending = pending
         self.save_config_pending = True
         logging.info("save_config: set [%s] %s = %s", section, option, svalue)
     def remove_section(self, section):
         self.autosave.fileconfig.remove_section(section)
+        if section in self.status_save_pending:
```
Arksine, line 354: This is correct in that it would remove items in the "status_save_pending" dict, however a removed item not added in this session won't be indicated.  For example, if a user saved a mesh profile in this session and then subsequently removed it then it is correctly handled...however if the profile was saved in a previous section and it was removed there is no indication that the removal is pending.

Arksine:
I understand your frustration, however it is isn't personal, its just a matter of resources.  When time is limited and split among multiple projects reviewers do need to prioritize our work, and indeed that is prioritization is going to be based in part on items of specific interest to us.  Its easy to miss an open PR as there are always multiple items competing for our attention.

With specific regard to this pull request, it looks ok to me, I just had one observation about pending removals.

KevinOConnor:
> That I find good (though still frustrating for me) behavior, but on Klipper PRs just sit there for ages with no input whatsoever from the repo owners, which I find very disappointing, but will not spend more of my time on.

I apologize for the silence.  A few months ago, I tried to relay through Discord, github, and Discourse that I'd be unable to review all PRs.  Going forward I hope that we'll get more members of the community involved with PRs.

-Kevin

klippy/configfile.py
Misterke: I think I understand what you mean, but let me phrase it my way to see whether we're on the same line: the status_save_pending is not really reflecting when an item is removed as the removal of an item is not flagged explicitly in the dict as a removal, just the absence of the value.

An alternative could be to add or keep the entry in the dict, but give it a None value.  Would that be better?

Arksine: You could assign the section a `None` value if its removed, however it might require some additional tracking to make sure it was there when the session started.  Another option would be to report the autosave data as a dict.  Then users would have a view of exactly what the "saved" portion of the configuration will be when they issue `SAVE_CONFIG`.  I'd be curious as to what Kevin and Meteyou would think is the best approach.

Misterke: I guess I'm still somehow missing your exact point.  We seem to be on the same line on what the goal should be: having a clear view on the current settings AND how those differ from the already saved settings (which is also the settings at start of the session as a SAVE_CONFIG triggers a restart).

My main goal would be that a GUI (Ex. Mainsail) or a set of macros should be able to easily figure out which config values have not yet been saved to reflect that to the user or - which is what I use it for - store those in some other way that does not require a restart and then restore them on the next startup to avoid they get lost.  Why would they get lost?  Because after ex. z-adjust at the beginning of a print, you cannot do a SAVE_CONFIG without killing the print and I have totally separate system that monitors the printer and shuts down power when it is cold for 15 minutes.  So, if power is taken off 15 minutes after the print ended, my z-offset is gone - were it not for my macros saving stuff in saved variables and then at startup restoring all that.  Question then is: why do I need to make such tricky macros for this?

In general I guess the problem is running deeper inside Klipper.  There are settings which for very practical reasons have to be applied at the start of a session (pin-configurations and other hardware-related stuff).  However, there are numerous settings which can be dynamically modified at run-time ((additional) z-offset, bed profiles, max.speeds/accel, ...) yet for some reason also these values can only be saved in combination with a restart.  Ultimately I think it would make more sense to have 2 separate types of config, one for stuff that cannot change dynamically (and might even be read-only, like the pin-config) and one for stuff that is dynamic and could hence be saved at any point.  Even then an "original values" vs "current values" view via 2 dicts might be useful, but the need for it would already be much lower.

dewi-ny-je: Having to restart to save various settings is indeed annoying.
Even a "save without kill" would be a definite improvement

KevinOConnor:
FWIW, I don't have a strong preference on how a deleted section is represented in save_config_pending_items.  Setting the section to None or an empty dict seems fine to me.

-Kevin

klippy/configfile.py
Misterke: @Arksine , I'm having problems figuring out what was part of the original config or not.  Something like:
```
    def remove_section(self, section):
        if self.autosave.fileconfig.has_section(section):
            self.autosave.fileconfig.remove_section(section)
            pending = dict(self.status_save_pending)
            pending[section] = None
            self.status_save_pending = pending
            self.save_config_pending = True
        elif section in self.status_save_pending and self.status_save_pending[section] is not None:
            pending = dict(self.status_save_pending)
            del pending[section]
            self.status_save_pending = pending
            self.save_config_pending = True
```
does not seem to work because autosave.fileconfig is reflecting current status, not the original unchanged config.  Am I overlooking something or do I really have to make a copy of the initial config (RawConfigParser).  The latter seems pretty heavy and I wonder whether knowing the distinction between those two is worth it?

Arksine: Having given it some thought I'm not sure that the distinction is necessary.  If `self.status_save_pending` has a section set to `None` then a front end or macro will conclude that the section will not be saved when `SAVE_CONFIG` is run.   I think the result it the same regardless of whether or not the section was initially loaded.

Misterke: Ok.  I do still want to check the effect of these None values further.  If that behaves nicely, I'll update the pull-request with this change.

Misterke: It required some extra fixes to get through the tests - one real issue due to the "None" value now possible for a section which I had apparently not covered in my usage the past days, but luckily was exposed by the tests.

Misterke:
I hope this can be merged now.  I've rebased on top of the latest master and handled the remark of @Arksine .

Arksine (approved):

Arksine:
Looks good to me.

KevinOConnor:
Thanks.

-Kevin

PR #5274, opened 2022-02-17
fix rotation_distance for sapphire pro

BastelKlug:
Hi, the sapphire pro config contains wrong rotation_distance values for X, Y and Z. I have changed them and it works perfectly. TwoTrees must have decided to use 20T instead of 16T pulleys and T8x2 lead screw instead of T8x8.

For this reason I have started the PR :)

### DIFF
--- a/config/printer-twotrees-sapphire-pro-2020.cfg
+++ b/config/printer-twotrees-sapphire-pro-2020.cfg
@@ -18,7 +18,8 @@ step_pin: PE3
 dir_pin: !PE2
 enable_pin: !PE4
 microsteps: 16
-rotation_distance: 32
+rotation_distance: 40
+#rotation_distance: 32
 endstop_pin: !PA15
 position_endstop: 0
 position_max: 230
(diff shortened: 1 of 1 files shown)


ld3300:
Hi @BastelKlug,
While I can't test the code, I don't see any problems.  Is there a reason you commented out the previous values instead of removing them completely?  If they are not applicable at all you should delete the lines and update the commit.

Also please edit your original commit comment on here to add the "Signed-off-by: My Name <myemail@address.org>" as described [Here](https://www.klipper3d.org/CONTRIBUTING.html?h=signed#format-of-commit-messages)

config/printer-twotrees-sapphire-pro-2020.cfg
```diff
@@ -18,7 +18,8 @@ step_pin: PE3
 dir_pin: !PE2
 enable_pin: !PE4
 microsteps: 16
-rotation_distance: 32
+rotation_distance: 40
+#rotation_distance: 32
```
Misterke, line 22: Why keep the old rotation distance as a comment?  Does this mean there are multiple flavors of the Sapphire Pro?  If so, would a comment explaining when to use what not make more sense than just a second value in a comment?

PR #5281, opened 2022-02-17
temperature_fan: add missing round function to get_status temperature

meteyou:
The only thing i noticed about temperature_fan was that the rounding command was missing.

### DIFF
--- a/klippy/extras/temperature_fan.py
+++ b/klippy/extras/temperature_fan.py
@@ -72,7 +72,7 @@ class TemperatureFan:
         return self.max_speed
     def get_status(self, eventtime):
         status = self.fan.get_status(eventtime)
-        status["temperature"] = self.last_temp
+        status["temperature"] = round(self.last_temp, 2)
         status["target"] = self.target_temp
         return status
     cmd_SET_TEMPERATURE_FAN_TARGET_help = \


ld3300:
Hi @meteyou 
I can't comment on the commit directly, but you will need to edit your comment above to include the "Signed-off-by:" statement to have your commit accepted.  See [Here](https://www.klipper3d.org/CONTRIBUTING.html?h=signed#format-of-commit-messages)

meteyou:
@ld3300 thx for the info, but i see the `Signed-off-by:` part in the commit itself. this should be default in my IDE...

image

ld3300:
my bad, didn't know you could do that.  Thank you.

meteyou:
@ld3300 i use jetbrains pycharm for klipper (python projects)

KevinOConnor:
Thanks.  Why do we want to round the reported value?

-Kevin

meteyou:
@KevinOConnor this was originally a PR by @Arksine to simply omit unnecessary comma digits for the api.

this was the PR: https://github.com/Klipper3d/klipper/pull/4795

KevinOConnor:
Thanks.

-Kevin

Issue #5290, opened 2022-02-20
GET_POSITION command returns incorrect gcode position with offsets.


TypQxQ:
Issuing a GET_POSITION command when using offsets should return the position where the g-code command expects it to be:

Using `Offsets: X:11.28 Y:4.73 Z:7.05`, I issue the command `G0 X500 Y5`

GET_POSITION now returns:
```
mcu: stepper_x:-79039 stepper_x1:-79039 stepper_y:82647 stepper_y1:82647 stepper_z:-37 stepper_z1:-25 stepper_z2:-44
stepper: stepper_x:521.012500 stepper_x1:521.012500 stepper_y:501.550000 stepper_y1:501.550000 stepper_z:35.136873 stepper_z1:35.136873 stepper_z2:35.136873
kinematic: X:511.281250 Y:9.731250 Z:35.136873
toolhead: X:511.280000 Y:9.730000 Z:35.136873 E:0.000000
gcode: X:500.000000 Y:5.000000 Z:28.086873 E:0.000000
gcode base: X:11.280000 Y:4.730000 Z:7.050000 E:0.000000
gcode homing: X:11.280000 Y:4.730000 Z:7.050000
```

I expect the gcode to show the position I send with gcode, not the toolhead location:
`gcode: X:500.000000 Y:5.000000 Z:28.086873 E:0.000000`

In gcode_move.py:
In GET_POSITION, the gcode: line returns the self.last_position.
This is updated in G1 for every move: `self.last_position[pos] = v + self.base_position[pos]`

So, the returned gcode: line should return the `self.last_position - self.base_position` to account for the offset.

TypQxQ:
The GET_POSITION "gcode" is "relative to the coordinate system specified in the config file".
"The M114 command (gcode_move.get_status()['gcode_position']) will report the last g-code position relative to the current g-code coordinate system."

Reference: https://www.klipper3d.org/Code_Overview.html#coordinate-systems

PR #5291, opened 2022-02-20
GET_POSITION gcode position fix

TypQxQ:
The gcode position now returns the position with offsets applied. This is the position you reached with a G1 command.
This fixes issue #5290.
Signed-off-by: Andrei Ignat andrei@ignat.se

### DIFF
--- a/klippy/extras/gcode_move.py
+++ b/klippy/extras/gcode_move.py
@@ -256,11 +256,8 @@ class GCodeMove:
         kin_pos = " ".join(["%s:%.6f" % (a, v) for a, v in kinfo])
         toolhead_pos = " ".join(["%s:%.6f" % (a, v) for a, v in zip(
             "XYZE", toolhead.get_position())])
-        last_pos = self.last_position
-        for pos in range(len(last_pos)):
-            last_pos[pos] = last_pos[pos] - self.base_position[pos]
         gcode_pos = " ".join(["%s:%.6f"  % (a, v)
-                              for a, v in zip("XYZE", last_pos)])
+                              for a, v in zip("XYZE", self._get_gcode_position())])
         base_pos = " ".join(["%s:%.6f"  % (a, v)
                              for a, v in zip("XYZE", self.base_position)])
         homing_pos = " ".join(["%s:%.6f"  % (a, v)
--- a/klippy/extras/gcode_move.py
+++ b/klippy/extras/gcode_move.py
@@ -256,8 +256,11 @@ class GCodeMove:
         kin_pos = " ".join(["%s:%.6f" % (a, v) for a, v in kinfo])
         toolhead_pos = " ".join(["%s:%.6f" % (a, v) for a, v in zip(
             "XYZE", toolhead.get_position())])
+        last_pos = self.last_position
+        for pos in range(len(last_pos)):
+            last_pos[pos] = last_pos[pos] - self.base_position[pos]
         gcode_pos = " ".join(["%s:%.6f"  % (a, v)
-                              for a, v in zip("XYZE", self.last_position)])
+                              for a, v in zip("XYZE", last_pos)])
         base_pos = " ".join(["%s:%.6f"  % (a, v)
                              for a, v in zip("XYZE", self.base_position)])
         homing_pos = " ".join(["%s:%.6f"  % (a, v)
--- a/klippy/extras/gcode_move.py
+++ b/klippy/extras/gcode_move.py
@@ -257,7 +257,7 @@ class GCodeMove:
         toolhead_pos = " ".join(["%s:%.6f" % (a, v) for a, v in zip(
             "XYZE", toolhead.get_position())])
         gcode_pos = " ".join(["%s:%.6f"  % (a, v)
-                              for a, v in zip("XYZE", self._get_gcode_position())])
+                         for a, v in zip("XYZE", self._get_gcode_position())])
         base_pos = " ".join(["%s:%.6f"  % (a, v)
                              for a, v in zip("XYZE", self.base_position)])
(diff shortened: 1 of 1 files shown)


klippy/extras/gcode_move.py
```diff
@@ -256,8 +256,11 @@ def cmd_GET_POSITION(self, gcmd):
         kin_pos = " ".join(["%s:%.6f" % (a, v) for a, v in kinfo])
         toolhead_pos = " ".join(["%s:%.6f" % (a, v) for a, v in zip(
             "XYZE", toolhead.get_position())])
+        last_pos = self.last_position
+        for pos in range(len(last_pos)):
+            last_pos[pos] = last_pos[pos] - self.base_position[pos]
         gcode_pos = " ".join(["%s:%.6f"  % (a, v)
-                              for a, v in zip("XYZE", self.last_position)])
```
jschuh, line 260: I think you just need to change this line to 
```
                              for a, v in zip("XYZE", self._get_gcode_position())])
```

jschuh:
@dmbutyugin, I think this PR is identifying a real bug (although I suggested a change to the PR). `GET_POSITION` already provides the unadjusted position of the toolhead in the "kinematic" and "toolhead" lines. So, it seems to follow that the "gcode" line should specifically be the adjusted gcode position (which would also match the implementation of the `M114` command in this file).

For context, here's the current output I get:

```
mcu: stepper_x:76917 stepper_y:76697 stepper_z:-36047
stepper: stepper_x:240.003125 stepper_y:0.000000 stepper_z:10.975000
kinematic: X:120.001563 Y:120.001562 Z:10.975000
toolhead: X:120.001563 Y:120.001562 Z:10.975000 E:0.000000
gcode: X:120.001563 Y:120.001562 Z:10.975000 E:0.000000
gcode base: X:0.000000 Y:0.000000 Z:0.975000 E:0.000000
gcode homing: X:0.000000 Y:0.000000 Z:0.975000
```

And here's the output I get after replacing `self.last_position` in that line of code with `self._get_gcode_position()`. (Notice the difference is in only the "gcode" line.)

```
mcu: stepper_x:76915 stepper_y:76699 stepper_z:-36048
stepper: stepper_x:240.003125 stepper_y:0.000000 stepper_z:10.975000
kinematic: X:120.001563 Y:120.001562 Z:10.975000
toolhead: X:120.001563 Y:120.001562 Z:10.975000 E:0.000000
gcode: X:120.001563 Y:120.001562 Z:10.000000 E:0.000000
gcode base: X:0.000000 Y:0.000000 Z:0.975000 E:0.000000
gcode homing: X:0.000000 Y:0.000000 Z:0.975000
```

klippy/extras/gcode_move.py
dmbutyugin: This is a good idea! Not only it is simpler, it will also report correct E position (accounting for self.extrude_factor)

dmbutyugin:
@jschuh Thank your for your explanation. I was confused and thought that the fix is about the toolhead position. Then it makes sense in my opinion.

TypQxQ:
@jschuh Thank you for explaining and finding a better way to fix it!
I just checked the change you proposed and I confirm that it works as expected.

klippy/extras/gcode_move.py
TypQxQ: This is much better than the original fix!
Thank you!

TypQxQ (commented):
Thank you!

dmbutyugin:
TBH, the 4b0bb4a07dd10091623409fed60aeeacb75aa049 makes indentation a bit strange, it might be worthwhile to add line breaks elsewhere instead. But otherwise the fix looks very reasonable, thanks!

TypQxQ:
@dmbutyugin I am struggling with the code formatting 😄 
I don't hold my breath it ever getting easier for me.

dmbutyugin:
@TypQxQ you may try it, for instance, as
```
        gcode_pos = " ".join(
                ["%s:%.6f"  % (a, v)
                 for a, v in zip("XYZE", self._get_gcode_position())])
```

dmbutyugin (approved):
OK, thanks!

klippy/extras/gcode_move.py
```diff
@@ -256,8 +256,9 @@ def cmd_GET_POSITION(self, gcmd):
         kin_pos = " ".join(["%s:%.6f" % (a, v) for a, v in kinfo])
         toolhead_pos = " ".join(["%s:%.6f" % (a, v) for a, v in zip(
             "XYZE", toolhead.get_position())])
-        gcode_pos = " ".join(["%s:%.6f"  % (a, v)
-                              for a, v in zip("XYZE", self.last_position)])
+        gcode_pos = " ".join(
+                 ["%s:%.6f"  % (a, v)
+                 for a, v in zip("XYZE", self._get_gcode_position())])
```
jschuh, line 261: I've noticed that most of the Klipper code aligns the arguments to the opening parenthesis when they spillover like this, so I formatted it this way when I tested it out this morning:

```
         gcode_pos = " ".join(["%s:%.6f"  % (a, v)
                              for a, v in zip("XYZE",
                                              self._get_gcode_position())])
```

KevinOConnor:
The current output is intentional - see: https://www.klipper3d.org/Code_Overview.html#coordinate-systems .  I'm not sure what the gain of changing the output is - to obtain the g-code position relative to the g-code coordinate system one can use the `M114` command.

-Kevin

KevinOConnor:
FYI, I committed c31f9815 so G-Codes.md has a link to the above info.

-Kevin

TypQxQ:
@KevinOConnor Thank you! I think understand now.

PR #5294, opened 2022-02-21
Support [extruder] sections without a stepper

KevinOConnor:
This PR makes it possible to define an extruder config section without an extruder stepper motor.  This may be useful for dual hotend systems that share a single stepper motor.

A bit unrelated, but I'm also added a commit here that removes the default for the "extruder" parameter of `[extruder_stepper]` config sections.  With the recent extruder changes, I suspect that the default of synchronizing all extruder_stepper motors to the "extruder" motion queue is confusing.  So, after this PR, users will be required to specify the "extruder" parameter of `[extruder_stepper]` config sections (either by setting it to something like `extruder: extruder` to set the default motion queue or set it as `extruder: ` to not set a default motion queue at startup).

@drumclock - fyi.

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,12 @@ All dates in this document are approximate.
 
 ## Changes
 
+20220304: There is no longer a default for the `extruder` parameter of
+[extruder_stepper](Config_Reference.md#extruder_stepper) config
+sections. If desired, specify `extruder: extruder` explicitly to
+associate the stepper motor with the "extruder" motion queue at
+startup.
+
 20220210: The `SYNC_STEPPER_TO_EXTRUDER` command is deprecated; the
 `SET_EXTRUDER_STEP_DISTANCE` command is deprecated; the
 [extruder](Config_Reference.md#extruder) `shared_heater` config option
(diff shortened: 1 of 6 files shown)


DrumClock:
Hi @KevinOConnor 
I try it and I get the following results:
`pressure_advance` and `pressure_advance_smooth_time` parameters
cannot be specified without a stepper definition.
```
Option 'pressure_advance' is not valid in section 'extruder1'
Option 'pressure_advance_smooth_time' is not valid in section 'extruder1'
```

**It should be like this ?**

Maybe I would leave it in the hotend definition and then assign it to 
the sync stepper using `SET_PRESSURE_ADVANCE EXTRUDER.`

The point is that if the second hotend needed different parameters, they would have to be stored in variables and then used. 
This way they would be define in the configuration and it would probably be clearer.
(just my idea, I don't know how it is further connected with FW)

**Everything else works as needed.**  
The change is made by the `SET_HOTEND macro`

Test configuration:  [printer.cfg.txt](https://github.com/Klipper3d/klipper/files/8122761/printer.cfg.txt)

KevinOConnor:
Yes - today pressure advance is associated with the *stepper* and not the extruder motion queue.  So, if there is no stepper associated with extruder1 then it is not possible to assign a pressure advance value to it.

I know this can be confusing, but the alternatives don't seem much better.

Thanks.
-Kevin

DrumClock:
Ok, this can be solved by a variable in the macro.

Is there anything else I need to test?

-Petr

PR #5316, opened 2022-02-28
stm32: add compile options and docs for stm32f103x6

mirokymac:
Signed-off-by: Miroky Mac <toufubomb@gmail.com>


KevinOConnor:
This change doesn't look correct to me, as it does not include the necessary changes to src/stm32/Kconfig .  Also, I'm not sure we want to make custom code changes for each variant of the stm32f103, as usually the same code can run on all the variants (with the possible exception of memory size).

-Kevin

mirokymac:
> This change doesn't look correct to me, as it does not include the necessary changes to src/stm32/Kconfig . Also, I'm not sure we want to make custom code changes for each variant of the stm32f103, as usually the same code can run on all the variants (with the possible exception of memory size).
> 
> -Kevin

The src/stm32/Kconfig has been added.

F103ld has 1 less group of TIM, UART and SPI, beside the smaller memory.
I just want to cut some cost on the pricy STM32F103 md or hd for the smaller project like data collection or toolhead boards.

PR #5320, opened 2022-03-02
Tip on Blobs when printing arcs

bhelm:
After switching to Klipper, i had a big Problem with blobbing (while the same GCode ran fine on the stock ender5+ Firmware). It took me days to figure this out and i feel it should be mentioned on the slicers page. There is also a [Youtube-Video on this](https://www.youtube.com/watch?v=Hvw3DrVAeTA).

Feel free to add/correct what is needed :)

### DIFF
--- a/docs/Slicers.md
+++ b/docs/Slicers.md
@@ -87,3 +87,34 @@ Klipper's maximum extrusion cross-section check.
 
 In contrast, it is okay (and often helpful) to use a slicer's
 "retract" setting, "wipe" setting, and/or "wipe on retract" setting.
+
+## Set lower resolution or use Arcs / Arc-Welder when you have small blobs/zits on curved/round walls
+
+Especially when printing at higher speeds, one may notice blobs/zits
+on curved/round surfaces. These can occur when the slicer creates a
+lot of small lines to over-accurately reproduce an arc, possibly
+with alternating extrusion amounts. These tiny G-Code lines can cause
+klipper to shutter slightly, causing these blobs. You can tell these
+blobs apart from other issues by the fact that these are
+reproducible on the same spots if you print the same model twice.
+If you print a 3d benchy, you may see them on the front hull lines.
+
+There are two ways to resolve this; first is to lower the resolution
+of the slicer. In Cura, these settings can be found under
+"Mesh Fixes" -> "Maximum Resolution", "Maximum Travel Resolution"
+and "Maximum Devaiation".
+
+The second way is to use arcs instead of lines in the slicer. These
+are G02/G03 G-Code-Commands that may be enabled in your slicer, as
+not all printers support these. There is also a tool available
+called Arc-Welder, which post-processes your G-Code and combines
+linear moves into arcs. Depending on your Slicer, Arc-Welder may
+be available as a Plugin.
+
+Klipper needs to be configured to accept these G02/G03 commands by
+adding a new config section [gcode_arcs](Config_Reference.md#gcode_arcs).
+
+Klipper currently takes the G02/G03 commands and converts them
+back into linear commands, defined by the resolution setting.
+While it may sound a bit pointless, it makes sure that these curved
+walls have a proper resolution without confusing extrusion commands.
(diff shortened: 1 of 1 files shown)


jakep82:
I disagree strongly with both of these recommendations, and I don't think they should be suggested as solutions.  If you're using the Octoprint the solution is [here](https://www.klipper3d.org/Config_Reference.html#virtual_sdcard).

Biorn1950:
The problem you described is due to octoprint. Not klipper.

dmbutyugin:
Indeed a more appropriate solution for Klipper specifically is to use `[virtual_sdcard]` which is supported with all web UIs, including octoprint, fluidd and mainsail.

bhelm:
@Biorn1950 I was not using octoprint, im with fluidd on a pi 3 b+
@dmbutyugin also i was using the [virtual_sdcard] from day 1.

As the only working fix comes down to generate a gcode without many tiny moves in arcs. It happens with the default settings for the ender 5 plus on cura **and** superslicer. These default settings are printer dependent, so if the slicer has a very high resolution dialed in for specific printers that works for them, it can mean it stops working as soon as someone switches over to Klipper.

That happened to me and it was hard to figure it out... i read the slicers doc.. checked everything and thought "ok, my G-Code printed on the stock firmware, i have all options set like it was recommended by the klipper slicers doc, so the G-Code is fine".
then debugged everything else.. printed 15 bencies... before i found that youtube video that told me that these tiny arc moves can cause havoc.

I also printed the same GCode on the stock firmware without these blobs.

It is a slicer setting (on some printer profiles on some slicers) that works on the stock firmware but does cause quality issues after switching to Klipper. So i feel it needs to be mentioned on the klipper documentation, especially as it is very hard to pin it down to its root cause.

Biorn1950:
@bhelm very very odd the guys I saw reported this issue was using exclusively Octoprint. That why I answered that.

dewi-ny-je:
I personally am not completely against mentioning the lowering of the resolution, provided it is explicitly mentioned as backup solution when other ones don't work (virtual sd card, fluidd/Mainsail, ...). It must be clear which is a workaround to be avoided as much as possible.
Also, it should be stated that slowdowns could be caused by a too low serial speed, for example is due to some issues it goes down to 115 kbps.

What I am surely against is the arc welding. If you use it, first of all the welding process introduces inaccuracies, second you are INCREASING the load on Klipper because arcs are converted into small moves again. Sure, you reduce the number of GCode lines, but that's unlikely an issue if Fluidd/Mainsail are used and the serial speed is set properly.

Sineos:
I also vote against adding such generalized advice:
* On the one hand such effects depend on your MCU hardware: With a 8bit board such limit is reached far earlier than with a 400 MHz 32 bit board
* Microstep resolution, stepper driver and stepper angle will have an influence
* Then surely the extrusion system (direct vs bowden, flow rate etc) will have an effect
* Postprocessing for arc welding is just adding another layer of complexity and source for errors or bugs: Multiple conversions equals multiple inaccuracies like rounding, approximations etc. Just check the issue tracker of the ArcWelderPlugin to get an idea

bhelm:
@Sineos I agree that using arc welder should be done with caution. I have changed the text to point out that arcwelder may introduce other problems.

KevinOConnor:
Serial port speed settings between OctoPrint and Klipper have no effect at all.  The speed setting is completely ignored.

A too low serial port baud rate between Klipper host and Klipper micro-controller will never result in print pauses.  If the channel does not have sufficient bandwidth then an error will be raised.

A too slow micro-controller will never result in print pauses.  If the micro-controller is unable to keep up with requests then an error will be raised.

-Kevin

jakep82:
It's been marked not for mainline, so it's not going to be merged.  There's no need for further discussion or to push more revisions.  

These suggestions are just workarounds to cover up whatever actual issues exists to cause pauses during a print.  In every case I've seen previously it was caused by Octoprint.  If that isn't the case here, then a post should be opened on Discourse with logs demonstrating the problem so analysis can be performed.

bhelm:
ok, sorry for wasting everyone's time, i tried to help.

As it sounds that this is more a bug on my side than a general problem with klipper, i may investigate and open an issue on it on a later time.

PR #5322, opened 2022-03-03
Multi-site mkdocs configuration

dingyifei:
it addresses the search index problem by building multiple sites at once.


dingyifei:
Update on the status of these changes:
build.sh works locally
workflow still needs modification.

Sineos:
Thanks for keeping it up 👍 
Surely you have seen https://github.com/ultrabug/mkdocs-static-i18n

dingyifei:
> Thanks for keeping it up 👍 
> 
> Surely you have seen https://github.com/ultrabug/mkdocs-static-i18n

Static18n is the solution we currently used, it seems like klipper3d.org's search function is unusable at the current state and the optimal solution seems to be creating this multi site builder.

dingyifei:
It seems like making a custom site builder script can fix some existing issues too. 

Current roadmap: 

- [ ] Workflow push to gh-pages
- [ ] add gettext for navigation
- [ ] add navigation to Weblate
- [ ] Fix config file 404 error
- [ ] Fix Font for other languages
- [ ] Separate language code for indexing support

dingyifei:
It generally works now, the demo sites are:

https://dingyifei.github.io/test-build-klipper3d/zh_HS/
https://dingyifei.github.io/test-build-klipper3d/
The alternate( switch lang ) button doesn't work cuz it only works with domains like https://dingyifei.github.io/ to https://dingyifei.github.io/zh_HS

dingyifei:
There are some known issues with my new commit, where it can't correctly replace the site directory. I'm working on a fix

dingyifei:
Resolved, the new test sites are
https://dingyifei.github.io/test-build-klipper3d/zh-Hant/
https://dingyifei.github.io/test-build-klipper3d/zh/
https://dingyifei.github.io/test-build-klipper3d/

dingyifei:
@mental405 

Could you take review this PR?

This PR fixed search index issue using multiple MkDocs sites.

The basics are:

1. Build.sh build the English documentation first with static configuration. If it fails then the main documentation has something wrong.
2. Build.sh fetch klipper-translations and move files into the proper directories
3. build.sh construct yml for languages and build sites into site/lang
4. English site is moved to default
5. Site folder is pushed to gh-pages through GitHub workflow.

KevinOConnor:
Yes, I'll review this PR.  However, this looks to have some substantial changes and that will take some time to do a full review.

In the meantime, I've disabled the translation support on klipper3d.org (commit c067d4df).  I think this should help give additional time for review.

Also, the current website wasn't building correctly.  https://github.com/Klipper3d/klipper/runs/5469904641?check_suite_focus=true - looks like the `klipper-translations/active_translations` has a definition for `zh-Hant` that isn't supported (`ERROR    -  Config value: 'plugins'. Error: Plugin 'i18n' value: 'languages'. Error: Language code values must be either ISO-639-1 lower case or represented with they territory/region/county codes, received 'zh-Hant' expected forms examples: 'en' or 'en_US'.`).  I'm not sure if that is due to the preparation for this PR or if it is a separate error.

-Kevin

dingyifei:
> Also, the current website wasn't building correctly.  https://github.com/Klipper3d/klipper/runs/5469904641?check_suite_focus=true - looks like the `klipper-translations/active_translations` has a definition for `zh-Hant` that isn't supported (`ERROR    -  Config value: 'plugins'. Error: Plugin 'i18n' value: 'languages'. Error: Language code values must be either ISO-639-1 lower case or represented with they territory/region/county codes, received 'zh-Hant' expected forms examples: 'en' or 'en_US'.`).  I'm not sure if that is due to the preparation for this PR or if it is a separate error.

this is due to the preparation for this PR, since the language code for material mkdocs, material mkdocs search feature, and static i18n use different language codes and they are incompatible with each other. I reused the previous language code column for site language code, leading to this error. The current active_translations is only compatible with the new solution.

dingyifei:
I realized my script might have a flaw, when build.sh fail the workflow wouldn't deploy to gh-pages. The no-translation build should be a fallback option that still gets deployed regardless.

Should I separate build.sh or allow it to take different parameters to address these issues?

dingyifei:
Done

The new deploy workflow:

1. Attempt to run `deploy` to deploy a multi-site MkDocs using build.sh
2. if `deploy` returned a `failed` state, `deploy_fallback` run using build_fallback.sh, which only utilize the `Klipper` repository files
3. Regardless of `deploy`'s final state, the English site (default site) is available.

Example:
https://github.com/dingyifei/test-build-klipper3d/actions/runs/1959944481 `deploy` successful, fallback doesn't run
https://github.com/dingyifei/test-build-klipper3d/actions/runs/1959911403 `deploy` failed, fallback run

dingyifei:
I fixed a few problems in build_fallback.sh and the fallback deploy workflow. They are running correctly now.

dingyifei:
Hi Kevin,
Any updates?

docs/_klipper3d/extra.yml
```diff
@@ -0,0 +1,11 @@
+# Site usage statistics
```
dingyifei, line 1: File header

docs/_klipper3d/nav_en.yml
```diff
@@ -0,0 +1,56 @@
+# Navigation hierarchy (this should mimic the layout of Overview.md)
```
dingyifei, line 1: Explain why is it "en"

dingyifei (commented):
Looks like the majority of my files are ready, but I'll have to squash come commits and add some file headers to describe how everything works together.

dingyifei:
https://dingyifei.github.io/testKlipperSite/
https://dingyifei.github.io/testKlipperSite/zh
https://dingyifei.github.io/testKlipperSite/zh-Hant
https://github.com/dingyifei/testKlipperSite/actions/runs/2009854025
https://github.com/dingyifei/test-build-klipper3d/actions/runs/2006985755
The script seems to be working now, and I have consolidated the commits into four, so it's clearer.

KevinOConnor:
Thanks for working on this.

As high-level feedback, I'm struggling to understand all the config and site transformations the scripts in this PR makes.  I feel I need to have a good understanding of how the website is deployed as it's something that will likely need regular maintenance as the project progresses.  Some questions/comments:
1. Why is the existing mkdocs.yml file renamed and split?  The split is making the PR harder to review.  If the split isn't necessary then probably best to look at config cleanup after this PR.
2. Why are the docs moved into an `en/` directory at the beginning of the scripts, and then moved back to the main `site/` directory at the end?  I'm concerned this moving of files around could break html links.
3. Why are there two loops in the build.sh script?  What's done in the second loop that's separate from the transformations in the first loop?
4. The fallback mechanism is complicating the PR and I don't think adds much.  If the site deployment fails, then the previous iteration of the site will still be available and I'll get an email about the error.  That seems like the correct behavior - maintaining two different deployment mechanisms seems like it could make things worse.

Thanks again,
-Kevin

dingyifei:
It basically builds multiple sites in different files and deploy the whole thing onto GitHub pages.

> Why is the existing mkdocs.yml file renamed and split? The split is making the PR harder to review. If the split isn't necessary then probably best to look at config cleanup after this PR.

It is possible to keep mkdocs.yml.  Initially I was thinking would be easier to build sites with different plugins and modifications with multiple ymls. It seems a bit redundant now.



> Why are the docs moved into an en/ directory at the beginning of the scripts, and then moved back to the main site/ directory at the end? I'm concerned this moving of files around could break html links.

I'll change this by making the English site build before the rest of the sites, indeed it is redundant ( build a site/ after site/<lang>/ removes site/<lang>/)

> Why are there two loops in the build.sh script? What's done in the second loop that's separate from the transformations in the first loop?

The second loop build the sites and the first loop handle the file moving. It is impossible to build site only in the first loop because the first loop appends all languages' relative directory to extra.yml so the alternate button on every language functions the same.

> The fallback mechanism is complicating the PR and I don't think adds much. If the site deployment fails, then the previous iteration of the site will still be available and I'll get an email about the error. That seems like the correct behavior - maintaining two different deployment mechanisms seems like it could make things worse.

Agree, it is always possible to revert changes that breaks the deployment.

dingyifei:
any update?

KevinOConnor:
Thanks.  This still has the copy to the `en/` sub-directory.  Is that needed?  I'm concerned it will break existing html links.

Separately, as a minor comment, using `#*#` seems like it causes some grief in the regex - why not use something like `#-#` or just tie to the known field names.

Otherwise, seems fine to me.  Due to scheduling on my side, I'll look to commit this early next week.

-Kevin

dingyifei:
> Thanks. This still has the copy to the en/ sub-directory. Is that needed? I'm concerned it will break existing html links.

I will check if the relative links are broken, based on mkdocs output, these links have correct file correspond to them.

> Separately, as a minor comment, using #*# seems like it causes some grief in the regex - why not use something like #-# or just tie to the known field names.

Thank you for pointing this out, indeed, tie them to the known field names makes it easier. I'll make these edits as soon as I get some time.

dingyifei:
> Separately, as a minor comment, using #*# seems like it causes some grief in the regex - why not use something like #-# or just tie to the known field names.

It seems like sed only read line by line and can't be used to match the field across multiple lines (had trouble with directly matching lang: en). I changed it to match #-# now.

Additionally, I have checked the links, the "../*" links became absolute links to the Klipper repository and is correct. The relative links pointing to files in img and prints shouldn't cause any issue since these folders are copied.

dingyifei:
Any updates?

dingyifei:
> Thanks for working on this.
> 
> 
> 
> I struggled to get this to build locally for testing.  I'm also not confident in the copying of the docs/ directory around (eg, it broke the loading of extra.css).
> 
> 
> 
> I made some changes to the scripts and added some documentation to the docs/_klipper3d/README file.  My updates are at https://github.com/KevinOConnor/klipper-dev/tree/work-translations-20220405 .  What do you think about that modified approach?  (Main script is at: https://github.com/KevinOConnor/klipper-dev/blob/work-translations-20220405/docs/_klipper3d/build-translations.sh )
> 
> 
> 
> -Kevin

It looks good to me. I think the sections in mkdocs.yml that are replaced using sed still need some kind of warning to prevent broken site build without warning (if sed replaced nothing, the script proceed to build) from accidental modification of the mkdocs.yml without changing the script.

KevinOConnor:
> I think the sections in mkdocs.yml that are replaced using sed still need some kind of warning to prevent broken site build without warning (if sed replaced nothing, the script proceeds to build) from accidental modification of the mkdocs.yml without changing the script.

I was able to update the script so that it isn't necessary to update the `site_dir` option (we can create a symbolic link in the script instead of updating the yml file).  That leaves only the "lang:" and "language:" options.  I don't think those options are likely to change without being noticed, but I'm open to alternatives.

-Kevin

KevinOConnor:
I tried just removing the explicit language settings from the mkdocs.yml file and everything still seems to work.  Maybe a simple answer is to not set these options?

-Kevin

dingyifei:
> I tried just removing the explicit language settings from the mkdocs.yml file and everything still seems to work.  Maybe a simple answer is to not set these options?
> 
> 
> 
> -Kevin

These explicit language settings change the behavior of search index and site language settings. They are default to en, but if they are not explicitly set, the search index won't be optimized for the language of the documentation and things like the text on search bar and buttons are not going to be the corresponding to the language. 

The problem with markdown-material is the search and the site language are often different. For many languages, the hack to have an "okay" search index is to use another language's language code on search's language setting.

KevinOConnor:
Okay.  Note that I get this during the build though:
`INFO     -  Option search.lang 'zh' is not supported, falling back to 'en'`

I separated out the language config updating to its own commit on that dev branch ( https://github.com/KevinOConnor/klipper-dev/commit/bb00ef7d13dc9649dc0ac63dd6e892e6d2977593 ).

Let me know how you would like to proceed - go forward with those three commits, go forward with just the first two commits (no language config setting for now), or take a different approach?

-Kevin

dingyifei:
> Okay.  Note that I get this during the build though:
> 
> `INFO     -  Option search.lang 'zh' is not supported, falling back to 'en'`
> 
> 
> 
> I separated out the language config updating to its own commit on that dev branch ( https://github.com/KevinOConnor/klipper-dev/commit/bb00ef7d13dc9649dc0ac63dd6e892e6d2977593 ).
> 
> 
> 
> Let me know how you would like to proceed - go forward with those three commits, go forward with just the first two commits (no language config setting for now), or take a different approach?
> 
> 
> 
> -Kevin



https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-search/?h=search#built-in-search-plugin

I think I misconfigured the search language in active_translations, while lunr(used for material MkDocs search) support zh, the current workaround for search language is to set it to Japanese. I'll change that when I have some time.

I would like to proceed with all 3 commits.

KevinOConnor:
Okay, thanks.  I merged in those commits and it looks like the klipper3d.org site is being generated as expected.

-Kevin

dingyifei:
I noticed scripts files are missing. I'm not sure why it's `....`, which is probably outside of `klipper/`, but the `scripts` folder might need to be copied to `"${WORK_DIR}lang/${langsite}"`.

```bash
WARNING  -  Documentation file 'FAQ.md' contains a link to '..../scripts/install-octopi.sh' which is not found in the documentation files.
WARNING  -  Documentation file 'FAQ.md' contains a link to '..../scripts/install-octopi.sh' which is not found in the documentation files.
WARNING  -  Documentation file 'FAQ.md' contains a link to '..../scripts/klipper-start.sh' which is not found in the documentation files.
```

dingyifei:
nvm, it seems like it's a translated string error. Everything is working

PR #5330, opened 2022-03-06
gmcd to gcmd

Ludy87:
change var gmcd to var gcmd


KevinOConnor:
Thanks for reporting.  Commits need to have a signed-off-by line for me to commit directly (see https://www.klipper3d.org/CONTRIBUTING.html ).  However, I did update the code to fix these bugs (commit a43973b0 and 98d3fffd).

-Kevin

PR #5333, opened 2022-03-07
display_status: stops M73 from setting progress to 0 if P missing

pedrolamas:
Following up on discussion in Discord, this PR is to ensure that an `M73` command with no `P` parameter does not set progress to 0 as it currently does.

This issue is easily reproducible in Prusa/SuperSlicer by enabling the "Firmware: Stealth mode" option together with "Firmware: Print remaining times" option:

image

After slicing, the g-code output will have 2 consecutive `M73` lines, but only the first line has a `P` param, and the second line will then set progress to 0.

image

Neither [Marlin](https://marlinfw.org/docs/gcode/M073.html) nor [RepRap](https://www.reprap.org/wiki/G-code#M73:_Set.2FGet_build_percentage) have any indication to assume `P0` if `P` is missing, so I believe this should be a safe fix to apply.

Issue was first reported here: https://github.com/fluidd-core/fluidd/issues/449

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,8 @@ All dates in this document are approximate.
 
 ## Changes
 
+20220307: `M73` will no longer set print progress to 0 if `P` is missing.
+
 20220304: There is no longer a default for the `extruder` parameter of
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #5334, opened 2022-03-08
Update printer-tronxy-x5sa-pro-2020.cfg

SlackHacky:


Fisheiyy (commented):
This is just so unnecessary, there is no need to make a whole PR to change 4 words.

KevinOConnor:
I don't have a preference on wording, but it would be necessary to lookup the original author of this config and ping them on changes like this.  Also, commits need to have a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html and lines should be wrapped to 80 columns ( https://www.klipper3d.org/Example_Configs.html ).

-Kevin

PR #5338, opened 2022-03-08
scripts: Migrate `check_whitespace.py` to python3

mdtusz:
Modified strings used in pattern matching to b-strings as are expected
with python3.

Signed-off-by: Miklos Tusz <miklos@aon3d.com>


KevinOConnor:
Thanks.

-Kevin

PR #5340, opened 2022-03-08
Add important step for consistent 100mm/s

hernandez87v:
docs: Add detail about minimum layer print time for required speed to be met.

Signed-off-by: Vlad Hernandez hernandez87v@gmail.com


KevinOConnor:
Thanks.  However, I don't think this change to the documentation is necessary.  As far as I know all the common slicers will default to a sufficient minimum layer time for the given object sliced using the given suggested parameters.  If a user has configured something unusual, then it's up to that user to diagnose their config - we can't reasonably list every slicer parameter that may be misconfigured.

-Kevin

PR #5341, opened 2022-03-09
stm32: set USB clock source to PLLQ for STM32G0B0

bigtreetech:
* The USB clock source of `STM32G0B0` have not `HSI48` but only `PLLQ`, So we set the USB clock source to `PLLQ`. In this way, both `STM32G0B0` and `STM32G0B1` can use the same `firmware.bin` file


KevinOConnor:
Thanks.  I committed this along with 4ce2d379 on top.

Separately, this sets USBSEL=2 which is in contrast to the reference manual https://www.st.com/resource/en/reference_manual/rm0444-stm32g0x1-advanced-armbased-32bit-mcus-stmicroelectronics.pdf which states USBSEL=1 should be used to enable PLLQCLK.  In the past I tried enabling the PLLQ for USB and could never get it to work - I guess this is why.  Do you have any further information on this discrepancy?

-Kevin

PR #5346, opened 2022-03-11
adjust to build tests nicely on mac

kpishere:
Addresses the one missing file on a mac in comparison to a linux machine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 out
 *.so
 *.pyc
+*.dSYM
 .config
 .config.old
 klippy/.version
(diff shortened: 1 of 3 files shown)


kpishere:
@KevinOConnor Any chance of this getting accepted? Makes further contributions easier for me. Pending one is ADS1x1x support where it can be used on bed and/or extruder thermocouples. I expect that one is of sufficient interest to people.

KevinOConnor:
Thanks.  However, I don't think these changes are a good idea right now.  There are many many possible build/test/run environments and the maintenance burden to support them would be too high for a small project like Klipper.  We instead look for the build/test/run environments to use commonly available Linux distros.  Many users can implement this on non-Linux systems by using a container or virtual machine.

-Kevin

PR #5360, opened 2022-03-17
Linear fan control for the temperature_fan

bhelm:
This adds a "linear" fan control to the temperature_fan. It scales the fan speed according to a linear temperature range. For example if the target_temp is 50° and the linear_offset is 10°, it will start at 40° to raise the fan speed from min_speed to max_speed on 50°.

Context: I have a very loud PSU fan that goes from 0 to 9100 RPM without any steps between. I wanted a better fan control, so i added a thermistor to the PSU's heatsink. While playing around with the temperature_fan settings, i realized that neither watermark nor pid fit for this use-case. Watermark works just like the stock fan behavior and would give no improvement. PID did also not seem like a good fit, as heat generated by the PSU is not constant. I have tried it, but the fan started oscillating in short intervals what is also annoying by itself. With the linear control, the fan slowly speeds up and down as the temperature on the PSU changes, giving a reasonable fan speed and is easily predictable on how it behaves.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2435,7 +2435,7 @@ watched component.
 
 ### [temperature_fan]
 
-Temperature-triggered cooling fans (one may define any number of
+Temperature-controlled cooling fans (one may define any number of
 sections with a "temperature_fan" prefix). A "temperature fan" is a
 fan that will be enabled whenever its associated sensor is above a set
 temperature. By default, a temperature_fan has a shutdown_speed equal
@@ -2457,14 +2457,20 @@ information.
 #tachometer_ppr:
 #tachometer_poll_interval:
 #   See the "fan" section for a description of the above parameters.
-#sensor_type:
-#sensor_pin:
 #control:
+#   In addition to "watermark" and "pid", temperature_fan has an additional
+#   "linear" control that sets the fan speed to a value between min_speed
+#   and max_speed, starting from a temperature offset below the target_temp.
+#linear_offset: 10.0
+#   on "linear" control, this sets the temperature offset below the
+#   target_temp where the linear scaling starts
 #pid_Kp:
 #pid_Ki:
 #pid_Kd:
 #pid_deriv_time:
 #max_delta:
+#sensor_type:
+#sensor_pin:
 #min_temp:
 #max_temp:
 #   See the "extruder" section for a description of the above parameters.
--- a/klippy/extras/temperature_fan.py
+++ b/klippy/extras/temperature_fan.py
@@ -35,7 +35,11 @@ class TemperatureFan:
             'target_temp', 40. if self.max_temp > 40. else self.max_temp,
             minval=self.min_temp, maxval=self.max_temp)
         self.target_temp = self.target_temp_conf
-        algos = {'watermark': ControlBangBang, 'pid': ControlPID}
+        algos = {
+            'watermark': ControlBangBang,
+            'pid': ControlPID,
+            'linear': ControlLinear
+        }
         algo = config.getchoice('control', algos)
         self.control = algo(self, config)
         self.next_speed_time = 0.
@@ -132,6 +136,31 @@ class ControlBangBang:
         else:
             self.temperature_fan.set_speed(read_time,
                                            self.temperature_fan.get_max_speed())
+
+######################################################################
+# Linear control algo
+# controls the fan speed based on a linear temperature range between
+# the fans min and max speeds
+######################################################################
+
+class ControlLinear:
+    def __init__(self, temperature_fan, config):
+        self.temperature_fan = temperature_fan
+        self.linear_offset = config.getfloat('linear_offset', 10, above=0.)
+        self.min_speed = self.temperature_fan.get_min_speed()
+        self.max_speed = self.temperature_fan.get_max_speed()
+    def temperature_callback(self, read_time, temp):
+        current_temp, target_temp = self.temperature_fan.get_temp(read_time)
+        temp_diff = target_temp - current_temp
+        # calculate linear speed range between 0.0 and 1.0
+        speed = 1 - (temp_diff / self.linear_offset)
+        # convert it to respect min/max speed
+        speed = self.min_speed + (speed * (self.max_speed - self.min_speed))
+        if speed > self.max_speed:
+            speed = self.max_speed
+        elif speed < self.min_speed:
+            speed = self.min_speed
+        self.temperature_fan.set_speed(read_time, speed)
 
 ######################################################################
 # Proportional Integral Derivative (PID) control algo


bhelm:
OK, here is my "Self-Review"

I have been using it since 3 Weeks without trouble. As it is a new function and does not change existing code (except from registrering it as an option), even if there were a defect, it would not cause anything else than the linear sacling function to go bad.
I actually wonder that it did not get added yet. We are building machines with PSU's, Stepper Controllers, Steppers and MCUs partially living inside cases that need ventilation and proper cooling. When fan noise is a factor, a linear control is a classic for these applications.
All Signed-Off, so i think yes
I read the doc and followed the other code i have seen.
Yes
It is only the linear control function and its documentation in a single commit

docs/Config_Reference.md
```diff
@@ -2457,14 +2457,20 @@ information.
...
 #   See the "fan" section for a description of the above parameters.
-#sensor_type:
-#sensor_pin:
 #control:
+#   In addition to "watermark" and "pid", temperature_fan has an additional
+#   "linear" control that sets the fan speed to a value between min_speed
+#   and max_speed, starting from a temperature offset below the target_temp.
+#linear_offset: 10.0
+#   on "linear" control, this sets the temperature offset below the
+#   target_temp where the linear scaling starts
```
dewi-ny-je, line 2466: Not clear to me from the explanation you gave.
Do you maybe mean
"this sets the temperature offset which causes the fan to spin at max or min speed, depending on the sign of the temperature difference"

klippy/extras/temperature_fan.py
```diff
@@ -133,6 +137,31 @@ def temperature_callback(self, read_time, temp):
...
+        self.linear_offset = config.getfloat('linear_offset', 10, above=0.)
+        self.min_speed = self.temperature_fan.get_min_speed()
+        self.max_speed = self.temperature_fan.get_max_speed()
+    def temperature_callback(self, read_time, temp):
+        current_temp, target_temp = self.temperature_fan.get_temp(read_time)
+        temp_diff = target_temp - current_temp
+        # calculate linear speed range between 0.0 and 1.0
+        speed = 1 - (temp_diff / self.linear_offset)
+        # convert it to respect min/max speed
+        speed = self.min_speed + (speed * (self.max_speed - self.min_speed))
```
dewi-ny-je, line 158: Test case: T=50, target=45, lin=10, min=0, max=100
speed= 1-(-5/10)= 1.5
speed= 0+1.5*(100-0)= 100 (capped)
So it goes 100% even if the target temp is lower than the current one.
Maybe the sign is wrong.

dewi-ny-je (commented):
Checked briefly for content and formulas, not for syntax or consistency with guidelines

docs/Config_Reference.md
dewi-ny-je: Or also, more textbook, less user friendly:
"this sets the temperature offset which causes the maximum compensation (the fan speed is set to one of the extremes)"

klippy/extras/temperature_fan.py
dewi-ny-je: I'm not sure about coding style, but you are reusing the variable name speed for two things which are different. Memory-wise it won't change anything since a new variable has to be allocated in memory, but legibility suffers.
Use "compensation =" for the first "speed =" (line 156)

dewi-ny-je: Also, with T = Ttarget the first "speed" is 1, which means the fan goes to full?

I think this should be rewritten along the principle:

speed = current speed + temp_diff/self.linear_offset * (self.max_speed - self.min_speed)

Which means: T = target temp -> no changes.
T > target temp -> speed decreased. How much it depends on the linear_offset.

I notice that this is becoming a proportional setup, with no D component and with a make-up I(ntegral) component, but as you coded, it also makes little sense because it works strange (to me! I may have misunderstood).

Sineos: Mybe I'm wrong but for me it reads like:
`temp_diff = target_temp - current_temp`
Test case: target=45, lin=10, min=0, max=100

At 35°C current temp:
temp_diff = 45-35=10
speed = 1-(10/10) = 0
--> Correct, since we want to start when the temperature is above 35

At 40°C
temp_diff = 45-40=5
speed = 1-(5/10) = 0.5
--> Correct, fan at half power

At 45°C
temp_diff = 45-45=0
speed = 1-(0/10) = 1
--> Correct, fan at full power

dewi-ny-je: My mistake, I read it again and I undestood.

But then the explanation is poorly written.
Please change it from "linear_offset" to "linear_starting_temp " and the explanation to something along the lines of:

```
#   In addition to "watermark" and "pid", temperature_fan has an additional
#   "linear" control that sets the fan speed to a value between min_speed
#   and max_speed, starting from linear_starting_temp and the target_temp.
#linear_starting_temperature: 30.0
#   on "linear" control, this sets the temperature at which the linear scaling starts
```

Why making things complicated when they can be straightforward? :)

Sineos: I agree, the term `linear_offset` is not self explanatory. Maybe it would be simpler not to provide a delta but an absolute starting temperature and called it `start_temperature`, `lower_threshold` or something along these lines.

bhelm: Thanks for reviewing!

@dewi-ny-je glad you got it correctly now :)

image
@Sineos One can change the fans target temp on runtime using "SET_TEMPERATURE_FAN_TARGET TEMPERATURE_FAN=psu TARGET=45" (or using fluidd for example). Having a absolute "linear_starting_temp" would mean that one would be able to change the target temp below the starting_temp. That case then would need to be handled one way or the other and may yield (for the users point of view) unexpected results that may require additional documentation. Changing it closer to the staring temp would mean a more narrow scaling range, which may also be undesired.
I can think of the use-case that someone likes his components to run as cool as possible (to increase lifetime) most of the time, but also wants to be able to make them more quiet (for example on night time) when required. Another case is that one wants to test different target temps during a print to find a good noise-temp trade-off.

As @dewi-ny-je had initial issues understanding what it was meant to do, the documentation may need improvement. As i am no native speaker, i appreciate any help on that. I have initially thought a while about the name, as i was not easy to find something self explanatory that was not relatively long. here are some thoughts:

1. linear_* should be the prefix to make clear that this option only relates to the linear scaling.
2. linear_offset (current)
3. linear_delta (i dont like the word delta, as it sounds a bit too scientific for what this option does)
4. linear_start_offset (a bit more verbose)
5. linear_scaling_range (that one could be better, as it defines over how many degrees the scaling takes place)
6. linear_scaling_amount (not sure on that)

I am open for ideas!

dewi-ny-je: I see your point on setting the delta, so maybe it's ok if you use "linear_scaling_range" but also the description I provided, which makes easier to understand what it does.

Concerning your use case, I thought that a fixed starting point made sense because making the fans quieter would mean just increasing the target temperature: why changing the (in my idea, fixed) starting one? at (for example) 35 °C the cpu simply needs no cooling, ever. It makes sense to me, as user, to fix a starting temperature when sor dure I don't need cooling, and change the upper extreme depending on the other conditions. But I guess few people will change it on the fly anyway... so the "linear_scaling_range" is fine to me.

KevinOConnor:
Thanks.  However, I'm not sure how this differs from the PID with Ki and Kd set to zero.  That is, if only Kp is set then the PID is a linear feedback system.

Separately, it does seem that many people would like better control of their fans.  If I get chance, I'll try to extend the recent `SET_LED_TEMPLATE` support ( https://www.klipper3d.org/G-Codes.html#set_led_template ) to something similar for output_pin and fan_generic config sections.  If one could assign the fan speed to a Jinja2 template it should make the Klipper fan support more flexible.

-Kevin

dewi-ny-je:
> Thanks. However, I'm not sure how this differs from the PID with Ki and Kd set to zero. That is, if only Kp is set then the PID is a linear feedback system.
> 
> Separately, it does seem that many people would like better control of their fans. If I get chance, I'll try to extend the recent `SET_LED_TEMPLATE` support ( https://www.klipper3d.org/G-Codes.html#set_led_template ) to something similar for output_pin and fan_generic config sections. If one could assign the fan speed to a Jinja2 template it should make the Klipper fan support more flexible.
> 
> -Kevin

Indeed, where it operates in reverse mode, therefore likely equivalent to a case where the target temperature is the base temperature ("target - offset") and the pid_Kp parameter has a sign opposite to the usual one (since usually the control increases below the target, and here the goal is to increase proportionally above the target).

However the Configuration page provides no units of measure for pid_Kp, therefore it's not obvious at all how to set it to obtain what it's wanted. It would be nice to be able to calculate its value without try and fail.

bhelm:
Thanks Kevin for your input. I have tried the pid control with these parameters and it works similar to the linear control. I still have no idea what these params do but with some trial & error i was able to get something that works.
```
pid_Kp: 20
pid_Ki: 0.000001 # cant be 0 or "float division by zero" error
pid_Kd: 0

control = pid```

So while there is a way to archive that, i think a dedicated function is more user-friendly and straight forward. Alternatively the pid control would need more documentation to make clear that it can and is intended to also be used as a linear cooling control.

I have no idea how a Jinja2 template can define a fan curve, but if that means that one could specify temperature -> fan speed pairs to make a custom fan curve, this would indeed be nice. Question is if this would make a linear control obsolete. It would if it comes with some basic fan curves that can used by adding one or two lines of config to a fresh klipper install.

**Comment by bhelm** [comment]
> And that would be equivalent to which "linear" settings? Have you found out what "20" means in practical terms? Like % change per degree difference? Or similar
I have not looked what 20 means. just tested until i found a value that had the same speed at the same temp as 10° linear_start_range.

I have done some tests by changing the target temp with these results:
Target Temp / Fanspeed / actual temp
45° 11% (33.7 currently)
42° 35% (33.5 currently)
40° 50% (33.3 currently)
35° 85% (33.0° currently)

That is pretty close.

**Comment by dewi-ny-je** [comment]
Interesting. I would have expected the linear component of the PID algorithm to be smaller when the error (target-actual temp) is smaller, therefore the last row should have the smallest fan speed. Probably the PID algorithm is modified or inverted compared to the PID algorithm for heaters.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">

</head>

<body link="#0563C1" vlink="#954F72">



target T | fan | actual T | linear alg. | P(ID) alg.
-- | -- | -- | -- | --
45 | 0,11 | 33,7 | -0,13 | 0,435
42 | 0,35 | 33,5 | 0,15 | 0,575
40 | 0,5 | 33,3 | 0,33 | 0,665
35 | 0,85 | 33 | 0,8 | 0,9



</body>

</html>

For the P(ID) algorithm column I used the formula:
1-(target - actual)/20

So I'm not sure how to interpret the value of pid_Kp

**Comment by dewi-ny-je** [comment]
I think the easiest way is to simply suggest 

pid_Kp = 255 / linear_offset

**Comment by KevinOConnor** [comment]
I fixed (hopefully) the pid_Ki=0 bug (commit 22297229).  I also added some documentation on the PID parameters (commit 5a409d5d).

For the temperature_fan, the PID is:
`fan_speed = fan_max_speed - (Kp*error + Ki*integral(error) - Kd*derivative(error)) / 255`
Where "error" is "target_temperature - measured_temperature", and a "fan_speed" of 1.0 (or greater) is full on.

So, if pid_Ki and pid_Kd are set to zero then the result is `fan_speed = fan_max_speed - Kp*error/255`

The PID code in temperature_fan.py was written by @wizhippo .

> I have no idea how a Jinja2 template can define a fan curve,

If you configure something like the following:
```
[led my_fan_hack]
red_pin: PA1 # Set to your fan pin

[display_template fan_temp]
text: { (printer.heater_bed.temperature - 50.0) / 10.0 }
```
and activate it with:
```
SET_LED_TEMPLATE LED=my_fan_hack TEMPLATE=fan_temp
```
then you should observe the fan changing speed as the bed temperature changes.

Obviously using an led definition to control a fan is a bit quirky, so I have been thinking about adding explicit support for templates to fan_generic (and output_pin) objects.

-Kevin

bhelm:
@dewi-ny-je i forgot to tell you that my min_speed is 0.1 which has an influence how the linear scaling works.

> I fixed (hopefully) the pid_Ki=0 bug (commit [2229722](https://github.com/Klipper3d/klipper/commit/22297229b7be6e2fd9d6496c887fff2bfc8ca032)). I also added some documentation on the PID parameters (commit [5a409d5](https://github.com/Klipper3d/klipper/commit/5a409d5d3112d43da6e2bdb90cd3dbc62e491186)).
>
yes, the documentation is much better now.
 
I think a dedicated linear scaling function is (from the users perspective) much easier to understand and use than a formula for the more complex pid control that yields a not obvious value.

If there is a consensus that the linear scaling function is a good thing to add, i will rename the parameter and improve the documentation with the suggestions.

> > I have no idea how a Jinja2 template can define a fan curve,
> 
> If you configure something like the following:
> 
> ```
> [led my_fan_hack]
> red_pin: PA1 # Set to your fan pin
> 
> [display_template fan_temp]
> text: { (printer.heater_bed.temperature - 50.0) / 10.0 }
> ```
> 
> and activate it with:
> 
> ```
> SET_LED_TEMPLATE LED=my_fan_hack TEMPLATE=fan_temp
> ```
> 
> then you should observe the fan changing speed as the bed temperature changes.
> 
Ok i understand. Using jinja2 templating for a custom fan function is a nice idea.

If i think of a even better fan control, i think of a fan curve defined by "temperature -> fanspeed" pairs that get interpolated in between. I did not do that because the added complexity and the linear control is sufficient for most cases i can think of.

Let me know if you think i should update the PR with better documentation or not.

dewi-ny-je:
> I think a dedicated linear scaling function is (from the users perspective) much easier to understand and use than a formula for the more complex pid control that yields a not obvious value.

I see your point, but you agree that it is a duplication of features, right? only ease of use would change, and likely for a small number of users: not only because probably few are interested in a linear control, but also because obtaining the pid_Kp value from the formula now available in the docs is kind of high school level math :)

Since nevertheless ease of use may be an advantage, why not writing a macro which takes the input you proposed and outputs the pid_Kp value? the user would then put it into the config. I'm not an expert, but I think you could even be able to write the config file directly from a macro (at least, I know that it's possible to ready any setting from any block). 

No duplication of code, no need for an approval by Kevin, and so on.

bhelm:
I think ease of use and a clear statement what a feature does is the main priority. That pid control can do linear scaling is not that obvious and using it requires some calculation, making it more complex than it needs to be. While you can do all kind of stuff with macros, i think this would be a workaround for a workaround.

If code duplication is the main concern, i would change the linear scaling to use the pid control in the background - basically wrapping it and doing the calculation for the user. The possible downside of this is that if the pid control function gets changed, it could have side effects for the linear scaling. I think that comes down to personal preference of the developer/maintainer. 

So let me know if wrapping the pid function in the background would be the preferred way to do it, then i will Update the PR.

dewi-ny-je:
Thanks for your feedback.
I was only raising the point so that it could be discussed further, now it's up to Kevin to decide, I think the picture is quite complete.

jakep82:
From the contributing document:

`New modules, new options, and new parameters should not provide similar functionality to existing modules - if the differences are arbitrary than it's preferable to utilize the existing system or refactor the existing code.`

If the desired functionality is already present in Klipper, it seems that updating documentation to explain it might be a better path than adding duplicate functionality.

KevinOConnor:
I think the bot may have been too aggressive here.  I'm not against changing the documentation (or some other change).  However, I'm not sure what is being proposed, and I'm not sure a change is required.

Cheers,
-Kevin

bhelm:
To conclude:
- This PR proposed a linear scaling function that has two parameters (target and offset, where it scales from min_speed (at target-offset) to max_speed (target). The linear_offset parameter should be renamed to linear_scaling_range and the documentation improved if this function should be merged.
- It is possible to archive a similar linear scaling functionality using pid_control, but it requires more words to explain and a calculation the user needs to do to set pid value correctly. so it is less obvious that pid control can do linear scaling and less user friendly.
- I offered that i could change the linear scaling function to use pid control in the background if reducing lines of code is a priority over simplicity.

While there seems to be a consensus that linear scaling is useful and should be more accessible (either by providing more documentation for the pid control or by a dedicated function), its up to the maintainer to decide which route is preferred. Its also a question of principle whenever a single complex function that does everything is preferred over multiple simpler functions that focus on specific tasks (pid control could also do watermark if i guess correctly). I am on the [KISS](https://en.wikipedia.org/wiki/KISS_principle) side on this matter.

I try to help/improve and i am also fine if the result is that this pr is not required.

Greetings :)

dewi-ny-je:
> I think the bot may have been too aggressive here. I'm not against changing the documentation (or some other change). However, I'm not sure what is being proposed, and I'm not sure a change is required.
> 
> Cheers, -Kevin

Kevin, I think the bot should not close pulls tagged "reviewer needed" at all... 
Provided that reviewer needed is assigned only to pulls which have been in principle approved for merge.

Pull -> pending feedback or discussion going on -> positive outcome in your opinion -> reviewer needed (and no more automatic closing) -> any amount of time it's needed -> merge.

PR #5361, opened 2022-03-18
Config:  Compile settings and config for Kingroon KP5L

candre23:


Fisheiyy (changes requested):
Remove all the macros and the commented out stuff at the bottom of the config, also clear up some of the comments and remove the virtual_sdcard section at the top of the config, also remove all of the BLTouch stuff unless a BLTouch comes stock on the printer.

Arksine:
To add to the review, it is necessary for config submisisons to follow the [guidelines outlined here](https://www.klipper3d.org/Example_Configs.html).  Also, all commits should adhere to the format in the [contributing guidelines](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages).

PR #5365, opened 2022-03-19
Updated and created new TwoTrees configs to avoid confusion

laurienzu:
I updated existing configs and created a new one to avoid confusion. The printer is now sold with a different name and there were some errors in existing configs. This should help the community to avoid problems and asking :)

Signed-off-by:  Lorenzo Cascone <laurienzudesign@gmail.com>


laurienzu:
I think the build fails for the new filename but I don't really know why

KevinOConnor:
Thanks.  In general the changes look fine to me.  I'd like confirmation from @thucar and @pranas .

Also, all three configs add a `[heater_fan]` config section that I think should read as:
```
[heater_fan extruder]
pin: PB0
```
instead of the large blocks that are currently added.  (Otherwise it seems contrary to https://www.klipper3d.org/Example_Configs.html points 7f and 9.)

-Kevin

laurienzu:
> Thanks. In general the changes look fine to me. I'd like confirmation from @thucar and @pranas .
> 
> Also, all three configs add a `[heater_fan]` config section that I think should read as:
> 
> ```
> [heater_fan extruder]
> pin: PB0
> ```
> 
> instead of the large blocks that are currently added. (Otherwise it seems contrary to https://www.klipper3d.org/Example_Configs.html points 7f and 9.)
> 
> -Kevin

Thanks to you for reviewing this, fixed [heater_fan] config section and I hope it will be merged soon 😊

PR #5367, opened 2022-03-20
docs: Fix config snippet for polar

charlespick:
Previously the section for stepper_bed was intertwined with printer. This commit makes it match (../config/example-polar.cfg).

Signed-off-by: Charles Pickering <me@charlespick.xyz>


KevinOConnor:
Thanks.

-Kevin

PR #5372, opened 2022-03-23
Allow inversion of CS pin for SPI busses

mhier:
This is necessary for implementing support of the DRV8711 stepper driver (separate PR will follow after this one has been merged).

### DIFF
--- a/klippy/extras/bus.py
+++ b/klippy/extras/bus.py
@@ -39,7 +39,9 @@ def resolve_bus_name(mcu, param, bus):
 
 # Helper code for working with devices connected to an MCU via an SPI bus
 class MCU_SPI:
-    def __init__(self, mcu, bus, pin, mode, speed, sw_pins=None):
+    def __init__(self, mcu, bus, pin, mode, speed, sw_pins=None,
+                 invert_cs=False):
+
         self.mcu = mcu
         self.bus = bus
         # Config SPI object (set all CS pins high before spi_set_bus commands)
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I think it is fine to add mcu support for inverting the cs pin.  However, I don't think this should be a user-facing option.  (That is, it should remain invalid to specify something like `cs_pin: !PA8`.)  An inverted cs pin is an attribute of the hardware that the developer will know and should define when using the `bus.SPI` system.

-Kevin

KevinOConnor:
As an example, see how the spi mode is configued by the python module code instead of via user config parameters.

-Kevin

mhier:
I changed it now. Not sure if/where this can/should be documented properly somehow though...

KevinOConnor:
Thanks.  In general, it looks fine to me.

As a minor comment, sometimes the new flag uses the name `invert_cs`, sometimes `invert_pin`, and sometimes `!pin_active_value`.  I think the code would be more clear if we used the same name (I'd say `invert_cs`) in all the code.

Because this change will require an mcu flash for many users, it's something I'll look to hold off for to see if we can batch it with other mcu changes.  So, expect a couple of weeks before merging.

-Kevin

mhier:
I changed the name to `invert_cs` always and rebased to the latest master.

KevinOConnor:
Thanks.  I committed this, but renamed the field to `cs_active_high` (commit 77937846).

-Kevin

KevinOConnor:
Unfortunately, this change broke the spi_flash tool.  I can update the spi_flash tool to use the new message format, but then that will prevent spi_flash from being able to flash older version of Klipper to the latest version.

@Arksine I know you had a mechanism to account for this issue the last time it came up.  What's your recommendation for this?  Should we add support for both forms of `config_spi`?

-Kevin

Arksine:
Yes, I think adding support for the new protocol in addition to the current one will work.  I'm about to take a look at it, I think I should be able to get a fix for it shortly.

PR #5375, opened 2022-03-25
HTU21D - SI7021 is has id 0x15

SirUli:
This error was most probably caused by a copy/paste error. All vendor documentation for SI7021 (e.g. https://www.silabs.com/documents/public/data-sheets/Si7021-A20.pdf) points to 21 and it actually only works when adapted to 0x15 (=21).

This PR was tested in real production on a Raspberry Pico MCU.

### DIFF
--- a/klippy/extras/htu21d.py
+++ b/klippy/extras/htu21d.py
@@ -14,7 +14,7 @@ from . import bus
 #       HTU21D - Tested on Linux MCU.
 #       Si7013 - Untested
 #       Si7020 - Untested
-#       Si7021 - Untested
+#       Si7021 - Tested on Pico MCU
 #       SHT21  - Untested
 #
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Maybe @dianlight could comment on this?

-Kevin

dianlight:
it is ok. 

L.

KevinOConnor:
Okay, thanks.

The submission would need to have a signed-off-by containing your full real name though ( https://www.klipper3d.org/CONTRIBUTING.html ).

-Kevin

SirUli:
@KevinOConnor sorry touch me a couple attempts to do it properly. I hope i now followed the conventions as needed.

KevinOConnor:
Thanks.

-Kevin

PR #5376, opened 2022-03-25

shubukin:
Change rotation_distance from32 to 40 what is on tho printers right.


laurienzu:
this was modified on Commit [4c8d1b8] so it can be closed I think

PR #5380, opened 2022-03-26
printer-biqu-b1-se-plus.cfg

maxcalavera81:
Biqu B1 SE Plus Klipper printer pins configuration


Sineos:
Please also check against https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review
Especially No 3 the signed off by line.

PR #5385, opened 2022-03-28

MyMakibox:
Matplotlib python library is required for graph generation as decribed later in the document.

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -67,9 +67,9 @@ Note that resonance measurements and shaper auto-calibration require additional
 software dependencies not installed by default. First, you will have to run on
 your Raspberry Pi the following command:
 ```
-~/klippy-env/bin/pip install -v numpy
+~/klippy-env/bin/pip install -v numpy matplotlib
 ```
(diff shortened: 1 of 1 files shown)


dmbutyugin:
Yes, the instructions appear to be correct as they are today. matplotlib is not needed by the code that runs inside Klipper.

MyMakibox:
Following instructions on the page, the user generates an error because matplotlib is not installed in the virtual environment.. That seems undesirable. An alternative fix would be to change instructions so that the user is in a different virtual environment with appropriate dependencies when generating graphs. That alternative fix seems more complex.

Sineos:
> Following instructions on the page, the user generates an error because matplotlib is not installed in the virtual environment.

Ref: https://www.klipper3d.org/Measuring_Resonances.html#software-installation

The instructions are correct:
* For the functionality **within** Klipper only numpy is needed. In Klipper (using from a Web-UI, e.g. Mainsail, Octoprint etc) no graphs are generated
* For the advanced functionality **outside** of Klipper, matplotlib is needed and the instructions are calling for it. This functionality is only called from a Linux shell

PR #5390, opened 2022-03-31
Fix incorrect default value for max_freq in [resonance_tester] documentation

cbc02009:
module: Documentation

According to [resonance_tester.py](https://github.com/Klipper3d/klipper/blob/master/klippy/extras/resonance_tester.py#L54) the default value for max_freq is actually 133.3333... Hz and not 120 like the docs say. Updated the docs to match the code.

Signed-off-by: Christopher Conroy <cbc02009@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1545,8 +1545,8 @@ section of the measuring resonances guide for more information on
 #   for more details on using this feature.
 #min_freq: 5
 #   Minimum frequency to test for resonances. The default is 5 Hz.
-#max_freq: 120
-#   Maximum frequency to test for resonances. The default is 120 Hz.
+#max_freq: 133.33
+#   Maximum frequency to test for resonances. The default is 133.33 Hz.
 #accel_per_hz: 75
 #   This parameter is used to determine which acceleration to use to
 #   test a specific frequency: accel = accel_per_hz * freq. Higher the


dewi-ny-je:
Setting a max value to a fractional frequency, when the testing is performed at integer steps, makes little sense. I think the code should be updated to the docs, in this case, or in any case both updated to an integer value

cbc02009:
I completely agree that it shouldn't be a fraction, but since I wasn't sure what the logic behind the original choice was, I figured it was safest to update the docs instead. I'd be happy to change the default value instead if that's what is preferred.

bernhardberger:
I just noticed that too. There definitely is a discrepancy there.

KevinOConnor:
Looks good to me.  @dmbutyugin - fyi.

-Kevin

dmbutyugin:
@KevinOConnor, thank you, yes, the change looks good to me too.

@dewi-ny-je FWIW, the testing is not performed in integer steps - the frequency changes smoothly from min to max value. It is only reported in integer steps in the console to not overwhelm the user with myriads of messages.

dmbutyugin (approved):

KevinOConnor:
Thanks.  I committed this (unfortunately, hitting the wrong button on github, but it still merged okay).

-Kevin

PR #5396, opened 2022-04-03
Add config for CR-10 Smart Pro, including installation instructions

iblue:
The file is pretty self-explanatory. 

Copyright: The file is a derivative work of a reddit thread. Permission to create a pull request and therefore acceptance of GPL terms is given implicitly by the authors. See https://www.reddit.com/r/klippers/comments/tr4jie/comment/i3012vj/?utm_source=reddit&utm_medium=web2x&context=3


manu7irl (changes requested):
Hi by going over very quickly, there is too much non related stuff, keep it bare bone. 
The macros should be removed, and some minor comments on the extruder section:
max_extrude_only_distance should be way less as it is the sprite pro extruder on it, which is direct drive.
and if you could use gear_ratio to reflect the real rotation_distance based on that.
Also remove presure_advance settings.

All in all it looks good to me.

iblue:
Thanks for the feedback. Will look into it in the next days.

jakep82:
Guidelines for example configs are [here](https://www.klipper3d.org/Example_Configs.html).

In addition to what manu7irl mentioned, you also need to add this to printers.test for regression testing, remove the [include] and [bed_mesh] sections, and the square_corner_velocity should be left at the default value of 5.  Using a SQV value of 1 will lead to poor results when printing curves.

It's also unlikely to be accepted with instructions at the top for installing Klipperwrt.  If a user wants to use that software, they can seek out the current install instructions in the Klipperwrt documentation.  It's not something we would want to maintain here.

iblue:
So, anymore feedback? Or can this be merged?

config/printer-creality-cr10-smart-pro-2022.cfg
```diff
@@ -0,0 +1,162 @@
...
+microsteps: 16
+endstop_pin: probe:z_virtual_endstop
+position_min: -1.5
+position_max: 400
+homing_speed: 4
+second_homing_speed: 1
+homing_retract_dist: 2.0
+
+[extruder]
+max_extrude_only_distance: 1000.0
```
Sineos, line 57: Should be removed

Sineos (changes requested):

config/printer-creality-cr10-smart-pro-2022.cfg
iblue: Aye. Done.

Sineos (approved):

KevinOConnor:
I committed this.

Separately, it's preferable to specify a gear_ratio for extruders that use a gearing mechanism, as then the extruder rotation_distance will be the circumference of the wheel that pushes the filament.

-Kevin

iblue:
Just for reference, after nagging Creality for a month, they released the source. Official pinout can be found here: https://github.com/CrealityOfficial/CR-10-Smart-Pro/blob/main/Marlin/src/pins/stm32f1/pins_CREALITY_V4.h

PR #5398, opened 2022-04-05
wrong pin for stepper_z in set gpio pins at boot

bkw777:
set gpio pins at mcu start up: There is no reason to set PB5 high at mcu startup, and stepper_z enable_pin PB11 is missing from the list.


bkw777:
I have a v2 (e3d) if that makes a difference, and later in the same config file the enable pin is defined and that part is correct.

I could supply a little more evidense if you want like meter readings or something.

KevinOConnor:
Okay, thanks - commit a10b2805.

-Kevin

PR #5402, opened 2022-04-09
docs: Add link to linux mcu neopixels config docs

agentk:
The reference to the linux mcu was helpful to understand what a linux mcu was but didn't help with understanding the technical reasons why it's not supported. I hope people will find linking to the closed issue provides more context.

Signed-off-by: Karl Bowden <karl@bearded.sexy>


docs/Config_Reference.md
```diff
@@ -2548,7 +2548,7 @@ sections with a "neopixel" prefix). See the
 [command reference](G-Codes.md#led) for more information.
 
 Note that the [linux mcu](RPi_microcontroller.md) implementation does
-not currently support directly connected neopixels.
+not currently support directly connected neopixels. See closed (issue #3366)[https://github.com/Klipper3d/klipper/issues/3366] for technical details.
```
Sineos, line 2551: It is preferable not to link to an issue from the documentation. Instead add something along the lines

> The used Linux kernel GPIO interface is not fast enough to provide the required pulse rates.

Sineos (changes requested):

docs/Config_Reference.md
agentk: I had noticed a lack of references to issues in the docs. In this specific case, the nuance of also seeing an issue has been opened, discussed and closed if hard to convey without linking to the issue.

Sineos: Well, I do not see any more need: The current design by using the kernel interface does not allow this scenario

agentk: That’s a very clear way of phrasing it. Thank you @Sineos. I’ll update the PR.

Sineos (approved):

KevinOConnor:
Thanks.

-Kevin

PR #5404, opened 2022-04-10
Kconfig: Add STM32F072 16KiB bootloader option

alstoepp:


Arksine:
Hi.  If I may ask, which bootloaders are available that require the 8 and 16 KiB offsets for the F072?  If they are open source bootloaders I think it would be useful to add them to the [documentaton](https://www.klipper3d.org/Bootloaders.html).  Thanks.

alstoepp:
For 8KiB [this](https://github.com/Arksine/CanBoot) bootloader could be adaptet for the STM32F072. For 16KiB [this](https://github.com/Telekatz/MSC-stm32f103-bootloader) bootloader works.

Arksine:
Thanks.  The first bootloader linked is one I wrote.  It was more proof of concept and it won't work with a recent version of Klipper.  That said, I do have plans to bring it up to date and once that is complete I will make the appropriate PRs to patch Klipper.  I don't think its a good idea to add the 8 KiB offset until its ready, as it implies that there is an 8 KiB bootloader available for the F072 now.

As for the second, that looks interesting.  I presume you are using it?  It would be a good idea to add documentation for it in [bootloaders.md](https://github.com/Klipper3d/klipper/blob/master/docs/Bootloaders.md).  Basically just follow the pattern shown for other bootloaders, with a description of the bootloader, a brief how-to on flashing it, with any appropriate links necesary.

Arksine (approved):
Looks good to me.

KevinOConnor:
Thanks.

-Kevin

PR #5405, opened 2022-04-10
Better EZBoard V1.x Config

houseofbugs:
Added in examples for EZNeo, EZABL, Added in missing settings for controller fan (FAN_2). Overall better baseline for these boards for users.


houseofbugs:
This was tested on an Ender 3 Pro with all the upgrades installed.

houseofbugs:
> If you're changing the file name, you also need to update test/klippy/printers.test so the regression test doesn't fail.
> 
> I can't comment on the changes you made, but I do know that the `[output_pin beeper]` section needs to be commented out or removed. You also need to sign off with your real name and email. See [here](https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages.)

Thanks. I updated that as well. But I am not sharing my personal email due to privacy reasons. Name is Timothy Hoogland, I own TH3D Studio. They can use our company support email Support@TH3DStudio.com

KevinOConnor:
Thanks.  You can use a support email address in your signed-off-by line if that is a valid contact address.  However, we'll still need an explicit signed-off-by line as it indicates you agree to the terms at https://github.com/Klipper3d/klipper/blob/master/docs/developer-certificate-of-origin as described at https://www.klipper3d.org/CONTRIBUTING.html .

Separately, adding (or recommending) a max_extrude_cross_section is against our guidelines ( https://www.klipper3d.org/Example_Configs.html point 9).  Also, I don't understand the beeper and display comments - the intent of the generic config files is to describe the board - not the possible peripherals attached to the board.  (Point 6a in link above.)  So, unless the board can't possibly handle any type of display other than an st7920, I don't think it makes sense to add those comments.

-Kevin

houseofbugs:
The support email is a valid address and I have access to it.

The board is intended to be used with only the st7920. It's not marketed or recommended for any other types. The only other possible one would be one of the "emulated" types but those are very rare to be used with this. The goal was to have a full config that has all the features laid out like the controller fan (this is usually used on all setups but also does not hurt anything if defined and nothing is connected).

I was just trying to make a better baseline for EZBoard users but it seems you don't want to have commented out items (aside from a BLTouch) to make setting up my board easier for people with your firmware. I'll just stick to maintaining these files separately on [our github repo for EZBoard Klipper stuff](https://github.com/th3dstudio/ezboard-klipper-stuff) here since this looks like it's going to be an issue because of the rules you have but its your repo so I am not going to try and argue with you.

I was simply trying to make a better config to help people get running with the EZBoard quicker so they don't have to waste as much time trying to guess things that need to be setup for most printer builds. Feel free to make whatever changes you want so this is compliant with your rules, otherwise just close it out. I am not spending any more time on it since we're going to maintain our own repo with examples there.

emersonsc:
As a "new" user, I can't for the life of me see why you have a problem with the manufacturer providing a thorough config file. I'm fact, what I have found to be a common theme with Klipper config files is they are horribly incomplete, and typically requires one having to pull information off of several config files or finding personal config files to make a complete one for a stock printer.

I can see the need for rule 9, but your pointing out the issue with the screen config under rule 6a is so backwards. Again this is the manufacturer of the board who knows what the board is intended to be used with so he wants his users to have a as-complete-as-possible config for those who want to migrate over to this software. Unfortunately it seems you don't want folks to have a better and easier experience with your software. 80% of my config came from help from others in the community because your documention is confusing and your config files lack. How about be a bit more open to ideas that aren't yours and help this community move forward and progress instead of turning it into a group of elitist snobs. Just my 2 cents.

houseofbugs:
Clearly @KevinOConnor is more concerned with his arbitrary "rules" than allowing a manufacturer to provide a config for their own damn product for this overly complex firmware project. I'm closing this out and will continue to recommend Marlin to people.

KevinOConnor:
I appreciate the feedback Tim.  I'm a bit puzzled by your latest messages however.  Earlier in the thread you clearly indicated that you did not want to spend further time on this, yet you are continuing to invest time on it.  I understand if you don't want to invest time; I too place a high-value on my time.  As a surprise to no one, you are unlikely to find people willing to engage after declaring that you are unwilling to engage.

It's unfortunate that you prefer another firmware, but I understand.  I recommend people use the firmware they are most comfortable with.  We aren't selling anything here.  For what it is worth, it seems to me that you had already made your choice of firmware before you even opened this PR.

Best of luck,
-Kevin

houseofbugs:
> It's unfortunate that you prefer another firmware, but I understand. I recommend people use the firmware they are most comfortable with. We aren't selling anything here. For what it is worth, it seems to me that you had already made your choice of firmware before you even opened this PR.
> 
> Best of luck, -Kevin

I prefer firmware that is not run by a bunch of elitists like yourselves.

PR #5406, opened 2022-04-10
temperature_fan: clarification about PID

dewi-ny-je:
Clarified that the temperature_fan operated in reverse compared to a heater.


dewi-ny-je:
Thanks and sorry for missing the signed-off part.

PR #5412, opened 2022-04-13
Virtual SD Card: Use python file buffering

kageurufu:
virtual_sd_card.py used it's own buffering, reading blocks then manually
 splitting lines to read files line by line.

While this would be normal in a direct IO environment, python has it's
 own IO buffering by default.

See the `buffering` argument to `open()` in the python docs for details.
https://docs.python.org/2/library/functions.html#open
https://docs.python.org/3/library/functions.html#open

The second problem this solves is handling unicode gcode files, which
 currently error during pause/resume, as python's `len()` returns
 character length for strings, not byte length. In python2, strings are
 bytes anyway, so this doesn't matter, and Klipper on python2 fails to
 read UTF-8 encoded files. Under python3, this means the calculated
 file_position based on read characters is wrong, and on resuming, will
 seek earlier positions in the file.

By using python's internal file buffering, we can simply call
 `readline()` to get the next complete line. `readline()` includes line
 endings, which isn't a problem for Klipper, and returns an empty string
 to signify the end of the file. This also removes the error state where
 the `next_file_position` gets out of sync and Klipper had to previously
 discard it's buffer.

I've monitored IO load using this, and saw consistent IO during printing
 with and without the changes. I've also tested pause/resume (using a
 M600 filament swapped multi-material print) and it no longer errors.

Signed-off-by: Franklyn Tackitt <franklyn@tackitt.net>

### DIFF
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -221,39 +221,28 @@ class VirtualSD:
             return self.reactor.NEVER
         self.print_stats.note_start()
         gcode_mutex = self.gcode.get_mutex()
-        partial_input = ""
-        lines = []
         error_message = None
         while not self.must_pause_work:
-            if not lines:
-                # Read more data
-                try:
-                    data = self.current_file.read(8192)
-                except:
-                    logging.exception("virtual_sdcard read")
-                    break
-                if not data:
-                    # End of file
-                    self.current_file.close()
-                    self.current_file = None
-                    logging.info("Finished SD card print")
-                    self.gcode.respond_raw("Done printing file")
-                    break
-                lines = data.split('\n')
-                lines[0] = partial_input + lines[0]
-                partial_input = lines.pop()
-                lines.reverse()
-                self.reactor.pause(self.reactor.NOW)
-                continue
+            try:
+                line = self.current_file.readline()
+            except:
+                logging.exception("virtual_sdcard read")
+
+            self.next_file_position = self.current_file.tell()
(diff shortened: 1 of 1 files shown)


garethky:
LGTM!

Nitpick: The docs says that if no buffering option is specified the system default is used. From what I read the system default is 8K and the code currently keeps an 8K buffer. Should we force it to keep an 8K buffer like the existing code? Maybe we want to use system default or 8K, whichever is larger?

(‘ io.DEFAULT_BUFFER_SIZE’ should be 8K but maybe there is some weird distro that makes it smaller?)

kageurufu:
@garethky That's a good point, although it uses the file's block size by default, not io.DEFAULT_BUFFER_SIZE, which is 4096 for most filesystems. 8K (or a calculated multiple of st_blksize similar to 8K) would be fine too.

KevinOConnor:
Thanks.  In general, it seems fine to use Python's buffering instead of manual buffering.  However, this code makes some subtle behavior changes that I'm not sure about:
1. It now calls `pause()` on every command instead of every read.
2. It seems to break `set_file_position()` which could be called from a g-code command within the file being read.

It might be simpler to just fix the python3 compatibility issue.

-Kevin

kageurufu:
I still like the idea of this PR, I'm working on fixing the `pause()` timing by checking the buffer's stream for changes, and i'll investigate fixing set_file_position(). It should be simple to just have that use `seek()`, I thought I had it working though (pause/resume works, for example)

PR #5414, opened 2022-04-13
Add SET_LED_HSI, providing HSI colorspace support

kageurufu:
This adds `SET_LED_HSI LED=<config_name> HUE=<value> SATURATION=<value> INTENSITY=<value> [INDEX=<index>] [TRANSMIT=0] [SYNC=1]` allowing users to set colors using the HSI colorspace. When using RGBW LEDs, the White channel will be used to increase brightness, giving brighter cleaner colors, instead of using green to desaturate and brighten a purple light.

This is in python since the HSI to RGBW code uses math functions that are not available in the jinja environment.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -714,6 +714,14 @@ timeout. If careful timing is not needed, the optional SYNC=0
 parameter can be specified to apply the changes without resetting the
 idle timeout.
 
+#### SET_LED_HSI
+`SET_LED_HSI LED=<config_name> HUE=<value> SATURATION=<value>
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I'm a little confused on how this command would work with common LEDs that don't have a white led component?

Also, this change would need to be rebased to the latest code, and it should not make whitespace changes to the import lines.  Finally, I would think a different color scheme would be more appropriate as parameters to the existing SET_LED command instead of a new command.

-Kevin

kageurufu:
I've been talking with another klipper user about the interface and conversion logic for this, including niceities like taking the LED's white color temperature as a factor.

PR #5417, opened 2022-04-14
Added layer_height to the pressure advance formula

SvenHinz:
In the TUNING TOWER the pressure advance setting is changing every layer, but it was calculated like it's changing every 1 mm.


theophile:
This is not correct. The tuning tower adjusts the PA level at each layer change to ensure that it's scaling at the specified increment per mm.

SvenHinz:
OK, now I understand.

PR #5420, opened 2022-04-14
stm32: add support for CanBoot bootloader

Arksine:
This pull request adds support for the [CanBoot](https://github.com/Arksine/CanBoot) bootloader on STM32F103, STM32F042, and STM32F072 devices.  Specifically, this implements a new CAN admin command that may be used to request the bootloader.  Included is a script that can be used to upload software and query nodes running the bootloader for UUIDs.  The script requires Python 3.7+, however it has no dependencies so it should run without an issue on Debian 10+.  If necessary I can rewrite it to run on Python 2.7.  

I have tested on the F103 and F072, it should work on the F042 without issue however it would be good if someone can test it.

Those wishing to test CanBoot for the F103 should be aware that it is important to select the correct flash page size in the menu config.  On models with 64 KiB or 128 KiB of flash this should be 1 KiB. Models with 256 KiB of flash have a page size of 2KiB.  


smackoz:
This will be very helpful.

eddietheengineer:
Hello Arksine, I have tested this with a Huvud (F103) and a Mellow SHT36 (F072) at 250k baud. After flashing the new bootloader with an STlink, I was able to flash over CAN using Klipper! 

I am using a Waveshare CAN hat with 250k data rate, untwisted wires, and no termination resistor on the toolhead board side--the Mellow one failed a time or two flashing over CAN before it succeeded but it eventually worked, probably because of my "improper" setup and Waveshare hat.

Thanks for your work on this.

eddietheengineer:
Commenting here to document what we talked about, it looks like Huvud/etc boards without a reset button need to have a full erase with STMCubeProgrammer before flashing the CAN bootloader. I've flashed all of my CAN devices now (2 Huvud 0.61, 5 Huvud 0.5, 1 Mellow SHT36) successfully using this method!

raymondh2:
Tested a MCP2515, MCP2518FD, and a USB Canable adapter all at 500kbit/s and was able to flash with all of them

smackoz:
> Commenting here to document what we talked about, it looks like Huvud/etc boards without a reset button need to have a full erase with STMCubeProgrammer before flashing the CAN bootloader. I've flashed all of my CAN devices now (2 Huvud 0.61, 5 Huvud 0.5, 1 Mellow SHT36) successfully using this method!

Helpful to know. Still waiting on my Huvud 0.61 boards.

daagaak:
Also works for me on a Huvud 0.61 + CANable (candlelight) @ 500kbit/s

smackoz:
> Still waiting on my Huvud 0.61 boards.

They have shipped!!!

Benoitone:
work fine on my Mellow SHT36 and SHT42

JNP-1:
Tested with a MCP2515, Canable and UCCB1.2 all at 500kbit/s and three nodes (STM32F103). Everything worked as expected.

TitusLabs:
Works perfect with my custom boards based on a STM32F103. I tested it with the MCP2515 pi head and a Canable. Many Thanks for this!!

JNP-1:
After this commit
https://github.com/Klipper3d/klipper/commit/77937846402f011c4c49a97027606902acea7a12
the flash_can.py script can no longer read a UUID.
Klipper can still communicate with the known UUIDs.
Can anyone else confirm this?

Arksine:
I'll look into it.  AFAIK the SPI changes shouldn't affect CanBoot or the flash_can script.

JNP-1:
I also do not know what SPI should have to do with the bootloader.
To be sure I read out the MCUs with the STLink.
At address 0x08000000 I can see the bootloader and at address 0x08002000 I can see the Klipper firmware.
I have this behavior on two systems, once with v0.10.0-378-g77937846 and once with v0.10.0-379-g7db07c46.

JNP-1:
I had made a mistake in the branch. After I corrected it, everything works as usual again.

KevinOConnor:
Thanks.  Looks like very useful functionality!

In general, it looks fine to me.  One note - it looks like the `flash_can.py` script has an upstream location in the CanBoot repo, and as such I think it should live in the Klipper lib/ directory (and not the scripts/ directory) if it is also present in Klipper.

Not blockers, but I have a couple of other comments:
1. It seems odd that we'll have two slightly different CAN enumeration strategies (one for Klipper and one for CanBoot).  I wonder if we could extend the existing protocol to support both programs?  For example, we could extend the CANBUS_RESP_NEED_NODEID message to include a program identifier in the 8th byte (0 for klipper, and 1 for CanBoot).
2. I'm a little leery about adding another set of hardcoded "enter the bootloader" functions.  It would be great if we could auto-detect the bootloader type at runtime - then we could enable that functionality unconditionally and use it in for both USB and CAN bootloader requests.

Thinking out loud, one way to detect the bootloader would be to put a 64bit signature in the bootloader code that Klipper could detect.  For example, on ARM boards we could look at the 8-bytes before the initial code vector to see if it matches a particular constant.  For example:
```c
    uint32_t *bl_vectors = (void*)0;
    uint64_t *bsig = (void*)(bl_vectors[1] - 8);
    if (*bsig == CANBOOT_SIGNATURE)
        ...
```

It should also be possible to enter the bootloader using a cross-platform mechanism.  All current chips do not reset memory on a reset, so if we put a signature in memory before rebooting then the bootloader should be able to detect this.  On ARM boards we could use the bootloader stack address for the memory location to use.  For example:
```c
    uint32_t *bl_vectors = (void*)0;
    uint64_t *reqsig = (void*)bl_vectors[0];
    *reqsig = CANBOOT_REQUEST_BOOTLOADER;
```

Thoughts?
-Kevin

KevinOConnor:
> Do we need another admin command to identify the program itself?

Maybe add another command for SET_NODEID.  For example, CANBUS_CMD_SET_NODEID=0x01 today, but we could rename it to CANBUS_CMD_SET_KLIPPER_NODEID=0x01 and CANBUS_CMD_SET_CANBOOT_NODEID=0x11.  Perhaps use a convention that the top four bits of SET_NODEID is the program type (and match it with the 8th byte of CANBUS_RESP_NEED_NODEID).  Just "thinking out loud".  There's nothing wrong with your original approach - just wondering if there are alternatives that may enhance future flexibility.

> I'd have to do some research to determine how to insert a constant at the specific address 8 bytes prior to the initial vector.

I was thinking of something like:
```diff
--- a/src/generic/armcm_boot.c
+++ b/src/generic/armcm_boot.c
@@ -22,7 +22,7 @@ extern uint32_t _stack_end;
  * Basic interrupt handlers
  ****************************************************************/
 
-static void __noreturn
+void __noreturn __visible
 reset_handler_stage_two(void)
 {
     int i;
@@ -78,17 +78,21 @@ reset_handler_stage_two(void)
         ;
 }
 
+#define CANBOOT_SIGNATURE 0x21746f6f426e6143 // CanBoot!
+
 // Initial code entry point - invoked by the processor after a reset
 // Reset interrupts and stack to take control from bootloaders
-void
-ResetHandler(void)
-{
-    __disable_irq();
-
-    // Explicitly load the stack pointer, jump to stage two
-    asm volatile("mov sp, %0\n bx %1"
-                 : : "r"(&_stack_end), "r"(reset_handler_stage_two));
-}
+asm(".section .text.ResetHandler\n"
+    ".8byte " __stringify(CANBOOT_SIGNATURE) "\n"
+    ".global ResetHandler\n"
+    ".type ResetHandler, %function\n"
+    "ResetHandler:\n"
+    "    cpsid i\n"
+    "    ldr r3, =_stack_end\n"
+    "    mov sp, r3\n"
+    "    b reset_handler_stage_two\n"
+    "    .pool\n");
+extern void ResetHandler();
 DECL_ARMCM_IRQ(ResetHandler, -15);
 
 // Code called for any undefined interrupts
```

-Kevin

KevinOConnor:
> I do know how to put it directly after the vector table by modifying the linker script.

Although that is also possible, it's less convenient because every ARM chip has a different vector table length.  Using a dummy ResetHandler that just jumps to the real boot code is a convenient way to embed a constant at an easily obtainable address.

Another option is to use a more elaborate signature and have Klipper scan the first 16KiB of flash.  At first glance though, that seems like more work.

> I didn't move Klipper's stack, so it is writing to the last 8 bytes of the end...it works but I'm not sure that this is desirable.

It shouldn't be necessary to move Klipper's stack.  As long as the memory setting is done with irqs disabled and `NVIC_SystemReset()` is called immediately after the assignment then I think it should be safe.  We do this trick currently with `usb_reboot_for_dfu_bootloader()` in the stm32f0.c/stm32f4.c code.

Thanks for looking at this.
-Kevin

KevinOConnor:
Oh - in retrospect, arm has a goofy 1 byte offset on code labels, so the signature check might need to be something like `uint64_t *bsig = (void*)(bl_vectors[1] - 9);`.

-Kevin

Arksine:
> I was thinking of something like:

That is a nice trick, worked perfectly.  Tested on both a BluePill (F103) and Nucleo (F072).  I have made the changes to CanBoot and pushed them.  CanBoot also now uses Klipper's admin IDs.  Canboot specific commands (SET_NODE_ID) start at 0x10.

For this PR, I have made the following changes:
1) Moved `flash_can.py` to `lib/canboot`
2) Implemented the cross-platform bootloader request in `canbus.c`. 
3) Report the "application id" in the `CANBUS_RESP_NEED_NODEID` response.
4) Removed the platform specific changes to the stm32fx.c files.
5) Modified `canbus_query.py` to report the application.  It will always report Klipper if dlc only contains 7 bytes so it should work with anyone running the previous code.

Arksine:
> I was thinking of something like:

That is a nice trick, worked perfectly.  Tested on both a BluePill (F103) and Nucleo (F072).  I have made the changes to CanBoot and pushed them.  CanBoot also now uses Klipper's admin IDs.  Canboot specific commands (SET_NODE_ID) start at 0x10.

For this PR, I have made the following changes:
1) Moved `flash_can.py` to `lib/canboot`
2) Implemented the cross-platform bootloader request in `canbus.c`. 
3) Report the "application id" in the `CANBUS_RESP_NEED_NODEID` response.
4) Removed the platform specific changes to the stm32fx.c files.
5) Modified `canbus_query.py` to report the application.  It will always report Klipper if dlc only contains 7 bytes so it should work with anyone running the previous code.

scripts/canbus_query.py
```diff
@@ -37,7 +37,14 @@ def query_unassigned(canbus_iface):
...
         found_ids[uuid] = 1
-        sys.stdout.write("Found canbus_uuid=%012x\n" % (uuid,))
+        app = "Klipper"
+        if msg.dlc > 7:
+            if msg.data[7] == 1:
+                app = "CanBoot"
+            elif msg.data[7] != 0:
+                app = "Unknown"
+        sys.stdout.write("Found canbus_uuid=%012x, Application: %s\n"
+                         % (uuid, app))
```
KevinOConnor, line 47: Minor, but it may be nice to tie the application id to the SET_NODEID_CMD identifier (thus perhaps fewer "magic values").  Something like:
```python
    AppNames = {CMD_SET_KLIPPER_NODEID: "Klipper", CMD_SET_CANBOOT_NODEID: "CanBoot"}
    ...
    app_id = CMD_SET_KLIPPER_NODEID
    if msg.dlc > 7:
        app_id = msg.data[7]
    app_name = AppNames.get(app_id, "Unknown")
    ...
```

src/generic/canbus.c
```diff
@@ -153,6 +161,21 @@ can_process_set_nodeid(uint32_t id, uint32_t len, uint8_t *data)
     }
 }
 
+static void
+can_process_request_bootloader(uint32_t id, uint32_t len, uint8_t *data)
+{
+    if (!can_check_uuid(id, len, data))
+        return;
+    irq_disable();
```
KevinOConnor, line 169: This `irq_disable()` should be right before the `*req_sig = ...` line.  Otherwise, if the canboot bootloader is not found then the function will return with irqs disabled which could cause havoc.  Only the assignment and reset needs to be protected.

src/generic/canbus.c
```diff
@@ -153,6 +161,21 @@ can_process_set_nodeid(uint32_t id, uint32_t len, uint8_t *data)
...
+    uint32_t *bl_vectors = (uint32_t *)(CONFIG_FLASH_START & 0xFF000000);
+    uint64_t *boot_sig = (uint64_t *)(bl_vectors[1] - 9);
+    if (*boot_sig == CANBOOT_SIGNATURE) {
```
KevinOConnor, line 172: We also need to check for alignment here - as some ARM platforms will crash on an unaligned integer access.  Something like:
```c
    if (boot_sig == ALIGN(boot_sig, 8) && boot_sig == CANBOOT_SIGNATURE && req_sig == ALIGN(req_sig, 8)) {
        ...
```

I guess the ResetHandler assembler should also have a `.balign 8` directive just before the CANBOOT_SIGNATURE to ensure it is 8-byte aligned.

KevinOConnor (commented):
Thanks!  I have a few minor comments (see below).  I'll also try to test this on my boards in the next couple of days.

Separately, I think we can rename the existing `usb_request_bootloader()` code to `request_bootloader()` and call it from both CAN and USB code.  However, something like that can be done after merging this.

-Kevin

src/generic/canbus.c
Arksine: Oops, that would be bad.  Fixed.

src/generic/canbus.c
Arksine: This should be fixed .  I had to do some type casting to keep the compiler from complaining, but the result should be correct.  I also added `balign 8` to the asm in CanBoot.

Arksine:
I went ahead and pushed the new changes as separate commits, hoping that it will make reviewing a bit easier.  When everything is acceptable  I'll squash them.

Arksine:
Should we add a check like the following?
```diff
diff --git a/src/generic/canbus.c b/src/generic/canbus.c
index dc470c25..c604d9d0 100644
--- a/src/generic/canbus.c
+++ b/src/generic/canbus.c
@@ -163,7 +163,7 @@ can_process_set_nodeid(uint32_t id, uint32_t len, uint8_t *data)
 static void
 can_process_request_bootloader(uint32_t id, uint32_t len, uint8_t *data)
 {
-    if (!can_check_uuid(id, len, data))
+    if (!can_check_uuid(id, len, data) || !(CONFIG_FLASH_START & 0xFFFFFF))
         return;
     uint32_t *bl_vectors = (uint32_t *)(CONFIG_FLASH_START & 0xFF000000);
     uint64_t *boot_sig = (uint64_t *)(bl_vectors[1] - 9);
```
Basically make sure that a bootloader offset has been configured so we don't attempt to read Klipper's vector table.   I *think* that between the alignment check and the boot signature check we prevent anything bad from happening, but it can't hurt to abort when we know a bootloader is not present.  That said, if we plan to merge everything into a `request_bootloader` we would need to handle it differently there, so perhaps its a good idea to wait.

lib/README
```diff
@@ -122,3 +122,9 @@ has been modfied to enable the US code page (437) and enable label
 support.  The file "diskio.c" in the original archive is not included
 as it is only serves as an example for how to implement the media api
 callbacks.
+
+The canboot directory contains code from:
+  https://github.com/Arksine/CanBoot
+revision f0d7bb7f9c189de2c22af9eb7210487ed5695f58.  The Pyhon module,
```
KevinOConnor, line 128: Minor typo - "Pyhon" vs "Python".

src/generic/canbus.c
```diff
@@ -90,8 +91,13 @@ console_sendf(const struct command_encoder *ce, va_list args)
 // Available commands and responses
 #define CANBUS_CMD_QUERY_UNASSIGNED 0x00
 #define CANBUS_CMD_SET_NODEID 0x01
+#define CANBUS_CMD_REQUEST_BOOTLOADER 0x02
 #define CANBUS_RESP_NEED_NODEID 0x20
 
+// CanBoot definitions
+#define CANBOOT_SIGNATURE 0x21746f6f426e6143
+#define CANBOOT_REQUEST   0x5984E3FA6CA1589B
```
KevinOConnor, line 99: Just out of curiosity, what's the background to 0x5984E3FA6CA1589B?  The 0x21746f6f426e6143 is "CanBoot!" in ascii - which will show up nicely in a hexdump.

KevinOConnor (commented):
Thanks.  Changes look fine to me.  I'll give a few days to see if there are further comments.

Also, some minor comments below.

-Kevin

KevinOConnor:
> Should we add a check like the following?
> Basically make sure that a bootloader offset has been configured so we don't attempt to read Klipper's vector table. 

FWIW, I don't think it matters much one way or the other.  No harm in checking, but no harm in rebooting in this case either.  (Afterall, if the user requests a reboot into a boot loader that doesn't exist, there's no harm in rebooting.)

Cheers,
-Kevin

src/generic/canbus.c
Arksine: It was random.  I can create an ascii signature if you prefer.

Arksine:
I fixed the "Python" typo and squashed the commits together.  It would be great if those testing this branch could build and flash the latest version of CanBoot and test that everything still works as expected with their devices.

JNP-1:
All devices (STM32F103) I have tested work as expected.
Double reset works without problems and always jumps into the bootloader.
Both applications are recognized reliably every time.
Very good work.

eddietheengineer:
Needed to use:

python3 lib/canboot/flash_can.py -q

to query CAN IDs and:

python3 lib/canboot/flash_can.py -i can0 -f ~/klipper/out/klipper.bin -u aabbccddeeff

to flash

eddietheengineer:
Needed to use:

python3 lib/canboot/flash_can.py -q

to query CAN IDs and:

python3 lib/canboot/flash_can.py -i can0 -f ~/klipper/out/klipper.bin -u aabbccddeeff

to flash

KevinOConnor:
Looking at the canboot code, it seems there is a different nodeid to canbus id mapping - Klipper uses `canbusid = nodeid * 2 + 256` while canboot uses `canbusid = nodeid * 2 + 512`.  Is this difference intentional?  It seems like it could lead to confusion as a klipper nodeid of 130 would conflict with a canboot nodeid of 2.

-Kevin

Arksine:
Yes, that was intentional, I did so in an attempt to avoid ID conflicts.   Since nodes are assigned sequentially I reasoned that it is unlikely that a user would have over 128 nodes, but I agree that it is not impossible.

One solution would be a "clear node id" admin command in Klipper, which should reduce the chance of collisions.  That said, I do believe it would still be a good idea to offset CanBoot's ID's differently from Klipper's.

pawlowskialex:
Is it conceivable for `CanBoot` and `flash_can.py` to be compatible with ST's [AN3154](https://www.st.com/resource/en/application_note/an3154-can-protocol-used-in-the-stm32-bootloader-stmicroelectronics.pdf), so that we have a unified update process for F0/F1 and F4 (that have a built-in CAN bootloader)?

Arksine:
@pawlowskialex While I'm sure its possible to implement ST's protocol I don't think its suitable for CanBoot.  We want to keep the protocol simple and deliver the bootloader in the smallest possible footprint.  I'm also unsure if there could be any licensing issues, I suspect not but better to err on the side of caution.  That said, if someone wants to fork CanBoot and take a crack at it then by all means they are welcome to do so.

Since we are discussing the protocol, its a good time let everyone know that I [reworked it to add CRC validation](https://github.com/Arksine/CanBoot/pull/12).  Previously the protocol limited the command size to 8 bytes, relying on the CAN frame itself to perform the CRC validation.  Blocks of data weren't CRC checked, instead the process relied on reading out the flash memory and validating it determine if the upload succeeded.   On some devices validation would consistently fail due to dropped or reordered frames, however the upload may still have been successful.

With the new protocol the bootloader can reject a bad block before writing to flash, prompting `flash_can.py` to resend.  During validation `flash_can.py` can re-request a block if frames are dropped (or reordered).   As soon as the new protocol is merged I will update `flash_can.py` in this PR, after which I believe it will be ready to merge as well.

CKay3010:
Hi everybody,
i tested it on FLY SHT 36, SHT 42 and on Huvud 0.61 for the STM32F072 klipper need´s a 8kb Bootloader offset.

The funktion with Pin sate is working

Arksine:
Hi everyone.  WIth Kevin's assistance we have managed to improve CanBoot quite a bit over the last several days.    I have added the latest version of `flash_can.py` to this PR, it is now in sync with CanBoot master.  It will be necessary to pull the CanBoot changes, rebuild and program the latest version.

Verification should now complete successfully even on "flaky" can devices, although it make take some time if `flash_can.py` needs to request blocks multiple times.

CKay3010:
I tried it with the klipper can boot build and it is woking with switching to bootloader for  Huvud 0.61

JNP-1:
Tested with a STM32F103 toolhead board. Everything works without problems.

KevinOConnor:
Thanks!  I committed this along with be4b8b17 on top.

-Kevin

Mosher23:
Hi,
I have tested canboot and it works with Mellow SHT42 board.
But when i qeurry for UUIDs i do not get any results. (I had my id before canboot)
`pi@mainsail:~/klipper $ python3 lib/canboot/flash_can.py -i can0 -q`
`Resetting all bootloader node IDs...`
`Checking for canboot nodes...`
`Query Complete`
`pi@mainsail:~/klipper $`

Mosher23:
Thank you for this information.
Could you please add to the installation instruction this information and also how to reset the board/device?

PR #5426, opened 2022-04-17
In-code documentation for DFRobot Gravity Expansion Board for Pico

thinkier:
I'm currently in the process of acquiring the said expansion board and I think my research would come in useful for other humans looking at the source in the future.

The comments makes a note of the physical pins on the development board and matches it against the klipper names of said pin groups.

image from dfrobot's website describing its pins and corresponding GPIO


Sineos:
I think this would be better placed in the "Configs" section of https://klipper.discourse.group.
Developers typically would not need this information and users, for whom it might be useful, typically do not look through source files.

thinkier:
Fair enough, I'll close the PR.

PR #5428, opened 2022-04-18
Add hints to check printer config file comments in Installation.md

nleskiw:
Add suggestion to check the comments at the top of printer config file for users' printer in `~/klipper/config/` when doing a `make menuconfig` for the correct options to select in that configuration screen, and also for alternate firmware flashing methods if `make flash` doesn't work.

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -49,6 +49,10 @@ cd ~/klipper/
 make menuconfig
 ```
 
+It is recommended that you review the comments at the top of the printer
+config file in `~/klipper/config/` for your printer. This has hints on 
+which options to select in this menu. 
+ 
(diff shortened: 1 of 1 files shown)


ld3300:
I can't do anything to approve, but I like the concept.  Make sure you add the Signed-off-by: line to your commit so it can be merged when reviewed.

KevinOConnor:
Thanks.  I agree that the installation document should do a better job of referencing the instructions in the example config file, however I think the document needs some reorganization to make that more clear at the start.  I've attempted this in PR #5496 .

-Kevin

KevinOConnor:
I committed PR #5496 .

-Kevin

PR #5431, opened 2022-04-18
cleanup code & missing import/class

Ludy87:
- 🐛 add error class
- ⚰️ remove obsolete imports
- ⚡️ declaration conflict
- 🐛 add missing import
- 🌱 `print` Python2 to Python3

### DIFF
--- a/klippy/extras/buttons.py
+++ b/klippy/extras/buttons.py
@@ -3,7 +3,6 @@
 # Copyright (C) 2018  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import logging
 
 
 ######################################################################
--- a/klippy/extras/delta_calibrate.py
+++ b/klippy/extras/delta_calibrate.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2017-2019  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import math, logging, collections
+import math, logging
 import mathutil
 from . import probe
 
--- a/klippy/extras/display/hd44780.py
+++ b/klippy/extras/display/hd44780.py
@@ -4,7 +4,6 @@
 # Copyright (C) 2018  Eric Callahan <arksine.code@gmail.com>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import logging
 
 BACKGROUND_PRIORITY_CLOCK = 0x7fffffff00000000
 LINE_LENGTH_DEFAULT=20
--- a/klippy/extras/display/hd44780_spi.py
+++ b/klippy/extras/display/hd44780_spi.py
@@ -5,7 +5,6 @@
 # Copyright (C) 2021  Marc-Andre Denis <marcadenis@msn.com>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import logging
 from .. import bus
 
 LINE_LENGTH_DEFAULT=20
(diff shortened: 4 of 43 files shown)


dewi-ny-je:
I think the Signed-off line should contain a real name and surname.

Haxk20:
> I think the Signed-off line should contain a real name and surname.

Correct. Signed-off-by is to represent the committer has the rights to submit this code and agrees to DCO (Developer Certificate of Origin)

Ludy87:
> > I think the Signed-off line should contain a real name and surname.
> 
> Correct. Signed-off-by is to represent the committer has the rights to submit this code and agrees to DCO (Developer Certificate of Origin)

How meaningful is the name, exactly zero. The signature is verified as is. Where is the problem?

Haxk20:
> > > I think the Signed-off line should contain a real name and surname.
> > 
> > 
> > Correct. Signed-off-by is to represent the committer has the rights to submit this code and agrees to DCO (Developer Certificate of Origin)
> 
> How meaningful is the name, exactly zero. The signature is verified as is. Where is the problem?

The name is very meaningful as it gives people a way of contacting you in case of issues with the commit(s).

Plus klipper contributions REQUIRE you to use your real name and valid email address in sign off.

Sineos:
Reference to the signed-off line: https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages
DCO can be downloaded here: https://www.klipper3d.org/developer-certificate-of-origin

The intention behind the DCO and this requirement is to keep the project legally on the safe side. Github accounts or "nick names" are nothing you can draw upon in a legal sense and the regulations in the DCO reference to you as a natural person and not as a GH account.
The Linux Foundation who introduced this DCO in 2004 learned it the hard way that it is necessary to establish such safeguards.

For many contribution it might seem overdone but where do you want to draw the line? So the most straight forward approach is to require it for each and every contribution.

Ludy87:
How do we want to continue with the PR now?

Haxk20:
> How do we want to continue with the PR now?

As Sineos mentioned above you will have to use real name and working email for Sign off.

Then Conor and Sine will do reviews

Ludy87:
okay, otherwise the bugs remain?

Haxk20:
> okay, otherwise the bugs remain?

Well since real name has to be used for legal reasons. So if you are asking if the pull request will be merged without using your real name. Thats up to Conor. Considering its to avoid legal trouble. Yea i would say the pull request wont be merged otherwise.

Haxk20:
But excluding the sign off for now.

The commit names have to follow the used standard which is "<the subsystem of klipper you are touching>: Explaining the change"

So for example the error class commit name would look like "motan: Add error class"

Ludy87:
Sorry, but when bugs are fixed, what rights are meant to be violated? I accept the way, but in some things it is too strict and should be reconsidered.

dewi-ny-je:
> Sorry, but when bugs are fixed, what rights are meant to be violated? I accept the way, but in some things it is too strict and should be reconsidered.

I searched online because I was curious and I found

https://drewdevault.com/2021/04/12/DCO.html
https://opensource.stackexchange.com/questions/10138/is-it-in-line-with-the-dco-that-a-github-sign-off-needs-and-publishes-full-name
https://stackoverflow.com/questions/62969381/is-it-in-line-with-the-dco-that-a-github-sign-off-needs-and-publishes-full-name

I'm not sure it solves anything, but it may help.

Ludy87:
Do as you see fit, only real names are not required in the DCO. It would be a pity that a name that cannot be controlled is mandatory. a nickname is common and will always be traceable.

KevinOConnor:
Thanks - I fixed the two bugs I could find in this PR (commit de14d106 and 5517a856).  (The stepstats.py script would needs a port to python3 beyond just the change to `print`.)

It is the current policy to require a signed-off-by line containing one's full real name for contributions.  This policy is in keeping with other popular projects (eg, Linux).  I understand there are downsides to this policy, however I plan to continue with it, at least for the short to medium term.

Cheers,
-Kevin

PR #5442, opened 2022-04-21
Update Command_Templates.md

tispokes:
added very useful hint to [save_variables config section]

### DIFF
--- a/docs/Command_Templates.md
+++ b/docs/Command_Templates.md
@@ -351,5 +351,4 @@ gcode:
   ACTIVATE_EXTRUDER extruder={svv.currentextruder}
 ```
 
-Be aware to only use small letters for the variable name, otherwise you may can't overwrite the stored data. Even if the console confirm with `Variable Saved
-`.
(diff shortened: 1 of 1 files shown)


Sineos:
Already mentioned in the documentation:
https://www.klipper3d.org/Command_Templates.html?h=contain+upper+case#variables

tispokes:
> Already mentioned in the documentation: https://www.klipper3d.org/Command_Templates.html?h=contain+upper+case#variables

Oh, then I was not able to find this. Some kind of cross reference would be nice in such cases. Like a footnote

docs/Command_Templates.md
```diff
@@ -350,3 +350,5 @@ gcode:
   {% set svv = printer.save_variables.variables %}
   ACTIVATE_EXTRUDER extruder={svv.currentextruder}
 ```
+
+Be aware to only use small letters for the variable name, otherwise you may can't overwrite the stored data. Even if the console confirm with `Variable Saved`.
```
dewi-ny-je, line 354: "you may be unable to overwrite"

PR #5444, opened 2022-04-22
Config_Reference: Change description in [screws_tilt_adjust]

JamesH1978:
Original description for screw1 was ambiguous to the fact that it is the probe that needs to be over the screw and not the nozzle, and that it is expressed in nozzle coordinates.


KevinOConnor:
Thanks.  I committed a slightly different wording (commit 07a90902).

-Kevin

PR #5445, opened 2022-04-22
printer-creality-ender3pro-2020: Update info on chip types

JamesH1978:
Added flash info that some Ender 3 Pro's shipped in 2022 have the GD32F103 chip


Sineos:
Acc to https://chowdera.com/2022/04/202204091601119439.html there are quite some differences between STM32 and GD32 chips.

@KevinOConnor is this something where we can expect subtle errors creeping up in the support channels in the future?

KevinOConnor:

As to support issues - alas, I don't know.  It seems several boards are using this stm32f103 clone.

-Kevin

Issue #3379, opened 2020-09-27
installing to stm32f103zet6 (trigorilla pro board) - can't connect.


sprig:
Hi,

I'm using a trigorilla pro board which has an STM32F103ZET6 chip on it. It also has jumpers for to put the chip in bootloader mode so that one can freely flash the chip. USART1 is connected to USB via a pl2303 it appears, and I'm able to flash and run marlin 2.0.x successfully on it.

I tried building and flashing klipper. Everything appears to run smoothly, except that I can't actually connect to it. build config and log attached. I tried flashing both via the makefile as well as manually with `stm32flash`.

Pointers/help welcome.

P.S.
The lines 
```
Got error 0 in read: (0)Success
Timeout on serial connect
Unable to open port: [Errno 2] could not open port /dev/ttyUSB0: [Errno 2] No such file or directory: '/dev/ttyUSB0'
```
from the log file are not from cable errors or the like - they were from when I moved the jumpers back into normal operation and restarted the board.

[config.txt](https://github.com/KevinOConnor/klipper/files/5289156/config.txt)
[klippy.log](https://github.com/KevinOConnor/klipper/files/5289136/klippy.log)

KevinOConnor:
That error indicates the host is unable to contact the micro-controller.  It generally means the board was either not successfully flashed or the settings in "make menuconfig" are not correct.

It's never been necessary for any other board, but you could try a different baud rate.  (I see references that the max for the pl2303 is 230400 baud.)  If so, make sure to change both in "make menuconfig" and in the printer.cfg file (see https://www.klipper3d.org/FAQ.html#how-do-i-change-the-serial-baud-rate ).

-Kevin

Issue #2822, opened 2020-05-03
TriGorrila board.


Hairyloon:
Did I not post my post about my board blowing up or did it get taken down? No matter, I've got a new board now, but I can't find it listed in the config folder.
Is a TriGorrilla 1.1, which I believe comes from AnyCubic. There are two AnyCubic files: i3 and 4max: which is it more likely to be?
Or do I have to go through all the pins manually?

Hywelmartin:
believe it's more or less a generic-ramps.cfg

klipper-gitissuebot:
Hi @Hairyloon,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

Unfortunately, too many people have opened tickets without providing the log. That consumes developer time; time that would be better spent enhancing the software. If this ticket references an event that has occurred while running the software then the Klipper log must be attached to this ticket. Otherwise, this ticket will be automatically closed in a few days.

For information on obtaining the Klipper log file see: https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md

The log can still be attached to this ticket - just add a comment and attach the log to that comment.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

ruevs:
The Anycubic deltas use the TriGorilla board.
For example:
https://github.com/KevinOConnor/klipper/blob/master/config/printer-anycubic-kossel-plus-2017.cfg

Hairyloon:
> believe it's more or less a generic-ramps.cfg

Yes, but the question is one of pin assignments or aliases.

Hairyloon:
> The Anycubic deltas use the TriGorilla board.
> For example:
> https://github.com/KevinOConnor/klipper/blob/master/config/printer-anycubic-kossel-plus-2017.cfg

Ah, good stuff. Thank you.
Is it Ok to leave this open for now, in case I have trouble translating it into Cartesian?

Hywelmartin:
[TriGoRiLLa_PINs.pdf](https://github.com/KevinOConnor/klipper/files/4571448/TriGoRiLLa_PINs.pdf)
here is the pinout

schnello:
>Is it Ok to leave this open for now, in case I have trouble translating it into Cartesian?

Use the i3config. Same Board but Cartesian.
[...]

PR #5446, opened 2022-04-22
printer-anycubic-predator-2020.cfg

lorenzing:
module: Anycubic-Predator configuration

Hello everyone, I share with you the configuration file of the Anycubic Predator printer, with original Trigorilla Pro board.

Signed-off-by: Lorenzo Franco (lorenzo.franco@lorenzing.com)


KevinOConnor:
Thanks.  However, this config would need to follow the guidance at https://www.klipper3d.org/Example_Configs.html - at a minimum point 4 and 7e.

It's also a good idea to set the appropriate gear_ratio for the extruder file (point 7e).

Also, all known Klipper boards work with the default baud of 250000, so a config should not recommend setting a different baud rate.

-Kevin

lorenzing:
> Thanks. However, this config would need to follow the guidance at https://www.klipper3d.org/Example_Configs.html - at a minimum point 4 and 7d.
> 
> It's also a good idea to set the appropriate gear_ratio for the extruder file (point 7e).
> 
> Also, all known Klipper boards work with the default baud of 250000, so a config should not recommend setting a different baud rate.
> 
> -Kevin

Thanks @KevinOConnor, I made the required changes except the baud rate. 
The trigorilla pro is a problematic board (there are other issues regarding it, for example:  #3379 ).
Unfortunately, the maximum baudrate at which I managed to obtain stable operation is 115200. If you want I can still put 230400 which is the theoretical maximum for pl2303.
To help other users I can create a generic file on the trigorilla pro.

Thanks so much
Lorenzo

lorenzing:
I have a problem, the printer works correctly but after entering the configuration in test / klippy / printers.test I see that for this mcu the pins are reserved by default for the usb: RESERVE_PINS_USB = PA11, PA12
The problem is that on the trigorilla pro the pin PA12 is dedicated to the endstop of the Y axis while for the USB Serial communication it uses the USART1
- USART1 TX = PA9 (to RX pin of the PL2303)
- USART1 RX = PA10 (to TX pin of the PL2303)

How can I solve? Thanks

lorenzing:
I am attaching the klippy log of the working printer:
[klippy.log.txt](https://github.com/Klipper3d/klipper/files/8603652/klippy.log.txt)

KevinOConnor:
> I have a problem, the printer works correctly but after entering the configuration in test / klippy / printers.test I see that for this mcu the pins are reserved by default for the usb

If the board uses serial, then you would add it to the section in printers.test for `Printers using the stm32f103 via serial`.  But, the example config file instructs users to enable USB, while it seems you want to instruct them to use UART on PA10/PA9?

> Unfortunately, the maximum baudrate at which I managed to obtain stable operation is 115200.

Okay, can you provide more detail on what occurs when you use the default of 250000 baud?  Can you provide a Klipper log file from the test?

Otherwise, the config looks fine to me.

Thanks,
-Kevin

lorenzing:
> If the board uses serial, then you would add it to the section in printers.test for Printers using the stm32f103 via serial. But, the example config file instructs users to enable USB, while it seems you want to instruct them to use UART on PA10/PA9?

The problem is that, on this board the usb is mapped to the UART on PA10 / PA9 through a pl2303. I have seen several users with communication problems between klipper and this board ( #2822 #3379 ).
When I compile the firmware I go to indicate the UART on PA10 / PA9 but this corresponds to the USB port in reality.
The pin scheme that I was able to partially verify is that: https://3dmixers.com/m/123405-trigorilla-pro-schematics-and-reverse-engineering

> Okay, can you provide more detail on what occurs when you use the default of 250000 baud? Can you provide a Klipper log file from the test?

Yes sure, I've recompiled and flashed the board with firmware at 250000 baudrate (it does the same even if I put 230400 baud):
```
[display_status]
=======================
Delta max build height 454.51mm (radius tapered above 391.31mm)
Delta max build radius 209.65mm (moves slowed past 188.70mm and 205.20mm)
Extruder max_extrude_ratio=0.266081
Starting serial connect
Timeout on serial connect
Wait for identify_response
  File "/home/pi/klipper/klippy/serialhdl.py", line 68, in _get_identify_data
    params = self.send_with_response(msg, 'identify_response')
  File "/home/pi/klipper/klippy/serialhdl.py", line 188, in send_with_response
    return src.get_response(cmd, self.default_cmd_queue)
  File "/home/pi/klipper/klippy/serialhdl.py", line 242, in get_response
    self.serial.raw_send_wait_ack(cmd, minclock, minclock, cmd_queue)
  File "/home/pi/klipper/klippy/serialhdl.py", line 180, in raw_send_wait_ack
    raise error("Serial connection closed")
error: Serial connection closed
Timeout on serial connect
```
[klippy.log_250000.txt](https://github.com/Klipper3d/klipper/files/8604770/klippy.log_250000.txt)

KevinOConnor:
Thanks.  Two comments here:
1. I'm confused by the comments at the top of this config file.  It says to use "usb for communications" in "make menuconfig", but clearly that can't be correct as one would need to select "uart for communication" on this board.
2. I'd like to see a few others confirm that this board does not work at a baud of 250000.  It is possible this is the first board that requires a baud of 115200.  However, without exaggeration, I've seen well over 100 claims that various boards needed a particular baud rating that were not correct (every board Klipper supports to date works at 250000 baud).  So, given the prevalent amount of misinformation on baud rate I feel getting confirmation is important.

Cheers,
-Kevin

lorenzing:
Thanks Kevin and sorry for the delay!
> 
> 1. I'm confused by the comments at the top of this config file.  It says to use "usb for communications" in "make menuconfig", but clearly that can't be correct as one would need to select "uart for communication" on this board.
>
Sorry I made a copy and paste mistake.

> 2. I'd like to see a few others confirm that this board does not work at a baud of 250000.  It is possible this is the first board that requires a baud of 115200.  However, without exaggeration, I've seen well over 100 claims that various boards needed a particular baud rating that were not correct (every board Klipper supports to date works at 250000 baud).  So, given the prevalent amount of misinformation on baud rate I feel getting confirmation is important.
> 
I found a person who seems to have the same problem (issue #3379), unfortunately she never showed up again. 
I ask you if possible to keep this pull request open and in the meantime I write on the facebook group of this printer if someone can do some tests at different baudrates.

Best regards,
Lorenzo

lorenzing:
I don't know if I can post the link to an external site but here they talk about the same problem, always trigorilla pro but on another printer. If you can't, delete the comment.
[https://drucktipps3d.de/forum/topic/i3-mega-trigorilla-pro-und-klipper/](url)

rubydesign:
Sorry to see this closed. It is the best config for a predator that i've seen.
Unfortunately i can't confirm now because i installed an octopus even before finding this.

But @lorenzing what did you do with the bed levelling probe? Did i just miss that?

PR #5448, opened 2022-04-23
spi_flash: support spi protocol changes

Arksine:
This patch adds support for the latest `spi_config` command, while retaining support for the previous version.

While the patch worked as expected in my testing, there was a peculiar behavior that I noticed.  If Klipper (or `spi_flash`) attempts to configure the MCU with the incorrect spi command, the board requires a power cycle to flash correctly with `spi_flash`.  As long as this condition does not occur I can freely flash between older and newer versions of Klipper.

This may be an issue with the bootloader for the SKR Mini E3 v1.2, I plan to test some other boards to see if I can reproduce.  I don't believe that this was introduced with this change or by 77937846402f011c4c49a97027606902acea7a12.


KevinOConnor:
Thanks!

-Kevin

PR #5450, opened 2022-04-23
spi_flash: resolve configuration issues

Arksine:
This patch should resolve the issues previously discussed.  Two changes are added:
1) The `spi_flash` module now validates that the MCU is configured after sending a configuration.
2) The SPI frequency is reduced to 400kHz, which is within the recommended guidelines for SD Card initialization.  While a typical sd card would increase the frequency after init, it isn't necessary for our application.

I have tested these changes on an SKR Pro v1.1 and SKR Mini E3 v1.2.   I plan to test on an SKR 1.3 and Flyboard Mini as well, however I wanted to get this PR up to give the opportunity for others to test.  In particular it would be good to make sure that the SPI frequency change doesn't negatively impact other boards, if necessary we can specify the frequency on a per-board basis.

### DIFF
--- a/scripts/spi_flash/spi_flash.py
+++ b/scripts/spi_flash/spi_flash.py
@@ -119,6 +119,9 @@ FINALIZE_CFG_CMD = "finalize_config crc=%d"
 class SPIFlashError(Exception):
     pass
 
+class MCUConfigError(SPIFlashError):
+    pass
+
 class SPIDirect:
     def __init__(self, ser):
         self.oid = SPI_OID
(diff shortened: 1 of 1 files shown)


meteyou:
tested on a skr 1.3 (table setup) and looks fine:

meteyou:
tested with a fysetc spider v2.2 and looks also fine:

image

meteyou:
tested with a mks sbase (table setup) and looks also fine:

image

KevinOConnor:
Thanks.  It looks fine to me.  Let me know when you are comfortable committing.

-Kevin

Arksine:
I think its good.  Works on my all of my boards here.  Thanks!

PR #5452, opened 2022-04-24
Extrusion speed limit. Closes #2745

### LINKED ISSUE #2745: [FEATURE] Extruder maximum volumetric speed?
I tried a larger nozzle today and stumbled upon the maximum volumetric rate limitation of the hotend.
It means that every hotend has a limited capacity which is expressed as the “maximum volumetric speed” it can handle. If you attempt to move more filament than this through the hotend, you’ll eventually encounter a variety of problems, including:
Low extrusion and uneven layers.
Extruder click and skips.
Nozzle and hotend jams.
Increased extruder motor heat.

[Calibrating maximum volumetric rate](http://projects.ttlexceeded.com/3dprinting_calibrating_volumetric_rate.html)

This maximum volumetric rate is dependent on many variables, printing temp, nozzle size, filament characteristics.

Some slicers have this maximum volumetric speed limit builtin (prusaslic3r for example)

I was wondering if this kind of feature is needed in Klipper?
Let's say a change in extruder kinematics to limit print move speed (if the user wants).
User can specify maximum volumetric speed in config or dynamically via gcode for the extruder. The unit is cubic mm in sec. By setting this limit the extruder will check a print move
and if the extruder volumetric speed will exceed the limit then the print move speed will be clamped down to the safe range.

masterxq:
Hello *,
I implemented an extrusion speed limit. It will reduce the speed of movements with extrusion (excluding retraction), if they hit the limit. The PR includes this new commands:
- SET_FILAMENT_DIAMETER
- SET_MAX_EXTRUDER_VELOCITY
- GET_MAX_EXTRUDER_VELOCITY
For additional information check doc/G-Codes.md

The configuration now allows to set:
`max_extruder_velocity`
it defaults to the same limit as `max_extrude_only_velocity == max_velocity * def_max_extrude_ratio`
See docs/Config_Reference.md for details.

The volumetric flow can not be set directly in the configuration. This would possibly induce confusion because its dependencies:
volumeric_flow = function_of(max_extruder_velocity, filament_diameter)
However, it still can be set thought the set cmd.

This is an important functionality for being able to implement material profiles and switch between them for reprint without reslicing. For e.g. ongoing this: #49.

The formulas have been triple checked, and multiple test prints has been successfully made.

Closes #2745


KevinOConnor:
Interesting.

My high-level feedback is that this would need to demonstrate a "quality improvement" and/or demonstrate that a notable number of users are utilizing it.  (For reference, there is a blurb on this at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review ).  For example, demonstrate higher quality prints when using this module with standard slicer settings.  You might also want to follow up on Klipper Discourse to mention your repo and provide tests that other interested users can run.

FWIW, this isn't something that I would be interested in on my own printer.  (I dislike the similar capabilities in Marlin and RRF.)  I find the interaction between the `max_speed` and `max_extruder_velocity` to be confusing - simple test prints can produce confusing results because more than one config parameter controls speed, and because it's harder to find the root cause to print defects (again due to the interaction of multiple config settings).

Thanks.
-Kevin

nemonada23:
Hi,
At least I have been looking for this functionality for a long time, because it allows me to slice only once for multiple printers that have different machine limits regarding the maximum extruder speeds.

masterxq:
Hello Kevin,
thanks for your review, nice to see a straight strategy inside the base code.
**Let's first talk about users that don't want to set an extrusion limit.**
> FWIW, this isn't something that I would be interested in on my own printer:
1. What about let it default to `0` and skip the complete check if it set to `0`. The changes would have no effect in this case and users who not set this setting will never have problems with it.
2. Add a comment in the config reference: If you are unsure leave it unset. (Even my personal opinion is that this is not the best choice. I want the machine limits always to be respected. But Klipper possibly have slightly other main targets. More about this later)

**About quality**
> For example, demonstrate higher quality prints when using this module with standard slicer settings. You might also want to follow up on Klipper Discourse to mention your repo and provide tests that other interested users can run.

Directly quality improvements can be obtained if the default limits in the slicer is set close to the maximum capability of the extruder and the hotend. The extruded material will be proportional smaller than the expected extrusion.
1. The user can define this limit while setup and changing hardware as this is the moment he will have the knowledge to do so, and they will be respected forever. If he goes above these limits in the slicer the quality will be better without intervention.
2. Reprinting a sliced print comes close to the machine limits without reslicing can be done in significant higher quality by lower the maximum volumetric flow rate to runtime (with a script or a command). There will be no additional lose of speed of time for section where the speed is not coming close to the machine limits or for travels.

As I also see the usability, flexibility and safety as quality improvement of the software, I see the following additional points:
1. As nemonada23 pointed out. It can avoid to reslice for similar printers with other machine limits.
2. Printers can have many users that are new to 3D printing (as mine). It is hard to control every print and every slicer setup. This results in broken prints. The other users target is not to learn 3D Printing but to have functional parts for their projects.
3. As already mentioned it possibly to configure some material profiles. This allows to reprint without reslicing. The maximum volumetric rate is a very important limit that is needs to be set by changing from e.g. PLA to PETG (16 mm³/s vs 10 mm³/s). All in all this gives more flexibility of printing, and the **option** of saving much of time by controlling the most important parameters of the printer in the printer instead of in the slicer. What I see as one of the most significant reasons to switch to a more powerful host controller as klipper does. If wished, I can prove that high quality prints can be reprinted with other material without reslicing by using material profiles implemented to klipper. The slicer does not do that much magic!
4. Being able to test the difference of quality by setting different maximum extruder speed. Eg. with parametric Test.
5. It feels really nice to know that given machine limits will be respected!

**Let's talk about user base:**
I have about 20 (including myself) friends that comes and using my printers in the weekend. I have some kind of hackerspace here. 10 of them have their own printer. 2 of these 10 will newer be able to configure there printer correctly. 3 of them will continue using marlin firmware. The remaining 5 decided to at least to test klipper as they did see it running on my printers.
- 1 of them is missing this feature as much as he goes to implement it his self :)
- 1 of them was searching this feature near as long as me.
- 1 of them was at least asking me how to set it.
- 1 of them did not talk his configuration.
- 1 was not able to set it up, so I did.

But ok, 3 of 5 users did look for this feature. 1 maybe because they are used to set this limit in the past. But I would say 2 of 5 users including me really need this.

I have some other points, but most of them are just about usability, flexibility and safety...

**At the last end lets talk about my goals and what you can expect from me or not**
I have limited time. So I will never be a core developer. Additionally, I'm used to writing low level code. But I will implement what is needed to be able to use klipper as I or my friends want it. I have no problem with sharing my code. If something doesn't meet the requirements I will try to change it.
But I can accept if klipper goes another way and use my own repo and syncing it against your master. As said I'm willing to share my code and respect as many rules as I can. But it is an important thing that my direction (of usability and flexibility) is wanted by you and the users. Even if there will be features in my code that significant increases the quality of prints the most of them will only make the printer easy to use! So maybe we should have a short video dialog by 20 minutes and talk about this, because it will take me many more hours of coding for respecting the given conventions. It would be a waste of time the code going in the wrong direction. It would even be a waste of your time. So possibly we really should take the short way and clarify this in a dialog. Perhaps there can also be found a solution in the middle, code reviews of an experienced project member are nice to have to avoid fatal mistakes and can also result in saving time. The compromise could be to let the first PR be a draft and if the feature is interesting for somebody I could clean it up.

In every case thousands of thanks for the thousands of hours YOU offered in this project. Really, excellent work!

Let me hear your thought about the PR and also about the common direction.

dewi-ny-je:
I would use and I regard the feature as a sort of safety parameter, like maximum temperature.
We set a max temp to avoid damage, even if the user Gcode sets it higher.
I would set this flow limit as safety for the extruder: it may happen due to user mistakes that the extrusion is set too high, and this would strip the filament (especially on single drive extruders), which requires lengthy disassembly for a number of extruder configurations.

I'm quite sure that the 100 users required would be met VERY easily.

In theory it's not needed, like limiting max temperature (I'm not talking about the runout protection). In practice, it's a safety which is useful to have.

KevinOConnor:
@dewi-ny-je 

Thanks for the feedback.

> I would use and I regard the feature as a sort of safety parameter

It seems to me that this safety feature already exists in Klipper and is called `max_extrude_cross_section`.  A significant difference between that parameter and this PR is that the current Klipper code checks for "invalid" commands and raises an error, while this PR checks for "questionable" commands and replaces them with "different" commands.  In general, I'm leery of this approach because I've found it "hides" real problems.  I greatly prefer code that "does what it is requested to do" or states "I can't do that".  In my experience, the approach of silently "reinterpreting" commands leads to persistent low quality results.

Admittedly, "move speed" is a bit fuzzy because of acceleration and because some existing checks already "silently" limit it.

Also, admittedly, the use of a "cross section" instead of the more common "flow rate" may make `max_extrude_cross_section` hard to use.  It's certainly possible that the current safety check could be improved.

Cheers,
-Kevin

KevinOConnor:
@masterxq - Thanks.  I appreciate the detailed feedback.  I think I understand your feedback.

To put some context to my earlier message, I think there is a fundamental trade-off here: the cost of additional configuration complexity, documentation complexity, and code complexity weighed against the additional capability to accomplish real-world tasks.

I regularly receive feedback that the Klipper printer.cfg file is daunting to users.  It's been common feedback since Klipper started and I continue to see it regularly today.

It seems to me this PR is adding additional config/documentation/code complexity to enable a different user workflow.  (Specifically, the ability to set max flow rate in the firmware instead of in the slicer.)  It seems to me this would not fundamentally improve the resulting 3d prints, but is a change to the user approach to obtaining them.

I'm reluctant to make the judgement call on the above trade-off.  I'd really like to see lots of other users try it out and tell me their results (in particular, I'd like to see the feedback from long-term Klipper contributors).  We can have a conversation on this, but I don't really feel I'm the person to convince.

> What about let it default to 0 and skip the complete check if it set to 0.

FWIW, I agree it is important to set a good default for parameters, but I'm not sure this would address the config/documentation/code complexity trade-off.

> This allows to reprint without reslicing.

FWIW, it's not immediately clear to me that one can realistically do this and get good results in general.  This is something I'd like to see lots of users try out and report on.  It's also not clear to me that the firmware can implement max flow rate as well as the slicer can implement it (for example, the slicer is in a better position to handle overhangs and bridging).

Finally, let me state that I'm open to changes to Klipper here.  I'm reluctant to be make a judgement call on this based just on the code.  Common slicers already implement max flow rate code.  I'd like to see feedback from Klipper users that they are accomplishing great things with a Klipper implementation of max flow rate.

Thanks again.
-Kevin

dewi-ny-je:
> > This allows to reprint without reslicing.
> 
> FWIW, it's not immediately clear to me that one can realistically do this and get good results in general. This is something I'd like to see lots of users try out and report on. It's also not clear to me that the firmware can implement max flow rate as well as the slicer can implement it (for example, the slicer is in a better position to handle overhangs and bridging).

I'm with you on this.

I myself try to keep the Gcode as independent from the material as much as possible, for example by not defining temperatures in it but by using macros on the printer, setting a flow multiplier via macros which I can edit without reslicing, but material- or printer-independent gcode is a chimera, there are simply too many parameters to configure: fan speed, flow rate, temperatures, bridging settings, ... As you already stated.

It works in my case if switching between similar materials or brands or colors (it also works between PLA and wood PLA), but a gcode for PLA will never work on ABS, and any bridging settings for anything but PETG will be very poor with PETG.

If the goal is to separate gcode from material and machine, the added complexity is in my opinion not worth the result.

masterxq:
Hello,
I think the discussion should **not be reduced on one point** I have mentioned. And as a newcomer that did go thought the configuration and was able to observe others that was going through that process, I can say the problem was not the amount of Options, but the time invested in searching missing options. Specially the `max_extrude_only_velocity` was very misleading also because the corresponding `max_extrude_velocity` configuration was completely missing.

Ok, but I'm as mentioned before, I have no time to change the software praxis and this is not the goal. I need a software that is usable for beginners and allows a good workflow inside projects. And there is no need that the PR is accepted to reach this goal. I can easily share my code and macros internal. If the discussion concludes some when I will happily implement a list of modifications that are needed or wished.
Just for advice, there will not be found a good conclusion if the discussion is reduced to one of many points. Even if this could be falsified there are still a lot of other cases that could matter. And the user's response is often hard to interpret, the devil is in the detail.

Just for information and not to be part of the discussion. Our material profiles currently support these parameters:
MATERIAL_TYPE, MATERIAL_VENDOR, PA_VALUE, E_TEMP, B_TEMP, ACCELERATION, ACCEL_TO_DECEL, VOLUMETRIC_SPEED, FILAMENT_DIAMETER, UNRETRACT_WAIT, RECTRACT_DIST, UNRETRACT_EXTRA_LENGTH, RETRACT_SPEED, UNRETRACT_SPEED, ZHOP_HEIGHT, ZHOP_VELOCITY

This is also working well for material (of same basic type) changing while printing, we did not test changing material **type**.

@dewi-ny-je you are right, it was necessary to reimplement Z-HOP in a save way for reaching good results with PETG. We are only printing with PETG and PLA here, ABS will come, but it's mainly untested. The printer's user will be asked if he wants to overwrite the given Material by the slicer if the Material type differs before the print starts. After the latest update about 10 prints could be printed in excellent quality and had 0 fails, with overwritten material profiles ofc. I'm not sure bridging is tested well, most of our prints avoids this. If somebody is interested in this we can start a discussion in the forum.

B-Man1:
With the recent uptick in print speed as a measure of quality in a printer, I was surprised that extrusion limit was not already in the printer.cfg file.  Acceleration limits, and speed limits are already there, but when trying to max out the speed of my voron (the whole reason I bought a voron to begin with) I found myself manually having to look through the cura "flow rate" visualization tool and trying to make sure all line types were under the limit I had found for my hotend, going back and forth to fine tune the speeds of all the different line types, playing with line widths, layer heights, layer cooling times etc.  I found myself really wishing that there was a limit that I could set in firmware that would catch and stop flowrates above the max that I had missed when combing through the cura visualization tool.

I also found myself wondering if the cura visualization tool was even accurate.  I don't think cura knows the acceleration limits of my printer for example, so if the visualization tool in cura says the flow rate of a specific line in layer X is Y, but my printer more faster than cura thinks it will, then the flowrate in reality might be higher than what cura predicted.  Hence a firmware limit on flowrate made a lot of sense to me.

With the speed boat challenge, the development of high flow hot ends, high flow nozzles etc, the current trend in 3d printing seems to be speed, and the physical limits of FDM print speed are more and more commonly being due to extrusion flow rate.  Having a flow rate limit value in the printer.cfg file in my opinion would further drive development in 3d printing speed and reliability.

Sineos:
Although I still believe this is better handled in the Slicer, @B-Man1 has a point here. Especially since some slicers, e.g. Cura, are lacking such a feature.

dewi-ny-je:
@B-Man1 whether the visualisation is correct or not and whether the slicer has a flow rate limiting feature or not, the maximum flowrate is simply speed*width*thickness. That is not a particularly strong argument.

Limits in firmware are there to protect the printer itself.
Is the printer going to be damaged if you push too fast filament? I don't think so, since at most it will skip steps or grind filament. Cleaning the gears can indeed be very annoying but it's not a damage.

Also, flow rate limit depends on nozzle size.
It's not really a "property of the printer".

B-Man1:
@Sineos I'm not sure the slicer CAN do this.  Not accurately like the firmware could anyway.  The slicer would need to know how quickly the printer is moving, which it can only estimate without knowing things like acceleration, (which I don't think it does, in the case of cura anyway, which may be why the print time estimates of cura different so greatly from the actual print times of my printer).

@dewi-ny-je I do think that grinding filament could lead to damage of the printer.  It could lead to fillament being stuck in a hot hotend for hours, heat creep, jammed fillament, or burning filament (especially if nozzle temps had been increased in the pursuit of speed).

Not really issues previously but with more and more people pushing the limits of speed with their printers I do think that extrusion rate is an important physical limitation of the hotend (I agree this is more of a hotend property than a printer property, but Klipper does not separate the two)

I would like an extrusion limit implemented from more of a print quality perspective than a "printer protection" one.  When I am trying to print fast,  I try to keep the flowrate at or below the limit because I already know I can't go beyond it and if I do there will be print quality issues if the printer exceeds that limit.  Cura doesn't seem to know how fast my printer is actually moving so it is only able to estimate flowrate pre-print, rather than calculate and limit it mid-print

Sineos:
Prusa Slicer and Super Slicer both have respective options for acceleration, speed etc and are capable of volumetric speed management. Unfortunately Cura does not.

dewi-ny-je:
@Sineos 
> Prusa Slicer and Super Slicer both have respective options for acceleration, speed etc and are capable of volumetric speed management. Unfortunately Cura does not.

In which use case a simple set max speed*line width*layer thickness won't provide you with the max volumetric speed? All of which are set in the slicer.
I think I'm missing a use case here.

Sineos:
My view:
* Recent printers are capable of printing well beyond 100 mm/s
* The limiting factor on such printers is the volumetric flow rate of the hotend
* If the slicer does not support volumetric speed management, you have to manually make sure that you do not exceed the hotend's capability
* This is especially interesting for example, when you want to print your perimeters at slower speed but crank up the speed for the infill
* Of course you can do the calculation manually and put it into the slicer but as soon as you modify either layer height, extrusion width or speed, you will have to recheck 

So it can be viewed as a "comfort feature" as well as a sanity check to avoid failed prints. Still the firmware is not the best place IMO since only the slicer can account for different feature types, like infill, overhangs, bridges etc.

The argument about not having to reslice models when using different printers / material types is something I personally do not care about (same as hardware retraction). May be interesting for people running printer farms but typically they are commercial and so I care even less in the context of an open source project (unless they pay / contribute to the project).

If the feature is working and does not create unexpected effects or a maintenance burden, so be it, otherwise I do not have any special feelings about it.

dewi-ny-je:
@Sineos 

> * Of course you can do the calculation manually and put it into the slicer but as soon as you modify either layer height, extrusion width or speed, you will have to recheck
> 
> So it can be viewed as a "comfort feature" as well as a sanity check to avoid failed prints. Still the firmware is not the best place IMO since only the slicer can account for different feature types, like infill, overhangs, bridges etc.

Also, some (minor) adaptation of the slicer is needed anyway, since max flow depends on nozzle therefore the slicer will need to have profiles setting the max flow rate in the slicer according to the nozzle chosen.

LucidWolf:
I would like to add my vote to adding this pull request if it can default to off similar to pressure advance and shaping.  I am a new user to Klipper and trying to maximize my flow rates and have made mistakes increasing layer width while not reducing speed and exceeding my flow rate of the hot end.  I know i should just "get gud" at my slicer settings, but it is nice to have a firmware safety rail.  Since this failure can occur later in the print due to a geometric feature and flow rate exceed that makes a nice feature to derail the hotend at a fast travel far into what looks like a successful print.  

I do have concern with limiting things in firmware because of the mentioned above issues with debugging.  Very hard to tell what happened when what is purely commanded by G-code is different than actual.  This leads to my next point about a job log.  

This might be better for another topic as Klipper adds more features similar to the pressure advance and input shaping.  It would be nice to have a planner summary letting me know the safety activated and I "should feel bad".  I think this sort of "warning log" for a print would be a nice feature in general where Klipper planner who is much better at knowing actual with shaping and pressure advance could dump things like extrusion rate limited due to max velocity, commanded acceleration over limits, ... etc.  This could also be a place you squirrel away changes to flow rate baby stepping and such as the print is running and the user is changing parameters.   Care will have to be taken to not overload this log file.  I would love to have an actual speed summary after a print.  Similar to what i can see in Fluidd while it is printing all for debugging my print quality.  Right now i am kind of sudo using that for tuning while I print but its not very accurate especially with my retraction/prime being counted in the flow rate.

napter:
> @Sineos
> 
> > Prusa Slicer and Super Slicer both have respective options for acceleration, speed etc and are capable of volumetric speed management. Unfortunately Cura does not.
> 
> In which use case a simple set max speed * line width * layer thickness won't provide you with the max volumetric speed? All of which are set in the slicer. I think I'm missing a use case here.


Not all lines are printed at the same thickness - and now with Arachne as the default engine in Cura and PrusaSlicer, line thickn ess can vary quite a lot throughout the print. A max speed wouldn't help because you need the thick lines to print slowly (they will max out the volumetric rate at a lower speed) and the thin lines to print quickly. If you set your speed limit to the speed that your thickest line can be printed that will significantly slow down the print. Here is a heat map of volumetric flow when using volumetric flow instead of max speed to determine the speed, as you can see it is nearly all the same flow rate:
image

Here is the same print with a speed heat map:
image

The print speed varies by quite a lot depending on the line thickness. If you were to use the max speed of the slowest line this print would take significantly longer. 

It is possible to set the max volumetric rate PrusaSlicer but I agree it would be nice if it could be adjusted with gcode at the start of or during a print.

B-Man1:
How do I add this to my klipper Installation?  I am constantly finding myself having to reduce the federate of my printer as it prints due to exceeding the flow rate limit of my print head.  I have to do this manually by reducing the overall federate, which reduces the speed of everything, walls, infill, top and bottom surfaces etc.  Typically it is only the infill that needs to be reduced but I have no way to do that. All I can do is reduce everything, which sucks. I would love to be able to just a flow rate limit while a print is printing and have klipper automatically adjust speed to make sure it is extruding below that limit.  This would be a huge efficiency and print reliability improvement

masterxq:
@B-Man1 sorry for late reply.

https://klipper.discourse.group/t/feature-request-volumetric-flow-rate-limit/2739/2

TommyC81:
I find this interesting and useful, and was looking for this option in the documentation without finding it. I'm surprised it's not already part of Klipper and how this can be developed further and implemented as an optional setting.

hg42:
I think many speed settings should in reality be flow settings.

Speed was only used as a limit, because it is the parameter controlled by gcode.
The relation between flow and speed was constant, because nozzle size, width and height were constant.

Speed itself was a limiting factor with 8bit MCUs because of the step rate limit they could produce and maybe because of the motors.

Today we have a much different situation:

* width is variable (e.g. Archne)
* height may be variable (e.g. infill every second layer)
* all components can go really fast (500mm/s with standard components is no problem)
* extruders are light weight
* input shaper etc. allow to print faster without loosing quality
* heaters have more power (e.g. 115W is cheap enough)
* nozzle size may be bigger (0.6 instead of 0.4 or even bigger nozzles for big parts) so the factor flow/speed is bigger
etc. pp.

I think it's a fact, that the flow rate is the main limiting factor today for the maximum speed and acceleration for xy-movement.
Speed itself is far below it's limits.

Consequently, Slicers like Prusa already have changed their strategies. They allow setting most speeds to auto and set max speed and max volumetric flow (and max speed can be very high and then simply ignored, because the flow is the real limit).
This simplifies a lot, because you can set one parameter and get them all.
The only exceptions are speeds that are lower to create better quality, e.g. for bridges, overhangs, outer perimeters.
I think these are also flow limits in reality.
One exception would be the first layer speed, I think that's not a flow problem, but more about acceleration in curves.

All arguments that were given against limiting the flow (debugging, unexpected behavior because of the limit) also apply to a speed limit (in fact it was the same thing when width and height were constant). No one has ever questioned the speed limit.

I also agree with @B-Man1 that the slicer doesn't have the whole physical picture, it only creates a plan. The printer decides what is possible. That's why it has acceleration limits, and it can also do other things like input shaping, 6th order calculations. So the final flow is only known by the printer (or firmware). I would like to set my slicers to the maximum values corresponding with model features and let the printer limit the physical parameters.

Also, we have tool changers. If changing a tool, certain limits change. These can be applied automatically when changing the tool.
The same applies to changing materials.
With big displays and comfortable GUIs, I don't even need a PC. I have an sdcard (or internal storage) with lots of models, I just choose one, then a material and a tool and then start the print. Why should I need to power on a PC and slice again only because a material or tool has different flow limits?

Some arguments against this PR are about quality.
Quality limitations related to speeds are usually far away from the flow limit. E.g. outer perimeters or first layers, bridges etc. are configured with low speeds. They don't even come near the flow limit. Or if they do, the limitation only leads to a better quality, becasue it's always a slow down.

AtmacOfficial:
> Common slicers already implement max flow rate code. I'd like to see feedback from Klipper users that they are accomplishing great things with a Klipper implementation of max flow rate.

Certain slicers have issues when it come to multi material printing and how they emit E-steps. Specifically for Single Extruder Multi Material, Slic3r (Prusa/Super) will ignore volumetric flow rates during loading moves on a filament swap/prime tower. This is normally covered by the machine limits as specified in the slicer, however the slicer will not emit those with the Klipper Flavor of Gcode. It becomes an issue were neither side (the slicer, nor the printer) is putting a check on what the Gcode is actually telling the printer to do. 

It should also be noted that for Single Extruder Multi Material, max_extrude_cross_section has to be effectively disabled. For the same instances that would make max_extruder_velocity useful, max_extrude_cross_section just causes an error and the print to fail. 
(and because neither side (the slicer, nor the printer) is putting a check on the Esteps, there is no way around this without post processing the Gcode doing effectively what max_extruder_velocity would do anyways)

PR #5457, opened 2022-04-25
Kconfig: Add 64kib bootloader offset option to STM32F401

JamesH1978:
This is needed for the Creality Ender 3 S1 with the STM32F401 chips to enable a 64kib bootloader offset


KevinOConnor:
Thanks.

-Kevin

PR #5460, opened 2022-04-26
config: Pull up endstop_pins on Fysetc S6/Spider

arkeet:
My Fysetc S6 V2.1 (essentially the same as V2) needed software pull-ups to make the endstops work. After checking the schematics for S6 V1.2 and Spider, those boards need software pull-ups as well.


Sineos (approved):
Indeed, neither hardware pull-ups nor any filtering caps on the endstop pins. Kind of a poor design decision.

KevinOConnor:
Thanks.

-Kevin

PR #5468, opened 2022-04-30
manual_stepper: Fix long moves

cmroche:
Description of the problem can be found in this thread: https://klipper.discourse.group/t/long-manual-stepper-moves-cause-timer-too-close/2740

Introduces better support for long moves on slow host processors which may otherwise generate a timer too close error when step generation takes too long.

Addresses a problem coming up in the ERCF project somewhat frequently https://github.com/EtteGit/EnragedRabbitProject/issues/75

Tests are passing

### DIFF
--- a/klippy/extras/manual_stepper.py
+++ b/klippy/extras/manual_stepper.py
@@ -6,9 +6,14 @@
 import stepper, chelper
 from . import force_move
 
+MOVE_BATCH_TIME = 0.500
+FLUSH_DELAY = 0.001
 class ManualStepper:
     def __init__(self, config):
         self.printer = config.get_printer()
+        self.reactor = self.printer.get_reactor()
+        self.all_mcus = [
+            m for n, m in self.printer.lookup_objects(module='mcu')]
         if config.get('endstop_pin', None) is not None:
             self.can_home = True
             self.rail = stepper.PrinterRail(
@@ -20,6 +25,8 @@ class ManualStepper:
             self.steppers = [self.rail]
         self.velocity = config.getfloat('velocity', 5., above=0.)
         self.accel = self.homing_accel = config.getfloat('accel', 0., minval=0.)
+        self.move_flush_time = config.getfloat(
+            'move_flush_time', 0.050, above=0.)
         self.next_cmd_time = 0.
         # Setup iterative solver
         ffi_main, ffi_lib = chelper.get_ffi()
@@ -56,6 +63,8 @@ class ManualStepper:
     def do_set_position(self, setpos):
         self.rail.set_position([setpos, 0., 0.])
     def do_move(self, movepos, speed, accel, sync=True):
+        toolhead = self.printer.lookup_object('toolhead')
+        batch_time = MOVE_BATCH_TIME
         self.sync_print_time()
         cp = self.rail.get_commanded_position()
         dist = movepos - cp
@@ -65,11 +74,23 @@ class ManualStepper:
                           accel_t, cruise_t, accel_t,
                           cp, 0., 0., axis_r, 0., 0.,
                           0., cruise_v, accel)
-        self.next_cmd_time = self.next_cmd_time + accel_t + cruise_t + accel_t
-        self.rail.generate_steps(self.next_cmd_time)
-        self.trapq_finalize_moves(self.trapq, self.next_cmd_time + 99999.9)
-        toolhead = self.printer.lookup_object('toolhead')
-        toolhead.note_kinematic_activity(self.next_cmd_time)
+        next_print_time = self.next_cmd_time + accel_t + cruise_t + accel_t
+        while 1:
+            segment_time = self.next_cmd_time + batch_time
+            self.next_cmd_time = min(segment_time, next_print_time)
+            self.rail.generate_steps(self.next_cmd_time)
+            finalize_time = self.next_cmd_time - FLUSH_DELAY
+            if self.next_cmd_time >= next_print_time:
+                finalize_time = self.reactor.NEVER
+            self.trapq_finalize_moves(self.trapq, finalize_time)
+            toolhead.note_kinematic_activity(self.next_cmd_time)
+            for m in self.all_mcus:
+                flush_time = self.next_cmd_time
+                if self.next_cmd_time < next_print_time:
+                    flush_time -= self.move_flush_time
+                m.flush_moves(flush_time)
+            if self.next_cmd_time >= next_print_time:
+                break
         if sync:
             self.sync_print_time()
     def do_homing_move(self, movepos, speed, accel, triggered, check_trigger):
--- a/klippy/extras/manual_stepper.py
+++ b/klippy/extras/manual_stepper.py
@@ -84,8 +84,8 @@ class ManualStepper:
                 finalize_time = self.reactor.NEVER
             self.trapq_finalize_moves(self.trapq, finalize_time)
             toolhead.note_kinematic_activity(self.next_cmd_time)
+            flush_time = self.next_cmd_time
             for m in self.all_mcus:
-                flush_time = self.next_cmd_time
                 if self.next_cmd_time < next_print_time:
                     flush_time -= self.move_flush_time
                 m.flush_moves(flush_time)


klippy/extras/manual_stepper.py
```diff
@@ -65,11 +74,23 @@ def do_move(self, movepos, speed, accel, sync=True):
...
+            self.rail.generate_steps(self.next_cmd_time)
+            finalize_time = self.next_cmd_time - FLUSH_DELAY
+            if self.next_cmd_time >= next_print_time:
+                finalize_time = self.reactor.NEVER
+            self.trapq_finalize_moves(self.trapq, finalize_time)
+            toolhead.note_kinematic_activity(self.next_cmd_time)
+            for m in self.all_mcus:
+                flush_time = self.next_cmd_time
+                if self.next_cmd_time < next_print_time:
+                    flush_time -= self.move_flush_time
```
dmbutyugin, line 90: It might make sense to move flush_time calculation outside of 'for m in ...' loop, as it does not depend on the mcu. So it might be a bit easier to understand in the future.

dmbutyugin (commented):
Thank you!

I have a couple of high-level questions on how the new code works:

- Does it work with `STOP_ON_ENDSTOP=...` parameter if `MOVE=` is set fairly large and the endstop is triggered early? E.g. `MANUAL_STEPPER STEPPER=my_stepper SET_POSITION=0 SPEED=10 MOVE=100 STOP_ON_ENDSTOP=1` and the endstop is triggered at 20 mm.
- On a slow machine, the current code may generate a 'timer too close' error, as you reported. How does your new code work in a similar situation if `MANUAL_STEPPER` command is executed with `SYNC=0` and there are subsequent G1/G0 toolhead moves that can run in parallel? Is it possible that there will be a similar 'timer too close' error when processing these subsequent G0/G1 commands? E.g.
```
G28
MANUAL_STEPPER STEPPER=my_stepper SET_POSITION=0 SPEED=10 MOVE=1000 SYNC=0
G0 X100 Y100 Z20 F600
```

cmroche:
> * Does it work with `STOP_ON_ENDSTOP=...` parameter if `MOVE=` is set fairly large and the endstop is triggered early? E.g. `MANUAL_STEPPER STEPPER=my_stepper SET_POSITION=0 SPEED=10 MOVE=100 STOP_ON_ENDSTOP=1` and the endstop is triggered at 20 mm.

When `STOP_ON_ENDSTOP=1` is specified a homing move is used which is instead run through **toolhead.py**. This currently causes a timer issue unfortunately, I'll look into applying the same pattern there as a fix as well.

> * On a slow machine, the current code may generate a 'timer too close' error, as you reported. How does your new code work in a similar situation if `MANUAL_STEPPER` command is executed with `SYNC=0` and there are subsequent G1/G0 toolhead moves that can run in parallel? Is it possible that there will be a similar 'timer too close' error when processing these subsequent G0/G1 commands? E.g.
> 
> ```
> G28
> MANUAL_STEPPER STEPPER=my_stepper SET_POSITION=0 SPEED=10 MOVE=1000 SYNC=0
> G0 X100 Y100 Z20 F600
> ```

I tested, this works without any issue.

klippy/extras/manual_stepper.py
dmbutyugin: But I'm afraid that now the code is incorrect. The whole block of `flush_time` calculation should have been moved outside the loop, otherwise the code
```
                if self.next_cmd_time < next_print_time:
                    flush_time -= self.move_flush_time
```
will be executed for each mcu, and so flush_time will be decreasing for each mcu.

dmbutyugin:
> > * Does it work with `STOP_ON_ENDSTOP=...` parameter if `MOVE=` is set fairly large and the endstop is triggered early? E.g. `MANUAL_STEPPER STEPPER=my_stepper SET_POSITION=0 SPEED=10 MOVE=100 STOP_ON_ENDSTOP=1` and the endstop is triggered at 20 mm.
> 
> I tested, this generated "Timer rescheduled in the past." I'll investigate if it's a similar problem to what the change solves.

Note that the homing move is generated by `ManualStepper.drip_move`, which in turn just calls `do_move`. The error might have something to do with the endstop triggering, but the stepper commands being generated further regardless of this. Actually, previously I discussed with Kevin how to implement the manual stepper homing such that it detects endstop trigger and does not wait until the full homing move 'completes' to continue GCode execution. The suggestion was to split the homing move generation into small batches and use `drip_completion` object to test whether the endstop triggered (similar to how Toolhead.drip_move works). Worst case, you may need to implement something like that to fix the error you observe (in fact, you have implemented most of it already). But maybe the problem lies elsewhere.


> > * On a slow machine, the current code may generate a 'timer too close' error, as you reported. How does your new code work in a similar situation if `MANUAL_STEPPER` command is executed with `SYNC=0` and there are subsequent G1/G0 toolhead moves that can run in parallel? Is it possible that there will be a similar 'timer too close' error when processing these subsequent G0/G1 commands? E.g.
> > 
> > ```
> > G28
> > MANUAL_STEPPER STEPPER=my_stepper SET_POSITION=0 SPEED=10 MOVE=1000 SYNC=0
> > G0 X100 Y100 Z20 F600
> > ```
> 
> I tested, this works without any issue.

Thanks! Can you explain, how does `G0 ...` work with your modified code in your test? Is it scheduled to start later, not together with the beginning of `MANUAL_STEPPER` command? The reason I am asking is because I generated the debug serial output with microcontroller commands for this sequence of commands
```
SET_KINEMATIC_POSITION X=0 Y=0 Z=10
MANUAL_STEPPER STEPPER=ms ENABLE=1 SET_POSITION=0 SPEED=10 ACCEL=100 MOVE=100 SYNC=1
G0 X100 F600
```
 with the mainline Klipper and your changes, and the diffs are:
```diff
--- test_manual_move_baseline.txt
+++ test_manual_move_modified.txt
@@ -29,17 +29,42 @@
 queue_digital_out oid=13 clock=4000000 on_ticks=1
 set_next_step_dir oid=0 dir=1
 queue_step oid=0 interval=4320000 count=1 add=0
-set_next_step_dir oid=1 dir=1
-queue_step oid=1 interval=4032660 count=1 add=0
-queue_digital_out oid=14 clock=4000000 on_ticks=0
-queue_step oid=1 interval=23757 count=2 add=-3507
-queue_step oid=1 interval=20000 count=394 add=0
 queue_step oid=0 interval=234256 count=1 add=0
 queue_step oid=0 interval=161036 count=2 add=-29688
 queue_step oid=0 interval=113110 count=2 add=-11540
 queue_step oid=0 interval=92085 count=3 add=-6136
 queue_step oid=0 interval=75281 count=4 add=-3448
-queue_step oid=0 interval=64000 count=2475 add=0
+queue_step oid=0 interval=64004 count=99 add=0
+queue_step oid=0 interval=64000 count=125 add=0
+queue_step oid=0 interval=64000 count=125 add=0
......................................................................................
+queue_step oid=0 interval=64000 count=125 add=0
+queue_step oid=0 interval=64000 count=125 add=0
+queue_step oid=0 interval=63604 count=3 add=2127
+queue_step oid=0 interval=71710 count=3 add=3994
+queue_step oid=0 interval=85670 count=2 add=7036
+queue_step oid=0 interval=101070 count=2 add=12540
+queue_step oid=0 interval=130848 count=2 add=30688
+queue_step oid=0 interval=234256 count=1 add=0
+queue_digital_out oid=14 clock=4000000 on_ticks=0
+set_next_step_dir oid=1 dir=1
+queue_step oid=1 interval=4032660 count=1 add=0
+queue_step oid=1 interval=23757 count=2 add=-3507
+queue_step oid=1 interval=20000 count=394 add=0
 queue_step oid=1 interval=20000 count=400 add=0
 queue_step oid=1 interval=20000 count=400 add=0
 queue_step oid=1 interval=20000 count=400 add=0
@@ -61,11 +86,5 @@
 queue_step oid=1 interval=20000 count=400 add=0
 queue_step oid=1 interval=20000 count=2 add=0
 queue_step oid=1 interval=24006 count=1 add=0
-queue_step oid=0 interval=65348 count=3 add=3168
-queue_step oid=0 interval=75205 count=3 add=5084
-queue_step oid=0 interval=92412 count=2 add=9158
-queue_step oid=0 interval=113110 count=2 add=18238
-queue_step oid=0 interval=161286 count=1 add=0
-queue_step oid=0 interval=234256 count=1 add=0
 queue_digital_out oid=14 clock=165685333 on_ticks=1
 queue_digital_out oid=13 clock=165685333 on_ticks=0
```
Note that oid=14 and oid=1 are for the X stepper, and oid=13 and oid=0 - for the manual stepper. Note that the modified code first generates full step output for the manual stepper (oid=0) and only then generates the steps for X stepper (oid=1). But the start point for the X steps is the same `queue_step oid=1 interval=4032660 count=1 add=0`. So, I was worried a bit that if the generation of all steps for the manual stepper takes too much time, then the MCU may error on the first X step with 'timer in the past' error. However, if Klipper adjusts the start of X steps to the next feasible time, it should be fine.

cmroche:
> Thanks! Can you explain, how does G0 ... work with your modified code in your test? Is it scheduled to start later, not together with the beginning of MANUAL_STEPPER command? The reason I am asking is because I generated the debug serial output with microcontroller commands for this sequence of commands

You are absolutely right it doesn't, and I wasn't paying attention to the move sequence of the head when I was testing! Of course it doesn't work, the new scheduling will send movement commands to the MCU before the `G0` is processed. I'll have to rethink this problem, I'm not sure how to tackle it.

dmbutyugin:
Consider also the following snippet (change against the master branch), though it very different from your current implementation:
```diff
diff --git a/klippy/extras/manual_stepper.py b/klippy/extras/manual_stepper.py
index 223e13f1..1bbb7283 100644
--- a/klippy/extras/manual_stepper.py
+++ b/klippy/extras/manual_stepper.py
@@ -6,6 +6,8 @@
 import stepper, chelper
 from . import force_move
 
+FLUSH_DELAY = 0.001
+
 class ManualStepper:
     def __init__(self, config):
         self.printer = config.get_printer()
@@ -34,6 +36,10 @@ class ManualStepper:
         gcode.register_mux_command('MANUAL_STEPPER', "STEPPER",
                                    stepper_name, self.cmd_MANUAL_STEPPER,
                                    desc=self.cmd_MANUAL_STEPPER_help)
+        self.printer.register_event_handler("klippy:connect", self.connect)
+    def connect(self):
+        toolhead = self.printer.lookup_object('toolhead')
+        toolhead.register_step_generator(self.rail.generate_steps)
     def sync_print_time(self):
         toolhead = self.printer.lookup_object('toolhead')
         print_time = toolhead.get_last_move_time()
@@ -57,6 +63,8 @@ class ManualStepper:
         self.rail.set_position([setpos, 0., 0.])
     def do_move(self, movepos, speed, accel, sync=True):
         self.sync_print_time()
+        finalize_time = self.next_cmd_time - FLUSH_DELAY
+        self.trapq_finalize_moves(self.trapq, finalize_time)
         cp = self.rail.get_commanded_position()
         dist = movepos - cp
         axis_r, accel_t, cruise_t, cruise_v = force_move.calc_move_time(
@@ -66,8 +74,6 @@ class ManualStepper:
                           cp, 0., 0., axis_r, 0., 0.,
                           0., cruise_v, accel)
         self.next_cmd_time = self.next_cmd_time + accel_t + cruise_t + accel_t
-        self.rail.generate_steps(self.next_cmd_time)
-        self.trapq_finalize_moves(self.trapq, self.next_cmd_time + 99999.9)
         toolhead = self.printer.lookup_object('toolhead')
         toolhead.note_kinematic_activity(self.next_cmd_time)
         if sync:
diff --git a/klippy/toolhead.py b/klippy/toolhead.py
index b7285420..b33283cc 100644
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -342,8 +342,8 @@ class ToolHead:
         self.idle_flush_print_time = 0.
         flush_time = self.last_kin_move_time + self.kin_flush_delay
         flush_time = max(flush_time, self.print_time - self.kin_flush_delay)
+        self._update_move_time(max(self.print_time, self.last_kin_flush_time, flush_time))
         self.last_kin_flush_time = max(self.last_kin_flush_time, flush_time)
-        self._update_move_time(max(self.print_time, self.last_kin_flush_time))
     def _flush_lookahead(self):
         if self.special_queuing_state:
             return self.flush_step_generation()
```

TBH, the thing I'm very uncertain about is the suggested change in the `toolhead.py`. I'm not sure if moving `self.last_kin_flush_time` after the `_update_move_time` call is correct or if it will break something else.

KevinOConnor:
Thanks for looking at this.  I agree the manual_stepper code should be improved so that it does not raise errors on long moves.

Alas, this is a tricky part of the code.  (Likely an area of the Klipper code that could be improved.)  I fear the changes on this PR could cause other errors.  Specifically, the call of `m.flush_moves(flush_time)` from manual_stepper.py is problematic because that call was designed to always be passed an incrementing time - calling it from both toolhead.py and manual_stepper.py makes it possible for the time to "go backwards".  Also, I'd say this low-level code would probably need a refactoring, as duplicating it in multiple locations is likely to lead to a maintenance burden.

At the root of the issue is that the stepper code (and more generally, anything using the move_queue - see #4128) needs explicit flush handling.  Currently that is done in toolhead.py, but that code arrangement makes it difficult to schedule stepper actions asynchronously from commanded movement on the toolhead.

Alas, I don't have good suggestions for improving this.  It is something I've thought about in the past, but any possible solutions I've come up with require significant code refactoring.

-Kevin

dmbutyugin:
@KevinOConnor Thanks for also taking a look at this!

> Currently that is done in toolhead.py, but that code arrangement makes it difficult to schedule stepper actions asynchronously from commanded movement on the toolhead.

So the snippet I've posted (which should not be considered a final working suggestion, but more like a code in a proof of concept stage) tries to address the difficulty of having to flush a separate move queue for the manual stepper by deferring this action to the toolhead itself (by registering the manual stepper rail as a step generator, similar to how extruder steppers do it). Naturally, it may have some defects and bugs. In fact, I've found a couple of small issues already, and so I pushed my code to a separate [branch](https://github.com/dmbutyugin/klipper/tree/manual-stepper-long-moves) for the ease of tracking. But in general, that seems to achieve the desired effect (that Klipper can generate the steppers steps for `MANUAL_STEPPER ... SYNC=0` and subsequent `G0`/`G1` commands in parallel and in batches, as well as a stand-alone `MANUAL_STEPPER` command):
```
SET_KINEMATIC_POSITION X=0 Y=0 Z=10
MANUAL_STEPPER STEPPER=ms ENABLE=1 SET_POSITION=0 SPEED=10 ACCEL=100 MOVE=200 SYNC=0
G0 X100 F600
```
generates
```
queue_step oid=0 interval=4320000 count=1 add=0
queue_step oid=1 interval=4032660 count=1 add=0
queue_step oid=1 interval=23757 count=2 add=-3507
queue_step oid=1 interval=20000 count=394 add=0
queue_step oid=0 interval=234256 count=1 add=0
queue_step oid=0 interval=161036 count=2 add=-29688
queue_step oid=0 interval=113110 count=2 add=-11540
queue_step oid=0 interval=92085 count=3 add=-6136
queue_step oid=0 interval=75281 count=4 add=-3448
queue_step oid=0 interval=64004 count=99 add=0
queue_step oid=0 interval=64000 count=125 add=0
queue_step oid=1 interval=20000 count=400 add=0
queue_step oid=0 interval=64000 count=125 add=0
queue_step oid=1 interval=20000 count=400 add=0
....................................................................................................
queue_step oid=1 interval=20000 count=400 add=0
queue_step oid=0 interval=64000 count=126 add=0
queue_step oid=1 interval=20000 count=2 add=0
queue_step oid=1 interval=24006 count=1 add=0
queue_step oid=0 interval=64000 count=125 add=0
queue_step oid=0 interval=64000 count=125 add=0
....................................................................................................
queue_step oid=0 interval=64000 count=125 add=0
queue_step oid=0 interval=64977 count=4 add=3407
queue_step oid=0 interval=79896 count=3 add=6082
queue_step oid=0 interval=101467 count=2 add=12143
queue_step oid=0 interval=130848 count=2 add=30688
queue_step oid=0 interval=234256 count=1 add=0
```

and just `MANUAL_STEPPER STEPPER=ms ENABLE=1 SET_POSITION=0 SPEED=10 ACCEL=100 MOVE=200 SYNC=0` generates
```
queue_step oid=0 interval=4320000 count=1 add=0
queue_step oid=0 interval=234256 count=1 add=0
queue_step oid=0 interval=161036 count=2 add=-29688
queue_step oid=0 interval=113110 count=2 add=-11540
queue_step oid=0 interval=92085 count=3 add=-6136
queue_step oid=0 interval=75281 count=4 add=-3448
queue_step oid=0 interval=64004 count=99 add=0
queue_step oid=0 interval=64000 count=125 add=0
queue_step oid=0 interval=64000 count=125 add=0
....................................................................................................
queue_step oid=0 interval=64000 count=125 add=0
queue_step oid=0 interval=64000 count=125 add=0
queue_step oid=0 interval=63604 count=3 add=2127
queue_step oid=0 interval=71710 count=3 add=3994
queue_step oid=0 interval=85670 count=2 add=7036
queue_step oid=0 interval=101070 count=2 add=12540
queue_step oid=0 interval=130848 count=2 add=30688
queue_step oid=0 interval=234256 count=1 add=0

```


> FWIW, moving `_update_move_time()` seems suspect to me. The intent of setting `last_kin_flush_time` prior to the call is to force a behavour change in `_update_move_time()`.

I agree this may be problematic in other regards. However, this change is not part of the original PR, so it cannot cause issues there.

This is a part of my alternative proposed changes that I've added in order to fix a simple case of a single command `MANUAL_STEPPER`: without this change, the code in the `flush_step_generation` inadvertently advances `last_kin_flush_time` which then 'breaks' the loop in the `_update_move_time` and makes it generate all steps for the manual stepper in one go instead of processing it in batches `MOVE_BATCH_TIME`.

cmroche:
This clearly needs more though and work, I'll go through the feedback and come back to the problem in the next week or two. Thanks.

cmroche:
> FWIW, moving `_update_move_time()` seems suspect to me. The intent of setting `last_kin_flush_time` prior to the call is to force a behavour change in `_update_move_time()`.

What's the reason for that behaviour change? Wouldn't this be introducing the same problem in all places that we're trying to fix here if flushing a long remaining move on the toolhead?

KevinOConnor:
> What's the reason for that behaviour change?

If the time isn't updated then it is possible that moves buffered elsewhere wont get properly flushed.

It may be possible to incrementally flush the queue and then update the final timestamp.  I'm not sure.

-Kevin

PR #5469, opened 2022-04-30
Add support to different flash size and ram size of STM32F103

secmaster:
As STM32F103 series have different setup for flash size and ram size, update the Kconfig to support it.
Reference to https://www.st.com/en/microcontrollers-microprocessors/stm32f103.html

Signed-off-by: Chun Yu([secmaster@gmail.com](mailto:secmaster@gmail.com))


KevinOConnor:
Thanks.  It's not clear to me why a user would want to specify the f103 series type.  I'm concerned that most users would not know the series type and the question could lead to confusion.  Klipper never uses more than about 16KiB of ram, so defining larger memory models doesn't provide any benefit, and similarly Klipper doesn't fit in small flash models so there is no value in defining them.  Maybe I've missed something?

Separately, commits need to have a signed-off-by line - see https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

secmaster:
Hi Kelvin
First of all thanks for your comments, actually this is my first time to create pull request and i am sorry that i didn't foolow the commit requirements, let me see how to fix it

As per your question, stm32f103 is a common mcu (it is being called as blue pill) and its has many different sub series with different flash size and ram size
e.g i bought the one from china is stm32f103c6t6 only got 32k flash and 10k memory. but the original menuconfig only support the most common stm32f103c8t6 ie 64k flash and 20k ram. Thus, i failed to compile the klipper firmware with it. and therefore i modified the Kconfig to support different sub series of stm32f103 with different flash and ram side

The submenu item for stm32f103 will only appear when users select it, it should not impact the selection for other stm32 menus
Thanks

KevinOConnor:
Thanks.  I fear the sub-menu will lead to confusion as most users wont know what sub-type they have.  Instead, feel free to add a new stm32 "processor model" with a description like "stm32f103ct86 (rare stm32f103 with only 6K of ram)".

Cheers,
-Kevin

secmaster:
In the past, when we purchase the Stm32 blue pill, the majority will be c8t6 series, but due the shortage of MCUs chips, their price became double, and the seller now promote the c6t6 series with cheaper price
When I key in “Stm32 blue pill” in Amazon, it will show both the c8t6 and c6t6, so, people like me may not aware the difference and buy the wrong one which cannot compile due to flash/Ram size issue

https://www.amazon.com/s?k=stm32+blue+pill

jevermeister:
> Hi Kelvin First of all thanks for your comments, actually this is my first time to create pull request and i am sorry that i didn't foolow the commit requirements, let me see how to fix it
> 
> As per your question, stm32f103 is a common mcu (it is being called as blue pill) and its has many different sub series with different flash size and ram size e.g i bought the one from china is stm32f103c6t6 only got 32k flash and 10k memory. but the original menuconfig only support the most common stm32f103c8t6 ie 64k flash and 20k ram. Thus, i failed to compile the klipper firmware with it. and therefore i modified the Kconfig to support different sub series of stm32f103 with different flash and ram side
> 
> The submenu item for stm32f103 will only appear when users select it, it should not impact the selection for other stm32 menus Thanks

Just stumbled upon this. So with your changes Klipper will run on the F103C6T6 blue pill with 32k flash and 10k RAM?

secmaster:
> 

Yes, pls give it a try and let me know the result

KevinOConnor:
Thanks.  I added a Kconfig option to limit the build to just 10KiB of ram on the stm32f103 (cf9d96434).

-Kevin

PR #5470, opened 2022-04-30
klippy: Reworked the MCU protocol error message

Drachenkaetzchen:
The MCU protocol error message often confuses users, especially after getting bombarded with the config format error line. This PR tries to improve it and has the following goals:

- Put the technical error at the end to prevent confusion and avoid the immediate jump to help channels instead of continuing to read
- Inform the user first what the type of error is (In this case: MCU Protocol error) and not at the end of the long error message
- Give the users a clear instruction what to do (pretty much unchanged), but possibly more approachable because the user is not confused by the technical error anymore
- Shows the version numbers of Klipper separate from all MCU version numbers so that it is more obvious that this is the version of Klipper
- Separates the version numbers between outdated MCUs and up-to-date MCUs, so that it becomes obvious if a MCU flash failed because the MCU doesn't disappear from the outdated list
- Previously, when viewing the errors in a web frontend, additional line breaks made it extremely hard to read the message, sometimes with as little as one word on one line. The error messages are now formatted to not have rogue line breaks, improving readability.

Signed-off-by: Felicia Alexa Hummel <felicia@drachenkatze.org>

### DIFF
--- a/klippy/klippy.py
+++ b/klippy/klippy.py
@@ -145,16 +145,8 @@ class Printer:
             m.add_printer_objects(config)
         # Validate that there are no undefined parameters in the config file
         pconfig.check_unused_options(config)
-    def _get_versions(self):
-        try:
-            parts = ["%s=%s" % (n.split()[-1], m.get_status()['mcu_version'])
-                     for n, m in self.lookup_objects('mcu')]
-            parts.insert(0, "host=%s" % (self.start_args['software_version']))
-            return "\nKnown versions: %s\n" % (", ".join(parts),)
-        except:
-            logging.exception("Error in _get_versions()")
-            return ""
-    def _get_outdated_versions(self):
+
+    def _build_protocol_error_message(self):
         try:
             host_version = self.start_args['software_version']
 
@@ -200,7 +192,7 @@ class Printer:
             msg = [message_protocol_error1,
                    "",
                    ' '.join(message_protocol_error2.splitlines())[1:],
-                   self._get_outdated_versions(),
+                   self._build_protocol_error_message(),
                    ' '.join(message_protocol_error3.splitlines())[1:],
                    "",
                    str(e)]
--- a/klippy/klippy.py
+++ b/klippy/klippy.py
@@ -182,7 +182,7 @@ class Printer:
                ' '.join(message_protocol_error3.splitlines())[1:],
                "",
                str(e)]
-        
+
         return "\n".join(msg)
 
     def _connect(self, eventtime):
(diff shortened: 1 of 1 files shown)


Drachenkaetzchen:
Background:

I'm Drachenkatze, a very active supporter on the Voron Discord. People approach us because they receive the protocol error message. Often they're overwhelmed by the technical information on the first line, preventing them from continuing reading. Other times, they're puzzled why they're still receiving the error despite updating their MCU boards - without realizing that the flash process did not work, or that they did update the wrong MCU.

This PR aims to address that, making the error more approachable by end-users and in turn reducing the supporter workload quite a bit.

I presented this in form of a draft to the #klipper-development a few months back, with some positive responses but also some "people just don't read" responses. I disagree with the latter - if we make it harder than it needs to be, yes. But if you argue like that then we could also throw E_PROTOCOL_ERROR without any further info at them.

I also interviewed quite many users what prevented them from interpreting the current error message, and I incorporated the results into this PR.

Note that many Klipper users are not tech-savvy. They're often first-time Linux users because of Klipper. Let's not make it harder for them than it needs to be.

Here are the old and new error messages as they would appear in Mainsail:
oldmessage
newmessage

KevinOConnor:
Thanks.  We can certainly change the mcu protocol error message, and I agree the proposed format is an improvement.

I have a few minor comments on the code though:
1. After this change, `_get_versions()` is no longer used and should be removed.
2. The try/exclude block should be limited to just the `mcu.get_status()` call, and not applied to the entire `_get_outdated_versions()` function (otherwise, it can cause confusing results if real code errors are suppressed).  For example:
```python
            for name, mcu in self.lookup_objects('mcu'):
                try:
                    mcu_version = mcu.get_status()['mcu_version']
                except:
                    logging.exception(...)
                    continue
```
3. It seems a little confusing to me to build the message in both `_get_outdated_versions()` and in the caller - maybe rename the function to `_build_protocol_error()` and build the entire message there?

-Kevin

Drachenkaetzchen:
You're right, I missed to remove `_get_versions()`. I also fixed the other issues. I'm not very happy with the way the string building is done, but right now I have no better idea - and at least it's somewhat readable. 

I took the liberty to use `_build_protocol_error_message` instead of `_build_protocol_error` to make it clear what the method does.

KevinOConnor:
Thanks.  I committed this change along with 7e76bd56 on top.

Note that I restored the previous linebreaks.  The error messages are currently written to have line breaks after 60 characters because otherwise they look really bad on OctoPrint.  I understand they don't render great on Mainsail, but I think we'll need to find a general solution for that.

-Kevin

PR #5471, opened 2022-05-01
Config: Added new configs to reflect newer hardware revisions

NexGen-3D-Printing:
I have added two new configuration files:

1. printer-twotrees-sapphire-pro-sp-3-2022.cfg: Added new config for the latest version of the TwoTrees SP-3 Pro.
The latest printer uses a newer board revision, physically different to previous version, with updated MCU, different driver type, different driver quantity and larger boot offset

- Drivers: 4 x Integrated TCM2225
- MCU: STM32F407
- Boot Offset: 32kb

2. generic-mks-robin-nano-s-v1.3.cfg: Added new config for the latest version of the Makerbase Robin Nano v1.x range.
The newer board revision from Makerbase is physically different to previous version, with updated MCU, different driver type, different driver quantity and larger boot offset.

- Drivers: 4 x Integrated TCM2225
- MCU: STM32F407
- Boot Offset: 32kb

### DIFF
--- a/config/generic-mks-robin-nano-s-v1.3.cfg
+++ b/config/generic-mks-robin-nano-s-v1.3.cfg
@@ -1,0 +1,103 @@
+# This file contains common pin mappings for the Makerbase Robin Nano-S v1.3
+# (TCM2225 drivers for X, Y, Z and E).
+
+# INSTRUCTIONS FOR COMPILING
+# To use this config, the firmware should be compiled for the STM32F407.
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  In general it looks fine, but commits need to have a signed-off-by line (see https://www.klipper3d.org/CONTRIBUTING.html ) and these config files need to be added to the regression test suite (https://www.klipper3d.org/Example_Configs.html point 4).

-Kevin

PR #5474, opened 2022-05-02
config:Create configuration file for Fly-Gemini-v1 motherboard

kluoyun:
Create a configuration file for Mellow's Fly-Gemini-v1 motherboard

Signed-off-by: Xiaokui Zhao <xiaok@zxkxz.cn>


KevinOConnor:
Thanks.  The config file looks fine to me, but it needs to be added to the regression test suite (https://www.klipper3d.org/Example_Configs.html point 4).

-Kevin

kluoyun:
> Thanks. The config file looks fine to me, but it needs to be added to the regression test suite (https://www.klipper3d.org/Example_Configs.html point 4).
> 
> -Kevin

Hi Kevin, thanks for pointing this out, I have added the STM32F405 and the config file to the test case.

KevinOConnor:
Thanks.

-Kevin

PR #5477, opened 2022-05-03
Create printer-biqu-b1-se-plus.cfg

maxcalavera81:


config/printer-biqu-b1-se-plus.cfg
```diff
@@ -0,0 +1,219 @@
...
+########################################
+
+########################################
+# Heater Bed Pins
+########################################
+[heater_bed]
+heater_pin: PD7
+sensor_type: Generic 3950
+sensor_pin: PA1
+control: pid
```
ld3300, line 122: Might be best to remove the PID section, and let the user decide later if they want to use it for their bed.

config/printer-biqu-b1-se-plus.cfg
```diff
@@ -0,0 +1,219 @@
...
+speed: 100
+z_hop: 5
+z_hop_speed: 5
+
+[output_pin probe_enable]
+pin: PE5
+value: 1
+
+########################################
+# Bed Mesh configuration               # exemple off bed mesh config
```
ld3300, line 164: Since bed mesh is a part of the general configuration steps it should be left out of printer example files.

config/printer-biqu-b1-se-plus.cfg
```diff
@@ -0,0 +1,219 @@
...
+
+########################################
+# Filament Sensor configuration
+########################################
+[filament_switch_sensor Sensor_Filamento]
+switch_pin: !PC2
+pause_on_runout: false #pause handled by macro
+runout_gcode:
+  M600 #calls a macro with the name M600 (so it can be called by slicers too)
+insert_gcode:
```
ld3300, line 210: You can remove optional features like insert_gcode.  Having it call a throwaway gcode will end up being an error a user may have to try and figure out

ld3300:
Hi @maxcalavera81,
Thank you for contributing so others have a place to start.  I'm not a reviewer, so can't approve anything, but I have put some notes in your commit for you to consider.  It might also be helpful to users to have a link to look for the potential board problem you mention at the top of the file.

Please also add the 'Signed-off-by:' section as required in the contributing guidelines [here](https://github.com/Klipper3d/klipper/blob/722ad4a1d2798d600ad34c8c8e7d74171e018aef/docs/CONTRIBUTING.md#format-of-commit-messages).

Best,
-David

config/printer-biqu-b1-se-plus.cfg
maxcalavera81: I can delete it, but then when klipper checks the configuration it will give an error because these parameters are missing.  Is it really convenient to delete?

config/printer-biqu-b1-se-plus.cfg
maxcalavera81: The biqu b1 se plus has automatic leveling as standard.  I left it commented because it can be very useful for those who will use it....

config/printer-biqu-b1-se-plus.cfg
maxcalavera81: On this point you are right.  I will delete as soon as I see it

maxcalavera81:
Good Morning.  I appreciate the feedback and have already answered some of the questions.  I just didn't quite understand the last part of the signature.

"Please also add the 'Signed-off-by:' section as required in the contributing guidelines [here](https://github.com/Klipper3d/klipper/blob/722ad4a1d2798d600ad34c8c8e7d74171e018aef/docs/CONTRIBUTING.md#format-of-commit-messages)."

config/printer-biqu-b1-se-plus.cfg
ld3300: I am wrong on that one, sorry.

config/printer-biqu-b1-se-plus.cfg
ld3300: It will be useful and I expect they will benefit, but bed mesh is already in the startup documentation under the installation and configuration section, so to minimize the amount of information someone needs to get the printer moving correctly leaving out information that will come along later will help minimize the learning curve.  May be best to get another opinion on this one.  I do see the viewpoint of value added that you are expressing.

ld3300:
You should go up and edit the first comment in this string and add

Signed-off-by: Real Name <realmail@email.com>

It states that you have the right to commit this information (which you do) and that you agree to its distribution. I would read the terms of the developer certificate of origin a few sentences down in that link.  The devs won't merge your request without it.

maxcalavera81:
Thanks

Signed-off-by: Bruno Melo <brunomelo81@gmail.com>

KevinOConnor:
Thanks.  In general it looks fine to me, but I have a few comments:
1. The top of the config file says it is for the BigTreeTech SKR 2, that needs to be updated to say it is for the Biqu b1..
2. I don't understand the two comment lines in the stepper_x config section - probably best to remove them.
3. Similar for stepper_y.
4. I don't understand the stepper_x, stepper_y, stepper_z rotation_distance.  They look wrong and/or require a comment describing how they were obtained.  (Note that one should not use a measure-and-trim method to obtain rotation_distance on kinematic steppers - https://www.klipper3d.org/Rotation_Distance.html ).
5. I don't understand the comment in [probe] samples_result.  It should probably be removed.
6. I don't understand the comment in save_z_home home_xy_position - it should probably be removed.
7. It is fine to have a bed_mesh section in `printer-....cfg` config files.  If you wish to leave it commented out, then that is fine too.
8. The heater_fan sections should not explicitly specify default value (see https://www.klipper3d.org/Example_Configs.html point 7f).
9. This config file would need to be added to the printer regression tests (point 4 in link above).

-Kevin

maxcalavera81:
@KevinOConnor I've already corrected everything you suggest. Please, when you have some time, take another look. Thank you very much for your time.

Signed-off-by: Bruno Melo <brunomelo81@gmail.com>

KevinOConnor:
Thanks.  Seems the fan sections are still specifying default values (point 7f in docs link).  Also, we don't recommend setting a `hold_current` ( https://www.klipper3d.org/TMC_Drivers.html#prefer-to-not-specify-a-hold_current ) so unless you've got test results showing it is needed, I don't think it should be specified.

-Kevin

KevinOConnor:
Thanks.  In general, the config looks fine.  But I noticed that the filename does not end in a year - see ( https://www.klipper3d.org/Example_Configs.html point 2) and the file must be added to the regression test cases (point 4).

-Kevin

Fisheiyy:
@maxcalavera81 you have still yet to add the configuration file into the config test list that is required per https://www.klipper3d.org/Example_Configs.html point 4

maxcalavera81:
@Fisheiyy I'm very sorry, but I don't know how to do that. Could you please help? Thank you very much for your availability.

KevinOConnor:
Thanks.  I committed this with 4404c986 on top.

-Kevin

PR #5483, opened 2022-05-04
feat(Makefile): determine available python version

KwadFan:
Instead hardcoding python version let make see what version is available.

According to your discussion in https://github.com/Klipper3d/klipper/pull/4957, especcially
https://github.com/Klipper3d/klipper/pull/4957#issuecomment-1003405059

I understand your concerns using Global Variables from the OS might cause unwanted side effects.
But, obviously runs klipper very well with python3, only the crutial `make menuconfig` needs python2.

In my opinion it doesn't make much sense to install an outdated package to have the possibilty to use klipper on
modern OSes without even having python2 packages. So, if there are no reasons besides backwards compatibilty,
let make determine what is installed. This negotiates the unwanted Global Variables behavior but doesnt limit to install py2.

To see if klipper is working with python3 use the early image of MainsailOS (working on bullseye and ne Framework).

https://github.com/KwadFan/MainsailOS-ng/actions/runs/2271418151

feel free to test the Image.

As said to force users and maintainers to use outdated interpreter is not a good option.
In my opinion, you have to go with the development of OSes because your software is great but depends on a good maintained and running OS :) So, please accept that PR or get it fixed ASAP please.

Regards

Signed-off-by: Stephan Wendel <me@stephanwe.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,10 @@ OBJCOPY=$(CROSS_PREFIX)objcopy
 OBJDUMP=$(CROSS_PREFIX)objdump
 STRIP=$(CROSS_PREFIX)strip
 CPP=cpp
-PYTHON=python2
+
+# Determine python version
+PYTHON=$(shell if [ -n "$(command -v python2)" ]; then echo "python2"; \
+    else echo "python3"; fi)
 
 # Source files
 src-y =


KevinOConnor:
Thanks, but I don't think we need to detect python version. Simpler to just change the default to python3.  Commit 9c45f53b2.

-Kevin

th33xitus:
> Commit https://github.com/Klipper3d/klipper/commit/9c45f53b24754fb92e03836e22796d06ab1e9825.

Doesn't this now make py3 a Klipper dependency?

Before you required py2 installations on more modern linux distros which only shipped with py3 pre-installed.
Now you require a py3 installation on all systems which worked with py2 absolutely fine before.

This just reverses the problem, doesn't it? I think this PR was meant to eradicate this exact problem to require a different python version only for configuring and compiling the firmware.

KevinOConnor:
You might be right.  If it causes problems I'll revert it.

In either case users can simply run `make PYTHON=python2` or `make PYTHON=python3`.

Alternatively, we could change the Makefile to default to `PYTHON=python` so either is run.

I don't see a reason to auto-detect the version.

Cheers,
-Kevin

th33xitus:
I think it should be fine, now thinking about it a second time.
I assume the majority of Klipper users now also use Moonraker and a Klipper Webinterface.
Moonraker depends on Python 3, so those systems already have Python 3 installed in that case.

KwadFan:
> You might be right. If it causes problems I'll revert it.
> 
> In either case users can simply run `make PYTHON=python2` or `make PYTHON=python3`.
> 
> Alternatively, we could change the Makefile to default to `PYTHON=python` so either is run.
> 
> I don't see a reason to auto-detect the version.
> 
> Cheers, -Kevin

Yes, you are right, users **can**, but most of them will be confused what they have to do. So, a solution from your site is in my opinion the waaayyy better solution. Therefor, what should happen to this PR? Close it?

Regards

PR #5485, opened 2022-05-04
Add support for MPU-9250 (and MPU-6050) accelerometer

bluesforte:
Add support for mpu9250 accelerometer over I2C bus

Instead of going and getting an ADXL345, I had an MPU-9250 accelerometer/gyro lying around from a previous project and figured it'd be nice to have support for a new accelerometer in Klipper. I adapted the ADXL345 code in order to work on the MPU9250.

As this is an I2C device, I had to do some adapting and testing to ensure there was enough bandwidth; and thankfully the 400kbs it runs at seems to work just fine at 4000 samples/sec (its max sample rate) over a 1M USB cable. This should pave the way for other I2C accelerometers' support, although ideally I think there should probably be a nicer more abstract interface for accelerometers if we start adding more (though I have to give my compliments to Kevin O'Connor - I found the current system was quite straightforward to adapt).

Added documentation for the MPU-9250 as well. Hopefully it doesn't confuse users who bought ADXL345's; but I also think there's enough outside documentation all over the web on that device by now.

Also note that this should work with the MPU-6050 as well (since that is what the MPU-9250 is based on, and the register map is the same); though I haven't tested it on that device.

Signed-off-by: Harry Beyel <harry3b9@gmail.com>


bremsstrahlung:
Hi, I tested this PR with a mpu-6050, this device has device_id=0x68 instead of 0x73. With this changed it seems to give correct values. No actual resonance testing has been done by me so far

Sineos:
https://components101.com/sensors/MPU9250-9-dof-mems-sensor-module-datasheet-pinout-features-working

Seems SPI is supported as well?

bluesforte:
> Hi, I tested this PR with a mpu-6050, this device has device_id=0x68 instead of 0x73. With this changed it seems to give correct values. No actual resonance testing has been done by me so far

Awesome, thanks for checking that out.  I updated to include the 6050's device ID as well during the sanity check.

bluesforte:
> https://components101.com/sensors/MPU9250-9-dof-mems-sensor-module-datasheet-pinout-features-working
> 
> Seems SPI is supported as well?

Yes, SPI is supported by the chip but I2C must be disabled in the USER_CTRL register at device reset.  Support for SPI could probably be added pretty straightforward by wrapping all read/writes and adding a setting to the struct, but I didn't really see the need when I2C over a simple USB cable works well.

dmbutyugin:
Thank you for taking time to implement the support for another sensor! I have some high-level questions and feedback on this PR:

- Both [MPU-9250](https://invensense.tdk.com/products/motion-tracking/9-axis/mpu-9250/) and [MPU-6050](https://invensense.tdk.com/products/motion-tracking/6-axis/mpu-6050/) seem to be EOL'ed. Therefore, it is not entirely clear what is the value of adding the support for them to the mainline Klipper? Is there, perhaps, some non-EOL'ed chip that has the same or similar communication protocol?
- You have implemented the support of reading the data via I2C (400 kHz). What is the attainable accelerometer sample collection rate? I understand it will be around 1000 samples/sec, but I didn't find anything specific in the documentation. I see that, technically, the code supports up to 4000 samples/sec data collection rate (in `SAMPLE_RATE_DIVS`), but I'm not sure if all the recorded data can be collected via I2C. And could you show some results of the resonance measurement? I generally suspected that I2C interface won't be sufficiently fast for high-quality measurements, but perhaps I'm wrong and I2C will be sufficient for the resonance measurements purposes.
- I see that there is a lot of code duplication between the new sensor and adlx345. I understand that the communication protocols and data formats are different. However, ideally we should define some common helper classes like `AccelerometerQueryHelper`, `AccelerometerCommandHelper`, `AccelerometerChipHelper` and then create `ADXL345` and `MPU9250` classes that just provide/override the necessary methods for chip communication and data decoding.

bluesforte:
> Thank you for taking time to implement the support for another sensor! I have some high-level questions and feedback on this PR:
> 
>     * Both [MPU-9250](https://invensense.tdk.com/products/motion-tracking/9-axis/mpu-9250/) and [MPU-6050](https://invensense.tdk.com/products/motion-tracking/6-axis/mpu-6050/) seem to be EOL'ed. Therefore, it is not entirely clear what is the value of adding the support for them to the mainline Klipper? Is there, perhaps, some non-EOL'ed chip that has the same or similar communication protocol?
> 
>     * You have implemented the support of reading the data via I2C (400 kHz). What is the attainable accelerometer sample collection rate? I understand it will be around 1000 samples/sec, but I didn't find anything specific in the documentation. I see that, technically, the code supports up to 4000 samples/sec data collection rate (in `SAMPLE_RATE_DIVS`), but I'm not sure if all the recorded data can be collected via I2C. And could you show some results of the resonance measurement? I generally suspected that I2C interface won't be sufficiently fast for high-quality measurements, but perhaps I'm wrong and I2C will be sufficient for the resonance measurements purposes.
> 
>     * I see that there is a lot of code duplication between the new sensor and adlx345. I understand that the communication protocols and data formats are different. However, ideally we should define some common helper classes like `AccelerometerQueryHelper`, `AccelerometerCommandHelper`, `AccelerometerChipHelper` and then create `ADXL345` and `MPU9250` classes that just provide/override the necessary methods for chip communication and data decoding.


-- Though MPU-6050 and MPU-9250 are technically EOL, they are still very readily available from component suppliers (digikey, etc), as well as retailers (it's the 3rd result for me on amazon for "accelerometer breakout."  They are probably some of the most ubiquitous accelerometers for hobbyists.  Hence I think plenty of klipper users would get plenty of use out of this in the mainline.  It's also just good to have an alternate for the ADXL345 (at one point a while back it was sold out at a few retailers in my area).

-- As I mentioned in the original commit message, I2C seems to support 4000samples/sec perfectly fine.  My testing was that I dumped values and looked through them manually to make sure they made sense.  I also monitored the fifo size during measurements and saw that it was sufficiently draining the fifo on each transfer.  Here are my results for resonance compensation using the MPU9250:

shaper_calibrate_x

shaper_calibrate_y

While I don't have an ADXL345 to compare to, these results are in line with what I measured using the test print, and also using the recommended shaper got me great print results after.


-- Agree with your assessment on having helper classes.  However, I wanted to make this as standalone as possible given it's the first new accelerometer support as far as I know.  I thought it was best to first add the functionality, and then in a separate commit work on making it more modular like that.  This prevents risk of anything going wrong with the ADXL345 (which I can't test) during refactoring, and I thought would also make the code review a bit better.  I indeed started off by making these types of classes as you mentioned, but personally decided it would probably be better to keep them separate as a first step.  If you think that's a requirement, I'd be happy to work on the class approach, though with the caveat that others would have to test the ADXL345 functionality to ensure no regressions.

dmbutyugin:
> > Thank you for taking time to implement the support for another sensor! I have some high-level questions and feedback on this PR:
> > ```
> > * Both [MPU-9250](https://invensense.tdk.com/products/motion-tracking/9-axis/mpu-9250/) and [MPU-6050](https://invensense.tdk.com/products/motion-tracking/6-axis/mpu-6050/) seem to be EOL'ed. Therefore, it is not entirely clear what is the value of adding the support for them to the mainline Klipper? Is there, perhaps, some non-EOL'ed chip that has the same or similar communication protocol?
> > 
> -- Though MPU-6050 and MPU-9250 are technically EOL, they are still very readily available from component suppliers (digikey, etc), as well as retailers (it's the 3rd result for me on amazon for "accelerometer breakout." They are probably some of the most ubiquitous accelerometers for hobbyists. Hence I think plenty of klipper users would get plenty of use out of this in the mainline. It's also just good to have an alternate for the ADXL345 (at one point a while back it was sold out at a few retailers in my area). Also upon further research, it looks like this will work for pretty much all InvenSense accelerometers - the latest IAM-20381 seems to have the same register map and sample rates, so might just need its device ID added. I only listed the MPU6050 as that is effectively identical to the MPU9250 (other than device ID). If someone has an IAM-20381 or other InvenSense accelerometer, please test it too (similar to how @bremsstrahlung did)!

I see. In general, it would indeed be good to have some support of other accelerometers in Klipper. Maybe indeed there will be more interest in testing this code with other accelerometers, or even these ones? As per [CONTRIBUTING](https://www.klipper3d.org/CONTRIBUTING.html) guide, 'As a general "rule of thumb", submissions should target a user base of at least a 100 real-world users.' FWIW, I was able to find some user questions and requests for MPU6050 support in Klipper.


> -- As I mentioned in the original commit message, I2C seems to support 4000samples/sec perfectly fine. My testing was that I dumped values and looked through them manually to make sure they made sense. I also monitored the fifo size during measurements and saw that it was sufficiently draining the fifo on each transfer.

TBH, 4K samples/sec looks a bit too good to be true with I2C. FWIW, ADXL345 only supports 800 Hz data sampling (and output) rate with 400 kHz I2C.

Though you have 2K samples/sec by default. So, if that really works, maybe that would already be not too bad and make these accelerometers quite usable. Maybe you could provide a raw log from some of resonance testing (can be obtained with `TEST_RESONANCES AXIS=X OUTPUT=raw_data`)?

Separately, there was a question regarding I2C accelerometers, and some feedback was that I2C seems to be unstable:

https://klipper.discourse.group/t/mpu6050-support/84

But maybe it is not really (at least, in this case). What cable length did you use BTW?


> -- Agree with your assessment on having helper classes. However, I wanted to make this as standalone as possible given it's the first new accelerometer support as far as I know. I thought it was best to first add the functionality, and then in a separate commit work on making it more modular like that. This prevents risk of anything going wrong with the ADXL345 (which I can't test) during refactoring, and I thought would also make the code review a bit better. I indeed started off by making these types of classes as you mentioned, but personally decided it would probably be better to keep them separate as a first step. If you think that's a requirement, I'd be happy to work on the class approach, though with the caveat that others would have to test the ADXL345 functionality to ensure no regressions.

I think for basic testing of the code for your accelerometer, just copying the code from adxl345 and modifying to your needs seems the easiest. And for testing, that's sufficient. But integrating the code into Klipper mainline is another story. This will add a lot of technical debt that someone will have to pay. Otherwise, modifying anything in the accelerometer codebase or fixing bugs will be a nightmare. Also, if we have common base that's easily extendable, we may support other accelerometers in the future.

bluesforte:
To be completely honest, I feel as though you aren't really putting in the effort here as a lot of these questions were answered already.

1.  As the MPU-6050 and 9250 are super ubiquitous, I would hazard a guess that there are a *LOT* more than 100 users who would find this feature useful.
2.  Are you suggesting that because you "feel" like I2C should be too slow, that you don't believe the testing methods I mentioned, and attached resonance charts?  Some simple math:  4k samples/sec * 6B/sample * 8b/B = 192kb/sec.  Even with I2C overhead (start/ACK bits), and protocol overhead (checking fifo size), I'd estimate we're at around 250kb/sec.  At 400kHz, the throughput should be roughly 400kb/s.  I changed my sample rate by setting "rate: 4000" in the printer.cfg file.  I set the default to 2000 as that still worked for resonance testing for me and thought it might be more conservative in case someone had bad wiring etc.  Re: cable length - I mentioned already in the commit. I used a 1M USB cable (USB clocks at 48MHz so this should be way overkill).
3. Again, I considered putting these things into a separate class but it felt to me like it was making changes but adding more chance of introducing regressions or someone missing something during review.  The only classes I personally see value in abstracting are MPU9250QueryHelper/CommandHelper and ClockSyncRegression.
- MPU9250QueryHelper/CommandHelper makes sense to have this abstracted into a generic "AccelerometerQueryHelper" type class, but the actual benefits seem minimal.  Resonance testing measurements are abstracted already via the API interface.  Other functionality is pretty chip-specific so though the code may be duplicated in some parts, there shouldn't be cross-dependencies where modifying one accelerometer breaks the other.  Again, I can make these classes if it's really required for check-in, but I don't think the rationale you mentioned is really applicable here.  Doesn't seem like it's creating technical debt to me, though I do see the added value of having a higher-level interface so I'm open minded...Just remember that it will also require changes to adxl345.py and potentially other places with shared code.
- ClockSyncRegression seems relatively chip-specific, and given that the original author didn't put it in a separate file I figured there must be a reason (one that I came up with, is that there is no generic 'clockutils.py'-like file I could find to put it into).  So yes, this could definitely be abstracted, but for what benefit?  Functionality seems atomic enough, and other accelerometers may need this adjusted.

dmbutyugin:
> To be completely honest, I feel as though you aren't really putting in the effort here as a lot of these questions were answered already.

I'm sorry if you got that feeling. Please understand that code reviews are sometimes done by volunteers who may not do this all the time and do not have the full authority of the code, therefore they may err on the side of caution. I personally prefer to ask a question if I'm confused about something and I think it's better to ask a question, than to keep being confused and uncertain. And so if I missed something obvious or already mentioned - I beg your pardon. FWIW, I do not insist that it must be me who reviews this PR.

>     1. As the MPU-6050 and 9250 are super ubiquitous, I would hazard a guess that there are a _LOT_ more than 100 users who would find this feature useful.

Yes, and I did mention that I found users asking questions about support for these accelerometers. So this is a good indication that many users may be interested in that feature.

>     2. Some math:  4k samples/sec * 6B/sample * 8b/B = 192kb/sec.  Even with I2C overhead (start/ACK bits), and protocol overhead (checking fifo size), I'd estimate we're at around 300kb/sec.  At 400kHz, the throughput should be roughly 400kb/s.  I changed my sample rate by setting "rate: 4000" in the printer.cfg file.  I set the default to 2000 as that still worked for resonance testing for me and thought it might be more conservative in case someone had bad wiring etc.

I looked at I2C protocol schematics in MPU6050 datasheet and estimated 92 time units ("bits") required per sample, including reading the data, FIFO size and I2C protocol overhead. If I did not make a mistake, that is. That's pretty close to 100 "bits" per sample theoretically available at 400 kHz and 4K samples/sec. That's why I asked further questions. 

>  All that said, I'm not actually convinced that 4000kHz fifo output rate is actually attainable on the chip (it may still output to FIFO at 1kHz).  I'm looking more into the raw data now.

Yes, please, let's find out the true speed the accelerometer can work at.

>  Re: cable length - I mentioned already in the commit. I used a 1M USB cable (USB clocks at 48MHz so this should be way overkill).

Thanks, I did miss it in the initial message. Separately, USB carries data over a twisted pair as a differential signal, much less susceptible to external EMF interference. I2C (as well as SPI) do not transmit data over a differential pair, and generally are more susceptible to the electromagnetic interference. In any case, it's good that this cable of 1m works.


>     3. Again, I considered putting these things into a separate class but it felt to me like it was making changes but adding more chance of introducing regressions or someone missing something during review.  The only classes I personally see value in abstracting are MPU9250QueryHelper/CommandHelper and ClockSyncRegression.
> 
> 
>     * MPU9250QueryHelper/CommandHelper makes sense to have this abstracted into a generic "AccelerometerQueryHelper" type class, but the actual benefits seem minimal. 

MPU9250QueryHelper and ADXL345QueryHelper classes are identical. The benefit is to not have the same code in two places.

MPU9250CommandHelper ADXLCommandHelper are mostly identical, except the accelerometer names and the min/max register values. The former can be solved by replacing `adxl345`/`mpu9250` in the code with `self.name` (and a bit of cleanup of that value).

MPU9250 and ADLX345 have a good share of common code, including some complex clock processing. So, a shared common parent or helper class could be very beneficial.
   
> Other functionality is pretty chip-specific so though the code may be duplicated in some parts, there shouldn't be cross-dependencies where modifying one accelerometer breaks the other. 

It is the other way around: changes or fixes in one accelerometer may need to be backported to other accelerometers (e.g. improvements in the clock drift compensation).

> Again, I can make these classes if it's really required for check-in, but I don't think the rationale you mentioned is really applicable here.  Doesn't seem like it's creating technical debt to me, though I do see the added value of having a higher-level interface so I'm open minded.

My opinion is that this level of code duplication is hard to maintain moving forward. In its current form it also does not help with supporting other accelerometers at all. Ideally, it would be great to have a system similar to TMC drivers support, where adding another one is rather easy and straightforward, and requires minimal amount of code added that's really specific to that driver. Of course, the current accelerometer support is nowhere near that state, and it is unfair to ask you to make it that way. But at least it would make sense to not make things worse. But that's just my opinion, maybe Kevin would be open to merging the PR as-is, I do not know.

> Just remember that it will also require changes to adxl345.py and potentially other places with shared code.

Yes, that's of course to be expected. I can test your changes for adxl345 to make sure it still works.


>     * ClockSyncRegression seems relatively chip-specific, and given that the original author didn't put it in a separate file I figured there must be a reason (one that I came up with, is that there is no generic 'clockutils.py'-like file I could find to put it into).  So yes, this could definitely be abstracted, but for what benefit?  Functionality seems atomic enough, and other accelerometers may need this adjusted.

The ClockSyncRegression code in your implementation and in adxl345.py is literally identical. So, I'm not sure what chip-specific implementation part you are referring to. It was not moved to some clockutils library because, perhaps, it was not used in that form anywhere else so far.

bluesforte:
Some good points there.  I may not agree with all points, but there's plenty of good meat in there and I think it's faster to just write some code instead of more back and forth :).  I'll work on the abstraction shortly.

But first, your earlier point on checking into the raw data led to finding that it always seems to be outputting at 1kHz to FIFO, despite following the datasheet to have it set to 4kHz.  Will continue to do some more investigation, though might just accept that the datasheet is either missing some info or just wrong (can't seem to find many successful stories on searches of people getting it to output at 4kHz).  Then on to the refactor.

KevinOConnor:
Thanks.  It looks like useful functionality to me.

Some random comments.
1. I don't want to "nitpick", but I prefer unambiguous copyright statements, so avoid lines like "derived from ..." in the main copyright header.  I think in this case you want to list both my and your copyrights in the two new code files you added.  See src/atsamd/i2c.c for an example.
2. Note that Klipper's i2c support isn't great on mcus.  It should be fine on the "Linux mcu" because it uses the low-level Linux kernel i2c drivers.  If one were to wire this to an actual mcu, then I expect there would be random shutdowns due to noise and many more drops (most mcu implementations only implement 100K rate - despite what may be configured in the printer.cfg file).  This doesn't deter committing this code, but be aware of the limitations.
3. It's not necessary to add a note to Config_Changes.md - that file is intended to list changes that are not backwards compatible.
4. FWIW, I'm not sure I'd list mpu-9250 support so prominently in Measuring_Resonance.md .  I expect most users are primarily looking for the recommended setup and it seems to me that is still an adxl345.  I'll defer to Dmitry on this though.
5. FWIW, I think it makes sense to implement obvious code reuse.  For example, if a Python class is unchanged from its upstream source then just use the upstream source instead of duplicating it - something like add `from . import adxl345` and use `return adxl345.ADXL345QueryHelper(self.printer, conn)`.  Similarly for other areas where minor changes to the upstream could result in code reuse.

Finally, I do think this is useful and I don't see anything fundamentally preventing us from merging this support.  Expect us to have questions and comments along the way though.  We want to make sure we're comfortable with changes before they are committed - I don't think there is any "rush" to commit this to the master branch.

Thanks again,
-Kevin

Sineos:
I ordered such a board and compared it to an ADXL345. Tested the MPU board with a RPi 3b and attached on the RPi Pico. These are my results:

**Noise:**
* MPU / Pico: Axes noise for xy-axis accelerometer: 1.886053 (x), 6.492534 (y), 3.879832 (z)
* MPU / RPi: Axes noise for xy-axis accelerometer: 1.612284 (x), 1.862545 (y), 3.595899 (z)
* ADXL / Pico: Axes noise for xy-axis accelerometer: 63.127429 (x), 43.972955 (y), 68.205144 (z)

**Samples / second:**
* MPU / Pico: ~500 Hz
* MPU / RPi: ~500 Hz
* ADXL / Pico: ~ 3200 Hz

**Resonance Testing**

x

y

**Observation**
Having both chips defined in the cfg (`[adxl345]` and `[mpu9259]` section), leads to a strange error message:
```
mux command ACCELEROMETER_MEASURE CHIP None already registered ({None: >, 'adxl345': >, 'mpu9250': >})
```

**Raw Data**
https://drive.google.com/file/d/1uEJmvVqN2aLEDBOYDUCzoSC7rZnxVF0D/view?usp=sharing

dmbutyugin:
@bluesforte OK, sounds good, thanks!

> 4. FWIW, I'm not sure I'd list mpu-9250 support so prominently in Measuring_Resonance.md . I expect most users are primarily looking for the recommended setup and it seems to me that is still an adxl345. I'll defer to Dmitry on this though.

I actually think it is fine to mention all supported accelerometers in Measuring_Resonance.md. However, we should also mention possible limitations of the sensors, if any. In this particular instance, it is likely worthwhile to mention that the chip will have substantially lower sampling rate (the exact max rate TBD, as far as I understand). So, for a new user, MPU-9250/6050 may be not the first choice, but if they already own one, or it is much easier to obtain this sensor than the adxl345/343, then they may consider using it instead.

@Sineos Thank you for your testing. So, which of the two sensors was it, exactly? Then, a few thoughts on your results:

> **Noise**:

Noise seems to be much better on the MPU board. Whether it is because of much lower sampling rate (I think ADXL345 also has lower noise at low sampling rates), or because it has a better power chain, I do not know. But this is an interesting result nonetheless.

> Samples / second:
> * MPU / RPi: ~500 Hz

So, you got even smaller sampling rate, 500 instead of 1K samples/sec. That's different from what @bluesforte reported, and TBH a bit too low. I hope we'll figure out the real maximum sampling rate for this sensor.

BTW, these are the spectrograms obtains from your data in `graph_accelerometer.py -s --max_freq=500` mode:

|specgram_x_20220515_075224|specgram_x_20220515_072956|
|:--:|:--:|
|adxl345|mpu|

Note that those lines on the MPU chart that looks like reflections from the top of the spectrogram are not a defect of the MPU sensor. Instead, they are the result of the sampling rate - one can replicate them by generating simulated ideal accelerometer readings and running the script on the simulated ideal data. Though the final results are comparable:
adxl345_mpu_comparison
it is an open question how reliable is the data at higher frequencies. Even though today Klipper does not reliably test the resonances for frequencies above 133 Hz, this may change in the future (as some very still printers may have resonances above that threshold).

> Having both chips defined in the cfg ([adxl345] and [mpu9259] section), leads to a strange error message:

Oh, yes, that's a 'bug' in the default registration. Well, not really a bug, but I suppose a better error message should be raised in this case (as only a single unnamed accelerometer is allowed to be defined).

Sineos:
> So, which of the two sensors was it, exactly?

The ADXL was the standard Chinese one. Attached to the Pico with GPIO23 pulled high.
The MPU Board, I got off ebay: https://www.ebay.de/itm/181877281002. From what I read in the internet it is actually a MPU-9255 (if that matters anything) 

> So, you got even smaller sampling rate, 500 instead of 1K samples/sec.

This is what I got with the standard settings on both Raspberries. From what I read the Pico should also support 400 kHz for I2C but adding `i2c_speed: 400000` did not change anything. If there is a way to modify the speed setting, I'll happily test it.

bluesforte:
> > So, you got even smaller sampling rate, 500 instead of 1K samples/sec.
> 
> This is what I got with the standard settings on both Raspberries. From what I read the Pico should also support 400 kHz for I2C but adding `i2c_speed: 400000` did not change anything. If there is a way to modify the speed setting, I'll happily test it.

I'm just getting back to my testing now (was a bit busy last week).  On the Pi, you need to edit /boot/firmware/config.txt and add/uncomment dtparam=i2c_arm=on,i2c_arm_baudrate=400000 and reboot.  I'm not familiar with the Pico but it being a microcontroller, I assume you probably don't need to modify any dto's to have it work?  Keep in mind that the I2C speed and the device sample rate are two different settings.

That said, in [mpu9250] can you add rate: 4000 (the device sample rate)?  That should give you more samples (which right now seems to actually be 1000).  Working on figuring out why it's not actually running at the advertised 4000.

Sineos:
I did a quick test on my Pico with following settings:
```
[mpu9250]
i2c_mcu: pico
i2c_bus: i2c0a
i2c_speed: 400000
rate: 4000
```
The result was like before. Average sampling rate of 503 Hz.

@dalegaard Hope you could chime in: Is there any way to set the I2C speed for the RP2040?

bluesforte:
Update: figured out the issue about 4kHz sample rate: it was due to ACCEL_CONFIG_2 being set to 0x80 instead of 0x08 (doh).  
The previous config sets the sample rate to 1kHz, and assuming rate is set to 4000 there is no divider, so I'm still not sure what's going on for @Sineos.  Any more info you can provide would be good - do you see the same results when connected directly to the Pi?

Anyway, once I set the rate to *actual* 4kHz, my method of reading one sample at a time from FIFO seemed to be just barely too slow for the I2C bus (fifo size increased by 1-2 samples per tick).  So I changed it to read as much of the FIFO as possible each tick and that works.  Getting 4khz samples (example accelerometer_measure attached).  There is a huge amount of noise in some samples though, which leads me to believe that may be one reason why they implemented a LPF in the hardware.  I'm thinking 1kHz sample rate with higher quality is probably better than 4kHz with noise - others' thoughts?

[mpu9250-20220518_204207.csv](https://github.com/Klipper3d/klipper/files/8720807/mpu9250-20220518_204207.csv)
.

dewi-ny-je:
@bluesforte for info, CSV files are terrible because outside US they require changing decimal and field separator. Please use XLSX or ODS which save that information and which are also compressed.

dewi-ny-je:
In https://cdn.sparkfun.com/assets/learn_tutorials/5/5/0/MPU-9250-Register-Map.pdf (page 15) I read that the max bandwidth of the accelerometer with 4 kHz data rate is 1130 Hz.
Can maybe the readout be limited to 2 kHz to make everything more stable without losing any high frequencies?

Even using a 1 kHz data rate with the bandwidth of 460 Hz could be enough.

Sineos:
> Any more info you can provide would be good - do you see the same results when connected directly to the Pi?

* RPi 3B+ with `i2c_arm_baudrate=400000`
* `SET_ACCEL_CONFIG2 = 0x08`
* Settings:
```
[mpu9250]
i2c_mcu: rpi
i2c_bus: i2c.1
i2c_speed: 400000
rate: 4000
```

**Results:**
* 1007 Hz average sampling rate
* Raw Data:
[raw_data_x_20220519_093143.zip](https://github.com/Klipper3d/klipper/files/8728348/raw_data_x_20220519_093143.zip)
 
> CSV files are terrible because outside US they require changing decimal and field separator. Please use XLSX or ODS which save that information and which are also compressed.

@dewi-ny-je A CSV file is the result of running e.g. `TEST_RESONANCES AXIS=X OUTPUT=raw_data` and can be further analyzed like shown by @dmbutyugin above.

dmbutyugin:
@dewi-ny-je yeah, CSV file is generated by Klipper directly. Unfortunately, it is impractical to implement writing the XLSX or ODS formats within Klipper. But most of the time, people use Klipper scripts to process that data anyway, so it is not too bad.

@Sineos, I'm unsure why you are still getting just 1007 Hz after setting ACCEL_CONFIG_2 = 0x08. Maybe @bluesforte will push the exact code that works in his setup.

@bluesforte 
> So I changed it to read as much of the FIFO as possible each tick and that works. Getting 4khz samples (example accelerometer_measure attached - ignore the few bad datapoints in there for now as I know what caused that).

I'm not sure which method you implemented to read from FIFO exactly (you still need to check its size, right? Or are you just doing it less frequently, not after every data point read?). But looking at the data, I noticed the following:

| time | accel_x | accel_y | accel_z | | effective_freq |
|-- | -- | -- | -- | -- | --|
12.830844 | -153.228906 | 976.834277 | -8523.35791 |   | 5050.50505047876
12.831042 | -225.054956 | 919.373437 | -8485.050684 |   | 5050.50505052407
12.831638 | -167.594116 | 1010.353101 | -8595.18396 |   | 1677.8523489938
12.831836 | -110.133276 | 1082.17915 | -8556.876733 |   | 5050.50505052407
12.832034 | -186.747729 | 1225.83125 | -8523.35791 |   | 5050.50505047876
12.832233 | -210.689746 | 1019.929907 | -8528.146313 |   | 5025.12562813358
12.832431 | -210.689746 | 900.219824 | -8556.876733 |   | 5050.50505052407
12.83263 | -215.478149 | 1134.851587 | -8523.35791 |   | 5025.12562813358
12.833225 | -191.536133 | 1101.332764 | -8518.569507 |   | 1680.67226890596
12.833423 | -301.669409 | 991.199487 | -8528.146313 |   | 5050.50505052407
12.833622 | -282.515796 | 948.103857 | -8561.665137 |   | 5025.12562813358
12.83382 | -186.747729 | 976.834277 | -8528.146313 |   | 5050.50505052407
12.834019 | -258.573779 | 948.103857 | -8523.35791 |   | 5025.12562813358
12.834217 | -225.054956 | 981.622681 | -8537.72312 |   | 5050.50505047876
12.834415 | -167.594116 | 895.431421 | -8547.299927 |   | 5050.50505052407
12.834812 | -191.536133 | 933.738647 | -8537.72312 |   | 2518.89168766037
12.835011 | -244.208569 | 876.277808 | -8489.839087 |   | 5025.12562813358
12.835209 | -210.689746 | 981.622681 | -8518.569507 |   | 5050.50505047876

here *effective_freq[i] = 1 / (time[i] - time[i-1])*. So, the concerning part is a bit unstable sampling frequency. Most of the time it is around 5K, but occasionally it drops to 2.5K - 1.7K. Technically, FFT used in accelerometer data analysis requires that all samples are sampled at the same stable frequency. Practically, what we observe here may be the artifact of the FIFO reading - perhaps all the samples were indeed sampled at the stable ~4K frequency. However, if we attempt to 'fix' the data by, say, interpolating to the appropriate fixed rate time points, we would actually skew the data. So, even if this is an artifact of the FIFO reads, we should fix it one way or the other.

Separately, I'd say if 4K samples/sec reading speed over I2C is not attainable easily, but 2K is without major optimization and tuning, maybe we should stick with 2K sampling rate? It would already be pretty good, in my opinion, and comparable to adxl345 (3.2K samples/sec).

bluesforte:
> So, the concerning part is a bit unstable sampling frequency. Most of the time it is around 5K, but occasionally it drops to 2.5K - 1.7K. Technically, FFT used in accelerometer data analysis requires that all samples are sampled at the same stable frequency. Practically, what we observe here may be the artifact of the FIFO reading - perhaps all the samples were indeed sampled at the stable ~4K frequency. However, if we attempt to 'fix' the data by, say, interpolating to the appropriate fixed rate time points, we would actually skew the data. So, even if this is an artifact of the FIFO reads, we should fix it one way or the other.

Thanks for this analysis - I have a feeling that this is due to the fifo reads rather than the device, though I don't have any hard evidence for this yet (other than I believe the FIFO is written to as part of a PLL with internal clock source).  Does the ADXL have any of these artifacts?  Looking into the fifo status, it seems pretty consistent (within 6-8 samples each tick):

```
...
mcu 'rpi': #output: fifo: 30, to_read: 30, buf:0
mcu 'rpi': #output: fifo: 36, to_read: 36, buf:0
mcu 'rpi': #output: fifo: 36, to_read: 36, buf:0
mcu 'rpi': #output: fifo: 42, to_read: 42, buf:0
mcu 'rpi': #output: fifo: 30, to_read: 30, buf:0
mcu 'rpi': #output: fifo: 36, to_read: 36, buf:0
mcu 'rpi': #output: fifo: 36, to_read: 36, buf:0
mcu 'rpi': #output: fifo: 36, to_read: 36, buf:0
mcu 'rpi': #output: fifo: 42, to_read: 42, buf:0
mcu 'rpi': #output: fifo: 30, to_read: 30, buf:0
mcu 'rpi': #output: fifo: 36, to_read: 36, buf:0
mcu 'rpi': #output: fifo: 36, to_read: 36, buf:0
mcu 'rpi': #output: fifo: 42, to_read: 42, buf:0
mcu 'rpi': #output: fifo: 30, to_read: 30, buf:0
mcu 'rpi': #output: fifo: 42, to_read: 42, buf:0
...
```
I did a quick grep of the log and didn't see any outliers either.
My thinking is that one or more of these things are happening:
1. The wake up time between ticks is too long (possibly due to i2c transfer times)
2. The MPU's 0.503ms constant output delay needs compensation
3. ClockSyncRegression (or clock reporting) needs to be changed for MPU/i2c
4. MPU chip is inconsistent

I'm going to look into things in this order, but definitely interested in any other thoughts.

> Separately, I'd say if 4K samples/sec reading speed over I2C is not attainable easily, but 2K is without major optimization and tuning, maybe we should stick with 2K sampling rate? It would already be pretty good, in my opinion, and comparable to adxl345 (3.2K samples/sec).

Unfortunately, 2K sampling rate is not possible on these chips - the sample rate divider is only active while the LPF is active, so our only options are 4K, 1K, 500, 250, etc.  That said, it is possible to get a 4K sample rate with 1K output data rate, provided it uses the internal LPF.  This has a cutoff frequency of 218.1Hz, which (I think) is probably fine as the backup plan if 4K output data rate is not consistent enough.

Also, here is the new csv file after the read fixes (frequency variations noted are still present though).
[mpu9250-20220519_234445.csv](https://github.com/Klipper3d/klipper/files/8734856/mpu9250-20220519_234445.csv)



Side note:  I also found a bug that existed in sensor_adxl345.c but apparently never got hit.  The mcu's data buffer needs to have 1 extra byte in order to sendf the data back to the host successfully, so if the data buffer is full, the sendf will cause an EOF error and the host will go into shutdown.  I'll fix this separately after done here.

bluesforte:
Figured it out.  The issue (for anyone curious) was due to the way fifo buffer was being sent.  Previously (and in ADXL), it was sent in a consistent frame of 8 samples.  After the changes, the buffer was flushed every read, so the frame size was dynamic.  There is one line in the python _update_clock: 
```
msg_count = (sequence * SAMPLES_PER_BLOCK
                     + buffered // BYTES_PER_SAMPLE + fifo)
```
where it calculates based on a consistent frame size (SAMPLES_PER_BLOCK).  I changed it back to being a consistent frame size and the inv frequency is now consistent:

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/megam/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/megam/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">

</head>

<body link="#0563C1" vlink="#954F72">



#time | accel_x| accel_y | accel_z | inv_freq
-- | -- | -- | -- | --
7.363718 | -292.093 | 909.7966 | -8580.82 | 3984.064
7.363969 | -201.113 | 948.1039 | -8523.36 | 3984.064
7.364221 | -244.209 | 909.7966 | -8542.51 | 3968.254
7.364472 | -148.441 | 837.9706 | -8523.36 | 3984.064
7.364723 | -177.171 | 938.5271 | -8542.51 | 3984.064
7.364975 | -191.536 | 1024.718 | -8532.93 | 3968.254
7.365226 | -229.843 | 976.8343 | -8518.57 | 3984.064
7.365477 | -296.881 | 1024.718 | -8576.03 | 3984.064
7.365728 | -263.362 | 890.643 | -8580.82 | 3984.064



</body>

</html>

[mpu9250-20220520_072322.csv](https://github.com/Klipper3d/klipper/files/8737332/mpu9250-20220520_072322.csv)



Noise seems to remain quite low for me as well:
```
Send: MEASURE_AXES_NOISE
Recv: // Axes noise for xy-axis accelerometer: 0.975219 (x), 2.470163 (y), 0.997030 (z)
```

Sineos:
@bluesforte can you push an update?

Whatever I do, I cannot get more than 1k. I2C verified to be on 400Hz (Ubuntu decided to give me issues with setting it in config.txt)

```
ubuntu@ubuntu ~/klipper $ echo $((16#$(xxd /sys/class/i2c-adapter/i2c-1/of_node/clock-frequency | awk '{print $(NF-2)$(NF-1)}')))
400000
```

bluesforte:
> @bluesforte can you push an update?

Just did :) try again now (don't forget to reflash the firmware).  Thanks for helping test!!

Sineos:
It works! Good job 👍

* RPi 3B+, I2C baud rate 400kHz
* Average MPU sampling freq: 4025 Hz
* Axes noise for xy-axis accelerometer: 1.613752 (x), 23.129917 (y), 3.107884 (z) 

[raw_data_x_20220520_082408.zip](https://github.com/Klipper3d/klipper/files/8737862/raw_data_x_20220520_082408.zip)

Sineos:
Strange, wanted to cross-check on the Pico (firmware flashed ofc) but now I get I2C timeout errors

```
Transition to shutdown state: MCU 'pico' shutdown: i2c timeout
...
API Dump Helper start callback error
  File "/home/ubuntu/klipper/klippy/extras/motion_report.py", line 46, in _start
    self.startstop_cb(True)
  File "/home/ubuntu/klipper/klippy/extras/mpu9250.py", line 447, in _api_startstop
    self._start_measurements()
  File "/home/ubuntu/klipper/klippy/extras/mpu9250.py", line 385, in _start_measurements
    dev_id = self.read_reg(REG_DEVID)
  File "/home/ubuntu/klipper/klippy/extras/mpu9250.py", line 301, in read_reg
    params = self.i2c.i2c_read([reg], 1)
  File "/home/ubuntu/klipper/klippy/extras/bus.py", line 185, in i2c_read
    return self.i2c_read_cmd.send([self.oid, write, read_len])
  File "/home/ubuntu/klipper/klippy/mcu.py", line 515, in send
    return self._do_send([self._cmd.encode(data)], minclock, reqclock)
  File "/home/ubuntu/klipper/klippy/mcu.py", line 513, in _do_send
    raise self._error(str(e))
CommandError: Unable to obtain 'i2c_read_response' response
```

Sineos:
Works equally well with the Pico:
* Sample rate 4028 kHz
* Axes noise for xy-axis accelerometer: 3.255546 (x), 34.375449 (y), 5.588269 (z)

bluesforte:
At this point I think I'd say this is "functionally complete" unless anything else comes up (would be great if anyone with a MPU-6050 or newer invensense accelerometer could try it out as well).  I started working on abstracting the accelerometer code, which will probably be a much bigger change.  I leave it up to others as whether to merge this as-is or wait until the abstraction is done.

Re: abstraction, for architecture feedback: I was planning to follow a similar style of extras/display.  There would be an Accelerometer base class, and then each device or device family can inherit from it.  Will move ClockSyncRegression into __init__.py for accelerometer folder, since it seems crucial to current accelerometer reading architecture.

bluesforte:
I believe this should be a reasonable level of refactor and abstraction for the python code at least.

A quick overview of the major changes:
- motion_sensor is a module of device classes, similar to the way display is organized.  The idea here is that an accelerometer is really just a specific class of motion sensor, so wanted to make the name make sense for any sort of motion sensor support in the future (also the MPU-9250 supports gyro readings too, maybe support for that could be useful somewhere in the future [deltas?]).
- config and documentation changes associated with this - now the config section is [motion_sensor] with a "chip" option.
- general architecture of communication with the device is unchanged: any new device would need to implement a few abstract methods and commands similar to the way that was done originally with the adxl345.  I didn't see any real need to change this as it works well for the main current use case (resonance testing).

I did testing for the MPU9250; my resonance outputs, noise, etc were unchanged (as expected :)).  Testing for the MPU6050 and ADXL345 was *not* done other than a simple sanity config check, so testing for those devices is needed.

KevinOConnor:
Thanks for working on this.  Sorry for the delay in responding.

I agree it would be good to add support for the mcp9250 (and similar) sensors.

However, it looks like this PR would break existing configs that define an `[adxl345]` config section.  I fear that will "cause havoc" in the short term and I'm not sure I want to endure that pain.  What are your thoughts on keeping the existing naming?

I'm also a bit leery of using the "abstract base class" system for these sensors, as doing so tends to lead to code in one python file "peeking" into member variables in classes in another python file.  (There is a blurb on this at https://www.klipper3d.org/Code_Overview.html#adding-a-host-module .)  For example it looks like the new invensense.py uses `self.name` that is defined in motion_sensor.py .  In my experience Python code rapidly becomes unmaintainable when member variables are accessed across .py files.  (As developers often don't realize a change to a member variable in one class will break some other class.)  FWIW, as a result, I generally prefer the "helper class" abstraction instead of the "base class" abstraction.

Are there functional changes between b0aaf75 and the latest changes?  If so, I'm fine with merging in the base functionality and then discussing the refactoring in follow up PRs.  (Though we should probably come to a consensus on config naming first.)

-Kevin

kubik369:
@KevinOConnor I would propose keeping the `[adxl345]` name in as a deprecated option, for the time being. It can be thrown in the bag with other things when some other big breaking change comes along or more of the little things add up. In the meantime, it can be nudged towards the bottom of the documentation pages as a historical side note - this should minimise the number of new people who use it. As for the people who are already using it, it should be fairly simple to implement. If used, it can fallback to the `adxl345` option with a deprecation message. After removing this option, if required, a permanent config error can be shown when a user tries to set this up, pointing them towards the newer way.

The naming seems a little odd to me though. While an accelerometer is _technically_ a motion sensor, I would just use `accelerometer` instead of `motion_sensor`. When a motion sensor is mentioned, I think of a proximity sensor, google does too and I believe same would go for lot of people (majority?). This might just be my opinion though.

Great job @bluesforte ! I was looking into this as it irked me so much when people cited random reasons for I2C "not being usable" and you beat me to the punch! :hot_face:

dmbutyugin:
My couple of cents:

The section config schema like `[adxl345]`, `[mpu9250]` and such falls in line with other similar cases (e.g. TMC drivers, which are defined as `[tmc2209 stepper_x]` rather than `[tmc stepper_x] type:2209`). It also simplifies the documentation (that adxl345 supports only SPI and MPU9250 - I2C). So, I would vote to keep it that way, though maybe there are some compelling reasons to change it that I may have missed?

@KevinOConnor 
> Are there functional changes between https://github.com/Klipper3d/klipper/commit/b0aaf75e59d3e0fa52f381b15e5a51124a59870a and the latest changes? If so, I'm fine with merging in the base functionality and then discussing the refactoring in follow up PRs.

I agree that this may be a better option in the end, but did you mean some other previous commit by any chance (e.g. https://github.com/bluesforte/klipper-mpu9250/commit/2a7d75758714ca6ef82292ffbba94061d095f7be)? The one you referred to already included a bunch of refactoring. Though there were definitely useful changes to the documentation on enabling and configuring I2C property, for instance.

> Great job @bluesforte ! I was looking into this as it irked me so much when people cited random reasons for I2C "not being usable" and you beat me to the punch!

FWIW, the original sentence in the docs referred to I2C support in ADXL345, which is limited to 800 samples/sec data collection rate, which is, well, not great, compared to the available SPI support. It appears that MPU-9250/6050 sensors are much better in that regard.

KevinOConnor:
> I agree that this may be a better option in the end, but did you mean some other previous commit by any chance (e.g. https://github.com/bluesforte/klipper-mpu9250/commit/2a7d75758714ca6ef82292ffbba94061d095f7be)?

Yes - that is the commit I meant.  I had a cut-and-paste error.

-Kevin

bluesforte:
> However, it looks like this PR would break existing configs that define an `[adxl345]` config section. I fear that will "cause havoc" in the short term and I'm not sure I want to endure that pain. What are your thoughts on keeping the existing naming?

Agree with you on this being a potential pain.  I was afraid of the same thing while I was writing these changes.  The thought I had though was that this will at the very least need all users to reflash their mcu's (due to the new commands - and while I know users _should_ reflash on every update, a lot likely don't).  So that already bring many people to the command line and tinkering with settings, I thought.  That said, I'm not very strongly opinionated on this, and I think others like yourself have better insight into the support costs.  I wrote it that way as I think in an ideal world, having a device class like [display] or [motion_sensor] would make the most sense; but definitely not upsetting current users with non-functional config changes is a good feature too :).

As an alternate option, something similar to @kubik369 's suggestion, we could also have both section headers work (and potentially deprecate [adxl345] if we want to.

> I'm also a bit leery of using the "abstract base class" system for these sensors, as doing so tends to lead to code in one python file "peeking" into member variables in classes in another python file. (There is a blurb on this at https://www.klipper3d.org/Code_Overview.html#adding-a-host-module .) For example it looks like the new invensense.py uses `self.name` that is defined in motion_sensor.py . In my experience Python code rapidly becomes unmaintainable when member variables are accessed across .py files. (As developers often don't realize a change to a member variable in one class will break some other class.) FWIW, as a result, I generally prefer the "helper class" abstraction instead of the "base class" abstraction.

Also agree here.  I'm not a big fan of python's inheritance for the same reason.  I tried to limit variables used in subclasses to the ones in the ```__init__``` function, but of course python doesn't do anything against future subclasses overriding other vars.  I must have missed the blurb in the doc there, and in hindsight helper class probably would have been a bit better.  That said as long as future coders are responsible I think both methods should be OK and maintainable.  I could also do a refactor to a helper class in the future, and I don't think it would really impact the device classes or future code very much.

> 
> Are there functional changes between [b0aaf75](https://github.com/Klipper3d/klipper/commit/b0aaf75e59d3e0fa52f381b15e5a51124a59870a) and the latest changes? If so, I'm fine with merging in the base functionality and then discussing the refactoring in follow up PRs. (Though we should probably come to a consensus on config naming first.)

The only functional difference was putting the device to sleep when not actively taking measurements.  I don't think that's much to be concerned about.  There was a bunch of code cleanup and documentation that also got lumped in with the refactor but nothing relating to performance or functionality.

> The naming seems a little odd to me though. While an accelerometer is technically a motion sensor, I would just use accelerometer instead of motion_sensor. When a motion sensor is mentioned, I think of a proximity sensor, google does too and I believe same would go for lot of people (majority?). This might just be my opinion though.

Actually I had it originally as accelerometer but then took after Google's android "motion sensor" apis ;) https://developer.android.com/guide/topics/sensors/sensors_motion

> The section config schema like [adxl345], [mpu9250] and such falls in line with other similar cases (e.g. TMC drivers, which are defined as [tmc2209 stepper_x] rather than [tmc stepper_x] type:2209). It also simplifies the documentation (that adxl345 supports only SPI and MPU9250 - I2C). So, I would vote to keep it that way, though maybe there are some compelling reasons to change it that I may have missed?

The main reason I had changed it to [motion_sensor] was to align it with [display], which I think is kind of similar as a printer accessory that could take many forms.  The other main reason is to make it easier in the future to query and use features (e.g. does motion_sensor A have accelerometer support?  gyro support?  what API to subscribe to for this data?).  Feature querying is not supported today, though might be a useful feature in the future.  The APIs and file writing do exist, and I think it's a little better to have those be consistently named following the config.  But, I do see the merits for both ways of naming.  As you mentioned, understanding support for I2C vs SPI per device gets a little sketchy.  
As an aside, I think it'd be nice if ADXL345 and MPU9250 supported both I2C and SPI and let users decide which to use (I was actually working on a commit for mpu9250 +SPI support already - not sure what it'd take for the adxl though)
IMO, motion_sensor is a little more self-explanitory when seen in config and APIs, and also set up a little bit better for future work.  



To summarize, it seems like the biggest friction to getting this merged is really the config breaking change, so I propose another commit which would effectively make [adxl345] an alias for [motion_sensor adxl345] with chip:adxl345.  This should not break existing configs, yet still keeps the motion_sensor section header.  Thoughts on this? (@dmbutyugin @KevinOConnor 

If there are more reasons that I'm missing why in the future it'd be better to use the device name itself (i.e. ```[adxl345] / [mpu9250]```) vs using the device class ```[motion_sensor]```, then I'd be totally fine just reverting to the device naming (I'm not do-or-die on these names, though slightly favor [motion_sensor] as it is).

jakep82:
> As an aside, I think it'd be nice if ADXL345 and MPU9250 supported both I2C and SPI and let users decide which to use 

If the I2C interface of the ADXL345 is too slow to provide quality results, I don't think it makes sense to offer it as an option. It will only lead to an increase in help requests from people with poor results from IS.

bluesforte:
Just pushed a change that adds back support for [adxl345] header, and logs a warning if used.  I did a sanity check of the config and seems fine, someone else will have to verify that the settings and functionality still work as expected - specifically things that use the device name and chip commands like outputs to file (e.g. ACCELEROMETER_MEASURE gcode).

bluesforte:
> FWIW, I'm not confident we'll find many other chips that will fit a code api designed around the adxl345, and I'm not confident enough users would utilize these other chips to make it worthwhile to change the current adxl345 configuration.
> 
> My feedback is that I think we should merge the mpu9xxx functionality and only look at refactoring the config layout if we find alternate chips become popular with users.
> 
> -Kevin

Commited a revert to the old style.  While making this reversion, I actually gained a stronger taste for moving to [motion_sensor] - I think it makes the documentation a lot clearer and I'm not sure it'd be immediately apparent to a new user that adxl345 and mpu9250 do the same thing (provide accelerometer data).  That said I'd rather work on new stuff, and start seeing general adoption of the mpu devices (they seem to be strictly better than the adxl345 - higher accuracy and cheaper at least near me).  I'm also a new contributor here so I don't want to claim I know more about the users than you all! :D

Cheers
-Harry

bluesforte:
Code is ultimately just code and as long as functionality is there, I'm really not picky about it - no offense taken. While there was a bunch of time spent, hopefully it will be useful if this topic is revisited in the future (I.e. Supporting more motion sensors is useful).
FWIW, there were only a few major iterations, as I was trying to keep a reasonable review in mind - it may be easier to review one commit at a time rather than files changed overall, and keep in mind a lot of the changes post-refractor are just documentation changes.

My main request on this PR is that we treat mpuxxxx devices as first-class in klipper - I don't think there's any reason to favor the adxl in principle (other than the caveat you mentioned about i2c support on separate mcus - once this I2C support gets merged one could also add spi mode later though).

Definitely I agree with @dmbutyugin 's initial comments that it would be worthwhile to do this sort of full refactor here, though perhaps a more staged approach is more prudent. We could, merge the initial working commit https://github.com/Klipper3d/klipper/pull/5485/commits/2a7d75758714ca6ef82292ffbba94061d095f7be
And then I could open a new PR for the refactor to make it a little easier to review and we can be a little more deliberate about the discussion to change config or not (or wait for another breaking change to do so)

dmbutyugin:
> Definitely I agree with @dmbutyugin 's initial comments that it would be worthwhile to do this sort of full refactor here, though perhaps a more staged approach is more prudent. We could, merge the initial working commit https://github.com/Klipper3d/klipper/commit/2a7d75758714ca6ef82292ffbba94061d095f7be
And then I could open a new PR for the refactor to make it a little easier to review and we can be a little more deliberate about the discussion to change config or not (or wait for another breaking change to do so)

OK, I'm fine with that approach as well, if Kevin prefers it to be that way, since we agree that, in principle, the obvious code duplicates should be eliminated. I can also take a look at that later.

FWIW, perhaps we should try to improve the error message there if both `[adxl345]` and `[mcu9250]` accelerometers are added without an explicit name (Klipper would generate a pretty obscure error message in this case). Or was it already fixed there prior to that commit? And then I think you had some nice more specific instructions on how to enable I2C on the RPi so that the required I2C speed is achieved. Could you perhaps add it to the PR prior to merging?

One small comment about the configuration: if we keep the configuration as `[adxl345]` and `[mpu9250]` per that referenced commit (rather than `[motion_sensor]`), there will be pretty much no hassle for existing installations. Even though you add a new MCU command, that will not require the existing users of Klipper to reflash their MCUs. The error will be triggered only for those who try to use MPU-9250 or other new sensors. For example, I was recently experimenting with a new MCU API, and I didn't have to reflash the MCU where that API wasn't used. Which, in my opinion, speaks in favor of this configuration schema, since it won't break the existing installations and they won't have to change anything.

bluesforte:
> I'm fine with a squash and merge of [2a7d757](https://github.com/Klipper3d/klipper/commit/2a7d75758714ca6ef82292ffbba94061d095f7be) , followed by a discussion on refactoring after that.
> 
> Please confirm if you are in agreement and I'll go ahead and merge.
> 
> Thanks, -Kevin

Sounds good to me!

KevinOConnor:
Thanks.  I did a squash and merge on [2a7d757](https://github.com/Klipper3d/klipper/commit/2a7d75758714ca6ef82292ffbba94061d095f7be) and committed it as f55b9d3e.  Lets make the documentation and other changes on top.  I have some proposed changes as well which I will follow up on in a separate PR.

-Kevin

KevinOConnor:
Thanks.  I'll mark this PR as resolved.  We can use new PRs for additional work on the mpu9250 chips.

-Kevin

PR #5486, opened 2022-05-05

VoltexRB:
If Python2 is still the default python to execute python scripts, then, since numpy and matplotlib were installed for python3, explicitly calling python3 might be necessary for some.


dmbutyugin:
Thank you. However, `calibrate_shaper.py` script has a preamble
```
#!/usr/bin/env python3
```
So I'm not sure how is it possible that it will be executed with Python2? Is it possible that you are using some previous Klipper release?

PR #5491, opened 2022-05-08
gcode: Adds HANDLE_ERROR command

pedrolamas:
This adds a new `HANDLE_ERROR` command that by default does absolutely nothing(!), but when overridden, it will allow users to do specific actions in regards to the error message!

I thought on this when I experienced issue #3666: I got a "Move exceeds maximum extrusion" mid print and the print itself was not cancelled, leaving the nozzle over the printed parts still in full power (something that I think can be quite dangerous!)

So this change adds a new `HANDLE_ERROR` that when overriden, users can check the `params.ERROR` for the error message and act in accordance with it.

To fix the issue above, one could write such a macro:

```
[gcode_macro HANDLE_ERROR]
rename_existing: BASE_HANDLE_ERROR
gcode:
  {% set ERROR = params.ERROR | default("", true) %}
  {% if ERROR.startswith("Move exceeds maximum extrusion") %}
    CANCEL_PRINT
  {% endif %}
```

Ideally, we would have an `ERROR_CODE` that could be either a number or a string and work with that, but right now all we have is the message hence why I am using that above.

Jinja2 by default doesn't have regex support (though that could be added in Klipper), so regular substring is check (`startswith`) is the only way to do it.

I also made sure that if an error occurs inside the `HANDLE_ERROR` macro that it doesn't get called recursively!


KevinOConnor:
Thanks.  However, I'm concerned this could cause unexpected behaviour.  I'm not sure about the direction in general.

In particular, I'm concerned that an error raised from HANDLE_ERROR could cause an infinite loop (or result in improper error handling).  I'm also not sure that we want to encourage overriding commands.  Also, this ability wouldn't be able to distinguish between errors from gcode files and errors from the console (eg, I expect the common use-case would be to park the steppers and turn off heaters in case of an error in a gcode print, but then that error handler would also run if the user had a typo in the console, which they likely wouldn't want).

I think a simpler way to address #3666 would be to define a `[virtual_sdcard] error_commands:` parameter that is run when an error occurs in gcode prints.

Cheers,
-Kevin

pedrolamas:
Thanks @KevinOConnor, I've now made some changes following your proposed approach on the `[virtual_sdcard]`, so now there is an `on_error_gcode` that will run in case of error!

I've updated the main comment on the PR accordingly.

klippy/extras/virtual_sdcard.py
```diff
@@ -279,6 +283,12 @@ def work_handler(self, eventtime):
         self.cmd_from_sd = False
         if error_message is not None:
             self.print_stats.note_error(error_message)
+            try:
+                context = self.on_error_gcode.create_template_context()
+                context['error_message'] = error_message
+                self.on_error_gcode.run_gcode_from_command(context)
+            except e:
+                self.print_stats.note_error(str(e))
```
KevinOConnor, line 291: Thanks.  However, it's not valid to call `run_gcode_from_command()` outside of a command handler (unless the gcode mutex is obtained and held).

Also, I'm not sure why error_message is passed to the code handler - seems like additional complexity and I'm not sure what anyone will use it for.

Also, I'm not sure why print_stats.note_error() is called on an exception.  Might be better to just call `logging.exception()`.

Also, seems like this config option should have a default - maybe `TURN_OFF_HEATERS` and `M84` (same default as idle_timeout.py).

-Kevin

pedrolamas: > Thanks. However, it's not valid to call `run_gcode_from_command()` outside of a command handler (unless the gcode mutex is obtained and held).

Apologies Kevin, my grasp for the code is not complete and I forgot about that when I refactored this bit of code. I will fix that ASAP.

> Also, I'm not sure why error_message is passed to the code handler - seems like additional complexity and I'm not sure what anyone will use it for.

My reasoning here was that users might want to manage different errors in different manners.

This way, it would be up to the user if and how he wishes to use `error_message` to do different actions here.

> Also, I'm not sure why print_stats.note_error() is called on an exception. Might be better to just call `logging.exception()`.

I called `print_stats.note_error` so the error would appear on the web client. Without this, any error raised on `on_error_gcode` would not be shown to the user and just appear on the logs instead.

I'm happy to change this if you think it's better just to log the error.

> Also, seems like this config option should have a default - maybe `TURN_OFF_HEATERS` and `M84` (same default as idle_timeout.py).

Fair enough, my though on not putting anything was to keep existing behavior (do nothing!) instead of introducing a breaking change.

If you prefer that way, I will set a default of `TURN_OFF_HEATERS` and `M84` and add that to the configuration changes file so users are aware of this change of behavior.

pedrolamas: @KevinOConnor Following your comments, I've made some updates, specifically:

- removed the `error_message` parameter so now the template will run on any gcode error raised in virtual_sdcard
- changed from non-mutex guarded `template.run_gcode_from_command()` call to mutex guarded `gcode.run_script(template.render())` call

I also kept the existing "do nothing" default behavior as discussed.

KevinOConnor:
Thanks.

-Kevin

PR #5492, opened 2022-05-08
Implement HX711 load cell sensor.

JochenLoeser:
First of all thanks everybody for this amazing piece of software. I really love using it and it wanted to give something back to the community.

This is my first pull request ever. So please bear with me. I think, I have read and followed all the contributing guide lines.

I want to use a [HX711](http://en.aviaic.com/detail/730856.html) to read a load cell. This load cell will be mounted beneath my filament spool. The goal, at least for me, is to check at the start of a print if I have still got enough filament remaining to finish the print. I have seen some other pull request to use a load cell as a probe sensor. This is definitely not the possible with this code.

I have implemented a C code module to read the latest value via an mcu. I have implemented python code. The python code periodically queries the mcu for the latest value. Additionally:
* Calibration/Conversion from raw_value into user defined value. The conversion is unit agnostic. So it can be weight, percentage or whatever the user wants.
* New G code (QUERY_HX711 NAME=<sensor>) to query the latest value.
*  get_status() returns the latest raw_value and converted value.

I have tested the mcu code on an Raspberry PI and an BTT Octopus. I'm not sure if the code works on an AVR. For the communication I am creating my own clock signal. That's because it has some timing constrains and it needs 25 cycles. So I could not use SPI.  I have borrowed some code from the lcd_hd44780 source and adapted it. But in line 51 of my sensor_hx711.c there is now code to skip the delays if it is an AVR. Assuming, the AVR is clocked with 16MHz the total delay must be between 4 and 800 instructions.

Signed-off-by: Jochen Loeser <jochen_loeser@trimble.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3929,6 +3929,22 @@ adc2:
 #   above parameters.
 ```
 
+### [hx711]
+
+HX711 Load Cell Sensor (see
+[HX711 Load Cell Sensor](HX711_Load_Cell_Sensor.md)).
+
+```
+[hx711 my_sensor]
+dout_pin:
+# Pin connected to hx711's DOUT pin
+sclk_pin:
+# Pin connected to hx711's PD_SCK pin
+#calib_data: 0,0;1,1
+#   A semicolon ';' separated list of <raw_value>,<value> pairs to form the conversion.
+#   Default: 0,0;1,1
+```
+
 ## Board specific hardware support
 
 ### [sx1509]
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -607,6 +607,16 @@ above the supplied MINIMUM and/or at or below the supplied MAXIMUM.
 [TARGET=<target_temperature>]`: Sets the target temperature for a
 heater. If a target temperature is not supplied, the target is 0.
 
+### [hx711]
+
+The following command is available when a
+[hx711](Config_Reference.md#hx711)
+config section is enabled.
+
+#### QUERY_HX711
+`QUERY_HX711 NAME=<sensor_name>`: Queries last obtained value and
+raw value of the hx711 sensor.
+
 ### [idle_timeout]
 
 The idle_timeout module is automatically loaded.
--- a/docs/HX711_Load_Cell_Sensor.md
+++ b/docs/HX711_Load_Cell_Sensor.md
@@ -6,6 +6,8 @@ To use the HX711 load cell sensor, read
 [Config Reference](Config_Reference.md#hx711) and
 [G-Code documentation](G-Codes.md#hx711).
 
+HX711 is only supported on MCUs with bitbanging support.
+
 ## Calibration procedure
 
 To get sensor value you can use **QUERY_HX711** command in terminal.
--- a/docs/HX711_Load_Cell_Sensor.md
+++ b/docs/HX711_Load_Cell_Sensor.md
@@ -1,0 +1,43 @@
+# HX711 load cell sensor
+
+This document describes HX711 load cell sensor module.
+
+To use the HX711 load cell sensor, read
+[Config Reference](Config_Reference.md#hx711) and
+[G-Code documentation](G-Codes.md#hx711).
+
+## Calibration procedure
+
+To get sensor value you can use **QUERY_HX711** command in terminal.
+
+1. Load the cell with a known weight.
+
+2. Query the latest raw value.
+
+3. Repeat steps 1 and 2 at least one time with a different weight.
+
+4. When you have enought calibration points you can create the calibration.
+The format is a list of <raw_value>,<value> pairs separated by semicolons ';'.
+e.g.: 0,10;100;1010;1050;10510
+
+5. Save the calibration data in config parameter `calib_data`.
+
+The conversion is unit agnostic. This means you can convert the
+raw values in grams, pounds, or even percentage if you like.
+
+For example:
+Let's assume you are weighing your filament spool with this sensor.
+You gather calibration data with an empty spool(520) and with a full(18657) spool.
+When you enter `520,0;18657,100` as `calib_data`, you will receive a
+percentage value indicating how full the spool is.
+
+## Example
+A simple example to read the latest value of 'hx711 test':
+
+```
+[gcode_macro hx711]
+gcode:
+  {% set res = printer["hx711 test"] %}
+  { action_respond_info(res.value) }
+  { action_respond_info(res.raw_value) }
+```
--- a/klippy/extras/hx711.py
+++ b/klippy/extras/hx711.py
@@ -1,0 +1,87 @@
+# Support for hx711 load cell sensor
+#
+# Copyright (C) 2022  Jochen Loeser <jochen_loeser@trimble.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+class SensorHx711(object):
+    def __init__(self, config):
+        try:
+            import numpy as np
+        except:
+            raise config.error("HX711 requires numpy module")
+        self.printer = config.get_printer()
+        self._name = config.get_name().split()[-1]
+        self._last_value = None
+        # Create polynom
+        try:
+            data_str = config.get('calib_data', '0,0;1,1')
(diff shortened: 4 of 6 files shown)


Sineos:
Also see https://github.com/Klipper3d/klipper/pull/4738
and:
https://klipper.discourse.group/t/strain-gauge-load-cell-based-endstops/2134

Although not identical (different goals) the topics seem sufficiently related to warrant a common approach to avoid code duplication and "doing the same a bit differently".

KevinOConnor:
Thanks.

I have some questions on the implementation.  What is the purpose of the state-machine?  Wouldn't it be simpler to wakeup once, bit-bang out the sequence, and report the results back to the host?

I'd also be interested in hearing feedback from @mhier and @clearchris .

-Kevin

mhier:
I will test this on my AVR-based printer. I am very keen to integrate this with my load-cell probing algorithm (see #4738). This looks like a very good start, thank you!

So far I have not looked so deeply into the microcontroller parts of Klipper, so I cannot really judge that part of the implementation. I don't fully understand how the timing is realised. What is the purpose of the `rest_ticks` argument of the MCU command? It seems to be filled with a constant value on the Python side. Can't we directly get the number of ticks per millisecond in the MCU code?

What I don't understand about the state machine: You are reading the 24 bits in a loop in the same state. Why don't you do that directly in the `if(hx711_rdy(hx)) {}` branch instead of using a separate state for it? Same question for the reply: why do you first transition into a new state (i.e. wait for a new timer event) before sending the value to the host? That would simplify the state machine to two states.

On the high level side, I would suggest to provide a standardised interface like for any ordinary ADC. That way it would be possible to use the HX711 driver in high level modules (like the load cell probe) same way as other ADC drivers (load cells can also be read with general purpose ADCs when using an appropriate analogue electronics).

It would be really great if we could achieve a standardised interface also on the low level side in the MCU code, so something like the adc endstop (see https://github.com/garethky/klipper/commits/adc-endstop) can be used with a variety of hardware as well, but I think this will be more work and goes beyond the scope of this PR.

mhier:
One more think I just noticed: You are setting SCK to high after receiving the 24th bit, and back to low only when the next transfer is requested. This will violate the specifications which say SCK high time should not exceed 50 microseconds.

You should ideally also make the conversion gain configurable. This works by sending 25, 26 or 27 clock signals. This will correspond to a gain of 128, 32 or 64, respectively (note the strange order). You must not send only 24 clock signals, also 28 are not allowed.

I know the chip might be tolerant against such errors (I have made a preliminary driver based on the SPI bus implementation which always sends 32 clock signals and does not wait for the ready signal - that works somewhat but sometimes I get garbage), but it is better to stick to the specifications to avoid problems e.g. with different chip revisions or so.

clearchris:
> Wouldn't it be simpler to wakeup once, bit-bang out the sequence, and report the results back to the host?

Well, that's an interesting question.  For our current implementation we used this code as a base, which sets a timer for each rise and fall of the clock pin.  If we can just wake up once, run it through while accepting any interrupts, that would certainly simplify things and be preferred.  That said, an interruption that lasts longer than 50-60us could send the hx711 into sleep mode, and would not be ideal.

https://github.com/clearchris/klipper/commit/45779825cfc2b89b5016f3676ea2d97aed647ed2#diff-3826f42c2d5a441c6fb0c378099710dd3c49bfa7c0a04216a246a3618bb8deda

What do you think @KevinOConnor, would that be your preferred way to see it implemented?  Are there any pitfalls we should watch out for with this type of implementation?

> On the high level side, I would suggest to provide a standardised interface like for any ordinary ADC. That way it would be possible to use the HX711 driver in high level modules (like the load cell probe) same way as other ADC drivers (load cells can also be read with general purpose ADCs when using an appropriate analogue electronics).

Also interesting.  I'd question the use of the hx711 as a homing probe, not because it couldn't be done, but because there are plenty of other ADC chips which would perform the task better, with fewer compromises.  Yes the hx711 is possibly the most common and cheapest out there, but for a few more dollars, a more suitable chip can be used for homing (see @Sineos 's referenced thread).  

Towards @mhier 's original point though, homing seems to have much higher requirements than a simple ADC interface.  However, I very much appreciate any standardization that can be done, and have implemented an interface for using load cells in weighing that can accept any klipper ADC.  See https://github.com/clearchris/klipper/blob/weight-scale/klippy/extras/weight_scale.py  There's also a patch working with mainsail to display the scale output (  https://github.com/mainsail-crew/mainsail/pull/745 ) Simple weighing functions have only basic requirements and do work fine if a klipper ADC interface is implemented.  Most other implementations I have seen tie the function of the scale or the probe tightly to the hardware.

clearchris:
BTW @JochenLoeser if you are open to collaboration, @cpiercedev and myself are working on getting a branch going along the lines of @mhier 's suggestion.  We have swapped out the ADC underneath the weight_scale interface twice already with no issue, and have mainsail integration already completed.  Cpiercedev also has an integration working with SuperSlicer.

Your hx711 mcu code is much cleaner than what we have been working on, if we could get your code compliant with what Kevin would like to see, and implement an ADC interface like mhier suggested, we could have a filament scale solution end to end from SuperSlicer all through to Mainsail in pretty quick order.

clearchris:
> One more think I just noticed: You are setting SCK to high after receiving the 24th bit, and back to low only when the next transfer is requested. This will violate the specifications which say SCK high time should not exceed 50 microseconds.

Could be intentional.  Setting SCK high for more than 60us puts the chip into sleep mode.  This is fine if you aren't requesting a conversion.

mhier:
> > One more think I just noticed: You are setting SCK to high after receiving the 24th bit, and back to low only when the next transfer is requested. This will violate the specifications which say SCK high time should not exceed 50 microseconds.
> 
> Could be intentional. Setting SCK high for more than 60us puts the chip into sleep mode. This is fine if you aren't requesting a conversion.

All right, but one needs to send at least 25 clock periods, not only 24.

Anyway, I think for many applications we need a continuous read out. I think this is also required for the standard ADC interface.

Still, I went ahead and tested the code on my AVR (ATMega 2560) based printer. It works so far, I can read out the load cell values.

JochenLoeser:
Ok so I try to answer all the questions:

@KevinOConnor  + @mhier About the state machine:
This is my first code for this framework. I tried to follow the guidelines, especially: "[...]do any work that lasts more than a few micro-seconds."
The HX_READING state takes at least 12us. So I tried to separate the previous and next part of the communication into other states to minimize the impact.

@mhier Setting SCK high after finishing:
As @clearchris has already pointed out this puts the chip into sleep mode. That's also the reason I have to use a delay method in the reading code part. Otherwise I couldn't guarantee the timing constrains.
 
@mhier Adjustable Gain:
That was a design choice. My understanding is, that you don't only choose the gain factor but also the channel(but not independently). This means in theory a user could setup both channels and this means the complexity explodes.
* You have to check if two config definitions use the pins but for different channels.
* You have to keep track which measurement matches which configuration.
* At the end of a measurement you have to figure out which channel/gain to use next.
Side note: I even thought about using the clock signal for multiple sensors.

@mhier The 25th clock signal:
Take a look at [sensor_hx711.c line 179](https://github.com/JochenLoeser/klipper/blob/master/src/sensor_hx711.c#L179).

@clearchris : Sure I  would love to collaborate.

clearchris:
> This means in theory a user could setup both channels and this means the complexity explodes.
@JochenLoeser  Yep, I have been thinking about this for a while...  It's the only part we haven't implemented yet.

>Side note: I even thought about using the clock signal for multiple sensors.
We have this working in our implementation.   @cpiercedev has tested up to 8 hx711 chips on a rpi2040 IIRC.

We are currently on the klipper discord, but could move if you have an alternate place you would like to chat.  
Here's the link to discord:  https://discord.klipper3d.org/
I'm on there as clearchris.  My email is clearchris-at-hotmail-com (obfuscated for spam bots)
Your PMs are disabled on github.

KevinOConnor:
Thanks for working on this.  I do think it would be good to add hx711 support to Klipper.  I do think we should address some of the comments raised here before merging though.

Cheers,
-Kevin

trackhacs:
are here updates on this?

PR #5493, opened 2022-05-09
Fixed the I2C interface code of STM32G0MCU and added SET_LED_ON, SET_…

gghyoo:
Fixed the I2C interface code of STM32G0MCU and added SET_LED_ON, SET_LED_OFF, SET_DISPLAY_ON, SET_DISPLAY_OFF commands

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ out
 .config
 .config.old
 klippy/.version
+scripts/install-ubuntu-20.04.sh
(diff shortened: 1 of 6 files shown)


Sineos:
Thanks for the contribution. Please also check against https://www.klipper3d.org/CONTRIBUTING.html. As a minimum you will need a signed-off line as per point 3.

It also probably makes sense to split the PR into multiple PRs. As far as I can tell this PR is mixing changes / addition in at least 4 areas:
* Bug ix for I2C
* LED features
* Display features
* Installation script

On the one hand separate PRs will make reviewing easier and if issues arise then you will have dedicated commits to check / bisect.

PR #5497, opened 2022-05-10
hide USB option when using stm32f1 with HSI

kaidegit:
When using stm32f1 with internal clock, it is not recommend to use USB. What's more, existing code(may set main clock to 64MHz) may not generate a right clock to USB.

See STM32CubeMX
F1-HSI-Cube
or DataSheet 2.3.24 (DS5792 Page 22)
F1-HSI-DS

Signed-off-by: Kai Ye <2857693944@qq.com>


KevinOConnor:
Thanks.  I implemented this a slightly different way - commit 19a478de.

-Kevin

PR #5502, opened 2022-05-13
webhooks: poll file descriptor before writes

Arksine:
This attempts to resolve dropped connections between Moonraker and Klippy when large amounts of data sent over the Unix Socket.  This tends to affect users with large configurations when a client queries the `configfile` object...the socket's buffers are overwhelmed by the amount of data being transferred and the simple delay mechanism isn't enough to compensation.

The most significant change is that to the reactor.  A `register_writer()` method has been added, which allows a callback to be executed when a file descriptor is ready to write.  Each call to `register_writer()` adds a callback to a list, they are popped off as the socket notifies that writes are available.  When the list is empty the reactor will no longer poll the file descriptor for writes.  When a file descriptor is unregistered it notifies all pending writers by passing `None` to the eventtime argument.


KevinOConnor:
Thanks.  I agree it would be good to extend the reactor to support detection of writeable fds and extend webhooks to use that functionality.  I'm a little leery of some possible "corner cases" in this implementation.

In past experiences with "socket servers" I've seen some common failures that result in clients crashing the server.  In particular, I've seen two common failures: a client stops reading responses and the server eventually runs out of memory buffering them, and a client sends more requests than it can read and thus also causes the server to run out of memory.

Given that, I think it's good practice for the server to stop reading new commands from a client that is not reading responses, and to disconnect any client that has not cleared it's buffer in a reasonable amount of time (eg, 5 seconds).

I'll try to mock up some example code to demonstrate some solutions.

-Kevin

KevinOConnor:
FYI, I put up an example branch at https://github.com/KevinOConnor/klipper-dev/tree/work-pollwrite-20220523 .  This isn't intended as a working solution - it's for discussion on alternatives.

-Kevin

Arksine:
Hi Kevin, I just took a look at what you have and I think its a good start.  Disabling the read fd in Klipper may require that I make some adjustments in Moonraker to prevent it from disconnecting on my end, it uses asyncio streams to communicate with the unix socket so I'll have to see how it handles `EAGAIN` and `EWOULDBLOCK`.


Aside from polling writers in the reactor there is something else we can do to mitigate this issue.   "Coding in the config" is becoming more common, resulting in an extremely large configfile response.  We could register an endpoint in `configfile.py` that allows clients to query the configuration piece-wise.  While we can't remove `configfile` from `get_status()` without breaking macros and existing front ends, we could encourage front end developers to use the alternate endpoint.

KevinOConnor:
> Disabling the read fd in Klipper may require that I make some adjustments in Moonraker to prevent it from disconnecting on my end

Yeah - in the general case - that's one disadvantage of turning off reading - it can result in a lockup if the client is blocking on a write and the server is no longer reading because it is waiting for the client to read.  If that is a big issue we should be able to keep reading in klippy as long as we have a reasonable timeout to disconnect clients that aren't keeping up.

> We could register an endpoint in configfile.py that allows clients to query the configuration piece-wise.

I'd say this shouldn't be necessary.  The code should be able to move hundreds of megabytes within a few seconds - well beyond any reasonable config file.  We can also tweak the busy disconnect timeout to give clients a few more seconds if that is needed.

Cheers,
-Kevin

samwiseg0:
Just wanted to chime in here after talking with Arksine. I am experiencing issues with this more and more. My issue seems to manifest itself with mainsail more than fluidd. 

I would be willing to do some testing with this if needed since I can pretty easily replicate the issue.

newphreak:
I'm also available if any testing is needed. Jon#4961 on the Voron discord.
I think this bug is being provoked more and more with people running the klicky macro and the likes.

Arksine:
Just to chime in, I rebased this branch and pushed an alternative "quick and dirty" patch if we want to hold off on modifying the Reactor.  The write will pause for 2ms on error, with 1000 retries, effectively making the write time out 2 seconds rather than 10ms.   I believe that this would resolve the issue in the interim while we work on a more robust solution.

Alternatively, I don't see anything wrong with your implementation Kevin.  I looked at the asyncio streams implementation, as long as the socket is open it will continue to retry writing if it errors out with EWOULDBLOCK, so there shouldn't be an issue with unregistering the reader in Klipper.

KevinOConnor:
I don't have a strong preference on way forward.

@samwiseg0 , @newphreak - if you pull the code at https://github.com/KevinOConnor/klipper-dev/tree/work-pollwrite-20220523 does everything work properly?

-Kevin

newphreak:
That does seem to go into the loop again. Arksine's "quick and dirty" have worked flawlessly so far for me.
[klippy.log](https://github.com/Klipper3d/klipper/files/9015154/klippy.log)
[moonraker.log](https://github.com/Klipper3d/klipper/files/9015155/moonraker.log)
.

KevinOConnor:
FYI, I committed the work-pollwrite-20220523 code to the master Klipper branch.

-Kevin

PR #5506, opened 2022-05-15
Implement option for retraction after homing

pauhull:
When using sensorless homing, the carriage will be pressed into the frame after being stopped (as documented [here](https://github.com/Klipper3d/klipper/blob/master/docs/TMC_Drivers.md#using-macros-when-homing)). The suggested fix is to create a macro for homing or using homing overrides. This is a bad approach because it harms maintainability, disables compatibility with firmware features (such as safe z homing), and is not beginner-friendly. This PR adds a `homing_final_retract` setting to the stepper configuration. It uses the regular homing retract but stops before performing the actual second home. This disables second homing, but it is not needed when using sensorless homing anyway. Tested the code on my machine for a few days now and had no problems.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -195,6 +195,11 @@ position_max:
 #   better to use the default than to specify this parameter. The
 #   default is true if position_endstop is near position_max and false
 #   if near position_min.
+#homing_final_retract:
+#   Set to true to enable retraction after homing. This is useful for 
+#   sensorless homing, where you want to prevent the stepper from 
+#   pressing the carriage into the frame after homing. Uses 
+#   homing_retract_dist and homing_retract_speed.
 ```
 
 ### Cartesian Kinematics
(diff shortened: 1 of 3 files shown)


pauhull:
Yeah, you've got a valid point there, but I think it's a fair tradeoff to make. The implementation is quite compact because it re-uses some existing settings (that wouldn't be needed anyway). Also I prefer my printer.cfg waayy more with this one option set instead of having to write a 30 lines long macro that I wouldn't need otherwise. With most driver boards having TMCs nowadays I think it's reasonable to add this extra option and it will decrease overall complexity. IMO macros are a kind of last resort for when you want to do really specific things with your printer (and they are very good at that), but I think a widespread feature such as sensorless homing justifies a proper firmware implementation. 

Quick note to my PR: I know it's quite messy, I had a hard time with the whitespace checker. But I think the concept is clear, implementation could be done by someone who actually knows what he's doing :D 

Thanks

PR #5514, opened 2022-05-18
Update Config_Reference.md to correct default z-hop speed

camerony:
The default z-hop speed is actually 15 mm/s according to the code in safe_z_home.py


camerony:
added signed-off by line

KevinOConnor:
Thanks.

-Kevin

PR #5517, opened 2022-05-20
Add CAN support for STM32H743

JamesH1978:
Added the CAN interface options for STM32H743 based boards that use this connection

Signed off by: James Hartley <james@hartleyns.com>


KevinOConnor:
Thanks.  Can you provide a little more detail on how this change was tested?

-Kevin

JamesH1978:
Not tested as of yet :) I have an incoming SKR3 board i intend to use on CAN, but the option is not there for CAN in this series of chip, so I thought I'd PR to enable it. 

Its fine though, I can wait till I get the board and test it before and update the PR, not a problem.

Thanks
James

bondus:
STM32H has a CAN-FD interface that is more likely similar to what you find in the STM32G. 
Have a look at PR#5488

PR #5519, opened 2022-05-21
mkdocs: Replace TOC in Translations

dingyifei:

### DIFF
--- a/docs/_klipper3d/build-translations.sh
+++ b/docs/_klipper3d/build-translations.sh
@@ -27,6 +27,16 @@ while IFS="," read dirname langsite langdesc langsearch; do
   new_docs_dir="${WORK_DIR}lang/${langsite}/docs/"
   locale_dir="${TRANS_DIR}/docs/locales/${dirname}"
 
+  # read toc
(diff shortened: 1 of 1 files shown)


dingyifei:
Changes to klipper-translations are already merged

KevinOConnor:
Thanks.  Alas, I'm not sure I understand what this change does.  Can you give a high-level overview?  What's the problem being solved and how the solution works?

-Kevin

dingyifei:
> Thanks.  Alas, I'm not sure I understand what this change does.  Can you give a high-level overview?  What's the problem being solved and how the solution works?
> 
> 
> 
> -Kevin

A part of the table of content is not translatable because they are not automatically generated from corresponding xxx.md. They are in the mkdocs.yml. I made Navigation.md and Navigation.po in killer-translations to make it translatable. Then this script basically read the specific lines of Navigation.md for languages and replace the strings in mkdocs.yml with the line read.

KevinOConnor:
Okay, I understand.  Thanks.

-Kevin

Issue #5008, opened 2021-12-08
Incorrect parsing of M117, M118 commands when message starts with a digit


blalor:
`M117 ohai` and `M118 ohai` are properly handled.  `M117 0hai` and `M118 0hai` are not, with the `0` (I think actually `0h`) being parsed as a number.

The following is what appears in the Fluidd console:
```
06:17:57 $ M118 0hai
06:17:57 // Unknown command:"M118 0"
06:18:07 $ M117 0hai
06:18:07 // Unknown command:"M117 0"
06:18:18 $ M118 ohai
06:18:18 echo: ohai
```

`klippy.log` just says 
```
Unknown command:"M118 0"
Unknown command:"M117 0"
```

blalor:
Still present.

```
07:04:12 $ m115
07:04:12 // mcu build version:
         // mcu: v0.10.0-223-g4c8d24ae
07:04:12 // FIRMWARE_VERSION:v0.10.0-223-g4c8d24ae FIRMWARE_NAME:Klipper
07:04:23 $ M118 0hai
07:04:23 // Unknown command:"M118 0"
```

fleinze:
I experience the same issue. Seems it was introduced in PR #4727

The problem with the M117-Command is that any message starting with characters other than [A-Z_*/] will be ommitted.
`M117 5% complete` will display "complete" without "5%"

Sineos:
@pedrolamas maybe you can cross-check

pedrolamas:
Ok, found the issue and the good news (for me!) is that it is not related to that PR!

The problem is with [this regular expression](https://github.com/Klipper3d/klipper/blob/02d5f9754fc7f9e493c8bc5a6418e2a9ea9d7ae1/klippy/gcode.py#L175) that is used for command parsing.

This is ok:

```python
>>> args_r.split("M118 TEST")
['', 'M', '118 ', 'TEST', '']
```

From the above, the code will use index 1 and 2 and build "M118" as the command.

But if the first parameter starts with a number:

```python
>>> args_r.split("M118 0TEST")
['', 'M', '118 0', 'TEST', '']
```

The result of index 1 and 2 will be "M118 0" and that is obviously incorrect!

KevinOConnor:
Unfortunately, #4747 did introduce a regression.  (There is special handling of M117 in gcode.py to ensure the correct handler is called, but it doesn't fixup `gcode._command`.)

It should be fixed now (commit f97fd7c6).

-Kevin

[...]

PR #5528, opened 2022-05-26
Handle renamed M117 and M118 commands that start with numbers or special characters

### LINKED ISSUE #5527: Issue #5008 persists when M117 and M118 are renamed
When using a gcode macro to redefine the behavior of M117 or M118, the renamed base call to M117 or M118 will be affected by the bug mentioned in issue #5008 

Steps to reproduce:

1. Add the gcode macro example redefining M117 given in the documentation to `printer.cfg` and restart klipper
```
[gcode_macro M117]
rename_existing: M117.1
gcode:
    M117.1 { rawparams }
    M118 { rawparams }
```
2. Run the command `M117 test`. It will work as expected
3. Run the command `M117 123 test`. It will raise an error "Unknown command:"M117.1 123"

[klippy log file](https://github.com/Klipper3d/klipper/files/8781362/klippy.log)

tux2603:
Closes #5527

### DIFF
--- a/klippy/gcode.py
+++ b/klippy/gcode.py
@@ -30,6 +30,11 @@ class GCodeCommand:
         command = self._command
         if command.startswith("M117 ") or command.startswith("M118 "):
             command = command[:4]
+        elif command.startswith("M117.") or command.startswith("M118."):
+            # If there's a space, it means that there was a special character
+            space_index = command.find(" ")
+            if space_index >= 0:
+                command = command[:space_index]
         rawparams = self._commandline
         urawparams = rawparams.upper()
         if not urawparams.startswith(command):
@@ -282,6 +287,15 @@ class GCodeDispatch:
             if handler is not None:
                 handler(gcmd)
                 return
+        elif cmd.startswith("M117.") or cmd.startswith("M118."):
+            # Handle renamed M117/M118 gcode with numeric and special characters
+            space_index = cmd.find(' ')
+            if space_index >= 0:
+                cmd = cmd[:space_index]
+            handler = self.gcode_handlers.get(cmd, None)
+            if handler is not None:
+                handler(gcmd)
+                return
         elif cmd in ['M140', 'M104'] and not gcmd.get_float('S', 0.):
             # Don't warn about requests to turn off heaters when not present
             return
@@ -351,7 +365,6 @@ class GCodeDispatch:
             if cmd in self.gcode_help:
                 cmdhelp.append("%-10s: %s" % (cmd, self.gcode_help[cmd]))
         gcmd.respond_info("\n".join(cmdhelp), log=False)
-
 # Support reading gcode from a pseudo-tty interface
 class GCodeIO:
     def __init__(self, printer):


JamesH1978:
Please see https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially the format of the PR and the signed off by part that certifies that you agree to the [developer certificate of origin. WIthout the correct format, your PR is unlikely to be accecpted/reviewed.

Thanks
James

tux2603:
@JamesH1978 I squashed the commits together and added the signed off on the new commit

KevinOConnor:
Thanks.  However, I'm very leery of making any changes to "old gcode".  I don't think the code should have a special behavior for commands starting with `M117.` or `M118.` .

If desired, we could create a new "extended gcode command" to set the display strings (eg, `SET_DISPLAY_TEXT MSG=...`).  That is, I'm fine with adding new functionality using newer human readable commands; I don't think we should "add new features to old gcode".

Cheers,
-Kevin

tux2603:
@KevinOConnor This isn't adding any special behavior or new behavior to old gcode, it's just patching a bug with how the gcode is interpreted. The bug was originally reported in #5008, but it was only partially patched. This pull request extends the patch logic that you added to `M117` and `M118` gcodes in f97fd7c6e392e376bd1552afdd39a2150d3e21d0 to `M117.*` and `M118.*` gcodes.

Arksine:
I believe that Kevin is proposing an alternative method that could be used to successfully override M117.  This would require a relatively simple change to `display_status.py`, ie:
```diff
diff --git a/klippy/extras/display_status.py b/klippy/extras/display_status.py
index be7a58b8..270a8462 100644
--- a/klippy/extras/display_status.py
+++ b/klippy/extras/display_status.py
@@ -16,6 +16,9 @@ class DisplayStatus:
         gcode = self.printer.lookup_object('gcode')
         gcode.register_command('M73', self.cmd_M73)
         gcode.register_command('M117', self.cmd_M117)
+        gcode.register_command(
+            "SET_DISPLAY_TEXT", self.cmd_SET_DISPLAY_TEXT,
+            desc=self.cmd_SET_DISPLAY_TEXT_help)
     def get_status(self, eventtime):
         progress = self.progress
         if progress is not None and eventtime > self.expire_progress:
@@ -39,6 +42,10 @@ class DisplayStatus:
     def cmd_M117(self, gcmd):
         msg = gcmd.get_raw_command_parameters() or None
         self.message = msg
+    cmd_SET_DISPLAY_TEXT_help = "Set or clear the display message"
+    def cmd_SET_DISPLAY_TEXT(self, gcmd):
+        self.message = gcmd.get_str("MSG", None)
+
```
Then you could override M117 like so:
```
[gcode_macro M117]
rename_existing: M117.1
gcode:
  {% if rawparams %}
    SET_DISPLAY_TEXT MSG="{rawparams|replace('"', '\\"'}"
  {% else %}
    SET_DISPLAY_TEXT
``` 
The `replace` probably looks a bit strange, it should make sure double quotation marks in the message are escaped.  I just rattled this off, its untested.  If interested please feel free to give it a test and modify the PR (or submit a new one).

There is already an extended gcode for handling M118, `RESPOND`.

tux2603:
That would be an option, if we go that way the documentation will need to be updated to reflect the change. Currently there's an [example](http://www.klipper3d.org/Command_Templates.html#the-rawparams-variable) that overrides the `M117` gcode to demonstrate the use of `rawparams`. If we don't want to fix the bug, we should probably either find a different example for `rawparams` or add a note to the documentation about potential unexpected behavior.

Another potential solution that would be to hunt down and root up the bug at its source, which would require a bit more digging around in the code

tux2603:
I dug into to code, and it looks like the culprit at the root of the bug is in the `_process_commands` function. Essentially the function parses all commands the same which isn't the case in other firmwares, leading to the unexpected behavior. Since the setup guide has you use the Marlin gcode flavor, I looked through their firmware to find how `M117` and `M118` are handled:

```cpp
if (letter == 'M') switch (codenum) {
    TERN_(GCODE_MACROS, case 810 ... 819:)
    TERN_(EXPECTED_PRINTER_CHECK, case 16:)
    case 23: case 28: case 30: case 117 ... 118: case 928:
      string_arg = unescape_string(p);
      return;
    default: break;
  }
```

If the command is an `M117` or `M118` gcode, the parser stores the rest of the command as a single argument and exits early. The logic is pretty straight forward, and I should be able to get it added in to Klipper's parse function without any major issues

Arksine:
Conceptually it doesn't appear all that different from Klipper's handling of special commands, it just performs the handling at a different stage.  The difference is that Klipper is capable of registering custom gcode commands and overriding/renaming existing commands...correct me if I'm wrong but I don't believe Marlin has this ability.  If it did then I believe that the code above would fail to correctly parse renamed/custom commands.

tux2603:
You can add custom gcodes to marlin, but it requires you to write the c++ code to handle it and to recompile the firmware.

For klipper, I agree that it would be conceptually identical to handle the special cases in either place, the only major difference would be reducing repeated code. Currently that code is only repeated twice, so I don't see any major issue with keeping it outside of `_process_commands`

PR #5533, opened 2022-05-29
Creality CR10 v3 with BLTouch support config.

CoreRasurae:
Configuration for Stock Creality CR10 V3 with BLTouch and runout filament sensor (tested)


JamesH1978:
Please see https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md for more information on how you should format a PR, specifically the signed-off-by that ertifies that you agree to the developer certificate of origin.

That being said, the repo is for stock boards and printers, we do not usually merge configs that are not as the printer came out of the box from the factory. So can you confirm that CR10v3 ships with a BLTouch and a runout sensor? if not, then i would suggest you post this in the klipper discourse https://klipper.discourse.group/ there is a place in there for configs that are not stock.

Thanks
James

CoreRasurae:
Ok, I'll have a look in those links regarding the submission procedures, thanks.
All Creality CR10 V3 units ship with a runout sensor and support for the optional BLTouch, however the BLTouch is not included, it is just shipped as ready for BLTouch and has an easy installation process, so that it is a factory supported configuration. So can I include the BLTouch too? or must it be removed in this particular case?

JamesH1978:
In that case, I believe the convention is to leave the config for them in, but comment the lines out, so as not to confuse future users. If it is a factory option and the pins exist for dedicated ports, then this should be fine..

Thanks

James

CoreRasurae:
Yes, it is a factory option and it has dedicated ports for it in the mainboard.

Ok, I'll leave the BLTouch specific configuration commented.

Thanks,
Luís

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
+#Stock Creality CR-10 V3 with BL-Touch and runout filament sensor
+[stepper_x]
+step_pin: PF0 #ar54
+dir_pin: PF1 #ar55
+enable_pin: !PD7 #!ar38
+microsteps: 16
+rotation_distance: 40 
+#Step distance is deprecated and will be removed in future releases
+#step_distance: .0125 <=> 0.0125 mm * 200 steps *16 microsteps = 40 mm
```
Sineos, line 9: Please remove this and similar comments.

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
+#Stock Creality CR-10 V3 with BL-Touch and runout filament sensor
```
Sineos, line 1: Add information how to build the firmware. Also include the board type and revision

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+[stepper_x]
+step_pin: PF0 #ar54
+dir_pin: PF1 #ar55
```
Sineos, line 4: Does this printer have a 8bit board? If not, remove all the #arXX comments.

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+
+[safe_z_home]
+home_xy_position: 104.25,147.6 # Change coordinates to the center of your print bed relative to the probe
+speed: 80
+z_hop: 10               
+z_hop_speed: 10
+
+[extruder]
+max_extrude_only_distance: 150.0
+max_extrude_cross_section: 1.00 #For start G-Code that was set-up in ideaMaker
```
Sineos, line 46: Remove

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+step_pin: PA4 # ar26
+dir_pin: !PA6 # !ar28
+enable_pin: !PA2 # !ar24
+microsteps: 16
+rotation_distance: 7.7201944 # 16 microsteps * 200 steps/rotation / steps/mm
+#Correction formula is new_rotation_distance = old_rotation_distance * mmsExtracted / 100.0
+nozzle_diameter: 0.400
+filament_diameter: 1.750
+min_extrude_temp: 166 # 166C is the minimum extrude temperature, this allows to print at 170C, given possible temp fluctuation with PID algorithm
```
Sineos, line 55: What purpose does this comment serve?
`min_extrude_temp` is a safety setting to avoid damaging the extrusion system when trying to print cold

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+pid_kd = 104.458
+min_temp: 0
+max_temp: 255
+
+[heater_bed]
+heater_pin: PH5 #ar8
+sensor_type: ATC Semitec 104GT-2
+sensor_pin: PK6 #analog14
+control: pid
+#Stock PID configuration taken from Marlin
```
Sineos, line 71: Remove

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+pid_Kd: 954.96
+min_temp: 0
+max_temp: 130
+
+[fan]
+pin: PH6 #ar9
+
+[mcu]
+serial: /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
+#pin_map: arduino
```
Sineos, line 83: Remove

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+[bltouch]
+sensor_pin: ^PD2 #^ar19
+control_pin: PB5 #ar11
+set_output_mode: 5V
+pin_move_time: 0.4
+#   The amount of time (in seconds) to wait for the BLTouch pin to
+#   move up or down. The default is 0.680 seconds.
+stow_on_each_sample: False
+probe_with_touch_mode: False
+#pin_up_touch_mode_reports_triggered: False
```
Sineos, line 110: Remove if not needed

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+click_pin: ^!PC2 #^!ar35
+
+
+[bltouch]
+sensor_pin: ^PD2 #^ar19
+control_pin: PB5 #ar11
+set_output_mode: 5V
+pin_move_time: 0.4
+#   The amount of time (in seconds) to wait for the BLTouch pin to
+#   move up or down. The default is 0.680 seconds.
```
Sineos, line 107: Remove

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+#move_check_distance: 5.0
+#   The distance (in mm) along a move to check for split_delta_z.
+#   This is also the minimum length that a move can be split. Default
+#   is 5.0.
+mesh_pps: 2,2
+#   A comma separated pair of integers (X,Y) defining the number of
+#   points per segment to interpolate in the mesh along each axis. A
+#   "segment" can be defined as the space between each probed
+#   point. The user may enter a single value which will be applied
+#   to both axes.  Default is 2,2.
```
Sineos, line 150: Remove

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+algorithm: bicubic
+#   The interpolation algorithm to use. May be either "lagrange"
+#   or "bicubic". This option will not affect 3x3 grids, which
+#   are forced to use lagrange sampling.  Default is lagrange.
+bicubic_tension: .2
+#   When using the bicubic algorithm the tension parameter above
+#   may be applied to change the amount of slope interpolated.
+#   Larger numbers will increase the amount of slope, which
+#   results in more curvature in the mesh. Default is .2.
```
Sineos, line 159: Remove at is default anyway.

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+relative_reference_index: 24
+#   A point index in the mesh to reference all z values to. Enabling
+#   this parameter produces a mesh relative to the probed z position
+#   at the provided index.
```
Sineos, line 163: What is the reason for setting a RRI?

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+
+[pause_resume]
+recover_velocity: 50
+
+[filament_switch_sensor fil_runout_sensor]
+pause_on_runout: True
+#   When set to True, a PAUSE will execute immediately after a runout
+#   is detected. Note that if pause_on_runout is False and the
+#   runout_gcode is omitted then runout detection is disabled. Default
+#   is True.
```
Sineos, line 173: Remove all comments from the official documentation. Will create a maintenance burden documentation is updated.

printer-creality-cr10v3-bltouch.cfg
```diff
@@ -0,0 +1,229 @@
...
+#pause_delay: 0.5
+#   The amount of time to delay, in seconds, between the pause command
+#   dispatch and execution of the runout_gcode. It may be useful to
+#   increase this delay if OctoPrint exhibits strange pause behavior.
+#   Default is 0.5 seconds.
+
+
+[gcode_macro G29]
+gcode:
+    BED_MESH_CALIBRATE
```
Sineos, line 200: Remove

Sineos (changes requested):
Many thanks for your contribution. Please check the review comments.

And, as @JamesH1978 already mentioned, the signed-off lines as per https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md No. 3 is mandatory

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Comments were removed

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Done.
The printer comes with a Creality 3D v2.5.2 8-bit mainboard with an ATMega2560. The information was appended to the comment.

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Yes, it does have an 8-bit mainboard.

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Done

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: The min. extrude temp was set to that values, because I once printed a PLA temp tower that had a minimum temperature of 170ºC and as such it would cause a Klipper error, due to the PID algorithm stabilization procedure reaching a temperature slightly below 170ºC. I will remove this configuration line, since it was a leftover from a print test.

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Done

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: I believe this was to get the measurement referenced in respect to the middle of printer bed, but I will remove it from the configuration too

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Done

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Ok, I'll remove, but this is very relevant for compatibility with Marlin GCODE

printer-creality-cr10v3-bltouch.cfg
CoreRasurae: Removed

CoreRasurae:
Sorry I did a lot of commits to cleanup and implement the suggested modifications to the configuration. If you wish I can delete this PR and create a new clean one, otherwise if you think the Squash on merge is ok, let me know.
Thanks

Sineos (changes requested):
Otherwise LGTM 👍 
Thanks

Sineos (approved):

config/printer-creality-cr10v3-bltouch-2020.cfg
```diff
@@ -0,0 +1,179 @@
...
+#->Select communication interface: UART0
+#->You can select baud rate value for the serial port of: 250000 bps or lower
+#make clean
+#make
+#
+#If using OctoPi or similar, please first stop the Klipper service:
+#service klipper stop
+#make flash FLASH_DEVICE=/dev/ttyUSB0
+#and then re-enable it after flashing the firmware:
+#service klipper start
```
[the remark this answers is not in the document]
CoreRasurae: Removed

KevinOConnor:
Thanks.  Some comments:
1. Use "option: value" syntax instead of "option = value" syntax ( https://www.klipper3d.org/Example_Configs.html point 7d).
2. I'm confused by the filename.  It sounds like you want to call this file `printer-creality-cr10-v3-2020.cfg`
3. It would be preferable if the initial comment header was similar to other config files (eg, https://github.com/Klipper3d/klipper/blob/master/config/printer-creality-cr10s-2017.cfg ).  It's preferable not to copy generic installation instructions in every config file as that becomes a maintenance burden should the generic installation instructions change.

-Kevin

Fisheiyy:
This doesnt seem needed, there is a lot of just random comments that dont need to be there and if the printer doesnt ship stock with a BLTouch already installed or in the box then there is no reason for it to be an option in the config, technically most if not all printers now have a port for the BLTouch but that doesnt mean the config should have that option already, if they are going to add a BLTouch then it should be their own duty to configure the config accordingly

Sineos:
@Fisheiyy: I do not agree. If the board is able to host a BLtouch it is fine to include a commented configuration. Many users typically struggle with such a configuration. Especially true on ready made printers with scarce or no information of the board.

Fisheiyy:
> @Fisheiyy: I do not agree. If the board is able to host a BLtouch it is fine to include a commented configuration. Many users typically struggle with such a configuration. Especially true on ready made printers with scarce or no information of the board.

32 Bit Creality boards have BLTouch ports yet none of the configs have a commented out BLTouch section even though Creality sells their CRTouch and sometimes bundles that include the CRTouch along with the printer from them directly

JamesH1978:
I agree, if the board has the dedicated port for a peripheral, but it does not ship with that peripheral, i think a commented out section is fine to give the end user the option. This would include most modern creality board with bltouch ports and filament runout sensor ports. As the pins are mostly the same on the STM32 boards, this would not take much to update older configs too.

Thanks
James

Fisheiyy:
> I agree, if the board has the dedicated port for a peripheral, but it does not ship with that peripheral, i think a commented out section is fine to give the end user the option. This would include most modern creality board with bltouch ports and filament runout sensor ports. As the pins are mostly the same on the STM32 boards, this would not take much to update older configs too.
> 
> Thanks James

I understand having it there for those that dont know how but if they want to get more advanced with things like a BLTouch and modding their printer then that should be their own thing to figure out, there are many guides on youtube and such that very easily explain the concept and guide you on how to set it up and it just bloats many configs and can confuse those that dont want to get advanced and dont understand why its there but not being used.

CoreRasurae:
Hi again,

Sorry for jumping in, in the discussion... and what about having two configuration files?
One printer-creality-cr10-v3-2020.cfg and another printer-creality-cr10-v3-bltouch-2020.cfg, so that the former is for the standard CR-10 v3 and the latter is for the stanard CR-10 v3 with the BLTouch option? This would not confuse anyone... the user would have only to pick the most appropriate configuration for his printer.

Thanks,
Luís

Sineos:
It is quite common to have sections commented out.
General idea of the configs is to provide as much information on existing and potential hardware as possible. 
The same logic could be applied to a e.g. BTT Octopus board: Just because most users do not use the 200 fans the board supports, it still makes sense to list them as commented options.

@CoreRasurae leave the BLtouch as commented option

CoreRasurae:
Hi Kevin,

Well, regarding the atmega2560 in CR10 v3 printer boards, they do come pre-programmed with a bootloader which is compatible with the avrdude and particularly make flash. Regarding the atmega2560 chip itself, I'm not so sure, there are some Microchip/Atmel chips which started to be distributed with a factory programmed bootloader at some point in time. I'm not sure if the atmega2560 is one of those, so if using fresh atmega2560 chips, an AVR ISP programmer maybe required. The Creality board does have a AVR ISP connector in case of need, particularly if a faulty Atmega chip needs to be replaced.

Best regards,
Luís

PR #5535, opened 2022-05-30
config: Huvud KlipperToolhead config

charlespick:
Pin information and basic flashing instructions for the Huvud CAN bus toolhead board. V0.61

Signed-off-by: Charles Pickering <me@charlespick.xyz>


charlespick:
> 1. Please use "option: value" syntax

Might want to update the docs to clarify how having an option: value within an option: works. I've always found that confusing.

KevinOConnor:
Thanks.  I committed this with df394655 on top.

-Kevin

PR #5538, opened 2022-05-31
extruder: Add QUIET=1 option to SET_PRESSURE_ADVANCE

jamespharaoh:
By adding QUIET=1 to a SET_PRESSURE_ADVANCE command, no output will be
generated. This is useful when automatically modifying values during a
print, as discussed here for example:

https://www.reddit.com/r/klippers/comments/o98xes/pressure_advance_becomes_way_too_aggressive_when/


KevinOConnor:
Thanks.  In general, however, I'm leery of adding additional command complexity for "cosmetic" changes.  There is no harm in outputting additional user information by default.

Separately, changing pressure advance results in a full kinematic flush, so it would be a bad idea to call it with high frequency mid-print.

Cheers,
-Kevin

PR #5544, opened 2022-06-03
Add USB CanBoot support for stm32 devices

Arksine:
This series adds support for requesting CanBoot over USB.  Also included is a "bypass" mode when Klipper requests the "reset" command.  This should eliminate any delays introduced by the bootloader under normal conditions.

I included the previously discussed patch to stm32 can which fixes the tx mailbox issue on GD32 devices.  If its preferable I can remove it and add it to another PR, or you could cherry pick that commit and hold wait on the rest until you have a chance to properly review the code.


KevinOConnor:
Thanks.  The patches look fine to me.  I'll commit when you are ready.

A note on the last patch ("[flash_can: support entering the usb bootloader") - entering the bootloader is specific to Klipper and it might be nice to make CanBoot more payload agnostic.  (That is, it might be nice to be able to use CanBoot with other programs - as bootloaders in general are a "mess".) One thing that occurred to me is that we could make Klipper's flash_usb.py script CanBoot aware instead.  That is, flash_usb.py could reset the Klipper usb device and then check if it comes back up in CanBoot mode - if so it can automatically call flash_can.py .  Up to you.

-Kevin

Arksine:
Oh, that is a good point.  I reverted the changes to `flash_can.py` in the CanBoot repo and updated it in this PR.  I'll look into getting the USB IDs for CanBoot so we can update `flash_usb.py` to detect it.

KevinOConnor:
Since flash_can.py is currently in the lib/ directory, we'll want to keep it in sync with its upstream source (and update lib/README if there is a change).

I can merge the first 4 commits and hold off on the 5th if you'd like.  Up to you.

-Kevin

Arksine:
Sounds good. Yeah, I reverted the usb bootloader stuff in `flash_can.py`, so they should be in sync.

KevinOConnor:
Thanks.  I committed this along with c43f62bd on top.

-Kevin

PR #5553, opened 2022-06-07
Correct UART pins for Ender 3 S1 display cable

aartrost:
The display cable on my Ender 3 S1 is connected to `PA3/PA2` instead of `PA10/PA9`. Not sure if this applies to the S1 Pro as well, I guess it does but I don't want to commit anything I can't verify.

### DIFF
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -7,7 +7,7 @@
 
 # If you prefer a direct serial connection, in "make menuconfig"
 # select "Enable extra low-level configuration options" and select
-# serial (on USB PA10/PA9), which is broken out on the 10 pin IDC
+# Serial (on USART2 PA3/PA2), which is broken out on the 10 pin IDC
 # cable used for the LCD module as follows:
 # 3: Tx, 4: Rx, 9: GND, 10: VCC
 
--- a/config/printer-creality-ender3-s1pro-2022.cfg
+++ b/config/printer-creality-ender3-s1pro-2022.cfg
@@ -7,7 +7,7 @@
 
 # If you prefer a direct serial connection, in "make menuconfig"
 # select "Enable extra low-level configuration options" and select
-# serial (on USB PA10/PA9), which is broken out on the 10 pin IDC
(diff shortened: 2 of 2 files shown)


Fisheiyy:
You would have to change the pins in the comments for the S1 Pro config also, what is the reason for using UART to connect to the S1 rather than just the USB-C port?

aartrost:
> You would have to change the pins in the comments for the S1 Pro config also, but what is the reason for using UART to connect to the S1 rather than just the USB-C port?

Main reason is because I've installed my pi internally so there's no cables sticking out. I've heard that USB also incurs a higher latency but I haven't noticed any difference personally.

But aside from that I just wanted to save somebody a few hours of debugging because the ports in that example are incorrect.

Fisheiyy:
As long as you can get more proof that the pins are correct for more people or on more hardware then I would say this PR is fine to merge.

kevinakasam:
I can confirm that `PA3/PA2` are the correct pins. (Ender 3 S1 (non Pro) with STM32F401 chip).

Fisheiyy:
> I can confirm that `PA3/PA2` are the correct pins. (Ender 3 S1 (non Pro) with STM32F401 chip).

The pins should be the exact same for the S1 Pro and with the normal F103 chip, as long as @aartrost also fixes the pins in the comments for the S1 Pro this PR is fine to merge.

aartrost:
Good stuff! I have the F103 chip so the pinouts are indeed exactly the same. I'll update the S1 pro config right now

Fisheiyy (approved):
Good to merge @KevinOConnor

KevinOConnor:
Thanks.  Commits need a signed-off-by line for me to commit directly - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

aartrost:
> Thanks. Commits need a signed-off-by line for me to commit directly - see: https://www.klipper3d.org/CONTRIBUTING.html
> 
> -Kevin

Ah, missed that. Added the signed-off-by line and combined the merge into a single commit.

KevinOConnor:
Thanks.

-Kevin

PR #5558, opened 2022-06-08
CanBoot support updates

### LINKED ISSUE #5557: Klipper cannot reconnect to the MCU after firmware restart since 8b1e3c3fb25e3d3fbec49158ba9dba9b07c9be00
Stating with the commit aforementioned above, Klipper is not able to reconnect to the (SKR v1.3) MCU following a firmware reset.
Building and flashing the MCU firmware is enough to reproduce the issue, Klippy version does not seem to make difference.
On power-on, the connection builds up without problems.

Relevant klippy.log section:

```
mcu 'mcu': Starting serial connect
webhooks client 3036409280: New connection
webhooks client 3036409280: Client info {'program': 'Moonraker', 'version': '?'}
mcu 'mcu': Timeout on connect
mcu 'mcu': Wait for identify_response
  File "/usr/lib/klipper/klippy/serialhdl.py", line 69, in _get_identify_data
    params = self.send_with_response(msg, 'identify_response')
  File "/usr/lib/klipper/klippy/serialhdl.py", line 258, in send_with_response
    return src.get_response([cmd], self.default_cmd_queue)
  File "/usr/lib/klipper/klippy/serialhdl.py", line 315, in get_response
    self.serial.raw_send_wait_ack(cmds[-1], minclock, reqclock,
  File "/usr/lib/klipper/klippy/serialhdl.py", line 250, in raw_send_wait_ack
    self._error("Serial connection closed")
  File "/usr/lib/klipper/klippy/serialhdl.py", line 62, in _error
serialhdl.error: mcu 'mcu': Serial connection closed
mcu 'mcu': Timeout on connect
mcu 'mcu': Wait for identify_response
  File "/usr/lib/klipper/klippy/serialhdl.py", line 69, in _get_identify_data
    params = self.send_with_response(msg, 'identify_response')
  File "/usr/lib/klipper/klippy/serialhdl.py", line 258, in send_with_response
    return src.get_response([cmd], self.default_cmd_queue)
  File "/usr/lib/klipper/klippy/serialhdl.py", line 315, in get_response
    self.serial.raw_send_wait_ack(cmds[-1], minclock, reqclock,
  File "/usr/lib/klipper/klippy/serialhdl.py", line 250, in raw_send_wait_ack
    self._error("Serial connection closed")
  File "/usr/lib/klipper/klippy/serialhdl.py", line 62, in _error
serialhdl.error: mcu 'mcu': Serial connection closed
mcu 'mcu': Timeout on connect
mcu 'mcu': Wait for identify_response
```

Relevant config section:

```
[mcu]
restart_method: command
serial: dev/ttyS0

```

Arksine:
This PR introduces the ability to request CanBoot for lpc176x devices.  In addition, the `no-delete-null-pointer-checks` flag has been added to fix the regression introduced by 8b1e3c3fb25e3d3fbec49158ba9dba9b07c9be00, this resolves #5557.  Finally, I introduced some changes to `flash_usb.py` to detect canboot and attempt to flash.  This is just one way of handling it, we may decide we want use another approach.

I have tested the changes on SKR 1.3, both with CanBoot and with the Smoothieware bootloader to verify that the issue is resolved.  The smoothieware bootloader is a bit flaky on reset, but this has always been an issue.  If desirable you could cherrypick the first two commits (or just the change to the makefile) or I can drop changes to `flash_usb.py` and submit separately.


KevinOConnor:
Thanks.  The lpc176x changes look fine to me.

Some comments on the flash_usb.py change:
1. I don't think we want to add a dependency on udevadm.  It is possible to read the productid/vendorid via sysfs.  See how flash_picoboot() gets usbdir to read busnum and devnum - there are similar idProduct and idVendor files in the same directory.
2. I'm leery of adding the the explicit `~/klippy-env/` path to the script.  I fear it could cause unexpected changes in behavior that will confuse users.  Maybe an alternative is to install the system pyserial package as part of the Klipper installation process.
3. I'm not sure about the global check for canboot.  If the user is going to call flash_usb.py with a `/dev/serial/by-id/usb-CanBoot-...` path then could they just call flash_can.py directly as well?

Thanks again,
-Kevin

Arksine:
> I don't think we want to add a dependency on udevadm. It is possible to read the productid/vendorid via sysfs. See how flash_picoboot() gets usbdir to read busnum and devnum - there are similar idProduct and idVendor files in the same directory.

Sounds good, I'll switch to that method.

> I'm leery of adding the explicit ~/klippy-env/ path to the script. I fear it could cause unexpected changes in behavior that will confuse users. Maybe an alternative is to install the system pyserial package as part of the Klipper installation process.

No problem, I will remove it.  Since the makefile defaults to python3 I can use `sys.executable`.  I could also make sure that pyserial installed by attempting to import it.  If not, I could print out an error message with basic instructions:
```
Unable to locate pyserial, please install with:
python3 -m pip install pyserial
```

> I'm not sure about the global check for canboot. If the user is going to call flash_usb.py with a /dev/serial/by-id/usb-CanBoot-... path then could they just call flash_can.py directly as well?

That is true.  I added that check in case the user runs `make flash` while the MCU is running CanBoot.  I can remove it, I suspect that an attempt to enter the bootloader just wouldn't do anything and `flash_can.py` would get called anyway.

Arksine:
Ok, I made some changes to `flash-usb.py`:

1) I removed the global canboot check. If the user runs `make flash` while CanBoot is running `flash_usb.py` will do the right thing and call `flash_can.py`.
1) As suggested I removed the call to udevadm and instead parse the usb ids from sysfs
2) I use `sys.executable` to call `flash_can.py` and verify that pyserial is installed by attempting to import it.  I am making an assumption here that `flash_usb.py` will be run using python 3.  Perhaps I should add a check? 
3) If `flash_can.py` results in an error I don't raise an exception, I just exit the program.  That way the user doesn't get confused by the error message specific to dfu/hid.

I tested the changes on various devices and bootloaders:
- STM32F103 (bluepill) with CanBoot, HID, and stm32duino bootloaders
- STM32F407 with CanBoot and HID Bootloader
- SKR 1.3 with Smoothie and Canboot bootloaders

KevinOConnor:
Thanks.  I committed these changes.

FWIW, it may be better for flash_can.py to warn about pyserial instead of flash_usb.py.

-Kevin

PR #5559, opened 2022-06-08
DS18B20: fix string conversion error on startup

functionpointer:
Unfortunately while applying https://github.com/Klipper3d/klipper/pull/5233 one change was missed, so klipper now fails to start.


KevinOConnor:
Thanks.  I apologize for the bad merge.  I also added commit f42ce3e2 to try to catch errors like that in the future.

-Kevin

PR #5562, opened 2022-06-10
Allow MAX31865 and ADXL345 to work on same hardware SPI bus on RP2040

hafriedlander:
This allows configuring the MAX31865 to run in Mode 3, so it can share 
a hardware SPI bus with an ADXL345 (which only supports Mode 3).

Separately, this fixes a bug doing this on the RP2040. The RP2040 SPI 
driver previously set the SPI mode on transmit (after the SPI hardware is enabled) 
but that causes problems for some hardware. For instance, trying to run a 
MAX31865 in Mode 3 would fail.

Finally this adds spi0e, which is the SPI pin config for the Seeduino XIAO RP2040
(https://wiki.seeedstudio.com/XIAO-RP2040/)

You might think this patch introduces a regression that prevents running 
different modes or bus speeds on the same hardware SPI bus, but AFAICT 
this was never possible on RP2040 anyway (happy to be corrected if
someone knows better).

Signed-off-by: Hamish Friedlander <hafriedlander@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2189,6 +2189,8 @@ sensor_pin:
 #spi_software_miso_pin:
 #   See the "common SPI settings" section for a description of the
 #   above parameters.
+#spi_mode: 1
+#   For MAX31865 only, select SPI mode 1 or 3
 #tc_type: K
 #tc_use_50Hz_filter: False
 #tc_averaging_count: 1
--- a/klippy/extras/spi_temperature.py
+++ b/klippy/extras/spi_temperature.py
@@ -277,8 +277,9 @@ class MAX31865(SensorBase):
         rtd_reference_r = config.getfloat('rtd_reference_r', 430., above=0.)
         adc_to_resist = rtd_reference_r / float(MAX31865_ADC_MAX)
         self.adc_to_resist_div_nominal = adc_to_resist / rtd_nominal_r
+        spi_mode=config.getchoice('spi_mode', {1: 1, 3: 3}, default=1)
         SensorBase.__init__(self, config, "MAX31865",
-                            self.build_spi_init(config))
+                            self.build_spi_init(config), spi_mode=spi_mode)
     def calc_temp(self, adc, fault):
         if fault & 0x80:
             self.fault("Max31865 RTD input is disconnected")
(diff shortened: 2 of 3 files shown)


hafriedlander:
In case it's interesting, I think why changing the config in spi_prepare doesn't work is that there's no guarantee that changes to CR0 take effect immediately. But Klipper currently assumes it's synchronous, and the very next thing after setting CR0 is to assert CS.

The MAX31865 samples SCLK as soon as CS is asserted to detect Mode 1 or 3, and if the mode change hasn't taken effect yet, it'll see the wrong SCLK state.

The datasheet is silent about the delay between setting CR0 and it taking effect. But it does say
> Following reset, the PrimeCell SSP logic is disabled and must be configured when in this state.
which is a little ambiguous but I read as "we don't support changing configuration after operation is enabled".

Supporting evidence: it all works fine in Software SPI mode, which definitely synchronously sets the SCLK
state correctly before asserting CS.

KevinOConnor:
It's common to place multiple SPI devices on the same bus and there is no need for them to share the same mode.  The devices are supposed to ignore the miso/mosi pins when their CS pins aren't activated - thus they don't care what mode other devices use.

In contrast, the adxl345 specs state that it has problems sharing an spi bus, and recommends deploying logic gates should the device be on a shared bus - see the adxl345 spec for the necessary hardware wiring.

This change would introduce a regression as is.

If you think there is a defect in the current rp2040 code, then I'd recommend raising that separately.

-Kevin

hafriedlander:
I have the logic gate on the ADXL already.

I'll eventually set up a logic analyzer to prove / disprove it, but I don't believe mixed mode SPI currently works (reliably) on RP2040. 

I'm not going to argue too hard though. I'm aware this isn't as clean a solution as figuring out how to make mixed mode work. Happy enough to just maintain a patch for personal use.

hafriedlander:
I'll split the patch that adds spi0e out, since that's a clean change - unless you've got plans for a bus config that handles "any of a complicated mix of pins" more flexibly? (Theoretically there's still lots of pin combos that could work but can't be selected)

dalegaard:
Hi,

If the SPI system can't have mode changed while enabled, a fix could be to, in `spi_prepare`, check if the peripheral is currently disabled, or currently has a different mode, then disable it, change the mode, and then enable it again.

Best regards,
Lasse

PR #5563, opened 2022-06-11
Rework CANbus code to use a single function for packet reading

KevinOConnor:
The existing canbus code has two mechanisms for reading messages - one for admin messages and one for data messages.  This PR changes the canbus code to use a single read mechanism.  This simplifies the low-level canbus code and should make it easier to port to new MCUs.

As part of this change, I've also extended the stm32g0 code to support canbus on PB8/PB9 pins.  I needed this to test the changes.

@bigtreetech - fyi.

-Kevin

### DIFF
--- a/src/Kconfig
+++ b/src/Kconfig
@@ -56,15 +56,10 @@ config USB_VENDOR_ID
 config USB_DEVICE_ID
     default 0x614e
 config USB_SERIAL_NUMBER_CHIPID
-    depends on HAVE_CHIPID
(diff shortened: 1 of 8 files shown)


KevinOConnor:
FYI, I have rebased this branch and added a couple of additional changes:
* The low-level code has better handling for extended and remote frames.
* The setting of the canbus uuid was moved out of the low-level canbus hardware code - further simplifying that code.

-Kevin

PR #5569, opened 2022-06-12
delta_calibrate: Fix DELTA_ANALYZE to work with Python 3

CODeRUS:


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit directly (see https://www.klipper3d.org/CONTRIBUTING.html ).

I went ahead and made this buxfix - commit 7f9ea231.

-Kevin

PR #5575, opened 2022-06-13
flash_usb: Use hexid "2e8a:0003" for rp2040 already in bootloader mode

KevinOConnor:
Use a USB hex id to indicate that the device is already in bootloader mode.  This makes the rp2040 flashing code similar to the other boards.

@dalegaard - fyi.

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20220617: It was previously possible to flash an rp2040 in bootloader
+mode by running `make flash FLASH_DEVICE=first`. The equivalent
+command is now `make flash FLASH_DEVICE=2e8a:0003`.
+
 20220612: The rp2040 micro-controller now has a workaround for the
 "rp2040-e5" USB errata. This should make initial USB connections more
 reliable. However, it may result in a change in behavior for the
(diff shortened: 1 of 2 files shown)


dalegaard:
Hi Kevin,

FWIW I picked "first" because that's what it will look like to a user with multiple connected devices, and it's significantly easier to write than a USB ID :-)

Best regards,
Lasse

KevinOConnor:
I understand.  The other bootloaders all have a similar mode (hidboot, dfu-util, smoothieware) and use hex ids.  That is, providing a hex id to flash_usb.py means: flash the first device with the given usb hex id.

I understand that hex ids may not be intuitive, but I'm not sure "first" is all that intuitive either.

One of the things I was looking at is the possibility of improving auto-detection of flash method.  This may be useful in the future with canboot and with usb-canbus-bridge mode.  Having more consistent command-line arguments will make this easier (should we go down that route).

Cheers,
-Kevin

dalegaard:
Hi Kevin,

Sorry, I wasn't being too clear. I merely wanted to provide context for the original choice. I have no opinions for or against this change.

Best regards,
Lasse

Arksine:
When I was adding CanBoot support to `flash_usb.py` I considered entering the bootloader first, then use the USB IDs to detect the flash method.  I thought it best not to make such a drastic change at the time, but I do think it should be possible to do this reliably.

PR #5578, opened 2022-06-15
config: fix processor name in th3d ezboard lite v2.0 config

meteyou:
i don't have this board, but a user sent me a picture and also on the website this processor name is listed.

pic from Luissen#7248
image

[link to website](https://www.th3dstudio.com/product/ezboard-v2/)
image


KevinOConnor:
Thanks.  Internally, the f405 and f407 should be identical to Klipper, but no harm in changing the description.

-Kevin

PR #5581, opened 2022-06-16
Remove mpu9250 duplicate code and minor changes

KevinOConnor:
This PR changes mpu9250.py to use the existing adxl345.py "helper class" code.  This removes code duplication between the two files.

I don't have an mpu9250 so can't test these changes directly (the changes work on an adxl345).

In addition to the removal of code duplication, I have added the mpu9250 to the regression test cases and updated Config_Reference.md .

The last commit in this series inlines the `twos_complement()` function into the inner loop of `_extract_samples()`.  I haven't run benchmarks, but Python can have high overhead when calling Python functions from an "inner loop".  The twos-complement expansion can be done in a single line of code so inlining it isn't too painful.

@bluesforte , @dmbutyugin - fyi.

-Kevin


docs/Config_Reference.md
```diff
@@ -1525,6 +1525,24 @@ cs_pin:
...
+Support for mpu9250 and mpu6050 accelerometers (one may define any
+number of sections with an "mpu9250" prefix).
+
+```
+[mpu9250 my_accelerometer]
+#i2c_address:
+#   Default is 104 (0x68).
+#i2c_mcu:
+#i2c_bus:
+#i2c_speed:
```
dmbutyugin, line 1539: It looks like `i2c_speed` has a default value of 400000.

dmbutyugin:
Thanks, Kevin, the changes look good from my perspective. Also, perhaps @Sineos may be interested in testing this PR, as an owner of MPU-9250.

KevinOConnor:
FYI, I've updated this branch with improved `AccelCommandHelper()` support.  It should now be possible to again have a bare `[mpu9250]` config section.

> It looks like i2c_speed has a default value of 400000.

Thanks for reviewing.  I'm not really sure how to word that.  Basically every Klipper mcu ignores i2c_speed. (I think the rp2040 may do something with it, but nothing else.)  I did add commit 55d1c372 (to the master branch) in an effort to better document that.

-Kevin

Sineos:
Successfully tested the mpu9250 on a RPi Pico.

It also seems to work to have both a `[adxl345]` and a `[mpu9250]` section defined without a weird error message.

KevinOConnor:
> Successfully tested the mpu9250 on a RPi Pico.

Thanks.  Does that mean the rp2040 is successfully obtaining a 400K i2c rate?

-Kevin

bluesforte:
Just got back from a work trip and looks like you beat me to this and did a great job doing so.  Looks good to me and works on my printer as well.  Good ideas on inlining two's complement, and adding it to the test config too.

Sineos:
> Thanks. Does that mean the rp2040 is successfully obtaining a 400K i2c rate?

grafik

Average: 4026 Hz

bluesforte:
I believe that's the sampling frequency and not i2c baud rate.  However, given that you are getting 4000+Hz sample rate I believe that does mean you are running at 400kbaud.  In order to send 4kHz samples on the bus, it was pretty close to using all 400kbaud (while developing, sending only 1 sample per block just barely oversaturated the connection.  Sending 8 at a time like the adxl saved enough in overhead to fit just fine.  So I think if the rp2040 was running at any much less than 400kbaud it would have failed).

Sineos:
> I believe that's the sampling frequency and not i2c baud rate.

Yes, of course. At least I do not know how one would measure the baud rate directly. With a baud rate of 100k, I get around 1030 samples / second.

KevinOConnor:
Thanks.  I merged this PR.

-Kevin

KevinOConnor:
> Average: 4026 Hz

Thanks - that was the info I was looking for.  (And thanks should go to @dalegaard for implementing the rp2040 i2c support.)

FYI, I added a note to the docs on which MCUs support an i2c_speed of 400K (commit 52a8afba).

-Kevin

PR #5582, opened 2022-06-16
docs: Updated accelerometer instructions

dmbutyugin:

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -75,21 +75,21 @@ the system that may damage the electronics.
 ### Software installation
 
 Note that resonance measurements and shaper auto-calibration require additional
-software dependencies not installed by default. First, you will have to run on
-your Raspberry Pi the following command:
+software dependencies not installed by default. First, run on your Raspberry Pi
+the following commands:
+```
+sudo apt update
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  It looks good to me.  I'll give a few days for others to comment as well.

-Kevin

thijstriemstra (approved):
I'm looking into using klipper and was looking for more info on the resonance measurement and these additional docs will certainly help setup the mpu-9250 (i was going to use for a mecanum wheel robot but using it on a 3d printer is much cooler).

Sineos:
I'm running Ubuntu on my Pi 3B and for whatever reason had troubles setting the I2C speed via `config.txt`. I ended up doing it directly in the device tree:
`fdtput --type u /boot/firmware/bcm2710-rpi-3-b-plus.dtb /soc/i2c@7e804000 clock-frequency 400000`

It may be worthwhile to provide an information how to check the actual I2C speed as verification that everything worked. One possible way is:

`echo $((16#$(xxd /sys/class/i2c-adapter/i2c-1/of_node/clock-frequency | awk '{print $2$3}')))`

KevinOConnor:
Thanks.  I merged this PR.  I think we can add additional instructions via another PR on top.

-Kevin

PR #5583, opened 2022-06-16
Support for “USB to canbus bridge" mode

KevinOConnor:
This PR adds support for running Klipper as a "USB to canbus bridge" - that is, one can use the same mcu to both run Klipper and act as a "USB to CAN adapter".  This may be useful for those running a CANbus toolhead and have a Klipper main board that is also CANbus capable.

There is also a discussion on this feature at https://klipper.discourse.group/t/experimental-usb-to-canbus-bridge-mode/3136 .

See the CANBUS.md document in this PR for important notes - https://github.com/Klipper3d/klipper/blob/work-usbcan-20220608/docs/CANBUS.md#usb-to-can-bus-bridge-mode

-Kevin


RimmeVG:
> 

This week i have mks skipr as usb to can bridge connection to ebb sb2209 just setup. Changed auto to allow-hotplug. Updated klipper to latest version, flashed mcu's after.
When i start printer, homing and qgl works.
But after restart firmware i get error 19. Need to restart a few times before it works. 
So i have the same problem now as other guys before your fix.

PR #5584, opened 2022-06-17
bed_mesh: cache mesh status

### LINKED ISSUE #5580: mesh_pps causing excessive load even when idle
I'm setting up a machine using a RPi Zero 2 W, and I noticed while experimenting with a large bed mesh (probe_count: 29,26), that the klipper process is using an excessive amount of cpu time even when the machine is idle.

Since the mesh takes nearly an hour to make, I altered mesh_y_pps & mesh_x_pps in the saved config section to expedite testing. After a firmware restart, and allowing everything to settle in for few minutes, cpu load on the klipper process looks like the following. 

|PPS|Load|
|------|------|
|0,0 |~9%|
|1,1 |~18%|
|2,2|~23%|
|4,4|~41%|
|8,8|~108%|

Anything higher than 4,4 and klipper will shut down while doing something as simple as a G28, and give the following message.

> MCU 'mcu' shutdown: Timer too close
> This often indicates the host computer is overloaded. Check
> for other processes consuming excessive CPU time, high swap
> usage, disk errors, overheating, unstable voltage, or
> similar system problems on the host computer.

I know at some point it will always overload the pi, but load should not be this high even when sitting idle. 

If i get some time, i'm going to try and dig into the code today, and see what might be going on, as the log doesn't point to anything obvious.

[klippy.log](https://github.com/Klipper3d/klipper/files/8920488/klippy.log)
top

Arksine:
Cache the mesh status dict, only updating it when a mesh is loaded, cleared, or if there is a profile change.  This should resolve #5580.

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -124,6 +124,8 @@ class BedMesh:
         # Register transform
         gcode_move = self.printer.load_object(config, 'gcode_move')
         gcode_move.set_move_transform(self)
+        # initialize status dict
+        self.update_status()
     def handle_connect(self):
         self.toolhead = self.printer.lookup_object('toolhead')
         self.bmc.print_generated_points(logging.info)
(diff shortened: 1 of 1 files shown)


dans98:
I've tested this, and responded to #5580

KevinOConnor:
Thanks.

-Kevin

PR #5591, opened 2022-06-20
support XRA140X on rp2040, ADS1X1X, build on mac for rp2040

kpishere:
Add support for SPI based port expansion on RP2040 MCUs.

Add ADS1X1X support I2C AD converters for all MCUs.

Add a missing header file from linux for compiling on macs.


kpishere:
The code changes have been thoroughly tested on at least one physical unit.  There is another user using the same device with the XRA140X device and others have expressed interest and/or are currently using one of ADS1X1X devices.

On the two units using ADS1X1X, the extruder temperature is measured and so is the bed temperature.  Fail-over for a stuck bed temp have triggered as expected when the bed is in the stuck on state and the unit shuts down at 120 degC.  The GPO Expander pins switch to their input state and the power switches off as expected for heater bed and extruder.

kpishere:
A reference schematic is shown here.

[PrintPiPico.0.2.7.56.17.AM.pdf](https://github.com/Klipper3d/klipper/files/8943886/PrintPiPico.0.2.7.56.17.AM.pdf)

Sineos:
Thanks for your contribution. 

Without judging code etc, there are some formal things:
* PRs need a signed off line - see: [klipper3d.org/CONTRIBUTING.html](https://www.klipper3d.org/CONTRIBUTING.html)
* There is at least one unrelated change: README.md
* To facilitate reviewing and potentially bisecting, should problems occur, it is preferable to submit the different features as individual PRs or at least as separate commits.
* Are the cfgs needed? It does not seem there is any actual hardware available currently

PR #5592, opened 2022-06-20
STM32F0/STM32G0 I2C Bus Corrections

retsamedoc:
This is a fix for a simple oversight to the STM32F0_i2c.c codebase when the STM32G0 support was introduced.

The original pins ("PF1,PF0") used to create the i2c1a bus for the F0 series microcontrollers are now dedicated to an external crystal oscillator in the G0 series.

Found this problem while attempting to migrate from an BTT SKR Mini E3 V2.0 to V3.0 where I am using a SSD1036 on the EXP1 header.

This is the same fix @gghyoo had introduced in #5493 but (hopefully) signed-off correctly and following the procedure.


KevinOConnor:
Thanks.  Unfortunately this doesn't look correct.  The PB8/PB9 pins are also valid on stm32f0 so this new definition should be added to the file - not exclusive to the g0.  Ideally the bus names would match the i2c.c file.

-Kevin

KevinOConnor:
Yes - PB8/9 should be added to all.

-Kevin

retsamedoc:
And done.

I added PB8/PB9 to the STM32F0 as i2c1b (so as to not cause current users pain) but left it defined at i2c1a for the STM32G0 (since PF0/PF1 are NOT part of the i2c1 bus in that micro and having an i2c1b w/o i2c1a felt weird).

KevinOConnor:
Alas, the changes don't look correct.  The enumeration values have to match the array indexes for all permutations.

-Kevin

retsamedoc:
Fixed. Good catch.

KevinOConnor:
Thanks.  In general, it looks fine, but I think it may be possible to implement this without as many compile time `ifdefs`.  Could you take a look at #5733 as an alternative?

-Kevin

retsamedoc:
Fair enough. Not sure why IFDEFS are bad but #5733 does look cleaner. I'll close this.

PR #5593, opened 2022-06-21
display_status: add SET_DISPLAY_TEXT extended gcode

Arksine:
This is the alternative solution to #5527 as discussed in #5528.  The `SET_DISPLAY_TEXT` extended gcode may be used  to correctly set the message of an overridden M117 command when the message begins with a special character.


tux2603:
Line 141 in the command templates documentation would be needed to be updated to reflect the change. 

It'd probably also be worth mentioning that using `rename_existing` on M117 and M118 can lead to unexpected behavior.

PR #5603, opened 2022-06-25
flash_usb: allow to flash via usb-to-can-bridge

ayufan:
This is proof of concept to allow putting device into DFU when using `USBCAN`

This allows to send bootloader command over can bridge
and make the board to enter DFU bootloader
    
To use:
    
```
make flash FLASH_DEVICE=can0:78cc933511c6:0483:df11
```
    
It will first send bootloader request via `can0`
(it does require that device does not use canboot)

Based on: https://github.com/Klipper3d/klipper/pull/5583


scripts/flash_usb.py
```diff
@@ -24,6 +24,28 @@ def enter_bootloader(device):
...
+    try:
+        cansock.bind((interface,))
+        payload = [(canbus_uuid >> ((5 - i) * 8)) & 0xFF for i in range(6)]
+        payload.insert(0, CANBUS_CMD_REQUEST_BOOTLOADER)
+        can_id = CANBUS_ID_ADMIN
+        packet = struct.pack(CAN_FMT, can_id, len(payload), bytes(payload))
+        cansock.send(packet)
+
+        # Wait a few seconds for device to re-appear
+        time.sleep(3)
```
ayufan, line 43: I did not had good idea how to detect device.

KevinOConnor:
I merged #5583 and it looks like that caused github to do something strange to this PR.  Likely needs a new PR rebased to master.

-Kevin

PR #5607, opened 2022-06-26
Create printer-bq-prusa-i3-hephestos-2014.cfg

rcarmo:
This is a working config with full LCD and stepper settings (only thing missing is the kill switch, which did not appear to work)

### DIFF
--- a/config/printer-bq-hephestos-2014.cfg
+++ b/config/printer-bq-hephestos-2014.cfg
@@ -1,10 +1,12 @@
 # This file contains pin mappings for the BQ Prusa i3 Hephestos from 2014
+# (https://www.reprap.org/wiki/Prusa_i3_Hephestos)
 # It was sold in kit form, and uses a RAMPS board with HD44780 display without
 # heated bed or any modern amenities.
-# To use this config, the firmware should be compiled for the AVR
-# atmega2560.
+
+# To use this config, the firmware should be compiled for the AVR atmega2560.
 
 # See docs/Config_Reference.md for a description of parameters.
+
 [display]
 lcd_type: hd44780
 rs_pin: PH1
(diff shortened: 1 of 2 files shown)


rcarmo:
I've since added and tested the kill pin, and defined a set of bed leveling points that still allow access to the screws on the stock bed.

rcarmo:
Hello, I've done a self-review and this appears to tick all the boxes (also, it is my current working config). What else is needed?

KevinOConnor:
Thanks.  Commits need to have a signed-off-by line containing your full real name - see https://www.klipper3d.org/CONTRIBUTING.html .

Also, I fear a name of "bq-prusa-i3" will lead to a ton of confusion as users with a "prusa i3" may look there first.  I think some file name that doesn't contain "prusa" would be needed.

-Kevin

KevinOConnor:
For an old printer like this, maybe just call it `printer-hephestos-2014.cfg` and place the full name and wiki link in the comments at the top of the file?

-Kevin

rcarmo:
Done. I left "bq" in the name sinceother configs also refer to the manufacturer, and "bq hephestos" returns a much nicer set of Google matches than the plain model name, so this might be easier to find that way.

KevinOConnor:
Unfortunately, I had to revert this change.  The original commit did not add the config to the regression test cases and I could not simply add it myself as the config is not complete (at a minimum the mcu section is missing and the extruder pid settings are missing).

The config will need to be added to the regression test cases in order for it to be added to Klipper - see point 4 at https://www.klipper3d.org/Example_Configs.html .

-Kevin

rcarmo:
Oh, ok. I’ll try to get those in later in the weekend.

PR #5608, opened 2022-06-27
Draft: flash_usb: allow to flash via usb-to-can-bridge

ayufan:
This is proof of concept to allow putting device into DFU when using `USBCAN`

This allows to send bootloader command over can bridge
and make the board to enter DFU bootloader
    
To use:
    
```
make flash FLASH_DEVICE=can0:78cc933511c6:0483:df11
```
    
It does work. It will first send bootloader request via `can0`.


KevinOConnor:
FWIW, I'm not sure about the changes to flash_usb.py .  However, I agree we should refactor the mcu code so that there is a single `bootloader_request()` mechanism for both CANbus and USB.

-Kevin

PR #5610, opened 2022-06-28
ADS1X1X i2c sensor, XRA1403 for RP2040

kpishere:
Klipper Host changes : 

* Add support for ADS1X1X I2C based AD Converter chips.  Also tie into temperature math to allow use for extruder and header objects.  Updated configuration documentation with a new section for this sensor.

RP2040 MCU changes : 

* Fix a frequency bug on i2c bus to allow ADS1X1X host extensions to work as expected
* Add a test to verify clock rate on MCUs
* Add XRA1403 GPIO expander support for RP2040 MCUs - there are two users of this board (PickSix) and interest for other custom built RP2040 based boards.  GPIO expansion allowed for more steppers and 5V tolerant GPIO pins.  NOTE: Option defaults to N for this so that BigTreeTech Pico based board passes.  With the option enabled, the first 6 GPIO pins are reserved for SPI bus and interrupt to XRA1403 (This is what causes the BTT board to fail as test uses these pins for GPIO). 
* Added missing canbus header on MacOS compiles and more .gitignore files for 'mac-isms'

All above changes have been thoroughly tested on physical hardware.  I believe others have pulled the ADS1X1X code for their own use in other projects already.  (Physical board tested with is in the Wikipage of this originating fork)


KevinOConnor:
This PR would need to follow the guidelines at https://www.klipper3d.org/CONTRIBUTING.html .

In particular it would need a signed-off-by line.  It would also need to be "rebased" (`git rebase`) into commits of one topic each.  Given the wide variety of changes, it seems separate PRs would likely be appropriate as well.

You may also wish to start a discussion on Klipper Discourse on these changes.  Some of these changes seem unlikely to be accepted into the mainline Klipper branch (eg, the changes to rp2040 gpio seem unlikely to reach our 100 users benchmark).  The gitignore changes aren't correct - files introduced by the local build should be in the user's gitignore and not in each project's gitignore.

-Kevin

kpishere:
@KevinOConnor Ok. will keep this one here for now for keeping track of others.  The first one is ... kpishere:rp2040i2c  .  Likely will sit for longest depending in comments from others.  But at least it is independent and specific in this PR for reference.

kpishere:
@KevinOConnor The other PR is this, kpishere:ads1x1x  The likely easier one to accept and more widely used in the short term as an add-on to any board that has i2c bus. The devices are sold on many project boards ready to connect (adafruit, ali, etc.).

This will be it for submitted changes, you've already made it clear that mac changes aren't welcome. :)

PR #5614, opened 2022-06-28
Support PCA9685 pins through host mcu with i2c

lanserge:
Added support for PCA9685 pins (digital and pwm).
    Most of the code adapted from replicape.py with some modifications.
    Pins named P0 ... P15 could be used as servos pins.
    
    Example of working config on raspberry pi
    (raspberry pi connected with PCA9685 board with four wires
         Pin 1(3.3v) : VCC
         Pin 3(Data) : SDA
         Pin 5(Clock): SDC
         Pin 6(GND)  : GND):
    
    [mcu host]
    serial: /tmp/klipper_host_mcu
    
    [PCA9685]
    host_mcu: host # the host mcu PCA9685 connected to
    i2c_bus: 1 # i2c bus index (default 1)
    i2c_address: 112 # i2c address (default 112)
    cycle_time: 0.02 # period time for the PWM (default 0.02 as servo used)
    
    [servo test_servo_0]
    pin: PCA9685:P0


lanserge:
It is possible to add extra pins with simple four wires connection to PCA9685 boards.
Each connected board adds extra 16 pins that could be used as GPIO, PWM or servo pins.
Unfortunately the refresh frequency of the IC is limited to max 1626 Hz, so it is not suitable to drive stepper motors (too slow),
but good for any other task like driving servo, fan or heater.

PR #5617, opened 2022-06-30
rp2040/gpio|i2c|spi|spi_xra1403: Support GPIO expander on SPI bus

kpishere:
XRA1403 GPIO Expander is an SPI device providing 16 gpios that are
5V tolerant.  Interrupts are used in MCU to trigger refresh of memory
register in MCU memory. First 6 pins of RP2040 are consumed for CS,SPI,
and INT. I2C module gets cosmetic update replacing numbers with
macros and an un-used parameter change to value zero.  src/spicmds.c
module needed update to move defines to header file for access by
gpio.  Feature is default OFF unless selected in menuconfig.

Signed-off-by: Kevin Peck <kevindpeck@gmail.com>


KevinOConnor:
Thanks.  However, my high-level feedback is that I'm not sure that supporting spi gpio expanders in the micro-controller is a good trade-off between complexity and capability - at least for the Klipper master branch.  Adding support for expanders complicates the code, impacts benchmarks even when using regular gpio pins, and there aren't that many boards with these gpio expanders.

Separately, it may be possible to implement host side support for gpio expanders.  (See replicate.py and sx1509.py as examples.)  Implementing in the host has notably less code overhead - though it does restrict how the gpio expander pins may be used.

-Kevin

kpishere:
Host side support, I'd thought about that.  But, when it comes to homing, it is polling rapidly so I'd opted for the interrupt method so that the signal comes back as soon as possible.  If the feature isn't configured, the code isn't in the MCU build, it is only present if enabled.  Performance wise, on the RP2040, it has no measurable effect.  I see 2-3% MCU utilization always.  The highest activity, when homing, is like 5%.

In configuring those pins, the native GPIO pins are used for Step/Dir signals and the others on the expander are used for switching devices and sensing endstops etc.  Polling only happens within memory and does not trigger communication on SPI bus (that would be relatively slow).  The GPIO 0 pin is used for interrupt where, when there is a change on the expander pin, a read follows to get the new value for an in memory copy of the expander gpio values and the interrupt is cleared. 

I did experience a vulnerability where one of the expander gpio pins was an input and the level was in-determinant and causing many interrupts.  This caused SPI bus contention, the display failed and the temperature controls could be come erratic -- system did shut down safely though in such cases.

That isn't great but I'd bet something similar could happen with MCU in such a case too, having a floating input triggering many interrupts.

PR #5618, opened 2022-06-30
extras/ads1x1x.py: Add support for i2c based ADC chip

kpishere:
ADC1X1X is a series of i2c bus chips with high sample rate and multiple channels.
Each channel can be mapped to a heater device and use the existing sensor math
and/or lookup tables.  Documentation is added.  A sample configuration is also
included (printer-picksix-v02.cfg).

Signed-off-by: Kevin Peck <kevindpeck@gmail.com>

### DIFF
--- a/config/printer-picksix-v02.cfg
+++ b/config/printer-picksix-v02.cfg
@@ -1,0 +1,401 @@
+#
+# TMC Debug command:
+#
+# DUMP_TMC STEPPER=stepper_x
+# SET_HEATER_TEMPERATURE HEATER=heater_bed TARGET=40
+# TURN_OFF_HEATERS
+# FORCE_MOVE STEPPER=stepper_x DISTANCE=10 VELOCITY=60
+# FORCE_MOVE STEPPER=stepper_y DISTANCE=10 VELOCITY=60
+# QUERY_ENDSTOPS
+#
+# BED_MESH_CALIBRATE PROFILE=default METHOD=automatic
+#
+# GPIO Extender gpio30 - gpio45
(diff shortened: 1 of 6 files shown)


kpishere:
Yes, you'd listed all changes, not sure what else with ".etc". 

pins.py: That configuration check had to be removed because this device has up to four sensors at one address and it wasn't allowing the use of the same pins in a heater and in an extruder device.

adc_temperature.py: The PT100 related calculation changes, I think in the end I didn't use those, they can still be used but I can remove as they are "un-tested", well, they weren't the right calculation method for the sensor I'd used.

The config file is a working configuration example.  Up to you if you'd want it.  I'd figure some like to see it as example of working configuration to complement the configuration instructions.

Besides the above, all components in this commit allow you to wire the ADC and use any of its up to four sensors for a controlled heater device like extruder, bed heater, or enclosure heater as you would for any direct ADC GPIO pin on an MCU.

The ADC does MUX switching and then sampling in separate steps, so sampling ended up being a little more complicated as the switch and sample operations are open ended and you don't want to sample a value to the wrong channel.  Measuring an extruder temperature on a bed heater would shut down the MCU immediately.  So a singleton round-robin state machine was needed to control MUX 1, Sample, MUX 2, Sample, MUX 3, Sample, MUX 4, Sample, MUX1, .... kind of pattern in regular intervals.

Other driver implementations may want to take advantage of polling or interrupt use in these devices at higher sample rates but this driver is not intended to satisfy that, it is for low samples per second even though they handle much higher sample rates.

PR #5622, opened 2022-07-03
Update generic-bigtreetech-skr-3.cfg

kingtricky:
add clock reference


KevinOConnor:
Thanks.  I made the associated update (commit 50b4d4c4).

-Kevin

Issue #2977, opened 2020-06-11
Add support for HX711 based ABL sensor (used in CR-6 printers)


jschuh:
The HX711 load cell ADC is used as a bed level sensor on the upcoming [Creality CR-6](https://www.creality.com/cr-6-se-3d-printer-back-us-to-save-100-p00296p1.html) series of printers. It works by treating the extruder mounting bracket as a load cell, allowing it to detect the load generated by the nozzle as it touches the bed surface.

I found [an old, bitrotted PR](https://github.com/MarlinFirmware/Marlin/pull/2309) to implement this in Marlin, which shows that the code itself is actually quite straightforward. The patch also references [a blog post (in Russian)](https://habr.com/en/post/378637/) that provides a full demonstration using a stock 5kg load sensor (rather than a custom bracket), meaning there's the potential for broader use beyond just the CR-6 printers.

I picked up 1kg and 5kg load cells to experiment with this, so if there's interest I could take a crack at adding the support sometime in the next month or so.

klipper-gitissuebot:
Hi @jschuh,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

Unfortunately, too many people have opened tickets without providing the log. That consumes developer time; time that would be better spent enhancing the software. If this ticket references an event that has occurred while running the software then the Klipper log must be attached to this ticket. Otherwise, this ticket will be automatically closed in a few days.

For information on obtaining the Klipper log file see: https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md

The log can still be attached to this ticket - just add a comment and attach the log to that comment.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

FHeilmann:
The immediate issue I'm seing with this is that the HX711 does not have an "interrupt pin" which allows to have a digital pin indicating a threshhold has been crossed. This is problematic because the HX711's current load has to be read by shifting in a 24-bit value, which is quite expensive compared to the simple "read digital pin" which klipper currently does.

These changes would have to be implemented in the low level MCU code as well, which is a bit trickier than writing the python host code.

If you're willing to tinker a bit, I have previously written this sketch for an attiny85 (a small arduino mcu), which connects to a HX711 and provides a digital output, similar to an endstop.

[...]

PR #5623, opened 2022-07-03
add support for load-cell based probes

mhier:
Follow up of #4738 which somehow could not be re-opened. See also the original ticket #3496.


mhier:
To help the review process, I am trying to collect the information as mentioned in the contribution guide. Since I am the submitter, this will be somewhat personally biased of course, but I am trying to do my best anyway :-)

I have tested the code extensively on my printer (a Renkforce RF1000), and there are a few users with the same printer model using a slightly older version of the code (the one of the original submission) for almost a year now. The regression tests are passing (I did not run them manually but Github shows it's fine).

The relevance goes beyond my printer model, several people have already shown interest in the Klipper forums about this technology. Still the user base of this module will probably small in the beginning, since there are only a few printer models equipped with this technology off-the-shelf. One upcoming prominent example is the Prusa XL. It is also not too complicated to retrofit an existing printer with the technology. Hence I think the quoted minimum of 100 real world users will be achieved rather soon.

It should be noted though that other modules (submitted in separate PRs in the future) are likely required to use this module: An ADC driver is required. I can provide quickly a driver for the ADS1100 (just need to write documentation) also and have an interest to get a working driver for the HX711.

The code is 100% written by me and I have added the GNU GPLv3 licence to the file header. The commit is also properly signed-off by me.

The PR adds a new Python module in the extras folder and adds some code to the mathutils.py file (which is hopefully fine?). The module registers itself as a "probe" printer object, so it is mutually excluding with any other probe modules. It also provides the `get_status()` function to allow accessing some information from macros or the API. It uses an ADC pin which is assumed to follow the standard Klipper ADC interface. It should also follow the other recommendations (Section "Adding a host module" of the "Code overview" page), although this is probably someone else should really have a closer look.

Extensive documentation about the working principles and a setup guide for new printer models has been added as well as a reference documentation for the configuration parameters and the new G-Code commands. **I now see that I need to add documentation for the status variables in [Status_Reference.md](https://www.klipper3d.org/Status_Reference.html). Also the primary documentation page should be added to the Overview page and the index.** I will do this asap.

This PR consists of a single commit only. Other, related changes (ADC driver, extended functionality) will come in separate PRs to simplify the review process, as discussed in the original PR. **I now realised that the commit message does not follow the preferred format.** I will fix this, too.


So, reviewing my one submission has already done some good :-D

pawlowskialex:
Is it possible to use this on printers like V-Core 3 or Voron Trident, with 3 load cells mounted under every bed arm?

pawlowskialex:
One thing I've noticed in Bambu X1 reviews is how quickly does its z probe works. The printer has a lidar in addition to the load cell. is it possible to speed up load cell homing by using a second probe (induction, laser, etc) to augment the readings?

mhier:
> Is it possible to use this on printers like V-Core 3 or Voron Trident, with 3 load cells mounted under every bed arm?

I would assume yes, as long as all load cells are connected to the same ADC so the readout gives the average force applied to the bed. The only condition is that the force between bed and nozzle can be measured at all probing locations and has measurement values in a somewhat comparable order of magnitude across all locations. Specifically it is not important that the force-to-distance relation ("stiffness"/"spring constant") is exactly the same at all probing locations, since it will be determined in each probe run.

It looks to me that those printers also can adjust the actual bed tilt by moving the 3 Z sub-axes separately. The load cell probe algorithm will only allow you to determine how much the bed is tilted. You will need some separate code to compute and execute the correct sub-axes movements out of this. I am not sure whether this exists already, but surely it can be written.


> One thing I've noticed in Bambu X1 reviews is how quickly does its z probe work

The load-cell based probes are not quick at the moment. There are some ideas around how to improve speed, but this is some significant amount of development/research time away from production use. See e.g. this forum thread: https://klipper.discourse.group/t/strain-gauge-load-cell-based-endstops/2134

IMO in this PR we should not compare with completely different technologies. There are printers out there which already use load cells for probing, including those you quoted and my Renkforce RF1000. The speed is comparable to the original firmware of the Renkforce RF1000. In my opinion, speed is not so important, because this calibration is done only infrequently (the bed does not change its tilt and bending so often). The precise distance can be corrected with a single point measurement, which is sufficiently fast to be done even at the beginning of every print.

Fisheiyy:
@mhier Could you copy the details of this PR and put them into a new PR on [this fork of klipper](https://github.com/DangerKlippers/danger-klipper), a small group of guys is making a [community maintained bleeding edge fork of klipper](https://github.com/DangerKlippers/danger-klipper), you can join the discord about it [here](https://discord.gg/zxufeXvpzD), but please if you could copy this over, we would like you to have credit for the PR and help us implement this.

mhier:
I would rather hope that this gets integrated soon into the main branch of Klipper. @KevinOConnor can you give me a rough estimate when I can expect the next step here? :-)

KevinOConnor:
It's on my list of things to look at.  I don't have a timeframe though.

-Kevin

docs/Config_Reference.md
```diff
@@ -1835,6 +1835,42 @@ z_offset:
...
+[Load-cell probe guide](LoadCellProbe.md) and
+[command reference](G-Codes.md#load_cell_probe) for further
+information. Some of the parameters are determined in a semi-automatic
+[calibration procedure](LoadCellProbe.md#setup-for-new-printer-models).
+
+```
+[load_cell_probe]
+#
+# General parameters:
+#
```
KevinOConnor, line 1851: It would be preferable to layout the config reference in the same format as the other config sections.  Each parameter should state what its default is, or explicitly state that it must be provided.  FWIW, I'm not sure it helps for the reference to state which values the user can be guided to configuring.

docs/Config_Reference.md
```diff
@@ -1835,6 +1835,42 @@ z_offset:
...
+adc:
+#   ADC pin which digitizes the load cell signal.
+adc_rate:
+#   Rate (in samples per second) at which to request the ADC samples.
+max_abs_force:
+#   Maximum absolute value of measured force (in chosen unit), if exceeded
```
KevinOConnor, line 1857: FWIW, I find it confusing for the reference to be vague on units.  I think it would be more clear to just say `unit of grams`.

docs/Config_Reference.md
```diff
@@ -1835,6 +1835,42 @@ z_offset:
...
+#
+# Parameters determined by semi-automatic procedure described in the
+# load-cell probe guide.
+#
+force_calibration:
+#   Conversion factor to get from ADC value to physical unit. Default value is
+#   1 (no conversion).
+stiffness:
+#   Stiffness/"spring constant", i.e. force per distance. Default value is a
+#   safe value resulting in very small step sizes.
```
KevinOConnor, line 1868: FWIW, I find it confusing to state there is some vague default.  Probably best to state explicitly what the default is.  I'd refrain from saying it is "safe" as we really have no idea if it could cause damage to hardware or not.

docs/Config_Reference.md
```diff
@@ -3405,6 +3441,10 @@ lcd_type:
 #   See the display sections below for information on each type and
 #   additional parameters they provide. This parameter must be
 #   provided.
+#line_length:
+#   Set the number of characters per line for an hd44780 type lcd.
+#   Possible values are 20 (default) and 16. The number of lines is
+#   fixed to 4.
```
KevinOConnor, line 3447: This looks like a merge error.

docs/G-Codes.md
```diff
@@ -915,6 +915,56 @@ babystepping), and subtract if from the probe's z_offset.  This acts
...
+
+#### LCP_CALIB_STIFFNESS
+`LCP_CALIB_STIFFNESS [SAMPLES=<n>]`: Only used for the initial calibration.
+Determine the `stiffness` configuration parameter with the procedure defined in
+the [calibration procedure](LoadCellProbe.md#setup-for-new-printer-models). The
+optional number of samples is used for averaging (defaults to 10).
+
+#### LCP_INFO
+`LCP_INFO`: Print parameters used by the load cell probe algorithms to the
+console. Useful for debugging and testing.
```
KevinOConnor, line 965: I don't think we should add so many commands.  If necessary, look to add one or two commands and allow the commands to take on separate actions depending on the parameters.  Otherwise, when looking at the project as a whole, we end up with hundreds of modules all defining dozens of commands, and it just become unmanageable.

Please spell out the module name in the command prefix (eg, `LOAD_CELL_PROBE_CALIBRATE`) as `LCP` is ambiguous.

docs/LoadCellProbe.md
```diff
@@ -0,0 +1,288 @@
...
+[reference of the g-code commands](G-Codes.md#load_cell_probe).
+
+## Mechanical and electrical setup
+
+<p align="center">
+  <img src="img/load-cell-probe-setup.png" width="30%"/>
+</p>
+
+There may be many possibilities how to realise a probe based on load cells. The
+setup used for testing and developing this algorithm is as follows:
```
KevinOConnor, line 20: FWIW, I think the document would be easier for users if it started with a concrete set of steps necessary to calibrate and use their probe.  The theory sections are helpful, but I'd put them towards the end of the document.  In my experience, the majority of users just want to follow a guide with the steps to follow and are uninterested in obscure alternatives or theory

docs/Status_Reference.md
```diff
@@ -319,6 +319,20 @@ is defined):
...
   to the macro containing this reference.
 
+## load_cell_probe
+
+The following information is available in the
+[load_cell_probe](Config_Reference.md#load_cell_probe):
+- `last_force`: Returns the last force measured by the probes force
+  sensor. The force readout takes place contineously even if no probe
+  command is running, so this value can e.g. be placed on the display
+  to monitor the extrusion force (if applied to the sensor).
```
KevinOConnor, line 329: FWIW, I find the vague reference to extruder pressure to be confusing in a reference.  I think it would be preferable to succinctly state what the parameters report.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,611 @@
...
+import mathutil
+import sys
+import logging
+import random
+import math
+
+syspath = sys.path
+sys.path = syspath[1:]
+import statistics
+sys.path = syspath
```
KevinOConnor, line 17: It's not clear to me what this does.  It does not look like something we want to have in the code.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,611 @@
...
+
+    def _adc_callback(self, time, value):
+        # convert to physical unit
+        self._last_uncompensated_force = value * self._force_calibration
+        # First value after start: use as zero offset
+        if self._force_offset == None :
+          self._force_offset = self._last_uncompensated_force
+        self._last_time = time
+        # Store zero offset compensated value for display
+        self._last_force = self._last_uncompensated_force - self._force_offset
```
KevinOConnor, line 257: I'm not sure, but this doesn't look to be thread safe.  The _adc_callback() code is invoked from a background thread.  Thus, if the main thread inspected any of these variables it wouldn't be able to atomically access the set of variables (at any time, due to a race, a variable like `_last_time` might be to a different sample than a variable like `_last_force`).

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,611 @@
...
+        last_time = self._last_time
+        clocksync = self._mcu_adc.get_mcu()._clocksync
+        clock = clocksync.print_time_to_clock(last_time)
+        last_sys_time = clocksync.estimate_clock_systime(clock)
+        for n in range(1,10) :
+          # wait shortly after the timer has called _sample_timer
+          self._reactor.pause(last_sys_time + n*self._report_time + 0.0001)
+          if self._last_time != last_time:
+            return
+        # callback not called after 10 report time intervals -> error
```
KevinOConnor, line 270: I'm not sure what this code is attempting to do, but it does not look correct.

At a minimum, the code must not "peek into" the member variables of other classes (`get_mcu()._clocksync` is not valid).  This is described briefly in docs/Code_Overview.md .

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,611 @@
...
+        raise gcmd.error("Timeout waiting for ADC value.")
+
+
+    def _move_z_relative(self, length, wait=True):
+        pos = self.tool.get_position()
+        self.tool.manual_move([pos[0],pos[1],pos[2]+length], SPEED)
+        if wait:
+          self.tool.wait_moves()
```
KevinOConnor, line 278: Sometimes the code uses 4 spaces for indent, and sometimes 2 spaces for indent.  Klipper doesn't have a strict coding style, but the code should be internally consistent.  Please use a consistent indent (preferable 4 spaces per indent).

klippy/mathutil.py
```diff
@@ -135,3 +135,78 @@ def matrix_sub(m1, m2):
...
+
+
+######################################################################
+# Simple statistics
+######################################################################
+
+# Standard deviation (square-root of variance) of the given samples
+def std(samples) :
+    mean = 0.
+    for v in samples :
```
KevinOConnor, line 147: I think it would be preferable to put all this code in load_cell_probe.py for now (and not alter mathutil.py for now).  We can refactor the code at a later date if needed.  Linear regressions are used in several areas of the code today, and it's not clear to me that this implementation is a good match for all the use cases.  Many times we want to implement a "rolling average" linear regression, or at least calculate the linear regression as samples arrive (via accumulating sum and accumulating sum of squares).  For this PR, simpler to just keep all the code in load_cell_probe.py .

KevinOConnor (commented):
Some minor things I noticed.

-Kevin

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,611 @@
+# Support for bed level probes featuring a load cell at which the hotend is
+# suspended.
```
KevinOConnor, line 2: For the top "copyright block" please use just a single one-line comment at the top of the file.  Feel free to add as many comments as you need in a comment block after the top "copyright block".

Thanks,
-Kevin

mhier:
Before going into details of the comments I would like to address and resolve this general point first:

> Thanks. The main question I have is - are there a notable number of real-world users that can accomplish real-world probing after merging this code? It's not clear to me that there are real-world users with force sensors available using the current ADC code. If there isn't then we need to merge in the ADC code first (or at least have it available for review), as I don't want to merge in code that isn't actually useful.

This is a hen-egg problem. The ADC code is not useful without this module (or any other module which makes use of such kind of ADCs), while this module is not useful without the ADC code. The review process of this PR is taking already quite long. I would like to get this off the plate before cleaning up the ADC code and making it ready for review. Is it enough for you if I give you the definitive promise of delivering the "driver" module for the ADS1100 ADC quite fast after this PR is merged, and in addition a driver for the HX711 basically next thing after that? I have to admit, only HX711 driver will make this module usable "for the masses", but writing the HX711 driver is a bit more tedious, as it involves writing MCU code. I will definitively do that, because I have an interest myself to use it. That said, I should add that in theory this module should be useable with any existing ADC implementation, if the electrical setup does some tricks (convert the bipolar signal into an unipolar signal).

Can we agree on such procedure in the interest of time and efficiency?

In any case, you can have a look at the ADS1100 code anytime:
https://github.com/RF1000community/klipper/blob/master/klippy/extras/ads1100.py

I guess I will have to ask some questions about the ADC interface anyway...


> Otherwise, I have some comments, though I fear they may seem "nitpicky".

On the contrary, your points are all very valid. In most cases I wonder why I did not see that myself (and I am sorry I didn't)... I will address them soon.

KevinOConnor:
> This is a hen-egg problem. The ADC code is not useful without this module (or any other module which makes use of such kind of ADCs), while this module is not useful without the ADC code.

I think it would be preferable to review the ADC code first, as it is simpler and is also (at least nominally) useful in and of itself.  Having just one of the ads1100 or hx711 is fine, as that will give me a better understanding of how other sensors will work.

I have some high-level questions on how the ADC code will be implemented, and it seems you do as well.  I think we should finalize those details before committing code that depends on it.

> One of the problems of this long-lasting review process is though that the quality requirements seem to increase between the time I wrote the code and the review.

For what it is worth, I'm definitely trying not to "move the bar".  When I do a review, I'm looking for mostly superficial things - for example: do I understand what the code is doing, are the docs being updated, would I know roughly how to configure each config option if I had the hardware, are there obvious code errors, does the module overlap with existing modules, and similar.  I've done this for all PRs since I started Klipper.  If I find a few things in a review then I tend to point them out before completing the review.  I fear that can make it seem like I'm "moving the bar", as further submissions then lead to further comments.  However, due to the time it takes to do a review, I'm not sure there is a practical alternative.  I wish I knew of a better way to do reviews.  I do think reviewing PRs a good idea.

I do appreciate the time you've taken to work on this and all the open feedback you have provided.

-Kevin

mhier:
ah no, not again... I was just on vacation. Can you reopen this time?

mhier:
Thanks. I also converted this into a draft for the time being, since the comments by @KevinOConnor above will require the preparation of another PR first...

CustomMek:
> Thanks. The main question I have is - are there a notable number of real-world users that can accomplish real-world probing after merging this code? It's not clear to me that there are real-world users with force sensors available using the current ADC code. If there isn't then we need to merge in the ADC code first (or at least have it available for review), as I don't want to merge in code that isn't actually useful.
> 
> Otherwise, I have some comments, though I fear they may seem "nitpicky". Alas, I can only review superficial stuff as I'm not an expert on the load-cell code.
> 
> I'll attach my comments separately.
> 
> Cheers, -Kevin

Hi Kevin, i'm new to this community and being a mechanical design engineer i was until now just reading and trying to learn without being able to add anything usefull but in this case i think that giving my opinion will help in some way to respond at your "doubt" about the effective usefullness of this improvement.
As i'm not a programmer i will approach the discussion, let me say, from another prospective presenting a real use case to underlying why I support the idea that introducing a loadcell z for probing will be not only an usefull implementation but a real revolution for some circumstances.

I worked with friends on the mechanical layout of a multi extruder hot chamber 3d printer (150-200°C) and between dozeon of complex problems the probing was problably one of the most difficult to solve, for many reason the best way to probe the bed is do that with all (bed/chamber/hot-end) at the final working temperature of process, this means that with high temperature everywhere you can not use almost anything apart the nozzle imself.

You can not use a simple "clicky" switch and neither inductive or capacitive sensor because the temperature is too high, and you can not use an ultrasonic sensor placed in a remote position protected by a cooled heat shield because the chamber temperature is always different and this change also the speed of sound and consequently require a further calabration dependent to the setted temperaature, maybe you can use laser sensor but with expensive solution you will have agian other problems if you use glass surface, etc. 

Now, ammitting that you can find a way to place and use one of those sensor (thinking in terms of an "object" without coonsidering the technology ) and ammitting that you can make it work properly you still have another problem infact you need to compensate also the hot-end thermal expansion that is not included "by design" in this kind of probing. In this kind of machine the heat-end is often very long due to the presence of heat shield etc and reach very high temp so also this point is not negligibly so the end you have to consider also the offset of the nozzle tip respet to the sensor itself and you end with a sum of error that is major than the value you're trying to measure.

The conclusion,IMO, is that the best solution that will became the new gold standard for any type of machine it to probe the bed direclty with the nozzle and following this idea the loadcell is the waay to go, it cana be placed between the extruder and the hot-end,  out of the hot-chamber in a protected area and solve whole the problems.

Now i agree with the fact that the High Temp application will be not interesting for most but considering the patent expiration over the hot chamber i'm quite sure that will become more and more common to see in the future, also the new Prusa XL have this solution even if they havn't an hot chamber.

To be more clear about my idea i think that the loadcell will be the best solution if there's a way to fix the complex problems discussed above but i want to add other 2 other possible solutions that follow the principle of use the nozzle as a probe, just for brainstorming;
In both you need an extra linear rail that allow the tool to move a little vertically respect to the X carriage keeping it "rigid" in X-Y directions, in this case you need a mechanical to stop the moviment on the bottom and a preloaded spring on the top that keep the tool in contact with the stop mentioned; with this layout the tool can move upward a bit when the bed contact happen and now, this avoid mechanical issue and alloow to perform 2 kind of probing:

1) you can use a well calabrated clicky sensor fixed between the tool (which can move in Z direction as said) and the X carriage which is fixed on the gantry, doing this you can probe using the nozzle but from a remote position, this allow to avoid thermal problem, include the hot end expansion "by design" and is aabsolutely"transparent" for the software because at the end you have just remoted a simple switch sensor in another position, Now unfortunately this layout has been patented in 2021.

2) you can use a linear encoder or potentiometer instead a switch, you can record a "zero" (like in a digital caliper) before the probing start and than record the Z position when the contact happen, this is more precise, mechnically secure, and  covered by a patent.

Thank you for attention, i hope that some ideas will help in some way.

mhier:
@CustomMek Thank you for your insight and support.  This PR is not abandoned, and I am confident that we will be able to include this in the Klipper mainline eventually, but there are some technical issues to be addressed first. Kevin is right, merging this code would not immediately enable anyone to use it. We will need some ADC driver implementation first which can be used for this purpose. I had hoped to do that second, but Kevin prefers to have it first. It will take me a while either way, because I am simply doing too many things in parallel ;-)

I have a few comments though:

> Now i agree with the fact that the high temp. application will be not interesting for most 
I don't see this limited to high temperature applications in any way. Most benefits you are quoting affect normal printers as well. I do not have a high temperature printer. Still, the original hotend of my printer extends a lot when heating it up. Being able to directly measure this (over time!) was already helping a lot in understanding many printing issues. Nowadays I totally like that the sensor is always perfectly calibrated.

> considering the patent expiration
Most people are using their printer exclusively at home for non-commercial purposes. To my knowledge, patents don't play any role in that context. It is perfectly legal to build something at home protected by patents, as long as you don't sell it or produce anything with it that you sell. You can also discuss these things online, share blue prints and software for it, as long as everything stays non-commercial. The very idea of patents is to protect ideas from being commercialised by others without keeping the idea a secret. Hence this was in most cases never any real limitation. Same is true for the two probing options you are quoting.

> you can use a linear encoder or potentiometer instead of a switch, you can record a "zero" (like in a digital caliper) before the probing start and than record the Z position when the contact happen, this is more precise, mechnically secure, and covered by a patent.
Interesting! This is very close to what the load cell approach is doing. The linear encoder or potentiometer combined with a spring which pushes the probe down is essentially a load cell. Your measurement values will be in a very different range though and you might get away with simpler ADCs etc., so maybe with such setup you could even use the on-chip ADCs of your microcontroller (and hence probably start using this code right away). The software won't care for these differences.

CustomMek:
> @CustomMek Thank you for your insight and support. This PR is not abandoned, and I am confident that we will be able to include this in the Klipper mainline eventually, but there are some technical issues to be addressed first. Kevin is right, merging this code would not immediately enable anyone to use it. We will need some ADC driver implementation first which can be used for this purpose. I had hoped to do that second, but Kevin prefers to have it first. It will take me a while either way, because I am simply doing too many things in parallel ;-)
> 
> I have a few comments though:
> 
> > Now i agree with the fact that the high temp. application will be not interesting for most
> 
> I don't see this limited to high temperature applications in any way. Most benefits you are quoting affect normal printers as well. I do not have a high temperature printer. Still, the original hotend of my printer extends a lot when heating it up. Being able to directly measure this (over time!) was already helping a lot in understanding many printing issues. Nowadays I totally like that the sensor is always perfectly calibrated.
> 
> > considering the patent expiration
> 
> Most people are using their printer exclusively at home for non-commercial purposes. To my knowledge, patents don't play any role in that context. It is perfectly legal to build something at home protected by patents, as long as you don't sell it or produce anything with it that you sell. You can also discuss these things online, share blue prints and software for it, as long as everything stays non-commercial. The very idea of patents is to protect ideas from being commercialised by others without keeping the idea a secret. Hence this was in most cases never any real limitation. Same is true for the two probing options you are quoting.
> 
> > you can use a linear encoder or potentiometer instead of a switch, you can record a "zero" (like in a digital caliper) before the probing start and than record the Z position when the contact happen, this is more precise, mechnically secure, and covered by a patent.
> 
> Interesting! This is very close to what the load cell approach is doing. The linear encoder or potentiometer combined with a spring which pushes the probe down is essentially a load cell. Your measurement values will be in a very different range though and you might get away with simpler ADCs etc., so maybe with such setup you could even use the on-chip ADCs of your microcontroller (and hence probably start using this code right away). The software won't care for these differences.

Thank you @mhier , yes you're right the application is not limited to high temp printer,  i was only trying to emphasize that while with standard printer the nozzle probing is helpfull with hot chamber machine is problably the only way to probe correctly, this difference will be not obvious..

I agree with what you say about patents, for DIY printer may be not a problem but i'm not so sure, i'm not against patents and i think are important in the business of a company but i think that some important solutions (as this is in my opinion) must be keept available for everyone, private user and other company so, the only way to achieve this it to implement it in an opensource project to avoid any futher claims. Just because we're one topic: one hour ago has been announced that new Goliath hot-end for the VZbot that everyone is waiting has been cancelled from mass production because someone claims a patent infringement.

Anyway, i writed wrong about the solution with potentiometer/linear encoder, this is NOT covered by patent (or however i didn't find it).. the patented solution is the other one with the remote positioned switch.

If you like my proposal with potentiometer/linear encoder this was the solution i adopted in my design which is still in progress, just to underline other pros/cons respect to load cell:

- More freedom on placement, no need to place it between extruder and hot-end --> load cell has to be like a ring (inner hole) and this means less chance to find the right one.

- For what i found most loadcells work only in compression or in traction and not in both way so it's improbable that we can use it to probe the bed and also to obtain the extrusion force so, if we use the potentiometer we keep the possibility for implement later a the specific load cell dedicated to extrusion force.

- You have no noise, no need to set thresold, etc..in theory you have just to record an initial value i think (but i don't know how to built it in firmware)

- You need a extra degree of moviment with an additional linear carriage plus a spring as mentioned and this will be a a big cons, but over the fact it allow to be more safe in terms of mechanics it aallow to create an high quality IDEX system, if you add a small linear stage that can move each extruder in Z indipendently you can perform the bed probing with the extruder 1 first and than with the extruder 2, after this instead to changhe the Z position of the bed you can change the Z position of each extruder, in this way you have both the nozzle at the correct distance from the bed but also equal each other, doing this both can print simultaneously...and againt, with this you can do Z hop lifting the light extruder instead to move down all the bed (faster).

clowrey:
@mhier Really looking forward to your load cell implementation in Klipper! I will be adding one to my printers via the hx711 once it's all implemented - load cells do seem like the best fairly simple way to do bed probing with the nozzle itself. 

@CustomMek I also came to similar conclusion about the awesomeness of additional Z axis on each hotend for IDEX in particular. And including some type of sensor there makes sense. I know this isn't the correct place to discuss that... my username is Chester#6621 on discord if you want to discuss more.

CustomMek:
> @mhier Really looking forward to your load cell implementation in Klipper! I will be adding one to my printers via the hx711 once it's all implemented - load cells do seem like the best fairly simple way to do bed probing with the nozzle itself.
> 
> @CustomMek I also came to similar conclusion about the awesomeness of additional Z axis on each hotend for IDEX in particular. And including some type of sensor there makes sense. I know this isn't the correct place to discuss that... my username is Chester#6621 on discord if you want to discuss more.

@clowrey i'm happy to know that i'm not the only one who has made this consideration so yes, let talk about that. I sent you a request on discord.

webbbn:
I have been following this and look forward to having this included in Klipper. I am currently using a loadcell setup that is working quite well, but it uses a dedicated Xiao MC to read the hx711 using custom firmware.

I've tried several configurations, including a loadcell on the extruder and also a "clicky" on the extruder using a small linear rail. In my case, the loadcell on the extruder didn't get very good results because I am using water cooling on the hotend, and the tubes tend to jerk around enough to cause too many false positives.

The "clicky" design worked well, but didn't fully realize the goal that I was trying to achieve. Even though it's relatively small, there is still a calibration step required to determine the amount of motion required to activate the switch. With a good, rigid design, I think it would be stable enough to only require calibration once, but, with 3d printed, plastic parts, I was concerned that I would still have to calibrate it occasionally. It also adds quite a bit of complexity and some extra weight to the extruder component.

The design that I currently am running (on a Voron 2.4) is similar to what is discussed above, and replaces the bed mounts with loadcells. There are two loadcells in the back corners and one in the front middle. They're all wired to one hx711, which is connected to a Xiao. The Xiao has an input GPIO that is set high to activate and tare the sensor, and an output GPIO that acts as a probe switch. It gets set high when the load is sufficiently higher than the value that was present when it was activated/tared. I calibrate the threshold by typing in random values until it works sufficiently well, which is likely not the best approach. :-)

Fortunately, and somewhat surprisingly, I have not notices any significant difference in performance based on probe location. I was a bit concerned about the front corners being an issue, but they don't seem to be.

I probe slowly so that the it's not jamming the head into the bed, and the repeatability is pretty much as good as a bltouch. I have my probe accuracy thresholds set to around 0.04mm, and my QBL generally terminates with a value under 0.02. Sometimes I get an anomalous value, which could very well be due to my code, but Klipper is well designed to handle that.

I would be happy to test anything that I can with my setup. I would like to wire the hx711 directly to my controller board if/when a driver is available. Is there anything else I can do to test this?

mhier:
@webbbn Sorry for the late reply. I think the first thing to do is to write a proper hx711 support. This is not yet available. I will do this next year somewhat soonish (I hope to start with it in February). I will let you know.

ChipCE:
To me, this PR is very useful. Not all printer designer can develop the module them self. With this module there will be more loadcell based design in the future. Not just load cell but smd resistor or custom pcb with trace similar to strain gauge can also be used. 
This for example https://github.com/IvDm/Z-probe-on-smd-resistors-2512

DaemonWork:
Hi.
First of all I'd like to say thank you - to everyone contributing to Klipper.  Thank you!
I've been looking at the idea of load cell based probe and I'd like to drop my 2 cents into this discussion.

To be honest, I think this approach is not the way to go. I haven't really looked into Klippers source at alI, but it seems like it's not too trivial to add this sensor type to the list - as it's discussed here. And understandably so - we are talking about a single microcontroller running multiple things, where accuracy in time and space are most crucial. And we want this code to run reliably as well as with good performance.

We don't really need the load cell to act as a scale. We just need it to accurately detect where the nozzle touches the buildplate.
We don't need the load cell to function as a load cell - we need the load cell to function as a switch.

I think you're smart people here, and lots of good work has gone into this already. But I believe this is.. sort of philosophical understanding that needs to happen in this thread: Load cell is not a Z-Probe to be integrated into Klipper. Load cell / strain gauge is a component we need to utilize in order to build a Z-Probe.

In my opinion - load cell based z probes should be handled just the same as we handle piezo based probes, for example.
**We handle the probe signal logic separately.** If a probe needs a probe-specific logic to function as a probe... perhaps that means that the probe needs its own firmware, tiny, little, simple firmware? 
If we put it that way - doesn't it seem almost absurd to try and write a firmware of a probe into the firmware of the printer itself? Not just this probe - all the probes. Imagine if BLTouch came out like "here's a couple of wires for a coil and raw Hall sensor output... figure it out!"

Just throw in another MCU. 
Another MCU, another computing thread - just for the probe to be snappy and happy. 
And most importantly - actually work as a probe, instead of a load cell component, so Klipper - or any other 3D printer firmware can accept it as such.
Klipper stays cleaner. Load Cell Probe logic is easier to write. 
What's not to like...?
MCU costs Money? It's not a lot.

Like 3$ Attiny85 like this guy has done:
https://github.com/sylim5442/Strain-Gauge-Leveling-Probe
Or Ardy Nano. Like you see here:
https://www.instructables.com/Reprap-Load-Cell-Z-Probe/

Or you could design a circuit for the load cell to act as a probe with amp and comparator.

If you were to design this as a product - **_Load cell based_ Z-Probe** - that's what you would do.

Sineos:
Commercial product using a load cell: https://dyzedesign.com/horizon-auto-bed-leveling-sensor/
Works with Klipper out of the box

mhier:
If you separate out the signal processing into a dedicated microcontroller, you are degrading it effectively into a digital switch. The great advantage of the load cell based probe is that it is analogue. That way we can extrapolate the true surface position instead of measure just some position which is close to the surface but has some offset which needs to be calibrated away.

Commercial products may take this into account by choosing a smart design minimising the offset and maybe hence getting away with a fixed factory calibration. This project also aims at people who want to retrofit such system to their own printer. A pair of load cells is cheap, the support structures can be 3D printed, so this technology could be open for everyone, without breaking the bank.

And yes sure this is also about integrating existing printers. Besides: even the quoted commercial product above has apparently an analogue output for which this module could be used.

DaemonWork:
> If you separate out the signal processing into a dedicated microcontroller, you are degrading it effectively into a digital switch. The great advantage of the load cell based probe is that it is analogue. That way we can extrapolate the true surface position instead of measure just some position which is close to the surface but has some offset which needs to be calibrated away.

In my mind, the great advantage of strain gauge based probe is not the fact that it's analogue, but the fact that you can integrate it into the tool head so you can use the nozzle itself to probe, while also having other properties we need. That's where it all stems, no? And a probe in this context, as a Z-Probe for 3D printer, ultimately means that we need to deduce, I wouldn't say degrade, the work of the element to a point where it ultimately becomes a switch.

The analogue nature of the load cell could be used as a measure to accomplish a certain feature we want, but it's not a feature we look for. I might be missing some sort of method you're thinking of implementing which would in theory or practice make the probe more accurate.
But my point is that at the end of the day, the features we look for in a Z-Probe, and 3D printer in general, are accuracy, speed, dependability/stability and cost. Even if the temperature in the room/enclosure changes. Even if the cables/filament moves on the head. -Insert whatever other variables-.
I personally don't care at all if my probe is digital, analogue or quantum, as long as it does what it's supposed to do, so I can get to printing. I don't think the initial offset calibration is an issue. It should still be calibrated even if I change the nozzle and I think that's how most people feel. Getting it as plug-and-play as possible is of course a good thing, but even with analogue readings - there's tons of printers out there, each with different types of beds and frame/toolhead riggidity. I don't think the initial calibration could be avoided anyway unless you know all these variables.

If we can get the probe to work with a deviation of 0.01mm as a switch - that's 'good enough' in my book - and looking at the works of other people - it seems it's certainly possible with a strain gauge, amplifier and a cheap MCU.
If we can use the analogue readings of the cell to introduce a useful feature or further the accuracy of the readings that benefit the quality enough to be worth the extra cost or complexity - be it on hardware or software side - that would be great! 
But it seems like it should be a further development, a "v2", instead of being the core discussion of enabling load-cell probes on Klipper.

webbbn:
I've implanted and am using a form of a load-cell based z-probe that uses a separate microcontroller, and it works well enough, but I don't see why integrating that functionality into Klipper would cause such an issue.

Fundamentally, a load-cell based z-probe is an ADC with a set of functions around it to calibrate, and ultimately turn the output of it into a single value that Klipper can use. It's really not that much different from a thermistor. We don't demand that thermistors come with their own microcontroller with custom calibration procedures, etc. It's not too much of a burden to Klipper to read the value of the thermistor. That's actually more critical than a z-probe, since reading the z-probe is only used during homing and ABL, etc., not during printing.

The reason why I would like to see the functionality integrated into Klipper is because, what I did (throwing a microcontroller at the problem with a simple program that makes it look like a switch probe) is easy, but not generally useful or complete. Others have put much more effort into calibration, etc, which would be nice to formalize in Klipper to make it more generally useful.

FWIW, my load-cell based approach that I discuss above is still working great, except, I moved my Voron onto the same table as my second (bed slinger) printer, and the vibration from that printer can cause false detections if it's printing when I'm probing with the Voron. I can tweak the threshold and generally get it to work, but that's where more sophisticated calibration tools and procedures would be very useful.

clowrey:
I 100% support getting this code into Klipper so we have the ability to read the load cell ADC directly. It is the perfect job for the klipper MCU microcontroller either the main one or secondary ones.

Plus less hardware necessary is always good, all of the new CAN tool head PCBs use a secondary MCU and are programmed via Klipper - it makes perfect sense to have the load cell ADC IC connected to this toolhead MCU and programmed via Klipper which is what this pull request is moving towards. This way any config change to your load cell setup is done in your printer.cfg file vs some separate code base that has to be programmed through a totally different method. 

Examples of similar functionality are the input shaper accelerometer interfaces and many others.

DaemonWork:
> I've implanted and am using a form of a load-cell based z-probe that uses a separate microcontroller, and it works well enough, but I don't see why integrating that functionality into Klipper would cause such an issue.
:)
> Fundamentally, a load-cell based z-probe is an ADC with a set of functions around it to calibrate, and ultimately turn the output of it into a single value that Klipper can use. It's really not that much different from a thermistor. We don't demand that thermistors come with their own microcontroller with custom calibration procedures, etc. It's not too much of a burden to Klipper to read the value of the thermistor. That's actually more critical than a z-probe, since reading the z-probe is only used during homing and ABL, etc., not during printing.

In a sense, load cell works very much like thermistor does, true. Change in resistance, etc.

But how we use them is different. For thermistor we can PID calibrate and have predictable behavior throughout the print. We have the same heater, the same ~mass to heat. So even though we do wish constantly accurate temperature, because we can predict it, it isn't as important to ALWAYS keep an eye on it. Even if there is some sort of delay in the execution of code - even if it's a one whole second, which is huge - it would be something like 200C vs 202C... far from ideal, but not a disaster either.

Whereas for homing/probing you need to be **precise** when it comes to timing. 
1ms delay would mean 0.01mm deviation at 10mm/s.
 
> The reason why I would like to see the functionality integrated into Klipper is because, what I did (throwing a microcontroller at the problem with a simple program that makes it look like a switch probe) is easy, but not generally useful or complete. 

I don't really get this part. Throwing a MCU at the problem is easy, but not useful or complete?
Isn't the probe working? Isn't that useful? What's missing from it?
You go on to say "it's working great", except it's too sensitive in some scenarios, which can be tweaked/calibrated to work.

**I FEEL LIKE A VILLAIN HERE**

I want to say - this is not my goal to ruin your fun or ambitions or whatever. But looking at this, I do believe to see an issue - and I want to express it, or understand it better.

My whole argument is this:
I use Klipper.
Load Cell based Z-Probe seems awesome on paper, I want.
I google - I find these threads here and there, starting from 3-4 years ago - "I want load cell! Can Klipper? No? Okay.. :("

Discussions on implementing HX711 and Load Cell probes seem to start at 2020.

```
The immediate issue I'm seing with this is that the HX711 does not have an "interrupt pin" which allows to have a digital pin indicating a threshold has been crossed. This is problematic because the HX711's current load has to be read by shifting in a 24-bit value, which is quite expensive compared to the simple "read digital pin" which klipper currently does.

These changes would have to be implemented in the low level MCU code as well, which is a bit trickier than writing the python host code.
```
_Originally posted by @FHeilmann in https://github.com/Klipper3d/klipper/issues/2977#issuecomment-642582797_
In 2020, talking about adding Creality CR-6's load cell support.

To me it seems like there's issues with implementing this approach. And maybe rightly so.

**The first issue I have is this:**
This is, on surface level, without digging around, is presented as a NECESSITY to get a load cell based probe to work. Which it's not. Instead of obsessing around adding native whatever support into Klipper - I believe the correct first step would be to provide the actual information and guide to run these probes. 
Right now when I wish to find out if Klipper can support them, I google "Klipper Load Cell" I get all these issues and people saying nooo it wont work, we need to add support for this and that. 
When I google "Marlin Load Cell" I get guides by people who make them work. The same methods apply to Klipper, because they are firmware agnostic and effectively replace the endstop.

So -
Instead of saying "WE NEED to add support to Klipper"
It should be said "YOU NEED to buy 3$ part from AliExpress and install firmware from THIS github page and it will work."
- that's the information I'd like to see and know when I'm searching for information about load cell based probing on klipper.

As it stands anyone who is remotely interested in load cell probe and Klipper would be instantly discouraged. And this interest is surely to rise with this Prusa XL announcement.
Do we want to discourage people away from Klipper and load cell based z-probes? I don't think we do.
I think we want to promote both of them.

This way if one day we have more optimized and easier to calibrate options - more people already have the hardware.
And guess what - it works the opposite way as well. If there are more people who have the hardware - there's more interest to develop better tools and algorithms and whatnot.

**The second issue I have is this:**
When you do dig around and find out that it's actually rather easy to get it working on Klipper or any other firmware... It seems to actually makes sense to do it like that.

As I've tried to point out before, as well as FHeilmann in 2020 - doing the readings this way during homing is a lot more expensive than reading a digital pin.

It simply doesn't seem like a good idea to collect data, analyze and calculate on it - all on the single MCU core while executing the probing/homing sequence. Everything that could be left out from these moves - probably should be. Unless it offers significant benefit.

The whole philosophy behind Klipper is running minimal code on MCU and offloading the expensive calculations. So you get great performance even on that 16mhz AVR.

If the probe can function with its own simple firmware, it makes sense to do it that way. The only argument to put it directly onto Klipper MCU as far as I can see would be cost.
Except load cell + HX711 + ATTiny85 costs literally 5$ from AliExpress.

We should instead figure out the best way to run them before thinking about how to minimize the hardware. In fact, we should probably move on to higher accuracy / sample rate converters like ADS1256. It seems like HX711 and even ADS1100 in this case is the actual bottleneck when it comes to accuracy here. As per this table here:
https://klipper.discourse.group/t/strain-gauge-load-cell-based-endstops/2134/2
It seems like it corresponds pretty well to how accurate people can get these probes to be on HX711. Which is "Good enough" to work, but could be much better.
Load cells are supposedly really accurate. Choosing the correct  hardware (starting with types of cells/gauges and amps/ADC's) is essential.

Work from the ground up.

I have yet to see a good counter argument to this. What's the theory here? How would you use these resistance readings exactly, other than deducing them to simple signal of "Touchy" or "No touchy"? Why can't you do it on a dedicated MCU - why do you need these readings on Klippers MCU?

clowrey:
@DaemonWork "I FEEL LIKE A VILLAIN HERE"
Yes it seems like you are arguing just to prove you viewpoint is correct at this point.. There are obviously other opinions. More options better.. If you don't like this available option, don't use it..

webbbn:
I'm not going to refute that entire post. I think this addresses most of the main points.

First, the HX711 is a very rudimentary sensor. I would hope that we could support more advanced sensors, and I believe the sensor support is what is currently holding up integration of this PR. Even with an supporting microcontroller, the HX711 is not very good.

The actual sensor is not the issue here, it's the overarching support for calibration and integration into Klipper. That's what's missing in my simple "throw a microcontroller at it" approach, which you seem to be recommending. It works, but I don't think you really want to be sshing into your SBC and tweaking hard-coded parameters in the code that controls your probe with numbers that you pull out of your...

Sineos:
* This is a **pull request** on integrating load cell based probes with an analog output in Klipper
* As such it is work in progress / prone to change / not fit for production / targeted at interested and knowledgeable people who want to contribute to exactly this implementation
* This is a place to discuss / review the actual code behind this idea
* This is a place to collect actual feedback on this very implementation

* This is NOT a discussion if this even makes sense
* This is NOT the place to discuss that there are a myriad of alternative solutions

If you want to discuss a more broader scope of this topic or alternative approaches, then revert to https://klipper.discourse.group

mhier:
Besides, I have put it on draft for a reason. It is not yet ready for review!

sbtoonz:
Please reopen Creality K1 printers have the Hx711 sensor and use load cell meshing concepts

mhier:
Sorry, I mostly gave up on this one for now. General interest to integrate this seems to be rather low, people who are interested want to implement their own ideas instead (but have to make them work first). Also the hurdles to get something as complex as this into Klipper are rather high while the feedback time is rather long (and it seems to me that the formal requirements increase faster than the feedback rounds work). If at least some of these obstacles change truly, I will start working on this again :-)

Besides, I think the K1 uses a HX717, which is not so easy to come by as a breakout board, so development of the driver has to be done by an owner of that printer. But: The K1 already uses Klipper. Since Klipper is released under GPLv3, Creality is obliged to release the full (sic!) source code, which they are currently not doing. You can help by putting some pressure on Creality to release the source code of the .o and .so files they have placed into the Github project:

https://github.com/CrealityOfficial/K1_Series_Klipper/tree/main

It does not hurt if as many as possible customers "remind" them that they are legally obliged to do so! (And for the future: don't buy their products. This is not their first GPL violation and they certainly do this knowingly. This behaviour hurts open source projects like Klipper, so if you like Klipper and e.g. Cura, you should boycott Creality and a couple of other companies!)

sbtoonz:
in my experience dealing with GPL violations for lack of source code is
still up to the vendor it took HTC almost 1 year to supply me with
the source for *.ko files they were building alongside their android
kernel. My faith in Creality actually offering up the *.c files that
correspond to the *.o files is VERY low and I don't know enough C to write
a wrapper for their board

Their klipper is running from 12/2022 anyway it is missing tons of
important updates.

I appreciate the input :D

sbtoonz:
> Sorry, I mostly gave up on this one for now. General interest to integrate this seems to be rather low, people who are interested want to implement their own ideas instead (but have to make them work first). Also the hurdles to get something as complex as this into Klipper are rather high while the feedback time is rather long (and it seems to me that the formal requirements increase faster than the feedback rounds work). If at least some of these obstacles change truly, I will start working on this again :-)
> 
> Besides, I think the K1 uses a HX717, which is not so easy to come by as a breakout board, so development of the driver has to be done by an owner of that printer. But: The K1 already uses Klipper. Since Klipper is released under GPLv3, Creality is obliged to release the full (sic!) source code, which they are currently not doing. You can help by putting some pressure on Creality to release the source code of the .o and .so files they have placed into the Github project:
> 
> https://github.com/CrealityOfficial/K1_Series_Klipper/tree/main
> 
> It does not hurt if as many as possible customers "remind" them that they are legally obliged to do so! (And for the future: don't buy their products. This is not their first GPL violation and they certainly do this knowingly. This behaviour hurts open source projects like Klipper, so if you like Klipper and e.g. Cura, you should boycott Creality and a couple of other companies!)

I found all their source code for the objects that produced the .o files 

Its on the fork in my profile :D

mhier:
> Its on the fork in my profile :D

Nice, can you please point me to the individual files? I only found the source for that Python .so library (which is already quite interesting) but not for the .c files.

Sineos:
Isn't this covered here: https://github.com/CrealityOfficial/K1_Series_Klipper/commit/15fddc424eeda5b48fb66153e0664e16ea96f983?

sbtoonz:
The C files only seem to start existing on firmware 1.3.***

I’ve got it all running on latest klipper now and the python wrapper
provided on 1.2.9.15 is enough to make it work 🤷‍♂️

mhier:
> Isn't this covered here: [CrealityOfficial/K1_Series_Klipper@15fddc4](https://github.com/CrealityOfficial/K1_Series_Klipper/commit/15fddc424eeda5b48fb66153e0664e16ea96f983)?

No, that commit introduces binary files (.o files, .so files) for which there is no source code. @sbtoonz found at least one important source file (apparently the most important one?), but others are still missing.

sbtoonz:
> > Isn't this covered here: [CrealityOfficial/K1_Series_Klipper@15fddc4](https://github.com/CrealityOfficial/K1_Series_Klipper/commit/15fddc424eeda5b48fb66153e0664e16ea96f983)?
> 
> No, that commit introduces binary files (.o files, .so files) for which there is no source code. @sbtoonz found at least one important source file (apparently the most important one?), but others are still missing.

all Gd32 source is there
Source for lidar  scanner wrapper is there
Everything to make klipper at HEAD exists in my repo

There were some WONKY things creality does with firmware naming schema (for autoupdates) and i did port that over but unsure if needed yet

There were a couple loadcell *.C files I thought you might wanna look at too the hx71708 and hx711s
https://github.com/sbtoonz/klipper/blob/master/src/hx711s.c
https://github.com/sbtoonz/klipper/blob/master/src/hx71708.c

the whole squashed commit is here

https://github.com/sbtoonz/klipper/commit/66e8d682c7256a9779b21fecf7849cf9173735be

ChipCE:
I'm doing some test with load cell based probe too. It needs to run quite slow due to the sample speed of the adc(10sps or 80sps with hx711). ADC with lower resolution will faster, or more expensive adc will do better, but it will be nice to stick with an affordable option. I'm doing some test with 16bit adc and PGA. I will report is they are good enough
 If we implement something like "fist probe speed" and "second probe speed" into the probe module, we can run the first probe fast, dispose the value of the first probe, raise the toolhead a few mm and do the second probe at slow speed. That will help a lot with the overall probing speed.

mhier:
I don't really understand why people are so obsessed with speed here. I am doing full bed scans maybe once per year, if that takes 20 minutes I really don't care. If you need to re-calibrate the end switch position (e.g. after a nozzle change) you can do that with a single point measurement, which takes maybe 20 seconds. I am doing that in the start code at the beginning of every print, which involves some tricks to have a clean nozzle. Those tricks involve cooling down the nozzle by 20 degrees, which takes several minutes. Any speed-up of the probing process would be totally unnoticeable here either.

So why do you think speed is important for you?

ChipCE:
Not all setup are stable enough to do full bed scan once a year like you. Some setup have Z sagging when power down, some build sheet is uneven compare to others(swap build plate/build sheet), some build plate warp differently based on the temperature. 
About speed, if the probe sample speed is 10sps, you home with speed 10mm/s, the deviation bight be over 1mm? I am wrong with the calculation here?

mhier:
> Some setup have Z sagging when power down,
Sounds like a common error indifferent of the X/Y position -> Do a single point measurement.
Otherwise it sounds like a severe mechanical issue which should be fixed in hardware...

> some build sheet is uneven compare to others(swap build plate/build sheet)

Store one mesh per build plate to fix that.

> some build plate warp differently based on the temperature.

Store one mesh per temperature to fix that.


There will be a good solution for all these problems without re-leveling the whole built plate every day.


> About speed, if the probe sample speed is 10sps, you home with speed 10mm/s, the deviation bight be about 1mm? I am wrong with the calculation here?

My algorithm works differently, I have described it in detail in the .md documentation in this branch. But yes, at 10sps you cannot got 10mm/s because you might go 1mm to far into the bed which might destroy something. The HX711 can do already 80sps which should be ok in theory for 10mm/s to find the coarse position without destroying something. I am using a subsequent scanning procedure to find the exact position at which the nozzle touches the build plate first without applying any force. That is the position you want to know.

Also: I don't say it wouldn't be nice to make the scan faster. It is just not the first thing to bother. Try the working algorithm in reality with actual use cases for a while to gather experience. Only then you can really tell how important speed will be for you...

ChipCE:
@mhier I just read the doc. It's a very clever method. I think it will fit nice with the setup. Thanks for the feedback!

Btw, What do you think about using Instrumentation amplifier instead of HX711? We can feed the output of the amp directly into the mcu ADC.

mhier:
> Btw, What do you think about using Instrumentation amplifier instead of HX711? We can feed the output of the amp directly into the mcu ADC.

The MCUs usually have ADCs with only a limited number of bits (like 12) which may be insufficient. Also you need in principle differential ADCs while most MCUs only offer single-ended ADCs, although this can be fixed by shifting the signal to the middle of the ADC range (or just cutting off one side, if you are not interested in measuring extrusion forces which go into the other direction). Finally, you will need a very stable reference, which you probably won't get easily with an MCU either.

So maybe it works, but I doubt it will be better than using a HX711. The nice thing about a HX711 is that you can get it easily for little money. Most other setups will require some specialised electronics. If you can make your own PCBs this may not be a big issue, but such solution will not work for the masses.

ChipCE:
Yeah, I will need some kind of pcb for that. I'm testing INA126, it has differential ADC inputs, output single-ended analog(Offset-able with something like TL431) and gain of 10000. I'm not sure if I will be able to apply use your code for that. I will repost here if it works.

mhier:
My code should be generic enough to get this working, as long as the measurement range and signal-to-noise level is sufficient. I recommend using load cells which can take around 10kg in total and setting the electronics up in a way that you can measure a maximum force in the range 5 to 10kg. I have two load cells which are rated 5kg each and my ADC will clip around +-8kg. I have 15 bits signed (this is not a HX711 but an ADS1100 at 16 sps) and thus an LSB of around 1 grams. If you have fewer bits, you may need to find some middle ground, but as I said this might be difficult.

Feel free to contact me if you need help, and please report your success and/or failure :-)

KevinOConnor:
FYI, there has been a bunch of discussions on load cells on Klipper Discourse: https://klipper.discourse.group/search?q=load%20cell

-Kevin

mhier:
> FYI, there has been a bunch of discussions on load cells on Klipper Discourse: https://klipper.discourse.group/search?q=load%20cell

I gave up following that after being told to shut up when contributing my experience...

KevinOConnor:
Hi Martin.

The link above is a search of Klipper Discourse for "load cell", and that search returns several topics.

You are free to create a topic on Discourse and share your experiences there, as you are free to share your experiences here.  Others are also free to share their ideas and experiences there and here.

It's important to me that all the Klipper communication channels are inclusive, have a welcoming attitude, and upbeat tone.  If you feel that wasn't achieved feel free to privately reach out to the Klipper Discourse admins or myself.

-Kevin

mhier:
Thanks. I know that. I started a thread about this PR a long time ago and there was very little interest. It is just so many factors which in the end made me stop working on this at least for now. I had the hope that people can work together to bring this into a mergeable condition, but this seems not how it works. People who have the capabilities to contribute are interested more in their own ideas. Other people who could test the code have more interest in plug&play solutions. Even people in the community of my printer have little interest in switching to Klipper - everyone is just talking about it (often complaining that X or Y is still missing) but nobody wants to take the first step. I would have to spend a year or so exclusively (in my spare time) to get this ready on my own, but I have so many other projects which I could not do then.

So as long as nobody approaches me and offers substantial support (at least with testing and debugging), I will be using it happily for myself :-) Maybe this sounds selfish, but getting something like this ready for everyone is like 80% of the total work (so a lot *more* than the original implementation). I am not telling you (Kevin) anything new here, I guess...

sbtoonz:
> Thanks. I know that. I started a thread about this PR a long time ago and there was very little interest. It is just so many factors which in the end made me stop working on this at least for now. I had the hope that people can work together to bring this into a mergeable condition, but this seems not how it works. People who have the capabilities to contribute are interested more in their own ideas. Other people who could test the code have more interest in plug&play solutions. Even people in the community of my printer have little interest in switching to Klipper - everyone is just talking about it (often complaining that X or Y is still missing) but nobody wants to take the first step. I would have to spend a year or so exclusively (in my spare time) to get this ready on my own, but I have so many other projects which I could not do then.
> 
> So as long as nobody approaches me and offers substantial support (at least with testing and debugging), I will be using it happily for myself :-) Maybe this sounds selfish, but getting something like this ready for everyone is like 80% of the total work (so a lot _more_ than the original implementation). I am not telling you (Kevin) anything new here, I guess...

Let me see what I can do about getting this one to work, I have gotten data coming through with another users load cell stuff (garethky) 

if I can get it mocked up I would be happy to try and support with testing and debugging

ChipCE:
Hi, this is some of my test result.
-  Instrumentation amplifier works quite well the output can be feed directly to MCU ADC pin. You will need Atlases about 5000x gain to match the resolution of 16bit ADC with programmable gain like ADS1115. I used TL1167 on my test.
- I tried the strain gauge tape (the small 350 Ohms). They work fine-is, but the initial offset value might vary, depend on how well they installed.
- SMD resistor works like a charm, give them a try.

mhier:
> Hi, this is some of my test result.

Interesting. Did you test this with my code, i.e. are you able to use the MCU ADC to successfully probe the bed?

ChipCE:
> > Hi, this is some of my test result.
> 
> Interesting. Did you test this with my code, i.e. are you able to use the MCU ADC to successfully probe the bed?

Sorry for the late resonse.
I did not test with your module yet. I'm still stuck with some geometry improvement of the sensor and ADC circuit design.

sfcgeorge:
Load cell probing seems ultimate as 1 sensor can do both levelling and Z-height without any offset, I'm surprised more people aren't excited about it!

HX711 seems extremely common; most load cells on AliExpress seem to come with a HX711 breakout for ~$2. This would make the feature widely accessible. Creality also uses an HX711...

In the CR-6 SE printer the HX711 is used in a toolhead board BUT it also has an MCU that just thresholds the HX711 via a pot; effectively turning it into a regular endstop. AKA there's just a Z-endstop pin coming out of that board. I imagine this introduces latency and is less flexible/precise than having it in Klipper. Just thought I'd share it as a temporary approach since the board can be bought separately for $20 on AliExpress "cr-6 se board". There is an open-source reverse engineering of this PCB but sadly not the firmware so it doesn't help this PR https://github.com/CR6Community/Hardware

The Gaggiuino coffee machine project _also_ uses HX711 for weight profiling coffee flow etc, and they _do_ have open firmware! Here is the [scales file](https://github.com/Zer0-bit/gaggiuino/blob/release/stm32-blackpill/src/peripherals/scales.cpp) and the [HX711_2 library they use comes from here](https://github.com/banoz/HX711) with both .cpp and .h files and a licence. Might help in making a working HX711 driver if you haven't already come across it.

mhier:
> Here is the [scales file](https://github.com/Zer0-bit/gaggiuino/blob/release/stm32-blackpill/src/peripherals/scales.cpp) and the [HX711_2 library they use comes from here](https://github.com/banoz/HX711) with both .cpp and .h files and a licence. Might help in making a working HX711 driver if you haven't already come across it.

Thanks, the problem is not the lack of knowledge how to interface with the HX711, this is actually pretty easy. If you look in another branch in my repo you can even find a very rudimentary driver for the HX711 which works with my code in Klipper:

https://github.com/RF1000community/klipper/blob/wip-loadcellprobe/klippy/extras/hx711.py

It is implemented on the Python side and hence slow and unreliable, but for some first tests it should work. Yet there seems to be nobody who wants to try this approach, since people apparently want it to be very fast first (which I consider not particularly important for something I do a few times per year).

PR #5629, opened 2022-07-06
Anycubic 4 Max Pro 2.0: Added two internal systems (beeper and PSU control).

int-0x03:
Hello there! I have added to klipper config for Anycubic 4 Max Pro 2.0 two internal stock systems. First one is internal beeper, and I created M300: "Play Tone" G-Code macro, according to Marlin G-Codes.
Second one, I added M81: "Power Off" G-Code macro, according to Marlin G-Codes. This command will be wait until hotend cool down, and then power off printer. Stock firmware do the same things. 

-- Signed-off-by: Usachev Alexander Valer'evich <int_0x21@protonmail.com>

### DIFF
--- a/config/printer-anycubic-4maxpro-2.0-2021.cfg
+++ b/config/printer-anycubic-4maxpro-2.0-2021.cfg
@@ -140,6 +140,16 @@ screw4: 265, 5
 [filament_switch_sensor filament_sensor]
 switch_pin: ^!PC4
 
+[output_pin buzz]
+pin: PC6
+pwm: True
+
+[output_pin AUTO_POWEROFF]
+pin: PD0
+pwm: True
+cycle_time: 0.02
+value: 1
+
 
 # This macro (M300) uses internal integrated beeper
 # Just use it in your G-code for making sounds. Example: M300 S1000 P500
--- a/config/printer-anycubic-4maxpro-2.0-2021.cfg
+++ b/config/printer-anycubic-4maxpro-2.0-2021.cfg
@@ -11,7 +11,7 @@
 
 # For Anycubic 4Max Pro (not 2.0) owners:
 #       Be careful when using this config! This config tested only on Anycubic
-#       4Max Pro 2.0 with klipper v0.9.1-667-g31761500! At first, you should
+#       4Max Pro 2.0! At first, you should
 #       set homing_speed on 5, and run homing and click on the endstops with
 #       your fingers. It is necessary to make sure that all the motors are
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  It looks okay to me.  I'll give a couple of days to see if others have comments, and otherwise commit.

-Kevin

int-0x03:
Demo video :)

mental405:
Looks ok to me since it is just replicating stock firmware functionality.

int-0x03:
@mental405 It's a good question! Possible you can help me to improve this feature. Just give some advice. Let's deep down to marlin code. In Marlin autopoweroff code for this printer looks like:

```c++
void DwinTFTClass::PowerDown()
{
  //marlin PSU_OFF() don't work
  #if PIN_EXISTS(PS_ON)
    #ifdef DWIN_TFT_DEBUG
      SERIAL_ECHOLNPGM("TFT Serial Debug: Power down");
    #endif
    for(unsigned char i = 0; i < 3; i++)
    {
      WRITE(PS_ON_PIN, LOW);
      delay(10);
      WRITE(PS_ON_PIN, HIGH);
      delay(10);
    }
  #endif
}
```

I just made some calculations, and got 0.02 duty_cycle and pulse width 50%. Just do the same thing, but use PWM. I don't know, maybe this implementation may cause some performance problems, or any other type of problems, but I used it for a long time (about 1 year) and have no problems.

mental405:
I am thinking that is so the PS pin cycles between high and low a few times depending on how the relay is triggered. (Active HIGH vs Active Low) idk, its kind of confusing to me. I would try taking the relay pin and just pulling it low and see what happens. It might turn the thing off without needing the other tomfoolery.

int-0x03:
@mental405 Unfortunately when I was making this functionality for myself I tried HIGH on a pin and I tried LOW.  This does not work.  Only after I repeated the behavior of the Marlin code did I get the result.  Inside is a relay module from Anycubic.

KevinOConnor:
Alas, I don't see a signed-off-by line.  I'll need that to commit - see https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

int-0x03:
> Alas, I don't see a signed-off-by line.  I'll need that to commit - see https://www.klipper3d.org/CONTRIBUTING.html .
> 
> 
> 
> -Kevin

Ok, I will fix it, soon

int-0x03:
Sorry for late feedback. Done.

KevinOConnor:
Thanks.

-Kevin

PR #5632, opened 2022-07-07
Change of alias for heater_fan in multiple configs

JamesH1978:
This PR serves to fix a longstanding misnomer in some config files. Many configs state a nozzle_cooling_fan alias for what is usually a "hotend cooling fan". This causes ambiguity and confusion with the parts fan. I have identified all 24 files with this and changed them here.


KevinOConnor:
Thanks.  In general, the change looks fine to me.  However, I wonder if `heatbreak_cooling_fan` may be more descriptive?

-Kevin

JamesH1978:
Makes sense, but a lot of the community may not know what the heatbreak is... but i don't mind either one, its just that current description is just plain wrong or at least ambigous

James

qwewer0:
I think in function it is a cooling fan for the heatbreaks heatsink, so <s>either</s> Kevin's `heatbreak_cooling_fan` idea <s>or heatsink_cooling_fan</s> would be more appropriate.
(`heatsink_cooling_fan` could be mistaken for the boards cooling, as most if not all has heatsinks for the drives)

JamesH1978:
Ok, I agree :) files now changed to reflect `heatsink_cooling_fan`

Thanks
James

KevinOConnor:
Thanks.

-Kevin

PR #5635, opened 2022-07-09
Update printer-creality-ender3-s1-2021.cfg

ImminentFate:
Added additional comment detailing requirements (root folder needed on the SD card, adjusting the name of the firmware file) when saving firmware to flash to the STM32F401.

This is consistent with the findings from [mriscoc ](https://github.com/mriscoc/Ender3V2S1/wiki/How-to-install-the-firmware) as well as independent testing.

### DIFF
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -13,14 +13,10 @@
 
 # Flash this firmware by copying "out/klipper.bin" to a SD card and
 # turning on the printer with the card inserted. The firmware
-# filename must changed to "firmware.bin"
+# filename must be changed to "firmware.bin"
 
 # On the STM32F401 board, you may need to place the firmware file in a
-# folder on the SD card called "STM32F4_UPDATE". The firmware
-# filename must end in ".bin" and must not match the last filename
-# that was flashed.
-
-# You may need to disconnect your printer's screen before inserting the SD card.
+# folder on the SD card called "STM32F4_UPDATE".
 
 # See docs/Config_Reference.md for a description of parameters.
 
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -15,6 +15,13 @@
 # turning on the printer with the card inserted. The firmware
 # filename must changed to "firmware.bin"
 
+# On the STM32F401 board, you may need to place the firmware file in a
+# folder on the SD card called "STM32F4_UPDATE". The firmware
+# filename must end in ".bin" and must not match the last filename
+# that was flashed.
+
+# You may need to disconnect your printer's screen before inserting the SD card. 
+
 # See docs/Config_Reference.md for a description of parameters.
 
 [stepper_x]
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -20,7 +20,7 @@
 # filename must end in ".bin" and must not match the last filename
 # that was flashed.
 
-# You may need to disconnect your printer's screen before inserting the SD card. 
+# You may need to disconnect your printer's screen before inserting the SD card.
 
 # See docs/Config_Reference.md for a description of parameters.
 


Fisheiyy:
No, this isnt required, you can flash the F4 boards with just the firmware.bin on an sd card, no STM32F4_UPDATE folder is required and you dont need to disconnect any screens before flashing, many people have flashed F4 boards with just the firmware.bin file and no STM32F4_UPDATE folder just fine.

ImminentFate:
@Fisheiyy maybe many people can, but there's definitely a lot of people who cannot. 
Maybe I did something wrong when trying myself, but no other method of leaving it on the root level worked for me. 

1. firmware.bin on the root directory on the included SD card
2. firmware.bin on the root directory on the included SD card after formatting to 4K FAT32
3. [random name].bin on the root directory
4. firmware.bin / [random name].bin on a different SD card (tried: included SD card, 32GB microSD, new 16GB A2 SD card)

Not one of the above worked despite multiple attempts; I even recompiled the firmware on a Pi 4 in case my Pi Zero W was failing somewhere, with no luck. 

I then used the `/STM32F4_UPDATE` folder, and it succeeded, using the same firmware files I'd used before. 

Maybe it's the shipped firmware version? I notice a lot of people flashing onto V1.X.X firmware, mine came with 3.0.3 out of the box. I don't know, but there's multiple instances of people asking the same thing online, and I think that flat out saying "this isn't required" is a bit much when clearly it is in some cases.

Fisheiyy:
> @Fisheiyy maybe many people can, but there's definitely a lot of people who cannot. Maybe I did something wrong when trying myself, but no other method of leaving it on the root level worked for me.
> 
> 1. firmware.bin on the root directory on the included SD card
> 2. firmware.bin on the root directory on the included SD card after formatting to 4K FAT32
> 3. [random name].bin on the root directory
> 4. firmware.bin / [random name].bin on a different SD card (tried: included SD card, 32GB microSD, new 16GB A2 SD card)
> 
> Not one of the above worked despite multiple attempts; I even recompiled the firmware on a Pi 4 in case my Pi Zero W was failing somewhere, with no luck.
> 
> I then used the `/STM32F4_UPDATE` folder, and it succeeded, using the same firmware files I'd used before.
> 
> Maybe it's the shipped firmware version? I notice a lot of people flashing onto V1.X.X firmware, mine came with 3.0.3 out of the box. I don't know, but there's multiple instances of people asking the same thing online, and I think that flat out saying "this isn't required" is a bit much when clearly it is in some cases.

If you go back and look at the original PR to add the S1 and S1 Pro configs, people were initially having issues with flashing and even the STM32F4_UPDATE folder did nothing, then it was figured out the proper settings to compile, are you compiling with the proper settings? STM32F4 processor and 64KiB bootloader with USART on PA10/11 and it should just be accepted no problem without the need of any folder.

ImminentFate:
> If you go back and look at the original PR to add the S1 and S1 Pro configs, people were initially having issues with flashing and even the STM32F4_UPDATE folder did nothing, then it was figured out the proper settings to compile, are you compiling with the proper settings? STM32F4 processor and 64KiB bootloader with USART on PA10/11 and it should just be accepted no problem without the need of any folder.

Yes, all steps were followed correctly, like I said I used the exact same firmware files (even compared hashes in case they got corrupted somewhere) between each run; the only difference between what worked and what didn't, was the folder.

Fisheiyy:
> > If you go back and look at the original PR to add the S1 and S1 Pro configs, people were initially having issues with flashing and even the STM32F4_UPDATE folder did nothing, then it was figured out the proper settings to compile, are you compiling with the proper settings? STM32F4 processor and 64KiB bootloader with USART on PA10/11 and it should just be accepted no problem without the need of any folder.
> 
> Yes, all steps were followed correctly, like I said I used the exact same firmware files (even compared hashes in case they got corrupted somewhere) between each run; the only difference between what worked and what didn't, was the folder.

A new version of stock firmware shouldnt be the cause of any reason that it wouldnt update without, remove line 23 about the disconnecting screens because that just makes no sense, the bit about not having the same filename as the last flashed doesnt matter so line 20 and 21 can be removed and the stuff after the period on line 19 can be removed, and add a "be" after "must" on line 16, also if you are going to modify one config for the s1 then modify the other config for the s1 pro, they are essentially the same thing.

Harrypulvirenti:
I can confirm that putting the firmware file inside the `STM32F4_UPDATE` is required to correctly start the flash procedure.
I spend a couple of hours trying to understand why the process was not starting.
Please add this info in the documentation

Fisheiyy:
> > > If you go back and look at the original PR to add the S1 and S1 Pro configs, people were initially having issues with flashing and even the STM32F4_UPDATE folder did nothing, then it was figured out the proper settings to compile, are you compiling with the proper settings? STM32F4 processor and 64KiB bootloader with USART on PA10/11 and it should just be accepted no problem without the need of any folder.
> > 
> > 
> > Yes, all steps were followed correctly, like I said I used the exact same firmware files (even compared hashes in case they got corrupted somewhere) between each run; the only difference between what worked and what didn't, was the folder.
> 
> A new version of stock firmware shouldnt be the cause of any reason that it wouldnt update without.
> 
> * remove line 23 about the disconnecting screens because that just makes no sense
> * the bit about not having the same filename as the last flashed doesnt matter so line 20 and 21 can be removed
> * the stuff after the period on line 19 can be removed
> 
> And add a "be" after "must" on line 16, also if you are going to modify one config for the s1 then modify the other config for the s1 pro, they are essentially the same thing.

Are you going to make any changes?

Fisheiyy (changes requested):
Good, but now do it for the s1 pro config too.

KevinOConnor:
I think we can just remove the s1pro config and add a comment to the s1 config that it applies to both the s1 and s1pro.

-Kevin

Fisheiyy:
> I think we can just remove the s1pro config and add a comment to the s1 config that it applies to both the s1 and s1pro.
> 
> -Kevin

I guess that works since its only really the max temp for the hotend and bed that is changed, the s1 plus probably should have its own config still because more is changed than just that.

KevinOConnor:
The change looks fine to me, but it needs a signed-off-by line containing your full real name.  See: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

Fisheiyy:
This PR can just be closed, I have handled it in my new PR #5705

PR #3792, opened 2021-01-17
sd_flash: upload klipper firmware to an SD Card

Arksine:
This module implements SD Card file transfer in Klipper, allowing users to update MCU firmware without physical intervention on boards that only support firmware updates via SD Card.  This is accomplished using a host only module, no new MCU code has been introduced.   The advantage of this method is that it isn't restricted to USB connections, it should allow users connected to UART to update as well.

The current state of this module is "proof of concept".  While it is workable, the process to flash is not consistent with Klipper's other methods of flashing.  While I have plans to implement a script to bring it in line with Klipper's other methods, I thought that it would be useful to PR this so we can get user feedback and identify potential issues with certain boards and/or SD Cards.  I can also address issues with the current implementation and tweak it if necessary.  The module configuration and GCodes are documented, however the specific process to flash is not.  If we decide to merge before a script is implemented then I can add that documentation where we think it is appropriate.  For the time being I will document the process here.

Below are some working configurations that I have tested:
- SKR 1.3
```
[sd_flash skr_1_3]
spi_bus: ssp1
cs_pin: P0.6
detect_pin: !P0.27
enable_crc: True
```
- SKR Mini E3 V1.2 (Should work for V2.0 and the MZ 1.0):
```
[sd_flash skr_mini_e3]
spi_bus: spi1
cs_pin: PA4
detect_pin: !PC4
enable_crc: True
``` 
- Flyboard Mini
```
[sd_flash flyboard_mini]
spi_bus: spi2
cs_pin: PB12
detect_pin: !PB11
enable_crc: True
sd_current_firmware_path: FLY.CUR
```

It is possible to configure `sd_flash` for multple MCU's, each should be given a unique name (ie: `[sdflash skr_mini_1]`, `[sd_flash skr_mini_2]`, etc).

For testing, [check out this pull request](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally) on your machine and restart the Klipper service.  I also recommend deleting everything on your SD Card prior to testing, particularly the .CUR file.

The standard process for updating firmware using [sd_flash] is as follows:
1) Log in to the SBC via SSH
2) `cd ~/klipper`
3) `git pull`
4) `make clean`
5) `make menuconfig`
6) `make`
7) Run the `SDFLASH_UPLOAD` gcode.  This can be done using "echo" or via the serial console presented by Octoprint, Mainsail, Fluidd, etc.  It should look something like this:
```
[...]

PR #5642, opened 2022-07-13
spi_flash: add btt-octopus

djessup:
Adds a board definition for BTT Octopus SPI/SD card flashing.

Was only able to get it working using `swspi` (software SPI), but is tested and (mostly) working on a real BTT Octopus v1.0. 

Only caveat is the the Octopus board doesn't seem to pick up the new firmware.bin after a soft-reset and I had to power-cycle it before it loaded the new firmware. Some details here: https://github.com/Klipper3d/klipper/pull/3792#issuecomment-1182706828

### DIFF
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -70,6 +70,12 @@ BOARD_DEFS = {
         'mcu': "stm32f407xx",
         'spi_bus': "spi3a",
         "cs_pin": "PC9"
+    },
+    'btt-octopus': {
+        'mcu': "stm32f446xx",
+        "cs_pin": "PC11",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
     }
 }
 


djessup:
Indeed it seems it is. Wish I had found that PR before I wasted a couple of hours pouring over the schematics 🤦 

Seems a shame not to merge one of these, while it's not perfect that a power-cycle is needed, it's still better than no support at all for these boards.

I'll let one of the maintainers indicate if this should be closed with or without merging.

jakep82:
I don't think it makes sense to merge something with a known issue,and then try to explain through documentation that many people won't read why it only half works. It will lead to help requests when people inevitably try to use it expecting full functionality. 

A better solution IMO is to add SDIO support as was discussed in the previous PR.

zellneralex:
Again a half backed solution is a nightmare for the supporters. If you know what you do that may work for you but think about all the people that come to the various discords and ask for help. So from the support standpoint: No solution is better than a half backed solution that does tell the user it was failing.

djessup:
@zellneralex Thanks for the feedback, and I take your point - I wasn't considering the support angle.

@jakep82 Agreed that SDIO support would be nice!

I'll close this PR.

PR #5643, opened 2022-07-13
configfile: Save backup files to own directory

abmantis:
This changes the directory of the backup files to be `klipper_config/backups` instead of `klipper_config`, making the former directory cleaner.

### DIFF
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -402,10 +402,15 @@ class PrinterConfig:
         if cfgname.endswith(".cfg"):
             backup_name = cfgname[:-4] + datestr + ".cfg"
             temp_name = cfgname[:-4] + "_autosave.cfg"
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  This would be a notable change in behaviour that I think should be discussed on Klipper Discourse to see what reaction there is.  It's not something that I'm comfortable making a sole decision on.

On the technical side, be aware that an `os.rename()` isn't assured to work on Unix if the file is in a different directory.

-Kevin

abmantis:
@KevinOConnor created a thread here: https://klipper.discourse.group/t/save-backup-files-to-a-dedicated-directory/3657/2
Haven't get much reaction and no strong opinions. Indeed, having this configurable as suggested would be nice, but I'm not sure how/where to make it configurable...

PR #5649, opened 2022-07-19
Make installation paths and some variables configurable

kageurufu:
```sh
KLIPPER_VENV=$HOME/.venvs/klipper ./scripts/install-debian.sh
AURCLIENT='yay -S' ./scripts/install-arch.sh
```

I have been using a relocated virtualenv like this for a while now,
 and relocated all my virtualenvs to a single ~/venvs folder for
 organizations-sake. Personally, I set KLIPPER_VENV in my .profile
 and everything just works.

If you would like, I can update documentation to explain these
 options.

Signed-off-by: Franklyn Tackitt <git@frank.af>

### DIFF
--- a/scripts/flash-sdcard.sh
+++ b/scripts/flash-sdcard.sh
@@ -3,7 +3,7 @@
 #  unattended firmware updates on boards with "SD Card" bootloaders
 
 # Non-standard installations may need to change this location
-KLIPPY_ENV="${HOME}/klippy-env/bin/python"
+KLIPPY_ENV="${KLIPPER_VENV:-${HOME}/klippy-env/bin/python}"
 SRCDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/.. && pwd )"
 KLIPPER_BIN="${SRCDIR}/out/klipper.bin"
 KLIPPER_BIN_DEFAULT=$KLIPPER_BIN
--- a/scripts/install-arch.sh
+++ b/scripts/install-arch.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
 # This script installs Klipper on an Arch Linux system
 
-PYTHONDIR="${HOME}/klippy-env"
+PYTHONDIR="${KLIPPER_VENV:-${HOME}/klippy-env}"
 SYSTEMDDIR="/etc/systemd/system"
-AURCLIENT="pamac"
-KLIPPER_USER=$USER
-KLIPPER_GROUP=$KLIPPER_USER
+AURCLIENT="${AURCLIENT:-pamac build}"
+KLIPPER_USER="${KLIPPER_USER:-$USER}"
+KLIPPER_GROUP="${KLIPPER_GROUP:-$KLIPPER_USER}"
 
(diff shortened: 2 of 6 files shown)


Sineos:
I like this idea. Basically did the same just recently as the various packages like octoprint, moonraker etc. are causing a folder mess.

Arksine:
I agree that the folder configuration is unorganized.   I have plans to change this in Moonraker, however its a breaking change so I really need to find a way to automate it given the number of users on FluiddPi and MainsailOS.

kageurufu:
For reference, moonraker and mobileraker_companion has merged PRs supported this method of configuration, and I've submitted a PR to [KlipperScreen](https://github.com/jordanruthe/KlipperScreen/pull/680) as well

KevinOConnor:
Thanks.  I agree it would be good to have a standard way of specifying the klipper host paths.  However, I'm leery of having the scripts pull in arbitrary variables from the Unix shell environment.  In other projects I found this can lead to a "spaghetti nest" of hard to document dependencies between scripts.

I'm not sure about alternatives.

-Kevin

kageurufu:
So my idea behind using namespaced environment variables are ways to make it easy to tweak, and allow other scripts to just use the same `${KLIPPER_VENV:-$HOME/klippy-env}/bin/xxx` type expansion to easily access the virtualenv. Making it more complex, configured through something else makes it harder for external projects to properly locate the venv once its relocated. Its fairly common practice among unix utilities to work this way too. Using something like a .env file is pretty easy to use, but pulling values from a .env in python is a lot harder than just `os.environ['KLIPPER_VENV']`.

If you have an idea of another way to do this, I'm happy to change things.

KevinOConnor:
FWIW, I'm not confident that using environment variables nor a .env file is a good complexity trade-off.  I'm not sure what alternatives there are, but I'd likely lean towards leaving things as they are.

Is the main goal to find the appropriate `python` binary?

-Kevin

Sineos:
> Is the main goal to find the appropriate `python` binary?

It offers a convienient way to install Klipper's venv to a different path, e.g. `~/all_my_vens/klipper`
I think for more advanced users this is a nice twist. Especially since a lot of tools in the Klipper world are using venvs so the `home` folder gets pretty cluttered fast.
Normal users probably never bother (just imagine, you need to look it up in the documentation 😉 ) so I'd see no harm.

KevinOConnor:
> It offers a convienient way to install Klipper's venv to a different path

Right, but the only thing needed from that venv is the Python binary?  That's what I meant by my question earlier - is the main goal to find the (python venv environment containing the) appropriate python binary?

Cheers,
-Kevin

Sineos:
> Right, but the only thing needed from that venv is the Python binary?

I'm not sure, I understand what you are referring to. From a user (i.e. non developer) perspective:
* Many venvs of different installations are cluttering the home folder
* Having an easy way to move the different venvs into a dedicated folder would be nice

Purely a cosmetic thing for people with an obsessive-compulsory disorder WRT such things 😄

PR #5652, opened 2022-07-20
manual_probe: report status

pedrolamas:
Exposes a `ManualProbe` status, that will be set by `ManualProbeHelper`.

When not running, status is:

```python
{
  'is_active': False,
  'z_position': None,
  'z_position_lower': None,
  'z_position_upper': None
}
```

When a manual probe is initiated, `ManualProbeHelper` will set `active` to `True` and the remaining values will update accordingly.

This will allow frontends like Fluidd and Mainsail to show an "assistant/wizard" type interface to help with these commands!

This PR is a follow up on this Discourse topic: https://klipper.discourse.group/t/exposing-calibration-testz-accept-abort-state/3511

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -244,6 +244,17 @@ The following information is available for each `[led led_name]`,
   chain could be accessed at
   `printer["neopixel <config_name>"].color_data[1][2]`.
 
+## manual_probe
+
+The following information is available in the
+`manual_probe` object:
+- `is_active`: Returns True if a manual probing helper script is currently
+active.
+- `z_position`: The current height of the nozzle (as the printer currently
+understands it).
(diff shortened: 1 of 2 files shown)


pedrolamas:
Now that I think of it, the `prev_pos` and `next_pos` naming could probably be better as `prev_lower_pos` and `prev_upper_pos` respectively, or something similar...

KevinOConnor:
Thanks.  In general it seems fine to me.  I have a few comments:
1. Please spell out the full names - for example: `z_position` instead of `z_pos`.
2. The status reference documentation would need to be updated.

-Kevin

pedrolamas:
Thank you, Kevin!

I've just updated the PR with improved labels and added documentation.

KevinOConnor:
Thanks.

-Kevin

PR #5654, opened 2022-07-21
adc: Account for jitter when rescheduling samples

Fraxul:
Recompute the waketime for the ADC sample timer relative to the current time instead of the previous waketime when the hardware is busy. If `gpio_adc_sample` repeatedly asks for a delay, scheduling jitter will eventually cause us to compute a waketime in the past and trip the scheduler safety checks.

I ran into this issue when setting up Klipper on a new BTT SKR Mini E3 v3.0 board (STM32G0-based). Without this patch, the board would crash with "Rescheduled timer in the past" -- randomly, but typically within an hour after power-on/reset. After this patch, I haven't seen any more problems and have been able to complete several prints.

During initial debugging, I noticed in the crash log receive queue dump that the repeated `analog_in_state` messages for one of the channels went missing shortly before the crash; that's what led me down this path. [My sample crash log](https://gist.github.com/Fraxul/d6d72d8dcb3ad1226c3fe8d8a0f4ff57)

I also noted the same pattern in crash logs found when searching for other user reports of this issue. In all cases, there were missing `analog_in_state` reports immediately prior to the crash.
[Sample 1](https://klipper.discourse.group/t/mcu-mcu-shutdown-rescheduled-timer-in-the-past/3293) 
[Sample 2](https://klipper.discourse.group/t/skr-e3-mini-v3-mcu-mcu-shutdown-rescheduled-timer-in-the-past/3456)
[Sample 3](https://klipper.discourse.group/t/random-mcu-rescheduled-timer-errors/2812)
[Sample 4](https://github.com/bigtreetech/BIGTREETECH-SKR-mini-E3/issues/667)

### DIFF
--- a/src/stm32/stm32f0_adc.c
+++ b/src/stm32/stm32f0_adc.c
@@ -119,7 +119,20 @@ gpio_adc_sample(struct gpio_adc g)
             return 0;
         goto need_delay;
     }
+#if CONFIG_MACH_STM32G0
+    // Acknowledge EOC flag, clear Channel Config Ready before updating CHSELR
+    adc->ISR = ADC_ISR_EOC | ADC_ISR_CCRDY;
+    adc->ISR;
+#endif
+
     adc->CHSELR = g.chan;
+
+#if CONFIG_MACH_STM32G0
+    // Wait for Channel Config Ready
+    while (!(adc->ISR & ADC_ISR_CCRDY))
+        ;
+#endif
+
     adc->CR = CR_FLAGS | ADC_CR_ADSTART;
 
 need_delay:


iwm911:
Thank you @Fraxul, let's hope that this is the issue.

Zandhoeg:
Did the changes, rebooted pi, did a homing and got a crash. So for me this did not solve it. Unless I did something wrong.

Fraxul:
> Did the changes, rebooted pi, did a homing and got a crash. So for me this did not solve it. Unless I did something wrong
> [klippy_cut.txt](https://github.com/Klipper3d/klipper/files/9161775/klippy_cut.txt)

This does appear to be the same issue -- your microcontroller misses some `analog_in_state` reports for OIDs 24 and 13 right before it shuts down with "Rescheduled timer in the past".

Did you rebuild and flash the firmware after updating?

Zandhoeg:
@Fraxul no I did not rebuild the firmware. I will try this next then and see.
Everything is as per usual? USART2 on PA3/PA2 (according to the schematics). Seen some hinting towards PA10/PA9 and I do not recall what one I used as I type this =)

inOut47:
the fix is in the firmware, if you didn't rebuilt and flashed you didn't deployed the fix :)

Fraxul:
> @Fraxul no I did not rebuild the firmware. I will try this next then and see.
> Everything is as per usual? USART2 on PA3/PA2 (according to the schematics). Seen some hinting towards PA10/PA9 and I do not recall what one I used as I type this =)

Yep, no configuration changes are necessary. If you updated a checkout that you had previously configured, you should be able to just run `make` in there again and it'll rebuild with your previous config. You should see at least `Compiling out/src/adccmds.o` on the output (since that's the file I touched).

Zandhoeg:
I run two (different) boards so I had to do re-do both. But I was unsure what was what for the boards.
They are made and patched, so now we wait? =)

Fraxul:
Now we wait. Please report back with any crashes (or lack thereof)!

Zandhoeg:
> Now we wait. Please report back with any crashes (or lack thereof)!

I done the things I did before that caused the error and nothing so far. All that is lacking is the time aspect. So hopefully no crashed in the days to come will prove this as a solution.

Macsbig:
Sorry, this is probably a really lame question, but how do I patch the firmware?

Zandhoeg:
> Sorry, this is probably a really lame question, but how do I patch the firmware?

You do the `make menuconfig` for your board, then go to klipper/out on your RPi with a ftp program and transfer the file to a SD-card formated to FAT32. renmae the file from `klipper.bin` to `firmware.bin` and put the sd-card in the board.

Have the board powered off, then turn it on and it should flash, confirm with putting the card in your computer and see if it been renamed from `.bin` to `.cur`. Reboot the card and klipper etc and try it out.

Fraxul:
If you have an existing git checkout, you can fetch this pull request and create a branch from it. (These commands assumes your remote for this project is named `origin`, which is the default):

```
git fetch origin refs/pull/5654/head
git checkout -B adc-scheduling-fix FETCH_HEAD
```

This will create a local branch called `adc-scheduling-fix` with the code from this PR and switch your working copy to it. You can then build (run `make`) and flash as usual.

Macsbig:
I'm confused. That's just what I do to flash the board. So what are we doing that's different so the code is "patched" to stop the mcu error?

Macsbig:
Ok thanks. I'll have to do some Google-ing, most of what you just wrote went over my head lol. I'm not that experienced with git. 

Thanks for the info tho 👍

Zandhoeg:
> Ok thanks. I'll have to do some Google-ing, most of what you just wrote went over my head lol. I'm not that experienced with git.
> 
> Thanks for the info tho 👍

What you can do is go to [https://github.com/Klipper3d/klipper/pull/5654/files](https://github.com/Klipper3d/klipper/pull/5654/files) and read what lines been changed. So d/l the same file from your RPi, open it in notepad++ for example, and change the lines mentioned, save it and upload it again. Then reboot RPi and do the `make menuconfig` again and flash etc.

If you do not want to do the `git fetch` however I think editing like that will make your klipper show up as "dirty" (meaning you made changes/edits that are not officially added)

Zandhoeg:
> git fetch origin refs/pull/5654/head
> git checkout -B adc-scheduling-fix FETCH_HEAD

if I done a manual edit and flashed from that, would this change anything other than potentially remove the "dirty" tag?

Macsbig:
> > Ok thanks. I'll have to do some Google-ing, most of what you just wrote went over my head lol. I'm not that experienced with git.
> > Thanks for the info tho 👍
> 
> What you can do is go to https://github.com/Klipper3d/klipper/pull/5654/files and read what lines been changed. So d/l the same file from your RPi, open it in notepad++ for example, and change the lines mentioned, save it and upload it again. Then reboot RPi and do the `make menuconfig` again and flash etc.
> 
> If you do not want to do the `git fetch` however I think editing like that will make your klipper show up as "dirty" (meaning you made changes/edits that are not officially added)



Thanks for the ideas. I'll have a look, and a Google, and take a look over the next day or two. I have updated mainsail/klipper/moonraker/linux to the latest versions; as leaving it at the Mainsail OS downloaded versions didn't make any difference (thought as it all worked ok before the "updates" I'd try that.)

I really should get my head around git. It's a great resource.

KevinOConnor:
Thanks.  Interesting find.

However, the change here is not correct - the host is expecting the mcu to report the temperature at a precise interval - the mcu should not alter that schedule.  If the adc code is requesting too many "retry" attempts it indicates there is a defect in the low-level adc code for that micro-controller.  It will be necessary to find and fix that root error.

Is the issue limited to the stm32g0 mcus (src/stm32/stm32f0_adc.c)?

-Kevin

celtare21:
I can also confirm the bug on an SKR mini E3 V3. Was not happening on the V2. temperature_mcu also seems to go crazy after some time: 
image

Zandhoeg:
> I can also confirm the bug on a SKR mini E3 V3. Was not happening on the V2. 

I never had this on the SKR MINI E3 v2 either, however that uses the `STM32F103` and not the `STM32G0B1` so can't compare them as such, need to find another card with `STM32G0B1` (not Mini E3 v3) and see if that have the same issues/errors and work from there? Unless it is something more specific to this card than that.

celtare21:
> STM32G0B1

I think the SKR Pico also runs that. Maybe if anyone has one on hand and doesn't mind running a test.

jakep82:
> > STM32G0B1
> 
> 
> 
> I think the SKR Pico also runs that. Maybe if anyone has one on hand and doesn't mind running a test.

The SKR Pico does not use this MCU. It uses the RP2040.

Zandhoeg:
@KevinOConnor and @Fraxul even if the solution is wrong it seems to have done something right. Because so far I have not had a single error.
Then if everything is OK or of this might break something else I do not know, but so far it seem to have solved this. Had my printer running almost 24/7 since I applied the patch without issue.

iwm911:
Without a doubt it's a work around.

I think that given the magnitude of the issue it should be implemented for now while investigating the issue.

The SKR mini e3 v3 is a popular board and the bug render Klipper useless.

KevinOConnor:
There is definitely a pattern to all of the above error logs.  It seems to occur when multiple ADC channels are in use, and for some reason, all but one channel stops working.  Oddly, the one channel that does work continues to work normally.

That type of error seems consistent with the hardware not clearing the EOC bit.  For those that are experiencing the issue, could you go back to the master branch, apply the following change to the code, reflash the micro-controller, and see if the error occurs?

```diff
--- a/src/stm32/stm32f0_adc.c
+++ b/src/stm32/stm32f0_adc.c
@@ -131,7 +131,9 @@ uint16_t
 gpio_adc_read(struct gpio_adc g)
 {
     ADC_TypeDef *adc = ADC1;
-    return adc->DR;
+    uint16_t dr = adc->DR;
+    adc->ISR = ADC_ISR_EOC;
+    return dr;
 }
 
 // Cancel a sample that may have been started with gpio_adc_sample()
```

If this doesn't fix the issue, I'll put together a debugging branch to see if we can gain more data.

-Kevin

Fraxul:
I rolled my patch back, applied yours, and will run that for a while to see what happens.

KevinOConnor:
FYI, it looks like there is a requirement in the stm32g0 specification that the code does not follow: `When the software configures the channels (by programming ADC_CHSELR or changing CHSELRMOD or SCANDIR), it must wait until the CCRDY flag rises before configuring again or starting conversions, otherwise the new configuration (or the START bit) is ignored. Once the flag is asserted, if the software needs to configure again the channels, it must clear the CCRDY flag before proceeding with a new configuration.`

That could explain this error.  I'll put together a fix for the above tomorrow.  Probably not worth spending time on the previous code change.

-Kevin

Fraxul:
Still happens with the previous patch -- I got 3 layers into a Benchy before my luck ran out. It's definitely the same issue, but here's the log anyway: https://gist.github.com/Fraxul/f04cdad550e3bff6f17ba4520724c0ec

Fraxul:
Wrote a quick patch to wait for the CCRDY flag and updated the PR. Needs further testing, but at least my printer made it through a calibration cube without crashing.

KevinOConnor:
Thanks.  I made a code change for the CCRDY bit (commit 3796a319).  It's not clear if this change will fix the issue here, but it is a necessary change regardless.  I recommend anyone experiencing this issue checkout the latest Klipper master branch and see if the problem persists.

-Kevin

Zandhoeg:
@KevinOConnor so we revert the changes from this completely and use the [3796a31](https://github.com/Klipper3d/klipper/commit/3796a319599e84b58886ec6f733277bfe4f1a747) instead? Is there a danger using the first fix as in core security features may not work? Asking mainly as I am not able to swap yet due to a long print going.

Fraxul:
My printer completed a Benchy and idled overnight with no crashes on my version of the change (24155b8). I have now switched it over to 3796a31 and will be doing more testing.

Fraxul:
I've been running 3796a31 continuously for hours with no problems. I think it's safe to say that the problem is fixed in that revision.

KevinOConnor:
Thanks for troubleshooting and reporting this issue.  I know how hard it is to track these things down.

-Kevin

PR #5657, opened 2022-07-22
Hardware PWM support for STM32G0

### LINKED ISSUE #5655: PWM fan on SKR mini E3 V3
Hi!
I'm try enable hardware PWM on Klipper config To fan0 or fan1 or fan

And get this error.

> MCU Protocol error
> This is frequently caused by running an older version of the
> firmware on the MCU(s). Fix by recompiling and flashing the
> firmware.
> 
> Your Klipper version is: v0.10.0-526-ga151aa8c
> MCU(s) which should be updated:
> 
> Up-to-date MCU(s):
> mcu: Current version v0.10.0-526-ga151aa8c
> 
> Once the underlying issue is corrected, use the "RESTART"
> command to reload the config and restart the host software.
> 
> mcu 'mcu': Firmware constant 'PWM_MAX' not found

firmware compiled correctly for STM32G0B1

When fans are connected, with software PWM, the fans either shake hard or they squeak if try change cycle_time.
Tried Sunon 5015 24v maglev and Younuon 4020 dual ball/5015 24v

Pmant:
This will add support for hardware PWM on SKR mini E3 V3.0 fan headers. 
Some fans require hardware PWM to be PWM controlled, specifically Sunon fans will not work well with software PWM.
[Another person with the same issue](https://github.com/bigtreetech/BIGTREETECH-SKR-mini-E3/issues/669).

A Sunon 5015 24V part cooling fan will work well with this config:
```
[fan]
pin: PC6
off_below: 0.07
cycle_time: 0.00003125
hardware_pwm: True 
```

Additional pins could easily be added according to the [spec sheet](https://www.st.com/resource/en/datasheet/stm32g0b1rc.pdf) Table 15-20 but I'm not sure which TIMs to choose. TIM2 seems to be blocked and cannot be used for hardware PWM.

fixes #5655


JeremyKennedy:
Hi - I am trying to use this on my printer for a laser engraver as per [these](https://github.com/Klipper3d/klipper/blob/master/docs/Using_PWM_Tools.md) docs, which specifies hardware PWM. I'm using the SKR Mini E3 V3.0 (STM32G0B1) with this config:
```
[output_pin TOOL]
pin: PC6
pwm: True
hardware_pwm: True
cycle_time: 0.001
shutdown_value: 0
maximum_mcu_duration: 5
```

KevinOConnor:
Thanks.  I think it is a good idea to add the stm32g0 hardware pwm definitions.  I'm a little concerned about only added the pwm pins for the mini-e3 though.  It looks like #5714 may be a superset of this PR.

-Kevin

PR #5658, opened 2022-07-24
Fixed formatting of BME(P)280/680 to properly reflecting parsing.

Shuudoushi:
Simply moved "i2c_address" under "i2c_bus" to properly reflect the order in which the info is parsed.

Spent a solid hour trying to figure why I couldn't get something so simple working...

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2251,11 +2251,11 @@ temperature.
 
 ```
 sensor_type: BME280
+#i2c_mcu:
+#i2c_bus:
 #i2c_address:
 #   Default is 118 (0x76). Some BME280 sensors have an address of 119
 #   (0x77).
-#i2c_mcu:
-#i2c_bus:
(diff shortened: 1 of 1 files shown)


Shuudoushi:
It kept throwing errors for not being able to write to the I2C device, errors for "bus i2c.1 not found", and mcu errors with the address above mcu and bus. Moved address below, and works flawlessly.

KevinOConnor:
I'm not sure what error you were experiencing, but the order of config options within a config section does not impact the option parsing.  If ordering is causing some unusual issue then it will be necessary to track down and fix the root cause.

-Kevin

Shuudoushi:
I gone back to recreate the errors to get screenshots, but it's now working fine... Best guess, I ran into some strange edge case that corrected through another edit to the config. Which edit? No idea... For now I'm just going to close this PR until I can figure out exactly what was going on.

PR #5660, opened 2022-07-25
bed_screws: report status

pedrolamas:
Exposes the current status of `BED_SCREWS_ADJUST` under `bed_screws`:

When not running, status is:

```python
{
  'is_active': False,
  'state': None,
  'current_screw': 0,
  'accepted_screws': 0
}
```

When a `BED_SCREWS_ADJUST` command is initiated, it will set active to True and the remaining values will update accordingly.

I've decided not to expose the "calculated" screws data inside bed_screws.py as we can get that directly from the config.

This will allow frontends like Fluidd and Mainsail to show an "assistant/wizard" type interface to help with the `BED_SCREWS_ADJUST` command!

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -28,6 +28,17 @@ The following information is available in the
 - `profiles`: The set of currently defined profiles as setup
    using BED_MESH_PROFILE.
 
+## bed_screws
+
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #5662, opened 2022-07-26
Added the possibility to get where the "cone shape" of the build volume starts from Macros

Hywelmartin:
Added the possibility to get where the "cone shape" of the build volume starts from Macros

Signed-off-by: Martin Malmqvist Volcomosq@gmx.com

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -427,8 +427,8 @@ The following information is available in the `toolhead` object
 - `axis_minimum`, `axis_maximum`: The axis travel limits (mm) after
   homing.  It is possible to access the x, y, z components of this
   limit value (eg, `axis_minimum.x`, `axis_maximum.z`).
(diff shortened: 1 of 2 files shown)


mental405:
On first inspection this does not appear to be a major or problematic addition. You do, however, need to provide a developer sign off line in your PR before it can be merged.

Hywelmartin:
sorry forgot about that

KevinOConnor:
Thanks.  However, any "printer status" variables that are exported will require a corresponding change to docs/Status_Reference.md .  Also, I'm not sure "limit_z" is a public facing name for this variable.

-Kevin

Hywelmartin:
limit_z was the expression in delta.py so I kept it... Now it's cone_start_z if that's a more viable name.. mentioned in status_references as well. Thanks for your time

KevinOConnor:
Thanks.

-Kevin

PR #5663, opened 2022-07-26
Assorted canbus improvements

KevinOConnor:
This PR contains a few separate improvements that may be of interest to canbus users.

1. USB to canbus bridge nodes should now be automatically detected and these bridge nodes will be restarted last during a FIRMWARE_RESTART.  This should make it possible for the bridge nodes to forward restart messages to the downstream nodes before the bridge is reset.  In order for this feature to work, it is necessary to reflash the bridge mcu with the latest version of the code from this PR.
2. The internal "bootloader request" mechanism has been refactored.  It should now be possible to instruct a usb to canbus bridge node to enter the bootloader even if that node is not running Canboot.  (Note, though, flash_usb.py support has not been added, so actually flashing after entering the bootloader would currently still require a second step.)
3. The low-level host serialqueue.c timing for canbus nodes has been improved.  This may make a minor improvement to timing predictions when using slow canbus speeds (eg, 250000).  The updated code can make improved timing predictions based on minimum transmit times for canbus messages (similar to the existing timing code for uart communications).

-Kevin


gauix4:
Hello ! :)
Sorry if i'm not in the right place :/
Think you do the CanBus for STM32H7 (SKR 3 board) ? :)
thank you very much keep it up ! :)

jakep82:
> Hello ! :) Sorry if i'm not in the right place :/ Think you do the CanBus for STM32H7 (SKR 3 board) ? :) thank you very much keep it up ! :)

You're correct, this is not the right place.  Feature requests belong on [Discourse](https://www.klipper3d.org/Contact.html#i-have-a-feature-request)

RimmeVG:
> This PR contains a few separate improvements that may be of interest to canbus users.
> 
>     1. USB to canbus bridge nodes should now be automatically detected and these bridge nodes will be restarted last during a FIRMWARE_RESTART.  This should make it possible for the bridge nodes to forward restart messages to the downstream nodes before the bridge is reset.  In order for this feature to work, it is necessary to reflash the bridge mcu with the latest version of the code from this PR.
> 
>     2. The internal "bootloader request" mechanism has been refactored.  It should now be possible to instruct a usb to canbus bridge node to enter the bootloader even if that node is not running Canboot.  (Note, though, flash_usb.py support has not been added, so actually flashing after entering the bootloader would currently still require a second step.)
> 
>     3. The low-level host serialqueue.c timing for canbus nodes has been improved.  This may make a minor improvement to timing predictions when using slow canbus speeds (eg, 250000).  The updated code can make improved timing predictions based on minimum transmit times for canbus messages (similar to the existing timing code for uart communications).
> 
> 
> -Kevin

This week i have mks skipr as usb to can bridge connection to ebb sb2209 just setup. Changed auto to allow-hotplug. Updated klipper to latest version, flashed mcu's after.
When i start printer, homing and qgl works.
But after restart firmware i get error 19. Need to restart a few times before it works.
So i have the same problem now as other guys before your fix in point 1

PR #5668, opened 2022-07-29
stm32: add FDCAN support for STM32H743 (SKR-3 Series)

bigtreetech:


bigtreetech:
The function is normal in preliminary test,  And the code content is being cleaned up to meet the contribution criteria.

bigtreetech:
Hello. This PR is ready for merged. Has been tested on SKR-3 (STM32H743) and EBB (STM32G0B1)

src/stm32/fdcan.c
```diff
@@ -128,8 +145,14 @@ canbus_set_filter(uint32_t id)
...
     can_filter(1, id);
     can_filter(2, id + 1);
+
+#if CONFIG_MACH_STM32G0
     SOC_CAN->RXGFC = ((id ? 3 : 1) << FDCAN_RXGFC_LSS_Pos
                       | 0x02 << FDCAN_RXGFC_ANFS_Pos);
+#elif CONFIG_MACH_STM32H7
+    SOC_CAN->SIDFC |= 3 << FDCAN_SIDFC_LSS_Pos;
+    SOC_CAN->GFC = 0x02 << FDCAN_GFC_ANFS_Pos;
+#endif
```
KevinOConnor, line 155: Out of curiosity, why does the stm32g0 code use `id ? 3 : 1` while the h7 unconditionally uses `3`?

src/stm32/fdcan.c
```diff
@@ -251,6 +274,29 @@ can_init(void)
...
+    SOC_CAN->SIDFC |= flssa << FDCAN_SIDFC_FLSSA_Pos;
+
+    SOC_CAN->RXF0C |= ((f0sa << FDCAN_RXF0C_F0SA_Pos)
+                       | (3 << FDCAN_RXF0C_F0S_Pos));
+    SOC_CAN->RXESC |= ((7 << FDCAN_RXESC_F1DS_Pos)
+                       | (7 << FDCAN_RXESC_F0DS_Pos));
+
+    SOC_CAN->TXBC |= ((tbsa << FDCAN_TXBC_TBSA_Pos)
+                       | (3 << FDCAN_TXBC_TFQS_Pos));
+    SOC_CAN->TXESC |= 7 << FDCAN_TXESC_TBDS_Pos;
```
KevinOConnor, line 297: Out of curiosity, can these registers be set explicitly instead of using read-modify-write sequences?

KevinOConnor (commented):
Thanks.  Looks fine to me, but I have a few minor questions - see below.

-Kevin

src/stm32/fdcan.c
bigtreetech: Yes, Modified to write register value directly.

src/stm32/fdcan.c
bigtreetech: stm32h7 is also changed to `id ? 3 : 1`.

KevinOConnor:
Thanks.

-Kevin

PR #5672, opened 2022-07-31
add feedrate information to all movements (menu.cfg)

schnello:
For moves (10,1,0.1) the feedrate information was missing.
This commit updates the X and Y moves to F6000
Z to F1500

Signed-off-by: Christian Schnellrieder <schnellrieder.cs@gmail.com>

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,9 @@ All dates in this document are approximate.
 
 ## Changes
 
+20220821: The movement entries in the "menu.cfg" for 10/1/0.1 mm are
+expanded with the feedrate. For X and Y F6000. For Z F1500.
+
 20220616: It was previously possible to flash an rp2040 in bootloader
 mode by running `make flash FLASH_DEVICE=first`. The equivalent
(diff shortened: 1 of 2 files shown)


KevinOConnor:
In general, it seems okay to explicitly set the feedrate in menu actions, but it's hard to guess what an appropriate default feedrate is.  It might be best to open a topic on Klipper Discourse to see if others have ideas on what the values should be.  At a minimum, a change like this would need to be documented in docs/Config_Changes.md .

-Kevin

schnello:
Thanks for the feedback.

I used the same default values as Mainsail. But I will open a discussion in Discord and add the change also to the Config_Changes.md

schnello:
@KevinOConnor 
Update Config_Changes.md done
Discussion startet:
https://klipper.discourse.group/t/default-force-for-menu-cfg/3909

schnello:
@KevinOConnor 
So far there is no other opinion about it. 
From my point of view:
slow enough to be safe but fast enough to work well with it.

@mcmatrix 
From your side any input regarding this topic?

PR #5673, opened 2022-07-31
Added new kinematics for polargraph drawbot.

tsogoo:
Inspired by https://github.com/gwisp2/klipper-drawbot. 
Didn't include pulley radius for calculation of position. I thought it's too small to calculate.
Tested with arduino uno board with cnc shield.

### DIFF
--- a/config/example-polargraph-drawbot.cfg
+++ b/config/example-polargraph-drawbot.cfg
@@ -13,11 +13,11 @@
 # [drawbot]
 # width: 400               distance between left stepper and right stepper.
 # [homing_override]
-# set_position_x: 275      initial distance between left stepper and pen.
-# set_position_y: 275      initial distance between right stepper and pen.
+# set_position_x: 200      initial location x coordinate.
+# set_position_y: 200      initial location y coordinate..
 
 [mcu]
-# You may need to change this.
+# You may need to change this. ie: /dev/ttyUSB0
 serial: /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
 # Has to be the same as in firmware
 baud: 250000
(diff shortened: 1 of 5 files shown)


KevinOConnor:
Thanks.  I'm not sure I understand this PR - it looks like it does something with polar kinematics but all the links in the description refer to a corexy wire implementation.

In any case, commits need to have a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review .  It may also make sense to start a topic on Klipper Discourse to share your machine and implementation details.

Cheers,
-Kevin

dmbutyugin:
Based on the name, I think the PR is actually for a device like this:
https://en.wikipedia.org/wiki/Polar_plotter
https://www.instructables.com/Polargraph-Drawing-Machine

So, besides the requirement of a singed-off-by line for commits, it would be good to reference the actual machine design in the PR. But as per Kevin's comment, it might be worthwhile to post this work and get some feedback on Klipper Discourse or somewhere else first. Otherwise, the PR looks reasonable (with the exception of the fake stepper_z; it might be possible to 'fake' non-homed Z axis instead, but I'm not sure if that's entirely possible).

tsogoo:
> Hey there (especially @tsogoo ) I would love to participate here, because I think this is a great idea. I have a Klipper/Moonraker/Mainsail Setup at home connected to a Makelangelo 5 ( http://www.makelangelo.com/ ) Drawing robot and try to get it working. To be able to test it, i probably need a bit more insight in parameters and configuration, because the Machine has a Rumba board, endstops, a special home point... As github removed the possibility to communicate, i can not get in touch directly.

Thank you to interest this pull request. You can find some config parameter from https://github.com/tsogoo/klipper/blob/master/config/example-polargraph-drawbot.cfg. Currently It doesn't have any endstop calculation.
'width' parameter is distance between left step motor and right step motor.(400)
'set_position_x' and 'set_position_y' is initial position of gantry. Calibration point is (width/2, width/2).

[stepper_l] , [stepper_r]
rotation_distance: 32 # depends on pulley radius

PR #5676, opened 2022-07-31
Add BQ Hephestos printer

rcarmo:
Supersedes #5607, which had to be reverted. Also adds a few minor kinematics changes.


KevinOConnor:
Config lines should be in `option: value` format, not `option = value` format - see: https://www.klipper3d.org/Example_Configs.html

Also, can you explain the 0.8mm rotation distance of the Z axis?  If there is a gearing mechanism on the Z it is generally preferable to specify a `gear_ratio`.

-Kevin

rcarmo:
The twin Z-axis is directly driving a threaded rod. This is the value that works for me.

KevinOConnor:
Thanks.

-Kevin

PR #5679, opened 2022-08-01
Edit of configuration file

JackyHe398:
Add gcode support and better parking when pause and cancel.
Add beeper support


config/printer-anycubic-vyper-2021.cfg
```diff
@@ -141,3 +169,139 @@ gcode:
     G4 P300
     SET_PIN PIN=probe_reset_pin VALUE=1
     G4 P100
+
+[virtual_sdcard]
+path: ~/gcode_files
```
Sineos, line 174: Optional. Only needed when using Octoprint but not for the modern UIs

config/printer-anycubic-vyper-2021.cfg
```diff
@@ -141,3 +169,139 @@ gcode:
...
+gcode:
+    {% set e = params.E|default(10) %}        #edit to your retract length
+    PAUSE_BASE
+    {% if printer.extruder.can_extrude|lower == 'true' %}
+      G1 E-{e} F2100
+    {% else %}
+      {action_respond_info("Extruder not hot enough")}
+    {% endif %}
+    END_POS
+    M140 S55
```
Sineos, line 192: Not required. Optional. Not for base config files

config/printer-anycubic-vyper-2021.cfg
```diff
@@ -141,3 +169,139 @@ gcode:
...
+      {% set get_params = "" %}
+    {% endif %}
+    ##### end of definitions #####
+    G91
+    {% if printer.extruder.can_extrude|lower == 'true' %}
+      G1 E{e} F2100
+    {% else %}
+      {action_respond_info("Extruder not hot enough")}
+    {% endif %}
+    RESUME_BASE {get_params}
```
Sineos, line 217: Not required. Optional. Not for base config files

config/printer-anycubic-vyper-2021.cfg
```diff
@@ -141,3 +169,139 @@ gcode:
...
+
+[gcode_macro CANCEL_PRINT]
+description: Cancel the actual running print
+rename_existing: CANCEL_PRINT_BASE
+gcode:
+    TURN_OFF_HEATERS
+    CANCEL_PRINT_BASE
+    END_POS
+    M84
```
Sineos, line 226: Not required. Optional. Not for base config files

config/printer-anycubic-vyper-2021.cfg
```diff
@@ -141,3 +169,139 @@ gcode:
...
+    {% endif %}
+    ##### end of definitions #####
+    G91
+    {% if "xyz" in printer.toolhead.homed_axes %}
+      G1 Z{z_safe}
+      G90
+      G1 X{x} Y{y} F6000
+    {% else %}
+      {action_respond_info("Printer not homed")}
+    {% endif %}
```
Sineos, line 254: Not required. Optional. Not for base config files

config/printer-anycubic-vyper-2021.cfg
```diff
@@ -141,3 +169,139 @@ gcode:
...
+    SAVE_CONFIG
+
+[gcode_macro M600]
+gcode:
+    PAUSE
+
+[gcode_macro preheat]
+gcode:
+    SET_HEATER_TEMPERATURE HEATER=heater_bed TARGET=55
+    SET_HEATER_TEMPERATURE HEATER=extruder TARGET=190
```
Sineos, line 307: Not required. Optional. Not for base config files

config/printer-anycubic-vyper-2021.cfg
```diff
@@ -104,21 +106,47 @@ pin: PA0
...
+move_check_distance: 3
+horizontal_move_z: 2
 
 [filament_switch_sensor runout]
 pause_on_runout: True
 switch_pin: PA5
+runout_gcode:
+    M300
+insert_gcode:
+    M300
```
Sineos, line 149: See below

Sineos (changes requested):
The official configs should "mimic" the stock printer. Additional macros are typically not added to such configs unless necessary for stock functionality.

config/printer-anycubic-vyper-2021.cfg
JackyHe398: This is base on the current config file. I just add a function to the part of moving to park position. It raise the readibility.

config/printer-anycubic-vyper-2021.cfg
JackyHe398: same, this one is base on the current file. I just make it better to use. And I really think that is neccessary because when you pause the print, it just stop there. 

Those Gcode simulate how anycubic vyper move without klipper. In My view, Pause and Resume should be kept. But if not, maybe we need to delete the code in the current file.

JackyHe398:
> The official configs should "mimic" the stock printer. Additional macros are typically not added to such configs unless necessary for stock functionality.

Thank you for your advice. 

First, I think those macros are pretty important for klipper so i add them in. I can understand that they might not be neccessary when configing the klipper with printe.cfg. But klipper report a bunch of error without those code. So i think that they will be better here. 

Second, those gcode in PAUSE and RESUME is simulating the action of the vanlila anycubic vyper(The one without klipper). It will be better for me to add them.

For me, these config files are references for the new comers. With those macros, it will be easier. The pros will definitly delete it because they know what they want. I don't really mind to delete those if they are illegal. But is there any ways for me to distribute it on the official platform of klipper?

Thank you so much for pointing out those for me

PR #5683, opened 2022-08-01
scripts: Update Debian install script

SteveGotthardt:
Only 'buster' and 'stretch' have python-dev. The other releases have python2-dev

### DIFF
--- a/scripts/install-debian.sh
+++ b/scripts/install-debian.sh
@@ -1,18 +1,24 @@
 #!/bin/bash
-# This script installs Klipper on an debian
+# This script installs Klipper on debian
 #
 
 PYTHONDIR="${HOME}/klippy-env"
 SYSTEMDDIR="/etc/systemd/system"
 KLIPPER_USER=$USER
 KLIPPER_GROUP=$KLIPPER_USER
-DEBIAN_VER=$(/usr/bin/lsb_release -c -s)
 
+OS_ID=$(awk -F= '$1=="ID" {print $2}' /etc/os-release)
+if [ "$OS_ID" != "debian" ]
+then
+    echo "This script supports debian not $OS_ID"
+    exit
+fi
 
 # Step 1: Install system packages
 install_packages()
 {
-    if [ "$DEBIAN_VER" = "buster" ] || [ "$DEBIAN_VER" = "stretch" ];
+    DEBIAN_CODENAME=$(awk -F= '$1=="VERSION_CODENAME" {print $2}' /etc/os-release)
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  It looks okay to me.  @BlackStump - do you have any comments?

-Kevin

BlackStump:
all good as far as I can see

BlackStump:
I just tried a clean Debian install with this change and received a error-
/usr/bin/lsb_release: No such file or directory

I typically use a console version for a BeagleBone Black Install of Debian, LSB-Core is not installed on the console so maybe a check to see if LSB-Core is installed and if not install first.

BlackStump:
next problem for anything BeagleBone Black-
Package lsb-core is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source.

SteveGotthardt:
thanks I assumed all Debian provided lsb_release.
I will use:   /etc/os-release
can you do me a favor and tell me if that file exists on BB black?
thanks

BlackStump:
Just checked it does have
`debian@beaglebone:~$ cat /etc/os-release`
`PRETTY_NAME="Debian GNU/Linux 10 (buster)"`
`NAME="Debian GNU/Linux"`
`VERSION_ID="10"`
`VERSION="10 (buster)"`
`VERSION_CODENAME=buster`
`ID=debian`
`HOME_URL="https://www.debian.org/"`
`SUPPORT_URL="https://www.debian.org/support"`
`BUG_REPORT_URL="https://bugs.debian.org/"`

SteveGotthardt:
@BlackStump 
I have updated the script and tested on my machine and additional generated os-release files and I find it to work.
Thanks for your consideration.

th33xitus:
Wouldn't it be easier to just search the apt cache for the desired package and if found, add it to the PKGLIST? You would get rid of the os-release checks and the guard clause at the beginning.

KevinOConnor:
Just thinking out loud - any thoughts on creating a new `install-debian-buster.sh` script (or rename the existing script and have `install-debian.sh` be for latest distros)?  I'm a little leery about complex script actions in install scripts that effectively run as root.

-Kevin

SteveGotthardt:
@KevinOConnor 
Debian stretch seems to be the only distro name that needs the python-dev. Even Buster has python2-dev. So maybe rename to install-debian-stretch.sh  the others can use the python2-dev in the install-debian.sh

KevinOConnor:
That makes sense to me.

-Kevin

SteveGotthardt:
redo on the way

PR #5687, opened 2022-08-02
heaters: Reset prev_temp_int to zero on target==0

zarecor60:
prev_temp_int would hold its value through on-off cycle of the heater, resulting in inconsistent PID behavior between first run and later runs. Later step changes would have more overshoot due to more accumulated integral gain.

This resets the values of prev_temp_int and temp_int when target==0, resulting in more consistent overshoot & pid behavior between on/off cycles of heaters.

Signed-off-by: Trey Zarecor <zarecor60@gmail.com>

### DIFF
--- a/klippy/extras/heaters.py
+++ b/klippy/extras/heaters.py
@@ -198,7 +198,11 @@ class ControlPID:
                           + temp_diff) / self.min_deriv_time
         # Calculate accumulated temperature "error"
         temp_err = target_temp - temp
-        temp_integ = self.prev_temp_integ + temp_err * time_diff
+        if target_temp == 0:
+            temp_integ = 0
+            self.prev_temp_integ = 0
+        else:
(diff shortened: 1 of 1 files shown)


zarecor60:
This can be recreated using a tuned PID controlled heater using the following steps:
  1: Restart the klippy service
  2: Give the heater a step change, and measure the amount of overshoot
  3: Set the temperature back to zero and allow it to stabilize at ambient 
  4: Give the system the same step change as before, and measure this overshoot

This second step change consistently provides more overshoot in response to a step change.

KevinOConnor:
Thanks.  I agree that the PIDs mechanism for finding the base pwm value for a given temperature is a little "goofy", but I'm not sure that adding special cases to the code is an improvement.  (That is, I fear it would just be "goofy in a different way".)  There has been some discussions recently on the PID code on the Klipper Discourse server - might be worthwhile to discuss this there.

-Kevin

PR #5689, opened 2022-08-03
Update Creality Ender S1 Config

Harrypulvirenti:
I found that the provided configuration for the stock `Ender 3 S1` is moving the bed outside of the physical limits of the printer.

The result is that the bed is crashing heavily even during a bed mesh calibration.

These changes update the end-stop positions, axis limits and bed mesh points.

The new end-stop positions and limits are targeting the official printing area of 220*220mm

I also added a section for the bed screws position.

### DIFF
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -25,9 +25,9 @@ enable_pin: !PC3
 microsteps: 16
 rotation_distance: 40
 endstop_pin: !PA5
-position_endstop: -10
-position_max: 235
-position_min: -15
+position_endstop: -11
+position_max: 240
+position_min: -11
 homing_speed: 50
 
 [stepper_y]
@@ -37,9 +37,9 @@ enable_pin: !PC3
 microsteps: 16
 rotation_distance: 40
 endstop_pin: !PA6
-position_endstop: -10
-position_max: 241
-position_min: -15
+position_endstop: -8
+position_max: 224
+position_min: -8
 homing_speed: 50
 
 [stepper_z]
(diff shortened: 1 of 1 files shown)


config/printer-creality-ender3-s1-2021.cfg
```diff
@@ -109,18 +109,23 @@ probe_with_touch_mode: true
...
+screw3: 195, 195
+screw4: 20, 195
+
 [safe_z_home]
 home_xy_position: 147, 154
 speed: 75
-z_hop: 5
+z_hop: 10
 z_hop_speed: 5
-move_to_previous: true
```
Harrypulvirenti, line 123: Removed to keep the probe to the centre of the bed and, in this way, being able to execute correctly the `PROBE_CALIBRATE` command

Fisheiyy:
now addressed in #5705

Fisheiyy:
This can be closed

Harrypulvirenti:
@Fisheiyy done.
I hope that next time I will have the possibility to contribute as well 😉

steti:
It looks like https://github.com/Klipper3d/klipper/pull/5689 updates the max temperature for this config, but it does not address the axis limits that were changed here. I can confirm that the `position_endstop`, `position_max`, `position_min` as defined in this file on `master` cause my Ender 3 S1 to crash when executing a `PAUSE` macro. Updating my printer.cfg to the values changed in this PR causes it to stop crashing.

Fisheiyy:
> It looks like #5689 updates the max temperature for this config, but it does not address the axis limits that were changed here. I can confirm that the `position_endstop`, `position_max`, `position_min` as defined in this file on `master` cause my Ender 3 S1 to crash when executing a `PAUSE` macro. Updating my printer.cfg to the values changed in this PR causes it to stop crashing.

seems to be a machine to machine problem not a true problem with the config, many things could cause physical differences in locations of endstops, tensioners, and the distance between the v wheels for the bed because of the eccentric nuts

PR #5692, opened 2022-08-05
Update printer-anycubic-vyper-2021.cfg

JackyHe398:
Add beeper, LED support, change homing speed

module: printer-anycubic-vyper-2021.cfg

Add LED and beeper support

Signed-off-by: He King Hung hekinghung@gmail.com


ld3300:
You should probably remove all of the other changes your file is making to the existing commit so only the beeper part get pulled.

JackyHe398:
> You should probably remove all of the other changes your file is making to the existing commit so only the beeper part get pulled.

maybe I simply remove the virtual SD card part and change the headline. The value I edit is following the suggested value of anycubic/ the tested value. If it is a really big problem then pls tell me. I will change it. But if not, just simply leave it there~ Thank you for your advice

Fisheiyy (changes requested):
Everything needs to be changed back to default except for the [output_pin BEEPER_PIN] section

JackyHe398:
> Everything needs to be changed back to default except for the [output_pin BEEPER_PIN] section

ok

Fisheiyy (approved):
Good, now we wait to see if Kevin will merge this or not

KevinOConnor:
It's fine with me, but:
1. Commits need to have a signed-off-by line containing your full real name: https://www.klipper3d.org/CONTRIBUTING.html
2. Example configs should not explicitly set values to their defaults: https://www.klipper3d.org/Example_Configs.html point 7f.

@cryd-s - do you have any comments on this?

-Kevin

JackyHe398:
module: printer-anycubic-vyper-2021.cfg

Add LED and beeper support

Signed-off-by: He King Hung hekinghung@gmail.com

KevinOConnor:
Thanks.

-Kevin

PR #5697, opened 2022-08-08
Do not write to console when variables are saved

MapleLeafMakers:
Prevents `Variable Saved` from being written to the console every time SAVE_VARIABLE is called.

### DIFF
--- a/klippy/extras/save_variables.py
+++ b/klippy/extras/save_variables.py
@@ -54,7 +54,6 @@ class SaveVariables:
             msg = "Unable to save variable"
             logging.exception(msg)
             raise gcmd.error(msg)
-        gcmd.respond_info("Variable Saved")
         self.loadVariables()
     def get_status(self, eventtime):
         return {'variables': self.allVariables}


MapleLeafMakers:
Self Review:

- only a single line of code was removed, a call to gcmd.respond_info.
- All regression tests pass
- contains no debugging code
- no comments were added or removed
- no documentation changes are required

it provides a benefit to anyone making use of the save_variable command from a macro or other gcode that's called frequently. 
it does this by preventing the console from being completely spammed with 'VARIABLE SAVED' messages.

Yes

Yes.

No documentation update should be required.

Yes

zellneralex:
You forgot the signed-off-by: line. See https://www.klipper3d.org/CONTRIBUTING.html #3 see last sentence.

MapleLeafMakers:
> You forgot the signed-off-by: line. See https://www.klipper3d.org/CONTRIBUTING.html
> Point 3 last sentence.

hmmm... i'm seeing it at the end of my commit message, did I do something wrong here? 
image

zellneralex (commented):
Looks ok for me

zellneralex:
I overseen that, most put it in the description.

KevinOConnor:
Thanks.  I'm not sure about removing the message though, as it indicates that Klipper has successfully written out the disk file.  The "save variables" feature is intended to be used to persist settings across software restarts, so I'm a little surprised that it is being called with such a high frequency.  Can you provide a little background on what it is being used for in your setup?  Can the similar `SET_GCODE_VARIABLE` command be used instead?

Cheers,
-Kevin

MapleLeafMakers:
Perhaps I'm wrong about this, but I can't imagine a situation where someone is manually running SAVE_VARIABLES from the command line, it seems like the only sane use cases are inside other more complex macros, which can output their own more informative messages, so it's really just unnecessary noise in those cases.

Personally I use this in combination with a delayed_gcode that runs periodically to track how long my carbon filter has been on and offers a reminder to change the carbon.  Because it's often running for the duration of long print jobs, a sudden crash or power outage can result in tens untracked hours, So it writes to disk occasionally to minimize those losses.. over the course of a long print this completely fills the console for hundreds of pages with nothing but 'Variable saved'.

Zodiak1993:
Thanks for this. Maybe it would be better to integrate this here:

[save_variables]
#filename:
#   Required - provide a filename that would be used to save the
#   variables to disk e.g. ~/variables.cfg

# **console_output: true/false**

https://www.klipper3d.org/Config_Reference.html#save_variables

My console is also full of save variables ;)

MapleLeafMakers:
> Thanks for this. Maybe it would be better to integrate this here:
> 
> ```
> [save_variables]
> #filename:
> #   Required - provide a filename that would be used to save the
> #   variables to disk e.g. ~/variables.cfg
> 
> #  console_output: true/false
> ```
> 
> https://www.klipper3d.org/Config_Reference.html#save_variables
> 
> My console is also full of save variables ;)

I'm not totally opposed to this option, but making it configurable adds extra complexity, and 'future work' in terms of maintenance, and documentation.  

I think I prefer to approach this as a bug-fix, like this is basically debug output that shouldn't have been there in the first place.

Sineos:
Not an expert, but wouldn't `logging.info` work as a middle ground? It ends up in the log but not in the webinterface.

KevinOConnor:
Okay, thanks.

-Kevin

PR #5698, opened 2022-08-09
delayed_gcode: Add ability to query delayed_gcode state

ld3300:
Adds the command 'QUERY_DELAYED_GCODE' which returns either the remaining time of the macro or states 'not running'

Adds 'remaining' as a status reference for delayed_gcode.  Returns float of number of seconds remaining, or 0 if not running.

Signed-off-by: David O. Smith <davidosmith@gmail.com>


KevinOConnor:
Thanks.  It's fine if you want to report the trigger time in a "printer status" variable, but I don't think we should add a command just to report this info.  (Also, I wouldn't report the trigger time and not the time remaining as the latter is hard to use programaticaly.)  Finally, I'd defer to any comments from @arksine .

-Kevin

ld3300:
Thanks, Kevin,
Sorry, I'm not sure I understand.  My thought was that the command provided 2 options, one was for troubleshooting, the other was to periodically print time remaining during a long macro, but thinking it over now I see how you could do the same with the printer status variable, then the actual text output could be selected by the user.  We definitely don't need to add more to the g-codes page for users to sort through.  I will remove that.

I put remaining time as a printer status variable I don't actually know how to programmatically use trigger time.  My usage would be an "if remaining time > 0" or for long commands to get feedback like: "respond 'xxxx gcode %s remaining time' & ...remaining..."
I'm happy to learn, but I'm not sure how to use trigger time... would you compare this to the system_stats cputime to determine remaining time?

That is an easy change, I just want to make sure I understand it.

Thank you,
David

Arksine:
I don't have much to add, I do agree that reporting the trigger time via `get_status` would be optimal.  I suspect to make this useful we will need to add `eventtime` to the gcode_macro's Jinja2 context, ie:
```diff
diff --git a/klippy/extras/gcode_macro.py b/klippy/extras/gcode_macro.py
index 4f50c75a..7f027d9d 100644
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -97,6 +97,7 @@ class PrinterGCodeMacro:
     def create_template_context(self, eventtime=None):
         return {
             'printer': GetStatusWrapper(self.printer, eventtime),
+            'eventtime': eventtime or self.printer.get_reactor().monotonic(),
             'action_emergency_stop': self._action_emergency_stop,
             'action_respond_info': self._action_respond_info,
             'action_raise_error': self._action_raise_error,
```


Then the comparison would be something like:
```
{% if eventtime < printer["delayed_gcode my_gc"].trigger_time %}
# ...do stuff
{% endif %}
````

KevinOConnor:
Right - report the absolute trigger time and subtract it from the current time in macros that want the elapsed time.  It isn't necessary to export `eventtime` as one can export a print time and use the existing `printer.toolhead.estimated_print_time`

-Kevin

ld3300:
Sorry for the delay on this.  I have been unexpectedly busy.  I will have to play with the print time you mentioned.  I am not sure I get it, but most of what I have learned has been through trying stupid things, so I'm sure I'll get there.

PR #5606, opened 2022-06-26
Add MCU temp for Stm32h7

adelyser:
module: stm32h7 ADC
Add support for reading the mcu temp on the stm32h7

Signed-off-by: Aaron DeLyser [bluwolf@gmail.com](mailto:bluwolf@gmail.com)

PR #5596, opened 2022-06-23
stm32: stm32g0/h7 dfu support

bigtreetech:
### Description
As described in part `4.1 Bootloader activation` of [AN2606](https://www.st.com/content/ccc/resource/technical/document/application_note/b9/9b/16/3a/12/1e/40/0c/CD00167594.pdf/files/CD00167594.pdf/jcr:content/translations/en.CD00167594.pdf) manual.  Before jumping to bootloader user must:
* Disable all peripheral clocks
* Disable used PLL
* Disable interrupts
* Clear pending interrupts

image

So `check_usb_dfu_bootloader();` of STM32G0 is moved behind Reset clock registers, and add Reset clock part for STM32H7 

### now Issues
`dfu-util: Error during download get_status` will appear after `make flash`. 
`sudo dfu-util -p 1-1.2 -R -a 0 -s 0x8020000 -D out/klipper.bin` No error message
`sudo dfu-util -p 1-1.2 -R -a 0 -s 0x8020000:leave -D out/klipper.bin` has error message, So the error  is reported in the `:leave` process. But in fact, the MCU has normally left the DFU mode and entered the klipper application. We can completely ignore this error message.
And I upgraded `dfu-util` to 0.11, which is still the same error in `:leave`

PR #5703, opened 2022-08-11
Fix software switching to DFU mode with stm32f429 MCU

visilya:
stm32f4.c:
This fix make possible to go to DFU mode with STM32F429 and flash it using command:
make flash FLASH_DEVICE={FLASH_DEVICE_PATH}

Where FLASH_DEVICE_PATH could be found with:
ls /dev/serial/by-path/

Exmple:
make flash FLASH_DEVICE=/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.0

All flashes correctly but some errors are displayed after flash process finished. Just ignore them.
Same errors appears when you switch to DFU mode with jumper, it's a dfu-util errors, not klipper.

Signed-off-by: Ilya Vislotsky <i.vislotsky@gmail.com>

### DIFF
--- a/src/stm32/stm32f4.c
+++ b/src/stm32/stm32f4.c
@@ -211,7 +211,22 @@ usb_hid_bootloader(void)
 static void
 usb_reboot_for_dfu_bootloader(void)
 {
-    irq_disable();
+    if (CONFIG_MACH_STM32F429) 
+    {
+        void (*SysMemBootJump)(void);
+        volatile uint32_t addr = 0x1FFF0000;
+
+        SysTick->CTRL = 0;
+        SysTick->LOAD = 0;
+        SysTick->VAL = 0;
+
+        SYSCFG->MEMRMP = 0x01;
+        SysMemBootJump = (void (*)(void)) (*((uint32_t *)(addr + 4)));
+
+        __set_MSP(*(uint32_t *)addr);
+
+        SysMemBootJump();
+    }    irq_disable();
     *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;
     NVIC_SystemReset();
 }
--- a/src/stm32/stm32f4.c
+++ b/src/stm32/stm32f4.c
@@ -211,20 +211,16 @@ usb_hid_bootloader(void)
 static void
 usb_reboot_for_dfu_bootloader(void)
 {
-    if (CONFIG_MACH_STM32F429) 
+    if (CONFIG_MACH_STM32F429)
     {
         void (*SysMemBootJump)(void);
         volatile uint32_t addr = 0x1FFF0000;
-
         SysTick->CTRL = 0;
         SysTick->LOAD = 0;
         SysTick->VAL = 0;
-
         SYSCFG->MEMRMP = 0x01;
         SysMemBootJump = (void (*)(void)) (*((uint32_t *)(addr + 4)));
-
         __set_MSP(*(uint32_t *)addr);
-
         SysMemBootJump();
     }
     irq_disable();
--- a/src/stm32/stm32f4.c
+++ b/src/stm32/stm32f4.c
@@ -226,7 +226,8 @@ usb_reboot_for_dfu_bootloader(void)
         __set_MSP(*(uint32_t *)addr);
 
         SysMemBootJump();
-    }    irq_disable();
+    }
+    irq_disable();
     *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;
     NVIC_SystemReset();
 }


Sineos:
Thanks for your contribution.
At minimum a signed-off line as per https://www.klipper3d.org/CONTRIBUTING.html would be needed.

> All flashes correctly but some errors are displayed after flash process finished. Just ignore them.

A feature that ends with errors will not and should not be accepted. For the regular user it is not evident if these errors are real or not. This will also create a support burden in the future.

visilya:
> Thanks for your contribution. At minimum a signed-off line as per https://www.klipper3d.org/CONTRIBUTING.html would be needed.
> 
> > All flashes correctly but some errors are displayed after flash process finished. Just ignore them.
> 
> A feature that ends with errors will not and should not be accepted. For the regular user it is not evident if these errors are real or not. This will also create a support burden in the future.

I corrected message. Errors goes from dfu-util. There are same errors when flashing MCU without this commit but with hardware switching in DFU.

joaobarros:
> A feature that ends with errors will not and should not be accepted. For the regular user it is not evident if these errors are real or not. This will also create a support burden in the future.

This is not the first MCU that dfu-util can properly flash and spew output errors at the end of the process.

I understand your point but I’d rather have flash capability with output errors than no flash at all.

miklschmidt:
> > A feature that ends with errors will not and should not be accepted. For the regular user it is not evident if these errors are real or not. This will also create a support burden in the future.
> 
> This is not the first MCU that dfu-util can properly flash and spew output errors at the end of the process.
> 
> I understand your point but I’d rather have flash capability with output errors than no flash at all.

As a matter of fact, i don't think any of the boards i have on hand flash with no errors from dfu-util. Would have to double check to be sure, but i seem to remember this being the case for the octopus, octopus pro, skr 3, plus all the mellow and btt toolboards. @Sineos this seems like a rule that hasn't been followed until now, is there any ongoing effort to fix it for the rest of them?

Sineos:
If this contribution does not add any new errors (was not clear in the initial post) I would assume it is fine and the call is anyway on @KevinOConnor side.

Ramalama2:
> stm32f4.c: This fix make possible to go to DFU mode with STM32F429 and flash it using command: make flash FLASH_DEVICE={FLASH_DEVICE_PATH}
> 
> Where FLASH_DEVICE_PATH could be found with: ls /dev/serial/by-path/
> 
> Exmple: make flash FLASH_DEVICE=/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.0
> 
> All flashes correctly but some errors are displayed after flash process finished. Just ignore them. Same errors appears when you switch to DFU mode with jumper, it's a dfu-util errors, not klipper.
> 
> Signed-off-by: Ilya Vislotsky [i.vislotsky@gmail.com](mailto:i.vislotsky@gmail.com)

Do that errors come from dfu-util v0.9?
Because here i wasn't able to flash with 0.9 at all and updated from the testing (sid) branch to 0.11

0.11 is able to flash my (octopus pro) f429 at all and without any errors.

Cheers

visilya:
> > stm32f4.c: This fix make possible to go to DFU mode with STM32F429 and flash it using command: make flash FLASH_DEVICE={FLASH_DEVICE_PATH}
> > Where FLASH_DEVICE_PATH could be found with: ls /dev/serial/by-path/
> > Exmple: make flash FLASH_DEVICE=/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.0
> > All flashes correctly but some errors are displayed after flash process finished. Just ignore them. Same errors appears when you switch to DFU mode with jumper, it's a dfu-util errors, not klipper.
> > Signed-off-by: Ilya Vislotsky [i.vislotsky@gmail.com](mailto:i.vislotsky@gmail.com)
> 
> Do that errors come from dfu-util v0.9? Because here i wasn't able to flash with 0.9 at all and updated from the testing (sid) branch to 0.11
> 
> 0.11 is able to flash my (octopus pro) f429 at all and without any errors.
> 
> Cheers

Dfu-util 0.9 can't flash 429 at all. Ver.1.1 can. And it throw err that it can't switch 429 to normal mode (but it does). And another err about prefix, it could be solved with dfu-prefix util. I think this command could be added to make pipeline later.

Ramalama2:
> > > stm32f4.c: This fix make possible to go to DFU mode with STM32F429 and flash it using command: make flash FLASH_DEVICE={FLASH_DEVICE_PATH}
> > > Where FLASH_DEVICE_PATH could be found with: ls /dev/serial/by-path/
> > > Exmple: make flash FLASH_DEVICE=/dev/serial/by-path/platform-fd500000.pcie-pci-0000:01:00.0-usb-0:1.4:1.0
> > > All flashes correctly but some errors are displayed after flash process finished. Just ignore them. Same errors appears when you switch to DFU mode with jumper, it's a dfu-util errors, not klipper.
> > > Signed-off-by: Ilya Vislotsky [i.vislotsky@gmail.com](mailto:i.vislotsky@gmail.com)
> > 
> > 
> > Do that errors come from dfu-util v0.9? Because here i wasn't able to flash with 0.9 at all and updated from the testing (sid) branch to 0.11
> > 0.11 is able to flash my (octopus pro) f429 at all and without any errors.
> > Cheers
> 
> Dfu-util 0.9 can't flash 429 at all. Ver.1.1 can. And it throw err that it can't switch 429 to normal mode (but it does). And another err about prefix, it could be solved with dfu-prefix util. I think this command could be added to make pipeline later.

@KevinOConnor 
What is your opinion of dfu-util?
The background is simply that buster/bullseye comes with v0.9, that simply doesn't work for newer boards.
At least with f429, probably h7 either?

I see 2 solutions for this:
1. Ship an own package, but this has to much downsides i think
2. Add an hint in the docs, how to upgrade it on buster/bullseye?

@Sineos 
The errors shouldn't be a problem at all, because bigtreetech sended some pull requests either, which throw the errors already and no one even questioned it, it was simply merged.
https://github.com/Klipper3d/klipper/pull/5596

And i seen some other pull request that were merged with errors or not correct implementations either like this:
https://github.com/Klipper3d/klipper/pull/5606

There are more examples i can filter out, but hopefully you get the point. Sometimes it's better to have sth that works, even if there are errors that we can ignore, or not working at all.
The possibility to flash f429 is very important, because the only workable way right now for the normal user, is to make an sdcard with the file, dissasembly the printer to enter the sdcard slot of the octopus, insert it and reset.
Flashing with dfu-util v0.9 is not possible either. Using the boot0 jumper is the same pain as with sdcard.

Cheers & Thx @visilya for this PR, that makes at least my life easier 👍

KevinOConnor:
Thanks.  I'm not sure what this code change is intended to do however.  The existing code should already "reset the cpu" and "jump to memory address at 0x1fff0004".  If there is a particular board that isn't working then I think it will be necessary to find the root cause.

As for dfu-util, it seems to be a separate topic from the code in this PR.

Cheers,
-Kevin

Ramalama2:
> Thanks. I'm not sure what this code change is intended to do however. The existing code should already "reset the cpu" and "jump to memory address at 0x1fff0004". If there is a particular board that isn't working then I think it will be necessary to find the root cause.
> 
> As for dfu-util, it seems to be a separate topic from the code in this PR.
> 
> Cheers, -Kevin

On st32f429 (Octopus Pro), the existing code isn't able to restart into dfu mode.
But i can retest this commit and mainline and post details between mainline and this commit if it helps.
Just need to wire up my printer again.

PR #5704, opened 2022-08-11
Add missing dependency for hid-flash to install script

chestwood96:
Module: INSTALL, added missing dependency

hid-flash requires pkg-config, this does not get installed by the install script. (and produces really misleading errors when missing)

I added it as a dependency which caused it to be installed by the install script and kiauh which allowed hid-flash to work right out of the box.

Signed-off-by: Adrian Joachim [adi.joachim12@gmail.com](mailto:adi.joachim12@gmail.com)

### DIFF
--- a/scripts/install-debian.sh
+++ b/scripts/install-debian.sh
@@ -20,7 +20,7 @@ install_packages()
     PKGLIST="${PKGLIST} avrdude gcc-avr binutils-avr avr-libc"
     # ARM chip installation and building
     PKGLIST="${PKGLIST} stm32flash libnewlib-arm-none-eabi"
-    PKGLIST="${PKGLIST} gcc-arm-none-eabi binutils-arm-none-eabi libusb-1.0"
+    PKGLIST="${PKGLIST} gcc-arm-none-eabi binutils-arm-none-eabi libusb-1.0 pkg-config"
 
     # Update system package info
     report_status "Running apt-get update..."


KevinOConnor:
It looks fine to me, but I'll let @BlackStump add any comments.  Also, why is this added to install-debian.sh but not needed in install-octopi.sh?

-Kevin

BlackStump:
If this it required to get HID-Flash working on Debian then It is all good by me.
 I would think if install-octopi.sh is used on a non RPI it would be required there also.
I am out of the loop somewhat, moving house and now renovating house so 3D print gear is mostly packed away until reno is finished.

chestwood96:
> It looks fine to me, but I'll let @BlackStump add any comments. Also, why is this added to install-debian.sh but not needed in install-octopi.sh?


It may be required in other places (or it may get installed as a dependency of a dependency in those) but I can only really test the debian one without setting up test systems for the other platforms. I honestly don't know enough about the whole install system to comment on the other ones. 

I initially though it was a kiauh issue but was told by the developers that they pull the dependencies from the install-debian script and adding it there made it work.

BlackStump:
I just checked the HID-Flash after running the updated script, no problems to report. It was certainly not working on a Beaglebone Black before.
`debian@beaglebone:~/klipper/lib/hidflash$ make`
`gcc -c -Wall -std=gnu99 -I . `pkg-config libusb-1.0 --cflags` main.c -o main.o`
`gcc -c -Wall -std=gnu99 -I . `pkg-config libusb-1.0 --cflags` hid-libusb.c -o hi `
`d-libusb.o`
`gcc -c -Wall -std=gnu99 -I . `pkg-config libusb-1.0 --cflags` rs232.c -o rs232.o`
`gcc  -no-pie main.o hid-libusb.o rs232.o `pkg-config libusb-1.0 --libs` -lrt -lp`
`thread -o hid-flash`

BlackStump:
If install-octopi.sh is run on a Raspberry Pi then there is no issues with HID-Flash if using another SBC then the appropriate script should be used for that distro.

KevinOConnor:
Thanks.

-Kevin

PR #5571, opened 2022-06-13
Add Ender 3 S1 Plus config & Fix max temp for S1

Fisheiyy:
Signed-off-by: Rob Casper [myallneedsemail@gmail.com](mailto:myallneedsemail@gmail.com)

The [Ender 3 S1 Plus](https://www.creality3dofficial.com/products/ender-3s1-plus-3d-printer-us-direct) has been released although I have yet to see any reviews over it or much talk but thought I might as well get the config added for when people do start to review it and talk about it (hopefully soon), I also fixed the max temperature for the hotend and bed for the S1 non-pro based off the [official diagram](https://www.creality3dofficial.com/files/goods/20220531/s1vsprovsPlus11.jpg) on the Creality website.

PR #5705, opened 2022-08-11
Add Ender 3 S1 Plus

Fisheiyy:
Signed-off-by: Rob Casper [myallneedsemail@gmail.com](mailto:myallneedsemail@gmail.com)

Follow up PR to #5571


KevinOConnor:
Thanks.  The changes to the s1 config look fine to me.

I'm not sure about the addition of the "s1 plus" config.  Please separate out the addition of s1-plus to a separate PR.  I'd like to understand how much it differs from the existing "s1" config.  Also, please confirm that config has been tested on the actual hardware, and please confirm that printer has the same issue with variable f103/f401 control boards.

-Kevin

antiLegion:
I'm willing to test if needed. I have an Ender 3 S1 Plus.
My git skills are only beginner/intermediate though.

I attempted to flash the firmware yesterday but got stuck in boot/could never connect klipper to fluidd.
Had to go back to stock -- this could totally be user error though. I'm a veteran flasher of other firmwares, won't take me long to pick up. I am attempting a second time right now just to make sure I didn't bork something due to user error.

Fisheiyy:
> Thanks. The changes to the s1 config look fine to me.
> 
> I'm not sure about the addition of the "s1 plus" config. Please separate out the addition of s1-plus to a separate PR. I'd like to understand how much it differs from the existing "s1" config. Also, please confirm that config has been tested on the actual hardware, and please confirm that printer has the same issue with variable f103/f401 control boards.
> 
> -Kevin

They are all under the S1 family so they all share the same general structure, which would include motherboard(s) and related variable MCU problems, the S1 Plus has a much larger print volume, I would argue its a change more worthy of its own config rather than the S1 Pro since that was just a maximum temperature change.

Fisheiyy:
> I'm willing to test if needed. I have an Ender 3 S1 Plus. My git skills are only beginner/intermediate though.
> 
> I attempted to flash the firmware yesterday but got stuck in boot/could never connect klipper to fluidd. Had to go back to stock -- this could totally be user error though. I'm a veteran flasher of other firmwares, won't take me long to pick up. I am attempting a second time right now just to make sure I didn't bork something due to user error.

Check your MCU and flash accordingly, if you are trying to flash F103 firmware on a F401 it wont work.

KevinOConnor:
Thanks.  I have no strong preference for adding s1plus config.  Please do separate it out to a separate PR though, and I need to see confirmation that it has been tested and is working on the actual hardware.

-Kevin

antiLegion:
I have successfully installed this on my Ender 3 S1 Plus.
It works.
I have only noted one odd issue.
When auto bed leveling, upon homing, and after beginning the 12 point leveling procedure, at point 4 the print head advances slightly too far and then violently corrects itself. Meaning that it adjust back the 1/2mm that it went too far, but it does so swiftly enough to cause what sounds like the head hitting the end of the track.

It only does this the first time after homing the axii.
I know this is probably difficult to understand from my description. Or maybe it is intended behavior. I am after all very new to 3D printing and a GitHub noob.

It did not do this on the stock firmware. Would a video help?

Fisheiyy:
> antiLegion

A video would help, its 16 point levelling and I dont know anything about going too far and then correcting itself.

Fisheiyy:
@KevinOConnor #5771 Now contains the fixed to the S1 Config and this is PR only adds the S1 Plus Config now.

antiLegion:
Just tested the latest config.
Homing then calibrating no longer causes a bang/crash.
Looks good! Great work y'all!

Fisheiyy:
@KevinOConnor Ready to merge afaik

KevinOConnor:
Thanks.

-Kevin

PR #5707, opened 2022-08-12
Update CANBUS.md

saft6luck:
Add useful information about needed HW and CAN termination.
Note: For the length of stub lines depends on CAN bus speed and the real CAN specification might be too complex. As for a 3D printer the size is not very relevant, I would go for the "longest distance", which is called the CAN bus and should get the termination.


zellneralex (commented):
You need a sign-off line.

The rest looks ok to me.

PR #5709, opened 2022-08-14
Update printer-creality-ender3-s1-2021.cfg

taz420nj:
Clarified firmware build instructions for this printer to avoid 'no mcu communication' issue after flash due to confusing wording, added important detail for F401 variant, added option for physical Z endstop switch.


Fisheiyy:
This is just a bunch of unneeded comment changes, yes it ships with an endstop but it is not installed stock from the factory and most people wont need it anyway since they have the probe z home, but its just not a necessary addition. and I have handled the F401 variant specific things in my PR #5705

Fisheiyy:
This can be closed

PR #5711, opened 2022-08-17
Fix the STM32H743 mcu temp on SKR-3

adelyser:
module: stm32h7 ADC
Fix mcu temperature on SKR-3

Signed-off-by: Aaron DeLyser [bluwolf@gmail.com](mailto:bluwolf@gmail.com)

### DIFF
--- a/src/stm32/stm32h7_adc.c
+++ b/src/stm32/stm32h7_adc.c
@@ -190,7 +190,7 @@ gpio_adc_setup(uint32_t pin)
     }
 
     if (pin == ADC_TEMPERATURE_PIN) {
-        ADC3_COMMON->CCR = ADC_CCR_TSEN;
+        ADC3_COMMON->CCR |= ADC_CCR_TSEN;
     } else {
         gpio_peripheral(pin, GPIO_ANALOG, 0);
     }


KevinOConnor:
Thanks.

-Kevin

nadavmatalon:
Although the above actions list seem to indicate this fix has been merged into the master branch (specifically: "KevinOConnor merged commit [9e4994c] into Klipper3d:master [on Aug 19]", this doesn't seem to be the case.

The line change to "ADC3_COMMON->CCR |= ADC_CCR_TSEN;" - which indeed fixes the issue - isn't present in the master branch code. 

Instead the master branch code contains this line: "ADC_TS->CCR |= ADC_CCR_TSEN;" which doesn't work and raises the error.

I verified this is the case by manually changing the line and recompiling the bin file for the device.

Could this fix be incorporated into the master branch code?

adelyser:
> Although the above actions list seem to indicate this fix has been merged into the master branch (specifically: "KevinOConnor merged commit [[9e4994c](https://github.com/Klipper3d/klipper/commit/9e4994cbdb5a3b1017a0dcca9808efde0de153d4)] into Klipper3d:master [on Aug 19]", this doesn't seem to be the case.
> 
> The line change to "ADC3_COMMON->CCR |= ADC_CCR_TSEN;" - which indeed fixes the issue - isn't present in the master branch code.
> 
> Instead the master branch code contains this line: "ADC_TS->CCR |= ADC_CCR_TSEN;" which doesn't work and raises the error.
> 
> I verified this is the case by manually changing the line and recompiling the bin file for the device.
> 
> Would it be possible to verify that this fix is incorporated into the master branch code?

My PR was merged. The issue is #5744 changed the line to support the stm32l412. I have added a comment about that, I’ll have to test to make sure other things are also not broken.

nadavmatalon:
Thanks for the quick reply! 

The addition of stm32l412 support into a source file specifically designed to address the adc function on  the stm32h7 chip is rather baffling, but I guess there are other considerations affecting this. 

Regardless, happy to hear you'll be looking into this.

PR #5712, opened 2022-08-17
temperature_sensors: add ATC Semitec 104NT-4-R025H43G thermistor

bjj:
I believe this has been substituted into "E3D V6 all-metal hotends"
based on measurements with a thermocouple and practical experiments.

The beta is significantly different from the -42G resulting in a
10-15C offset at printing temperatures.

Signed-off-by: Ben Jackson <ben@ben.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -712,9 +712,10 @@ heater_pin:
 #   periods) to the heater. The default is 1.0.
 sensor_type:
 #   Type of sensor - common thermistors are "EPCOS 100K B57560G104F",
-#   "ATC Semitec 104GT-2", "ATC Semitec 104NT-4-R025H42G", "Generic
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  Adding a new thermistor is fine.  I'm not sure what "I believe this has been substituted" means though - e3d is pretty good with docs - I'm sure they would say if they use a new thermistor.

-Kevin

KevinOConnor:
Any updates on this?  I'm fine with adding a new thermistor, but I would like further clarity on the E3D comments.

-Kevin

bjj:
I went through an exchange with E3D support which did not give me any more insight into the situation.

They did state directly that the part they use is 104NT-4-R025H42G.

If you like, I can amend this commit to have a more neutral description for the pull request.

PR #5713, opened 2022-08-17
temperature_mcu: add reference_voltage

ClfStehrer:
This allows for boards that don't run their ADC with 3.3V to output accurate MCU temperature readings.

Signed-off-by: Peter Stehrer <peter.stehrer@cloudflight.io>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2328,6 +2328,8 @@ monitor these temperatures.
 sensor_type: temperature_mcu
 #sensor_mcu: mcu
 #   The micro-controller to read from. The default is "mcu".
+#reference_voltage:
+#   The reference voltage for the ADC of the mcu. Default is 3.3
 #sensor_temperature1:
 #sensor_adc1:
 #   Specify the above two parameters (a temperature in Celsius and an
--- a/klippy/extras/temperature_mcu.py
+++ b/klippy/extras/temperature_mcu.py
@@ -123,10 +123,10 @@ class PrinterTemperatureMCU:
         hot_temp = ((cal1 >> 12) & 0xff) + ((cal1 >> 20) & 0xf) / 10.
         room_1v = get1v((cal1 >> 24) & 0xff)
         hot_1v = get1v((cal2 >> 0) & 0xff)
-        room_adc = ((cal2 >> 8) & 0xfff)
-        * room_1v / (self.reference_voltage * 4095.)
-        hot_adc = ((cal2 >> 20) & 0xfff)
-        * hot_1v / (self.reference_voltage * 4095.)
+        room_adc = (
+            (cal2 >> 8) & 0xfff) * room_1v / (self.reference_voltage * 4095.)
+        hot_adc = (
+            (cal2 >> 20) & 0xfff) * hot_1v / (self.reference_voltage * 4095.)
         self.slope = (hot_temp - room_temp) / (hot_adc - room_adc)
         self.base_temperature = self.calc_base(room_temp, room_adc)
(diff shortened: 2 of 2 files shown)


KevinOConnor:
Thanks.  I'm not sure this would meet our "100 users" threshold though - https://www.klipper3d.org/CONTRIBUTING.html .  What board did you need this on, and what was the voltage that you needed?

-Kevin

Fisheiyy:
> Thanks. I'm not sure this would meet our "100 users" threshold though - https://www.klipper3d.org/CONTRIBUTING.html . What board did you need this on, and what was the voltage that you needed?
> 
> -Kevin

It may not meet the 100 users threshold, but its probably still good to add it, for the boards that dont run ADC on 3.3V and end up becoming mainstream or in a printer that is very popular.

ClfStehrer:
After a quick chat in discord, as requested here's some more information:

It's a DIY board using a RP2040 as MCU with the ADC at 3V.
As for the style-issues, I'll handle those :)

KevinOConnor:
Thanks for the info.  Out of curiosity, does the current Klipper mcu code run unmodified on this DIY board?

I regularly get feedback that the Klipper config is overwhelming to users.  As such, I'm leery of adding new config options.  In general, I prefer to see a common "real world" usage pattern in place before adding a new user facing option.

That said, I'm fine with adding a `self.reference_voltage` internal variable to the temprature_mcu code, as that is a general code improvement.  (And just assign it as `self.reference_voltage = 3.3` until we observe more use cases.)

Another thought (which I don't necessarily recommend) would be to refactor the existing `[adc_temperature] adc_voltage` config to a more general `[mcu] adc_voltage` config option.

-Kevin

ClfStehrer:
Yep the current klipper MCU code does run & without any issues (as the MCU is just a RP2040) - the only issue is that the temperature value returned is ~50° when the MCU is at 30°. (

I can understand not wanting to add more configuration options, however the MCU temperature is not a feature that is used by most users - and even if, as the self.reference_voltage has a default value of 3.3V it's not something a user would have to configure.

I'm fine with not exposing the reference_voltage for now, since for validation of that board only a very minor change in `temperature_mcu.py`.

Just beware that I'll be back when that DIY board is available more widely to people :)

KevinOConnor:
Thanks.  If  you want to update this PR to make the code changes without the config entry, then I'll commit it.  We can look at exporting the config when more devices are ready.

-Kevin

mattthebaker:
I ran into this when testing the mcu temp sensor on my stm32l4 work in progress -- I have used 3.0v regulators on some of my projects, because they can run nicely off of 3.3v or 5.0v sources without being in "passthrough" conduction at 3.3v.

I realize I am probably late to the game to comment on this, but most mcu's have a built in voltage reference and ability to measure their supply voltage similar to how the mcu temperature endpoints functions. I would have maybe gone for having the code monitor the devices supply voltage and applying measured values to the temp correction code, instead of a config option.

During the chip shortage I sometimes just used 3.3v and 3.0v regulators interchangeably depending on availability.

tehmaze:
> Thanks. I'm not sure this would meet our "100 users" threshold though - https://www.klipper3d.org/CONTRIBUTING.html . What board did you need this on, and what was the voltage that you needed?

Also working on an integration with a Maker Nano RP2040 board, that has a 3V ADC reference. I guess the paradox here is that while it's not meeting the "100 users"-threshold, there may be more users of this flag if it would allow them to use different model (RP2040 based) boards. Not having the ADC reference configurable means, for these users, they have to either buy different hardware or look for a different way to control their printers.

PR #5714, opened 2022-08-18
stm32: Add Hardware PWM support for STM32G0 processors

bjj:
Uses existing common code for STM32. Adds a table for device-
specific PWM mappings. Adds support for enabling all TIM timer
devices. Makes it a runtime error to enable devices the code
doesn't know how to enable.

I have verified performance of the fan pins (PC6, PC7, PB15)
on the SKR Mini E3 V3.

Signed-off-by: Ben Jackson <ben@ben.com>


bjj:
There is a many-to-many relationship that is not explicitly handled by this code (or the existing code): Many pins can be driven by more than one PWM output, and most PWM outputs can be driven on more than one pin. (Even more if the "N" outputs were used). As it is, the code knows one PWM driver for each pin. If that conflicts due to the PWM being in-use on another pin, it doesn't notice. It also doesn't know the other PWM sources for that pin to choose an alternative.

I assume in most cases, very few pins need PWM support in Klipper so this does not come up.

While testing, I observed (via a meter with a Hz setting) that both hardware and software PWM went twice as fast as I expected (e.g. given `cycle_time: 0.01` I saw 200Hz PWM). I need to get out my scope and double check this result.

bjj:
fixes https://github.com/Klipper3d/klipper/issues/5655 and unfortunately partially duplicates https://github.com/Klipper3d/klipper/pull/5657

KevinOConnor:
Thanks.  In general, it looks fine to me.  The one part I'm not sure about is the call to `shutdown()` in `lookup_clock_line()`.  The `lookup_clock_line()` code is generally called from the early boot process which is invoked before it is valid to call `shutdown()`.  (Specifically, calling `shutdown()` before `sched_main()` would result is a `longjmp()` call with uninitialized state.)

Probably best to focus this PR on the hard pwm changes and defer structuralal change to `lookup_clock_line()`.

Cheers,
-Kevin

bjj:
Removed the `shutdown()` on unknown peripheral and just return a `.bit=0` struct which will do nothing in the caller. Possibly the undefined behavior invoked by over-shifting did the same thing in the original code!

KevinOConnor:
Thanks.

-Kevin

PR #5715, opened 2022-08-18
scripts: Support BigTreeTech Octopus boards in flash-sdcard

dewhisna:
Since SDIO is effectively a superset of SPI, with respect to SD Cards, it's possible to use SPI Flash to still flash boards that normally use SDIO.

Unfortunately, the hardware SPI peripheral pin assignments don't usually line up to allow hardware SPI on these (such as is the case with the STM32F407, STM32F429, and STM32F446 MCUs on the BTT Octopus boards), but Software SPI is still an option.

I have personally tried this on my Octopus v1.1 board with the STM32F446 MCU and it worked with no issues.  According to BTT schematics, documentation, and information available, all of the current Octopus and Octopus Pro variants with the F407, F429, and F446 MCUs use the same SDIO pinout configuration.

Therefore, this PR defines the pins for the Octopus with those three MCUs and aliases the names for the various board configurations available for purchase on the BIQU website.

### DIFF
--- a/docs/SDCard_Updates.md
+++ b/docs/SDCard_Updates.md
@@ -89,7 +89,8 @@ This procedure is automated during the upload process.
   the current version.
 - Only boards that use SPI for SD Card communication are supported.
   Boards that use SDIO, such as the Flymaker Flyboard and MKS Robin Nano
-  V1/V2, will not work.
+  V1/V2, will not work in SDIO mode.  However, it's usually possible to
+  flash such boards using Software SPI mode instead.
 
 ## Board Definitions
 
@@ -127,8 +128,9 @@ and the following additional field should be specified:
   the SD Card in the format of `miso,mosi,sclk`.
 
 It should be exceedingly rare that Software SPI is necessary, typically only
-boards with design errors will require it. The `btt-skr-pro` board definition
-provides an example.
+boards with design errors or boards that normally only support SDIO mode for
+their SD Card will require it. The `btt-skr-pro` board definition provides an
+example.
 
 Prior to creating a new board definition one should check to see if an
 existing board definition meets the criteria necessary for the new board.
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -44,6 +44,24 @@ BOARD_DEFS = {
         "firmware_path": "Robin_e3.bin",
         "current_firmware_path": "Robin_e3.cur"
     },
+    'btt-octopus-f407-v1': {
+        'mcu': "stm32f407xx",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        "cs_pin": "PC11"
+    },
+    'btt-octopus-f429-v1': {
+        'mcu': "stm32f429xx",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        "cs_pin": "PC11"
+    },
+    'btt-octopus-f446-v1': {
+        'mcu': "stm32f446xx",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        "cs_pin": "PC11"
+    },
     'btt-skr-pro': {
         'mcu': "stm32f407xx",
         'spi_bus': "swspi",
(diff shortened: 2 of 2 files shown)


KevinOConnor:
Thanks.  However, as I understand it, this has been raised a few times before and there is concern that the current flashing tools will not work correctly on the SDIO boards.  Specifically, it was found that these boards require a hard power reset after a flash in order to reset the sdcard.  Due to this, the consensus was to not list support for these boards.  (I don't own these boards and do not have an opinion on this topic.)

Cheers,
-Kevin

dewhisna:
Hmm..  I don't know.  It must be a per-board thing, which is why I only did this for the Octopus that I am actually using on this printer and didn't try to be an overachiever and do it for all SDIO board.  I need to be able to update it this way since it's a pain to tear the entire printer apart to put it in DFU mode anytime I want to do a firmware update.

Eventually, I will be doing something similar for the new SKR 3 board too, as I'm in the process of setting that one up on another printer, and it also uses SDIO.  I figured I'd address that one only after it's complete and I've tested it (so I didn't include it either for now).

It does seem to work fine for me.  Here's the log from my reflashing session with it, in case you see something I don't (since this is my first time trying to use the flash-sdcard script):

```
$ ./scripts/flash-sdcard.sh /dev/serial/by-id/usb-Klipper_stm32f446xx_2C0022001950534146373920-if00 btt-octopus-f446-v1.1
Flashing /home/pi/klipper/out/klipper.bin to /dev/serial/by-id/usb-Klipper_stm32f446xx_2C0022001950534146373920-if00
Checking FatFS CFFI Build...
Connecting to MCU...Connected
MCU needs restart: is_config=1, is_shutdown=0
Attempting MCU Reset...Done
Waiting for device to reconnect.....Done
Connecting to MCU...Connected

SD Card Information:
Version: 2.0
SDHC/SDXC: True
Write Protected: False
Sectors: 31116288
manufacturer_id: 3
oem_id: SD
product_name: SL16G
product_revision: 8.128
serial_number: 5A650D40
manufacturing_date: 11/2015
capacity: 14.8 GiB
fs_type: FAT32
volume_label: b''
volume_serial: 808739888
Validating Upload...Done
Firmware Upload Complete: firmware.bin, Size: 26548, Checksum (SHA1): CE8323BB39B435533ABE8240327AD9A8FFAE168C
Attempting MCU Reset...Done
Waiting for device to reconnect....Done
Connecting to MCU...Connected
Found and deleted firmware.bin after reset
Firmware Flash Successful
Current Firmware: v0.10.0-546-ga709ba43
Attempting MCU Reset...Done
SD Card Flash Complete
```

Prior to running this, I of course shut down Klipper, and when it was done, I restarted it, and the printer reconnected with Mainsail just fine, meaning it was at least running code again.

I figured there still might be some boards out there with SDIO that this method might not work on and is why I said "it's usually possible to flash such boards using Software SPI mode" in my updates to the docs rather than definitively saying it's possible.

dewhisna:
Though as I'm here looking at this closer, this line in the output `Found and deleted firmware.bin after reset` might be a clue that it does need a hard reset to reflash it, as you were talking about.  Hmm...  So maybe you are correct.

At least this means I can get the file transferred to the SD Card just fine, but I might need to change that script to not delete the file but to tell me to hit the reset button first ... to me, that would be acceptable.  I just don't want to have to turn the printer upside down and remove about 20 screws every time to reflash it.  An SD Card extender might work, but could cause read issues for it.

You guys can decide what you want to do with this PR ... having to hit the reset button manually is still an acceptable solution for me (and probably most users in the same situation).

dewhisna:
Oh shoot... After some experimenting tonight, I get it now...  I had thought perhaps there just needed to be an actual signal on the reset pin to release the GPIO pins from SWSPI so they could get reassigned back to SDIO by the bootloader.  But no, just a hardware reset signal on the main board isn't enough.  The SD Card itself needs to be power cycled so that it can switch modes.

OK... Since I still need to flash my board this way, I think I will just hand-modify my `spi_flash.py` file to stop after transferring the `firmware.bin` file and just exit, skipping the verification steps.  My RPi is powered from the printer's power supply, so I have to shut it down too when I power cycle and I'll just have to do all that manually.  But it sure beats having to tear the printer apart each time.

Since that's not a viable solution for this script for most other users, I'll close this PR.  I'll keep my branch with these changes on GitHub in case anyone else stumbles across this discussion and wants to see the pin mapping for the Octopus boards or something.  Sorry for the noise here ...

dewhisna:
@KevinOConnor -- I suppose another possibility, instead of me only butchering my local `spi_flash.py` file and it not benefiting anyone else, is that I could modify it to add another optional per-board setting in `board_defs.py`, something like a `skip_verify` boolean that defaults to `False`.  If it's set to `True` for a board, it will exit after the flashing operation with a note to the user that they must manually reset their board and that if their bootloader uses SDIO mode they will also have to do a full power cycle to complete the flash operation.

It seems more and more boards are using SDIO, with both of the boards I'm using of the Octopus and the new SKR 3 being prime examples, and it would greatly help others more easily update their Klipper firmware.

So, just let me know if you want me to do that.  I can push it as another commit to this PR and reopen this once I've finished testing it.

PR #5718, opened 2022-08-19
No more whitespace/80+ character errors

Fisheiyy:
Remove trailing whitespace and 80+ character line check errors


Fisheiyy:
not sure why it put this pr under klipper not dangerklipper

PR #5724, opened 2022-08-22
klippy/extras: Ignore non-standard klippy modules

whi-tw:
# What

* Add a .gitignore to `klippy/extras` to ignore all python files except specific exceptions.
* Add a script to make adding new files to the exclusion list easier.

# Why

When adding modules locally (ie. z_calibration, ercf etc.), the local clone of klipper becomes dirty. This makes it somewhat harder for update managers to determine the current state of the repository.

Ideally, custom modules could be loaded from a separate directory to avoid this in future, but for now this will keep the working copy 'clean'.

In the case of a new module being added with the same name as an existing custom one, it will still blow up in the same way as before.


Sineos:
From a support POV, I strongly vote against this idea. The dirty information is very valuable to determin if the installtion is modified.

Arksine:
Untracked files shouldn't make a repo report as dirty, only modification of tracked files.  There are two reasons we want to make sure that the code is pristine.
1) As Sineos mentioned, code modification makes support impossible.  If a user has an issue the first response recommendation will be to run pristine code.
2) Dirty repos can result in merge conflicts.  It isn't possible for Moonraker or other update methods to reliably automate conflict resolution.

Piezoid:
@dans98 [suggested allowing the use of an out of source extras driectory](https://klipper.discourse.group/t/adding-support-for-a-users-extras-directory/3175).

For support, the logs could contain a "tainted" message with the list of loaded unofficial modules and their hashes.

You can't control what users are doing with their installation. They will use extra modules either because they want to or need to. I think that making this process more standardized on way or another will save headaches for both the users and the support team.

PR #5725, opened 2022-08-22
scripts: Support SDIO in flash-sdcard using split verification step and add BTT Octopus

dewhisna:
This PR replaces #5715 (I didn't realize you had to reopen a PR on GitHub **before** force pushing updates to the branch or it won't let you reopen it).

This changes the `flash-sdcard.sh`/`spi_flash.py` scripts to allow for a split flash/verification session as required to program boards that use SDIO for their SD Cards in order to perform a manual reboot/power-cycle.

It also adds support for the BigTreeTech Octopus boards, which use SDIO.  Here's what this new SDIO procedure looks like from a user-perspective, where I reflashed my Octopus board from `ga709ba43` to `g9e4994cb` while testing these changes:

First, I ran the verify step to confirm it worked independently and verify the device matched my existing `ga709ba43` build:

```
$ ./scripts/flash-sdcard.sh -c /dev/ttyACM0 btt-octopus-f446-v1.1
Flashing /home/pi/klipper/out/klipper.bin to /dev/ttyACM0
Checking FatFS CFFI Build...
Connecting to MCU...Connected
MCU needs restart: is_config=1, is_shutdown=0
Attempting MCU Reset...Done
Waiting for device to reconnect....Done
Connecting to MCU...Connected
Firmware Flash Successful
Current Firmware: v0.10.0-546-ga709ba43
Attempting MCU Reset...Done
SD Card Flash Complete

```

I then updated and reflashed it:

```
$ make clean
$ git pull
$ make
$ ./scripts/flash-sdcard.sh /dev/ttyACM0 btt-octopus-f446-v1.1
Flashing /home/pi/klipper/out/klipper.bin to /dev/ttyACM0
Checking FatFS CFFI Build...
Connecting to MCU...Connected
MCU needs restart: is_config=1, is_shutdown=0
Attempting MCU Reset...Done
Waiting for device to reconnect....Done
Connecting to MCU...Connected

SD Card Information:
Version: 2.0
SDHC/SDXC: True
Write Protected: False
Sectors: 31116288
manufacturer_id: 3
oem_id: SD
product_name: SL16G
product_revision: 8.128
serial_number: 5A650D40
manufacturing_date: 11/2015
capacity: 14.8 GiB
fs_type: FAT32
volume_label: b''
volume_serial: 808739888
Validating Upload...Done
Firmware Upload Complete: firmware.bin, Size: 26548, Checksum (SHA1): 224A11A4B6492EA0B867F8C9F8448FB47362FF73
Attempting MCU Reset...Done
Waiting for device to reconnect.....Done
Connecting to MCU...Connected
This board requires a manual reboot to complete the flash process.
If the board's bootloader uses SDIO mode for its SDCard, then a full
power cycle is required.  Please perform the power cycle now and then
rerun this utility with the 'check' option to verify flash.
Attempting MCU Reset...Done
SD Card Flash Complete
```

Then after I power-cycled my board and RPi and resumed with the verification step:

```
$ ./scripts/flash-sdcard.sh -c /dev/ttyACM0 btt-octopus-f446-v1.1
Flashing /home/pi/klipper/out/klipper.bin to /dev/ttyACM0
Checking FatFS CFFI Build...
Connecting to MCU...Connected
MCU needs restart: is_config=1, is_shutdown=0
Attempting MCU Reset...Done
Waiting for device to reconnect.....Done
Connecting to MCU...Connected
Firmware Flash Successful
Current Firmware: v0.10.0-554-g9e4994cb
Attempting MCU Reset...Done
SD Card Flash Complete
```

Note, as illustrated here during my testing, the new check/verify operation can also be used to verify if the device matches the build folder.  That can be used on all boards, not just those that require the split or that use SDIO.

I used `c` and `check` for the new command-line option instead of `v` and `verify` to avoid a conflict with `verbose` that was already defined.

### DIFF
--- a/scripts/spi_flash/spi_flash.py
+++ b/scripts/spi_flash/spi_flash.py
@@ -785,6 +785,13 @@ class SDCardSPI:
             if err_msgs:
                 raise OSError("\n".join(err_msgs))
 
+SDIO_WARNING = """
+This board requires a manual reboot to complete the flash process.
+If the board's bootloader uses SDIO mode for its SDCard, then a full
+power cycle is required.  Please perform the power cycle now and then
+rerun this utility with the 'check' option to verify flash.
+"""
+
 class MCUConnection:
     def __init__(self, k_reactor, device, baud, board_cfg):
         self.reactor = k_reactor
@@ -989,6 +996,9 @@ class MCUConnection:
         return sd_chksm
 
     def verify_flash(self, req_chksm, old_dictionary, req_dictionary):
+        if bool(self.board_config.get('skip_verify', False)):
+            output_line(SDIO_WARNING)
+            return
         output("Verifying Flash...")
         validation_passed = False
         msgparser = self._serial.get_msgparser()
@@ -1063,6 +1073,7 @@ class SPIFlash:
         self.firmware_checksum = None
         self.task_complete = False
         self.need_upload = True
+        self.need_verify = True
         self.old_dictionary = None
         self.new_dictionary = None
         if args['klipper_dict_path'] is not None:
@@ -1092,7 +1103,7 @@ class SPIFlash:
                 raise SPIFlashError("Unable to reconnect")
         output_line("Done")
 
-    def run_reset(self, eventtime):
+    def run_reset_upload(self, eventtime):
         # Reset MCU to default state if necessary
         self.mcu_conn.connect()
         if self.mcu_conn.check_need_restart():
@@ -1101,6 +1112,16 @@ class SPIFlash:
         else:
             self.need_upload = False
             self.run_sdcard_upload(eventtime)
+
+    def run_reset_verify(self, eventtime):
+        # Reset MCU to default state if necessary
+        self.mcu_conn.connect()
+        if self.mcu_conn.check_need_restart():
+            self.mcu_conn.reset()
+            self.task_complete = True
+        else:
+            self.need_verify = False
+            self.run_verify(eventtime)
 
     def run_sdcard_upload(self, eventtime):
         # Reconnect and upload
@@ -1121,7 +1142,8 @@ class SPIFlash:
 
     def run_verify(self, eventtime):
         # Reconnect and verify
-        self.mcu_conn.connect()
+        if not self.mcu_conn.connected:
+            self.mcu_conn.connect()
         self.mcu_conn.configure_mcu()
         self.mcu_conn.verify_flash(self.firmware_checksum, self.old_dictionary,
                                    self.new_dictionary)
@@ -1148,12 +1170,18 @@ class SPIFlash:
             self.mcu_conn = k_reactor = None
 
     def run(self):
-        self.run_reactor_task(self.run_reset)
-        self._wait_for_reconnect()
-        if self.need_upload:
-            self.run_reactor_task(self.run_sdcard_upload)
+        if not bool(self.board_config.get('verify_only', False)):
+            self.run_reactor_task(self.run_reset_upload)
             self._wait_for_reconnect()
-        self.run_reactor_task(self.run_verify)
+            if self.need_upload:
+                self.run_reactor_task(self.run_sdcard_upload)
+                self._wait_for_reconnect()
+            self.run_reactor_task(self.run_verify)
+        else:
+            self.run_reactor_task(self.run_reset_verify)
+            if self.need_verify:
+                self._wait_for_reconnect()
+                self.run_reactor_task(self.run_verify)
 
 def main():
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Thanks.  I don't have an opinion on this topic.  Maybe @Arksine has some comments?

-Kevin

Arksine:
It seems like a reasonable compromise to allow uploads for SDIO based cards and present a message that they require a hard reset.   I'm not sure about the `--check` option.  Given that the flash isn't fully automated, it probably isn't necessary.  As far as I can tell it isn't actually performing a meaningful check, as no hash is passed to the `run_verify` method and it isn't comparing dictionaries either.  Its basically comparing values initialized to `None` and reporting success.

dewhisna:
@Arksine -- I suppose you are correct about the SHA1 hash on the verify.  But, unless I'm mistaken, even on the original single-step process, that SHA1 hash is only done on the upload, and that's still being done here on the upload step (see my comments with the PR and the test run output) -- it checks the hash there to make sure it made it to the SD Card correctly.

The verify step is more of checking to make sure that the version you have flashed matches the version in your build folder to make sure that the board really did flash it.

Here's where I did a test run of the verify step with an intentional version mismatch:

```
Attempting MCU Reset...Done
Waiting for device to reconnect.....Done
Connecting to MCU...Connected
Verifying Flash...
SD Card Flash Error: Version Mismatch: Got 'v0.10.0-546-ga709ba43...', expected 'v0.10.0-554-g9e4994cb...'
```

So, that's why I elected to create the separate `--check` option -- so that you can easily make sure the version matches.  Though I suppose you could just as easily launch the browser and read the version number from the MCU info and compare that to your git cache.  But I thought having a nearly identical process for the SDIO boards would be preferred with the only caveat being that it requires a two-step process with a reboot in between.

scripts/spi_flash/spi_flash.py
```diff
@@ -989,6 +989,16 @@ def sdcard_upload(self):
...
     def verify_flash(self, req_chksm, old_dictionary, req_dictionary):
+        if bool(self.board_config.get('skip_verify', False)):
+            output("This board requires a manual reboot to")
+            output_line(" complete the flash process.")
+            output("If the board's bootloader uses SDIO mode")
+            output_line(" for its SDCard, then a full")
+            output("power cycle is required.  Please perform")
+            output_line(" the power cycle now and then")
+            output("rerun this utility with the 'check' option")
+            output_line(" to verify flash.")
```
Arksine, line 1000: It might be more clear to call output_line once here:
```python
output_line(
    "This board requires a manual reboot to complete the flash process.\n"
    "If the board's bootloader uses SDIO mode for its SDCard, then a full\n"
    "power cycle is required.  Please perform  the power cycle now and then\n"
    "rerun this utility with the 'check' option  to verify flash."
)
```

Alternatively you could define the message as a  docstring similar to how Klipper defines its
startup messages in `klippy.py`, in which case it isn't necessary to include newline characters.

Arksine:
Thanks.  I took another look at the code and see how its working now.  I can see some utility in it.  I left a minor comment, everything else looks ok to me.

scripts/spi_flash/spi_flash.py
dewhisna: Yes, it certainly would be more clear.  I originally had it as four calls to `output_line`, one per printed line, but the over restrictive sanity checks yelled at me when I opened the PR because the lines were more than 80 characters long.  So, I broke it into the less readable and maintainable lines that I had.

I can certainly change that to the auto-concatenating strings like you show with the single function call, and see if that passes the 80-character limit.  The lines printed are each less than 80-characters, as I tend to keep output lines under that in scripts for readability in the terminal windows, but with the indentation for Python, they might still exceed the 80-character limit in the check.

KevinOConnor:
```python
SDIO_WARNING = """
This board requires a manual reboot to complete the flash process.
If the board's bootloader uses SDIO mode for its SDCard, then a full
power cycle is required.  Please perform  the power cycle now and then
rerun this utility with the 'check' option  to verify flash.
"""

class MCUConnection:
...
        ...
        output_line(SDIO_WARNING)
```

dewhisna: That works.  Changes have been made and pushed.  It is much easier to read that way.  Though I still think the sanity checks are a bit over aggressive, IMHO.  On the other hand, it would still fit onto punch cards (ah, the good 'ole days of computers when the worst thing you had to worry about was dropping your stack of punch cards and having to spend the rest of the day manually sorting them).

Arksine:
This looks good to me.  FWIW #5759 has some overlap with this PR, although I think that this PR is still likely to be useful even if full SDIO support is added for the F4.

KevinOConnor:
Thanks.

-Kevin

PR #5732, opened 2022-08-25
fan: add power_pin option

dalegaard:
Hi,

This adds a `power_pin` option to fans. When the fan PWM target is set to a non-zero value, the `power_pin` will be pulled high.

The intended use case is for fans with dedicated PWM inputs. Some of these don't support going to 0% (off) but stay on at some low value. To be able to turn fans like these off, the fan ground can be switched normally(via a standard fan port ground-switching FET), while the `pin` can be connected directly to the fan PWM input. Using the dedicated PWM pin is preferable to ground switched PWM as the fan controller can provide much more accurate and linear control over the fan speed.

With the rising popularity of 4028s, this I believe this will soon(if not already) be useful to >100 printers. Many of the 4028s people are using are intended as server fans, where PWMing to zero is not generally desired.

Best regards,
Lasse

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2446,6 +2446,12 @@ pin:
 #   enough for fans below 10000 RPM at 2 PPR. This must be smaller than
 #   30/(tachometer_ppr*rpm), with some margin, where rpm is the
 #   maximum speed (in RPM) of the fan.
+#enable_pin:
+#   Optional pin to enable power to the fan. This can be useful for fans
+#   with dedicated PWM inputs. Some of these fans stay on even at 0% PWM
+#   input. In such a case, the PWM pin can be used normally, and e.g. a
+#   ground-switched FET(standard fan pin) can be used to control power to
+#   the fan.
 ```
 
 ### [heater_fan]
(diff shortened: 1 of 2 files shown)


amp-man:
I also would like to use a 4028 fan as a parts cooling fan, but at the moment there aren't that many models available to buy because of supply constraints and it would be great to open up the amount of fan models that would be suitable to use for parts cooling.

joaobarros:
@dalegaard thank you for this.

Haven't tested yet but regarding userbase, this is important for hundreds of RatRig V-Core owners that recently got 4028s, myself included in several printers.

KevinOConnor:
Thanks.  It looks fine to me.  As a high-level comment, I'm curious about the choice of `power_pin` as the config name.  I wonder if it may cause confusion with other options such as `max_power`.  Would a name like `enable_pin` be an alternative?

-Kevin

dalegaard:
Hi @KevinOConnor 

`enable_pin` works for me. In my head the pin turned on power to the fan, so it was the "power pin" :-) I'll update with new config name and force push.

Best regards,
Lasse

dalegaard:
Done!

Best regards,
Lasse

KevinOConnor:
Thanks.

-Kevin

PR #5734, opened 2022-08-25
extruder_stepper: Support configuring pressure advance from config

dmbutyugin:
Also added status report for extruder_stepper objects with the
current pressure advance parameters values.

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -120,6 +120,13 @@ The following information is available in the
 - `excluded_objects`: An array of strings listing the names of excluded objects.
 - `current_object`: The name of the object currently being printed.
 
+## extruder_stepper
+
+The following information is available for extruder_stepper objects (as well as
+[extruder](Config_Reference.md#extruder) objects):
+- `pressure_advance`: The current [pressure advance](Pressure_Advance.md) value.
+- `smooth_time`: The current pressure advance smooth time.
+
 ## fan
 
 The following information is available in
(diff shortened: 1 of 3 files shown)


dmbutyugin:
Note that `SET_PRESSURE_ADVANCE EXTRUDER=...` command effectively expects the ExtruderStepper object (and Extruder object can act as ExtruderStepper). So, it is expected that the PA is configured on a per-ExtruderStepper basis. However, it is not possible to configure the PA in the `printer.cfg`:
```
[extruder_stepper extruder_s1]
pressure_advance_smooth_time: 0.02
```
^ this does not work ^

This PR fixes this inconsistency and allows a user to configure the PA in the `printer.cfg` (specifically, configuring smooth_time may be very useful for complicated extruder setups, e.g. mixing extruders). It also adds `get_status` method for ExtruderStepper objects such that the PA params can be queried in macros.

KevinOConnor:
Thanks.  The reason I didn't add pressure_advance parameters to extruder_stepper objects is that I figured almost everyone would configure PA via macros anyway.  That said, I don't have an issue with adding the config option (for consistency).

So, if you have a use case for it, then let me know and I will commit this PR.

-Kevin

dmbutyugin:
Thanks, Kevin. It is more about configuring `pressure_advance_smooth_time` from the config rather than `pressure_advance` itself. I see smooth time being more of a property of an extrusion system rather than the filament, so it makes more sense that way. I'd also rather avoid specifying it in every filament profile that I have. Besides, currently the default for the extruder_stepper for smooth_time is 0, which disables PA, and this might be unexpected to the users (if they just run `SET_PRESSURE_ADVANCE EXTRUDER=... ADVANCE=...` without further inspecting its output).

My case is a mixing extruder with two stepper motors pushing two filaments into a common melting heatblock (Cyclops-style). And I found it beneficial to configure the smooth time different from 0.04 (which isn't even the default for `extruder_stepper`).

KevinOConnor:
Okay, thanks.

-Kevin

PR #5735, opened 2022-08-26
dump_mcu: helper script for mcu flash backup

Arksine:
This utility uses logic from the `DUMP` command in `console.py` to read out a section of the MCU's flash and save it to disk.  The motivation behind this was to make it a bit easier for users unfamiliar with `console.py` to backup their stock bootloader.  Usage is similar to console.py:
```
usage: dump_mcu.py [-h] [-b <baud rate>] [-c <canbus iface>]
                   [-i <canbus nodeid>] [-s <read start>] [-l <read length>]
                   <device> <outfile>

MCU Flash Dump Utility

positional arguments:
  <device>              Device Serial Port
  <outfile>             Path to output file

optional arguments:
  -h, --help            show this help message and exit
  -b <baud rate>, --baud <baud rate>
                        Baud Rate
  -c <canbus iface>, --canbus_iface <canbus iface>
                        Use CAN bus interface; <device> is the chip UUID
  -i <canbus nodeid>, --canbus_nodeid <canbus nodeid>
                        The CAN nodeid to use (default 64)
  -s <read start>, --read_start <read start>
                        Flash address to start reading
  -l <read length>, --read_length <read length>
                        Number of bytes to read
```

```
# serial
~/klippy-env/bin/python ~/klipper/scripts/dump_mcu.py -l 0x4000 /dev/ttyACM0 ~/backup.bin
# can
~/klippy-env/bin/python ~/klipper/scripts/dump_mcu.py -l 0x4000 -c <uuid> ~/backup.bin
```

If you think this is worth adding I can write documentation.  I'm thinking that `bootloaders.md` might be a better place for this than `debugging.md`.

### DIFF
--- a/scripts/dump_mcu.py
+++ b/scripts/dump_mcu.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python
+# MCU flash dump assistant
+#
+# Copyright (C) 2022 Eric Callahan <arksine.code@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+import sys
+import argparse
+import os
+import traceback
+import logging
+KLIPPER_DIR = os.path.abspath(os.path.join(
+    os.path.dirname(__file__), "../"))
+sys.path.append(os.path.join(KLIPPER_DIR, "klippy"))
+import reactor
+import serialhdl
+import clocksync
+
+###########################################################
+#
+# Helper methods
+#
+###########################################################
+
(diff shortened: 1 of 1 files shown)


KevinOConnor:

Documentation is always nice.  Might be worth putting in Debugging.md as it is sometimes useful to dump fields other than the bootloader.  I have no strong opinion though.

-Kevin

PR #5736, opened 2022-08-26
Add module for load-cell based probes

Fisheiyy:
Adds support for load-cell based probes.

These probes utilize a load-cell sensor measuring the contact force between hotend and nozzle, the provided algorithm is capable to deal with several issues like baseline wandering and hysteresis effects which are often inherent to such systems.
It provides a truly offset-free measurement of the bed surface without the need to calibrate the probe.


Fisheiyy:
oopss

PR #5737, opened 2022-08-26
Add SET_DIGIPOT command to mcp4018 implementation

jake-b:
Added a `SET_DIGIPOT` command to the mcp4018 implementation.

Previously the mcp4018 was read only, and set at the time of configuration.  This allows you to change the value during a print, which is needed for some older printers that need to lower the stepper current during preheating. (smaller power supplies needed all the power to heat the bed, so the steppers are lowered to just enough to hold position while the bed comes up to temp, and then raised for printing.)

By way of testing, I printed an entire Voron 2.4 on a heavily modified Makerbot Replicator 2X with a Mightyboard Rev G that uses the mcp4018 to control stepper current.

Documentation updates are included, along with instructions on how to implement a `G130` macro, for compatibility with the start g-code that is commonly used with vintage printers that use mcp4018 to adjust their vRef voltages.

Thanks for your consideration.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -746,6 +746,54 @@ scheduled to run after the stepper move completes, however if a manual
 stepper move uses SYNC=0 then future G-Code movement commands may run
 in parallel with the stepper movement.
 
+### [mcp4018]
+
+The following command is available when a
+[mcp4018 config section](Config_Reference.md#mcp4018) is
+enabled.
+
+#### SET_DIGIPOT
+`SET_DIGIPOT DIGIPOT=config_name WIPER=<value>`: This command will
+change the current value of the digipot.  This value should typically
+be between 0.0 and 1.0, unless a 'scale' is defined in the config.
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  In general it looks fine to me.  As a minor thing, I'd ask that we not add complex macros to G-Codes.md .  If you feel an example macro is needed then maybe add it to config/sample-macros.cfg (or similar).  Alternatively, if there is a particular Klipper printer config that needs this feature then maybe add the macro to that printer config.

Also, please use the existing `gcmd.getfloat()` range checks instead of implementing them manually (eg, `gcmd.get_float('WIPER', minval=0., maxval=self.scale)`).

-Kevin

jake-b:
Thanks Kevin.  I've moved the sample macro out of `G-Codes.md` and into `sample-macros.cfg`.  A full printer config is probably outside the scope of this pull request for now.  Thanks for showing me the `get_float` range check parameters, also implemented.

KevinOConnor:
Thanks.

-Kevin

PR #5741, opened 2022-08-28
Wrong uart_address assignment for Y and Z - BTT SKR CR6

JamesH1978:
As found on discord by user David Carey.

uart_address for y and z are swapped. Discovered while setting up sensorless homing.

Signed off by: James Hartley <james@hartleyns.com>


KevinOConnor:
Thanks.  @KoeRt888 - do you have any comments on this?

-Kevin

KoeRt888:
That could be. I took the uart addresses from a very early version of the marlin firmware of the BTT repository. Might be that either I mixed it up or it was wrong in the version that I have taken the addresses from.

KevinOConnor:
Thanks.

-Kevin

PR #5733, opened 2022-08-25
Fix i2c on stm32g0

KevinOConnor:
This is an alternate approach to PR #5592 - it fixes i2c on stm32g0, but uses less compile time `#ifdefs`.

@retsamedoc - FYI.

-Kevin

PR #5744, opened 2022-08-29
stm32l4: add stm32l412 support with adc, i2c, spi, usb

mattthebaker:
Add support for stm32l412, with most peripherals functional, including usb clock recovery.


mattthebaker:
I should mention this needs PR #5733 or a similar solution for I2C pin alternate function. STM32L4 uses alternate function index 4 for I2C pins.

KevinOConnor:
Thanks.  In general it looks good to me.  I have some high-level comments:

1. Please update lib/README to reflect the changes made to the lib/ directory.
2. Can you check if maybe src/stm32/stm32h7_adc.c is a better starting point for stm32l4 adc?
3. Can you confirm that the delay and BTABLE assignment is needed on stm32l4?  Seems a bit unusual.
4. Can you confirm that stm32l4 can not reuse the `gpio_peripheral()` function from src/stm32/gpioperiph.c ?
5. In regard to i2c, if stm32l4 needs function 4, why does this patch add them with `GPIO_FUNCTION(1)`?

Separately, I agree we need to do something about the i2c bus names.  Not sure what the solution is.  One possibility would be to use bus names like `i2c1_PA9_PA10` instead of the ambiguous names in use now.

-Kevin

mattthebaker:
Thanks for the feedback. This code was written awhile ago, but sat idle on due to the chip shortage.

1. Done.
2. H7 wasn't around when originally written, but it does look like a better fit. I will rewrite and retest based on the H7 driver.
3. I will retest this to confirm. I cannot remember if it was critical or not, and it was present when I tested after rebasing.
4. Done, reworked to use the general implementation.
5. Fixed, this was an error. The board I have for testing I2C is on another branch due to an unusual clock setup not likely to go on mainline, and I missed a commit when packaging.

I will follow up with a complete candidate once I can rewrite the ADC driver and retest the USB.

The proposed i2c naming scheme would be a definite improvement for the user to configure a bus without referencing the code or mcu DECL constants. It may even be a viable solution to allow mixed and matched bus pin combinations for any board, though there may end up being a large number of permutations.

The primary board this commit is for will need something similar to the proposed i2c naming scheme in order to mix and match spi bus pins, but I wanted to avoid lumping that in. For spi in particular the number of possible permutations is probably too much with the extra pin. I believe the sercom on samd already has a run time pin selection system in place that could be implemented similarly.

mattthebaker:
Following up on points 2 and 3.

2. ADC driver reimplemented based on H7. Due to how that module was written, it required more total line changes than from F0. The implementation has a different `gpio_adc_cancel_sample` then F0 and I am not sure how to force that to happen for code coverage, but I did not find any issues in test.
3. I did not find any unexpected behaviors in test after removing the `BTABLE` write and delay so they are removed.

**It is not quite ready to merge.**

I left the F0 based implementation in the repository, if you have an opinion on which implementation would be better to incorporate, I will roll back the other.

Thanks,
Matt

KevinOConnor:
Thanks.  Seems it would be worthwhile to go with the stm32h7_adc.c work.  FWIW, the `gpio_adc_cancel_sample()` in stm32h7_adc.c should be refactored to look like the code in stm32f0_adc.c.  @adelyser - do you have any comments?

-Kevin

mattthebaker:
Fixed merge conflicts and rolled back F0 based ADC.

Was torn on whether to re-enumerate all the I2C pin configs for L4 or work them in as done. I can rework this if it would be cleaner the other way -- readability wise its nice not to have #defines interleaved, but there are more lines of code to edit/maintain.

KevinOConnor:
Thanks!
-Kevin

adelyser:
The change to the MCU Temperature line in stm32h7_adc.c broke the MCU function on the STM32H7xx boards. I don’t really understand why the STM32L4 was added to the H7 files, it should have its own file (stm32l4_adc.c). I can’t wrap my brain around the idea of hiding the L4 processor code in the H7 files……

Also, with all those changes to the STM32H7 file, nobody has tested anything against an STM32H7XX board.

KevinOConnor:
It is unfortunate that the stm32h7 code is not working correctly.  Have you been able to identify the part of the change that caused the regression?  I do think we want to reuse the code between chip variants, so if we can identify the regression then I think we should fix that.

I understand the confusion with a file name of "stm32h7_adc.c".  To me it does not mean "stm32h7 code for handling adc"; to me it means "code for handling the adc hardware found on at least the stm32h7".  It is important to me that where a hardware vendor has reused hardware between chip variants that we also try to reuse software.  I understand this can introduce frustrating regressions when refactoring introduces a regression, but I believe it is a huge benefit for long term maintenance.  Otherwise, we'd have hundreds of nearly identical code files just for the stm32 line alone, each with their own set of bugs.

-Kevin

adelyser:
> It is unfortunate that the stm32h7 code is not working correctly. Have you been able to identify the part of the change that caused the regression? I do think we want to reuse the code between chip variants, so if we can identify the regression then I think we should fix that.
> 
> I understand the confusion with a file name of "stm32h7_adc.c". To me it does not mean "stm32h7 code for handling adc"; to me it means "code for handling the adc hardware found on at least the stm32h7". It is important to me that where a hardware vendor has reused hardware between chip variants that we also try to reuse software. I understand this can introduce frustrating regressions when refactoring introduces a regression, but I believe it is a huge benefit for long term maintenance. Otherwise, we'd have hundreds of nearly identical code files just for the stm32 line alone, each with their own set of bugs.
> 
> -Kevin

I'm not following your logic, the STM32L4 and STM32H7 are not hardware similar. In fact, most of the stm32h7_adc.c file is switched out for the STM32L4 because of this. What am I missing here?

KevinOConnor:
The adc on the L4 and H7 have identical run-time code - only the initialization and pin configuration is different.  It seems clear to me that stm used similar hardware blocks to implement the adc, and we should therefore try to reuse code between them.

I do agree that the code could be refactored to improve maintainability.  The chip specific initialization could be moved into `stm32h7_init()` and `stm32l4_init()` functions to reduce the number of compile time `ifdefs`.

-Kevin

mattthebaker:
> The change to the MCU Temperature line in stm32h7_adc.c broke the MCU function on the STM32H7xx boards. I don’t really understand why the STM32L4 was added to the H7 files, it should have its own file (stm32l4_adc.c). I can’t wrap my brain around the idea of hiding the L4 processor code in the H7 files……
> 
> Also, with all those changes to the STM32H7 file, nobody has tested anything against an STM32H7XX board.

Hi @adelyser, are you saying here that it has not been tested on an H7, or that it has been tested on an H7 and it does not work?

I reviewed the code again and I don't see anything that looks like it would affect the MCU temperature on the H7, but if its broken I will fix it. Which board is it broken on?

Matt

adelyser:
> > The change to the MCU Temperature line in stm32h7_adc.c broke the MCU function on the STM32H7xx boards. I don’t really understand why the STM32L4 was added to the H7 files, it should have its own file (stm32l4_adc.c). I can’t wrap my brain around the idea of hiding the L4 processor code in the H7 files……
> > Also, with all those changes to the STM32H7 file, nobody has tested anything against an STM32H7XX board.
> 
> Hi @adelyser, are you saying here that it has not been tested on an H7, or that it has been tested on an H7 and it does not work?
> 
> I reviewed the code again and I don't see anything that looks like it would affect the MCU temperature on the H7, but if its broken I will fix it. Which board is it broken on?
> 
> Matt

I'm saying it would be nice to test it on an H7 to ensure it still works.

That said, I did just run the master code (been out of town), and it's working on my devices, so I'm not sure why it wouldn't be working on the report that I received. I will continue to test though.

PR #5746, opened 2022-08-29
samd21: fix erroneous register bit reference.

mattthebaker:

### DIFF
--- a/src/atsamd/i2c.c
+++ b/src/atsamd/i2c.c
@@ -25,8 +25,6 @@ i2c_init(uint32_t bus, SercomI2cm *si)
     si->CTRLA.reg = 0;
     uint32_t areg = (SERCOM_I2CM_CTRLA_LOWTOUTEN
                      | SERCOM_I2CM_CTRLA_INACTOUT(3)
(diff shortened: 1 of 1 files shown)


mattthebaker:
This needs some discussion, the original code was setting bits which are not correct or reserved by referencing the wrong defines. However, updating the code may change bus behavior. I believe the change reflects the original code intention.

KevinOConnor:
Thanks.  Good catch.  I'd say it is probably safer to just remove those two erroneous bits.  It looks like I wrote the original code, but I don't know why I tried to include those two bits.  Since the code seems to be working without setting those bits, probably safer to avoid setting them now.

Maybe @FHeilmann has some comments on this.

-Kevin

FHeilmann:
> Maybe @FHeilmann has some comments on this.

Thanks for the tag but unfortunately I have nothing to add to this.

Fh

mattthebaker:
Agree, that would be the prudent choice. They appear related to automatic bus error or timeout recovery, I did not dig deep and it as awhile ago. Device is working without.

I force pushed with bits removed instead.

Matt

KevinOConnor:
Thanks.

-Kevin

PR #5748, opened 2022-08-31
[WIP] stm32: Support for stm32mp1

jernejp21:
Added support for STM32MP1's Cortex-M core.

I am developing new controller board, and in the future I would like to use STM32MP1 to run both linux (on Cortex A cores) and control motors (on Cortex M core). Currently you need 2 boards to run linux and MCU. With STM32MP1 you have all in one SoC.

Signed-off-by: Jernej Pangerc [jernej.pangerc@tutamail.com](mailto:jernej.pangerc@tutamail.com)


Fisheiyy:
Very cool idea, has CAN support too

KevinOConnor:
Thanks.  Is this code still a "work in progress" or is it ready for widespread usage?

Some high-level questions:
1. I noticed that this is using a custom linker script - can you provide some info on why this is needed on the mp1?  Does it have a different arm vector table / startup system than other stm32 chips?
2. How do the new stm32mp1_serial.c and stm32fp1_adc.c differ from the existing stm32h7_serial.c and stm32h7_adc.c files?

-Kevin

jernejp21:
Current code is ready to use, but there are still some features missing. SPI, I2C, CAN and HW PWM are not supported yet. I would also like to add OpenAMP for communication between cores - creating virtual UART, which replaces USB or UART cable connection.

1. Linker script: this chip doesn't have flash. CA7 cores are booted from SD card (like Raspberry Pi) and CM4 core has only RAM memory. You have to load CM4 code from linux and run it from linux. RETRAM is used for vector table, SRAM1 is used for .text, SRAM2 is used for .data.
I think it would be possible to add RAM regions to existing linker file, but it would need a lof of #ifdef #else

image

2. ADC difference: STM32H7 has 3 ADC channels, STM32MP1 has 2 ADC channels. There are also some registers named differently. *LDORDY registers are missing from CMSIS* - I have added those registers to STM32CubeMP1 and added a pull request to STM32 repo. I haven't done those changes for STM32CubeH7.
Serial difference: it could be done with #ifdef #else, but stm32h7_serial.c would be misleading, if both h7 and mp1 code is in this file. I am still not sure if I should keep UART communication on MP1 or not, since the idea is to use virtual UART.

Should I make more changes (more finished product) and then create a new pull request, or should I keep making changes in this pull request's branch and when done it will be merged?

jernejp21:
I will put code for serial into stm32h7_serial, and will do the same for adc. I will also implement other features - hardware PWM, SPI, I2C, CAN on OpenAMP. For the time being I have marked this PR as WIP.

If pull requests should be only finished code, you can also close this PR and I will try another PR at later time, when the code for STM32MP1 is more mature.

KevinOConnor:
Interesting, thanks.  For what it is worth, I'd try to keep the initial commit to the basic parts (compiling, communication, and gpios).  Then add the more advanced parts afterwards.  That will ease the review process.

As to the linker scripts, if they need to be added that is fine.  I just want to make sure I understand them.  If they do need to be added, they should be in the src/stm32/ directory and not in the src/generic/ directory.  From your description, it's not clear to me why the custom linker scripts are needed though - you can't put all the code and data in a single set of pages?  Perhaps some limitation with the Linux deployer?  Again, it is fine if it is needed, I just want to make sure I understand it.

For communication, just so I understand, you are currently running physical uart wires between host CA7 and CM4 mcu, but plan to replace that with a memory based IPC system?

-Kevin

jernejp21:
Thank you for the comments.

```
RETRAM_interrupts (xrw)  : ORIGIN = 0x00000000,  LENGTH = 0x00000600
SRAM1_text        (xrw)  : ORIGIN = 0x10000000,  LENGTH = 128K
SRAM2_data        (xrw)  : ORIGIN = 0x10020000,  LENGTH = 128K
SRAM3_ipc_shm     (xrw)  : ORIGIN = 0x10040000,  LENGTH = 64K
SRAM4             (xrw)  : ORIGIN = 0x10050000,  LENGTH = 64K
```

Interrupts need to be in `RETRAM_interrupts` and `SRAM3_ipc_shm` is used for OpenAMP. `SRAM1_text` could be renamed to `rom` and `SRAM2_data` could be renamed to `ram`. I will try to to implement this in existing linker file and add #if #else for stm32mp1.

Yes, currently I am using USART pins and USB connector to communicate between CA7 and CM4. The plan is to use IPC for production, but if you want to develop klipper (python part) you can still connect to PC via USART.

I have already fixed serial and ADC (put in stm32h7 files). When I implement changes into linker file, I will make another push.

KevinOConnor:
> I will try to to implement this in existing linker file and add #if #else for stm32mp1.

I don't think that will improve things.  If it needs a separate linker then that is fine.  We do this on other platforms already (like the rp2040).  I just want to make sure I understand the build process.

It does sound like we should wait until IPC based communication is ready before merging this.

Thanks,
-Kevin

PR #5749, opened 2022-09-02
Support requesting bootloader when using UART

KevinOConnor:
This PR adds a "side-channel" mechanism for entering the bootloader when using serial communication. This makes it easier to automate flashing scripts. USB and canbus have similar side-channel mechanisms - this PR extends the concept to UART.

The mechanism to enter the bootloader over serial is to send a special string on the serial port. For example:
`echo -e '~~~ \x1c Request Serial Bootloader!!!~~~' > /dev/ttyAMA0`
The string is specially crafted so that the Klipper message parser will not fragment it.

I have enabled this support on lpc176x and stm32 (as there is UART CanBoot support for these boards).

@Arksine - thoughts?

-Kevin

### DIFF
--- a/src/Kconfig
+++ b/src/Kconfig
@@ -44,6 +44,8 @@ source "src/simulator/Kconfig"
 # Generic configuration options for serial ports
 config SERIAL
     bool
+config SERIAL_BOOTLOADER_SIDECHANNEL
+    bool
 config SERIAL_BAUD
     depends on SERIAL
     int "Baud rate for serial port" if LOW_LEVEL_OPTIONS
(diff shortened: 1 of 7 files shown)


Arksine:
Looks good to me.  Should we create a `flash_uart.py` or similar to perform the request?  I could probably add an option in `flash_can.py` to do it otherwise.

KevinOConnor:
Thanks.  I committed this branch.  I changed the request string slightly - it is now something like:
`echo -e '~~~ \x1c Request Serial Bootloader!! ~~~' > /dev/ttyAMA0`

As for scripts, I'm not sure.  It does seem like something that should be in the Klipper repo.  The Klipper flashing scripts could use a bit of an overhaul though.  (FWIW, ideally, I think there would be just one `flash_helper.py` script that could call all the others as needed.)

-Kevin

PR #5751, opened 2022-09-03
Toolhead: Enforcing max_accel configuration option on M204

compihu:
Enforces max_accel configuration option, to prevent physical damage of the printer (e.g. losing steps on some axes on a delta which can drive the toolhead into the buildplate)

### DIFF
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -211,7 +211,8 @@ class ToolHead:
                                             self._handle_shutdown)
         # Velocity and acceleration control
         self.max_velocity = config.getfloat('max_velocity', above=0.)
-        self.max_accel = config.getfloat('max_accel', above=0.)
+        self.max_accel_config = config.getfloat('max_accel', above=0.)
+        self.max_accel = self.max_accel_config
         self.requested_accel_to_decel = config.getfloat(
             'max_accel_to_decel', self.max_accel * 0.5, above=0.)
         self.max_accel_to_decel = self.requested_accel_to_decel
@@ -592,7 +593,7 @@ class ToolHead:
                                   % (gcmd.get_commandline(),))
                 return
             accel = min(p, t)
-        self.max_accel = accel
+        self.max_accel = min(self.max_accel_config, accel)
         self._calc_junction_deviation()
 
 def add_printer_objects(config):


Sineos:
Thanks. At minimum you need to follow https://www.klipper3d.org/CONTRIBUTING.html, especially the signed-off line

compihu:
> Thanks. At minimum you need to follow https://www.klipper3d.org/CONTRIBUTING.html, especially the signed-off line

Signed-off-by added to the commit message. Let me know please if anything else is necessary.

Sineos:
@compihu One of the core developers need to have a look and review it. All have a busy life so I kindly ask for some patience.

compihu:
> @compihu One of the core developers need to have a look and review it. All have a busy life so I kindly ask for some patience.

Thanks for the info

dmbutyugin:
At least in case of `SET_VELOCITY_LIMIT`, the change to allow it to set acceleration higher than what's in config was intentional. Arguably, `M204` may be different in that regard. But I do not know what would be a reasonable behavior of `M204` with respect to `[printer].max_accel` and acceleration limit set by `SET_VELOCITY_LIMIT`: should `M204` only respect the limit from `max_accel` but not from `SET_VELOCITY_LIMIT` (but then why?), or should it respect both? So, it might be that there is no reasonable all-round option and the current behavior is the best you can do in the general case.

Separately, if users wish to have this behavior for their printer, they can achieve it by overriding `M204` command with a macro. Config values are available to macros, so it is possible to enforce the maximum acceleration set by `M204` (and also update `accel_to_decel`, for instance).

danorder:
I'm not sure if i  fully agree with this. On one hand you could change the current behavior as a safety limit however this could also be a pain. If anything a max accel editing method may be viable similar to marlins option.  Alternatively the printer should have max velocity  set to the max safest value with acceleration to deal with this already. Further in slicers with per filament settings like prusa / super slicer is also nice to be able to override or use generic profiles without making changes to the config or adding macros which simplifies things some what for beginners. Im not sure how many profiles would work properly without the setup defaults. Typically delta can run most profiles provided they are for the correct extruder type.  

My other concern is some of us do use accel / jerk macros already which can scale accel to decel becoming broken.  Another  fear would be a situation of having to use a high acc value similar to 999999 as a workaround to this. Leading to high acc on things like homing , probing , or a possible garbled gcode situation.  I can also see how a typo in the slicer could lead to damages but I can't see this being a issue with max velocity set appropriately for what the machine can do. what we can't control ultimately  is how well or safe users will make configs outside of the github examples. another solution may be a max acc limit for homing specifically / probing as a config option. (ignoring homing override) Im not a huge fan of high acc defaults being in use regarding those moves either. This could also reduce some clutter in configs if a Homing_Max_acceleration option were available.

compihu:
@danorder Pretend that you're a printer manufacturer: Would you like to make sure that all the **limits** you configured are obeyed preventing warranty issues where your customer accidentally made damages by simply misconfiguring a slicer or entering some dumb value on the UI?

dmbutyugin:
My personal stance is only about `SET_VELOCITY_LIMIT`: there are situations when it is very useful to let it set maximum acceleration over the configured limit and this saves a lot of hassle. For example, this is true for both resonance testing methods. I do not have a very strong opinion about `M204` either way, but I just find it difficult to find a different reasonable behavior for it.


> @danorder Pretend that you're a printer manufacturer: Would you like to make sure that all the **limits** you configured are obeyed preventing warranty issues where your customer accidentally made damages by simply misconfiguring a slicer or entering some dumb value on the UI?

The same thing can happen with the `printer.cfg` limits. And here I do speak from my personal experience: at some point I bumped `max_accel` to 10K on Ender 3 for testing, then went on vacation, forgot about it, and then started to get occasional layer shifts on Y. So, it can go both ways.

In general, I do understand the concern, but (a) Klipper and warranty do not play well together and (b) there are many other ways to damage the printer by a wrong configuration. Ideally, you should have per-printer profiles (which, for instance, PrusaSlicer and SuperSlicer support) that you configure once, make sure that they are sane (including velocity and acceleration limits, if you chose to specify them; I personally don't), and then use them for a long time. I totally agree that the fact that such options are exposed in the UIs in such a way that it is easy to make a mistake and enter the wrong value is really bad. However, I do not see this a problem of Klipper, rather, it would be great if UIs (Fluidd/Mainsail) had some safety checks in place (e.g. they can take the original configured value from the config status exported by Klipper and warn a user if it tries to set too large limit). For example, Octoprint immediately indicates if you are trying to enter obscure temperature and Cura also does that if you specify some parameter outside of a reasonable range; I do not see why other software cannot do the same.


> This could also reduce some clutter in configs if a Homing_Max_acceleration option were available.

@danorder, Are there any situations where this can be useful? For example, for X/Y, typically the default acceleration is good enough, and Klipper supports separate Z acceleration for pretty much all kinematics where Z may not perform as well (even on Deltas, where there is no input shaping for Z axis and using the same high acceleration for Z is undesirable).

KevinOConnor:
I don't think we should make this change.  It is a change in behavior that may break existing configurations, and I don't see a notable gain from it.  If the user sends in poor commands then they will get poor quality results.  This isn't a safety issue as far as I'm aware - just a quality issue.

-Kevin

PR #5752, opened 2022-09-03
add function display IP

EchoHeim:
Display the current host ip address on the 12864 for easy viewing

### DIFF
--- a/klippy/extras/CatchIP.py
+++ b/klippy/extras/CatchIP.py
@@ -1,0 +1,19 @@
+# query local ip address
+
+import socket
+import time
+import traceback
+
+def get_host_ip():
+
+    ip='0.0.0.0'
+    try:
+        s=socket.socket(socket.AF_INET,socket.SOCK_DGRAM)
+        s.connect(('8.8.8.8',80))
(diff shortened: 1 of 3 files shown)


Sineos:
Not sure reading the code correctly but it looks like a hard error when the IP could not be obtained. If true then this is not a robust solution. Apart from this I see no reason for my printer to contact Google DNS.

At minimum you need to follow https://www.klipper3d.org/CONTRIBUTING.html, especially the signed-off line

EchoHeim:
The screen shows 0.0.0.0 when the ip cannot be obtained.
Google DNS is only used as a basis to verify whether the network is connected or not, and has no other role.

zellneralex (approved):
Tried it on my machine and looks good to me

jschuh:
This doesn't feel like the right approach. Klipper doesn't interact with the network, so this is just sort of grabbing an arbitrary IPv4 address (which will usually be correct, but on some systems won't).

Unless I'm misunderstanding the goal, it seems like what you really want is for Moonraker to provide Klipper with the set of addresses it's listening on?

PR #5756, opened 2022-09-04
Add a BME280 report time config option to reduce mcu load

Blinker73:
module: BME280, add bme280_report_time option

Having hooked up 2 BME280 chips on two i2c busses of the same RPI, I see a high "MCU RPI Load" utilization
Looking at the code it reported every .8 seconds.
I do not need the sensors to be this precise and tested reporting every 30 seconds instead which brings down the MCU load significantly but with a burst every 30 seconds

This PR is to add an option to specify a custom report time similar to the one in the LM75 temperature sensor
The previous REPORT_TIME if the BME280 was used as a default/minimum value.
This change reads the report time value from a bme280_report_time option in the config files

Signed-off-by: Sylvain Dansereau brutus_dansereau@hotmail.com

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2162,6 +2162,9 @@ sensor_type: BME280
 #i2c_speed:
 #   See the "common I2C settings" section for a description of the
 #   above parameters.
+#bme280_report_time:
+#   Interval in seconds between readings. Default is 0.8, with minimum
+#   0.8.
 ```
 
 ### HTU21D sensor
--- a/klippy/extras/bme280.py
+++ b/klippy/extras/bme280.py
@@ -6,7 +6,7 @@
 import logging
 from . import bus
 
-REPORT_TIME = .8
+BME280_MIN_REPORT_TIME = .8
 BME280_CHIP_ADDR = 0x76
 BME280_REGS = {
     'RESET': 0xE0, 'CTRL_HUM': 0xF2,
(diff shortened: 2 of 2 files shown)


Blinker73 (commented):
I reviewed this change according to the steps in the CONTRIBUTING document and everything looks good to me

1) Is the submission free of defects and is it ready to be widely deployed?
		Yes, I am running it now

2) Does the submission provide a "high impact" benefit to real-world users performing real-world tasks?
		Yes, other sensors have similar options and CPU load is reduced

3) Is the copyright of the submission clear, non-gratuitous, and compatible?
		Yes, I coded it for free

4) Does the submission follow guidelines specified in the Klipper documentation?
		Yes, I based the code on a similar functionality of another sensor

5) Is the Klipper documentation updated to reflect new changes?
		Yes

6) Are commits well formed, address a single topic per commit, and independent?
		Yes, (if commits will be merged using squash)

KevinOConnor:
Thanks.  I'm not sure what real world tasks can be accomplished with this new config option?  I regularly receive feedback that the printer config file is daunting to users and I'd prefer not to add more options unless there is a well understood requirement for it.

A high "MCU RPI Load" does not, in and of itself, indicate a problem.  The statistic is a heuristic based on scheduling jitter of "low priority tasks" in the micro-controller.  Sporadic i2c requests will naturally increase the scheduling jitter, but that does not indicate a problem.

Cheers,
-Kevin

ReXT3D:
> I am planning to add one of those sensors on a canbus board and I wonder how much of an effect it will have on the canbus traffic. for now I can't get the chip working and I think it's because i2c.c does not allow to use the PB3 PB4 pins of the I2C on the BTT EBB36 V1.2 board.

FWIW, the sensor works just fine and without any issues over CAN, have a look here: https://github.com/Klipper3d/klipper/pull/5733#issuecomment-1241411851

PR #5758, opened 2022-09-06
virtual_sdcard: read new lines as-is

### LINKED ISSUE #5755: Wrong `file_position` reported under Python 3 when file has CRLF line endings
We've had reports in Fluidd of incorrect progress being reported and only now we found why this was happening!

Klipper uses the python inbuilt `open` function here:

https://github.com/Klipper3d/klipper/blob/9f8f3cd77d97339bd825053b064ae4d08d5d6902/klippy/extras/virtual_sdcard.py#L180

This works just fine under python 2, however this function seems to just ignore the CR of files with CRLF line endings!

To prove this, I ran the following steps:

- Started a clean debian docker container with `docker run --rm -it debian bash`
- `apt update && apt install -y python2 python3`
- Created a very simple file with CRLF by running `echo -ne "pedro\r\nlamas">test.txt`

After the above, I ran `python2` and these commands:
```python
f = open('test.txt', 'r')
data = f.read(100)
f.close()
for x in range(len(data)):
    print(ord(data[x]))
```

Output from python2:
```text
112
101
100
114
111
13
10
108
97
109
97
115
```

I can see clearly the CR (code 13) on the above output.

Also, doing `len(data)` returns 12, which is what I expected.

Running the same on `python3`, and this is the output:

```text
112
101
100
114
111
10
108
97
109
97
115
```

We can see clearly that CR (code 13) on the above output is missing!

Also, doing `len(data)` will return 11 instead of 12.

Bottom line is that `open` function under python3 is ignoring CR chars, thus leading to an incorrect `file_position` being reported.

pedrolamas:
The existing code works perfectly under python2: the internal `open` function will open the file and allow reading of the data with unaltered line-endings (meaning that lines will be CRLF or LF ended)

But python3 has changed the behavior of the `open` function to be an alias of `io.open`. And that brings in the Universal Line Endings as default, meaning that any CRLF will be read as LF.

Context [here](https://docs.python.org/3/howto/pyporting.html):

> You should also use io.open() for opening files instead of the built-in open() function as the io module is consistent from Python 2 to 3 while the built-in open() function is not (in Python 3 it’s actually io.open()).

This changes the behavior by ensuring we read line-endings as-is (by using `io.open` and passing in `newline=''`)

From my tests, this change will work fine with python2 and python3.

Fixes #5755 

### DIFF
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2018  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import os, logging
+import os, logging, io
 
 VALID_GCODE_EXTS = ['gcode', 'g', 'gco']
 
@@ -177,7 +177,7 @@ class VirtualSD:
             if fname not in flist:
                 fname = files_by_lower[fname.lower()]
             fname = os.path.join(self.sdcard_dirname, fname)
-            f = open(fname, 'r')
+            f = io.open(fname, 'r', newline='')
             f.seek(0, os.SEEK_END)
             fsize = f.tell()
             f.seek(0)


KevinOConnor:
Thanks.

-Kevin

PR #5763, opened 2022-09-09
Add TMC Linearity Correction

TheRealOrange:
Ported the TMC2130 linearity correction algorithm from Prusa firmware to klipper, for TMC2130 and TMC5160 drivers.

Note that even though the `SET_TMC_WAVE` gcode has been added, do not try to use the calibration tower provided [here](https://help.prusa3d.com/article/extruder-linearity-correction-calibration_2254) without modification, as the `SET_TMC_STEP` gcode is not supported by klipper and I have not added it.

I have tested this code with my printer which runs TMC2130 drivers by calling `SET_TMC_WAVE FAC=x` for 0 to 200, and I have performed multiple prints with no issues. My code changes adhere to the review guidelines as far as I can tell.

## Reason
Many people who use TMC2130 drivers, especially when not using a geared extruder, see a moire pattern on their prints, which is aesthetically unappealing. This aims to address that by porting over the Prusa code for linearity correction.


dmbutyugin:
Thank you for taking time to port this. Since you've tested it, I'd be curious to learn how it performs in real-life prints and if you managed to improve the print quality and by how much? Can you share some example prints before/after that demonstrate the improvements? And also explain the process you used to tune the parameter. 

The reason I am asking is because around a 1-2 years ago I also tried to play with this sine wave table of tmc2130/tmc5160. At the time, I implemented a generic compression code
```
    struct tmc_sine_regs {
        uint32_t mslut[8];
        uint32_t mslutstart, mslutsel;
    };
    struct tmc_sine_regs* tmc_sine_compress(double func[256], uint8_t scale);
```
that could compress arbitrary 256-valued function into the corresponding register values based on the TMC encoding scheme or report that the error is too large. At the time I tried different functions, including the one suggested by Prusa research. Alas, I found that, at least on my hardware, the results from the Prusa correction weren't that helpful and considerably different functions are necessary. However, I tried this primarily to optimize the motion of the toolhead rather than the extruder. And ultimately I abandoned this effort because the improvements were not very significant and it wasn't clear how to really properly tune this sine table (the process suggested by TMC is quite fiddly and impractical in a small workshop setup). Later experimentation also suggested that updating the table for 1 full step may be insufficient, it actually may be necessary to have different wave forms for 2 or even 4 steps. But I admit that it may have just been my bad luck and maybe there is some hardware where updating the sine wave table might be beneficial. And so hence my question.

Ramalama2:
> 

Just for the testing steps.
As far i understand:
1. We need to generate in the slicer a squared tower.
2. Add every 20 layers: SET_TMC_WAVE FAC=[i+10]
3. And check simply which 20 layers were the best?
4. Then probably fine tune it, if we have a "best" range
5. Print examples with and without wave, and show the results here?

Cheers

TheRealOrange:
It's a bit more complicated than that. What the Prusa code does is that it deliberatesly offsets the phase of the microstep table at each layer height to produce a pronounced moire diagonal pattern, which I am unable to replicate yet because I have yet to add code to schedule steps based on MSCNT. 

However, specific prints with coincidentally correct perimeter lengths will show an obvious pattern, and when I encounter one of these, I will compare it with linearity correction enabled Vs disabled and post the results here.

 Currently unable to test as I am lacking time.

RomanOvchinnikov:
> Currently unable to test as I am lacking time.

I can help with printing tests on TMC2130 (extruder direct BMG clone), if you provide method with parameters you desire.

KevinOConnor:
Thanks for working on this.  There were some questions on the implementation and how much improvement one can measure from this change.  I understand this takes time; there is no rush.  We can reopen this PR when a submitter has sufficient time and this PR is ready to be merged.  (Or, a new PR can be created at that time.)

Klipper Discourse may also be used for general discussions on the topic.

Cheers,
-Kevin

Ramalama2:
> Thanks for working on this. There were some questions on the implementation and how much improvement one can measure from this change. I understand this takes time; there is no rush. We can reopen this PR when a submitter has sufficient time and this PR is ready to be merged. (Or, a new PR can be created at that time.)
> 
> Klipper Discourse may also be used for general discussions on the topic.
> 
> Cheers, -Kevin

Thanks Kevin 👍

Riconec:
is this feature also available for tmc2209?

PR #5767, opened 2022-09-10
mpu9250: add MPU6500

dingyifei:


dingyifei:
I have tested MPU6500 with a pico, the results are beautiful

Additionally, MPU9250 and 6050 are obsolete parts, and most of them on the market have a really high failure rate. I think these should be mentioned in the doc (havent add them yet)

y
x

KevinOConnor:
Thanks.  Maybe @dmbutyugin or @bluesforte have comments on this?

-Kevin

docs/Measuring_Resonances.md
```diff
@@ -31,9 +33,17 @@ and **will not work**. The recommended connection scheme:
...
-accelerometer has been tested to work over I2C on the RPi at 400kbaud.
-Recommended connection scheme for I2C:
+Fritzing wiring diagrams for some of the ADXL345 boards:
+
+![ADXL345-Rpi](img/adxl345-fritzing.png)
+
+#### MPU-9250/MPU-6050/MPU-6500
+
+Alternatives to the ADXL345 are MPU-9250, MPU-6050, and MPU-6500. These
+accelerometer has been tested to work over I2C on the RPi or RP2040(pico) at
```
dmbutyugin, line 43: s/accelerometer has/accelerometers have/

docs/Measuring_Resonances.md
```diff
@@ -42,14 +52,24 @@ Recommended connection scheme for I2C:
 | SDA | 03 | GPIO02 (SDA1) |
 | SCL | 05 | GPIO03 (SCL1) |
 
+![MPU-9250 connected to RPI](img/MPU9250-PI-fritzing.png)
```
dmbutyugin, line 55: The image links here appear to be broken and do not render in the viewer:
https://github.com/Klipper3d/klipper/blob/04d85a04cd7ea4a0d96fd346c4751f768545e453/docs/Measuring_Resonances.md
*I suspect this has to do with the wrong upper/lower case letters in the name.

Separately, the images seem to be pretty low-res, not sure how they look like in the doc though.

docs/Measuring_Resonances.md
```diff
@@ -42,14 +52,24 @@ Recommended connection scheme for I2C:
...
+| SCL | 02 | GP1 (I2C0 SCL) |
 
+![MPU-9250 connected to PICO](img/MPU9250-PICO-fritzing.png)
 
-Double-check your wiring before powering up the Raspberry Pi to prevent
-damaging it or the accelerometer.
+***I2C protocol is not reliable on long distance, consider hanging PICO
+within 5cm cable length away from the accelerometer!***
+
+***Double-check your wiring before powering up to prevent
```
dmbutyugin, line 71: This comment now appears to be applicable only to MPU* accelerometers, and the users of ADXL345 may skip it and not read it. I would suggest to either duplicate it in ADXL345 section, or alternatively add it as a first paragraph to `### Wiring` section before going to any of the accelerometers.

dmbutyugin (commented):
Thank you! I had a few comments on the documentation changes. Implementation-wise, it is pretty trivial, so I don't see anything off here. However, it would be good if someone who have or used one of these MPU accelerometers took a look at the documentation update and shared their feedback, e.g. @bluesforte or @Sineos

docs/Measuring_Resonances.md
Sineos: I could reliably run the MPU950 with a Cat5e cable of 50cm. My feeling is that the ADXL and MPU are equally demanding WRT cable quality.

Maybe worth to resurrect #5059

dingyifei: It is probably a problem with my setup. I have all the wires basically in parallel vs twisted pairs, and I couldn't get anything to work  at around 30-10 cm and I had to shorten it to around 5 to work. Maybe it's a problem with pico?

docs/Measuring_Resonances.md
dingyifei: Yes they are a-bit blurry. I'll try to generate some new ones.

docs/Measuring_Resonances.md
Sineos: Tested both Pico and RPi. Could not see any differences between these two WRT cable.
Any regular Dupont jumper cable were giving me serious issues, even on the regular 10 to 20 cm lengths.
No problems since I switched to a shielded ethernet cable (Cat 5e) and [this wiring setup](https://github.com/Klipper3d/klipper/pull/5059#issuecomment-1001714082)

dingyifei:
One thing I'm still unsure of is to add WMI of all MPUs, because the refurbished chips appear to have various deviceID due to internal issues but will function as a normal MPU-6050

The workaround for slimeVR is
`return deviceId == 0x68 || deviceId == 0x70 || deviceId == 0x71 || deviceId == 0x73 || deviceId == 0x75; // Allow any MPUs`

docs/Measuring_Resonances.md
dingyifei: I have updated the document with some more general suggestions. I'll still run some tests tomorrow and see if different combinations/lengths have different effects.

Sineos:
I'm not sure that it makes sense to add this pull up resistor reference. A user who is not able to setup a proper cable most likely will do more harm than good with this information.

Otherwise, LGTM 👍

docs/Measuring_Resonances.md
dmbutyugin: I was actually referring only to the 'Double-check your wiring before powering up to prevent...' comment, and not to the I2C cable length limit. Alas, github review pulled a larger block than I intended, sorry for the confusion. So, I think that warning about double-checking the wiring makes sense to have for all wiring setups.

dmbutyugin: Though I'm fine with the current version of the docs, modulo the fact that I haven't been using twisted pair and I cannot confirm or say contrary to the 'Suggested twisted pair order'(s). Also, the same as what Sineos has said, I'm not sure I'd go as far as to suggest replacing the pull up resistors. It is a fiddly and potentially dangerous process, and I suppose the people who have the equipment and skills to pull that are already aware of that possibility, WDYT?

dmbutyugin:
> Should I add Device ID of all MPUs, because the refurbished chips with defects appear to have various deviceID but function as a normal MPU-6050. If some people get mpus with these weird ids, they are still likely to have working MPUs.

OK, I'm fine with it. At this rate, maybe you can now define a list of acceptable device IDs rather than having that long list of comparisons?

docs/Measuring_Resonances.md
```diff
@@ -31,9 +50,31 @@ and **will not work**. The recommended connection scheme:
...
+GND+SDO
+3.3V+SDA
+GND+SCL
+```
+
+If you still experience signal integrity issues and can't shorten your cable, replacing the pull up resistors to lower resistance on SDA and SCL may help. [This document](https://www.ti.com/lit/an/slva689/slva689.pdf) explained the calculation behind choosing the right pull up resistor value. **This may cause DAMAGE to the controller/RPI and Accelerometer!!!**
+
+##### MPU-9250/MPU-6050/MPU-6500
+
+**MPU-6050 and MPU-9250 are obsolete parts! MPU-6500 are replacements of MPU-6050 at a lower failure rate.**
```
dmbutyugin, line 71: > MPU-6050 and MPU-9250 are obsolete parts!

The exclamation mark is probably too much here, it's not like it is something to pay super close attention to.

dingyifei:
> > Should I add Device ID of all MPUs, because the refurbished chips with defects appear to have various deviceID but function as a normal MPU-6050. If some people get mpus with these weird ids, they are still likely to have working MPUs.
> 
> 
> 
> OK, I'm fine with it. At this rate, maybe you can now define a list of acceptable device IDs rather than having that long list of comparisons?
> 
> 

Yes, I this line of code is a SlimeVR Arduino workaround, I'll add a warning message for suspicious IDs and use a list for the known IDs.

dingyifei:
I haven't tested the last two changes I've made, but I have included a few more MPU IDs. Note that I changed MPU9250's ID and added MPU9255 with the same ID as the previous MPU-9250 based on the official documents. It is common to get these MPUs just by luck. 0x75 and 0x69 are IDs of MPU 6xxx/9xxx with some internal connection error. Based on SlimeVR community discord, these usually work as usual but some reported worse measurements/more noise.

dingyifei:
Sorry abt the delays..... my testing plan...... is not going well...... still couldn't find enough time to crimp a piece of cable for 50 times

docs/Measuring_Resonances.md
bluesforte: I would probably suggest just removing the whole last line.  I haven't seen any evidence of problems with MPU-6050 (and in fact some searches suggest quite the opposite - that they are still used in many projects working for years!).  imo, I think we should leave it to the reader to have done their research on the parts and pick the one they like.

BTW - even the MPU-6500 is not currently recommended for new designs.  It seems invensense is moving to the ICM-* products to replace (if curious you can see a [fork with tested support for the ICM-20948 here](https://github.com/bluesforte/klipper-mpu9250/tree/mpu-blues-icm)).

bluesforte:
Just saw the notification now for my input on this.  Good changes - always a fan of improving the docs and really like the new fritzing diagrams.  Good to have the device ids in a list now that there are more than 2.

Only real feedback I have is the one I left above re: mpu devices' production status.  My opinion is that it's not worth calling out because a) even if they aren't 'officially' supported, all the mpu devices are still very readily available for purchase at retailers, and b) I think users can do their own research/shopping and figure out which model they want to go with, since all of them are plenty accurate to perform input shaping.

dingyifei:
> Just saw the notification now for my input on this.  Good changes - always a fan of improving the docs and really like the new fritzing diagrams.  Good to have the device ids in a list now that there are more than 2.
> 
> 
> 
> Only real feedback I have is the one I left above re: mpu devices' production status.  My opinion is that it's not worth calling out because a) even if they aren't 'officially' supported, all the mpu devices are still very readily available for purchase at retailers, and b) I think users can do their own research/shopping and figure out which model they want to go with, since all of them are plenty accurate to perform input shaping.

I'll take that line out, it seems to be redundant. In the case they find something wrong with their mpu, existing sources with better explanations are readily available.

docs/Measuring_Resonances.md
```diff
@@ -99,7 +141,10 @@ Afterwards, check and follow the instructions in the
 Make sure the Linux SPI driver is enabled by running `sudo
```
dmbutyugin, line 141: I think this sentence 'Make sure the Linux SPI driver is enabled by running...' can be moved at the start of the next section for adxl345. Then it will follow the same logic as the recommendation to enable I2C with the MPU chip

KevinOConnor:
Thanks.  On the top right hand side of this github PR page there is an "assignees" option - if you assign this PR to yourself then it'll turn off the bot.

-Kevin

dmbutyugin:
Hmm, for some reason, I thought that it is handled via 'Reviewers' list, but seems to be not the case. Tried the 'assignees' now.

dmbutyugin (commented):
@dingyifei, thank you!

Besides these couple of comments to change the documentation, I tested your branch, and it works all fine:
```
Found mpu-6500 with id 70
MPU9250 starting 'mpu_test' measurements
accelerometer measurements started
.............
```
I used the powering I suggested in the comment:
3V3 (pin 36) -> VCC
GND (pin 38) -> GND
and the `i2c0e` bus (but this should not matter, the suggested I2C connection is fine).

dingyifei:
> @dingyifei, thank you!
> 
> 
> 
> Besides these couple of comments to change the documentation, I tested your branch, and it works all fine:
> 
> ```
> 
> Found mpu-6500 with id 70
> 
> MPU9250 starting 'mpu_test' measurements
> 
> accelerometer measurements started
> 
> .............
> 
> ```
> 
> I used the powering I suggested in the comment:
> 
> 3V3 (pin 36) -> VCC
> 
> GND (pin 38) -> GND
> 
> and the `i2c0e` bus (but this should not matter, the suggested I2C connection is fine).

Hi, I'm really busy in the last few weeks, thank you for giving these comments. I'm really sorry for being inactive for so long.

dmbutyugin:
@dingyifei No worries, thank you for your contribution!

Sineos:
Nice work @dingyifei 👍. Some comments:

If you are already working on it, you could also add the combination for ADXL and Pi Pico. All the details, including a fritzing can be found here: https://klipper.discourse.group/t/raspberry-pi-pico-adxl345-portable-resonance-measurement/1757

There is also a report about numpy compatibility here: https://github.com/Klipper3d/klipper/issues/5831
I cannot reproduce, since on my system automatically v1.16.6 got installed. Acc to the [documentation](https://numpy.org/devdocs/release/1.19.0-notes.html), v1.19 and above is no longer compatible with Python2. 

My tests have shown that it is generally preferable to use the `Vin` port of the ADXLs instead 3V3 (if both are present). An updated fritzing can be found here: https://klipper.discourse.group/t/adxl345-different-between-rpi-pico-and-rpi-3b/1842/30?u=sineos

In the same discourse thread, it was found that pulling GPIO23 high on the Pico, e.g. with
```
[static_digital_output pico_3V3pwm]
pin: pico:gpio23
```
is beneficial, due to more stable power supply. It has quite an effect on the ADXL. Did not test with the MPU but surely does not hurt there either.

dmbutyugin:
I don't mind making additional changes to the PR, but it is also better to have something complete merge than to have never-ending PR that grows bigger and bigger :) So, @dingyifei up to you. I think the suggestion about pulling `gpio23` high on Raspberry Pi Pico specifically is actually pretty good (though it is only applicable to Raspberry Pi Pico, for example I've seen standalone boards like FYSETC PortableInputShaper based on Pico that do not require that).

NumPy issue would require further looking. My impression was that `pip` should figure out the correct compatible version. At least, it does a good job for Python2 installations. Not sure what's going on there.

Also, @Sineos, feel free to submit your own PR with ADXL-related suggested changes ;)

Sineos:
> Also, @Sineos, feel free to submit your own PR with ADXL-related suggested changes ;)

I could and would have done so already but I do not feel comfortable with the real name sign off. My personal policy is to keep my real name off the internet as much as possible. So no facebook, social media, extensive use of VPNs, etc. Maybe I'm just too old, too paranoid, whatever. Not going to change this for a few lines of documentation PRs 😄 

@dingyifei take what you deem suitable and thanks for the effort put in!

docs/Measuring_Resonances.md
```diff
@@ -42,14 +80,18 @@ Recommended connection scheme for I2C:
...
+![MPU-9250 connected to RPI](img/mpu9250-PI-fritzing.png)
 
-Fritzing wiring diagrams for some of the ADXL345 boards:
-
-![ADXL345-Rpi](img/adxl345-fritzing.png)
+Recommended connection scheme for I2C(i2c0a) on the RP2040:
 
+| MPU-9250 pin | RP2040 pin | RPi pin name |
+|:--:|:--:|:--:|
+| 3V3 (or VCC) | 39 | VSYS |
```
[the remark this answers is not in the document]
dingyifei: I don't think most MPU boards come with a 3.3v pin. They all have a little LDO that can take 5v and the input is marked as VCC. I think some errors happened when I was making the chart. Is there any particular reason to use 3.3v out v.s. the 5v?

dingyifei:
I think everything is getting a bit messy and hard to keep track of. I have modified the PR according to most suggestions. I know many people got ADXL 345 working with STM and other chips so I just added a line to mention the possibility.

I think we should merge everything in its current state.

docs/Measuring_Resonances.md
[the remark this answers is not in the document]
Sineos: If the MPU board has level shifters and you connect it to 5V you have a good chance to kill the host since most host boards are not 5v tolerant. In addition you cannot rely on the fact that there is always a LDO

Board without LDO:
https://www.sparkfun.com/products/11028

There are ADXLs with Level Shifters, don't know if there are MPU boards with them as well.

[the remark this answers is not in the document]
dmbutyugin: Yeah, as Sineos said, the main worry was about frying either an accelerometer or a Pi in case of voltages mismatches. If an MPU board has a logic level shifter, RPi may be damaged, if an MPU doesn't have an LDO, the accelerometer may get damaged. An example of a board with the logic level shifters (on some pins):
https://learn.adafruit.com/mpu6050-6-dof-accelerometer-and-gyro/downloads
So, thanks for updating, now it looks good.

dmbutyugin:
@dingyifei Thanks! I took a look and overall looks good! Just a couple of notices / suggestions:

> I know many people got ADXL 345 working with STM and other chips so I just added a line to mention the possibility.

The main issue there is voltage level compatibility. Many MCU boards only have 5V available, and then depending on whether the accelerometer has a logic level shifter and whether the SPI pins are 5V tolerant or not it might or might not be safe to connect. That's why it wasn't suggested by default. But overall, what you wrote is actually a reasonable middle ground. If possible, I would highlight the issue:
```
**Note: Some MCUs will work with an ADXL345 in SPI mode (e.g. Pi Pico), wiring and
configuration will vary according to your specific board and avaliable pins.
Double-check the acceptable and actual voltage levels on both the MCU board and
the accelerometer for all pins before proceeding.**
```

Another small comment is that your wiring diagram for Pico states `i2c0a` bus, but the sample configuration - `i2c1a`. I would make it consistent and have the same bus in both cases (I don't care which one specifically).

> I think we should merge everything in its current state.

I'll give @Sineos a chance to take another look at the text, but modulo these couple of tiny suggestions, I agree.

Sineos:
LGTM

KevinOConnor:
Thanks.  Sorry for the delay.  I've now merged this PR.

-Kevin

PR #5771, opened 2022-09-11
Fix Ender 3 S1 Config

Fisheiyy:
I fixed the max temperature for the hotend and bed for the S1 non-pro based off the [official diagram](https://www.creality3dofficial.com/files/goods/20220531/s1vsprovsPlus11.jpg) on the Creality website, fixed some crashing issues reported in #5689, and added the F401 Motherboard Variant clarifications per #5709.

Signed-off-by: Rob Casper [myallneedsemail@gmail.com](mailto:myallneedsemail@gmail.com)


KevinOConnor:
Thanks, it looks fine, but I don't see a signed-off-by line.

-Kevin

Fisheiyy:
> Thanks, it looks fine, but I don't see a signed-off-by line.
> 
> -Kevin

My bad, adding that now

KevinOConnor:
Thanks.

-Kevin

kpavey:
Will this work on the Ender 3 S1 Pro as is?

Fisheiyy:
> Will this work on the Ender 3 S1 Pro as is?

Yes this config was already merged into main, so it is the default config you see in the config_examples folder, it will work just fine on the S1 but if you want to use the high temperature aspect of it you need to change the max temperature to 300.

carlhaun:
Hello :

Prior to flashing my new Ender 3 S1 Pro will I still have to get the numbers off of the motherboard processor ?  I have read that you had a couple of different .bin files for the new Ender 3 s1 Pro which depends upon the serial number of the processor...seems like there would be an OctoPrint command for this ...?

Thanks !

Fisheiyy:
> Hello :
> 
> Prior to flashing my new Ender 3 S1 Pro will I still have to get the numbers off of the motherboard processor ?  I have read that you had a couple of different .bin files for the new Ender 3 s1 Pro which depends upon the serial number of the processor...seems like there would be an OctoPrint command for this ...?
> 
> Thanks !
> 
> 

There is no way to find this out without looking at the actual mcu, there is no commands, yes you still need to find the mcu type you have and need to make the firmware for that version

carlhaun:
Thanks Fisheiyy !

I think I-am-a-gonna wait a bit until I flash Klipper onto my new toy then....

Your prompt response is greatly appreciated  -

grantland:
The stepper values here cause my S1 to crash when doing the bed leveling mesh due to `pos_y` being out of bounds.

FWIW I've pulled _just_ the `stepper_*` and the `bed_mesh` values from #5689 and it seems to be working for me now.

Fisheiyy:
> grantland

https://github.com/Klipper3d/klipper/pull/5689#issuecomment-1363351595

Fisheiyy:
> The stepper values here cause my S1 to crash when doing the bed leveling mesh due to `pos_y` being out of bounds.
> 
> FWIW I've pulled _just_ the `stepper_*` and the `bed_mesh` values from #5689 and it seems to be working for me now.

https://github.com/Klipper3d/klipper/pull/5689#issuecomment-1363351595

PR #5777, opened 2022-09-18
Swap acceleration and velocity in Wanhao D9 example

mdshw5:
I originally had these values set incorrectly, and this fixes my earlier mistake.


KevinOConnor:
Thanks.   The change is fine, but it would need to have a signed-off-by line.

-Kevin

mdshw5:
@KevinOConnor of course. I'll close and open a new PR with appropriate signed-off commit messages. Sorry for the time waste.

PR #5784, opened 2022-09-22
Update instructions for generic-th3d-ezboard-lite-v2.0.cfg

Apulo:
Corrected the build instructions for the TH3D EZBoard V2 to include the command to convert Klipper.elf to a SREC bin format named firmware.bin. The SREC format is required for the bootloader installed on the board.


KevinOConnor:
Closing as it looks like a dup of #5785 

-Kevin

PR #5785, opened 2022-09-22
Update generic-th3d-ezboard-lite-v2.0.cfg

Apulo:
Corrected the build instructions for the TH3D EZBoard V2 to include the command to convert Klipper.elf to a SREC bin format named firmware.bin. The SREC format is required for the bootloader installed on the board.


Apulo:
Signed-off-by: Anthony Dellett <anthony.dellett@gmail.com>

KevinOConnor:
Thanks.  As a minor thing, I think it might be more clear if the instructions used the form `arm-none-eabi-objcopy -O srec out/klipper.elf out/firmware.bin` so as to avoid needing to change directory.

-Kevin

Apulo:
Thanks! I will do that, I also noticed that I have a couple of typos in
there. I'll fix those also.

PR #5788, opened 2022-09-24
Add grouping for LEDs

julianschill:
This adds the functionality to create groups of LEDs to either separate a chain or combine multiple chains to one big LED chain.

Configuration is like this:
```
[led_group mygroup]
leds:
  neopixel:grbw (2-4,6-7,10)
  neopixel:grb
  dotstar:my_dotstar (32-35)
  ```

The indexing is then done in the order of the group definition. 
The code was partly taken out from the LED effect module, created by Paul McGowan and maintained by me.

Signed-off-by: Julian Schill <j.schill@web.de>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2829,7 +2829,18 @@ PCA9632 LED support. The PCA9632 is used on the FlashForge Dreamer.
 #initial_WHITE: 0.0
 #   See the "led" section for information on these parameters.
 ```
+### [led_group]
 
+Groups LEDs together into LED groups, which then can be used like any LED
+chain.
+```
+[led_group my_group]
+#leds:
+#   A list of LEDs to put in the group. Every chain in a new line with the chain
+#   name, optionally followed by the index range in brackets. Ranges can be
+#   combined by separating them with a comma. Example:
+#   neopixel:my_neopixel (2-4,6-7,10)
+```
 ## Additional servos, buttons, and other pins
 
 ### [servo]
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2838,8 +2838,10 @@ chain.
 #leds:
 #   A list of LEDs to put in the group. Every chain in a new line with the chain
 #   name, optionally followed by the index range in brackets. Ranges can be
-#   combined by separating them with a comma. Example:
-#   neopixel:my_neopixel (2-4,6-7,10)
+#   combined by separating them with a comma. LEDs are indexed in the order they
+#   are configured. Example:
+#   my_neopixel (2-4,6-7,10)
+#   my_dotstar
 ```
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  Can you provide some background on why this feature is useful?  Why would users want to organize LEDs into "virtual chains"?  Can they use macros or display_templates for the same effect?

Separately, at a minimum, this code would need to be reworked to not "peek" into member variables of external classes.

-Kevin

julianschill:
Ican thonk of two setups where this is useful. First setup would be a really long chain, where e.g. the chamber LEDs are followed by a toolhead Logo LED and two nozzle LEDs. This is a common setup for Voron printers. So the users could define their virtual chains "chamber", "logo", "nozzle" and set them directly without memorizing the indeces. You could also assign display templates much easier to those LEDs. 

Second application are several chains on different pins, that should be switched together. Like 4 chains on the top of the printer connected to four pins on the MCU. So you could switch them on and off with one command or apply a display template on all of them. 

I think this could partly done by macros, but it would be much more convenient to have such an abstraction layer. This could then also be used then by the frontends to switch those LEDs directly. So you could add a color selector for each LED group.

I will rework the code to use getter and setter functions.

meteyou:
currently, I am also working on neopixel groups in mainsail. however, these are only "virtual" and therefore a single GUI solution. furthermore, I therefore also have to send a lot of gcode every time several neopixels are addressed. this would also look cleaner with the groups in klipper.

image

because I currently only have the stealthburner with neopixels (i personally don't use many light effects in my printers), I can only show it from this setup. mainsail always sends 2 lines for the nozzle LEDs. but if users have 10 neopixels per group, it will be a lot of gcode...

so I would very much welcome this function in klipper. just my 2 cents...

KevinOConnor:
Thanks.  I noticed a couple of things looking through this PR:
1. An "extra" module should not store a reference to the "config" object.  This just leads to confusion, as it's generally not valid to read config parameters after the initial printer setup phase.  In some cases, it looks like this is just done for error reporting (where `raise self.printer.config_error()` is an easy replacement).  I'm not sure about the other cases.
2. FWIW, the config format seems overly complicated to me.  I fear adding in a new "syntax" will just be a burden to users.

Finally, I'm not sure I'm a good candidate to review this PR.  I don't have any extensive LED setups and it's not a particular area of interest to me.  Klipper needs more core developers willing to review PRs - without that, I fear this PR will not get sufficient review attention.

-Kevin

Riconec:
sad it was closed. I have mini12864 with 3 neopixels and I wanted to split them into display backlight and encoder leds to maybe have effects on encoder but I don't sure I can do that in stock klipper

jonferreira:
Shame this was dismissed. If you daisy chain LED Rings,for example,this is extremely useful!

PR #5789, opened 2022-09-25
Added: RAMPS-FD v2 config example

soimy:
Since no one is doing this, I figure this myself based on marlin configuration and pin out map of RAMPS-FD v2


KevinOConnor:
Config files will need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html and commits need to have a signed-off-by line https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

PR #5790, opened 2022-09-25
STM32H7xx: Separate out each SPI pin function number

### LINKED ISSUE #5773: SPI3 not correctly assigned on STM32H7XX MCU based boards
the issue is this line here https://github.com/Klipper3d/klipper/blob/master/src/stm32/stm32h7_spi.c#L70
it assigns all 3 SPI3 pins to function 6
but on the H742x/H743x at least that would configure PB4 and PB3 for SPI3 but PB5 would end up talking to the I2C4 device! To get PB5 to talk to SPI you need to set it's alternate function number to 7 not 6, which it appears that the current code doesn't let you do.

Look at page 89 of datasheet: https://www.st.com/resource/en/datasheet/stm32h743vi.pdf 
This issue specifically affects the fly-super8pro board, that uses hardware SPI 3 for TMC driver control

@adelyser I believe you were the original author of this code.

adelyser:
STM32H7xx

Fixes #5773 

Separated out the SPI functions for each pin. On the STM32H743, SPI3 has a combination of pins (PB3, PB4, and PB5), where PB5 uses a different function number from the other 2 pins. This change resolves this issue.

Signed-off-by: Aaron DeLyser [bluwolf@gmail.com](mailto:bluwolf@gmail.com)

### DIFF
--- a/src/stm32/stm32h7_spi.c
+++ b/src/stm32/stm32h7_spi.c
@@ -12,7 +12,7 @@
 
 struct spi_info {
     SPI_TypeDef *spi;
-    uint8_t miso_pin, mosi_pin, sck_pin, function;
+    uint8_t miso_pin, miso_function, mosi_pin, mosi_function, sck_pin, sck_function;
 };
 
 DECL_ENUMERATION("spi_bus", "spi2", __COUNTER__);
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Just out of curiosity, is there a reason to define spi3 in this case, as it seems like users could just use spi1 for PB4,PB5,PB4.

-Kevin

adelyser:
> Thanks. Just out of curiosity, is there a reason to define spi3 in this case, as it seems like users could just use spi1 for PB4,PB5,PB4.
> 
> -Kevin

Well, you are totally right on that! I was just looking into an issue and overlooked that the spi pins are used elsewhere. I think if we don’t proceed with this change, we should remove that particular spi combination so others don’t try to use it

KevinOConnor:
Okay, thanks.  For now, I think we can just remove the incorrect definition (commit e4c8a247).  The spi and i2c definitions, on all platforms, could use a bit of a rework so that it is easier to choose appropriate pins.

-Kevin

PR #5792, opened 2022-09-26
controller_fan: Add min_speed configuration

trippwill:
Adds a min_speed configuration value that allows the user to override the minimum speed. The min_speed value is used on startup and after the idle_speed timeout.

This allows a single fan in a combined electronics enclosure to always run at a defined speed to keep air flowing even when the configured steppers and
heaters are not active.

Signed-off-by: Charles Willis (5862883+trippwill@users.noreply.github.com)

### DIFF
--- a/klippy/extras/controller_fan.py
+++ b/klippy/extras/controller_fan.py
@@ -23,6 +23,8 @@ class ControllerFan:
         self.idle_speed = config.getfloat(
             'idle_speed', default=self.fan_speed, minval=0., maxval=1.)
         self.idle_timeout = config.getint("idle_timeout", default=30, minval=0)
+        self.min_speed = config.getfloat(
+            'min_speed', default=0., minval=0., maxval=1.)
         self.heater_names = config.getlist("heater", ("extruder",))
         self.last_on = self.idle_timeout
         self.last_speed = 0.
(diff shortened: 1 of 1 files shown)


trippwill:
If the change is acceptable I can push another commit with the updated docs.

ReXT3D:
I can confirm that this works as intended.  One thing worth noting is that if `min_speed` is set to a value lower than `off_below`, the fan will be off.  This is good from overall logic perspective but could potentially be confusing for some users.  In other words, it might be useful to state in the documentation that `min_speed`, when set, has to be no lower than `off_below`.

FWIW I personally think this is a very useful change for the exact reason you stated.

trippwill:
Does it make more sense for min_speed deafult and/or min value to be set to off_below?

trippwill:
Also, I run this change on my own SKR Pico and it works as expected (and solves my Raspberry PI 3 overheating issue :-) ).

ReXT3D:
> Does it make more sense for min_speed deafult and/or min value to be set to off_below?

You will have to wait for Kevin or an official reviewer to comment.

> Also, no notes or discord discussion I could find on why PR 4810 was withdrawn. It's exactly the same as this change.

I am not sure, the author just suddenly withdrew it...

trippwill:
> > Does it make more sense for min_speed deafult and/or min value to be set to off_below?
> 
> You will have to wait for Kevin or an official reviewer to comment.

Giving this some additional thought, I think @ReXT3D original suggestion to include a note in the documentation makes more sense than tying min_speed to off_below. For controller_fan, the default behavior is meant to be turning the fan off, so a default of 0. makes the most sense.

PR #5800, opened 2022-09-29
z_thermal_adjust: hotfix for M105 internal error

alchemyEngine:
M105 expects a get_temp method for all sensors with a gcode_id defined. z_thermal_adjust crashes when the gcode_id parameter is set and M105 is called. 

Port get_temp and stats (for good measure?) methods from temperature_sensor.

Signed-off-by: Robert Pazdzior <robertp@norbital.com>


KevinOConnor:
Thanks.

-Kevin

PR #5803, opened 2022-10-03
config: Allow nozzle_diameter to be slightly larger than filament_dia…

sprior:
Allow for the configured nozzle_diameter to be slightly larger than the filament_diameter.  And example of this is a 1.8mm CHT nozzle with 1.75mm filament which some have had success with.


sprior:
Is there anymore I need to do to get this accepted?

dmbutyugin:
Please note that the commits need to be formatted according to the following requirements:
https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages
(you need to provide a Signed-off-by line that indicates you agree to the [developer certificate of origin](https://www.klipper3d.org/developer-certificate-of-origin)).

Otherwise, note that nozzle diameter is only used for some sanity-checking of the gcode, so even if you use a nozzle of 1.8mm, you don't really have to set it that way in the config. Just setting it to something like 1mm should generally suffice. But in principle the change can be merged, I think.

PR #5804, opened 2022-10-05
Add support for CANbus on atsame51 chips

KevinOConnor:
This PR adds support for CANbus (and USB to CAN bridge mode) on atsame51 chips.

-Kevin


ReXT3D:
I was eager to try this on my Duet3 Mini 5+, but it unfortunately fails to compile:

```
pi@creality:~/klipper $ make KCONFIG_CONFIG=config.duet3mini
  Creating symbolic link out/board
  Building out/autoconf.h
  Compiling out/src/sched.o
  Compiling out/src/command.o
  Compiling out/src/basecmd.o
  Compiling out/src/debugcmds.o
  Compiling out/src/initial_pins.o
  Compiling out/src/gpiocmds.o
  Compiling out/src/stepper.o
  Compiling out/src/endstop.o
  Compiling out/src/trsync.o
  Compiling out/src/adccmds.o
  Compiling out/src/spicmds.o
  Compiling out/src/thermocouple.o
  Compiling out/src/i2ccmds.o
  Compiling out/src/spi_software.o
  Compiling out/src/sensor_adxl345.o
  Compiling out/src/sensor_angle.o
  Compiling out/src/sensor_mpu9250.o
  Compiling out/src/lcd_st7920.o
  Compiling out/src/lcd_hd44780.o
  Compiling out/src/buttons.o
  Compiling out/src/tmcuart.o
  Compiling out/src/neopixel.o
  Compiling out/src/pulse_counter.o
  Compiling out/src/atsamd/main.o
In file included from src/atsamd/internal.h:13:0,
                 from src/atsamd/main.c:11:
lib/same51/include/same51.h:53:4: error: #error Library does not support the specified device.
   #error Library does not support the specified device.
    ^~~~~
In file included from src/atsamd/main.c:11:0:
src/atsamd/internal.h:27:1: error: unknown type name 'Sercom'
 Sercom * sercom_enable_pclock(uint32_t sercom_id);
 ^~~~~~
src/atsamd/main.c: In function 'bootloader_request':
src/atsamd/main.c:24:20: error: 'HSRAM_ADDR' undeclared (first use in this function)
     writel((void*)(HSRAM_ADDR + HSRAM_SIZE - 4), 0xf01669ef);
                    ^~~~~~~~~~
src/atsamd/main.c:24:20: note: each undeclared identifier is reported only once for each function it appears in
src/atsamd/main.c:24:33: error: 'HSRAM_SIZE' undeclared (first use in this function); did you mean 'ARRAY_SIZE'?
     writel((void*)(HSRAM_ADDR + HSRAM_SIZE - 4), 0xf01669ef);
                                 ^~~~~~~~~~
                                 ARRAY_SIZE
src/atsamd/main.c:26:5: warning: implicit declaration of function 'NVIC_SystemReset' [-Wimplicit-function-declaration]
     NVIC_SystemReset();
     ^~~~~~~~~~~~~~~~
src/atsamd/main.c: In function 'armcm_main':
src/atsamd/main.c:33:5: warning: implicit declaration of function 'SystemInit' [-Wimplicit-function-declaration]
     SystemInit();
     ^~~~~~~~~~
make: *** [Makefile:64: out/src/atsamd/main.o] Error 1
pi@creality:~/klipper $
```

For reference, here is the most basic configuration I tried:

image

KevinOConnor:
Ah, thanks for looking at it.  I thought the Duet3 used at atsame51 chip, but it looks like it actually uses an atsame54 chip.  I've removed the misleading definition for "same51p20" which doesn't actually exist.

I'll try to see if adding support for same54 chips is straight forward.

-Kevin

ReXT3D:
> Ah, thanks for looking at it. I thought the Duet3 used at atsame51 chip, but it looks like it actually uses an atsame54 chip. I've removed the misleading definition for "same51p20" which doesn't actually exist.
> 
> I'll try to see if adding support for same54 chips is straight forward.
> 
> -Kevin

Thank you, that would explain why I could not find "same51p20" in the data sheet.

We are all currently compiling for SAMD51P20 target to run on the Mini 5+ which is indeed equipped with SAME54P20.  On the surface, the E54P20 has 2xCAN and Ethernet added to the D51P20 feature set.

KevinOConnor:
I added support for the "same54p20" chip and rebased this branch.  I don't have that particular chip, so let me know how testing goes (success or failure).

-Kevin

ReXT3D:
Thanks very much again for implementing this Kevin.

The rebased branch builds just fine now for SAME54P20 and everything appears to be functioning normally when using the traditional USB communication interface.  Unfortunately selecting 'USB to CAN bus bridge' communication interface option results in MCU that fails to be recognized by Klipper (there is no `/dev/serial/by-id/usb-Klipper_same54p20a_xxxxxxx` device).

Here are the config options that I used, all correct to the best of my knowledge:

image

Is there anything specific that you would like me to do to help troubleshoot at a lower level?

KevinOConnor:
If you configure the mcu in "usb to can bridge mode" then it will not show up in Linux as a serial device.  It should show up in Linux as a usb-to-canbus adapter.  You'll need to configure a `can0` interface in Linux, find the device canbus id, and configure the device using canbus in the printer.cfg file: https://www.klipper3d.org/CANBUS.html

-Kevin

ReXT3D:
> If you configure the mcu in "usb to can bridge mode" then it will not show up in Linux as a serial device. It should show up in Linux as a usb-to-canbus adapter. You'll need to configure a `can0` interface in Linux, find the device canbus_uuid, and configure the device using canbus in the printer.cfg file: https://www.klipper3d.org/CANBUS.html
> 
> -Kevin

Thanks for the clarification Kevin.  I misunderstood the implementation and assumed that the MCU will continue connecting to the host via serial over USB and its CAN interface will be available at the same time as CAN over USB.  Will test and report.

ReXT3D:
> > I also assume that the Mini cannot be re-flashed with new Klipper binary over the CAN bridge?
> 
> I don't know enough about the mini to help with flashing. In general, there is currently no fully automated "make flash" for "usb to can bridge" mcus. It is possible to put the board into bootloader mode with flash_can.py - https://klipper.discourse.group/t/octopus-pro-canboot-can-bus-bridge/3734/22
> 
> -Kevin

Thanks for the flashing info and the link Kevin, I will review.

I can also now happily confirm that the native Duet3 Mini 5+ CAN interface is working beautifully on my (messy) test bed a.k.a. WIP Voron 2.4:

IMG_1273

I have not done any longer term robustness testing, but EBB42 works beautifully at 1000000 bus speed including repeated `MEASURE_AXES_NOISE` requests with no reported CAN errors or anomalies.

Thanks again for incorporating this!

KevinOConnor:
Thanks for testing.  I committed this PR.  I also updated the duet3 mini config to recommend compiling for the same54p20 (commit eb9cda91).

-Kevin

PR #5807, opened 2022-10-05
Add SKR2, SKR3 and Creality 4.2.2/7 to spi_flash

JamesH1978:
As discussed with Arksine, he has created new working settings for the Creality 4.2.x boards as the original 4.2.7 entry did not work, plus addition of SKR2 and SKR3. Tested on 4.2.2 and SKR2 and SKR3EZ.

These are using the new skip_verify functionality that was recently merged due to them using SDIO

Also removed a double definition for `monster8` it was in main definitions and aliased.

### DIFF
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -87,6 +87,26 @@ BOARD_DEFS = {
         'spi_bus': 'spi3a',
         'cs_pin': 'PA15'
     },
+    'btt-skr-2': {
+        'mcu': "stm32f407xx",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        'cs_pin': "PC11",
+        'skip_verify': True
+    },
+    'btt-skr-3': {
+        'mcu': 'stm32h743xx',
+        'spi_bus': 'spi3a',
+        'cs_pin': 'PA15',
+        'skip_verify': True
+    },
+    'creality-v4.2.2': {
+        'mcu': "stm32f103xe",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        'cs_pin': "PC11",
+        'skip_verify': True
+    },
     'monster8': {
         'mcu': "stm32f407xx",
         'spi_bus': "spi3a",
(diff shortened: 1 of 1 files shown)


scripts/spi_flash/board_defs.py
```diff
@@ -87,6 +87,26 @@
         'spi_bus': 'spi3a',
         'cs_pin': 'PA15'
     },
+    'btt-skr-2': {
+        'mcu': "stm32f407xx",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        'cs_pin': "PC11",
+        'skip_verify': True
+    },
```
Arksine, line 96: The `btt-skr-2` looks like it could be added as an alias for `btt-octopus-f407-v1`.

scripts/spi_flash/board_defs.py
```diff
@@ -123,7 +143,7 @@
     'btt-skr-mini-mz': BOARD_DEFS['btt-skr-mini'],
     'btt-skr-e3-dip': BOARD_DEFS['btt-skr-mini'],
     'btt002-v1': BOARD_DEFS['btt-skr-mini'],
-    'creality-v4.2.7': BOARD_DEFS['btt-skr-mini'],
+    'creality-v4.2.7': BOARD_DEFS['creality-v4.2.2'],
```
Arksine, line 146: We should be able to remove this since it has a board definition.

Arksine:
Thanks.  The additions look good, I just had some minor observations that I added in the review.

scripts/spi_flash/board_defs.py
JamesH1978: Noted, will change that to be an alias

scripts/spi_flash/board_defs.py
JamesH1978: I removed the original 4.2.7 definition and replaced it with a 4.2.2 and then created an alias for 4.2.7

Arksine (approved):

KevinOConnor:
Thanks.

-Kevin

adelyser:
This doesn't work on my SKR3 EZ. I had to define the board pins as follows for it to work:

```
    'btt-skr-3': {
        'mcu': 'stm32h743xx',
        'spi_bus': 'swspi',
        'spi_pins': "PC8,PD2,PC12",
        'cs_pin': 'PC11',
        'skip_verify': True
    },
```

I'm curious how you got it to work using `spi3a`

PR #5812, opened 2022-10-07
config: Add BigTreeTech Manta M4P & M8P board

bigtreetech:

### DIFF
--- a/config/generic-bigtreetech-manta-m4p.cfg
+++ b/config/generic-bigtreetech-manta-m4p.cfg
@@ -1,0 +1,175 @@
+# This file contains common pin mappings for the BIGTREETECH Manta M4P
+# To use this config, the firmware should be compiled for the
+# STM32G0B1 with a "8KiB bootloader" and USB communication.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[stepper_x]
+step_pin: PC6
+dir_pin: PA14
+enable_pin: !PC7
+microsteps: 16
+rotation_distance: 40
+endstop_pin: ^PC0
+position_endstop: 0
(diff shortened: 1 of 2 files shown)


JamesH1978:
Can I ask why you are proposing removing flashing instructions from the SKR Mini E3 v3 file? These are still perfectly valid and as far as I know there is no other way to flash it, at least for the initial flash.

Thanks
James

bigtreetech:
> Can I ask why you are proposing removing flashing instructions from the SKR Mini E3 v3 file? These are still perfectly valid and as far as I know there is no other way to flash it, at least for the initial flash.
> 
> Thanks James

According to the instructions [here](https://github.com/Klipper3d/klipper/pull/5596) 
From my personal point of view `make flash FLASH_DEVICE=/dev/serial/by-id/usb-Klipper_stm32g0b1xx` DFU updates are valid for `stm32g0b1`
If `removing flashing instructions from the SKR Mini E3 v3` will cause misunderstanding, I can restore it.

JamesH1978:
Yes that option may be valid, but that only works once there has been an initial Klipper flash. For subsequent flashes it is perfectly valid. I do not believe this board ships with a default klipper flash therefore a klipper by-id will not be available. I am not entirely sure if the dfu flash will work using the default marlin by-id, or if a klipper flash is a prerequisite for this to work. I do not have this board to test. 

Thanks
James

bigtreetech:
Yes, It need klipper running before use DFU.  I will restore it now

KevinOConnor:
Thanks.  I committed this change and also added the two config files to the regression tests (commit 2d4ce1b9).

-Kevin

PR #5818, opened 2022-10-08
Config: Add Duet3 6hc config

lukeslaboratory:
Module: Config for Duet3 6HC

Signed-off-by: Luke Ashley <luke@lukeslabonline.com>


KevinOConnor:
Thanks. I think it is fine to add this printer config file, but it will need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html . At a minimum point 7, point 7b, point 7d, and point 6a.

-Kevin

PR #5820, opened 2022-10-10
Updated Voron 2 config

FatalBulletHit:


FatalBulletHit:
Fails by design (config needs to be adjusted by user)

PR #5821, opened 2022-10-10
toolhead: Capture current junction_deviation in a Move class

dmbutyugin:
If a maximum acceleration is changed between two consecutive moves, this allows to correctly compute the junction velocity between them.

### DIFF
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -17,6 +17,7 @@ class Move:
         self.start_pos = tuple(start_pos)
         self.end_pos = tuple(end_pos)
         self.accel = toolhead.max_accel
+        self.junction_deviation = toolhead.junction_deviation
         self.timing_callbacks = []
         velocity = min(speed, toolhead.max_velocity)
         self.is_kinematic_move = True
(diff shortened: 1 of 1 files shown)


dmbutyugin:
A sample GCode that demonstrates this problem is as follows:
```
SET_KINEMATIC_POSITION X=0 Y=0 Z=10
SET_VELOCITY_LIMIT ACCEL=1000 ACCEL_TO_DECEL=500
G1 X100 Y0 F6000
SET_VELOCITY_LIMIT ACCEL=10000 ACCEL_TO_DECEL=5000
G1 X100 Y100 F6000
```
in this case, the mainline Klipper computes the junction velocity approx equal to 1.58 mm/sec (assuming scv = 5 mm/sec). However, if there was no second `SET_VELOCITY_LIMIT` command, it would be 5 mm/sec. With this fix, the cornering velocity will be computed as 5 mm/sec in this case. I think this is a more appropriate behavior in this case.

KevinOConnor:
Thanks.  I agree there is a "corner case" here.  It's not clear to me that we need to "fix" this, but I also don't see any issues with your proposed code.  I guess I'll give a few days to see if there are any other comments and otherwise look to commit.

Cheers,
-Kevin

dmbutyugin:
Than you, Kevin. I stumbled upon this bug when making a different test for input shaping - if two gcode moves are collinear, but the second one bumps an acceleration to a very high value, then even the junction velocity for a straight line junction drops considerably. Also, some folks doing dynamic acceleration control may experience unexpected issues. And since the fix is simple and straightforward, I see no reason not to fix it.

KevinOConnor:
Makes sense.  I committed this change.

Thanks.
-Kevin

PR #5822, opened 2022-10-11
Add config file for the V4 version of the GT2560 boards

Keybored02:
The current generic-gt2560.cfg file is based on revision 3.0 of the board in question, which ceased production long ago. V4.0 is among the newest version to be packed in Geeetech products. V4.1 is theoretically also compatible, although I haven't been able to test it. 
Signed-off-by: Matteo Parenti, matteoparenti02@gmail.com

### DIFF
--- a/config/generic-gt2560-v4 .cfg
+++ b/config/generic-gt2560-v4 .cfg
@@ -15,7 +15,7 @@ max_z_velocity: 20
 max_z_accel: 500
 
 #[bltouch]
-#sensor_pin: ^PC7 #^PC7 default for z_min port,  ^PE7 for Z1-
+#sensor_pin: ^PC7              #^PC7 default for z_min port,  ^PE7 for Z1-
 #control_pin: PB5
 
 [stepper_x]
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks. I think it is fine to add this printer config file, but it will need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html . At a minimum point 3, point 4, point 6a, point 7, point 7e.

Can you explain the unusual `step_pulse_duration`?  What driver is it using?  Probably worth adding a comment to the config.

The commit will also need to have a signed-off-by line - https://www.klipper3d.org/CONTRIBUTING.html

The config file should be named: `generic-gt2560-v4.cfg`

Cheers,
-Kevin

Keybored02:
Hi,
Thanks for feedback!

`step_pulse_duration` was changed as the stock GEEETECH A4988 drivers were losing steps at more sustained speeds. This might not happen with different drivers, therefore I changed it back to the default value and added a comment to alert the user.

I also reverted back `rotation_distance` to the value indicated in `generic-gt2560.config` as the value I was using was intended for a different extruder. I'm not sure how to behave with a `generic` configuration and such variable values, so I guessed the stock one was the better choice. 

I did not comment out [display] as it is included in `generic-gt2560.config`. I did add a comment to indicate which display was it meant for.  

I commented out `[bltouch]`.

I renamed the file to `generic-gt2560-v4.cfg`.

I should have ticked all boxes, let me know if you need anything else modified/formatted.

Cheers!
Matteo

PR #5827, opened 2022-10-13
board_defs.py Update to fix SKR-3 config

JamesH1978:
it seems I made an error with my code for the SKR 3, and I copied the code from the wrong host , this PR fixes that. I have just retested with the right code and works as expected, this is confirmed by @adelyser who brought the issue to my attention.


KevinOConnor:
Thanks.

-Kevin

PR #5830, opened 2022-10-14
config: create config for Geeetech 301

konsumverweigerer:
Create a working config for the Geeetech 301 printer with an addition to flash_usb.py to use stm32flash

Signed-off-by: Peter Gruber <gruberp@googlemail.com>


konsumverweigerer:
The stm32flash uses the USB connection to the board. The board contains an onboard  usb-to-serial and is connected to the serial interface of the STM32F103VET6. All communication is done via this USB.

KevinOConnor:
Thanks.  The config looks fine (though it would be preferable to state "enable usart(on pins x and y)" in the comments as opposed to saying "disable usb".

I don't agree with adding stm32flash support to flash_usb.py .

Cheers,
-Kevin

KevinOConnor:
Thanks.  I merged this commit along with 6be0ba78 (in keeping with point 6b at https://www.klipper3d.org/Example_Configs.html ).

However, after I merged this, I realized the config should have a year in the filename (the approximate date the printer was purchased) as mentioned in point 2 of the above link.

Do you know what the approximate year of the printer is?

-Kevin

konsumverweigerer:
> Do you know what the approximate year of the printer is?

Its the 2019 version

KevinOConnor:
Thanks.  Commit d883c57d.

-Kevin

MCTRACO:
Hi, when flash through the onboard usb the firmware  i am getting error: Failed to init device.
I am using a jumper on boot0 and pressing the rest button but it is still not working
Anything im doing wrong?

PR #5833, opened 2022-10-15
config: Ender 3 s1 pro Fixed possible interference between y axis roller and belt tensioner.

jumika:
I've noticed that the y axis roller crashes into the belt tensioner housing during the bed probe routine on the S1 pro. Therefore i've added comments about limiting the y axis max position and adjusting the bed mesh section for this machine.


Sineos:
@Fisheiyy any comments?

Fisheiyy:
This might just be your machine, The S1 and S1 Pro are identical size wise, I have had a couple people test with S1's and S1 Pro's and not have any issues, this could also be coming from tightening the bed rollers since it will move them a little bit to add grip.

PR #5834, opened 2022-10-16
config: Add legacy Duet boards

woheedev:
Added a basic config for the legacy duet boards (0.6/0.8.5)


KevinOConnor:
Thanks.  However, the example configs need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html (at a minimum point 4).  You will also need to provide a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

PR #5835, opened 2022-10-16
spi_flash: Fix definition for btt-skr-mini-e3-v3

nhchiu:
The original config gets an error: "SPIFlashError: Failed to Initialize SD Card." when flashing skr-mini-e3-v3 through UART.
Same result with several sd cards with different brands and capacity.
Screenshot 2022-10-17 010749

Changing spi_bus to swspi 	made it work.
Screenshot 2022-10-17 010425

Signed-off-by:  Nathan Chiu <nhchiu2009@gmail.com>

### DIFF
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -28,7 +28,8 @@ BOARD_DEFS = {
     },
     'btt-skr-mini-v3': {
         'mcu': "stm32g0b1xx",
-        'spi_bus': "spi1",
+        'spi_bus': "swspi",
+        'spi_pins': "PA6,PA7,PA5",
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Okay, but it will be necessary to find the root cause of the error.  Making changes without understanding the root cause increases the risk of introducing a regression for other users.

-Kevin

nhchiu:
Sorry. I've checked the STM32g0b1 pin schematic and the spi1 bus is using the same pins as the SWSPI pins I assigned.
But my knowledge about MCU is too limited so I couldn't figure out why using SWSPI instead makes it work.
I opened the PR wishing someone could take a look at it... Thanks.

PR #5837, opened 2022-10-17
adxl345: Apply correct scaling for X,Y and Z axes

dmbutyugin:
According to ADXL345/ADXL343 datasheets, at 3.3V supply voltage, which is most frequent in the various boards, the sensitivity of X and Y axes changes to 265 LSB/g from 256 LSB/g at 2.5V.

### DIFF
--- a/klippy/extras/adxl345.py
+++ b/klippy/extras/adxl345.py
@@ -24,7 +24,8 @@ ADXL345_DEV_ID = 0xe5
 SET_FIFO_CTL = 0x90
 
 FREEFALL_ACCEL = 9.80665 * 1000.
-SCALE = 0.0039 * FREEFALL_ACCEL # 3.9mg/LSB * Earth gravity in mm/s**2
+SCALE_XY = 0.003774 * FREEFALL_ACCEL # 1 / 265 (at 3.3V) mg/LSB
+SCALE_Z  = 0.003906 * FREEFALL_ACCEL # 1 / 256 (at 3.3V) mg/LSB
 
 Accel_Measurement = collections.namedtuple(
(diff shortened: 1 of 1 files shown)


freakydude:
Hi. 

I didn't not try your code yet, but it's sounds very logical to me. If it is mentioned in the specs explicitly, then there must be a differanction made. 

But something else prompted me to comment this. You are writing that the accuracy melted down from 5% to 1%, What would be, if we did it the other way around?

Let's say, we try to calibrate this individual ADXL345 module with the expected 9.81m/s² (or if someone knows it bettter, with the individual value). 

Pseudo Calibration Code:
1. For every axies
   1. Do one run with SCALE == 1, results into an unscaled acceleration. 
   2. Put the unscaled acceleration to 9.81m/s², reults into Calibrated-ScaleXY multiplier
   3. Use this "calibrated" SCALEDXY for further calculations and measuerements. 

Does this make sense?

With best regards,
Frank

dmbutyugin:
@freakydude,

Sure, it is technically possible. One would have to put an accelerometer in 6 possible orientations (2 opposing directions for each of the 3 axis) or better even more, run measurements for a dozen of seconds or more, and then use the averaged data for each measurement to determine the correct offset and scaling factors (e.g. using least-squares method). That is quite involved though, so I doubt many would do it. The benefits are also unclear at this time (having the same scale on each axis, even if the scale is incorrect, is a bit different story).

KevinOConnor:
Thanks.  Nice catch.  Looks good to me.  I'll give a couple of days to see if there are any other comments, but otherwise look to commit.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #5838, opened 2022-10-17
stm32g0: Expand stm32g0 family

leptun:
Added stm32g070, stm32g071 and stm32g0b0 support.

This PR should make sure that the stm32g0b0 uses TIM3 instead of TIM2 (eg, see BTT Manta M4P which might have a stm32g0b0ret6 where TIM2 doesn't exist).
I also expanded the support to stm32g070 and stm32g071 since those were the only nucleo-g0 boards I found in stock at the moment. I'm still waiting for them to be delivered, so I haven't tested this PR yet unfortunately. I'm submitting it early so that I may get some feedback on the code in advance (marked as draft for now). If anyone has a board with stm32g0b0 and can test this, I would be grateful.


leptun:
I've checked and it works on STM32G070RB and STM32G071RB. Can't really check the G0Bx variants since I can't get my hands on such a board.

KevinOConnor:
Thanks.  In general, I'm fine with adding support for additional boards like this.  However, this commit adds several "ifdefs" to the code - some of which look incorrect (eg, the stm32f0_i2c.c and stm32f0_serial.c).  I wonder if there is a way to do this with less ifdefs.

Separately, will the stm32g0b1 binary run unmodified on an stm32g071 chip?

-Kevin

leptun:
> Separately, will the stm32g0b1 binary run unmodified on an stm32g071 chip?

It will not because of the different SRAM size. But otherwise it should work as long as the peripherals used exist also there (some don't such as CAN and some I2C)

> I wonder if there is a way to do this with less ifdefs.

I'm not sure how I could do it with less ifdefs. In any case, I've been using the stm32g071rb as my X/Y axis microcontroller for the past week and I haven't had issues with that. I'm using it over UART (the default nucleo board interface)

leptun:
@KevinOConnor I've rebased the changes to the latest master. I'm still not entirely sure what I need to change so that this PR can get merged.
> However, this commit adds several "ifdefs" to the code - some of which look incorrect (eg, the stm32f0_i2c.c and stm32f0_serial.c). I wonder if there is a way to do this with less ifdefs.

I'm not entirely sure what I could do about all these ifdefs. It's disappointing that the chips differ in this way, but I don't see a way to have those features besides using ifdefs. The only change I can think of is instead of doing for example `#ifdef USB_BASE` I use `#if CONFIG_xxx`, but I'm not sure if that will result in cleaner code.

KevinOConnor:
Some of the changes here are definitely incorrect - for example, a change to the `i2c_bus` ifdefs in stm32f0_i2c.c must have a corresponding change to the ifdefs in the `DECL_ENUMERATION` part of that file (and the enumeration number must exactly match the `i2c_bus` array offset).

> > Separately, will the stm32g0b1 binary run unmodified on an stm32g071 chip?

> It will not because of the different SRAM size.

If you set the ram size of the stm32g0 build to 0x9000 does it then work?  Klipper never uses more than about 16KiB of ram, so we can reduce the size of the existing stm32g0 definition if that helps the code run on more chips.

-Kevin

KevinOConnor:
Yes - certainly the g0x0 variants will need different support.  For some reason I thought most of the ifdefs were due to the g0b1 vs g071 - is that not the case?

-Kevin

leptun:
> Yes - certainly the g0x0 variants will need different support. For some reason I thought most of the ifdefs were due to the g0b1 vs g071 - is that not the case?

The main reason for the huge number of defines is because of the G07x series. The G0B0 alone is not that hard to implement alongside the G0B1, but the G07x series requires quite a bit more defines for them to work.

I'm still torn whether I should reduce the ram of the G0Bx series. It's not like you would want to use a G0Bx series binary on the G07x series microcontrollers, so unifying the memory usage doesn't yield any real improvement in code quality. It just makes a smaller diff, but at the cost of having a misleading RAM_SIZE config value. As for 0x0 vs 0x1 series, those are not interchangeable at all due to the timer code. A 0x0 binary should work on the 0x1, but not the other way around. And even so, it is preferable to have a 32bit system timer, not a software extended 16bit timer.
I've pushed an update for the I2C code. I hope this is correct now. I also had to rebase on top of the STM32G4 changes. Please let me know what else I should change to push this PR forward.

Speaking of the STM32G4, I also pushed a fix for the SWD accidental disable on that mcu family. However, I'm not entirely sure that this was a problem in the first place since I don't have that G4 model MCU avaliable to test. All things considered, the situation looked very similar to the STM32G0, so I ported the fix to that MCU family as well.

KevinOConnor:
Okay, thanks for the clarification.

> The G0B0 alone is not that hard to implement alongside the G0B1

Would it be possible to break out the G0B0 code into a separate commit (and/or PR) then?  I know that's more work, but it would make review easier.  I'm still a little leery about the ifdefs (mainly because I don't want to introduce a regression on one of the other stm32 targets).  I don't see an issue with adding ifdefs if that is the best way to support the 07x chips, but it would be easier to review them if it's just those chips (and not 0bx).

-Kevin

leptun:
Sure. I'll make a separate PR for that in the coming days. Should I include the G0 and G4 SWD bugfix also?

KevinOConnor:
Yes - please make the swd fix separate and ping @mattthebaker on it.

-Kevin

mattthebaker:
The G4 SWD change looks fine, that would explain why I needed to make `openocd` hold the device in reset before initiating programming. I didn't actually need to use the debug interface for debug or I would have noticed this. Future boards are less likely to have resetb wired to the debug header so this should save someone some trouble.

The i2c `DECL` statements for L4 look like they have different ordering than the structure in the latest commit.

leptun:
@mattthebaker Thanks.
@KevinOConnor I think the PR is ready again

leptun:
I've pushed a commit addressing the requests. I did not have time to test them however, but I think it should be fine as it still compiles.

KevinOConnor:
Thanks.

-Kevin

PR #5841, opened 2022-10-18
stm32f446: Allows to select clock frequency by menuconfig

marbocub:
A few of TRONXY STM32F446 controller including XY-3 SE is driven at 120MHz by the stock firmware.
The klipper firmware drives STM32F446 at 180MHz which is maximum frequency then didn't work stable with the XY-3 SE. It often causes "hard fault" interrupts and then restarted by IWDG.

Allows the user to select the clock frequency of STM32F446 in menuconfig. Selection of 120MHz, 144MHz and 168MHz are for setting up PLLQ to 48MHz.

Signed-off-by: Mitsunori YOSHIDA <marbocub@gmail.com>

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -139,15 +139,28 @@ config MCU
     default "stm32h750xx" if MACH_STM32H750
     default "stm32l412xx" if MACH_STM32L412
 
+choice
+    prompt "MPU Frequency" if MACH_STM32F446
+    config CLOCK_FREQ_180MHZ
+        bool "180MHz"
+    config CLOCK_FREQ_168MHZ
+        bool "168MHz"
+    config CLOCK_FREQ_144MHZ
+        bool "144MHz"
+    config CLOCK_FREQ_120MHZ
+        bool "120MHz"
+endchoice
+
 config CLOCK_FREQ
     int
     default 48000000 if MACH_STM32F0
     default 64000000 if MACH_STM32F103 && STM32_CLOCK_REF_INTERNAL
     default 72000000 if MACH_STM32F103
-    default 120000000 if MACH_STM32F207
+    default 120000000 if MACH_STM32F207 || CLOCK_FREQ_120MHZ
     default 84000000 if MACH_STM32F401
-    default 168000000 if MACH_STM32F4x5
-    default 180000000 if MACH_STM32F446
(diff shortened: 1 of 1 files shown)


marbocub:
Hi Kevin
Thank you for your comment.

First, the stock frequency of the XY-3's STM32F446 as lower as others is maybe thermal issues because the controller mounted upside down in the box under the heated bed.

Second, I checked clock frequencies and clock selectors by the Klipper's RCC register values before PR.
Below is my understanding.

The main PLL changes the three frequencies of PLLCLK, PLLQ and PLLR. Then I examined the impact of changing the frequency of PLLCLK by three points: peripheral frequencies, PLLQ, and PLLR.

(1) Peripheral dividers do not need change, even to keep the APB1 frequency below 45MHz of the spec but still as high as possible. Therefore, the APB1 and APB2 clocks are changed proportionately changing PLLCLK.
(2) PLLQ is used to SDIO and USB48MHz through PLL48CLK, and is better equal to 48MHz.
(3) PLLR is only used to SPDIFRX-IN and is no problem.

The real frequency impact of changing the PLLCLK from 180MHz to 120MHz shows in the table below.

|  PLLCLK | 180MHz | 120MHz |
| --- | --- | --- |
| PLLQ | 51.43MHz | 48MHz |
| PLLR | 60MHz | 60MHz |
| APB1/2 | 45MHz | 30MHz |
| (AHB) | 180MHz | 120MHz |
| (MCO2) | 180MHz | 120MHz |

The PLLQ is to be better and the PLLR is the same. 
The APB1/2 are slowing down, however, I think it is no real problem because get_pclock_frequency() returns the correct frequency of the APB1/2.


I would appreciate it if you could reconsider including an option to change the Klipper firmware's STM32F446 frequency to 120MHz for machines whose stock CPU frequency has been lowered due to thermal issues.

KevinOConnor:
Thanks.  Chip thermal issues seems highly unlikely to me, as anything heating the chip to that extreme will certainly overheat the chip at 120mhz as well.

If it is a thermal issue, it should be easy to confirm - run the pristine kliipper code, enable a `[temperature_sensor]` config section using a `temperature_mcu` sensor, and run a print until a fault is raised.  Then repeat with the mcu code modified to 120mhz.  Post both klipper logs here.  That will allow easy analysis of the temperature curves with different mcu frequencies.

Cheers,
-Kevin

PR #5842, opened 2022-10-19
neopixel: update max led count

Nailig:
Increase sanity check on max nof leds. I lost my sanity and have 194 RGBW leds (= 776 nof addresses)


julianschill:
I would vote for removing this check completely. I could run succesfully 700 Neopixels on one MCU.

KevinOConnor:
Thanks.  I don't have a particular preference on the maximum neopixel chain size.  However, at very large values it can result in command processing delays within the mcu and could consume too much memory in some mcus.  So, making it unlimited isn't a good option.

At the current limit of 500 the theoretical minimum transmit time is 0.625ms (1250ns * 500).  It should be okay to double that, but much larger could result in obscure processing issues.

-Kevin

PR #5844, opened 2022-10-20
rp2040: Added canboot bootloader support for rp2040.

sh83:
This PR add support for the canboot bootloader on rp2040 mcu. The corresponding canboot pr is here https://github.com/Arksine/CanBoot/pull/45
The most changes are trivial.
The makefile was modified to produce binary instead of uf2  if bootloader is used.. 
boot2 generation is skipped if bootloader is used. 
Option was added to the KConfig.
The nontrivial part of the PR - bootloader offset. On RP2040 first 256 bytes of the flash are used for boot2, so user bootloader is placed at the strange offset.

Signed-off-by: Alex Malishev <malishev@gmail.com>

### DIFF
--- a/src/atsam/Kconfig
+++ b/src/atsam/Kconfig
@@ -70,6 +70,10 @@ config FLASH_START
     default 0x400000 if MACH_SAM4 || MACH_SAME70
     default 0x80000
 
+config BOOTLOADER_START
+   hex
+   default 0x0
+
 config FLASH_SIZE
     hex
     default 0x80000
--- a/src/atsamd/Kconfig
+++ b/src/atsamd/Kconfig
@@ -154,6 +154,10 @@ config FLASH_START
     default 0x2000 if FLASH_START_2000
     default 0x0000
 
+config BOOTLOADER_START
+   hex
+   default 0x0
+
 choice
     prompt "Communication interface"
     config ATSAMD_USB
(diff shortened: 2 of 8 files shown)


KevinOConnor:
On a closer look, I think I misunderstood the changes you made to armcm_reset.c, thus the diff above isn't complete.  I'm leery of adding board specific code to the src/generic/ files.  I think we'll have to refactor the build so that the absolute flash start is defined in each board Kconfig so that the rp2040 can define its custom flash start address.

-Kevin

sh83:
Thank you for your review.
I have added CONFIG_BOOTLOADER_START variable. It initialized in Kconfig to the same value as before (CONFIG_FLASH_SIZE & 0xFF000000). For rp2040 it have custom value, as it should.

sh83:
Should I change or fix something in this PR?

KevinOConnor:
Thanks.  I'd prefer to avoid the "ifdefs" in the makefile if possible.  How about PR #5874 as an alternative?

-Kevin

sh83:
@KevinOConnor , thank you. You PR looks correct to me. I will test it on real hw with canboot and report later.

sh83:
Well, alternative PR is tested and interacts well with CanBoot. I assume this one could be closed.

PR #5846, opened 2022-10-22
Use consistent memory position/size on stm32h743

KevinOConnor:
Since Klipper never uses more than about 16KiB of ram, there isn't a gain to using a different memory map on the stm32h743 chip.  Make that chip the same as stm32h750 and other stm32 chips.

@D4SK  @adelyser - fyi.

-Kevin

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -166,7 +166,6 @@ config FLASH_SIZE
 
 config RAM_START
     hex
-    default 0x24000000 if MACH_STM32H743
     default 0x20000000
 
 config RAM_SIZE
@@ -181,8 +180,7 @@ config RAM_SIZE
     default 0x10000 if MACH_STM32F401
     default 0x20000 if MACH_STM32F4x5 || MACH_STM32F446
     default 0x24000 if MACH_STM32G0B1
-    default 0x20000 if MACH_STM32H750
-    default 0x80000 if MACH_STM32H743
+    default 0x20000 if MACH_STM32H7
 
 config STACK_SIZE
     int


adelyser:
I had issues using that address space when I was developing on the board. Since @D4SK has tested the STM32H750 already, I would prefer to keep the STM32H743 as it was.

KevinOConnor:
I'd really prefer to keep all the mcus as much the same as possible.  It certainly seems odd that the h743 is the only stm32 chip that uses a different ram address.  That chip has an identical address map as the h750, which can use the standard 0x20000000 address.  If there is a reason the h743 can't use that address I think we definitely should find out why, as it might be a problem that impacts other chips.

-Kevin

bigtreetech:
Just my experience:
When I was debugging Marlin earlier, H743 SDMMC1 was connected to the AXI bus, and the buffer for reading SDMMC1 data must be on AXI SRAM(0x24000000), otherwise data errors will occur. I'm not sure whether there are similar requirements in the peripherals used by Klipper
image

Hundsbuah:
Ill pull the PR later and print a xyz cube. Ill report back

adelyser:
I pulled the PR and it seems to work on my SKR3 and the SKR-SE-BX boards. I guess my issues earlier were due to other config problems. I'm ok with the change.

Hundsbuah:
printed successfully on stm32h743 skr3 - no problems so far

2022-10-31 19_53_24-Window

[klipper_pr_5846.zip](https://github.com/Klipper3d/klipper/files/9904002/klipper_pr_5846.zip)

KevinOConnor:

-Kevin

PR #5858, opened 2022-10-25
Feature/mixing extruder

konsumverweigerer:
Support for mixing (n-in 1-out) nozzles. It uses the sync to stepper/extruder and adjusts the rotation distance for the affected steppers.

### DIFF
--- a/config/sample-mixing-extruder.cfg
+++ b/config/sample-mixing-extruder.cfg
@@ -1,0 +1,82 @@
+# This file contains a configuration snippet for a printer using a
+# mixing extruder (3 in - 1 out). The 3 extruder motors should be defined
+# as extruder, extruder1 and extruder2
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[mixing_extruder]
+steppers: extruder,extruder1,extruder2
+
+
+[gcode_macro M163]
+gcode:
+    SET_MIXING_EXTRUDER STEPPER={params.S} SCALE={params.P}
+
+[gcode_macro M164]
+gcode:
+    SAVE_MIXING_EXTRUDERS EXTRUDER={params.S|default(ACTIVE)}
+
+[gcode_macro M165]
+gcode:
+    SET_MIXING_EXTRUDER STEPPER=0 SCALE={params.A|default(0)|float}
+    SET_MIXING_EXTRUDER STEPPER=1 SCALE={params.B|default(0)|float}}
+    SET_MIXING_EXTRUDER STEPPER=2 SCALE={params.C|default(0)|float}}
+    SAVE_MIXING_EXTRUDERS EXTRUDER=ACTIVE
+
+[gcode_macro M166]
+gcode:
+    {% if params.A|float > 0 and params.Z|float > 0 and params.I|int >= 0 and params.J|int >= 0 %}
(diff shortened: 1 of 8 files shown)


konsumverweigerer:
Ok I will look into points 1, 2 and 4 ...

Regarding 3.: I need to poll the initial rotation distance on the steppers and for that I need to make a method call on the ExtruderStepper (I added the get/set_rotation_distance in extruder.py for that) and this was the way I saw to get to the ExtruderSteppers having only their names.

Regarding 6.: Thanks for the pointer. I hadn't seen the this move_transform possibility and with that I could add the position hook only when needed: when an gradient is active (you are right this is only needed for gradients)

Regarding 6.: The G1 is only a demo how flexible the basic command is (eg. the duet g-code flavor also supports a "G1 Ennn:nnn:nnn..." command: https://docs.duet3d.com/User_manual/Reference/Gcodes) and it make manual testing much easier

Regarding 7.: The extruder_name is the extruder which is controlled by the G0/G1 commands and the steppers are the steppers I sync with the extruder. An alternative solution would be that the first stepper in the list is the extruder by convention.

for the 8.: when I do some reorganization I will probably remove the multiple mixing_extruders altogether since they are only for convenience to hold some presets for mixings and that could also easily be done with gcode macros if required

KevinOConnor:
Thanks.

> I need to poll the initial rotation distance on the steppers and for that I need to make a method call on the ExtruderStepper (I added the get/set_rotation_distance in extruder.py for that) and this was the way I saw to get to the ExtruderSteppers having only their names.

I think a simpler way would be to add a `get_extruder_stepper()` method call to the extruder.py and extruder_stepper.py .

> The extruder_name is the extruder which is controlled by the G0/G1 commands and the steppers are the steppers I sync with the extruder. An alternative solution would be that the first stepper in the list is the extruder by convention.

I'm a little confused on the naming.  In Klipper today, an "extruder" is the combination of a "hotend heater" and an "extruder stepper".  Given that, wouldn't a "mixing extruder" be the combination of an "extruder" and n "extruder steppers"?  Why would a mixing extruder have multiple hotends?

-Kevin

konsumverweigerer:
> > I need to poll the initial rotation distance on the steppers and for that I need to make a method call on the ExtruderStepper (I added the get/set_rotation_distance in extruder.py for that) and this was the way I saw to get to the ExtruderSteppers having only their names.
> 
> I think a simpler way would be to add a `get_extruder_stepper()` method call to the extruder.py and extruder_stepper.py .

That would be a solution for the extruder.py . But for the extruder_stepper.py that would not help since they are not registered with the printer as far as I see... 

> > The extruder_name is the extruder which is controlled by the G0/G1 commands and the steppers are the steppers I sync with the extruder. An alternative solution would be that the first stepper in the list is the extruder by convention.
> 
> I'm a little confused on the naming.  In Klipper today, an "extruder" is the combination of a "hotend heater" and an "extruder stepper".  Given that, wouldn't a "mixing extruder" be the combination of an "extruder" and n "extruder steppers"?  Why would a mixing extruder have multiple hotends?

exacly so extruder_name is the one extruder (which is used for the sync_to_extruder call target) and the steppers list are extruder steppers (where I set the rotation distance); one of which may be that extruder. Possibly I could figure out the extruder_name from the toolhead ...

KevinOConnor:
> But for the extruder_stepper.py that would not help since they are not registered with the printer as far as I see...

All "extras" modules are registered in the `printer` object via the config name.  For example, `printer.lookup_object("extruder_stepper my_extra_stepper")`.

-Kevin

PR #5860, opened 2022-10-26
stm32: fix USART ORE status flag handling

leptun:
If an USART RX overrun happened on a stm32g0/f0/h7, the ORE flag would get set by hardware. This flag would also trigger an interrupt. The problem was that this flag was never cleared on these 3 mcu families since the ORE flag clear sequence is different to all of the older chips. As a result, the MCU would get stuck in the USART ISR and trigger a watchdog reset instead of just causing a communication error that can be recovered from.
Since the ORE flag is not used in any meaningful way anyway, it was disabled during the init sequence for the STM32 G0/F0/H7.

### DIFF
--- a/src/stm32/serial.c
+++ b/src/stm32/serial.c
@@ -57,8 +57,11 @@ void
 USARTx_IRQHandler(void)
 {
     uint32_t sr = USARTx->SR;
-    if (sr & (USART_SR_RXNE | USART_SR_ORE))
+    if (sr & (USART_SR_RXNE | USART_SR_ORE)) {
+        // The ORE flag is automatically cleared by reading SR, followed
+        // by reading DR.
         serial_rx_byte(USARTx->DR);
+    }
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Nice catch!  Patch looks fine, but as a minor comment I think it is preferable to explicitly set hardware registers where possible - that is use: `USARTx->CR3 = USART_CR3_OVRDIS`. (Also mentioned briefly in docs/Code_Overview.md .)

Separately, it seems we should merge stm32h7_serial.c into stm32f0_serial.c as they have basically the same code.  I'll look into that.

-Kevin

leptun:
@KevinOConnor I used `|=` since I didn't want to affect the other bits in the register. The datasheet says this register is initialized to 0, so it probably is safe to change it to `=`. I'll change this now.

leptun:
@KevinOConnor I updated the code.

KevinOConnor:
Thanks.

-Kevin

PR #5862, opened 2022-10-28
Mightyboard flash support

dockterj:


dockterj:
shouldn't have gone to upstream

PR #5869, opened 2022-10-30
config: Allow nozzle_diameter slightly > filament_diameter

sprior:
Allow for the configured nozzle_diameter to be slightly larger than the filament_diameter. An example of this is a 1.8mm CHT nozzle with 1.75mm filament which some have had success with.

Signed-off-by: Steve Prior <sprior@geekster.com>

### DIFF
--- a/klippy/kinematics/extruder.py
+++ b/klippy/kinematics/extruder.py
@@ -164,7 +164,7 @@ class PrinterExtruder:
         # Setup kinematic checks
         self.nozzle_diameter = config.getfloat('nozzle_diameter', above=0.)
         filament_diameter = config.getfloat(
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I'm a little concerned that this change will provide a confusing error message to users that incorrectly configure a filament_diameter - as the new message will return something like "Option 'filament_diameter' in section 'extruder' must have minimum of 1.6625" and users will have no idea where 1.6625 came from..

As mentioned by @dmbutyugin in the previous PR, it's easy enough to alter the nozzle_diameter for this uncommon case.

Cheers,
-Kevin

PR #5870, opened 2022-10-31
Use stm32f0_serial.c on stm32h7 chips

KevinOConnor:
The stm32h7 uses similar usart hardware as the stm32f0 and stm32g0 chips.  Use the same code implementation for all these chips.

This came up in PR #5860 - it's preferable to use the same code for these chips so that bug fixes get applied to all the chips.

@D4SK , @adelyser - FYI.

-Kevin


src/stm32/stm32f0_serial.c
```diff
@@ -16,30 +16,58 @@
...
-  #define USARTx_FUNCTION GPIO_FUNCTION(1)
+  #define USARTx_FUNCTION GPIO_FUNCTION(CONFIG_MACH_STM32H7 ? 7 : 1)
   #define USARTx USART2
   #define USARTx_IRQn USART2_IRQn
 #elif CONFIG_STM32_SERIAL_USART2_ALT_PA15_PA14
   DECL_CONSTANT_STR("RESERVE_PINS_serial", "PA15,PA14");
   #define GPIO_Rx GPIO('A', 15)
   #define GPIO_Tx GPIO('A', 14)
-  #define USARTx_FUNCTION GPIO_FUNCTION(1)
+  #define USARTx_FUNCTION GPIO_FUNCTION(CONFIG_MACH_STM32H7 ? 7 : 1)
```
adelyser, line 40: There isn't a PA14,PA15 serial connection on the STM32 H7, this config is dependent on the STM32F0 switch

KevinOConnor: Thanks.  I removed that part of the change.

-Kevin

D4SK:
I just tested this and it works great.

KevinOConnor:

-Kevin

PR #5874, opened 2022-11-03
Support CanBoot bootloader on rp2040

KevinOConnor:
This is an alternative to PR #5844 .

-Kevin

### DIFF
--- a/src/atsam/Kconfig
+++ b/src/atsam/Kconfig
@@ -74,6 +74,10 @@ config FLASH_SIZE
     hex
     default 0x80000
 
+config FLASH_BOOT_ADDRESS
+    hex
+    default 0x0
+
 config RAM_START
     hex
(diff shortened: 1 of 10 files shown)


sh83:
This PR was tested on rp2040 dev board with usb connection mode.
1. The standalone mode (without bootloader) klipper.uf2 flashed and booted well, create usb serial device.
2. The bootloader mode klipper.bin flashed with CanBoot master. It boots well, create usb serial device, with ``stty -F /dev/ttyACM0 1200`` command it reboot back to CanBoot. So I assume CanBoot detection is correct as well.
I have found no regression or issue.
Thank your, @KevinOConnor.

Undefining:
Tested this PR in combination with the SB2040 and a BTT Octopus as CAN bridge. Was able to flash klipper over CAN using canboot.

PR #5877, opened 2022-11-05
bed_mesh: profile fixes

### LINKED ISSUE #5875: Autogenerated Config Missing Keys
Hi,

After running BED_MESH_CALIBRATE PROFILE=default METHOD=automatic I'm getting an error saying the autogenerated config doesnt have `mesh_min` specified. Similarly, I'm getting an error saying `control` is not present in the autogenerated config for the `heater_bed` section.

This is similar to [#3085](https://github.com/Klipper3d/klipper/issues/3085) based on searching with `is:issue autogenerated `

```
Option 'control' in section 'heater_bed' must be specified

Printer is halted
```

Hardware:
Raspberry Pi 3B+
CR-10 Smart (Stock)
Config taken from a community post due to no template being available for this model. I'm willing to submit a PR for it after I get a decently working test print.

Latest log and config:
[klippy.log](https://github.com/Klipper3d/klipper/files/9932974/klippy.log)
[printer.txt](https://github.com/Klipper3d/klipper/files/9933018/printer.txt)

Arksine:
This makes a couple minor changes to the profile behavior for `bed_mesh`:

1) Require that a profile name be specified for the `PROFILE` parameter in `BED_MESH_CALIBRATE` and all parameters in `BED_MESH_PROFILE`.  This resolves #5875.

2) Introduce an `initial_profile` option.  This allows users to specify which profile is loaded at startup, or to bypass loading a profile at startup by specifying a profile name that doesn't exist (my preference is to set this to none).

### DIFF
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -399,6 +399,26 @@ is not desirable the _default_ profile can be removed as follows:
 Any other saved profile can be removed in the same fashion, replacing
 _default_ with the named profile you wish to remove.
 
+
+#### Loading the default profile
+
+Previous versions of `bed_mesh` always loaded the profile named _default_
+on startup if it was present.  This behavior has been removed in favor of
+allowing the user to determine when a profile is loaded.  If a user wishes to
+load the `default` profile it is recommended to add
+`BED_MESH_PROFILE LOAD=default` to either their `START_PRINT` macro or their
+slicer's "Start G-Code" configuration, whichever is applicable.
+
+Alternatively the old behavior of loading a profile at startup can be
+restored with a `[delayed_gcode]`:
+
+```ini
+[delayed_gcode bed_mesh_init]
+initial_duration: .01
+gcode:
+  BED_MESH_PROFILE LOAD=default
+```
+
 ### Output
 
 `BED_MESH_OUTPUT PGP=[0 | 1]`
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,11 @@ All dates in this document are approximate.
 
 ## Changes
 
+20230201:  The `[bed_mesh]` module no longer loads the `default` profile
+on startup.  It is recommended that users who use the `default` profile
+add `BED_MESH_PROFILE LOAD=default` to their `START_PRINT` macro (or
+to their slicer's "Start G-Code" configuration when applicable).
+
 20230103: It is now possible with the flash-sdcard.sh script to flash
 both variants of the Bigtreetech SKR-2, STM32F407 and STM32F429.
 This means that the original tag of btt-skr2 now has changed to either
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -129,7 +129,6 @@ class BedMesh:
     def handle_connect(self):
         self.toolhead = self.printer.lookup_object('toolhead')
         self.bmc.print_generated_points(logging.info)
-        self.pmgr.initialize()
     def set_mesh(self, mesh):
         if mesh is not None and self.fade_end != self.FADE_DISABLE:
             self.log_fade_complete = True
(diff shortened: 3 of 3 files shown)


Arksine:
Sounds good.  Yes, the `initial_profile` is really to allow users to select either a specific profile or skip loading a profile (this is my preference).  I agree that perhaps its better to disable loading the default profile.   If a user really wants to load one at startup they could use a `delayed_gcode` to do it, that is something I could note in `bed_mesh.md`.

KevinOConnor:
Okay, thanks.  I cherry-picked commit 5e34b450.

Separately, it seems a bit odd that one would use a delayed_gcode - I would think it would be simpler for a user to update their START_PRINT macro (or corresponding slicer setup).

Cheers,
-Kevin

Arksine:
Thanks.  I updated this PR to remove profile initialization.

> Separately, it seems a bit odd that one would use a delayed_gcode - I would think it would be simpler for a user to update their START_PRINT macro (or corresponding slicer setup).

Generally I agree, however I documented it in the event that a user wanted to restore the prior behavior.  There may be a situation where a user wants the profile active before the print has started, although in general practice I don't think that makes much sense.  I can remove that documentation  that if you prefer.

KevinOConnor:
Thanks.  Seems fine, but something to change after the next release.

-Kevin

KevinOConnor:
FYI, I agree with making this change.  I suspect it might cause a lot of pain for existing users though.  I wonder if there is some way to announce this change so it's better understood before we roll it out.

-Kevin

zellneralex:
Moonraker has integrated announcement and to my knowledge Klipper and moonraker are automatically subscribed.
That would reach at least most Fluidd/Mainsail users.

@Arksine can help to setup an Klipper announcement, he has done that for moonraker in the past.

Arksine:
Sure, we can do an announcement.  [Moonlight](https://github.com/arksine/moonlight) is already set up to watch Klipper's issue tracker.  All Kevin would need to do is create an issue in this repo and tag it with an `announcement` label.  Within an hour the announcement should be distributed to all running instances of Moonraker with an internet connection.

Additionally I started looking into pulling announcements from the discourse using their API.  I believe its possible without credentials, so its something I'm willing to look at adding for future announcements related to Klipper.

KevinOConnor:
That makes sense.  We could make an announcement now and schedule the change on something like Feb 1st?

The announcement could be something like:

An upcoming version of Klipper will no longer load a default bed_mesh configuration at startup.  If you are using a `bed_mesh` configuration today then it is recommended to add an explicit `BED_MESH_PROFILE LOAD=default` command to your START_PRINT macro (or equivalent slicer start commands).  This change to Klipper is planned for February 1st 2023.

Thoughts?

-Kevin

Arksine:
Sounds good to me.  The of title of the issue will be emphasized, and the announcement will include a link that takes them directly to the issue.  When you close the issue it will be dropped from the feed.

I went ahead and fixed up `config_changed.md`, forward dating it to Feb 1 2023.

KevinOConnor:
Okay - I created the issue (#5983).

Separately, for what it is worth, I'd recommend having Config_Changes.md and bed_mesh documents point users to updating START_PRINT (or slicer start).  I don't feel strongly about it though, so up to you.

-Kevin

Arksine:
Looks good!

image

I went ahead and made those changes to the docs.  I agree that loading the profile in the start macro is the best way to deal with it.  I did leave a note in the Bed Mesh docs on how to restore the old behavior with a `[delayed_gcode]`, but recommended that they just load the profile before a print.

djessup:
Just confirming this appeared in Mainsail for me just now - the system works!

qwewer0:
On Fluidd it is a bit cut down, but its there.

image

pedrolamas:
> On Fluidd it is a bit cut down, but its there.

Pressing "more information" button should send to the full article... I wonder if I should just allow those notifications to be something like 10 lines or so... 🤔

Sineos:
I'd even go a step further since I assume that a lot of users readily ignore the little "1": Priority announcement should be shown (full-screen) automatically until actively cleared

KevinOConnor:
Thanks.

-Kevin

PR #2787, opened 2020-04-26
linux: userspace GPIO control via libgpiod

andryblack:
Allow use host GPIO pins for non-realtime purposes. In my setup HD44780 display connected directly to Raspberry-pi zero. Multi-mcu configuration with linux host-mcu and stm32f4 custom board for printing. I use my own lightweight print server, and now use Klipper. For build with GPIO support need libgpiod headers and library. (libgpiod-dev on Debian). It is early implementation: only one gpiochip, dnt implement pullup/pulldown for input (dnt supported by 4.x kernels). It is can be rewritten directly to gpiod ioctrls for remove library dependency.
At compare to PR #2781 use gpiod interface instead of (deprecated) sysfs gpio interface get code cleanly, faster and possibly extend to digital-buses (bulk gpios). At near future I work on implement events on input pins, encoder too connected to host board at my setup.

related issues: #1814, #925

Update: encoder worked with hardware pull-up resistors.

PR #5878, opened 2022-11-05
linux_mcu:  use correct gpio bias flags

Arksine:
This replaces stale gpiod bias flags introduced in the original pull request, #2787.   According to [this comment](https://github.com/Klipper3d/klipper/pull/2787#issuecomment-623004661) the decision was made to remove the libgpiod dependency and use ioctl's directly, however the bias flags were not updated.


KevinOConnor:
Thanks.

-Kevin

PR #5879, opened 2022-11-05
safe_z_home: Fix final z hop to use relative z coordinates

redstone99:
After a G28 z-axis homing, there is a final z hop. It was hoping to height z_hop as an absolute z height rather than relative.

If the z-axis home leaves the head at a z height higher than z_hop, e.g. because you were using a probe to do z homing, this meant that it was z hopping in the negative z direction, which could result in crashing the toolhead.

Signed-off-by: Joshua Redstone <redstone@gmail.com>

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,12 @@ All dates in this document are approximate.
 
 ## Changes
 
+20221114: Previously, with safe_z_home, it was possible that the
+z_hop after the g28 homing would go in the negative z direction.
+Now, a z_hop is performed after g28 only if it results in a positive
+hop, mirroring the behavior of the z_hop the occurs before
+the g28 homing.
+
 20220616: It was previously possible to flash an rp2040 in bootloader
 mode by running `make flash FLASH_DEVICE=first`. The equivalent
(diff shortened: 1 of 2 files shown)


FHeilmann:
Good change but I think your description is backwards. You changed the zhop from an absolute hop to `self.zhop` (which has the issues you mention) to a relative one to `current_z + self.zhop`

Just FYI :)

redstone99:
Thanks for the comments. I added a new commit with hopefully a clearer commit message.

KevinOConnor:
Thanks.  I'm fine with the change, but this looks like a change in behaviour that should be documented in docs/Config_Changes.md .

-Kevin

redstone99:
Thanks for the suggestions. I added a change description to Config_Changes.md.
Also, looking at the code more closely, I think it's more consistent to have the z hop after the g28 behave in the same way as the zhop before the g28.  That is, hopping to self.z_hop as an absolute height, but only if it's higher than the current height. This behavior is also consistent with the documentation in Config_Reference.

KevinOConnor:
Thanks.  I committed this along with 9727f00e on top.

-Kevin

PR #5881, opened 2022-11-06
statistics: Extend system stats with unixtime

denis-efimov:
For the moment, it is not possible to make macros based on the host's system time, such as bed leveling once a day or turning things on and off based on real time.

Extending system stats with unixtime, we bring the possibilities of macros to a whole new level.

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -380,7 +380,7 @@ The following information is available in
 
 The following information is available in the `system_stats` object
 (this object is always available):
-- `sysload`, `cputime`, `memavail`: Information on the host operating
+- `sysload`, `cputime`, `unixtime`, `memavail`: Information on the host operating
   system and process load.
 
 ## temperature sensors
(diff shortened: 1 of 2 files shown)


30350n:
Hey! I'm currently looking into creating an air filter (nevermore) timing macro to track the time the filter has been used and this is exactly the functionality I'm looking for. The change seems relatively small and straight forward, could this get reviewed and merged?

pedrolamas:
I think you would be better adding this as a Jinja2 function (check `rawparams` implementation, you can add a function on that level) than as part of a status object.

The reason be that a change in a status object will be pushed to Moonraker, and that will trigger a bunch of other stuff that seem unnecessary to me...

Arksine:
The `get_status()` methods used to update printer status are also called when the macro is evaluated.  There would be very little difference between the method outlined in this PR and adding `time.time()` to the Jinja 2 context.  All expressions in a `gcode macro` are evaluated, then then template is rendered, typically resulting in one or more gcode commands.  These commands are then queued for execution.  If anything is blocking the gcode queue then the execution of said commands will be delayed.

30350n:
Thanks, that makes sense. In that case, [here's](#6010) a new PR.

PR #5883, opened 2022-11-07
virtual_pins: initial implementation of virtual pins

pedrolamas:
This is a follow up on https://klipper.discourse.group/t/new-virtual-pins-module-for-testing-purposes/4727/6

As the AVR atmega644p only has 30 usable pins, that can be quite low when one wants to test features while developing for Fluidd or Mainsail.

This is where virtual-pins comes in!

On the “printer.cfg”, add a `[virtual_pins]` section, and from that point on, try setting a few pins to `virtual_pin:<pin-name>`.

That will add a new SET_VIRTUAL_PIN gcode command that allow to set a virtual pin value (pass `PIN=<pin-name>` and `VALUE=<value>`)

The global or individual pins state can be retrieved from the status of `printer.virtual_pins`.

Note: I understand that this might not be of usefulness in Klipper mainline (this is mostly a debug helper feature!), but in the end, its impact is only seen when the section is added, so I am sending a PR so that this can be properly evaluated.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1562,6 +1562,13 @@ Marlin/RepRapFirmware compatible M486 G-Code macro.
 [exclude_object]
 ```
 
+### [virtual_pins]
+
+Enables support for virtual pins, unseful for testing purposes.
+
+See the [command reference](G-Codes.md#virtual_pins) for additional
+information.
+
 ## Resonance compensation
 
 ### [input_shaper]
--- a/docs/Debugging.md
+++ b/docs/Debugging.md
@@ -278,3 +278,15 @@ using gtkwave with:
 ```
 gtkwave avrsim.vcd
 ```
+
+### Using virtual-pins
+
+The AVR atmega644p has a total of 40 pins, but only 30 of those can actually be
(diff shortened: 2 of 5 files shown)


KevinOConnor:
Interesting.  FWIW, I'm not sure it needs to be in the mainline Klipper branch though.

Cheers,
-Kevin

eliteSchwein:
Would be handy for all developers

zellneralex:
Could be handy to generate toggle switches inside the UI (Fluidd/Mainsail) und use them than as an enable feature insight other macros. Currently I recommend people using a Linux mau and assigning uns used gpio pins to get that functionality.

so there would be also a use case for main klipper but I am not sure if I would find 100 users for that.

pedrolamas:
I have my doubts that one would find 100 users for this feature... having said that, I have bundled this module directly on my Klipper Simulavr Docker image so I can easily test things while I work on Fluidd development, and that sure helps a lot as I don't need to go fiddle with what pins I still have free to use! 😉

Another advantage is that when someone shares a configuration with me, I can just copy-paste the part I am interested in checking, prefix any of the pins with `virtual-pin:`, restart Klipper, and test!

PR #5889, opened 2022-11-08
rp2040: move all code to the ram

sh83:
This is some kind of premature optimization. In rp2040 we have a lot of ram (256kb) and pretty slow main flash. The klipper binary ~30KiB is large enough for the cache (16KiB), but small enough to fit into the ram. In this PR we move all  code to the ram (except ResetHandler and reset_handler_stage_two used for bootstrap).
Positive effect on interrupt latency jitter and general performance is expected.

Signed-off-by: Alex Malishev <malishev@gmail.com>


KevinOConnor:
Thanks.  In general, I agree that moving all the code to ram is technically preferable on the rp2040.  However, it's not clear to me that in practice we would observe an improvement.  I'm leery of adding complexity to the code, and making the rp2040 less like other chips, unless we can measure a meaningful improvement.

Does this change improve the benchmarks?  Is there a realistic test that could be run to measure the improvement?

Cheers,
-Kevin

sh83:
@KevinOConnor Thank you for the review. I run stepper benchmark and I did not see any difference. Quite possible what all "hot" klipper mcu code fits into cache well. I will try to make more complicated benchmark, what use more functionality, but it is not easy task. 

Meanwhile I modified PR to avoid code changes. Now it is handled at linker script level. It do not relocate vector table and couple of functions,  cause they fit cache 100%, cause other code already in ram.
I believe in this form PR is pretty inexpensive way to avoid unexpected slowdown in rare conditions caused by cache miss.

KevinOConnor:
Thanks.  As high-level feedback, I'm leery of making this change.  I'm concerned that the changes are subtle and could make the build a little more fragile.  In particular, it's unclear that rp2040 code is dependent on the names of the functions in armcm_boot.c, and if one were to change the armcm_boot.c code it could result in an unbootable rp2040 (with no build time errors).

Separately, this code would not move the vector table, memcpy, or memset into ram - which could lead to surprising results.

So, I agree that it would be technically better to move all the code into ram.  However, I'm not sure the complexity of doing that is worthwhile.  In particular, since it is likely that all the run-time rp2040 code will fit in the cache.

-Kevin

PR #5892, opened 2022-11-10
stm32: add STM32H723 support

bigtreetech:
* Update `lib/stm32h7` to the lastest version.
* Add STM32H723 support

# known issues
* `make flash FLASH_DEVICE=/dev/serive/by-id/usb-Klipper_stm32h723xxxx` It cannot be executed normally unless the `*(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG` flag bit is written twice or waiting for a period of time after the flag bit is written and before  `NVIC_SystemReset();`


src/stm32/stm32h7.c
```diff
@@ -210,6 +215,10 @@ clock_setup(void)
 static void
 usb_reboot_for_dfu_bootloader(void)
 {
+#ifdef STM32H723xx
```
bigtreetech, line 218: It's strange here. It's still in debugging. Any useful experience and suggestions are welcome

src/stm32/stm32h7.c
```diff
@@ -210,6 +215,10 @@ clock_setup(void)
...
+    irq_disable();
+    *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;
+#endif
     irq_disable();
     *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;
```
sh83, line 223: Just wild guess: Do this variant work?
*(volatile uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;

bigtreetech: Tested `volatile`, it does not work, and execute `barrier()` after `*(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG`, also has no effect.

sh83: Another guess is to use SCB_CleanDCache();
Preferably with barrier () or volatile.

bigtreetech: Thank you very much for your help.

```
irq_disable();
*(volatile uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;
barrier();
SCB_CleanDCache();
```
Just tested and still not working.

KevinOConnor:
That's odd.  Out of curiosity, does the following work:
```c
#include "board/io.h" // readl
...

static void
usb_reboot_for_dfu_bootloader(void)
{
    irq_disable();
    *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;
    readl(USB_BOOT_FLAG_ADDR); // Make sure write is flushed
    NVIC_SystemReset();
}
```

bigtreetech:
> That's odd. Out of curiosity, does the following work:
> 
> ```c
> #include "board/io.h" // readl
> ...
> 
> static void
> usb_reboot_for_dfu_bootloader(void)
> {
>     irq_disable();
>     *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG;
>     readl(USB_BOOT_FLAG_ADDR); // Make sure write is flushed
>     NVIC_SystemReset();
> }
> ```


image
image
Still not working. It's really odd.


image
image

KevinOConnor:
Thanks.  I agree it would be useful to add support for stm32h723 chips.  Some notes:
1. I think we should look to commit this in early December, after the v0.11.0 release.
2. Would it be possible to use files from the v1.9.0 release of stm32CubeH7 instead of v1.10.0?  The v1.9.1 and later releases update the copyright statement in the files to be confusing (making a reference to a LICENSE file that doesn't actually exist).  I'd prefer to stay with the previous (more clear copyright statement) if we can.
3. The `while` loop in `usb_reboot_for_dfu_bootloader()` look strange.  If this issue is due to the dcache, can it just be disabled with something like: `irq_disable(); SCB_DisableDCache(); *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG; NVIC_SystemReset();`
4. I think this PR will need to be broken up into multiple commits due to its size - for example, one commit updating lib/stm32h7, one commit adding stm32h723  support, one commit fixing the bootloader reboot.  I can help with this.

-Kevin

bigtreetech:
* lib/stm32h7 has been updated to v1.9.0
* `irq_disable(); SCB_DisableDCache(); *(uint64_t*)USB_BOOT_FLAG_ADDR = USB_BOOT_FLAG; NVIC_SystemReset();` doesn't work. It's really strange. Before finding the root cause of the problem, I don't add the change of DFU to avoid affecting the merge of this PR. The DFU will create another new fix PR.

KevinOConnor:
Thanks.  I fixed up the merge conflicts and merged this.  I also made some minor whitespace changes to stm32h7_adc.c and slightly reworked the ifdefs in usbotg.c .

-Kevin

PR #5897, opened 2022-11-11
docs: Dual Carriage

grigi:
Adds documentation for IDEX printers.
This includes:
* Offset calibration macros, docs and prints
* Fan control improvements
* Tool changing macros
* Sample config
* Some extra comments around IDEX printers

Based on #4508

Signed-off-by: Nickolas Grigoriadis <nagrigoriadis@gmail.com>


KevinOConnor:
Thanks.  As high-level feedback, this PR seems to be a mix of a few different "orthogonal" topics:
1. Bug fix to the Config_Reference.md document.
2. An example of advanced fan switching when using a tool changer.
3. A Z offset tuning mechanism and a guide to using it.

If you want to split out the bug fix, I'll go ahead and commit it.

The fan changes seem a bit complex to me - I'm not sure if it would be more help than confusion for a new user.  That said, I don't see an issue with including something like that - perhaps in its own sample config file.

I'm not sure I fully understand the z offset tuning macros and guide.  I understand that z tuning is finicky on many idex printers and careful calibration is needed.  It's not clear to me why that should be in the Klipper docs though, as I'd think this is a fundamental calibration task that an owner of an idex printer would need to do regardless of firmware.  Said another way, why should the Klipper docs advocate one particular z tuning mechanism over others?

Cheers,
-Kevin

grigi:
Thanks for the review Kevin.

Ok, so split this PR up into the following:
1. Bug fix to the Config_Reference.md document.
2. Tuning guide
3. Fan switching example

That seems fair. 

On the tuning guide, I can try streamlining it more.

Why should we advocate a tuning mechanism?
Doesn't Klipper already advocate for the paper tuning mechanism for Z for example?
I could simplify the tuning guide some more. 
It's based on common tuning strategies from other IDEX printers. Most of them have menu driven macros to get gross calibration done easily, so some of it was to try emulate that.

Re the macro, It's doing both X & Y lines (I can remove Y as it's less useful at this stage). It also travels back along the printing path to make the join cleaner, as chances are your printer isn't well tuned when you're doing this calibration step. I'll add some comments right by the macro to clarify.

I'll have a look at splitting this PR up and simplifying/clarifying it for you in the next week.

KevinOConnor:
I'm not sure the bot should have closed this ticket.  I'm also not sure what the state of the ticket is though.

-Kevin

PR #5898, opened 2022-11-11
scripts: convert to python3

junland:
Commit converts python scripts within the `scripts/` directory to python3.

Signed-off-by: John Unland <unlandj2012@gmail.com>

### DIFF
--- a/scripts/buildcommands.py
+++ b/scripts/buildcommands.py
@@ -1,7 +1,8 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 # Script to handle build time requests embedded in C code.
 #
-# Copyright (C) 2016-2021  Kevin O'Connor <kevin@koconnor.net>
+# Copyright (C) 2016-2021 Kevin O'Connor <kevin@koconnor.net>
+# Copyright (C) 2022 John Unland
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import sys, os, subprocess, optparse, logging, shlex, socket, time, traceback
@@ -42,7 +43,7 @@ class HandleCallList:
         pass
     def generate_code(self, options):
         code = []
-        for funcname, funcs in self.call_lists.items():
+        for funcname, funcs in list(self.call_lists.items()):
             func_code = ['    extern void %s(void);\n    %s();' % (f, f)
                          for f in funcs]
             if funcname == 'ctr_run_taskfuncs':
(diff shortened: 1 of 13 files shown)


scripts/buildcommands.py
```diff
@@ -42,7 +43,7 @@ def update_data_dictionary(self, data):
         pass
     def generate_code(self, options):
         code = []
-        for funcname, funcs in self.call_lists.items():
+        for funcname, funcs in list(self.call_lists.items()):
```
grigi, line 46: In python2 `dict.items()` returns a list, in python3 it returns an generator.
So you added `list()` to convert the iterable generator to a list, and then iterate over it.

This seems completely unnecessary.
Please only convert to list where it's really required, else you're just introducing unnecessary complexity and a performance hit.

scripts/checkstack.py
```diff
@@ -180,14 +179,14 @@ def main():
             elif insn in ('rcall', 'call'):
                 cur.noteCall(insnaddr, calladdr, stackusage + 2)
             else:
-                print("unknown call", ref)
+                print(("unknown call", ref))
```
grigi, line 182: passing in a single tuple to `print()` will force the result to be printed as:
```
('unknown call',  <ref>)
```

Which is not desirable, leaving the print as before will give the desired result of:
```
unknown call <ref>
```

scripts/logextract.py
```diff
@@ -489,9 +487,9 @@ def parse_line(self, line_num, line):
     def get_lines(self):
         # Ignore old stats
         all_ts = []
-        for mcu_name, mcu in self.mcus.items():
-            all_ts.extend(mcu.sent_seq_to_time.values())
-            all_ts.extend(mcu.receive_seq_to_time.values())
+        for mcu_name, mcu in list(self.mcus.items()):
+            all_ts.extend(list(mcu.sent_seq_to_time.values()))
```
grigi, line 491: `list.extend()` accepts an iterable and does the right thing. Also no need to convert to list.

grigi (changes requested):
Pretty much all the code changes that I can see are unnecessary and just reduces readability and introduces code bloat.
Other than the hash-bang to call a python3 instead of python2 I don't see a benefit.

Possibly the hashbang needs to be just `python` so it will run on whichever is the default in the environment?

Arksine:
In addition to @grigi's feedback I'd point out that there are items missing as well.  For example, even with these changes `webhooks.py` will not work on Python 3 as the strings must be converted to byte strings.

Also it isnt necessary (or desirable) to include yourself in the Copyright a file when making a modification.  Generally that is reserved for those who created the file.

junland:
@Arksine @grigi Thanks, for what it's worth I did used a Py2 to Py3 converter to get most of the code converted over and did a visual inspection. Granted I'm still rusty on python, but I'll close this PR and see if I can correct it. I may just submit separate PR's for a smaller chunk of the scripts. Thanks for the comments.

PR #5907, opened 2022-11-18
My AZSMZ Mini + AZSMZ Display config

irgendwer92:
Hi, 

in the last years i changed every single part of my corexy project, so i decided to reuse the leftovers in the "Trashcube". 

Part of this was configuring Klipper for this chinese Smoothieboard clone with the fitting display:
[AZSMZ Mini](https://de.aliexpress.com/item/32835645667.html) and [AZSMZ 12864 LCD](https://de.aliexpress.com/item/32837222770.html)

It's not directly supported by Smoothieware, so the last FW-Version was from 2017 and a pain to use. I'd never have reused it again without Klipper, love your work! Maybe someone also has that hacky abomination laying around, so i'd like to contribute this.

### DIFF
--- a/config/generic-azsmz-mini.cfg
+++ b/config/generic-azsmz-mini.cfg
@@ -1,0 +1,131 @@
+# This file contains common pin mappings for AZSMZ Mini and the fitting display. To use
+# this config, the firmware should be compiled for the LPC1769.
+
+# See docs/Config_Reference.md for a description of parameters.
+[include mainsail.cfg] 
+
+[stepper_x]
+step_pin: !P2.0
+dir_pin: !P0.5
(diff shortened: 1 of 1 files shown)


irgendwer92:
Sorry, that's my first pull request

Neko-vecter (changes requested):
line 5 "[include mainsail.cfg]" should be remove for different setup/config of Klipper. For example, fluidd or octoprint.

The formatting of the commit message need update. Following [CONTRIBUTING](https://www.klipper3d.org/CONTRIBUTING.html) Docs.

Also, you need add into printers.test file. More info in [Example_Configs](https://www.klipper3d.org/Example_Configs.html) Part 4.

PR #5911, opened 2022-11-24
stm32h7_adc: Fix potential race condition with ADC values

D4SK:
When playing around with some adc setups i found that if the timer checking the adc values is set to a time significantly after the conversion is ready, the values between different adc pins could get mixed up. With the stock klipper code this was extremely unlikely to happen as the adc request would have to arrive within a window of <1us.

This was already forshadowed by the comment 
 // ADSTART is not as long true as SR_STRT on stm32f4


D4SK:
closing this as it is addressed by #5913

PR #5912, opened 2022-11-25
Add delta implementation of Z_OFFSET_APPLY_ENDSTOP

janherich:
Current Z_OFFSET_APPLY_ENDSTOP command only works for printers with cartesian architecture which have separate Z axis defined. But this functionality (persisting Z babystep value to endstops) is exactly as convinient for delta printers, therefore this PR implements it.

### DIFF
--- a/klippy/extras/manual_probe.py
+++ b/klippy/extras/manual_probe.py
@@ -13,9 +13,25 @@ class ManualProbe:
         self.gcode_move = self.printer.load_object(config, "gcode_move")
         self.gcode.register_command('MANUAL_PROBE', self.cmd_MANUAL_PROBE,
                                     desc=self.cmd_MANUAL_PROBE_help)
+        # Endstop value for cartesian printers with separate Z axis
         zconfig = config.getsection('stepper_z')
         self.z_position_endstop = zconfig.getfloat('position_endstop', None,
                                                    note_valid=False)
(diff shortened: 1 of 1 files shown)


manu7irl:
Hi I think this method is really useful I got 3 deltas and running a delta_calibration each time my offset is changed, is driving me nuts.
I think that implementing that should be done by doing a contional state to see which type of kinematic is in use, and them apply the same command method than for a cartesian printer, if I am correct your implementation adds a new command named, z_offset_delta_apply_endstop.
I think that's a bit too much for users to run, also this should cause problem as fluidd and mainsail will have to adapt their codes for it.
So better to use the same name and apply the values differently based on the kinematic in use.
Correct me if I am wrong.
@janherich

janherich:
Hi @manu7irl this PR implements new command internally but it's hooked to the same GCODE command `Z_OFFSET_APPLY_ENDSTOP` exactly because of the reasons you listed, so UIs like Mainsail/Fluid don't have to bother with decision logic and no new code is required - from UI POV, it's the same functionality and printer architecture doesn't matter.

manu7irl:
Ok so I will try it on my deltas and report, please sign it off as Kevin requested.

janherich:
@KevinOConnor I signed off the commit as requested and also changed the logic to check for `delta` kinematics when registering the delta specific `Z_OFFSET_APPLY_ENDSTOP` gcode command

PR #5913, opened 2022-11-25
Apply race fixes to stm32h7_adc.c

KevinOConnor:
Improve handling of race conditions with hardware updates.  This is the same changes applied to stm32f0_adc.c in commit 88325b6c.

This is an alternate implementation of PR #5911 .

@adelyser , @mattthebaker , @D4SK - fyi.

-Kevin

### DIFF
--- a/src/stm32/stm32h7_adc.c
+++ b/src/stm32/stm32h7_adc.c
@@ -28,7 +28,6 @@
   // Number of samples is 2^OVERSAMPLES_EXPONENT (exponent can be 0-10)
   #define OVERSAMPLES_EXPONENT 3
   #define OVERSAMPLES (1 << OVERSAMPLES_EXPONENT)
-  #define ADC_MEAS_DELAY  (1 + 2.3666*OVERSAMPLES)
 
   // LDORDY registers are missing from CMSIS (only available on revision V!)
   #define ADC_ISR_LDORDY_Pos                  (12U)
@@ -44,7 +43,6 @@
   #define ADC_TS                              (ADC12_COMMON)
 
   #define OVERSAMPLES                         (0)
-  #define ADC_MEAS_DELAY                      (10)
 
 #elif CONFIG_MACH_STM32G4
   #define ADCIN_BANK_SIZE                     (19)
@@ -57,7 +55,6 @@
   #define ADC_CCR_TSEN                        (ADC_CCR_VSENSESEL)
 
   #define OVERSAMPLES                         (0)
-  #define ADC_MEAS_DELAY                      (10)
 #endif
 
 #define ADC_TEMPERATURE_PIN 0xfe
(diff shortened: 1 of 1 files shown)


KevinOConnor:
FYI, I don't have the stm32h7 hardware, so I have only "compile tested" this change.

-Kevin

D4SK:
Just tested this and it works fine.

adelyser:
I'll test this on my 2 printers here in the next couple days

src/stm32/stm32h7_adc.c
```diff
@@ -303,9 +299,11 @@ gpio_adc_cancel_sample(struct gpio_adc g)
 {
     ADC_TypeDef *adc = g.adc;
     irqstatus_t flag = irq_save();
-    // ADSTART is not as long true as SR_STRT on stm32f4
-    if ((adc->CR & ADC_CR_ADSTART || adc->ISR & ADC_ISR_EOC)
-        && adc->SQR1 == (g.chan << ADC_SQR1_SQ1_Pos))
-        gpio_adc_read(g);
+    if (adc->SQR1 == (g.chan << ADC_SQR1_SQ1_Pos)) {
+        if (adc->CR & ADC_CR_ADSTART)
+            adc->CR |= ADC_CR_ADSTP;
```
mattthebaker, line 304: It makes me a little nervous that there may be a race with this read modify write compared to the f0 version which is only a write. With very unfortunate timing, ADSTART may clear between the read and the write. 

Based on this excerpt from the L412 manual, it suggests ADSTP might be blocked from occurring (which is okay because it has already stopped), but it might be able to set the ADSTART bit.

"The software is allowed to write the ADSTP or JADSTP control bits of the ADC_CR register
only if the ADC is enabled, possibly converting, and if there is no pending request to disable
the ADC (ADSTART or JADSTART must be equal to 1 and ADDIS to 0)."

mattthebaker:
Looks good to me other than the inline comment, and the f0 version worked effectively on the l4 previously.

When I read "allowed" in the data sheet I get the impression that logic might block setting of the bit, but its hard to say.

src/stm32/stm32h7_adc.c
KevinOConnor: Good catch!  I modified this PR to try to avoid that race.

In general, I think stm32h7_adc.c should be reworked to avoid read-modify-write on all `ADC->CR` accesses.  That register has side-effects so using read-modify-write is just asking for subtle race conditions.  However, I don't want to conflict with your stm32g4 PR nor the stm32h723 PR.  So, something to look at after those PRs are merged.

-Kevin

KevinOConnor:
Thanks.  I was also able to test this on an stm32h723.  I have merged this PR.

-Kevin

PR #5916, opened 2022-11-29
scripts: Python3 migration

junland:
Here's my second attempt. I excluded `whconsole.py` and `buildcommands.py` as those two scripts would take a little more time / work to get migrated over. Also it was suggested in the last PR that setting the she bang to just `python` (Which most likely be symlinked to `python3`) so I did just that and left the scripts alone that explicitly required `python2` for compatibility sake.


ajmirsky:
I stumbled across another python 2/3 discrepancy in `parsedump.py` when expanding arc command support. It's included in that [pull request](https://github.com/Klipper3d/klipper/pull/5910) but you can cherrypick/patch from [`klippy/parsedump.py`](https://github.com/Klipper3d/klipper/pull/5910/commits/cc362082bef6a893c99912c710e030004bd043e1#diff-1bbb9cd3f60b825fec9c0ec2dea625cefaedda5f9245176c088a67bad7fbef0a) directly.

junland:
> I stumbled across another python 2/3 discrepancy in `parsedump.py` when expanding arc command support. It's included in that [pull request](https://github.com/Klipper3d/klipper/pull/5910) but you can cherrypick/patch from [`klippy/parsedump.py`](https://github.com/Klipper3d/klipper/pull/5910/commits/cc362082bef6a893c99912c710e030004bd043e1#diff-1bbb9cd3f60b825fec9c0ec2dea625cefaedda5f9245176c088a67bad7fbef0a) directly.

Thanks, I believe it was mentioned in my [previous PR](https://github.com/Klipper3d/klipper/pull/5898#issuecomment-1312423483) that `whconsole.py` would have this problem with strings / byte strings conversion. Though I'll hold off to send in that PR as I'm still rusty on my python.

KevinOConnor:
Thanks.  I agree we should update these scripts so that they can run with either python2 or python3.

It's important that white space changes aren't mixed with functional changes - as mentioned at https://www.klipper3d.org/CONTRIBUTING.html .

A big commit containing all these changes is difficult to review and makes it hard to use tools like `git bisect` and `git revert`.  I'd prefer to see one commit per tool, along with a declaration that the tool was tested on both python2 and python3.  I'd rather the software clearly state if it is python2 or python3 then for it to have subtle bugs.

Cheers,
-Kevin

junland:
> Thanks. I agree we should update these scripts so that they can run with either python2 or python3.
> 

python2 was [EOL'd / Sunset](https://www.python.org/doc/sunset-python-2/) Jan 1st. 2020. While I understand that these scripts should run both language versions, there should be a long term goal of fully moving / requiring python3. One less thing to test against is always good. :)

> It's important that white space changes aren't mixed with functional changes - as mentioned at https://www.klipper3d.org/CONTRIBUTING.html .
> 
> A big commit containing all these changes is difficult to review and makes it hard to use tools like `git bisect` and `git revert`. I'd prefer to see one commit per tool, along with a declaration that the tool was tested on both python2 and python3. I'd rather the software clearly state if it is python2 or python3 then for it to have subtle bugs.
> 
> Cheers, -Kevin

Gotcha, I'll reset the commits and see if I can make individual ones. I just put it as one big commit to reduce the commit log (it's also what we do at my work so its just a natural thing for me to do) Might be good to mention something about that in the contributing doc about number of commits in a PR.

KevinOConnor:
You can batch together changes together if you want, but one big commit seems a bit unwieldy.  (It doesn't need to be one commit per tool.)  The big thing is confidence that the tools will actually run correctly on python3.

Cheers,
-Kevin

junland:
Moving to this [PR](https://github.com/Klipper3d/klipper/pull/5931) as I messed up my git history.

Issue #2457, opened 2020-02-01
Delta: Incorrect position boundary checks


pienjo:
The maximum build dimensions seem to be calculated incorrectly on my delta printer (Arm length ~270mm, delta radius 149mm, build radius 120mm).

It claims that the max. build height is 239.86 mm(which corresponds to the minimum endstop position) with the radius tapered above 195.23mm. This isn't correct; if I move the toolhead to X0 Y-120 Z195 (which is on the boundary of the printable area according to Klipper, there is still ~35mm between the A and B carriages and their endstops. In other words: The maximum build height should've been around 230mm instead. 

Also, the tapering seems incorrect; I was expecting a build volume resembling a cylinder with a cone on top, with the base radius of the cone equal to the cylinder. This doesn't seem to be the case: At Z=195.5, the first Y coordinate which Klipper claims to be reachable is -44.4. At Z=200, the radius has reduced to -39.8 (at which point there is still a minimum of ~28 mm clearance between the endstop and the carriages). So instead of a cylinder with a connecting cone, it seems the build volume on my printer is a cylinder with a skinny, tall cone, whose base diameter is significantly smaller than the cylinder.

Any suggestions?

See logfile:
[klippy.log](https://github.com/KevinOConnor/klipper/files/4142709/klippy.log)

KevinOConnor:
> The maximum build height should've been around 230mm instead.

True, Klipper could have done better here.  Your delta arms are a little shorter than most delta printers, and the current code isn't taking into account build_radius being smaller than delta_radius.

> So instead of a cylinder with a connecting cone, it seems the build volume on my printer is a cylinder with a skinny, tall cone, whose base diameter is significantly smaller than the cylinder.

This is correct - the boundary constraints are currently a cylinder with a "tiny" cone on top of it.

Can you give some high-level info on what printer you have and the sizing of prints you are unable to complete (if that is the case)?

It's possible to address both of the above, but the math can get complex quickly.  It's known that the boundary constraints aren't perfect, but (to date) most people don't push the envelope at the very top of their delta printers.

-Kevin

pienjo:
> Your delta arms are a little shorter than most delta printers, and the current code isn't taking into account build_radius being smaller than delta_radius.

[...]

PR #5917, opened 2022-12-01
Better delta build volume constraint checking

janherich:
Existing code uses very restrictive build volume constraint checking with just narrow cone on top of fully cylinder for delta printers.

Code here implements more permissive and still safe build volume constraint checks.
To avoid calling expensive math.sqrt function in check_move method linear interpolation is used instead.

Tested on our small SpaceDelta3D printer which has radius of ~113mm, arm length of 240.7mm and build radius of 103mm.
Limit Z height at the edge of the build radius is a bit above 181mm (slightly improved over current klipper Implementation, since it takes build radius smaller than true radius into account).

With current implementation, the narrow cone on top of the full build volume cylinder has radius of only ~27mm, which means that even at 40-50mm from center, build height is capped at 181mm.
With changes implemented in this PR, over 10mm of extra build height is available at the same distance from center.

### DIFF
--- a/klippy/kinematics/delta.py
+++ b/klippy/kinematics/delta.py
@@ -65,6 +65,8 @@ class DeltaKinematics:
         self.min_z = config.getfloat('minimum_z_position', 0, maxval=self.max_z)
         self.limit_z = min([ep - arm
                             for ep, arm in zip(self.abs_endstops, arm_lengths)])
+        self.min_arm_length = min_arm_length = min(arm_lengths)
+        self.min_arm2 = min_arm_length**2
         logging.info(
             "Delta max build height %.2fmm (radius tapered above %.2fmm)"
             % (self.max_z, self.limit_z))
@@ -123,7 +125,11 @@ class DeltaKinematics:
         end_z = end_pos[2]
         limit_xy2 = self.max_xy2
         if end_z > self.limit_z:
-            limit_xy2 = min(limit_xy2, (self.max_z - end_z)**2)
+            above_z_limit = end_z - self.limit_z
+            allowed_radius = self.radius - math.sqrt(
+                self.min_arm2 - (self.min_arm_length - above_z_limit)**2
+            )
+            limit_xy2 = min(limit_xy2, allowed_radius**2)
         if end_xy2 > limit_xy2 or end_z > self.max_z or end_z < self.min_z:
             # Move out of range - verify not a homing move
             if (end_pos[:2] != self.home_position[:2]


janherich:
To better illustrate the differences, this is current build volume:
delta_current_boundary

And this is the boundary as implemented in PR:
delta_pr_boundary

Just for reference, this is true maximum boundary, which is a complex shape based on extraction of 3 spheres from build volume cylinder, as already mentioned in linked issue, even for planar X/Y move, can start in valid area, end in valid area and still pass through invalid area out of boundary, which would be quite problematic to efficiently check in klipper code:
delta_ideal_boundary

KevinOConnor:
Thanks.  Alas, I'm not sure the code on this PR is "correct".  The reason the current code only supports a "cone" is due to xy+z moves.

So, for example, if the cylinder has radius of 50 and ends at z=100, and the top of the cone is at z=150, then a move from x=49,y=0,z=151 to x=0,y=0,z=148 is likely invalid even though both the start and end positions are valid.  Unless I'm missing something, this PR would permit the move.

Admittedly, I think the current code would allow a move from within the cylinder to the cone which could also be invalid.

As mentioned in #2457, we could add the ability to turn off some of these checks.

If we don't want to validate xy+z moves then I don't think interpolation is necessary.  It's not necsessary to call `sqrt()` during the calculation as one can just compare the "squared" values directly.  That is, something like `end_xy2 <= arm_length2 - (arm_length - z_pos)**2`.

-Kevin

janherich:
Hi @KevinOConnor:
About the first point, while it's true that linear move in all three axes can indeed cross the valid zone, I fail to see how it's different from current implementation - move starting at the edge of the build cylinder just below `limit_z` and proceeding straight to the top of cone (x=0,y=0, z just below `max_z`) would be still considered valid in current implementation (and this PR as well), while travelling through invalid zone and causing carriage collision - I can check this on master if you wish although I prefer to not crash the printer intentionally. 

Unless I'm missing something the only way to protect against this is to consider both start/end of the move a detect line/surface intersection, or make the top cone have the same angle as the tangent of the circle formed by arm as it approaches the center and drag the cone all the way down, so build cylinder radius and cone base radius would be the same - this would result in a lot of wasted space at the edge of the build envelope.

Regarding second point of just comparing squared values directly, that was my initial idea, unfortunately I was not able to figure it out - your proposed solution (`end_xy2 <= arm_length2 - (arm_length - z_pos_above_limit_z)**2`) is unfortunately incorrect, as `(arm_length - z_pos)**2` terms give you square of the radius part you have to subtract from radius to get limit, not valid limit radius (or it's square) directly.

Given our hypothetical example (r=50, limit_z=100, max_z=150, arm_length=50), consider `end_xy2 = 625` (move ending at radius 25mm from center) and height just below `max z` (let's say 149). this makes the second term `arm_length2 - (arm_length 2 - z_pos_above_limit_z)**2` evaluate to `2499` which is more than `625` but the move is still invalid (straying 25mm from center at the very top of the build envelope).

Ultimately it boils to the equation `r = math.sqrt(end_xy2) + math.sqrt(- arm-length2 math.sqrt(- arm_length z_pos_above_limit_z))` and I don't know how to solve that without square root.

KevinOConnor:
Okay, I think I understand.

Although I agree a simple formula isn't possible, I still don't think the interpolation is necessary.  My high-level understanding is that a `sqrt()` operation is roughly the cost of a "divide" operation on modern computers.  I'd be surprised if interpolation improved performance (I suspect calling a python function would be dramatically more expensive than calling `sqrt()`).

I'm also still unsure about the issue with xy+z moves.  I agree the current system also has the issue.  I'll try to give it some additional thought.

Thanks,
-Kevin

janherich:
@KevinOConnor I did some research and it seems like `math.sqrt()` in python links to C `sqrt()` function in python interpreter, therefore it should be reasonable fast, and linear interpolation is unnecessary. 
I will update the PR with simple sqrt call instead of interpolation.

KevinOConnor:
Thanks!

-Kevin

ItsMeAubey:
I'm currently running into an issue with XY+Z moves - Homing, then trying to move to x=0 y=print_radius z=cone_start_z leads to crashing the rear tower into the endstop repeatedly. The start and end positions are valid, but positions between them are not. 

I'm able to workaround by moving to Z=cone_start_z before doing anything, but it would be nice to not have to remember to do this and risk damaging my printer if I forget to, or to avoid damage if a slicer decides to g28, then move to a position that is valid but which has invalid travel move positions. 

Alternatively, an optional setting to limit Z to cone_start_z and to move to it after homing would be a suitable workaround as the cone generally isn't useful print volume anyways. 

Please let me know if I can be of any assistance in testing anything if anybody decides to pick this issue up. Thank you!

KevinOConnor:
Unfortunately, checks on XY+Z moves into and out of the "cone" have never been fully validated.  It will take an interested developer to implement and submit code.  Admittedly, the documentation should probably be updated to note this limitation, though it is never recommended to submit invalid moves to Klipper.

-Kevin

PR #5919, opened 2022-12-02
scripts : parsedump python3 update

ajmirsky:
python 2 implicitly decodes byte arrays into strings. python 3 needs explicit conversion.

Signed-off-by: Andrew Mirsky <andrew@mirsky.net>

### DIFF
--- a/klippy/parsedump.py
+++ b/klippy/parsedump.py
@@ -13,6 +13,20 @@ def read_dictionary(filename):
     dfile.close()
     return dictionary
 
+# bytes to string differ between python 2 & 3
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  The underlying issue here was that parsedump.py (and some parts of msgproto.py) should have been using a `bytearray()` instead of `str` or `bytes`.  Most of msgproto.py was converted to this format last year.  I went ahead and updated the remaining parts of msgproto.py and parsedump.py (commit 336cc92a).

-Kevin

PR #5920, opened 2022-12-03
tmc: Configurable microstep lookup table

leptun:
Make all the tmc microstep lookup table registers configurable via the config file. It also loads the default values on startup if the parameters are not provided. TMC220x and TMC2660 do not support this feature (only TMC2130 and TMC5160 for now).
Without this PR these registers can't be configured at all (no fields defined, just the register address, so not even SET_TMC_FIELD works). Without this PR the default values are also not loaded, so it is possible to use an incorrect value in these registers if you switch from a firmware that uses them (eg, Prusa-Firmware for the MK3) and don't power cycle the printer.

And here is a request for this feature on the klipper forum: https://klipper.discourse.group/t/access-to-tmc-microstep-lookup-table-registers/2903?u=leptun. There are more, but this was the only direct request for this feature.

Todo: update docs after everything else is accepted

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3073,6 +3073,23 @@ run_current:
 #   set, "stealthChop" mode will be enabled if the stepper motor
 #   velocity is below this value. The default is 0, which disables
 #   "stealthChop" mode.
+#driver_MSLUT0: 2863314260
+#driver_MSLUT1: 1251300522
+#driver_MSLUT2: 608774441
+#driver_MSLUT3: 269500962
+#driver_MSLUT4: 4227858431
+#driver_MSLUT5: 3048961917
+#driver_MSLUT6: 1227445590
+#driver_MSLUT7: 4211234
+#driver_MSLUTSEL: 4294934614
+#driver_MSLUTSTART: 16187392
+#   These fields control the Microstep Table registers directly. The optimal
+#   wave table is specific to each motor and might vary with current. An
+#   optimal configuration will have minimal print artifacts caused by
+#   non-linear stepper movement. The values specified above are the default
+#   values used by the driver. The value must be specified as a decimal integer
+#   (hex form is not supported). In order to compute the wave table fields,
+#   see the tmc2130 "Calculation Sheet" from the Trinamic website.
 #driver_IHOLDDELAY: 8
 #driver_TPOWERDOWN: 0
 #driver_TBL: 1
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Thanks.  In general it looks fine to me.  Please don't make whitespace changes to the `Registers` field of tmc2130.py as part of this change.  Please do make the corresponding updates to docs/Config_Reference.md .  Finally, commits need to have a signed-off-by line.

-Kevin

KevinOConnor:
Thanks.  Looks fine to me.  I think @dmbutyugin played with these tables in the past, so I'll give a couple of days to see if he has any comments.  Otherwise I'll look to commit.

Cheers,
-Kevin

dmbutyugin:
Thanks Kevin and Alex. Indeed I looked into it previously. In general, I found that the optimal table would have to cover more than just a quarter of a sine. But this goes outside the scope of this PR, I think overall it makes sense to have this functionality to override these registers in Klipper. It may facilitate future experimentation, and this is something I had to do myself in my experimental branch anyway.

As a small note, MSLUTSEL and MSLUTSTART have internal fields actually (i.e. W0-W3 and X1-X3 for MSLUTSEL and START_SIN[0] and START_SIN90 for MSLUTSTART), but I'm not sure how critical it is to actually expose them to the users. I actually did that in my experimental branch because I used my own sine wave table compressor that produced these fields and I was copying them into TMC registers that way. And even though most users, I imagine, would be using TMC2130/TMC5160_Calculations spreadsheet, but it also seems to deal with those register fields. So, just a thought to add instead
```
+Fields["MSLUTSTART"] = {
+    "START_SIN":                0xFF << 0,
+    "START_SIN90":              0xFF << 16,
+}
+Fields["MSLUTSEL"] = {
+    "X3":                       0xFF << 24,
+    "X2":                       0xFF << 16,
+    "X1":                       0xFF << 8,
+    "W3":                       0x03 << 6,
+    "W2":                       0x03 << 4,
+    "W1":                       0x03 << 2,
+    "W0":                       0x03 << 0,
+}
```
But don't take my feedback as blocking. It's just a thought.

leptun:
I was also thinking whether I should make those as actual fields or to just define a single field that covers the entire register. In the end, I went with the latter option since those fields aren't really user configurable without also altering the MSLUT values. At that point, you might as well use the script again and regenerate all values. I'll leave this up to Kevin if he wants more configuration parameters in printer.cfg or have them combined as they are now.
As an interesting note, there are other tmc drivers which do actually have user configurable options in the MSLUTSTART register which don't require a complete regeneration of the MSLUT table, but those drivers are not supported in klipper just yet (I will make a PR for that at a later time :) ). For such a driver, it would make sense to split the fields like you showed above.

KevinOConnor:
FWIW, it looks to me like the sub-fields should be defined and used in Klipper.  They appear to be defined in the Trinamic specifications in a similar way to other fields that Klipper breaks out.

Cheers,
-Kevin

leptun:
Updated as requested

dmbutyugin:
Yes, looks good from my side, thanks!

KevinOConnor:
Thanks.

-Kevin

PR #5922, opened 2022-12-06
save_variables: Verify filename path is writeable at startup

jschuh:
I've helped a number of people diagnose confusing `save_variable` errors because they copied the `filename` path in from a setup guide, but the copied path was actually invalid on their config. So, this just tries to create an empty file at `filename` if one is not already present. That way the user will get an error pointing to the bad path at startup, rather than a more confusing `'dict object' has no attribute` error sometime later while the printer is running.

### DIFF
--- a/klippy/extras/save_variables.py
+++ b/klippy/extras/save_variables.py
@@ -12,6 +12,8 @@ class SaveVariables:
         self.filename = os.path.expanduser(config.get('filename'))
         self.allVariables = {}
         try:
+            if not os.path.exists(self.filename):
+                open(self.filename, "w").close()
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  In general, it looks fine to me.  I'm a little leery about using different open mechanisms for this check though - any reason not to use the same method that `cmd_SAVE_VARIABLE()` uses?  That is: `open(self.fielname, "w").close()`

-Kevin

jschuh:
Honestly, that line was just muscle memory on my part. Happy to change the write mode to "w" and make the close explicit since it will all have the same effect.

PR #3032, opened 2020-06-29
input_shaper: Initial support of input shaping

dmbutyugin:
Input shaping can help to reduce printer vibrations due to resonances
and eliminate or reduce ghosting in prints.

Signed-off-by: Dmitry Butyugin <dmbutyugin@google.com>

PR #5923, opened 2022-12-06
Add ZVDD & ZVDDD shapers

tehmaze:
Requested by popular demand, the ZVDD & ZVDDD variants as per https://www.researchgate.net/publication/316556412_INPUT_SHAPING_CONTROL_TO_REDUCE_RESIDUAL_VIBRATION_OF_A_FLEXIBLE_BEAM

### DIFF
--- a/klippy/extras/shaper_calibrate.py
+++ b/klippy/extras/shaper_calibrate.py
@@ -13,7 +13,8 @@ MAX_SHAPER_FREQ = 150.
 
 TEST_DAMPING_RATIOS=[0.075, 0.1, 0.15]
 
-AUTOTUNE_SHAPERS = ['zv', 'mzv', 'ei', '2hump_ei', '3hump_ei']
+AUTOTUNE_SHAPERS = ['zv', 'zvd', 'zvdd', 'zvddd', 'mzv',
+    'ei', 'ei2', '2hump_ei', '3hump_ei']
 
 ######################################################################
 # Frequency response calculation and shaper auto-tuning
--- a/klippy/extras/shaper_defs.py
+++ b/klippy/extras/shaper_defs.py
@@ -30,6 +30,22 @@ def get_zvd_shaper(shaper_freq, damping_ratio):
     T = [0., .5*t_d, t_d]
     return (A, T)
 
+def get_zvdd_shaper(shaper_freq, damping_ratio):
+    df = math.sqrt(1. - damping_ratio**2)
+    K = math.exp(-damping_ratio * math.pi / df)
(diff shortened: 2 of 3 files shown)


freakydude:
Tested it with my Artillery Sidewinder X2. Code changes seems to be valid. I also checked the formula in the paper and it seems the equal to the implemented one. 

Here are my graphical results in my case:

shaper_calibrate_x
shaper_calibrate_y

What I miss are some more technical details about the two new algorithms in the documentation. For example I don't know if I should prefer them to MVZ in my case or not.
[Technical Details](https://www.klipper3d.org/Resonance_Compensation.html#input-shapers])

But nice work so far. Thank you!

tehmaze:
I am not a mechanical engineer, so I don't find myself qualified to put any recommendations in the documentation. 

ZVD(D(D)) are most useful if you want to compensate X&Y on the same frequency configuration, they cover a wide range of frequencies, similar to EI. Which one is most suitable depends on many factors, the most determining factor is how far apart the X & Y resonant frequencies are and how wide the "side lobes" of the frequency peaks are, other factors are how quickly the frequencies respond to the impulses.

Sineos:
You always want to look after 3 values from the user perspective:
* Remaining vibration (`vibr` in text box) --> lowest possible value
* Least smoothing (`sm` in text box) --> lowest possible value
* Maximum recommended acceleration before stronger smoothing kicks in (`accel` in text box) --> highest possible value

According to @freakydude's measurement both ZVDD and ZVDDD will offer no benefit but it is only his machine.  Other mileage may vary.

KevinOConnor:
Thanks.  Can you give some additional background on why these additional shapers are desirable?

@dmbutyugin and I discussed these shapers earlier on, and it was thought they would not be particularly useful in practice - https://github.com/Klipper3d/klipper/pull/3032#issuecomment-651894869

-Kevin

dmbutyugin:
@tehmaze It is not difficult to add ZVDD and ZVDDD shapers. In fact, I originally had them implemented, but following a discussion with Kevin, we decided to drop them. It would seem that 2HUMP_EI should always be better then ZVDD and 3HUMP_EI should similarly be preferable over ZVDDD. In fact, a couple of charts posted here already demonstrate that a user could get slightly higher accelerations with 2HUMP_EI/3HUMP_EI than with ZVDD/ZVDDD. So, I wonder if you or anyone else came across the configurations where the ZVDD/ZVDDD shapers would be beneficial? Otherwise, right now they add more of a visual clutter to the calibration tool and 'too many choices' to choose from.

PR #5925, opened 2022-12-08
mcu: new HC32F460 support

SteveGotthardt:
new mcu support for the HC32F460
I have been using the code since Jan 2022. Others have started using/testing and it seems to work well.
The applicable configs (and names) have been updated to comply.


KevinOConnor:
Thanks.  In general, I think it is fine to add support for the hc32f460 chips, as we discussed in PR #5208 .

As with that PR, this PR would need a "git rebase" so that it is easier to review.  Unrelated changes would need to be removed from the PR (changes to gitignore shouldn't be made here, changes to the lcd code would need to be discussed separately).

I'm also not sure I understand why the Kconfig changes ask for specific printers - is there a reason for that?  Could we build a more generic binary in Kconfig and customize it in printer.cfg instead?

I'd like to further review the local code implementation, but I'd really like to see the "git rebase" done first as otherwise it is difficult to review.

-Kevin

SteveGotthardt:
Did the rebase work? whew...
I delayed a little because a kind klipper user (in Germany) sent me a Creality board from an Ender 2 pro. Creality is starting to use hc32f460. Anycubic too. 

I tried not to step on original klipper files - I had to deal with the Aquila X1 LCD driver circuitry that was inverted (I guess to level shift). Maybe there is a better way.

I realize the Kconfig has information in possibly the wrong spots. Since the printer.cfg doesn't contain the UART pins or Flash offset, those should be saved somewhere. I don't have a great grasp on the overall klipper vibe so I am probably missing that maybe I should put comments in the printer configs that tell the user what to pick in the menuconfig?

I would be good to get this work into master. It seems Creality has more printers and now a new hc32f460 board coming/here.

thanks

Sineos:
I'm not the final authority here but as far as I know @KevinOConnor it may be preferred to have the different changes, i.e.
* new MCU
* new printers
* LCD changes
in separate PRs.

KevinOConnor:
> Did the rebase work?

Alas, no.  If you navigate to `https://github.com/Klipper3d/klipper/pull/5925` you'll see tabs for "Conversation", "Commits", "Checks", and "Changed Files".  Next to "Commits" this PR shows 132.  All those commits need to be squashed down to a handful of commits (typically 3-5).  You can find information online about how to use "git rebase" to accomplish that.

Cheers,
-Kevin

SteveGotthardt:
too many files to process...

PR #5926, opened 2022-12-09
linux process: add 'GPIOI' for linux process

bigtreetech:
Add the Klipper control of `GPIOI` port for BTT CB1 (H616). Theoretically, it will not have a negative impact on other SoC.


KevinOConnor:
Thanks.

-Kevin

PR #5927, opened 2022-12-09
STM32 G0 and G4 fix SWD debug

leptun:
Do not clear the DBG_SWEN bit for the MCU families that have it. This bit is by default set to 1 on reset, but the flash latency initialization code would overwrite that bit with 0, disabling the SWD interface.
The G4 can just use a simple mask for the bit, but on the G0 family it's a bit more complicated since not all MCUs have that bit present. In this particular case, the G0x1 family has this bit, but on the G0x0 family it is defined as "reserved". Unfortunately, even though it is reserved, it still might exist there and function correctly (as was the case with my board), so it might be better to follow the instructions from the datasheet in this case and keep reserved bits to the reset value.

@mattthebaker ping as requested by Kevin.

### DIFF
--- a/src/stm32/stm32g0.c
+++ b/src/stm32/stm32g0.c
@@ -183,8 +183,11 @@ armcm_main(void)
 
     check_usb_dfu_bootloader();
 
-    // Set flash latency
-    FLASH->ACR = (2<<FLASH_ACR_LATENCY_Pos) | FLASH_ACR_ICEN | FLASH_ACR_PRFTEN;
+    // Set flash latency, cache and prefetch; use reset value as base
+    uint32_t acr = 0x00040600;
+    acr = (acr & ~FLASH_ACR_LATENCY) | (2<<FLASH_ACR_LATENCY_Pos);
+    acr |= FLASH_ACR_ICEN | FLASH_ACR_PRFTEN;
+    FLASH->ACR = acr;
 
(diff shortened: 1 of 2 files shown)


mattthebaker:
The G4 edit looks good.

Regarding reserved bits they should nearly always be kept at the reset value unless instructed otherwise by the manufacturer.

Is there much of a difference in reset values across the G0 line, such that the default value could be explicitly set in your edit? With bootloaders we can't always assume the present value in the register is correct.

leptun:
@mattthebaker The entire G0 line seems to have ACR initialized to 0x00040600. I guess I could hardcode that and work from there. It would have been nice if the ST headers contained also the default values for each register...

leptun:
Updated the PR. Now I use the datasheet reset value instead of read-modify-writing the ACR register.

mattthebaker:
Personally, I'd try to recreate the reset values from the defines, possibly adding the missing define for some families if they need it. It will just be more readable than a magic number. Kevin may have better guidance though.

leptun:
I did it like this since a few lines higher the same thing happens for some other registers which don't get cleaned up by a bootloader. I'll wait for what Kevin has to say.

KevinOConnor:
Thanks.  It looks fine to me.  I committed this PR.

-Kevin

PR #5928, opened 2022-12-09
stm32g0: add stm32g0b0 support

leptun:
Subset of changes from #5838.

### DIFF
--- a/src/stm32/stm32f0_serial.c
+++ b/src/stm32/stm32f0_serial.c
@@ -87,8 +87,10 @@
 #endif
 
 #if CONFIG_MACH_STM32G0
-  // The stm32g0 has slightly different register names
-  #define USART2_IRQn USART2_LPUART2_IRQn
+  // Some of the stm32g0 MCUs have slightly different register names
+  #ifdef LPUART2
+    #define USART2_IRQn USART2_LPUART2_IRQn
+  #endif
   #define USART_CR1_RXNEIE USART_CR1_RXNEIE_RXFNEIE
   #define USART_CR1_TXEIE USART_CR1_TXEIE_TXFNFIE
   #define USART_ISR_RXNE USART_ISR_RXNE_RXFNE
--- a/src/stm32/stm32f0_timer.c
+++ b/src/stm32/stm32f0_timer.c
@@ -20,14 +20,17 @@
  ****************************************************************/
 
 // Use 32bit TIM2 timer if available (otherwise use 16bit TIM3 timer)
-#ifdef TIM2
-#define TIMx TIM2
-#define TIMx_IRQn TIM2_IRQn
-#define HAVE_TIMER_32BIT 1
-#else
-#define TIMx TIM3
-#define TIMx_IRQn TIM3_IRQn
-#define HAVE_TIMER_32BIT 0
+#if defined(TIM2)
+  #define TIMx TIM2
+  #define TIMx_IRQn TIM2_IRQn
+  #define HAVE_TIMER_32BIT 1
+#elif defined(TIM3)
+  #define TIMx TIM3
+  #define TIMx_IRQn TIM3_IRQn
+  #define HAVE_TIMER_32BIT 0
+  #ifdef TIM4
+    #define TIM3_IRQn TIM3_TIM4_IRQn
+  #endif
 #endif
 
 static inline uint32_t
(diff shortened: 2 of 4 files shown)


src/stm32/stm32f0_serial.c
```diff
@@ -88,7 +88,9 @@
 
 #if CONFIG_MACH_STM32G0
   // The stm32g0 has slightly different register names
+  #ifdef LPUART2
   #define USART2_IRQn USART2_LPUART2_IRQn
+  #endif
   #define USART_CR1_RXNEIE USART_CR1_RXNEIE_RXFNEIE
```
KevinOConnor, line 94: I don't understand what this does.  Is it that USART2_IRQn actually exists on the stm32g0b0?  If so, could this be written as:
```c
    #ifndef USART2_IRQn
      #define USART2_IRQn USART2_LPUART2_IRQn
   #endif
```

src/stm32/stm32f0_timer.c
```diff
@@ -24,10 +24,16 @@
...
+#elif defined(TIM3)
 #define TIMx TIM3
+#ifdef TIM4
+#define TIMx_IRQn TIM3_TIM4_IRQn
+#else
 #define TIMx_IRQn TIM3_IRQn
+#endif
 #define HAVE_TIMER_32BIT 0
+#else
+#error No system timer defined
```
KevinOConnor, line 36: I'm lost in `ifdefs`.  What does this do?  Can this block of ifdefs be indented to make it more clear?

This looks like it adds an extra "no system timer defined" path - why?

src/stm32/stm32g0.c
```diff
@@ -56,7 +60,7 @@ lookup_clock_line(uint32_t periph_base)
         return (struct cline){.en=&RCC->APBENR2,.rst=&RCC->APBRSTR2,.bit=1<<18};
     if (periph_base == ADC1_BASE)
         return (struct cline){.en=&RCC->APBENR2,.rst=&RCC->APBRSTR2,.bit=1<<20};
-    if (periph_base >= APBPERIPH_BASE && periph_base <= LPTIM1_BASE)
+    if (periph_base >= APBPERIPH_BASE && periph_base < APBPERIPH_BASE + 0x8000)
```
KevinOConnor, line 63: What does this do?  What is 0x8000?

KevinOConnor (commented):
Thanks.  In general it looks okay, but I have some questions - see below.

-Kevin

src/stm32/stm32g0.c
leptun: 0x8000 is the offset from the APB bus base address at which the peripheral after LPTIM1 starts. The issue is that we don't have LPTIM1 on the G0x0, so we can't use that peripheral base address in the if statement.
As a further note, this if statement selects bits from a 32 bit register that controls power where the peripheral address neatly maps to the power bit. As such, at most 32 peripherals can be controlled by that register. Each peripheral is 0x400 in length in memory (see a few lines lower). 0x8000 is 0x400 * 32.
image

src/stm32/stm32f0_timer.c
leptun: The error was added as a prevention in case I messed something up with the timers definition. I'll change the ifdefs for more clarity.

src/stm32/stm32f0_serial.c
leptun: The G0x0 family is missing some (or most) of the lower power peripherals. As such, these IRQ vectors that are combined might not be like that on the G0x0 family. In this case, the G0b0 doesn't have LPUART2, so the USART2 vector is just `USART2_IRQn`

leptun: Unfortunately it can't be written as instructed since the IRQn thingies are actually part of an enum, not defines. So I had to revert to the old code. I did change the comment a bit however. Hopefully this makes the change a bit more clear.

src/stm32/stm32f0_timer.c
leptun: I removed the error case and added indentation. The same issue as with the LPUART2 also applies to the IRQn here.

KevinOConnor:
Thanks.  I committed this change.  I made some minor changes on top (commit f31540b3).

-Kevin

leptun:
:tada:. I'll rebase #5838 on top of these changes as soon as I have time.

PR #5929, opened 2022-12-10
Add bootloader 64Kib for Lerdge Board

DarkSoulRL:
I tested the Lerdge X board. The main functionality works. Currently there is no screen and touchscreen support.
Links for instructions and files:

https://www.lerdge.com/document/detail/Klipper-tutorial1
https://www.lerdge.com/public/upload/file/Klipper%20Firmware%20for%20Lerdge%20&%20Klipper%20printer%20config.zip

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -235,6 +235,8 @@ choice
         bool "48KiB bootloader (MKS Robin Nano V3)" if MACH_STM32F4x5
     config STM32_FLASH_START_10000
         bool "64KiB bootloader" if MACH_STM32F103 || MACH_STM32F446 || MACH_STM32F401
+    config STM32_FLASH_START_10000
+        bool "64KiB bootloader (Lerdge Board)" if MACH_STM32F4
 
     config STM32_FLASH_START_800
         bool "2KiB bootloader (HID Bootloader)" if MACH_STM32F103


Sineos:
While I appreciate that Lerdge is at least partially moving away from their dubious firmware policy, I still think this is not a good candidate for main line:
* Still requires a tool to encrypt the firmware
* Tool to do so is closed source
* Tool is Windows only

Maybe a post in the config section of Klipper discourse is the better place.

KevinOConnor:
Thanks.  I agree we should not make mention of Lerdge in the comments if the boards are not actually meaningfully supported.

That said, I don't see a harm in adding a 64KiB bootloader offset (without comment to board) if that helps users.

-Kevin

dolence:
I'm trying to compile the Lerdge K version Klipper firmwar as per instructions on Lerdge site. Firmware compile without errors but no .bin file is being generated. I think the kconfig could have changes since september 2022. Any thoughts?

docgalaxyblock:
Hey @KevinOConnor , I would like to redo a F407 64kbit bl PR.
The current Kconfig is: `bool "64KiB bootloader" if MACH_STM32F103 || MACH_STM32F446 || MACH_STM32F401`

should I just add `MACH_STM32F407` there or replace `MACH_STM32F446 || MACH_STM32F401` with `MACH_STM32F4` to enable it for the whole family?

Greetings 🎃

KevinOConnor:
Probably best to enable for all F4 chips.

Cheers,
-Kevin

PR #5930, opened 2022-12-10
Docs: make pico/mpu example cfg match example wiring

xommmmmmmm:
The [wiring diagram example](https://github.com/Klipper3d/klipper/blob/228259da1e74f0a24788d1cead23a8dabaf9cb38/docs/Measuring_Resonances.md) for pico+mpu shows i2c0a, but config example shows i2c1a and will cause timeout.

Additionally, pin -> pins in [current config reference](https://www.klipper3d.org/Config_Reference.html#static_digital_output).

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -195,7 +195,7 @@ serial: /dev/serial/by-id/<your PICO's serial ID>
 
 [mpu9250]
 i2c_mcu: pico
-i2c_bus: i2c1a
+i2c_bus: i2c0a
 
 [resonance_tester]
 accel_chip: mpu9250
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit directly - see https://www.klipper3d.org/CONTRIBUTING.html .

@dmbutyugin , @dingyifei - fyi.

-Kevin

dmbutyugin:
@xommmmmmmm Thanks for fixing it. I noticed the bus inconsistency, but alas it wasn't addressed in the original PR. So, modulo the Signed-off-by line, the change looks good to me.

PR #5932, opened 2022-12-11
scripts: Python 3 compatability

junland:
Here's my third attempt. I excluded `whconsole.py` and `buildcommands.py` as those two scripts would take a little more time / work to get migrated over. Also it was suggested in the last PR that setting the she bang to just python (Which most likely be symlinked to python3) so I did just that and left the scripts alone that explicitly required python2 for compatibility sake.

### DIFF
--- a/docs/Packaging.md
+++ b/docs/Packaging.md
@@ -14,7 +14,7 @@ klippy/chelper/__init__.py`.
 ## Compiling python code
 
 Many distributions have a policy of compiling all python code before packaging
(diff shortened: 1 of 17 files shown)


KevinOConnor:
Thanks.  Same comments as before though - don't mix whitespace changes with functional changes (and in general, whitespace changes will not be accepted).  Also, I need confirmation that each tool that is changed has been confirmed to actually work on python3.

Cheers,
-Kevin

PR #5933, opened 2022-12-12
feat: Adds Support for DGUS DWIN touch screens

gbkwiatt:
This is based on https://github.com/Desuuuu/klipper 
I've only resolved conflicts with latest klipper. 

I've also tested and it works fine with proper dwin screen, I am planning to add more support for switching/parking  dual extruders. But first the main work should be merged. 

I believe there was an attempt already to get that merge, but not sure why it wasn't ? 
But I am happy to work on it so it get merged eventually, as I know many people are starting to use that screen more often and it would be very beneficial to get that done. 

Here is a screen firmware that works with it
https://github.com/Desuuuu/DGUS-reloaded-Klipper

However firmware shouldn't be considered a part of it, since it can be adapted and created for everyone's needs. e.g. I've modified the whole firmware to be vertical not horizontal and added elements for dual extruders.

I am not entirely sure, if all the changes are needed, so please if someone can have a look. 

@Desuuuu  - I know there was some extra work in progress, to make klipper generate screen menu from klipper rather than being hardcoded in screen firmware, however that work was stale and was never finished, and this approach here I believe is good enough to be  a part of klipper.


manu7irl:
Thomasz, is this using dgus-display branch that helps with creating a
dynamic menu structure?
I have been using that on numerous occasions, on different printers for
more than a year no issues at all, and I think it is better than the first
attempt from desuuuu to make the display completely static.
What do you think?

gbkwiatt:
@manu7irl it's using his first attempt with static display firmware. I've read about his dynamic approach and if I am not mistaken it's missing some core features because it was work in progress and waa never finished as states in comments here https://github.com/Desuuuu/klipper/discussions/74

manu7irl:
@gbkwiatt, I use the new method and it works flawless for me for more than a year I tried also the previous one and I prefer @Desuuuu second implementation! because you are free to choose which uart to choose at run time, and no need to recompile the micro firmware anytime you change something, and the new config section and menu structure allows to edit menus dynamically, and that is gold for me! you should consider rebase on that and do the portrait thing there.
if you need me to share input or pictures let me know!
thanks for your time!

Sineos:
Did you align posting his code here as PR with @Desuuuu?

gbkwiatt:
> Did you align posting his code here as PR with @Desuuuu?

No, but since it's open source I can't see why it shouldn't be a part of klipper - by the look of it many people are using it forked anyway.

KevinOConnor:
This work would need to follow the copyright requirements described at https://www.klipper3d.org/CONTRIBUTING.html - in particular a clear copyright statement and a signed-off-by line.

The change would also need a `git rebase` and a developer committing to be a long-term owner of it.  I don't use these lcd-type screens so it is hard for me to review changes for them.

Cheers,
-Kevin

klippy/extras/t5uid1/dgus_reloaded/__init__.py
```diff
@@ -0,0 +1,110 @@
+# Package definition for the extras/t5uid1/dugs_reloaded directory
```
ammmze, line 1: Nit: typo...dugs vs dgus

theodonkulus:
I think this could actually use a reviewer, Stumbled upon this as starting to do my own changes. Maybe its worth contacting the original dev so things don't get out of sync?

theodonkulus:
@KevinOConnor What would be needed to get this pushed into mainline?

PR #5934, opened 2022-12-12
homing: add event to signal homed status to extras.

mattthebaker:

This commit adds a homed event that extras can act on.

Its initial use is to enable a displacement sensor based homing routine using the Beacon eddy current sensor. In the routine, the machine is first homed through standard procedures, the event indicates completion, and then a native displacement is measured at stationary z and applied to the machines position. This allows fast dive speeds for closing the gap, followed by high precision measurement at fixed z afterwards.

There are presently around 70 units of the new sensor in private circles and many more being shipped out around the end of the year. Klipper is the only supported firmware, so it will surpass the 100 threshold readily.

This is the only required change outside of a modularized extras module to support the sensor. Adding the signal will allow the module to be easily managed externally until the implementation is stable enough for mainline integration.

### DIFF
--- a/klippy/extras/homing.py
+++ b/klippy/extras/homing.py
@@ -266,6 +266,9 @@ class PrinterHoming:
         kin = self.printer.lookup_object('toolhead').get_kinematics()
         try:
             kin.home(homing_state)
+            axes_str = "".join([a for (a, i) in zip("xyz", [0, 1, 2])
+                                if i in axes])
+            self.printer.send_event("homing:homed", axes_str)
         except self.printer.command_error:
             if self.printer.is_shutdown():
                 raise self.printer.command_error(


KevinOConnor:
Thanks.  As a general rule though, I'm leery of adding incomplete code to Klipper.  I've found that, in other projects, doing that resulted in a consistent long-term maintenance burden.  It's not practical to meaningfully review the contributions, not practical to test it when making other code changes, and not practical to determine if the external projects still rely on it.

All that said, it doesn't look like the added event is needed here (and it doesn't look like gcode_move.py is the right place to generate the event).  The homing.py code already generates a series of events ("homing:home_rails_begin", "homing:home_rails_end", "homing:homing_move_begin", and "homing:homing_move_end") that are already used to do things similar to what you describe.  In particular, take a look at how endstop_phase.py alters the printer position based on the "homing"homing_rails_end" event.

Cheers,
-Kevin

dalegaard:
Hi,

On reading through `endstop_phase.py` it appears to adjust in stepper space. That's fine for a Cartesian but no good on a delta.

Once we've homed Z we are interested in performing an override of the cartesian space Z coordinate based on what the probe is measuring. I wasn't able to find a way to do this in the current code, as all the events you mention are only able to affect stepper space, not cartesian space. If I've misunderstood one of these, a pointer to how we can override Z from one of these events would be helpful.

Best regards,
Lasse

dalegaard:
Hi,

That was the conclusion I came to back when we wrote the code. Adding a single well defined event seemed like the most clean way to accomplish the task. Expanding the current events would complicate the code much more and add rarely exercised code paths(given that nothing in the current codebase would use the extensions).

We can go whichever path you prefer, we just need to know what that path is.

Best regards,
Lasse

theolas:
Hello, hoping this can be merged soon, this will greatly benefit myself and my customers.

mattthebaker:
Would like to revisit this pull request now that the code which requires it is open.

The reason why we want the signal still stands, and seems unique to the requirements of a direct displacement measuring sensor.

How the signal is used can be seen here: https://github.com/beacon3d/klipper/blob/master/klippy/extras/beacon.py#L1145

With the homing signal or a close equivalent the beacon code can be completely compartmentalized to the single python module. It will make it much easier for us to directly support the users and fix any issues that come up.

KevinOConnor:
Thanks for the update.  In general, I don't see an issue with adding an event to support a new module when it is added to Klipper.  However, as per my earlier comments, I'm not in favour of adding code to Klipper for external code.  My experience is that creates a maintenance burden.

-Kevin

PR #5936, opened 2022-12-13
Rework usb_reboot_for_dfu_bootloader() so it works on stm32h723

KevinOConnor:
I found that disabling d-cache and moving the dfu reboot flag to "axi sram" was sufficient for me to enter the bootloader on the stm32h723.  Without either of those changes, I was unable to enter the bootloader.

@bigtreetech - fyi.

-Kevin

### DIFF
--- a/src/stm32/stm32h7.c
+++ b/src/stm32/stm32h7.c
@@ -208,7 +208,7 @@ clock_setup(void)
  * Bootloader
  ****************************************************************/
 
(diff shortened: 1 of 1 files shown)


bigtreetech:
In the actual test, this change is programmed 10 times continuously, and 100% downloaded successfully

KevinOConnor:

-Kevin

KevinOConnor:
On further testing, it seems it is sufficient to clean the dcache instead of fully disabling it.  I made that change in commit 0c43ae1f.

-Kevin

PR #5937, opened 2022-12-13
Fix G-Code arc parsing code

tehmaze:
The changes in #5910 resulted in wrong behaviour, for `G2` and friends, a zero value for `I` and `J` are valid and should be accepted for input. However, the values should not be omitted from the G-Code.

Signed-off-by: Wijnand Modderman-Lenstra <maze@pyth0n.org>

### DIFF
--- a/klippy/extras/save_variables.py
+++ b/klippy/extras/save_variables.py
@@ -12,6 +12,8 @@ class SaveVariables:
         self.filename = os.path.expanduser(config.get('filename'))
         self.allVariables = {}
         try:
+            if not os.path.exists(self.filename):
+                open(self.filename, "w").close()
             self.loadVariables()
         except self.printer.command_error as e:
             raise config.error(str(e))


tehmaze:
The validity of this case is also illustrated in the [CNC Cookbook examples](https://www.cnccookbook.com/cnc-g-code-arc-circle-g02-g03/):

> This arc starts at X0Y2 and finishes at X2Y0. It’s center is at X0Y0. We could specify it in g-code like this:
> ```
> G02 (Set up the clockwise arc mode)
> X2Y0 I0J-2.0
> ```

KevinOConnor:
@ajmirsky - your thoughts on this?

-Kevin

fbeauKmi:
@tehmaze . Nice one, but do not forget that I J or K can be omitted, the example you bring can be written
```G2 X2Y0 J-2.0```
and it should work.
Probably, just change in line 85 will fix the issue.
```if not (asPlanar[0]  or asPlanar[1]) : ```
What's your thoughts ?

**Comment by tehmaze** [comment]
@fbeauKmi wrote:
> do not forget that `I` `J` or `K` can be omitted

That would require different logic, since it now generates `None` values where I would assume the default null-value should be `0`?

**Comment by ajmirsky** [comment]
@tehmaze thank you for fixing!

**Comment by KevinOConnor** [comment]
Thanks.  I'm leery of mixing a bugfix with code reformatting in a single commit.  Is the bugfix:
```diff
--- a/klippy/extras/gcode_arcs.py
+++ b/klippy/extras/gcode_arcs.py
@@ -84,7 +84,7 @@ class ArcSupport:
             asPlanar = [ gcmd.get_float(a, 0.) for i,a in enumerate('JK') ]
             axes = (Y_AXIS, Z_AXIS, X_AXIS)
 
-        if not asPlanar[0] or not asPlanar[1]:
+        if not (asPlanar[0] or asPlanar[1]):
             raise gcmd.error("G2/G3 requires IJ, IK or JK parameters")
 
         asE = gcmd.get_float("E", None)
```

-Kevin

SoftFever:
> 

I was about to submit a PR fo this regression, then I found this PR :)
Yeah, I believe `if not asPlanar[0] and not asPlanar[1]:` should fix the regression.

-SoftFever

tehmaze:
Sorry for the noise, since I can't re-open this PR, continuation in #5939

TravisWilder:
for reference: it is now https://github.com/Klipper3d/klipper/pull/5939

PR #5938, opened 2022-12-13
Simplify and unify stm32h7_adc code

KevinOConnor:
The stm32h7_adc.c code is used on a number of chips now - stm32g431, stm32l412, stm32h723, stm32h743, and stm32h750.  The adc code had a lot of ifdefs and slightly different behavior between chips.  Most of these differences were unnecessary.  This PR reworks the code to use less ifdefs.

I found and (hopefully) fixed a few bugs while doing this:
1. The stm32l412 should not use CKMODE=0.
2. On stm32h7 the adc actually runs at 25Mhz so the code should use BOOST=0b10.
3. The ISR clear of ADRDY should use explicit writes instead of read-modify-write.

I have also reworked the code to use explicit register writes where practical (instead of read-modify-write).

I have tested this on stm32h723.  I don't have stm32l4, stm32g4, stm32h743, nor stm32h750 chips to test.

@D4SK , @adelyser, @mattthebaker, @bigtreetech - FYI.

-Kevin

### DIFF
--- a/src/stm32/stm32g4.c
+++ b/src/stm32/stm32g4.c
@@ -43,6 +43,10 @@ lookup_clock_line(uint32_t periph_base)
                               .bit = 1 << pos};
 
     } else {
+        if (periph_base == ADC12_COMMON_BASE)
+            return (struct cline){.en = &RCC->AHB2ENR,
+                                  .rst = &RCC->AHB2RSTR,
+                                  .bit = RCC_AHB2ENR_ADC12EN};
         uint32_t pos = (periph_base - AHB2PERIPH_BASE) / 0x400;
         return (struct cline){.en = &RCC->AHB2ENR,
                               .rst = &RCC->AHB2RSTR,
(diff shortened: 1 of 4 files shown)


KevinOConnor:
In further testing, I found that avoiding some of the read-modify-write operations caused issues on adc1/adc2 of the stm32h723 (due to linear calibration flags in the CR register having side effects).  I've backed out that part of the change and rebase'd this branch.

-Kevin

bigtreetech:
I have tested on SKR 3 (both H743 and H723).  NTC thermistor results are normal,  But `temperature_mcu` of H723 will be about 6 ℃ higher than that of `master` branch

KevinOConnor:
> But temperature_mcu of H723 will be about 6 ℃ higher than that of master branch

Thanks.  It looks like this change in behaviour got introduced in commit `stm32: No need to use hardware oversampling in stm32h7_adc.c`.  However, it seems the issue is not related to oversampling, but due to insufficient time to sample the temperature sensor.  The stm32h723 datasheet states the sensor should have a sample time between 9us to 25.2us, but the code samples for 1.85us.  This timing is unchanged from the previous code, but I'm guessing the oversampling changed the characteristics of the results.

It looks like all of the chips using stm32h7_adc.c have a too low sampling time for the temperature sensor (g4 is 1.267 vs 5us, l4 is 2.375 vs 5us, h743/h750 is 2.58 vs 9us, h723 is 1.85 vs 9us).

I'm inclined to leave this PR as is and look to address the mcu temperature sampling time as a separate topic.

-Kevin

KevinOConnor:
I have rebased this PR to the latest master code.  I have also added a new commit to this series that increases the sampling time on all chips.  This should make all the chips compliant with the minimum sample time needed for the internal mcu temperature sensor.

I've tested this series on an stm32h723 chip.  I'm inclined to go forward with this series (even though it may impact stm32g4, stm32l4), as there are a number of fixes now on this PR.

-Kevin

PR #5939, opened 2022-12-14
Allow either one of I, J, K to be default-zero on G2/G3

tehmaze:
Implemented the suggested comments from @KevinOConnor and @ajmirsky in #5937 which I accidentally closed with a wrong force-push.

Signed-off-by: Wijnand Modderman-Lenstra <maze@pyth0n.org>


KevinOConnor:
Thanks.

-Kevin

PR #5941, opened 2022-12-15
workflows: Update various actions versions

junland:
PR bumps the versions of a couple of workflows actions. `build-test.yml` ran just fine, see [this ](https://github.com/junland/klipper/actions/runs/3707390702/jobs/6283738263)for reference. Also `stale-issue-bot.yml` had an update for `actions/stale`  from v3 to v6.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -6,10 +6,10 @@ jobs:
   build:
     runs-on: ubuntu-18.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Setup cache
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: ci_cache
         key: ${{ runner.os }}-build-${{ hashFiles('scripts/ci-install.sh') }}
@@ -21,7 +21,7 @@ jobs:
       run: ./scripts/ci-build.sh 2>&1
 
     - name: Upload micro-controller data dictionaries
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         name: data-dict
         path: ci_build/dict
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  What advantages are there in making this change?

-Kevin

junland:
* `actions/cache` gets small bug fixes. Nothing too drastic looking at the various releases leading up to the next major.
* `actions/setup-python` upgrades the default node version to 16 as the current one in v2 is EOL + lots of bug fixes.
* `actions/checkout` upgrades the default node version to 16 as well + updated core dependencies + bugfixes

KevinOConnor:
Okay, thanks.  However, I'd prefer not to introduce changes to the repo unless it provides some concrete feature that we need.

Cheers,
-Kevin

Arksine:
I noticed this PR in my inbox, so I thought it would be prudent to bring up that GitHub has [deprecated the old actions](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).  Presumably they will set a concrete date soonish, after which they will no longer function.  There are other deprecations as well (Ubuntu 18.04, the `set-output` command, etc), the result from a recent run displays the various [warnings](https://github.com/Klipper3d/klipper/actions/runs/3856420049).

junland:
> I noticed this PR in my inbox, so I thought it would be prudent to bring up that GitHub has [deprecated the old actions](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/). Presumably they will set a concrete date soonish, after which they will no longer function. There are other deprecations as well (Ubuntu 18.04, the `set-output` command, etc), the result from a recent run displays the various [warnings](https://github.com/Klipper3d/klipper/actions/runs/3856420049).

Oh yea, I almost forgot to mention that. Thanks for mentioning that.

@KevinOConnor would you accept this PR if I resubmit since it's closed? (Unless you can reopen it again and accept it.) Think the deprecation fixes and keeping up with the latest action modules is more than enough to get this into main.

PR #5944, opened 2022-12-17
Added precise location of [bed_screws] for Ender-3-s1 

fire1ce:
Added precise location of [bed_screws] for Ender-3-s1 config for better manual bed leveling. 

[bed_screws] locations points the tip of the nozzel above the location of the manual bed leveling scews for optimal leveling.

Signed-off-by: Stas Yakobov <stas@3os.org>

### DIFF
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -131,7 +131,7 @@ runout_gcode: PAUSE
 recover_velocity: 25
 
 [bed_screws]
-# Ender s1
+# Ender-3-s1 bed
 screw1: 20, 29
 screw2: 195, 29
 screw3: 195, 198
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -129,3 +129,10 @@ runout_gcode: PAUSE
 
 [pause_resume]
 recover_velocity: 25
+
+[bed_screws]
+# Ender s1
+screw1: 20, 29
+screw2: 195, 29
+screw3: 195, 198
+screw4: 20, 198
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -131,7 +131,6 @@ runout_gcode: PAUSE
 recover_velocity: 25
 
 [bed_screws]
-# Ender-3-s1 bed
 screw1: 20, 29
 screw2: 195, 29
 screw3: 195, 198


KevinOConnor:
@Fisheiyy - any thoughts on this?

-Kevin

Fisheiyy:
@fire1ce Is this necessary? I would like to keep the config as barebones as possible to reduce new-comer confusion, and the bed screws thing is confusing for me, because of the clock representation of rotations, so do you think a new user would understand it and be able to use this feature?

fire1ce:
@Fisheiyy I think u are looking at this from the wrong perspective. Let me clarify, Klipper firmware as a base line isn’t for the “noobies” also when i was first installing klipper i was looking for the most complete config there is. Which made me go looking for other ppl configs that i weren’t sure are even correct since the Official ones didn’t had some stuff. For the point of adding the Bed Screws position, Its a hardware part of this printer, not really something a user should change unless he modded the printers bed. At my point any Default Hardware configuration should be present at klipper initial config. Thats my opinion from my experience.

fire1ce:
Also i wanted to add that I've looks for those settings at probably more then 20 other people configs, there where all wrong. So i had to take the positions manually

Sineos:
Given that the location is correct and according to standard Ender 3 S1 then it definitively makes sense in my opinion.
Bed leveling is an important task and this makes it easier

config/printer-creality-ender3-s1-2021.cfg
```diff
@@ -129,3 +129,10 @@ runout_gcode: PAUSE
 
 [pause_resume]
 recover_velocity: 25
+
+[bed_screws]
+# Ender-3-s1 bed
```
Fisheiyy, line 134: Remove this and this can be merged.

Fisheiyy (changes requested):

KevinOConnor:
Thanks.  The change looks fine to me.  However, commits need to have a signed-off-by line - https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

fire1ce:
> Thanks. The change looks fine to me. However, commits need to have a signed-off-by line - https://www.klipper3d.org/CONTRIBUTING.html .
> 
> -Kevin

Hey Kevein, thanks for pointing this out. 
I have edited the first comment with the relevant information.

KevinOConnor:
Thanks.

-Kevin

Littlericket:
@KevinOConnor @Fisheiyy 

How about having screws_tilt_adjust instead of bed_screws, since e3s1 is equipped with zprobe by default anyways.

```
[screws_tilt_adjust]
screw1: 50, 67
screw1_name: front left
screw2: 226, 67
screw2_name: front right
screw3: 226, 225
screw3_name: back right
screw4: 50, 225
screw4_name: back left
horizontal_move_z: 10.
speed: 50.
screw_thread: CW-M3
```

PR #5945, opened 2022-12-18
scripts: support CanBoot on RP2040 in `make flash`

clee:
Allows `make flash` to work successfully with an RP2040 running CanBoot instead of the stock bootloader.

### DIFF
--- a/src/rp2040/Makefile
+++ b/src/rp2040/Makefile
@@ -45,7 +45,7 @@ $(OUT)klipper.uf2: $(OUT)klipper.elf $(OUT)lib/rp2040/elf2uf2/elf2uf2
 	@echo "  Creating uf2 file $@"
 	$(Q)$(OUT)lib/rp2040/elf2uf2/elf2uf2 $< $@
 
-target-$(CONFIG_RP2040_HAVE_STAGE2) += $(OUT)klipper.uf2
+rptarget-$(CONFIG_RP2040_HAVE_STAGE2) := $(OUT)klipper.uf2
 rplink-$(CONFIG_RP2040_HAVE_STAGE2) := $(OUT)src/rp2040/rp2040_link.ld
 stage2-$(CONFIG_RP2040_HAVE_STAGE2) := $(OUT)stage2.o
 
@@ -54,10 +54,11 @@ $(OUT)klipper.bin: $(OUT)klipper.elf
 	@echo "  Creating bin file $@"
 	$(Q)$(OBJCOPY) -O binary $< $@
 
-target-$(CONFIG_RP2040_HAVE_BOOTLOADER) += $(OUT)klipper.bin
+rptarget-$(CONFIG_RP2040_HAVE_BOOTLOADER) := $(OUT)klipper.bin
 rplink-$(CONFIG_RP2040_HAVE_BOOTLOADER) := $(OUT)src/generic/armcm_link.ld
 
 # Set klipper.elf linker rules
+target-y += $(rptarget-y)
 CFLAGS_klipper.elf += --specs=nano.specs --specs=nosys.specs -T $(rplink-y)
 OBJS_klipper.elf += $(stage2-y)
 $(OUT)klipper.elf: $(stage2-y) $(rplink-y)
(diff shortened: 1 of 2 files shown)


clee:
Okay, I figured out a way to get `make flash` to work, but I'm not sure if this is the right approach. Any thoughts?

```diff
--- a/src/rp2040/Makefile
+++ b/src/rp2040/Makefile
@@ -67,6 +67,12 @@ lib/rp2040_flash/rp2040_flash:
     @echo "  Building rp2040_flash"
     $(Q)make -C lib/rp2040_flash rp2040_flash
 
+ifeq ($(CONFIG_RP2040_HAVE_BOOTLOADER),y)
+flash: $(OUT)klipper.bin
+    @echo "  Flashing $< to $(FLASH_DEVICE)"
+    $(Q)$(PYTHON) ./scripts/flash_usb.py -t $(CONFIG_MCU) -d "$(FLASH_DEVICE)" $(if $(NOSUDO),--no-sudo) $(OUT)k
lipper.bin
+else
 flash: $(OUT)klipper.uf2 lib/rp2040_flash/rp2040_flash
     @echo "  Flashing $< to $(FLASH_DEVICE)"
     $(Q)$(PYTHON) ./scripts/flash_usb.py -t $(CONFIG_MCU) -d "$(FLASH_DEVICE)" $(if $(NOSUDO),--no-sudo) $(OUT)klipp
er.uf2
+endif
```

KevinOConnor:
You'll want something like this in the Makefile:
```diff
--- a/src/rp2040/Makefile
+++ b/src/rp2040/Makefile
@@ -45,7 +45,7 @@ $(OUT)klipper.uf2: $(OUT)klipper.elf $(OUT)lib/rp2040/elf2uf2/elf2uf2
        @echo "  Creating uf2 file $@"
        $(Q)$(OUT)lib/rp2040/elf2uf2/elf2uf2 $< $@
 
-target-$(CONFIG_RP2040_HAVE_STAGE2) += $(OUT)klipper.uf2
+rptarget-$(CONFIG_RP2040_HAVE_STAGE2) := $(OUT)klipper.uf2
 rplink-$(CONFIG_RP2040_HAVE_STAGE2) := $(OUT)src/rp2040/rp2040_link.ld
 stage2-$(CONFIG_RP2040_HAVE_STAGE2) := $(OUT)stage2.o
 
@@ -54,10 +54,11 @@ $(OUT)klipper.bin: $(OUT)klipper.elf
        @echo "  Creating bin file $@"
        $(Q)$(OBJCOPY) -O binary $< $@
 
-target-$(CONFIG_RP2040_HAVE_BOOTLOADER) += $(OUT)klipper.bin
+rptarget-$(CONFIG_RP2040_HAVE_BOOTLOADER) := $(OUT)klipper.bin
 rplink-$(CONFIG_RP2040_HAVE_BOOTLOADER) := $(OUT)src/generic/armcm_link.ld
 
 # Set klipper.elf linker rules
+target-y += $(rptarget-y)
 CFLAGS_klipper.elf += --specs=nano.specs --specs=nosys.specs -T $(rplink-y)
 OBJS_klipper.elf += $(stage2-y)
 $(OUT)klipper.elf: $(stage2-y) $(rplink-y)
@@ -69,4 +70,4 @@ lib/rp2040_flash/rp2040_flash:
 
 flash: $(OUT)klipper.uf2 lib/rp2040_flash/rp2040_flash
        @echo "  Flashing $< to $(FLASH_DEVICE)"
-       $(Q)$(PYTHON) ./scripts/flash_usb.py -t $(CONFIG_MCU) -d "$(FLASH_DEVICE)" $(if $(NOSUDO),--no-sudo) $(OUT)klipper.uf2
+       $(Q)$(PYTHON) ./scripts/flash_usb.py -t $(CONFIG_MCU) -d "$(FLASH_DEVICE)" $(if $(NOSUDO),--no-sudo) $(rptarget-y)
```

Unless I'm missing something, this would need to go in the same commit as the flash-usb.py changes, as otherwise users would likely flash the wrong thing to their boards?

-Kevin

clee:
> Unless I'm missing something, this would need to go in the same commit as the flash-usb.py changes, as otherwise users would likely flash the wrong thing to their boards?

Tested your Makefile approach on my SKR Pico and it worked here, so I'm calling it good. Merged that into my previous commit and re-pushed.

src/rp2040/Makefile
```diff
@@ -67,6 +68,7 @@ lib/rp2040_flash/rp2040_flash:
 	@echo "  Building rp2040_flash"
 	$(Q)make -C lib/rp2040_flash rp2040_flash
 
-flash: $(OUT)klipper.uf2 lib/rp2040_flash/rp2040_flash
+# Flash rules
```
KevinOConnor, line 71: Dup'ing this comment looks odd.

-Kevin

clee: … not entirely sure how that snuck in, but it's gone now!

KevinOConnor:
Thanks.

-Kevin

PR #5946, opened 2022-12-18
mcu: fresh commit and PR for HC32F460

SteveGotthardt:
Started over with just mcu and copied files to new branch.
This is just the mcu/library/src portion. I have configs, test, LCD fix to come...

Signed-off-by: Steven Gotthardt [gotthardt@gmail.com](mailto:gotthardt@gmail.com)


KevinOConnor:
Okay, thanks.  It's going to be very hard for me to review this PR though as it's 98 new files all bundled together as one.  It would be preferable if this could be split up into at least two commits (in the same PR) - one adding the lib/ files and one adding the klipper specific code.

That said, I'll try to take a look at this PR in the coming days.

-Kevin

src/hc32f460/Makefile
```diff
@@ -0,0 +1,38 @@
...
+src-y += hc32f460/interrupts.c
+src-y += hc32f460/gpio.c
+src-y += ../lib/hc32f460/mcu/common/system_hc32f460.c
+src-y += ../lib/hc32f460/driver/src/hc32f460_clk.c
+src-y += ../lib/hc32f460/driver/src/hc32f460_efm.c
+src-y += ../lib/hc32f460/driver/src/hc32f460_sram.c
+src-y += ../lib/hc32f460/driver/src/hc32f460_utility.c
+src-y += ../lib/hc32f460/driver/src/hc32f460_gpio.c
+src-y += ../lib/hc32f460/driver/src/hc32f460_pwc.c
+src-$(CONFIG_HAVE_GPIO_I2C) += hc32f460/i2c.c../lib/hc32f460/driver/src/hc32f460_i2c.c
```
KevinOConnor, line 25: This looks like dead code.  (There is no i2c.c file nor is i2c enabled in Kconfig.)  To avoid confusion, best to remove these lines until i2c support is ready.

src/hc32f460/adc.c
```diff
@@ -0,0 +1,155 @@
...
+
+
+struct gpio_adc
+gpio_adc_setup(uint32_t gpio)
+{
+    // validate pin in adc_pins table
+    int chan;
+    for (chan=0; ; chan++)
+    {
+        if (chan >= ADC1_CH_COUNT-1)   // don't allow internal
```
KevinOConnor, line 67: FWIW, I think the code would be more readable if this did `if (chan >= ARRAY_SIZE(adc_gpio))` as otherwise it's not clear the following access to `adc_gpio` is valid.

src/hc32f460/serial.c
```diff
@@ -0,0 +1,189 @@
...
+  #define USART_RX_PORT   (PortA)
+  #define USART_RX_PIN    (Pin07)
+  #define USART_TX_PORT   (PortA)
+  #define USART_TX_PIN    (Pin08)
+#endif
+
+// use USART 1 for serial connection
+#define USARTx          M4_USART1
+#define USART_ENABLE    PWC_FCG1_PERIPH_USART1 | PWC_FCG1_PERIPH_USART2 | \
+                        PWC_FCG1_PERIPH_USART3 | PWC_FCG1_PERIPH_USART4
```
KevinOConnor, line 57: Almost certainly this define should be wrapped in parenthesis.

KevinOConnor (commented):
Thanks.  In general it looks okay to me.  I have a few minor comments (see below).  Can you also confirm that this PR is still ready to merge from your side.

-Kevin

src/hc32f460/Makefile
SteveGotthardt: Thanks - I removed that

src/hc32f460/adc.c
SteveGotthardt: I incorporated that

src/hc32f460/serial.c
SteveGotthardt: wrapped

KevinOConnor:
Thanks.  I squashed this PR into two commits and merged them (94cbf5ff and 72b6bd7e).

-Kevin

PR #5947, opened 2022-12-19
safe_z_home: Add hooks to safe_z_home

ibash:
This adds hooks to safe_z_home which allow running custom gcode before/after each axis is homed.

I need to move the toolhead a little after homing x to prevent a crash. I also have a euclid probe, which means I need to deploy/retract the probe. Rather than recreating the z-hop logic in gcode (which isn't totally possible anyway, there's no way to mark z as not homed in gcode). I found it to be simpler to add hooks in safe_z_home.

This allows me a nice short config and is safer than homing override. e.g.

```
[safe_z_home]
home_xy_position: 150,150
speed: 25
z_hop: 15
z_hop_speed: 5
after_x:
  # move z a little to prevent crashing into a magnum pulley
  G1 X20 F3600
before_z:
  M401
after_z:
  M402
```

Signed-off-by: Islam Sharabash <islam.sharabash@gmail.com>

### DIFF
--- a/klippy/extras/safe_z_home.py
+++ b/klippy/extras/safe_z_home.py
@@ -3,7 +3,6 @@
 # Copyright (C) 2019 Florian Heilmann <Florian.Heilmann@gmx.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-
 class SafeZHoming:
     def __init__(self, config):
         self.printer = config.get_printer()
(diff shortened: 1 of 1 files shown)


ibash:
I can understand that, however I think it's becoming more common do things like deploy a probe before/after z.

Reproducing all the functionality safe_z_home does in homing_override would require some code changes to klipper to let gcode call `note_z_not_homed`. Even then, it's a huge PIA to implement since you need to parse out the requested axis and handle them separately. I know this because I tried implementing this in homing_override first, then made a code change to expose additional functionality to homing_override.

Adding hooks into safe_z_home was the most elegant and easy version.

KevinOConnor:
I understand.  However, I don't currently think this is a good candidate for the Klipper master branch.  There are already several ways to customize the homing process (eg, homing_override, overriding the G28 macro, using macros to home and avoiding G28 entirely).  I'm concerned that adding yet another way to customize homing increases complexity.

-Kevin

ibash:
Definitely valid concern, and I really do like that klipper keeps things simple. I'm happy to maintain my fork since I find it useful.

That said I think this change would allow the voron switchwire, legacy, v1, and v2 to all use `safe_z_home` instead of `homing_override`. Would need input from that community since I don't have a voron.

Potentially can simplify this monster of a `homing_override`: https://github.com/jlas1/Klicky-Probe/blob/2bfb9bff0f7248ce583f53de4798b05bfb2be10c/Klipper_macros/klicky-macros.cfg#L566

Maybe cc @jlas1 to get an opinion?

PR #5951, opened 2022-12-23
Set TRSYNC_TIMEOUT to 0.050 instead of 0.025

alexownsall:
This change is to support a slightly longer timeout for Multi-MCU setups, specifically in Can Bus scenarios where homing timeouts occur. Doesn't seem to cause any latency.


KevinOConnor:
Dup of #5816 - see comments there.

-Kevin

PR #5953, opened 2022-12-23
Create BTT E3 V3.0 Sovol SV01

parahooners:


JamesH1978:
We do not usually accept configs for a printer with non-stock options for inclusion in the repo. Our examples for a printer should be bone stock as if someone opened it new in a box. This is designed so not to confuse new users when their setup does not match a stock config due to 3rd party options being added. Unless this printer ships with a E3 Mini V3 from the factory? Custom configs are usually posted on the Discourse in the Config section https://klipper.discourse.group/

Thanks
James

PR #5954, opened 2022-12-24
docs: Fix typo in rotation distance

Sud-Puth:
The actual remaining filament distance will be 20mm and not 2mm as 70 - 50 = 20

Signed-off-by: Sudheer Puthana <sudheer@sudheer.co.in>

### DIFF
--- a/docs/Rotation_Distance.md
+++ b/docs/Rotation_Distance.md
@@ -75,8 +75,8 @@ Then use the following procedure to "measure and trim":
    `rotation_distance = <previous_rotation_distance> * <actual_extrude_distance> / <requested_extrude_distance>`
    Round the new rotation_distance to three decimal places.
 
-If the actual_extrude_distance differs from requested_extrude_distance
-by more than about 2mm then it is a good idea to perform the steps
+If the `actual_extrude_distance` differs from `requested_extrude_distance`
+by more than about 20mm then it is a good idea to perform the steps
 above a second time.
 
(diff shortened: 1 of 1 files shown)


jakep82:
This change is incorrect. I think you're confusing the initial mark distance with the requested extrude distance.

Sud-Puth:
> Use the digital calipers to measure the new distance between the extruder body and the mark on the filament

The mark on the filament was 70mm (`initial_mark_distance`), we extruded 50mm(`requested_extrude_distance`), now 20mm should remain. The new distance between the extruder body and the mark will be 20mm cause we extruded 50mm 

Kind of the same principle here - https://teachingtechyt.github.io/calibration.html#esteps (which I have been following for Marlin)
1) 
> we take a permanent marker and put a mark 120mm from the entry to the extruder
2)
> G1 E100 F50
3)
> Ideally, 20mm remains, which means exactly 100mm was extruded. If your distance is anything other than this, complete the form below to calculate the correct E-steps

jakep82:
You're not understanding the instructions. They ask you to compare the requested distance to the actual distance, not the initial mark distance. 

```

If the actual_extrude_distance differs from requested_extrude_distance by more than about 2mm then it is a good idea to perform the steps above a second time.

Sud-Puth:
Ahhh that makes sense, I shouldn't be doing late night shenanigans. Thanks for just pointing that out. Closing the PR. I was so used to the other instructions that I did not process the klipper doc

PR #5955, opened 2022-12-24
pid improvements

dans98:
This PR is related to improving PID control within the constraints Klipper
currently operates under. I've made changes to the calibration algorithm as 
well standard positional form algorithm. Additionally I've added a standard 
velocity form algorithm, that eliminates to potential for integral windup.     

Please see one of the 2 commit messages for greater details, test results, and 
links to relevant papers.

-Dan S.


Sineos:
I'm using both PID functions since quite a while and have not noticed anything negative. Working stable and nicely.

What I find especially useful is the `asymmetry` reading, providing an indication between heating capacity and heat loss. 
This lead to the conclusion that on my two printers the 50W cartridge in combination with the given hotend was too weak. Upgrading both to 70W fixed the asymmetry and are both controlled nicely by the algorithms.

Should @KevinOConnor consider this PR (I hope so), then this should be explained in the documentation.

I wish all of you a happy holiday season!

The-Futur1st:
I have used that for couple of days, works great, provides much better values than it was. 'pid_v' provides an ability to get rid of first big overshoot. My biggest thanks to author.

kejar31:
Tried this today.. works well

celtare21:
[klippy (7).log](https://github.com/Klipper3d/klipper/files/10314003/klippy.7.log)
Getting this when trying to use pid_v

dans98:
@celtare21 I see it, and i will fix that tonight. it looks like you are the first person to test this thats uses m109/m104, as that looks to be what triggered it.

celtare21:
> @celtare21 I see it, and i will fix that tonight. it looks like you are the first person to test this thats uses m109/m104, as that looks to be what triggered it.

Yeah I still use the slicer start/end gcode. Wanting for the fix then i'll try the velocity algorithm too, for now using the improved pid algo temps look both more stable and the overshoot is much lower than default

dans98:
@celtare21 you should be able to test it now!

Maik35:
Ich habe nach einem Klipperupdate das Problem das mein Drucker bei ca. 145C mit dem aufheizen abbricht was auch beim PID des Hotends so ist

freakydude:
> Ich habe nach einem Klipperupdate das Problem das mein Drucker bei ca. 145C mit dem aufheizen abbricht was auch beim PID des Hotends so ist

@Maik35 Hier gehts um was völlig anderes. Am besten du versuchst dein Glück bei Discord in der Klipper Community [https://discord.klipper3d.org/](https://discord.klipper3d.org/)

KevinOConnor:
Thanks.  Very interesting.  I have two high-level comments:
1. On a procedural level, this PR is going to be hard to review and merge as it seems to be mixing functional changes with unrelated changes.  (For example, changing the core algorithms in the same commit that renames variables, makes whitespace changes, changes filenames, and similar.)  It's important that changes be made in small functional chunks so that we can utilize tools like `git bisect` and `git revert`.  Viewing these commits right now it appears to be a "rewrite" of the code.  It's fine if we ultimately obtain totally new versions of the code after several commits.  But in my experience a single bulk replacement of the existing code with new code greatly increases the risk of introducing regressions.
2. It's unclear to me when I should use a `pid` or `pid_v` controller.  I suspect other users would be similarly unsure.  It's also not clear to me if there are a sufficient number of users that would obtain a real-world benefit to a `pid_v` controller (some notes on this at https://www.klipper3d.org/CONTRIBUTING.html ).  So, for the pid_v commit, I suspect we would need some kind of user facing document detailing the tests a user should run to determine if a pid_v controller should be used and the benefits the user should expect from utilizing it.

Cheers,
-Kevin

celtare21:
+1 On the documentation. Would love to see all of it properly documented, comparisons between options etc. As well as what would a "proper" pid tuning session be like.

freakydude:
I totally agree with Kevin. I played around a bit with it. On difference I noticed is, that it takes longer to reach the target temperature but (if it was the calibrated one) it is hold with less tolerance.

I tried it with the head bed... Same result but significant longer time to heat up.

Using a target temperature, not calibrated

dans98:
@KevinOConnor 

> Viewing these commits right now it appears to be a "rewrite" of the code.  It's fine if we ultimately obtain totally new versions of the code after several commits.  But in my experience a single bulk replacement of the existing code with new code greatly increases the risk of introducing regressions.

To an extent this is the case, but I will be happy to answer any questions about the changes i've made.  


> 2. It's unclear to me when I should use a `pid` or `pid_v` controller.  I suspect other users would be similarly unsure. It's also not clear to me if there are a sufficient number of users that would obtain a real-world benefit to a `pid_v` controller (some notes on this at https://www.klipper3d.org/CONTRIBUTING.html ).

Generally speaking pid_v will almost always be better, because it intrinsically prevents integral wind up. The only time pid_v wouldn't be better would be when used with noisy sensor readings, because it uses a 1st and 2nd derivative. Thankfully it's easy to tell if pid_v will work or not. You do a test run, and if it doesn't settle in it won't work. The only thing you can really change would be to use different pid tuning parameters (more on that in a moment).    

Astrom & Hagglund covers velocity at a high level int their books, for example section 3.5 of this one.
https://www.ucg.ac.me/skladiste/blog_2146/objava_92847/fajlovi/Astrom.pdf


Here are some comparison graphs from my hotend, using the same pid parameters.

The plot short hand:
origonal = the current control algorithm used by klipper
updated = the updated version of the original 
velocity form = the new velocity control algorithm



Plot of the full test runs.


zoomed in showing the initial warm up with no wind-up/overshoot.



This plot shows the summation of absolute variance from the target temp, once they have all settled in. Zero variance would be a perfectly strait horizontal line.  



Here is an example of my very noisy AC powered, high thermal mass bed, with a tortures target temperature of 40C. Velocity control just did not work with Ziegler-Nichols parameters, but when i switched to Cohen-Coon as @Sineos recommended I got far better results with positional and velocity form control, but velocity form was better overall.     

The plot short hand:
positional - the current control algorithm used by klipper
velocity - the new velocity control algorithm
zn - Ziegler-Nichols
cc - Cohen-Coon
calibraded - pid parameters generated directly from a calibration run.
manual - pid parameters generated from a calibration run that have been manually tweaked to yield better performance.

 
Plot of the full test runs.


zoomed in showing the initial warm up. Because of the high thermal mass and large delay associated with my bed i always have overshoot, but it's worth noting velocity form always has less.  


absolute variance





 >So, for the pid_v commit, I suspect we would need some kind of user facing document detailing the tests a user should run to determine if a pid_v controller should be used and the benefits the user should expect from utilizing it.
 
I'm, out of town at the moment, but i'll ty and write something up over the next few days, and i'll run it by the community on discourse.

freakydude:
That explains my observation. Also, I found your [Klipper Discourse](https://klipper.discourse.group/t/experimental-pid-improvement-changes/3604) mentioned in one of your commits earlier very helpful for understanding. 

In fact, you did 3 things:
- Modified the current PID algorithm
- Additional implemented a PID Velocity variant 
- Updated the calibration algorithm for both

As I switched to your branch and switched at the same time to pid_v, my extruder and heater bed heat up time inceased. In fact, not really. In my start code, the printer starts to print if the target temp is reached the first time within a tolerance of 2.5%, With PID it overshoots the target temp after that - but starts printing. With PID_V the print starts at a later time point (after the classic PID would be settled down to the target temp).

Great!

KevinOConnor:
Thanks for the additional information.

> Generally speaking pid_v will almost always be better

What does it mean to be "better" in this context?  Is it better in the mathematical sense of less deviation from requested, or better in the sense that users will obtain real-world prints with noticeably improved quality?

I do find this interesting.  For what it is worth though, all the pid_v graphs look roughly the same to me at a high-level.  All the extruder graphs settle at roughly 2 minutes, and the bed graphs (with the exception of "positional zn calibrated") generally settle between 3 to 5 minutes.  These times aren't significant when compared to a typical print time of 2-5 hours.  So, I'm not sure I'd add a new option to Klipper for pid_v.

Again, I do find this interesting, and I'll try to review your code to better understand it.  I'll start with a focus on the pid_calibrate changes.

Cheers,
-Kevin

dalegaard:
Hi,

I'm seeing much more stable temperatures in my machine with this patch. I run an 80W heater and a 24 CFM fan, and the upstream code was never able to effectively regulate this combo for me.

I'm not currently using the velocity form.

Best regards,
Lasse

dans98:
> What does it mean to be "better" in this context? Is it better in the mathematical sense of less deviation from requested, or better in the sense that users will obtain real-world prints with noticeably improved quality?

>These times aren't significant when compared to a typical print time of 2-5 hours. So, I'm not sure I'd add a new option to Klipper for pid_v.

@KevinOConnor

The two points above are linked in my opinion, so I will see if I can answer your question and justify why in my opinion pid_v should exist as an option. I apologize upfront as I’ve had a really 48 hours at work thanks to some major release issue (not my work), so this might get a little rambly.

The most obvious way pid_v is better is that it drastically reduces or completely removes overshoot because of integral wind-up. In the grand scheme of things (as you mentioned), overshoot doesn’t affect print quality, or print time. However,It does help reduce the amount of material you waste printing skirts or using a custom priming routine.

 For me that means I'm wasting several meters less filament per spool. I do a lot of small quick prints (less than 30 mins) for prototyping stuff, and I print hot so i get more oozing than others during the warmup. 

Where pid_v is better, but in a far less obvious way, is keeping the temp on target.


bang bang, positional pid, velocity pid, etc are all doing the same thing, trying to balance a thermal equation. In the case of a hot end, you have the heater on one side putting heat in. On the other you have ambient temp, print temp, print speed, volumetric flow rate, cooling fan speed etc playing a part in taking heat out.

Each heater has what I like to call a work envelope. At the low end it’s overpowered for the given print parameters and will struggle to hold temp. At the high end, it’s under powered, and will struggle to hold temp. In the middle of the envelope is the sweat spot. Pid_v shines its brightest when you move away from the sweet spot.

Given the plethora of printers, hotends, thermistors, beds, and bed heaters on the market, it would be foolish to say every single user is going to see a marked improvement in print quality. If someone is running their printer with the hot end and bed in the sweet spot they might see a small or no improvement. Someone with their bed or hot end in the sweet spot but not the other should see some improvement. Someone with neither heater in the sweet spot should see a more significant improvment.

On my printer the bed is the most problematic heater. So I’ll use graphs for that.

First I should explain what these graphs are showing in more detail. For each pwm update interval i’m calculating variance as abs(target temp - reported temp). The graph is showing  the individual variances summed over time.  Thus the higher up the Y axis a curve goes the worse it is at keeping the heater at the target temp.|

my bed calibrated at 40C


and again at 75C


For both graphs just look at the “positional cc calibrated” &  “velocity cc calibrated” curves. At 40C (well outside my beds sweet spot) you can see that velocity pid is far better at keeping the bed on target. At 75C (close to the middle of my beds sweet spot), the two algorithms are on par with each other. 

Also take a look at the Y axis, velocity control finished both tests around 300, so it held its level of control relatively constant, while positional control didn’t.

I don’t have any graphs of it on hand, but the same thing will happen as i get close to the upper limit of my bed 110-120C. Velocity control maintains better control than positional control.

Holding tighter temperature control on the bed has very visible effects on print quality. As [jslaker](https://klipper.discourse.group/u/jslaker) mentioned in discourse when the bed temp fluctuates you can get z banding/breathing, because beds do not heat up uniformly.  And as he also mentioned cnc kitchen  did a nice video on the topic.

The video is talking about bang bang vs pid, but the concept is universal. When the temp of the bed fluctuates, the bed will flex because the bed won't heat uniformly. Depending on what your bed is made from and how it is constrained, the effects can be from mild to extreme.

https://www.youtube.com/watch?v=9JyydfcOcD0


tighter control has positive effects on hot ends as well. if your hot end temp starts fluctuating to much you can get what looks like a case of inconsistent extrusion caused by a clocked nozzle or extruder that slips slightly.

jslaker:
```What does it mean to be "better" in this context? Is it better in the mathematical sense of less deviation from requested, or better in the sense that users will obtain real-world prints with noticeably improved quality?```

@KevinOConnor 

In addition to DanS' comments re: basic print quality, one of the issues I've personally encountered with the existing PID implementation is that auto-tuned variables will essentially never correctly converge for some bed/heater combinations due to overshoot. My suspicion is that the combination of a bed with high thermal mass with a thermistor on the opposite side of the bed from the heater (and therefore considerable lag time in response to saturated heater output) and high heater power is particularly susceptible. 

As you noted in #4427, using TEMPERATURE_WAIT instead of M190 can help bandaid over this, but the downside is that bed temperature will continue to oscillate as the PID controller keeps hunting for the set point. In my case, the PID controller was never able to fully recover from overshoot with auto-calibrated terms and would run hot throughout a print.

I was able to get things behaving properly after some extensive manual tuning, but, given the complexity of PID tuning, I'd think that's a pretty big ask of most users.

In my testing, my manual terms still heat more quickly (read: are considerably more aggressive) than what I'm seeing with these changes, but the biggest difference is that it generated terms that would consistently settle at the set point in a reasonable amount of time.

I have seen improvement with my hotend as well, but in that case, it's more "holds setpoint more consistently" rather than "never reaches setpoint." :)

**Comment by pki791** [comment]
I use the bambulab hotend which heats up very quick, also the PID tune process is done within one minute. The tune returns almost similar values as the master branch. But as i changed to PID_V algorythm the heater is much more stable, overshoot less and holds the temperature during print within 0,5deg, also the range of pwm is much smaller. With PID it shoots peranently over 50%, with PID_V it is less than 20%. I vote to merge at least the PID_V algo.

**Comment by dans98** [comment]
@KevinOConnor 

I'm working on the documentation now, is there a place you want me to put the spreadsheet i created? 
https://klipper.discourse.group/uploads/short-url/oR6MJ0s0I07T2LIXvgZB6iql5DY.zip

I thought about redoing it in pure js, but it seems its not possible to include html/js in md files.

**Comment by pki791** [comment]
What about a python script? Eventually i can help if You can't do python.

**Comment by dans98** [comment]
> 

I can for sure do a python script!

**Comment by pki791** [comment]
Maybe this would be even a way too much around, the PID-TUNE could return all the values calculated in the log.

**Comment by Sineos** [comment]
In my opinion:
* Klipper main line should return only one set of values for the PID
* The PR has to choose the best possible, read most commonly applicable, model
   1. Do the tuning
   2. Return values
   3. `SAVE_CONFIG`
   4. Done
* Further material / tips / options could be posted on discourse for the interested user to play with

Why?
* Avoid confusing the users
* Avoid documentation burden --> Which model to choose when and why
* Avoid support burden because one thing is 100% certain: Folks will find a way how to mess it up the one or the other way

**Comment by KevinOConnor** [comment]
> I'm working on the documentation now, is there a place you want me to put the spreadsheet i created?

Alas, I'm not sure what I'm looking at with that spreadsheet.  I don't recall reading about it on Discourse - got a link?

-Kevin

**Comment by pki791** [comment]
> I don't recall reading about it on Discourse - got a link?

Sure https://klipper.discourse.group/t/experimental-pid-improvement-changes/3604/63?u=pki

**Comment by MGunlogson** [comment]
I tried this and had great results! 

I have a bed that can't hold under 50C with current PID implementation, it overshoots and never corrects. This is a problem when printing TPU. The new algorithm using `PID_V` holds steady down to 35C!

**Comment by dans98** [comment]
@KevinOConnor  just an fyi, its going to take me a little while on the documentation. I just got assigned a new project today that's probably going to be 2 or 3 weeks of 10 to 12 hr days, but I'll try and squeeze time in on the weekends.

**Comment by dans98** [comment]
@KevinOConnor 

Just an fyi, I'm over the hump of my work project so I can not get back onto the documentation.

**Comment by freakydude** [comment]
Hi, I can't wait 😉 Any news?

**Comment by dans98** [comment]
> Hi, I can't wait 😉 Any news?

it's coming along slowly. 

I have to admit end user documentation is not one of my strong suits. Thus i end up writing a paragraph and then reworking it several times.

Additionally work has been one fire after another so far this year.

**Comment by vkhodygo** [comment]
@dans98 Do you think you can share what you've done so that other people could potentially help you?

**Comment by dans98** [comment]
@vkhodygo
The documentation is up.
https://github.com/dans98/klipper/blob/Final-PID-Improvements/docs/PID.md

it took longer than expected, as my trip was doubled in length. When I finally got home, we had a household issue that had me running none stop for the last week.

I stripped the documentation down to the bare bones, as my initial passes were probably to in-depth for the average person.

**Comment by freakydude** [comment]
I have reviewed the document. I understood the content very well. Good work. I made a pull request to improve the grammar and spelling in your branch @dans98 

I don't know if it helps and if it is the right way to do it. Use it however you want.

**Comment by dans98** [comment]
> I have reviewed the document. I understood the content very well. Good work. I made a pull request to improve the grammar and spelling in your branch @dans98
> 
> I don't know if it helps and if it is the right way to do it. Use it however you want.

can you wrap your lines at 80 characters? if memory serves, the the build will fail if it finds lines longer than 80 characters.

**Comment by qwewer0** [comment]
@dans98

```
3d printers don’t actively cool the hot end or bed, so they can never reach the ideal.
```

If you weren't thinking about the part cooling fan, then could you elaborate a bit more?

**Comment by dans98** [comment]
@qwewer0 

If you want to drop the temp of your hot end from 250C to 200C. The pwm signal will just go to zero initially, and how long it takes to reach 200c is predominantly a function of how fast your heater block loses heat to the surrounding atmosphere. in other words our heaters are a one way system. 

Consider a servo powered robot that can move forwards & backwards. If it needs to go forward the pwm signal will be positive. To go in reverse the pwm signal will be negative. This is a two way system. 

The servo powered robot can theoretically reach the ideal, because it can go in both directions.

does that make sense?

File: docs/PID.md [hunk 1]
```diff
@@ -0,0 +1,109 @@
+# PID
+PID control is a widely used control methodology in the 3d printing world.
+It’s ubiquitous when it comes to temperature control, be it with heaters for
+generating heat, or fans for removing it. The goal of this document is to
+provide a high level overview of what PID is, and how to best leverage it in
+Klipper.
```
qwewer0, line 6:
```suggestion
It’s ubiquitous when it comes to temperature control, be it with heaters to
generate or fans to remove heat. This document aims to provide a high-level
overview of what PID is and how to use it best in Klipper.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+In 1984 Karl Astrom and Tore Hagglund published their paper “Automatic Tuning
+of Simple Regulators with Specifications on Phase and Amplitude Margins”. In
+the paper they introduced an automated tuning methodology commonly referred to
+as the “Astrom-Hagglund method” or the “Relay method”.
+
+In 2019 Brandon Taysom & Carl Sorensen published their paper "Adaptive Relay
+Autotuning under Static and Non-static Disturbances with Application to
+Friction Stir Welding", that laid out a method to generate more accurate
+results from a relay test. This is the PID calibration method currently used by
+Klipper.
```
qwewer0, line 27:
```suggestion
In 2019 Brandon Taysom & Carl Sorensen published their paper "Adaptive Relay
Autotuning under Static and Non-static Disturbances with Application to
Friction Stir Welding", which laid out a method to generate more accurate
results from a relay test. This is the PID calibration method currently used by
Klipper.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+### PID Calibration
+
+As previously mentioned Klipper uses a relay test for calibration purposes. A
+standard relay test is conceptually simple. You turn the heater's power on and
+off to get it to oscillate about the target temperature, as seen in the
+following plot.
```
qwewer0, line 34:
```suggestion
As previously mentioned, Klipper uses a relay test for calibration purposes. A
standard relay test is conceptually simple. You turn the heater’s power on and
off to get it to oscillate about the target temperature, as seen in the
following graph.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+![simple relay test](img/pid_01.png)
+
+The above plot shows a common issue with a standard relay test. If the system
+being calibrated has too much or too little power for the chosen target
+temperature, it will produce biased and asymmetric results. As the plot above
+shows, the system spends more time with the power off, and has a greater
+amplitude above the target temperature than it does below.
```
qwewer0, line 42:
```suggestion
The above graph shows a common issue with a standard relay test. If the system
being calibrated has too much or too little power for the chosen target
temperature, it will produce biased and asymmetric results. As can be seen
above, the system spends more time in the off state than on and has a larger
amplitude above the target temperature than below.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+In an ideal system, both power on and off time, as well as the amplitude above
+and below the target temperature would be equal. 3d printers don’t actively
+cool the hot end or bed, so they can never reach the ideal.
```
qwewer0, line 46:
```suggestion
In an ideal system, both the on and off times and the amplitude above and below
the target temperature would be the same. 3D printers don’t actively cool the
hot end or bed, so they can never reach the ideal state.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+The following plot is a relay test that uses the methodology laid out by
+Taysom & Sorensen. After each iteration, the data is analyzed, and a new
+maximum power setting is calculated. As can bee seen from the plot, the system
+starts the test asymmetric, but ends it very symmetric.
```
qwewer0, line 51:
```suggestion
The following graph is a relay test based on the methodology laid out by
Taysom & Sorensen. After each iteration, the data is analyzed and a new maximum
power setting is calculated. As can be seen, the system starts the test
asymmetric but ends very symmetric.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+![advanced relay test](img/pid_02.png)
+
+ A topic that’s not often discussed in the 3dp community is the conditions in
+ which calibration should be performed. When a calibration test is performed
+ external variables should be minimized as much as possible, as the goal of the
+test is to model the system in steady state conditions free of external
+disturbances. For example, if you are calibrating a hot end, you do not want a
+bed or chamber heater actively heating up or cooling down. You want them off,
+or holding at their target temperature. Part cooling and chamber fans can also
+be problematic, as they can cause temperature fluctuations in the hot end.
```
qwewer0, line 62:
```suggestion
A topic that’s not often discussed in the 3D printing community is the
conditions in which calibration should be performed. When a calibration test is
performed external variables should be minimized as much as possible, as the
goal of the test is to model the system in a steady-state condition and free of
external disturbances. For example, if you are calibrating a hot end, you do
not want a bed or chamber heater actively heating up or cooling down. You want
them off, or holding at their target temperature. Part cooling and chamber fans
can also be problematic, as they can cause temperature fluctuations in the hot
end.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Pid Control Parameters
+
+Many methods exist for calculating control parameters, such as Ziegler-Nichols,
+Cohen-Coon, Kappa-Tau, Lambda, and many more. By default, classical
+Ziegler-Nichols parameters are generated. If A user wants to experiment with
+other flavors of Ziegler-Nichols, or Cohen-Coon parameters, they can extract
+the constants from the log as seen below and feed them into this
```
qwewer0, line 70:
```suggestion
Many methods exist for calculating control parameters, such as Ziegler-Nichols,
Cohen-Coon, Kappa-Tau, Lambda, and many more. By default, classical
Ziegler-Nichols parameters are generated. If a user wants to experiment with
other flavors of Ziegler-Nichols, or Cohen-Coon parameters, they can extract
the constants from the generated log as seen below, and enter them into the
following table.
```

After this, the order of the log and the table could be switched.

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+```
+Ziegler-Nichols constants: Ku=0.103092 Tu=41.800000
+Cohen-Coon constants: Km=-17.734845 Theta=6.600000 Tau=-10.182680
+```
+
+Classical Ziegler-Nichols parameters work in all scenarios. Cohen-Coon
+parameters work better with systems with a large amount of dead time/delay.
+For example, if a printer has a large thermal mass bed that’s slow to heat up
+and stabilize, Cohen-Coon parameters will generally do a better job at
+controlling it.
```
qwewer0, line 82:
```suggestion
For example, if a printer has a bed with a large thermal mass that’s slow to
heat up and stabilize, the Cohen-Coon parameters will generally do a better job
at controlling it.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Pid Control Algorithms
+
+Klipper currently supports two control algorithems positional, and velocity.
+The fundamental difference between the two algorithms is as follows. The
+positional algorithm calculates what the PWM value should be for the current
+time interval. The velocity algorithm calculates how much the previous PWM
+setting should be changed by, to get the PWM value for the current time
+interval.
```
qwewer0, line 91:
```suggestion
Klipper currently supports two control algorithms: Positional and Velocity.
The fundamental difference between the two algorithms is that the Positional
algorithm calculates what the PWM value should be for the current time
interval, and the Velocity algorithm calculates how much the previous PWM
setting should be changed to get the PWM value for the current time interval.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+Positional is the default algorithm, as it will work in every scenario. The
+Velocity algorithm can provide superior results to the Positional algorithm,
+but requires lower noise sensor readings, or a larger smoothing time setting.
```
qwewer0, line 95:
```suggestion
Positional is the default algorithm, as it will work in every scenario. The
Velocity algorithm can provide superior results to the Positional algorithm but
requires lower noise sensor readings, or a larger smoothing time setting.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+The most noticeable difference between the 2 algorithms is that given the same
+configuration parameters velocity control will eliminate or drastically reduce
+overshoot as seen in the plots below, because it is not susceptible to integral
+wind-up.
```
qwewer0, line 100:
```suggestion
The most noticeable difference between the two algorithms is that for the same
configuration parameters, velocity control will eliminate or drastically reduce
overshoot, as seen in the graphs below, as it isn’t susceptible to integral
wind-up.
```

docs/PID.md
```diff
@@ -0,0 +1,109 @@
...
+
+![algorithm comparison](img/pid_03.png)
+
+![zoomed algorithm comparison](img/pid_04.png)
+
+In some scenarios Velicty control will also be better at holding the heater at
+its target temperature, and rejecting disturbances. The primary reason for this
+is that velocity control is more like a standard second order differential
+equation. It takes into account position, velocity, and acceleration.
```
qwewer0, line 109:
```suggestion
In some cases, Velocity control will also be better at holding the heater at
its target temperature and suppressing disturbances. The primary reason for
this is that Velocity control is more like a standard second-order differential
equation. It takes into account position, velocity, and acceleration.
```

qwewer0 (changes requested):
Hope it fits the style of the whole document.

qwewer0:
> @qwewer0 I've incorporated a good deal of your changes into my local copy. What name and email address do you want me to use in the sign off?

Feel free to use:
```
Signed-off-by: Csongor Boros b.csongor93@gmail.com
```
And thanks for the question, I'm honored.

ReXT3D:
Really glad to see this work being lined up for a merge!

I have one minor comment that I mentioned before via Discord.  The current implementation and the spreadsheet uses Cohen-Coon constants that use unnatural sign convention.  Specifically take this Example from `PID.md`:

`Cohen-Coon constants: Km=-17.734845 Theta=6.600000 Tau=-10.182680`

In the above example, the negative sign of Km (process gain) would normally imply that the temperature is inversely proportional to the commanded PWM value.  This is obviously not the case.  Similarly, Tau represents the process time constant which in natural units is always positive.  I think it would be helpful to correct these sign anomalies in the code and in the spreadsheet so that the reported process parameters follow normal conventions.

Thanks @dans98 for the awesome effort!

dans98:
> 
>  The current implementation and the spreadsheet uses Cohen-Coon constants that use unnatural sign convention. Specifically take this Example from `PID.md`:
> 
> `Cohen-Coon constants: Km=-17.734845 Theta=6.600000 Tau=-10.182680`
> 


>  I think it would be helpful to correct these sign anomalies in the code and in the spreadsheet so that the reported process parameters follow normal conventions.
> 

@ReXT3D  As i mentioned when you messaged me back in October about this, that just how the math works out. If you just change the signs, you will get dramatically different results.

see equations 5 and 6 here for calculating τ and Km.
https://www.osti.gov/servlets/purl/1646635

You then use Θ, τ, & Km to calculate Ti & Td  and ultimately Kp, Ki, & kd.

ReXT3D:
> @ReXT3D As i mentioned when you messaged me back in October about this, that just how the math works out. If you just change the signs, you will get dramatically different results.

Roger that.  I did not dive into the math, but obviously the relay method described therein uses a different definition of process parameters.  I am obviously too influenced by characterizing the process by means of open loop manual step input method.

All good!

dans98:
> First of all I want to say, your PR is amazing, with the default Pid algorithm, my phaetus rapido was overshooting by up to 15° when heating up, with pid_v there is no overshoot at all and the temps are really stable.
> 
> But I have found a little bug: When you do a PID Autotune, the control value always reverts to "Pid" after saving the config. I have already written a pretty simple fix for it and will open a PR for it soon.

I'm glad to hear it's working out for you, I'll be on the lookout for the PR.

Zeanon:
> > First of all I want to say, your PR is amazing, with the default Pid algorithm, my phaetus rapido was overshooting by up to 15° when heating up, with pid_v there is no overshoot at all and the temps are really stable.
> > 
> > But I have found a little bug: When you do a PID Autotune, the control value always reverts to "Pid" after saving the config. I have already written a pretty simple fix for it and will open a PR for it soon.
> 
> I'm glad to hear it's working out for you, I'll be on the lookout for the PR. 

PR is out

dans98:
@LastZeanon Your changes have been merged.

dans98:
> One little thing I noticed: The symmetry reading IMO is worth mentioning in the docs. I think this is a highly useful information

Have you seen this?
https://github.com/dans98/klipper/blob/Final-PID-Improvements/docs/PID.md

Or are you talking about what shows up in the console while a calibration is running?

Sineos:
> Or are you talking about what shows up in the console while a calibration is running?

@dans98 exactly. The (a)symmetry value which can be interpreted as an heat gain / heat loss ratio.

Zeanon:
@dans98 I read that you wanted to implement PID Profiles when this PR is merged(at least I think I read that).
I have already done that(based on your Pid improvements branch, which is why I can't do a PR till this one is merged XD).
If you are interested, just ping me

dans98:
> @dans98 I read that you wanted to implement PID Profiles when this PR is merged(at least I think I read that). I have already done that(based on your Pid improvements branch, which is why I can't do a PR till this one is merged XD). If you are interested, just ping me

Glad to hear you have done it already, it's something I personally think is needed.

dans98:
@Sineos  I added some info to the documentation!

docs/PID.md
```diff
@@ -0,0 +1,144 @@
...
+3:17 PM   sample:4 pwm:0.5866 asymmetry:0.0169 tolerance:0.4134
+3:18 PM   sample:5 pwm:0.5852 asymmetry:0.0668 tolerance:0.0377
+3:18 PM   sample:6 pwm:0.5794 asymmetry:0.0168 tolerance:0.0142
+3:19 PM   sample:7 pwm:0.5780 asymmetry:-0.1169 tolerance:0.0086
+3:19 PM   PID parameters: pid_Kp=16.538 pid_Ki=0.801 pid_Kd=85.375
+               The SAVE_CONFIG command will update the printer config file
+               with these parameters and restart the printer.
+```
+
+When asymmetry starts off negative, It will not converge to zero. If Klipper
```
Sineos, line 74: When asymmetry starts off negative, **it** will

docs/PID.md
```diff
@@ -0,0 +1,144 @@
...
+does not error out, the calibration run will complete and provide good PID
+parameters, However the heater is less likely to handle disturbances as well
```
Sineos, line 76: parameters, **however** the heater

Sineos (changes requested):
Excellent, thanks

Sineos:
@dans98 and @KevinOConnor What is missing to get this landed?

Sineos:
@dans98 maybe a last comment regarding the documentation:
* I really appreciate the details about PID and its history
* From my support experience this level of detail will cause a TLDR reflex
* The "important", i.e. Klipper / tuning related information gets easily lost

Maybe a structure like:
* Basic tuning (very concise)
   * When to choose Positional vs Velocity
   * Prepare the test: Minimize external influences
   * How to interpret the tuning results, e.g. the asymmetry value
* Advanced tuning
   * Write file
   * Excel  
* Further reading
   * All the rest about different algos, history etc

KevinOConnor:
> What is missing to get this landed?

It's mainly lack of time on my side.  This PR is effectively a rewrite and unfortunately that makes for a large review burden (both because it's difficult to identify what has changed and because it increases the risk of a regression).  I think a `git rebase` of this PR into small "atomic" commits, and avoiding unneeded whitespace or variable renames would really help.

Cheers,
-Kevin

dans98:
@KevinOConnor  I do not believe a rebase is warranted.
https://github.com/Klipper3d/klipper/pull/5955/files

2 existing md files have had miniscule changes.
1 new md file has been added.
5 new files have been added in resources, to support the new md file.

The only 2 python files thats have been changed are heaters.py and pid_calibrate.py

in heaters.py the only change of significance is the temperature_update method on the pid control class and that's only a 33 line method, 13 of witch are comments/spacing. The other significant change is the new velocity pid algorithm.

in pid_calibrate.py the only change of significance is the rewrite of the ControlAutoTune class. In my opinion that cannot be broken down into atomic commits, as the algorithm/concept is far to different to the previous one.

Sineos:
@KevinOConnor 
Unfortunately I cannot support in the code review part. FWIW:
* I'm using it since months with very good success and no apparent issues
* There is quite some positive feedback in some discourse discussions 
* Various people here checked the documentation aspect and provided input. 
* Finally leaving two files with code changes for your review

dans98:
>  As it is, I'm not comfortable merging without also following the software engineering practices that I feel ensure good long term code stability.

It might be to 2 or 3 weeks till i can get to it, because I'm dealing with some family health issues, but I will vacation/free time for sure at the end of the month.

However I want to ask how are you determining what is good or bad code? do you have a set of volunteer beta testers, are you doing line by line code review, what's the setup?

I'm asking because when i first started looking at the pid controller over a year ago, I found a basic logic flaw that's been in place since version 0.2. 

https://github.com/Klipper3d/klipper/blob/v0.2.0/klippy/heater.py#L166

consider a target temp of 220 degrees, and an actual temp of 221 degrees. 

```
temp_err = self.heater.target_temp - temp
```
220 - 221 = -1

```
temp_integ = self.prev_temp_integ + temp_err * time_diff
```
time_diff is always 0.3

self.prev_temp_integ is the summation of previous errors. To make the remaining math simpler, lets just assume its zero.

thus temp_integ = 0 + -1 * 0.3 = -0.3

```
temp_integ = max(0., min(self.temp_integ_max, temp_integ))
```

self.temp_integ_max is set in init. in current master that's instantiated as follows
```
        self.temp_integ_max = 0.
        if self.Ki:
            self.temp_integ_max = self.heater_max_power / self.Ki
```
self.heater_max_power / self.ki should always be a positive value. thus to make the remaining math simpler lets assume self.temp_integ_max is 1.0

thus we have

min(1.0, -0.3) = -0.3

max(0., -0.3) = 0.

Now let's look at why using the max function like this is a fundamental flaw.
https://github.com/Klipper3d/klipper/blob/v0.2.0/klippy/heater.py#LL170C9-L170C77
```
co = int(self.Kp*temp_err + self.Ki*temp_integ - self.Kd*temp_deriv)
```
self.Ki * 0. = 0

Never letting temp_integ go negative, means the integral competent can also never go negative. Thus once the actual temp  is above the target temp, the Integral component is essentially useless, as it can do nothing to improve the situation.

Zeanon:
@dans98 I can try to split the reworked pid tune code into the smaller commits like Kevin asked if that's fine with you or try to help you out with other stuff on the pr as I think those changes are awesome and I really want them merged XD

dans98:
> @dans98 I can try to split the reworked pid tune code into the smaller commits like Kevin asked if that's fine with you or try to help you out with other stuff on the pr as I think those changes are awesome and I really want them merged XD

let me see what I can do with what remains of the weekend.

KevinOConnor:
> However I want to ask how are you determining what is good or bad code? do you have a set of volunteer beta testers, are you doing line by line code review, what's the setup?

Thank you for the feedback.

During a review, I try not to judge "good or bad" and instead I try to look for "defects" and to check that the commit follows the documented change processes.  I've written about this at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review .  To be sure, it's hard to separate out opinions from a review, and I'm sure I could do better at it.  I do try to focus on "defects" though.

The main concern I have with this PR is item 6 at the link above - specifically "are commits well formed, address a single topic per commit, and independent".  I did not originate this practice of isolating commits to a single topic - if you do a web search for "atomic commit" you'll find lots of references to it - for example: https://github.blog/2022-06-30-write-better-commits-build-better-projects/

> I'm asking because when i first started looking at the pid controller over a year ago, I found a basic logic flaw that's been in place since version 0.2.

I agree that looks like a defect in the current code.  If you submit a PR with just that fix then it will get fixed rapidly.  It was not clear to me that your new code addressed this defect.  (It looks to me like a simple fix is to change the min/max to: `temp_integ = max(-self.temp_integ_max, min(self.temp_integ_max, temp_integ))`)

From your description above it is now also clear to me that your new code introduces a regression.  The `time_diff` is not always 0.300.  The low-level mcu protocol does not ensure delivery of messages from mcu to host ( https://www.klipper3d.org/Protocol.html#message-flow ) so `time_diff` will normally be 0.300, but it may be 0.600 (or 0.900 or more) in the event of a lost message.  I attempted to write the pid code to gracefully handle these types of events.

FWIW, the above is a good example of why we really want to separate large changes into a series of small atomic changes that can be individually reviewed.

Cheers,
-Kevin

dans98:
@KevinOConnor 

> It was not clear to me that your new code addressed this defect. (It looks to me like a simple fix is to change the min/max to: `temp_integ = max(-self.temp_integ_max, min(self.temp_integ_max, temp_integ))`)

It does, by removing it, as well as the logic to set the temp_integ_max.  The concepts are not needed when you use a slightly more advanced version of conditional integration.  
 
 This if statement
 https://github.com/Klipper3d/klipper/blob/master/klippy/extras/heaters.py#L213
 
 Becomes this one
 https://github.com/dans98/klipper/blob/Final-PID-Improvements/klippy/extras/heaters.py#L218
 with the more advanced logic described by Astrom & Hagglund 
 
  https://drive.google.com/file/d/1m_uTwvOMMpjDP4tWROZ4jdOHU4BSwuSk/view

> A few different switching conditions are now considered. Orre simple ap-
proach is to switch off integration when the control error is large. Another
approach is to switch off integration during saturation. Both these methods
have the disadvantage that the controller may get stuck at a non-zero control
error if the integral term has a large value at the time of switch-off.
A method without this disadvantage is the following. Integration is switched
off when the controller is saturated and the integrator update is such that it
causes the control signal to become more saturated. Suppose, for example, that
the controller becomes saturated at the upper saturation. Integration is then switched off if the control error is positive, but not if it is negative.

Most of what i did to "improve" the current positional pid controller, was replace what looked to be arbitrary limitations, and replace them with more industry standard/recommended ones like the above. 

I also did stuff like switching the integral from the left hand integration rule, to the trapezoidal one as it is has less error.



> From your description above it is now also clear to me that your new code introduces a regression. The `time_diff` is not always 0.300. The low-level mcu protocol does not ensure delivery of messages from mcu to host ( https://www.klipper3d.org/Protocol.html#message-flow ) so `time_diff` will normally be 0.300, but it may be 0.600 (or 0.900 or more) in the event of a lost message. I attempted to write the pid code to gracefully handle these types of events

It would have been nice to have this as an inline comment. Honestly if the step size, can very by several multiples of the set constant, the system should error out in some way. especially when it comes to potential fire risks etc.   In all my testing, I never saw anything that looked like more than machine precision/rounding issues.


Yo can go ahead and close this pid request if you want, and i will submit some smaller ones with the fixes and improvements. The pid calibration one will most likely not be tinny though, as the algorithms are exceptionally different on multiple levels.

KevinOConnor:
> trapezoidal

I'm not sure if you are referring to this new code `(self.prev_err + err) / 2.`.  Averaging the latest temperature with the previous temperature would add an additional 150ms update delay to the PID.  I'm not sure that's a good idea.  I've seen reports that the current default of 1 second `smooth_time` may not be optimal, and many printers would benefit from a 0 second `smooth_time`.  Adding an additional systemic delay could be problematic on some heaters.

> Honestly if the step size, can very by several multiples of the set constant, the system should error out in some way. especially when it comes to potential fire risks etc.

FYI, the mcu itself has several checks on the heaters independent of the host communication - https://www.klipper3d.org/FAQ.html#will-the-heaters-be-left-on-if-the-raspberry-pi-crashes .  I don't think missing an occasional pid update is a safety hazard.

Cheers,
-Kevin

dans98:
@KevinOConnor where are you getting 150ms (0.15 seconds) from?

when the pid controller is initialized self.prev_err is set to ambient temp here.   
https://github.com/dans98/klipper/blob/Final-PID-Improvements/klippy/extras/heaters.py#L190

at the end of each temperature_update request the class attribute is updated here. 
https://github.com/dans98/klipper/blob/Final-PID-Improvements/klippy/extras/heaters.py#L214

saving a value to a class variable should be be many orders of magnitude less than 150ms.

Alos switching to the trapezoidal rule, isn't about machine precision level accuracy improvements, we are talking much much larger improvements.

The left hand and right hand numerical integrations rules, use rectangles to approximate the integral. Klipper uses the right hand rule (I always get them reversed.)
  https://github.com/Klipper3d/klipper/blob/master/klippy/extras/heaters.py#L201

`temp_err * time_diff 
`

here is an example image showing the differences between the 2.
compare



The trapezoidal rule uses as you might have guessed, a trapezoid to approximate the integral, hence why it does a much better job at approximating the integral.

here is an image comparing the 2 methods using the same step size.
trap
   

  Even better rules exist than the trapezoidal rule, but they come with additional constraints (step size needing to be constant), and pitfalls.

KevinOConnor:
> where are you getting 150ms (0.15 seconds) from?

The temperature update rate is every 300ms, and (in the typical case) averaging of two samples will provide an estimate of the value half way in time between the two discrete samples.  Thus, typically, resulting in a temperature that is 150ms in the past from the last temperature sample.

Since `err = target_temp - temp` and `(self.prev_err + err) / 2.`, in the common case that will be `target_temp - (prev_temp + temp)/2.` - which is an averaging of the temperature (aka, a low pass filter).

> The trapezoidal rule uses as you might have guessed, a trapezoid to approximate the integral, hence why it does a much better job at approximating the integral.

Okay, I understand now.  I missed this when I first saw the code.  Although the code is in effect smoothing the temperature, it isn't doing this to reduce measurement noise, it is doing it to obtain a more accurate integral.  That seems fine to me.

Thanks,
-Kevin

KevinOConnor:
Well, thinking about it further, over the long term the different integration methods are going to calculate the same results, as we are summing them anyway.  So, in effect, the only thing the new code does is smooth out some noise and delay the integral action by ~150ms.  I guess that's fine as the feedback from temperature integration isn't intended to be fast.

-Kevin

dans98:
@KevinOConnor 

The results will not be the same long term.

see the error estimations here, and note we can't use the midpoint rule as we can't generate data mid time step.
https://en.wikipedia.org/wiki/Riemann_sum#Riemann_summation_methods  

Simpsons rule  is even more accurate, but it can't be used in a live time dependent situation as it would actually cause delay/latency.
https://en.wikipedia.org/wiki/Simpson%27s_rule

KevinOConnor:
> The results will not be the same long term.

It seems to me it will be.  Putting aside rare boundary cases, your code basically does `self.int_sum += ((self.prev_err + err) / 2.) * self.dt` on each iteration.  That is equivalent (putting aside rare boundary cases) to `self.int_sum += ((target_temp - self.prev_temp)/2 + (target_temp - temp)/2) * self.dt`.  Over long time frames that is the same as summing the series `(target_temp - temp) * self.dt`, which is what the current code does.  That is, the new code is just delaying the adding of `(target_temp - temp) / 2 * self.dt` until the next iteration (putting rare boundary cases aside).

What am I missing?

-Kevin

dans98:
> What am I missing?

I probably have an unfair advantage as this is what my background is in, so i just intrinsically know a lot of this.

I'm done with work now, so i can provide some  examples with visuals.

 take a look at the plot of the equation y=-.2x^2+5x, 
 https://www.wolframalpha.com/input?i=plot+y%3D-.2x%5E2%2B5x+from+0+to+10 
 
 because its a simple continues equation in can be integrated exactly.
 If we integrate between x=0 and x=10 we get an answer of 183.333
 https://www.wolframalpha.com/input?i=integrate+y%3D-.2x%5E2%2B5x+between+0+and+10
 
 if we use the right hand method (what klipper is using) with a step size of 1 we get an answer of 198. That's an overestimation of the integral by ~8%.
 https://www.wolframalpha.com/input?i=integrate+-.2x%5E2%2B5x+from+0+to+10+with+right+endpoint+method+with+interval+size+of+1   
 
 If we switch to the trapezoidal rule, we get an answer of 183 That's an under estimation by  ~0.2%.
 https://www.wolframalpha.com/input?i=integrate+-.2x%5E2%2B5x+from+0+to+10+with+trapezoidal+method+with+interval+size+of+1  
 
 
 
 here is a sin wave example, and again the trapezoidal method is more accurate. It underestimated by ~1.3%, while the right hand rule underestimated by ~3.6% 

 https://www.wolframalpha.com/input?i=plot+y%3Dsin%28x%29%2B1+from+0+to+10
 https://www.wolframalpha.com/input?i=integrate+y%3Dsin%28x%29%2B1+between+0+and+10
 https://www.wolframalpha.com/input?i=integrate+y%3Dsin%28x%29%2B1+from+0+to+10+with+right+endpoint+method+with+interval+size+of+1
 https://www.wolframalpha.com/input?i=integrate+sin%28x%29%2B1+from+0+to+10+with+trapezoidal+method+with+interval+size+of+1
 
 In pretty much every senario the trapezoidal method is more accurate,
 
 
 If you were to take a numerical methods class, they likely won't even cover the right and left hand methods. They'll jump straight to the trapezoidal and Simpsons methods, and then progress to even more complex and accurate methods.

KevinOConnor:
> take a look at the plot of the equation y=-.2x^2+5x,

I understand that integration of a real continuous math function would be different depending on the method used.  However, we don't have a real continuous math function - we have a discrete set of temperatures.

Something like:
`temperature_series = [1., 2., 3., 4., 5., 6., 7., 8., 9.]`

You seem to be telling me that there is a fundamental difference between:
`s1 = sum([(temperature_series[i] + temperature_series[i+1])/2. for i in range(len(temperature_series)-1)])`
and:
`s2 = sum([temperature_series[i] for i in range(len(temperature_series))])`

But, there definitely is not a fundamental difference between these two summations.  The difference is between `s1` and `s2` is always `(temperature_series[0]+temperature_series[-1])/2`.

Try it - use any set of temperatures you want.

-Kevin

dans98:
@KevinOConnor 

Ok, just close the pull request then, i'm done trying to convince you of what anyone who has any knowledge of this topic knows to be fact. 

You are just jerking me around, and wasting my time now! I should have just listened to what so many others told me the very first time i started looking into contributing to klipper.

InsanityAutomation:
Dan's method matches what's in use internally by some series of Rockwell Automation PLC's standard PID functions on formulas followed, at least for a single acting function. Lacking a flaw in implementation, standard methods always trump arbitrary as it's predictable and expected by those familear with the field.

The autotune code block for the micro800 series PLC's is actually published in a readable form I believe if you wanted to look at something used on industrial systems as a reference. I'm not sure you can access any of the actual PID code without a TechConnect contract though.

CroXY3D:
Kevin,

your comments show a profound misunderstanding of how pid loops work.  Please step aside and let someone who know what he’s doing fix this

pid is clearly broken in klipper.  It is not stable and oscillates wildly especially with beds.

MGunlogson:
@KevinOConnor 

My memory of calc is hazy, but I believe the trapezoid rule is indeed more accurate when the underlying function is continuous, even if your samples are discrete. And temperature change definitely is continuous. 

Trapezoid rule is more accurate unless your underlying data is discrete. For example, a sensor that counts jellybeans passing by. You can't have half a jellybean so trapezoid rule won't increase accuracy in that case. 

Anyways, temp change isn't discrete values so AFAIK trapezoid rule will help quite a bit in our case.

freakydude:
I've been thinking about whether or not to say something about this. I've been following the Pull Request from the beginning and I don't like to see it just die now.

I suggest that we all calm down a bit here first and look at the situation from the balcony. 

Let me summarize my view:
This improvement has a good half year of discussion, implementation and various tests behind it. And a lot of heart and soul. That a bad regression is to be feared, I hold therefore leader rather improbably. The improvement is based on the initial paper, a theoretical improvement is not to be dismissed and at least here some people have already confirmed it in their scenario. None has reported a deterioration so far. 

On the other hand, the current PID implementation works well, so the risk of a new bug creeping in cannot be completely dismissed. 

To reduce this risk, there are usually a few remedies in software development. 
- As Kevin suggests, and I don't think anyone here has seriously objected, small atomic commits, so that changes are easily understandable, ideally sorted logically. How small is small enough, of course, is open to debate. If one algorithm is replaced by another, only a complete commit makes sense. Even the complete difference of the pull request is far from atomic, but not really much code either. For me it would be reviewable - maybe you both just do it live in Discord and have a beer ... 
- Tests. Automated tests are the very best way to detect bugs and thus minimize risk and fear. Of the existing tests, none has ever broken. I think we could talk about using this pull request as a reason to add more automated tests.
- Feature Toggles. If variant 1 and 2 are not good enough for other reasons, I would suggest to hide the customized PID algorithm (completely) behind a configuration entry "pid_new" and the velocity based one under "pid_v" and let all users out there run the code under "pid" as it has existed for a few years now. The solution allows that the code is mainline, but the users can switch to "pid_new" or "pid_v" experimentally for a longer while - until one dares to replace "pid" by "pid_new". 

I think the aspects are valid, from my point of view it needs an announcement (or assistance), then it will be done for sure.


The recent discussion around single lines of code and theoretical optimums of these I consider quite petty. Here I would recommend to the involved ones to swallow that. Optimizing can always be done in the end before everyone loses interest.

I would be happy if this ends well and maybe is an occasion for upcoming PRs. 
Many greetings and thanks for the great project

Zeanon:
> On the other hand, the current PID implementation works well, so the risk of a new bug creeping in cannot be completely dismissed. 

To be honest, the current implementation does work, but not that well for certain types of heaters, mainly high power PTC heaters like in the phaetus rapido(the reason why I switched to this implementation is that I had a temperature overshoot of 10°-15° when heating up which was far from ideal).
I get Kevin's point, which is why I already offered to help break down this PR and I also agree with you that one person is not enough to maintain a project of this size which is why I would suggest to get more people on board(I know there are already four).
I know this would need some time in the first place to get everyone up to date/discuss what is needed/expected from everyone but I think it would be worthwhile in the long run.
I can also again offer my help with the project, however that would be.

jslaker:
>To be honest, the current implementation does work, but not that well for certain types of heaters, mainly high power PTC heaters like in the phaetus rapido(the reason why I switched to this implementation is that I had a temperature overshoot of 10°-15° when heating up which was far from ideal).

To echo this: the impetus for @dans98 working on this code is that we observed certain printer setups where temperature would *never* correctly converge with the setpoint using the extent control loop and calibration routine. As I noted in a previous comment, there are years-old bug reports describing this behavior which makes sense given this code has been in place even longer.

The code in place now works reasonably well for *most* setups, but there are printers in the real world where it's not a matter of some marginal level of inaccuracy but being unable to start prints entirely without resorting to workarounds.

MGunlogson:
I concur with recent comments. I really want to see this merged. I've been looking forward to it more than any other Klipper feature. 

 I have both a weak bed heater and Rapido hotend. The current algo overshoots hotend 15C and can't hold "extreme" bed temps. The new algorithm (PID_V) doesn't overshoot and holds bed temps from 40C-120C. That's 10C lower and 15C hotter than current algorithm can manage. 

Many of us Klippers are using modified printers operating on the edges of heater capability. Improvement in PID control will allow me to print TPU and Polycarbonate without replacing my heaters or messing with PID parameters when I switch materials. 

Plz try to work this out y'all. @dans98 put a ton of work into it and it has real benefits for some users.

celtare21:
+1 on adding new features and improvements. To me, being reluctant to new stuff is only a step backwards. I understand not wanting to introduce regression or bugs, but staying in place is also a regression. Code is never perfect and we should always be looking to improve it. Just my 2 cents

dans98:
@vkhodygo 

> The most basic relevant example should do the trick.

The simplest way to see this is to take a a piece of paper and set up an X/Y axis. Then from left to right, place some random dots. Then again from left to right connect the random dots with strait lines, just like if you tried to plot the random sensor data like I did in this response.
https://github.com/Klipper3d/klipper/pull/5955#issuecomment-1368374527

They try and approximate the area under the interpolated dots with rectangles and trapezoids.  The rectangles will have errors, the trapezoids will not. If I have time over lunch I will do a white board example.

MGunlogson:
@KevinOConnor understandable. Perhaps we can follow a previous suggestion to have the new code split completely from the old implementation. "PID_NEW" for the modified version, and "PID_V" for new algo or something like that. That way if there's any bugs it won't affect existing installations. Tuning could be handled a similar way, with the new tuning algorithm used for the new PID algorithm and existing one remaining as is? 

That way we can get this merged so Klippers can test it and get some real world experience on its performance. Could mark it experimental in the docs too, to discourage newbies from using it until it's cooked for a while.

dans98:
in a week or to I should be able to work on updating the velocity to address the  the time intervals being able to very.

dans98:
> I have just encountered an interesting phenomenon: When setting my bed to 45° while printing, I got 50°(with pid profile tuned for 45°) When setting it to 45 manually it still kept creeping up When setting it to 50 with a profile tuned for 50, I got like 52 at 55 and above, I get what I request

@Zeanon 

where you running mainline, or my branch? 
If mine where you were running mine where you using pid or pid_v? 
also what do you mean by setting it to 45 manually?

@jslaker had this issue with the old controler and his bed. The temp would overshoot, and just never come back down to the tarket. He got around it with a lot of manual tuning if memory serves.

Technically speaking any pid controller can have this issue.
1) to much or to little smoothing.
2) sensor going bad and generating more noise. (i had this happen when a pt100 started going bad)
3) connection not fully seated and thus more noise (i had this after to many rounds of resonance testing)
4) pid parameters to different for the current environment. (i had this as well when enclosed my printer, and insulated it)

Zeanon:
Sorry, I should have clarified
I am running your PID code(but with my implementation of PID Profiles added, that's why I had tuned variables for 45 and 50 ready to go)
The initial problem(printing something and getting 50° even though 45 was requested) happened with pid_v.
After that print finished I waited a bit, restarted and then set the temp to 45 in the Webinterface to make sure it was not just a one off.
It also crept higher(up to 47/48° until the idle timeout kicked in).
I also did the same thing but changed the control to PID and the result was basically the same.
As for changes in environment or stuff: it could ofc be that my thermistor is dying or the wire is bad somewhere but I doubt that since every temp above 50 is rock solid.

dans98:
@Zeanon 

I'm ready to summit a PR just to add velocity pid control, and it incorporates your changes . What name and email address do you want me to use in the sign off?

Signed-off-by: first_name last_name username@domain

Zeanon:
> @Zeanon 
> 
> I'm ready to summit a PR just to add velocity pid control, and it incorporates your changes . What name and email address do you want me to use in the sign off?
> 
> Signed-off-by: first_name last_name username@domain

Thx for asking :)

Signed-off-by Vinzenz Hassert thezeanon@gmail.com

Would there be any changes to the velocity algorithm compared to the one in this PR? or would it be the same algorithm?(cause I want to keep using your improved auto tune since it provides way better results for me)

dans98:
@Zeanon 

> Would there be any changes to the velocity algorithm compared to the one in this PR? or would it be the same algorithm?(cause I want to keep using your improved auto tune since it provides way better results for me)

It's functionally the same, it's just updated to deal with the fact that we can't rely on the interval between pwm being constant.

to continue using my calibration revision, you should just be able to drop the calibration file into the appropriate location and restart. assuming the velocity pid pr request gets merged into master, I will then make another one to incorporate my calibration revision.

dans98:
for those that are interested here is the new minimal PR.
https://github.com/Klipper3d/klipper/pull/6272/

Zeanon:
Ah ok, then I'll have to grab the changes for the pwm intervals and drop them in my fork
Thx for the quick response

KevinOConnor:

-Kevin

pki791:
Is this new PID functionallity now included in the main release or still not?

PR #5956, opened 2022-12-27
Add SKR2 F429 chip variant to spi_flash

JamesH1978:
Later addition of a F429 variant SKR2 was released. Changed btt-skr2 to btt-skr-2-f407 and added a new alias for btt-skr-2-f429 linked to the octopus with same variant and pins

Thanks
James

Signed-off-by: James Hartley james@hartleyns.com


JamesH1978:
Thanks Kevin,

That is a very good point, i did not think to take into account the alias change, this would indeed be a breaking change if people are using this command and alias in scripts or standalone.

I will add an addition to Config_Changes.md to document this change, if @Arksine  is in agreement.

James

Arksine:
I'm good with this change.  Would have been nice if BTT versioned the variants differently.

KevinOConnor:
Thanks.

-Kevin

PR #5960, opened 2022-12-30
Support for mixing extruders

konsumverweigerer:
Support for mixing (n-in 1-out) nozzles. It uses the sync to stepper/extruder and adjusts the rotation distance for the affected steppers.

This is a redesign of the https://github.com/Klipper3d/klipper/pull/5858 PR and should address its shortcomings/problems.


konsumverweigerer:
Since this is based on  the #5858 PR which was already partially reviewed and the issues found are addressed in this PR.
So I reviewed this change according to the steps in the CONTRIBUTING document and everything looks good to me.

KevinOConnor:
Thanks.  At a high-level, the changes seems okay to me.  I do have some comments though:

1. The number of new commands need to be reduced.  Perhaps just two commands - `SET_MIXING_EXTRUDER` and `SET_MIXING_EXTRUDER_GRADIENT`.  Adding many more commands becomes a maintenance burden as it becomes hard to prevent modules from adding two commands with the same or similar names.  (This is also why new commands should have the module name - `MIXING_EXTRUDER` - in the name.)  It should be possible to use parameters to determine the actions to take in the command - for example, if `FACTORS` is not provided to `SET_MIXING_EXTRUDER` then the current status can be display, if `START_FACTORS` is present in `SET_MIXING_EXTRUDER_GRADIENT` then it can be interpreted as an "add" attempt, etc.
2. This PR adds a `get_status()` method - it is thus required to update the docs/Status_Reference.md document as well.  Also, the resulting status looks like it could be cpu intensive to calculate - many Klipper API clients just subscribe to everything, so we don't want any `get_status()` calls to be cpu intensive.

I also have some additional code comments to follow.
-Kevin

docs/Config_Reference.md
```diff
@@ -2082,6 +2082,26 @@ printer kinematics.
...
 ```
 
+### [mixing_extruder]
+
+A mixing printhead which has <n>in-1out mixing nozzle. When activated
+additional g-code are available. See [G-Codes](G-Codes.md#mixing_extruder)
+for a detailed description of the additional commands.
+
+```
+[mixing_extruder]
```
KevinOConnor, line 2092: It seems odd that this doesn't have a name as in `[mixing_extruder myname]`.

klippy/extras/mixing_extruder.py
```diff
@@ -0,0 +1,335 @@
...
+        self._activated = True
+        try:
+            for name in self.stepper_names:
+                try:
+                    stepper = (self.printer
+                               .lookup_object(name).extruder_stepper)
+                except:
+                    stepper = (self.printer
+                               .lookup_object("extruder_stepper " + name)
+                               .extruder_stepper)
```
KevinOConnor, line 89: This doesn't look a good idea to me.  If the user wants to use an "extruder_stepper mystepper" then they should specify that explicitly in the config file.  I don't think the code should try to infer what stepper is being requested.

klippy/extras/mixing_extruder.py
```diff
@@ -0,0 +1,335 @@
...
+                self._extruder_steppers.append(stepper)
+        except Exception as e:
+            while len(self.steppers) > 0:
+                self.steppers.pop()
+            logging.error("no steppers found: %s: %s",
+                          ", ".join(self.stepper_names), e)
```
KevinOConnor, line 95: If there is an error in the config, an error should be raised.  The code should not ignore a config error.

klippy/extras/mixing_extruder.py
```diff
@@ -0,0 +1,335 @@
...
+        self._trapqs = tuple(
+            stepper.stepper.get_trapq()
+            for stepper in self._extruder_steppers)
+    def get_status(self, eventtime):
+        return dict(
+            mixing_enabled="enabled" if self._enabled else "disabled",
+            mixing=",".join(
+                "%0.1f%%" % (m * 100.) for m in self._active_mixing),
+            mixing_ticks=",".join(
+                "%d" % (extruder_stepper.stepper.get_mcu_position())
```
KevinOConnor, line 108: This looks to me like it is "peeking" into a member of the extruder_stepper object.  As per the Klipper conventions documented in docs/Code_Overview.md this is not valid.

klippy/extras/mixing_extruder.py
```diff
@@ -0,0 +1,335 @@
...
+                                    for x in self._gradient_vector),
+                    enabled=str(self._gradient_enabled)).items())})
+        return status
+    def _check_scales(self, gcmd, s):
+        try:
+            scales = s.replace(',', ':').split(':')
+            if len(scales) != len(self._mixing):
+                raise gcmd.error("Could not configure gradient: invalid scales")
+            return tuple(float(scales[i]) for i in range(len(scales)))
+        except Exception:
```
KevinOConnor, line 281: It's not a good idea to catch any Exception as that often results in real code bugs not being reported properly.  (A real bug will end up corrupting internal state and when the code finally does error, the reported error wont indicate the real source of the issue.)  Prefer to catch just the exception needed.

klippy/extras/mixing_extruder.py
```diff
@@ -0,0 +1,335 @@
...
+    def cmd_RESET_MIXING_GRADIENT(self, gcmd):
+        self._gradient_enabled, self._gradients, self._gradient_method = \
+            False, [], 'linear'
+        self._disable_gradient()
+
+
+def load_config(config):
+    mixingextruder = GradientMixingExtruder(
+        config.getsection('mixing_extruder'))
+    logging.info("Added mixing extruders")
```
KevinOConnor, line 334: Debugging comments should be removed.

thijstriemstra:
This is awesome and I'm looking forward to testing this @konsumverweigerer (for a clay printer I'm working on).

klippy/extras/mixing_extruder.py
```diff
@@ -0,0 +1,320 @@
+# Code for supporting mixing extruders.
+#
+# Copyright (C) 2022 Peter Gruber <nokos@gmx.net>
```
thijstriemstra, line 3: happy new year: 2022-2023

docs/Config_Reference.md
konsumverweigerer: I don't see for what the name can be used.

klippy/extras/mixing_extruder.py
konsumverweigerer: I have adjusted it to use if/else instead of try since the "extruder" is actually fixed in name

klippy/extras/mixing_extruder.py
konsumverweigerer: Adjusted it, will now throw on config errors

klippy/extras/mixing_extruder.py
konsumverweigerer: I have refactored the extruder_stepper/stepper access and as I understand get_mcu_position() is a "public" member of the stepper

klippy/extras/mixing_extruder.py
konsumverweigerer: Have adjusted it to ValueException which may originated from a non-number passed from the user as scale

klippy/extras/mixing_extruder.py
konsumverweigerer: removed

konsumverweigerer:
>     1. The number of new commands need to be reduced.  Perhaps just two commands - `SET_MIXING_EXTRUDER` and `SET_MIXING_EXTRUDER_GRADIENT`. ...

Have reduced it to 2 commands
 
>     2. This PR adds a `get_status()` method - it is thus required to update the docs/Status_Reference.md document as well.  Also, the resulting status looks like it could be cpu intensive to calculate - many Klipper API clients just subscribe to everything, so we don't want any `get_status()` calls to be cpu intensive.

I have reduced the get_status to the minimum and documented it.The cpu cost should be minimal since calculating the stepper position on extruder steppers is straightforward due to no kinematics involved.

thijstriemstra:
> Unfortunately a reviewer has not assigned themselves to this GitHub Pull Request and it is therefore being closed.

That's not true though..

thijstriemstra:
please re-open.

konsumverweigerer:
@Sineos I have addressed and commented on the issues from the review.

Sineos:
@KevinOConnor, seems your input has been addressed. Anything missing?

KevinOConnor:
Unfortunately, it's primarily due to lack of time on my side.  I'll assign this to myself to turn off the bot.

-Kevin

Nathan22211:
like to note I have gotten a mixing hotend working in klipper to some extent without this PR. Just normal mixing so far as well as a random mix implementation to copy zonestar functionality. if you want I can test this PR to see if there's a difference

jpereira-ca:
I see that this is coming soon, if I'm not mistaken! @KevinOConnor, @Sineos, is that true? 
What is the approximate ETA on getting this merged?

jpereira-ca:
> like to note I have gotten a mixing hotend working in klipper to some extent without this PR. Just normal mixing so far as well as a random mix implementation to copy zonestar functionality. if you want I can test this PR to see if there's a difference
> 
> 

What slicer are you using? Cura?

ETE-Design:
@KevinOConnor Any plan for merging this?

Nathan22211:
@jpereira-ca prusa slicer mostly for painting.

Corodius:
Is there anything I could help with to get this possibly merged? As it seems pretty ready fpr testing. Even if it is just that, testing, I am happy to do so if needed - been running mixing hotend done wolith macros for a bit now.

konsumverweigerer:
> Hi, I also have a mixing printer and I'd like to test this feature, is it already merged in the main branch? How can I build and test this feature?

Hi @apalazzi : I usually have https://github.com/konsumverweigerer/klipper merged with this patch. The docs and the config/sample-mixing-extruder.cfg should be helpful to configure it for your printer

JulianStremel:
> Unfortunately, it's primarily due to lack of time on my side. I'll assign this to myself to turn off the bot.
> 
> -Kevin

I understand that your time is limited as everybody elses but is there any update on when to expect this to get merged ?

Some people are really looking forward for this.

Kind regards

JulianStremel:
> Hi, I also have a mixing printer and I'd like to test this feature, is it already merged in the main branch? How can I build and test this feature?

@konsumverweigerer 

I partially ported your pr over to the [Kalico](https://github.com/KalicoCrew/kalico/pull/512) for of klipper.

I just fixed some small stuff and removed gradient support for now because that made ci fail over there and i currently do not have the hardware available to test further.

I don*t want to plagerize your work and would like you to join me on developing this feature over there :)

Really love the klipper ecosystem and i fully understand that @KevinOConnor  is just as humen as everybody else so having a smaller community fork that works on such features is a great thing in my mind and maybe we cen get this feature mainline some day in the future.

konsumverweigerer:
Hi @JulianStremel 

> I don*t want to plagerize your work and would like you to join me on developing this feature over there :)

I don't mind you using the mixing code, I use it on my experimental setup ... and it definitely has room for improvement

I will check out the Kalico project to see if I can help to integrate it there.

PR #5961, opened 2022-12-31
Add TronXY D01 Plus V2 Direct Drive Printer

mcowger:
Add TronXY D01 Plus V2 Direct Drive Printer

### DIFF
--- a/config/printer-tronxy-d01-plus-v2-direct-drive.cfg
+++ b/config/printer-tronxy-d01-plus-v2-direct-drive.cfg
@@ -10,7 +10,7 @@
 # fmw_tronxy.bin file with your klipper.bin file in the update folder.
 # Note that you MUST use this filename, and keep all the rest of the files.
 
-# Place the update folder on the root of a microSD/TF card (so that the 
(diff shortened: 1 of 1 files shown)


JamesH1978:
If someone were to purchase a TronXY D01 Plus V2 would this have direct drive as standard, or is this a mod you have added yourself? 

Thanks
James

mcowger:
The printer is available in 2 (relevant) variants:  With a direct drive or with Bowden style. 

There are no modifications to the printer that i have made.

JamesH1978:
excellent, then someone will review it and make comment. We only allow stock printer configs so as not to confuse other users with same printer

Thanks
James

JamesH1978 (commented):
Hi,

Apologies for the bot, i thought i had assigned myself.

Ok you have missed a couple of things, please review https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying attention to the signed off by line, and we need you to check your submission conforms to the guidelines in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and you commit a change for the print regression test as layed out in that document.

Thanks
James

PR #5962, opened 2022-12-31
palette2: Fix errors with bad handling of NoneType in a few locations

cmroche:
Issue specific to Python 3, NoneType is being used to compare heartbeat time (actually caused by invalid condition operator), and by returning NoneType in timer functions.

Resolves the issue documented here: https://klipper.discourse.group/t/palette-2-connection-giving-error/4923/9


KevinOConnor:
Thanks.

-Kevin

PR #5963, opened 2023-01-01
bed_mesh: relative_reference_index range check

shiftingtech:
Added a simple range check to the relative_reference_index,
so that having it set too large leads to a reasonable error
instead of a confusing index out of range message.

Signed-off-by: Ben Eastep <shifting@shifting.ca>

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -295,8 +295,9 @@ class BedMeshCalibrate:
         self.mesh_config = collections.OrderedDict()
         self._init_mesh_config(config)
         self._generate_points(config.error)
-        if self.relative_reference_index >= len(self.points):
-            raise config.error("Invalid relative_reference_index")
+        if self.relative_reference_index != None:
+            if self.relative_reference_index >= len(self.points):
+                raise config.error("Invalid relative_reference_index")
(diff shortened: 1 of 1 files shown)


zellneralex:
Why not even enhance the message further like 
`Invalid relative_reference_index. Value must be within 0 and x`
where x is the length-1 value

KevinOConnor:
Thanks.  Could we just move the `getint('relative_reference_index', None)` to after `_generate_points()` and use `getint('relative_reference_index', None, minval=0, maxval=len(self.points)-1)`?

Maybe @Arksine has some thoughts?

-Kevin

shiftingtech:
That would essentially silently discard an invalid value,  correct?

My fear there, would be that it would lead to somebody suddenly getting a
mesh that's shifted shifted down, resulting in them driving the toolhead
into the bed with little warning.

Arksine:
Kevin's suggestion will work, it takes advantage of the range checking built in to the `configfile` implementation.  A value outside the specified range would result in a config error.

It would also be necessary to move `self.orig_config['rri'] = self.relative_reference_index` below the `getint` method.

shiftingtech:
Ah, okay.   I'll revisit it when I get a minute then

PR #5964, opened 2023-01-02
ci-install.sh: Add compatibility with ubuntu 22.04 LTS

sandlerr:
The python package consumed here `python-dev` is available under ubuntu 18.04 as defined in the github workflows for this repo, but it is missing from any newer ubuntu release:

https://packages.ubuntu.com/search?keywords=python-dev&searchon=names&exact=1&suite=all&section=all

Instead, ubuntu now makes this available as `python2-dev`:

https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=python2-dev&searchon=names

### DIFF
--- a/scripts/ci-install.sh
+++ b/scripts/ci-install.sh
@@ -14,9 +14,17 @@ mkdir -p ${BUILD_DIR} ${CACHE_DIR}
 ######################################################################
 # Install system dependencies
 ######################################################################
+PYTHON_PKG="python2-dev"
+if test -f /etc/os-release; then
+    . /etc/os-release
+    ubuntu_year=$(echo $VERSION_ID | cut -f 1 -d '.')
+    if [ $ubuntu_year -le 18 ]; then
+        PYTHON_PKG="python-dev"
+    fi
+fi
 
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I'm not sure it makes sense to complicate the build test scripts with version checks.  We could just update to a more recent version of Ubuntu for the build environment (as defined in .github/workflows/build-test.yaml).

-Kevin

sandlerr:
Hi Kevin, would you like me to update this PR accordingly? An updated Ubuntu LTS might change the toolchain version as well. I'm not sure if you produce any binary distributions such as Docker images, and if those are generated using the CI system, but if so there might be unintended consequences for your userbase. That's kind of why I hoped this would be less risky.

KevinOConnor:
No binaries are exported - just need to make sure the tests still pass.  I'll take a look at it.

Thanks,
-Kevin

KevinOConnor:
I updated the github build scripts to use Ubuntu 22.02 (commit df1137ef).

-Kevin

sandlerr:
Thanks!

PR #5965, opened 2023-01-03
Config for Neptune 3 Pro

kozross:
This is a config for the Neptune 3 Pro from Elegoo. I've kept the settings fairly conservative, based on mostly on `M503` output from the stock config and the references linked.

### DIFF
--- a/config/printer-elegoo-neptune3-pro-2023.cfg
+++ b/config/printer-elegoo-neptune3-pro-2023.cfg
@@ -1,0 +1,126 @@
+# This file contains pin mappings for the stock Elegoo Neptune 3 Pro (ZNP Robin Nano_DW V2.2)
+# To use this config, during "make menuconfig" select the  STM32F401 with a
+# "32KiB bootloader" and serial (on USART1 PA10/PA9) communication.
+
+# Note that the "make flash" command does not work with ZNP Robin boards.
+# After running "make", rename the out/klipper.bin file to out/ZNP_ROBIN_NANO.bin
+# Copy the file out/ZNP_ROBIN_NANO.bin to an SD card formatted to FAT32
+# and then restart the printer with the SD card inserted.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+# Core
+
+[mcu]
+serial: /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
+restart_method: command
+
+[printer]
+kinematics: cartesian
+max_velocity: 300
+max_accel: 3000
+max_z_velocity: 5
+max_z_accel: 100
+
+# Steppers
+
+[stepper_x]
+step_pin: PC12
+dir_pin: PB3
+enable_pin: !PD2
+microsteps: 16
+rotation_distance: 40
+endstop_pin: PA13
+position_min: -5
+position_endstop: -5
+position_max: 235
+homing_speed: 50
+
+[stepper_y]
+step_pin: PC11
+dir_pin: PA15
+enable_pin: !PC10
+microsteps: 16
+rotation_distance: 40
+endstop_pin: PB8
+position_endstop: 0
+position_max: 234
+homing_speed: 50
+
+[stepper_z]
+step_pin: PC7
+dir_pin: !PC9
+enable_pin: !PC8
+microsteps: 16
+rotation_distance: 8
+position_min: -0.8
+endstop_pin: probe:z_virtual_endstop
+position_max: 283
+homing_speed: 10
+
+[probe]
+pin: PA8
+# NOTE: Set this to a value based on your printer and bed.
+z_offset: 0
+x_offset: -28.5
+y_offset: 22
+
+[extruder]
+step_pin: PB10
+dir_pin: PB1
+enable_pin: !PC6
+microsteps: 16
+rotation_distance: 8.42
+nozzle_diameter: 0.400
+filament_diameter: 1.750
+heater_pin: PA6
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PC1
+control: pid
+# NOTE: These settings are for PETG, thus 240C at 30% fan.
+pid_Kp: 26.27
+pid_Ki: 1.607
+pid_Kd: 107.380
+min_temp: 0
+max_temp: 260
+max_extrude_only_distance: 100
+
+[heater_bed]
+heater_pin: PA5
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PC0
+control: pid
+# NOTE: These settings are for PETG, thus 80C bed temperature.
+pid_Kp: 70.173
+pid_Ki: 1.418
+pid_Kd: 868.388
+min_temp: 0
+max_temp: 100
+
+# Coooling
+
+[fan]
+pin: PA7
+
+[heater_fan hotend_fan]
+pin: PB0
+
+# Rest
+
+# This is put on the FAN3 pin.
+[led top_LEDs]
+white_pin: PB9
+cycle_time: 0.005
+
+[filament_switch_sensor filament_runout_sensor]
+switch_pin: PB4
+
+[safe_z_home]
+home_xy_position: 143.5, 93
+
+[bed_mesh]
+mesh_min: 10, 25
+mesh_max: 205, 220
+probe_count: 6, 6
+algorithm: bicubic
+speed: 100
--- a/test/klippy/printers.test
+++ b/test/klippy/printers.test
@@ -188,6 +188,7 @@ CONFIG ../../config/printer-twotrees-sapphire-pro-sp-3-2020.cfg
 # Printers using the stm32f401
 DICTIONARY stm32f401.dict
 CONFIG ../../config/printer-artillery-sidewinder-x2-2022.cfg
+CONFIG ../../config/printer-elegoo-neptune3-pro-2023.cfg
 
 # Printers using the stm32f405
 DICTIONARY stm32f405.dict


kozross:
@CleisonManriqueAguirre - I haven't 'waited' at all, since I only sent it 11 hours ago, or 6 hours before your comment. I'd encourage you to just send your PR.

CleisonManriqueAguirre:
https://github.com/CleisonManriqueAguirre/klipper/pull/1 . I first fork and the make a PR , Am I wrong ? . I am new on github but I wanna contribute ... Maybe I should wait

kozross:
According to [this](https://github.com/Klipper3d/klipper/actions/runs/3832296146/jobs/6522439701), it's called 'being on Windows', because Windows insists on CRLF line endings, unlike basically everyone else.

CleisonManriqueAguirre:
I am gladly of receiving your help . Now I have all clear and its without errors. Greetings from Peru .

mygrexit:
Hey @kozross, I have a suggestion: Wouldn't it be better to set the following home-position?

home_xy_position: 143.5, 93

115, 115 is the correct position for the nozzle, but it leads to PROBE_CALIBRATE moving away from the center in order to compensate the XY offset of the probe.

I'm not that familiar with Klipper, maybe there is a better way for fixing that, but I thought it would be good to let you know.

kozross:
@mygrexit - yes, that position is probably better. I used 115,115 out of laziness mainly. Will fix it and push.

config/printer-elegoo-neptune3-pro-2023.cfg
```diff
@@ -0,0 +1,138 @@
...
+# is normal, and the flash itself will only take a few seconds.
+#
+# See docs/Config_Reference.md for a description of parameters.
+
+# Core
+
+[mcu]
+serial: /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
+restart_method: command
+baud: 115200
```
JamesH1978, line 27: Please do not select a custom baud other than the default 250000. Unless there is a very good reason to set this. Same goes for the comment in the build instructions

JamesH1978 (changes requested):
Good afternoon,

Thank you for submitting this config to the repo, I have a few comments. You will need to add a signed offline as per https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

And you also need to add regression tests as per point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Thanks
James

config/printer-elegoo-neptune3-pro-2023.cfg
kozross: There _is_ a very good reason. The board will suffer from random disconnects otherwise. I've tested this fairly extensively, and in fact, the Marlin config for this board requires this baud rate (which was what tipped me off).

config/printer-elegoo-neptune3-pro-2023.cfg
```diff
@@ -0,0 +1,138 @@
...
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PC1
+control: pid
+# NOTE: These settings are for PETG, thus 240C at 30% fan.
+pid_Kp: 26.27
+pid_Ki: 1.607
+pid_Kd: 107.380
+min_temp: 15
+max_temp: 260
+min_extrude_temp: 135
```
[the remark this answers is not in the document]
kozross: `min_extrude_temp` I presume?

config/printer-elegoo-neptune3-pro-2023.cfg
JamesH1978: Then that is a good reason, though marlin baud rates work differently to klipper board connection, and 250000 default usually works. But as this is tested by you.

Thanks
James

kozross: I'll note this in the comments at the top to clarify why I set it this way.

config/printer-elegoo-neptune3-pro-2023.cfg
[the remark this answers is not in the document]
JamesH1978: Yes, default it is 170, as this i a good starting point for materials with a lower temperature such as PLA. The default is designed to stop a cold extruder which could be detrimental.

Thanks
James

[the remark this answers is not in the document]
kozross: OK, thanks. I will fix that.

kozross:
@JamesH1978 - this should now be compliant with everything. Please let me know if I've missed anything.

1271134893:
Excuse me, is this configuration open to WiFi? If not, can we consider adding WiFi.

kozross:
> Excuse me, is this configuration open to WiFi? If not, can we consider adding WiFi.

This doesn't really matter: the device running Klipper is what you connect to, not the printer itself. If the device running Klipper has wifi, you can reach it via wifi using something like Fluidd or Mainsail. The Klipper config has nothing to do with it.

JamesH1978 (commented):
Thank you for the changes that you have made. 

Please review point 7 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Your instruction header should follow the same patterns as other configs and should not include URLs, as these can change.

Thanks
James

BongoEADGC6:
@kozross Is the intention that this also works for the Neptune 3 Plus?

kozross:
@BongoEADGC6 - I don't have a Neptune 3 Plus, so I don't exactly know. Other than the obvious (bed dimensions, Z maximum, etc), we'd need to check if the pins on the board are the same, and the extruder and probe are the same. If that's the case, there's a good chance.

kozross:
@JamesH1978 - done. Hope this is good enough to merge now.

config/printer-elegoo-neptune3-pro-2023.cfg
```diff
@@ -0,0 +1,141 @@
...
+filament_diameter: 1.750
+heater_pin: PA6
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PC1
+control: pid
+# NOTE: These settings are for PETG, thus 240C at 30% fan.
+pid_Kp: 26.27
+pid_Ki: 1.607
+pid_Kd: 107.380
+min_temp: 15
```
JamesH1978, line 99: This may limit someone in an area where the ambient temperature is below 15, please set to 0

config/printer-elegoo-neptune3-pro-2023.cfg
```diff
@@ -0,0 +1,141 @@
...
+[heater_bed]
+heater_pin: PA5
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PC0
+control: pid
+# NOTE: These settings are for PETG, thus 80C bed temperature.
+pid_Kp: 70.173
+pid_Ki: 1.418
+pid_Kd: 868.388
+min_temp: 15
```
JamesH1978, line 112: same for here, this may limit someone in an area where the ambient temperature is below 15, please set to 0

JamesH1978 (changes requested):
Please sign off your second commit(and any subsequent commits) and change the min temp temp values as commented.

Thanks
James

config/printer-elegoo-neptune3-pro-2023.cfg
kozross: Isn't 0 the default? If so, should I just erase that line, as per the submission criteria?

JamesH1978: yes you can do that too

kozross:
@JamesH1978 Addressed your feedback, hopefully.

KevinOConnor:
Thanks.  However I'm leery of making a recommendation for a baud of 115200.  This would be the first printer that requires a custom baud rate in Klipper.  (Many, many people have an incorrect understanding of baud requirements.)  So, I'd like to understand how the 115200 requirement was determined and what the root cause analysis for that requirement is.

Cheers,
-Kevin

kozross:
@KevinOConnor - I was having some significant issues with communication dropout at the standard baud rate. This basically went away when I dropped down to the current rate. For what it's worth, this is the configuration I've been running since about mid-December 2022, for almost daily printing, and the issues went away when I stepped down the baud rate.

I'm willing to do more testing by using a different Klipper board, but it'll take me a few days to a week to be sure. As long as you don't mind waiting, I'm happy to do so.

KevinOConnor:
Thanks.  I don't mind waiting, though I'm not sure why you'd use a different board.

What host computer are you running Klipper on?  What type of physical connection is there between micro-controller and host?  Can you attach a Klipper log file here from an error event after returning host and mcu to the standard 250000?

-Kevin

kozross:
Host is a [Lime 2](https://www.olimex.com/Products/OLinuXino/A20/A20-OLinuXino-LIME2/open-source-hardware), and the physical connection is a USB cable. I'll switch back to 250,000 and report if anything happens, but if I don't see anything for a week, I'll assume it's solved and update the config.

KevinOConnor:
Okay, it sounds like there is a usb-to-serial chip on the printer board (usb from host to printer board, usb-to-serial chip on the printer board, serial from usb chip to micro-controller on the printer board).

Do you know what usb-to-serial chip is on the printer board?  This is usually obtained by running `lsusb` or by running `dmesg` just after plugging in the usb cable.

-Kevin

AllenRDCo:
I came across this last night. The USB/Serial chip is a CH340C. It's supposed to support up to 2 million baud, so 250k baud shouldn't be an issue. 

I'm going to be loading Klipper on my Neptune 3 Pro so I will also check to see if there is any issue like this on my setup.

kozross:
@AllenRDCo - you beat me to it. Yeah, that is indeed the chip on the Neptune 3 Pro's board.

I've upped my baud rate back up to the default: if I don't see issues for a week, I'll consider this solved, and will amend the config accordingly.

AllenRDCo:
Been running mine at 250000 baud for a couple days now, almost non-stop (lots of calibration). No issues with communication dropping for me.

kozross:
@KevinOConnor - OK, I've figured out why I was having weirdness issues with baud rate, and can confirm that the rate had nothing to do with it. It was actually some weird interference from my Klipper board, now resolved with a USB hub. I've fixed the config to no longer suggest (or use) a different baud rate.

I also found 3K accels worked fine without ringing, so I fixed that up too. H

Hope this is now good to merge. Let me know if I need to fix anything else.

config/printer-elegoo-neptune3-pro-2023.cfg
```diff
@@ -0,0 +1,135 @@
...
+# * Copy `out/ZNP_ROBIN_NANO.bin` to the root of an SD card formatted to FAT32
+# * Insert the card, then restart
+#
+# The display, if connected, will loop in the 'Updating firmware' screen. This
+# is normal, and the flash itself will only take a few seconds.
+#
+# See docs/Config_Reference.md for a description of parameters.
+
+# Core
+
```
JamesH1978, line 22: looks fine in terms of the config, but your intro section does not conform to the prescribed format, according to point 7 i in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md. I would suggest:
```
# This file contains pin mappings for the stock Elegoo Neptune 3 Pro (ZNP Robin Nano_DW V2.2)
# To use this config, during "make menuconfig" select the
# STM32F401 with a "32KiB bootloader" and serial (on USART1 PA10/PA9)
# communication.

# Note that the "make flash" command does not work with MKS Robin
# boards. After running "make", rename the out/klipper.bin file to out/ZNP_ROBIN_NANO.bin
# Copy the file out/ZNP_ROBIN_NANO.bin to an SD card formatted to FAT32
# and then restart the printer with the SD card inserted.

# See docs/Config_Reference.md for a description of parameters.```

I assume this board does not the usual robin script running on it to create a useable bin file? ie `./scripts/update_mks_robin.py`

Apart from that, all looks good.

Thanks
James

JamesH1978:
thanks, but that's not what I meant, if the script is not used to process the Robin file on this particular printer there no need to change that or create a script to do it. What I'm saying is that usually you have to run a special script on the bin file that doesn't just rename it, it changes the pattern of the internals of the file. But if you did not use it to create the bin for this printer then it is not needed

Thanks

James

kozross:
@JamesH1978: Nothing but a filename change is needed - the binary remains the same.

KevinOConnor:
Thanks.

-Kevin

PR #5966, opened 2023-01-03
Config file for Artillery genius pro

CleisonManriqueAguirre:
I recently created this config file for those who have the Artillery genius pro


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "cmake.sourceDirectory": "${workspaceFolder}/lib/rp2040/boot_stage2"
```
JamesH1978, line 2: Can i ask why this is needed?

config/example_artillery_genius_pro_2023.cfg
```diff
@@ -0,0 +1,153 @@
+[include mainsail.cfg]
```
JamesH1978, line 1: Please do not put custom includes in stock files, the end user may not have mainsail

config/example_artillery_genius_pro_2023.cfg
```diff
@@ -0,0 +1,153 @@
...
+pin: PC8
+off_below: 0.1
+[heater_fan extruder]
+pin: PC7
+off_below: 0.2
+[controller_fan case]
+pin: PC6
+off_below: 0.3
+idle_speed: 0.0
+[temperature_sensor mainboard]
```
JamesH1978, line 78: Please do not add custom temperature sensors for mcu's and hosts, these may not be available on the end users equipment

config/example_artillery_genius_pro_2023.cfg
```diff
@@ -0,0 +1,153 @@
...
+screw1_name: front left
+screw2: 143,63
+screw2_name: front right
+screw3: 143,183
+screw3_name: back right
+screw4: 23,183
+screw4_name: back left
+speed: 100.0
+screw_thread: CW-M5
+[mcu]
```
JamesH1978, line 124: You have stated [mcu] twice

JamesH1978 (changes requested):
Good afternoon,

Thank you for submitting this config to the repo, I have a few comments. You will need to add a signed offline as per https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

And you also need to add regression tests as per point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and review other needs in there too.

Thanks
James

config/example_artillery_genius_pro_2023.cfg
CleisonManriqueAguirre: Hi, they does work perfectly on Artillery genius .

config/example_artillery_genius_pro_2023.cfg
CleisonManriqueAguirre: Yes , I will solve it

JamesH1978 (commented):
Please refer to my original comment, you need to:

You will need to add a signed offline as per https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

You also need to add regression tests as per point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and review other needs in there too. Please make the file look the same as other configs, and include flash instructions at the top of the file. Also point 1v this would not be an example but a printer.

Thanks
James

PR #3828, opened 2021-01-23
Add samples_result option to [probe] last

Stephan3:
Most of my probes are showing the phenomena that the first probe result is slightly off. See the following examples.

Fotek PL08N:
```
23.1.2021, 10:23:57: : PROBE_ACCURACY at X:125.000 Y:78.000 Z:20.662 (samples=10 retract=1.000 speed=5.0 lift_speed=50.0)
23.1.2021, 10:23:57: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:23:58: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:23:58: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:23:59: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:23:59: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:24:00: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:24:01: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:24:01: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:24:02: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:24:02: : probe at 125.000,78.000 is z=4.700000
23.1.2021, 10:24:02: : probe accuracy results: maximum 4.700000, minimum 4.700000, range 0.000000, average 4.700000, median 4.700000, standard deviation 0.000000

23.1.2021, 10:24:08: : PROBE_ACCURACY at X:125.000 Y:78.000 Z:25.026 (samples=10 retract=1.000 speed=5.0 lift_speed=50.0)
23.1.2021, 10:24:10: : probe at 125.000,78.000 is z=4.637500
23.1.2021, 10:24:10: : probe at 125.000,78.000 is z=4.650000
23.1.2021, 10:24:11: : probe at 125.000,78.000 is z=4.650000
23.1.2021, 10:24:11: : probe at 125.000,78.000 is z=4.650000
23.1.2021, 10:24:12: : probe at 125.000,78.000 is z=4.650000
23.1.2021, 10:24:12: : probe at 125.000,78.000 is z=4.650000
23.1.2021, 10:24:13: : probe at 125.000,78.000 is z=4.650000
23.1.2021, 10:24:14: : probe at 125.000,78.000 is z=4.650000
23.1.2021, 10:24:14: : probe at 125.000,78.000 is z=4.650000
[...]

PR #5969, opened 2023-01-03
Richardjm/probe discard first

richardjm:
Supports the ability to discard the first n probes. Allowing a faster first probe before accurate subsequent measurements.

There are many users implementing this mechanism via manually replacing the probe.py file with a version that supports discarding the first probe.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1817,6 +1817,10 @@ z_offset:
 #   samples. If this tolerance is exceeded then either an error is
 #   reported or the attempt is restarted (see
 #   samples_tolerance_retries). The default is 0.100mm.
+#samples_discard_first: 0
+#   The number of samples to discard when starting to probe. This can
+#   be used if the first probe is often less accurate. The default is
+#   zero.
 #samples_tolerance_retries: 0
 #   The number of times to retry if a sample is found that exceeds
 #   samples_tolerance. On a retry, all current samples are discarded
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -48,6 +48,8 @@ class PrinterProbe:
                                                  minval=0.)
         self.samples_retries = config.getint('samples_tolerance_retries', 0,
                                              minval=0)
+        self.samples_discard = config.getint('samples_discard_first', 0,
+                                             above=-1)
         # Register z_virtual_endstop pin
         self.printer.lookup_object('pins').register_chip('probe', self)
         # Register homing event handlers
@@ -154,16 +156,24 @@ class PrinterProbe:
                                            self.samples_tolerance, minval=0.)
         samples_retries = gcmd.get_int("SAMPLES_TOLERANCE_RETRIES",
                                        self.samples_retries, minval=0)
+        samples_discard_first = gcmd.get_int("SAMPLES_DISCARD_FIRST",
+                                             self.samples_discard, above=-1)
         samples_result = gcmd.get("SAMPLES_RESULT", self.samples_result)
         must_notify_multi_probe = not self.multi_probe_pending
         if must_notify_multi_probe:
             self.multi_probe_begin()
         probexy = self.printer.lookup_object('toolhead').get_position()[:2]
         retries = 0
+        discards = 0
         positions = []
         while len(positions) < sample_count:
             # Probe position
             pos = self._probe(speed)
+            # Discard the first readings (probably at most one)
+            if discards <= samples_discard_first:
+                gcmd.respond_info("Probe sample discarded")
+                discards += 1
+                continue
             positions.append(pos)
             # Check samples tolerance
             z_positions = [p[2] for p in positions]
@@ -204,6 +214,8 @@ class PrinterProbe:
         sample_count = gcmd.get_int("SAMPLES", 10, minval=1)
         sample_retract_dist = gcmd.get_float("SAMPLE_RETRACT_DIST",
                                              self.sample_retract_dist, above=0.)
+        samples_discard_first = gcmd.get_int("SAMPLES_DISCARD_FIRST",
+                                             self.samples_discard, above=-1)
         toolhead = self.printer.lookup_object('toolhead')
         pos = toolhead.get_position()
         gcmd.respond_info("PROBE_ACCURACY at X:%.3f Y:%.3f Z:%.3f"
@@ -214,10 +226,16 @@ class PrinterProbe:
                              speed, lift_speed))
         # Probe bed sample_count times
         self.multi_probe_begin()
+        discards = 0
         positions = []
         while len(positions) < sample_count:
             # Probe position
             pos = self._probe(speed)
+            # Discard the first readings (probably at most one)
+            if discards <= samples_discard_first:
+                gcmd.respond_info("Probe sample discarded")
+                discards += 1
+                continue
             positions.append(pos)
             # Retract
             liftpos = [None, None, pos[2] + sample_retract_dist]
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -173,16 +173,16 @@ class PrinterProbe:
             if discards <= samples_discard_first:
                 gcmd.respond_info("Probe sample discarded")
                 discards += 1
-                continue
-            positions.append(pos)
-            # Check samples tolerance
-            z_positions = [p[2] for p in positions]
-            if max(z_positions) - min(z_positions) > samples_tolerance:
-                if retries >= samples_retries:
-                    raise gcmd.error("Probe samples exceed samples_tolerance")
-                gcmd.respond_info("Probe samples exceed tolerance. Retrying...")
-                retries += 1
-                positions = []
+            else:
+              positions.append(pos)
+              # Check samples tolerance
+              z_positions = [p[2] for p in positions]
+              if max(z_positions) - min(z_positions) > samples_tolerance:
+                  if retries >= samples_retries:
+                      raise gcmd.error("Probe samples exceed samples_tolerance")
+                  gcmd.respond_info("Probe samples exceed tolerance. Retrying...")
+                  retries += 1
+                  positions = []
             # Retract
             if len(positions) < sample_count:
                 self._move(probexy + [pos[2] + sample_retract_dist], lift_speed)
@@ -235,8 +235,8 @@ class PrinterProbe:
             if discards <= samples_discard_first:
                 gcmd.respond_info("Probe sample discarded")
                 discards += 1
-                continue
-            positions.append(pos)
+            else:
+              positions.append(pos)
             # Retract
             liftpos = [None, None, pos[2] + sample_retract_dist]
             self._move(liftpos, lift_speed)
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -49,7 +49,7 @@ class PrinterProbe:
         self.samples_retries = config.getint('samples_tolerance_retries', 0,
                                              minval=0)
         self.samples_discard = config.getint('samples_discard_first', 0,
-                                             above=-1)
+                                             minval=0)
         # Register z_virtual_endstop pin
         self.printer.lookup_object('pins').register_chip('probe', self)
         # Register homing event handlers
@@ -157,7 +157,7 @@ class PrinterProbe:
         samples_retries = gcmd.get_int("SAMPLES_TOLERANCE_RETRIES",
                                        self.samples_retries, minval=0)
         samples_discard_first = gcmd.get_int("SAMPLES_DISCARD_FIRST",
-                                             self.samples_discard, above=-1)
+                                             self.samples_discard, minval=0)
         samples_result = gcmd.get("SAMPLES_RESULT", self.samples_result)
         must_notify_multi_probe = not self.multi_probe_pending
         if must_notify_multi_probe:
@@ -215,7 +215,7 @@ class PrinterProbe:
         sample_retract_dist = gcmd.get_float("SAMPLE_RETRACT_DIST",
                                              self.sample_retract_dist, above=0.)
         samples_discard_first = gcmd.get_int("SAMPLES_DISCARD_FIRST",
-                                             self.samples_discard, above=-1)
+                                             self.samples_discard, minval=0)
         toolhead = self.printer.lookup_object('toolhead')
         pos = toolhead.get_position()
         gcmd.respond_info("PROBE_ACCURACY at X:%.3f Y:%.3f Z:%.3f"
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -170,7 +170,7 @@ class PrinterProbe:
             # Probe position
             pos = self._probe(speed)
             # Discard the first readings (probably at most one)
-            if discards <= samples_discard_first:
+            if discards < samples_discard_first:
                 gcmd.respond_info("Probe sample discarded")
                 discards += 1
             else:
@@ -179,10 +179,13 @@ class PrinterProbe:
               z_positions = [p[2] for p in positions]
               if max(z_positions) - min(z_positions) > samples_tolerance:
                   if retries >= samples_retries:
-                      raise gcmd.error("Probe samples exceed samples_tolerance")
-                  gcmd.respond_info("Probe samples exceed tolerance. Retrying...")
+                      raise gcmd.error(
+                                "Probe samples exceed samples_tolerance")
+                  gcmd.respond_info(
+                      "Probe samples exceed tolerance. Retrying...")
                   retries += 1
                   positions = []
+                  # Do not reset discards as they are done once per probe
             # Retract
             if len(positions) < sample_count:
                 self._move(probexy + [pos[2] + sample_retract_dist], lift_speed)
@@ -232,7 +235,7 @@ class PrinterProbe:
             # Probe position
             pos = self._probe(speed)
             # Discard the first readings (probably at most one)
-            if discards <= samples_discard_first:
+            if discards < samples_discard_first:
                 gcmd.respond_info("Probe sample discarded")
                 discards += 1
             else:
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -922,7 +922,7 @@ setting in the [probe config section](Config_Reference.md#probe).
 "open").
 
 #### PROBE_ACCURACY
-`PROBE_ACCURACY [PROBE_SPEED=<mm/s>] [SAMPLES=<count>]
+`PROBE_ACCURACY [PROBE_SPEED=<mm/s>] [LIFT_SPEED=<mm/s>] [SAMPLES=<count>]
 [SAMPLE_RETRACT_DIST=<mm>] [SAMPLES_DISCARD_FIRST=<count>]`: Calculate the
 maximum, minimum, average, median, and standard deviation of multiple probe
 samples. By default,10 SAMPLES are taken. Otherwise the optional parameters
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -911,7 +911,8 @@ see the [probe calibrate guide](Probe_Calibrate.md)).
 #### PROBE
 `PROBE [PROBE_SPEED=<mm/s>] [LIFT_SPEED=<mm/s>] [SAMPLES=<count>]
 [SAMPLE_RETRACT_DIST=<mm>] [SAMPLES_TOLERANCE=<mm>]
-[SAMPLES_TOLERANCE_RETRIES=<count>] [SAMPLES_RESULT=median|average]`:
+[SAMPLES_TOLERANCE_RETRIES=<count>] [SAMPLES_RESULT=median|average]
+[SAMPLES_DISCARD_FIRST=<count>]`:
 Move the nozzle downwards until the probe triggers. If any of the
 optional parameters are provided they override their equivalent
 setting in the [probe config section](Config_Reference.md#probe).
@@ -922,10 +923,10 @@ setting in the [probe config section](Config_Reference.md#probe).
 
 #### PROBE_ACCURACY
 `PROBE_ACCURACY [PROBE_SPEED=<mm/s>] [SAMPLES=<count>]
-[SAMPLE_RETRACT_DIST=<mm>]`: Calculate the maximum, minimum, average,
-median, and standard deviation of multiple probe samples. By default,
-10 SAMPLES are taken. Otherwise the optional parameters default to
-their equivalent setting in the probe config section.
+[SAMPLE_RETRACT_DIST=<mm>] [SAMPLES_DISCARD_FIRST=<count>]`: Calculate the
+maximum, minimum, average, median, and standard deviation of multiple probe
+samples. By default,10 SAMPLES are taken. Otherwise the optional parameters
+default to their equivalent setting in the probe config section.
 
 #### PROBE_CALIBRATE
 `PROBE_CALIBRATE [SPEED=<speed>] [<probe_parameter>=<value>]`: Run a


klippy/extras/probe.py
```diff
@@ -154,16 +156,24 @@ def run_probe(self, gcmd):
...
+        discards = 0
         positions = []
         while len(positions) < sample_count:
             # Probe position
             pos = self._probe(speed)
+            # Discard the first readings (probably at most one)
+            if discards <= samples_discard_first:
+                gcmd.respond_info("Probe sample discarded")
+                discards += 1
+                continue
```
Leandro-Rocha, line 176: This continue will skip the retract down the loop that must happen to avoid "Probe triggered prior to movement" error

klippy/extras/probe.py
```diff
@@ -214,10 +226,16 @@ def cmd_PROBE_ACCURACY(self, gcmd):
...
```
Leandro-Rocha, line 238: This continue will skip the retract down the loop that must happen to avoid "Probe triggered prior to movement" error

klippy/extras/probe.py
richardjm: Thanks, I believe I've fixed that.

klippy/extras/probe.py
richardjm: Thanks, I believe I've fixed that.

MugenMicko:
This would be amazing!! <3

mjoconr:
I also use something like this and feel its needed. First value from a probe cycle seem to always have a large difference from the following.

KevinOConnor:
Thanks.  However, there was a long discussion on this at #3828 .  My comments from that thread still apply - I don't think it makes sense to change the code without first diagnosing the root cause of inaccurate first probe results.

FWIW, my voron2 printer with klicky probe also shows this phenomenon.  So, maybe I'll get a chance to diagnose root cause (but it could be months away before I get to it).

-Kevin

GadgetAngel:
I would love to see this feature of discard_first. Why use a data point in the samples when you know that something in the motion system needs more time to settle and is causing the first sample to be thrown off a bit.

To me discarding the first sample fixes the issue with the motion system not being settled when the fist sample is taken.  

 I understand the difference between median and average for the samples_result option but to me using a data point that I know will always tend to be an outlier in the data set does not make sense.
 
 I have isolated my signals for the probe (Euclid) by using shielded cabling and I have switched the micro-switch to using a Kailh GM 4.0 Micro Switch (https://www.amazon.com/dp/B08K9FSSLH).  These Kailh GM 3.0 switches are used in mice for computers for years.
 
 I did extensive testing with PROBE_ACCURACY which you can look at with this URL: https://github.com/GadgetAngel/Klipperbackup_Voron2.4r2_LDO300kit/tree/master/probe_accuracy_test_suite_Test_Results and https://github.com/GadgetAngel/Klipperbackup_Voron2.4r2_LDO300kit/tree/master/Probe_Accuracy_Test_Results
 
 From my experience it is usually the first sample that is off in a set of samples.

terraphantm:
> 

My understanding is that it's a mechanical problem inherent to the belted Z drives. Swapping to a worm gear setup is known to cure the issue, but that brings its own problems

ChipCE:
> I don't think it makes sense to change the code without first diagnosing the root cause of inaccurate first probe results.

Depend on the setup, there will be faster first probe and slower on the others. Some probe have sampling rate (specially capacitive type have quite low sampling rate), so the homing speed will affect the first probe accuracy.

mattthebaker:
I have some unique data that may be interesting to the conversation. Everything I have seen points to the mechanics of a printer with this issue needing the z motion impulse from the first probe dive to "settle" the toolhead into a more stable resting position.

Here is a measurement of z resonance on one machine showing 20 um peak to peak from "only" 10mm/s dive speed. Faster speeds will have an even more violent stop.
image

Using ranging measurements on a corexy (troodon) with particularly sloppy mechanics, 10um (peak to peak) variation in z offset was found from only moving the toolhead out and back in the x and y -- that is, 10um variation in z offset cross coupled from presumably x-y backlash, belt tension asymmetry, poorly constrained rail cart or some related x/y issue. The z never moved so much as a microstep. An aggressive impulse from homing stop would help release asymmetric tension or just vibrate it into a more natural/stable position.

My two personal machines are leadscrew driven, bed-lift z axes and do not have this issue. A z impulse is not strongly coupled to the x-y, only through the frame, so if there were something in the mechanics that needed to settle, it would not receive a strong "bump" from the homing stop.

It may not specifically be something in the x-y settling, a belted z may instead have something that settles directly on the z axis.


My suggestion to get more definitive data would be to have code similar to this pull request just generate a z impulse without ever triggering the probe -- get it up to homing speed and just stop. This will isolate mechanical settling from anything that may be present in the probe or signaling. Samples subsequent to the impulse should not have the bad first sample if this theory is correct.

I would have already tested this out of curiosity but neither of my machines exhibit this issue.

KevinOConnor:
> Everything I have seen points to the mechanics of a printer with this issue needing the z motion impulse from the first probe dive to "settle" the toolhead into a more stable resting position.

Interesting.  Though, if this was the case, I'd expect to see it only on the first probe in a QUAD_GANTRY_LEVEL (or BED_MESH_CALIBRATE or Z_TILT_ADJUST) sequence.  However, I've seen the issue reported on each probe descent, even when previous probe descents occurred recently.

I agree that tests like you described would help track down root cause.

Cheers,
-Kevin

mattthebaker:
> Interesting. Though, if this was the case, I'd expect to see it only on the first probe in a QUAD_GANTRY_LEVEL (or BED_MESH_CALIBRATE or Z_TILT_ADJUST) sequence. However, I've seen the issue reported on each probe descent, even when previous probe descents occurred recently.

The mentioned data from the troodon suggests that any movement in the x-y could "unsettle" the toolhead and require the impulse again. Though again it may not always be due to x-y depending on the machine.

Fortunately this can also be tested relatively easily by probing, waiting awhile without moving in the x-y, and then probing again to check for bad first value. Maybe someone on this thread can test and report their findings, as it doesn't require any new code.

GadgetAngel:
> > Interesting. Though, if this was the case, I'd expect to see it only on the first probe in a QUAD_GANTRY_LEVEL (or BED_MESH_CALIBRATE or Z_TILT_ADJUST) sequence. However, I've seen the issue reported on each probe descent, even when previous probe descents occurred recently.
> 
> The mentioned data from the troodon suggests that any movement in the x-y could "unsettle" the toolhead and require the impulse again. Though again it may not always be due to x-y depending on the machine.
> 
> Fortunately this can also be tested relatively easily by probing, waiting awhile without moving in the x-y, and then probing again to check for bad first value. Maybe someone on this thread can test and report their findings, as it doesn't require any new code.

If you want to do further testing, I have a version of this in an extension to Klipper called ``z_v2settling_probe`` that @voidtrance and I worked on.  My GitHub repo is at https://github.com/GadgetAngel/voron2_settling_probe

@voidtrance has his version called ``settling_probe`` at https://github.com/voidtrance/voron-klipper-extensions

The difference between ``settling_probe`` and ``z_v2settling_probe`` is that ``z_v2settling_probe`` is a tool to help test out this issue. 

``z_v2settling_probe`` allows a command line parameter to be added to the following commands:
```BASH
PROBE
PROBE_ACCURACY
PROBE_CALIBRATE
QUAD_GANTRY_LEVEL
BED_MESH_CALIBRATE
PROBE_Z_ACCURACY
CALIBRATE_Z
```
The command line parameter the user can add to the above command enables the user to discard the first settling sample on the fly instead of having to `RESTART` or `FIRMWARE_RESTART` their 3D printer (which changes the state of the machine).  

The command line parameter is called ``SETTLING_PROBE`` so the commands would look as follows with `z_v2settling_probe` Klipper extension:
```BASH
PROBE
PROBE SETTLING_SAMPLE=0
PROBE SETTLING_SAMPLE=1

PROBE_ACCURACY
PROBE_ACCURACY SETTLING_SAMPLE=0
PROBE_ACCURACY SETTLING_SAMPLE=1

PROBE_CALIBRATE
PROBE_CALIBRATE SETTLING_SAMPLE=0
PROBE_CALIBRATE SETTLING_SAMPLE=1

QUAD_GANTRY_LEVEL
QUAD_GANTRY_LEVEL SETTLING_SAMPLE=0
QUAD_GANTRY_LEVEL SETTLING_SAMPLE=1

BED_MESH_CALIBRATE
BED_MESH_CALIBRATE SETTLING_SAMPLE=0
BED_MESH_CALIBRATE SETTLING_SAMPLE=1

PROBE_Z_ACCURACY
PROBE_Z_ACCURACY SETTLING_SAMPLE=0
PROBE_Z_ACCURACY SETTLING_SAMPLE=1

CALIBRATE_Z
CALIBRATE_Z SETTLING_SAMPLE=0
CALIBRATE_Z SETTLING_SAMPLE=1
```
A `SETTLING_SAMPLE=0` means you **do not want** to discard the first settling sample, while `SETTLING_SAMPLE=1` means that you **do want** to discard the first settling sample.

My Ender3 Pro with a BLTouch definitely has an issue but my Ender3 is down at the moment.  When I do have it back up I will be adding the Klipper extension `z_v2settling_probe` to that machine.

If someone else here wants to try out this extension and provide data I would like to see the test data.  

The `z_v2settling_probe` Klipper extension does not require you to change any of the Klipper files.  The Klipper `z_v2settling_probe` extension and the Klipper `settling_probe` extension just extends the `probe object` to so that you can choose to discard the first settling probe point from each sample group.

My hope is that tool can help settle this issue! No pun intended.

GarthSnyder:
I would like to see this PR integrated into the mainline release as well.

Not wanting Klipper to normalize (in the societal sense) sloppy construction seems like a reasonable goal. Klipper shouldn't be ushering users toward the wrong fixes for problems. And providing support for for work-arounds is pretty much equivalent to tacitly endorsing the use of those work-arounds.

However, I don't think first-probe error on mechanical probes generally falls into the category of things that are easily or straightforwardly fixable. My experience is the same as the OP's in #3828: every probe on every machine I've ever personally owned has exhibited this same issue. I worked long and hard in every case to minimize the first-probe deviation, but it never went away completely.

I would analogize this problem to that of "vertical fine artifacts" from belt noise. You can replace all your belts and all your pulleys, retension your belts, tighten your screws, input-shape your printer within an inch of its life, and you'll still never completely get rid of those artifacts. The best you can hope for is to get them so subtle that they can be detected only under the harshest lighting. There's a limit to what a given printer can achieve.

By not supporting a straightforward "ignore the first probe" option, Klipper is more or less pushing users toward other suboptimal hacks. For example, using median sampling or setting the probe tolerance so that a second round of probing is usually assured. 

I'd rather have an honest configuration that drops probes but reflects the tolerance I expect the probe to achieve in practice. Then if I see retries start to creep in, I know something's up and can investigate.

richardjm:
> If I may suggest a couple of modifications to the current PR:
> 
> 1. This setting should really apply to Z homing as well as PROBE. I looked at the code briefly, but homing seems to have a rather different implementation.
> 2. A minor nit: the current "Probe sample discarded" message creates misleading visual groupings:
> 
> ```
> probe at 30.000,270.000 is z=-0.659781
> Probe sample discarded
> probe at 30.000,270.000 is z=-0.655406
> probe at 30.000,270.000 is z=-0.656031
> probe at 30.000,270.000 is z=-0.656031
> probe at 270.000,270.000 is z=-0.664781
> Probe sample discarded
> probe at 270.000,270.000 is z=-0.659156
> probe at 270.000,270.000 is z=-0.659156
> probe at 270.000,270.000 is z=-0.658531
> probe at 270.000,30.000 is z=-0.670406
> Probe sample discarded
> probe at 270.000,30.000 is z=-0.664781
> probe at 270.000,30.000 is z=-0.661656
> probe at 270.000,30.000 is z=-0.659781
> probe at 270.000,30.000 is z=-0.659781
> ```
> 
> The message follows the discarded samples, so the four probes in each "group" are actually from two different probing sets. Better, I think, to just mark the discards inline:
> 
> ```
> probe at 30.000,270.000 is z=-0.659781 (discarded)
> probe at 30.000,270.000 is z=-0.655406
> probe at 30.000,270.000 is z=-0.656031
> probe at 30.000,270.000 is z=-0.656031
> probe at 270.000,270.000 is z=-0.664781 (discarded)
> probe at 270.000,270.000 is z=-0.659156
> probe at 270.000,270.000 is z=-0.659156
> probe at 270.000,270.000 is z=-0.658531
> probe at 270.000,30.000 is z=-0.670406 (discarded)
> probe at 270.000,30.000 is z=-0.664781
> probe at 270.000,30.000 is z=-0.661656
> probe at 270.000,30.000 is z=-0.659781
> probe at 270.000,30.000 is z=-0.659781
> ```
> 
> There's a fork of @richardjm's code that fixes this [here](https://github.com/GarthSnyder/klipper/tree/probe-discard-first).

Want me to integrate your changes into my PR? Happy to cherry pick across if you want.

richardjm:
> > > Interesting. Though, if this was the case, I'd expect to see it only on the first probe in a QUAD_GANTRY_LEVEL (or BED_MESH_CALIBRATE or Z_TILT_ADJUST) sequence. However, I've seen the issue reported on each probe descent, even when previous probe descents occurred recently.
> > 
> > 
> > The mentioned data from the troodon suggests that any movement in the x-y could "unsettle" the toolhead and require the impulse again. Though again it may not always be due to x-y depending on the machine.
> > Fortunately this can also be tested relatively easily by probing, waiting awhile without moving in the x-y, and then probing again to check for bad first value. Maybe someone on this thread can test and report their findings, as it doesn't require any new code.
> 
> If you want to do further testing, I have a version of this in an extension to Klipper called `z_v2settling_probe` that @voidtrance and I worked on. My GitHub repo is at https://github.com/GadgetAngel/voron2_settling_probe
> 
> @voidtrance has his version called `settling_probe` at https://github.com/voidtrance/voron-klipper-extensions
> 
> The difference between `settling_probe` and `z_v2settling_probe` is that `z_v2settling_probe` is a tool to help test out this issue.
> 
> `z_v2settling_probe` allows a command line parameter to be added to the following commands:
> 
> ```shell
> PROBE
> PROBE_ACCURACY
> PROBE_CALIBRATE
> QUAD_GANTRY_LEVEL
> BED_MESH_CALIBRATE
> PROBE_Z_ACCURACY
> CALIBRATE_Z
> ```
> 
> The command line parameter the user can add to the above command enables the user to discard the first settling sample on the fly instead of having to `RESTART` or `FIRMWARE_RESTART` their 3D printer (which changes the state of the machine).
> 
> The command line parameter is called `SETTLING_PROBE` so the commands would look as follows with `z_v2settling_probe` Klipper extension:
> 
> ```shell
> PROBE
> PROBE SETTLING_SAMPLE=0
> PROBE SETTLING_SAMPLE=1
> 
> PROBE_ACCURACY
> PROBE_ACCURACY SETTLING_SAMPLE=0
> PROBE_ACCURACY SETTLING_SAMPLE=1
> 
> PROBE_CALIBRATE
> PROBE_CALIBRATE SETTLING_SAMPLE=0
> PROBE_CALIBRATE SETTLING_SAMPLE=1
> 
> QUAD_GANTRY_LEVEL
> QUAD_GANTRY_LEVEL SETTLING_SAMPLE=0
> QUAD_GANTRY_LEVEL SETTLING_SAMPLE=1
> 
> BED_MESH_CALIBRATE
> BED_MESH_CALIBRATE SETTLING_SAMPLE=0
> BED_MESH_CALIBRATE SETTLING_SAMPLE=1
> 
> PROBE_Z_ACCURACY
> PROBE_Z_ACCURACY SETTLING_SAMPLE=0
> PROBE_Z_ACCURACY SETTLING_SAMPLE=1
> 
> CALIBRATE_Z
> CALIBRATE_Z SETTLING_SAMPLE=0
> CALIBRATE_Z SETTLING_SAMPLE=1
> ```
> 
> A `SETTLING_SAMPLE=0` means you **do not want** to discard the first settling sample, while `SETTLING_SAMPLE=1` means that you **do want** to discard the first settling sample.
> 
> My Ender3 Pro with a BLTouch definitely has an issue but my Ender3 is down at the moment. When I do have it back up I will be adding the Klipper extension `z_v2settling_probe` to that machine.
> 
> If someone else here wants to try out this extension and provide data I would like to see the test data.
> 
> The `z_v2settling_probe` Klipper extension does not require you to change any of the Klipper files. The Klipper `z_v2settling_probe` extension and the Klipper `settling_probe` extension just extends the `probe object` to so that you can choose to discard the first settling probe point from each sample group.
> 
> My hope is that tool can help settle this issue! No pun intended.

Just had a look at this but it relies on z_calibration which since moving to TAP I have removed so I can't make use of it

richardjm:
Ran a few tests today. Something possibly odd is going on. The result of a GET_POSITION doesn't seem to be matching the probe position during a PROBE_ACCURACY.

Am I missing something fundamental?
image

richardjm:
More tests. As can clearly be seen the first probe after moving back to CENTER is off in comparison.

Test commands
```
GO_EOD # moves to max x,y
CENTER Z=2 # moves to center at height of Z 2
PROBE_ACCURACY
PROBE_ACCURACY
```

```
10:14    CENTER z=2
10:14    PROBE_ACCURACY PROBE_SPEED=5 SAMPLE_RETRACT_DIST=2 LIFT_SPEED=5
10:14    PROBE_ACCURACY at X:125.000 Y:125.000 Z:2.012 (samples=10 retract=2.000 speed=5.0 lift_speed=5.0)
10:14    probe at 125.000,125.000 is z=-0.711875
10:14    probe at 125.000,125.000 is z=-0.708750
10:14    probe at 125.000,125.000 is z=-0.708750
10:14    probe at 125.000,125.000 is z=-0.708125
10:14    probe at 125.000,125.000 is z=-0.708125
10:14    probe at 125.000,125.000 is z=-0.708125
10:14    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe accuracy results: maximum -0.708125, minimum -0.711875, range 0.003750, average -0.708875, median -0.708750, standard deviation 0.001038
10:15    PROBE_ACCURACY PROBE_SPEED=5 SAMPLE_RETRACT_DIST=2 LIFT_SPEED=5
10:15    PROBE_ACCURACY at X:125.000 Y:125.000 Z:1.291 (samples=10 retract=2.000 speed=5.0 lift_speed=5.0)
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.709375
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.709375
10:15    probe at 125.000,125.000 is z=-0.709375
10:15    probe at 125.000,125.000 is z=-0.709375
10:15    probe accuracy results: maximum -0.708750, minimum -0.709375, range 0.000625, average -0.709000, median -0.708750, standard deviation 0.000306
10:15    go_eod
10:15    centre z=2
10:15    Unknown command:"CENTRE"
10:15    center z=2
10:15    PROBE_ACCURACY PROBE_SPEED=5 SAMPLE_RETRACT_DIST=2 LIFT_SPEED=5
10:15    PROBE_ACCURACY at X:125.000 Y:125.000 Z:2.012 (samples=10 retract=2.000 speed=5.0 lift_speed=5.0)
10:15    probe at 125.000,125.000 is z=-0.713125
10:15    probe at 125.000,125.000 is z=-0.709375
10:15    probe at 125.000,125.000 is z=-0.709375
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.708750
10:15    probe at 125.000,125.000 is z=-0.709375
10:16    probe at 125.000,125.000 is z=-0.708750
10:16    probe at 125.000,125.000 is z=-0.708750
10:16    probe at 125.000,125.000 is z=-0.709375
10:16    probe accuracy results: maximum -0.708750, minimum -0.713125, range 0.004375, average -0.709437, median -0.709062, standard deviation 0.001264
10:16    PROBE_ACCURACY PROBE_SPEED=5 SAMPLE_RETRACT_DIST=2 LIFT_SPEED=5
10:16    PROBE_ACCURACY at X:125.000 Y:125.000 Z:1.291 (samples=10 retract=2.000 speed=5.0 lift_speed=5.0)
10:16    probe at 125.000,125.000 is z=-0.708750
10:16    probe at 125.000,125.000 is z=-0.708750
10:16    probe at 125.000,125.000 is z=-0.708750
10:16    probe at 125.000,125.000 is z=-0.709375
10:16    probe at 125.000,125.000 is z=-0.709375
10:16    probe at 125.000,125.000 is z=-0.709375
10:16    probe at 125.000,125.000 is z=-0.709375
10:16    probe at 125.000,125.000 is z=-0.709375
10:16    probe at 125.000,125.000 is z=-0.710000
10:16    probe at 125.000,125.000 is z=-0.709375
10:16    probe accuracy results: maximum -0.708750, minimum -0.710000, range 0.001250, average -0.709250, median -0.709375, standard deviation 0.000375
```

Slidr666:
Over the last 2 years i have seen this phenomenon on 3-5 Printers each starting roughly a print a day
- Inductive Probes (Omron and knock offs)
- Klicky/KlickyNG
- Tap with Untap
- Tap with Optical Sensor 5V
(- IR Sensor was not tested)

Altough this is a not so scientific answer as those are not recorded observations: those mashines went through all kind of states from old/unmaintained with bad mechanics/stiffness to freshly (re-)built: it occured always.

Everyone is fighting it by using some sort of "hacks" like higher sample number and using median.
If someone finds the root cause that would be nice, but knowing this is not a severe problem with a simple solution would make a lot of people happy. I strongly vote for a "skip first probe" option.

GarthSnyder:
> Want me to integrate your changes into my PR? Happy to cherry pick across if you want.

Sure, pick away! 🙂 

> From the results here it certainly doesn't appear to be a mechanical issue so I'm agreeing with Kevin that discarding the first seems to be hiding something subtle in Klipper.

Could you elaborate on this conclusion a bit? Your samples match my experience: it's not "the first sample" that's off so much as "the first sample after moving the toolhead in X and/or Y." But this seems to me entirely compatible with a mechanical issue of some kind. (Of course, it's also consistent with several other theories...)

mattthebaker:
> More tests. As can clearly be seen the first probe after moving back to CENTER is off in comparison.
> 
> From the results here it certainly doesn't appear to be a mechanical issue so I'm agreeing with Kevin that discarding the first seems to be hiding something subtle in Klipper.

Moving your toolhead in the X-Y causes the next Z probe to be off. This is exactly what I was describing from the Troodon data I had, and is very likely a mechanical issue. For a CoreXY it could very well be asymmetric belt tension equalizing from the Z impulse of the first probe, or just a little slop in the motion system settling after being pulled left or right.

Is your machine a floating gantry type or does the bed move?

richardjm:
> > More tests. As can clearly be seen the first probe after moving back to CENTER is off in comparison.
> > From the results here it certainly doesn't appear to be a mechanical issue so I'm agreeing with Kevin that discarding the first seems to be hiding something subtle in Klipper.
> 
> Moving your toolhead in the X-Y causes the next Z probe to be off. This is exactly what I was describing from the Troodon data I had, and is very likely a mechanical issue. For a CoreXY it could very well be asymmetric belt tension equalizing from the Z impulse of the first probe, or just a little slop in the motion system settling after being pulled left or right.
> 
> Is your machine a floating gantry type or does the bed move?

I have a Voron 2.4. So it's a floating gantry type printer. I'm wondering if there is any way to check, maybe by wobbling the print head a bit without moving.

Sineos:
This is kind of a strange issue. If it was a mechanical thing, like snapping into place after an impulse etc, then one would think that other firmwares are affected by the same.

A very superficial search did not reveal similar report with RRF or Marlin

richardjm:
This mornings investigation changed from 64 microsteps on Z to 16 - still the first value is off.
```
06:58    GO_EOD
06:58    center Z=2
06:58    PROBE_ACCURACY PROBE_SPEED=5 SAMPLE_RETRACT_DIST=2 LIFT_SPEED=5
06:58    PROBE_ACCURACY at X:125.000 Y:125.000 Z:2.012 (samples=10 retract=2.000 speed=5.0 lift_speed=5.0)
06:58    probe at 125.000,125.000 is z=-0.715000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe accuracy results: maximum -0.710000, minimum -0.715000, range 0.005000, average -0.710500, median -0.710000, standard deviation 0.001500
06:58    PROBE_ACCURACY PROBE_SPEED=5 SAMPLE_RETRACT_DIST=2 LIFT_SPEED=5
06:58    PROBE_ACCURACY at X:125.000 Y:125.000 Z:1.290 (samples=10 retract=2.000 speed=5.0 lift_speed=5.0)
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe at 125.000,125.000 is z=-0.710000
06:58    probe accuracy results: maximum -0.710000, minimum -0.710000, range 0.000000, average -0.710000, median -0.710000, standard deviation 0.000000
```

mattthebaker:
> This is kind of a strange issue. If it was a mechanical thing, like snapping into place after an impulse etc, then one would think that other firmwares are affected by the same.
> 
> A very superficial search did not reveal similar report with RRF or Marlin

Marlin and RRF operate in real time, they have less reason to apply an instant stop with basically infinite jerk, they can use at least some deceleration on stop if they want. I can't be bothered to verify because I'm not remotely interested in RRF or Marlin.

Besides, I have high accuracy data from the Troodon showing Z offset drifts by 10um just from making moves in the X-Y. There are no impulses involved because its measured with a native displacement sensor. It can't be anything other than mechanical, and was shown to be a very poor tolerance linear rail.

You can argue whether other machines showing same symptom have the same root cause or not, but its +1 mechanical issue on that machine.

voidtrance:
While I am fully in support of adding this to Klipper, especially if multiple users with different printers are experiencing issues related to this, I think that it should be done in the form of an extra probe rather than discarding samples.

In my extensions (referenced in a comment @GadgetAngel), I chose to add an extra sample as to not lower the size of the sample set. From a quick look at the current PR, it seems that not only will the samples be discarded but the user can set a discard size large enough to where the probing won't have any samples to work with.

GarthSnyder:
> ...I think that it should be done in the form of an extra probe rather than discarding samples.

I believe that's in fact what the current PR is doing. It first makes `samples_discard_first` probes that are thrown away and then takes however many samples are specified by `samples`. However, the docs could be clearer that you are getting M discards plus N samples rather than N samples of which M are discarded.

voidtrance:
> > ...I think that it should be done in the form of an extra probe rather than discarding samples.
> 
> I believe that's in fact what the current PR is doing. It first makes `samples_discard_first` probes that are thrown away and then takes however many samples are specified by `samples`. However, the docs could be clearer that you are getting M discards plus N samples rather than N samples of which M are discarded.

You are correct, sir! I misread the loop condition.

richardjm:
> > If I may suggest a couple of modifications to the current PR:
> > 
> > 1. This setting should really apply to Z homing as well as PROBE. I looked at the code briefly, but homing seems to have a rather different implementation.
> > 2. A minor nit: the current "Probe sample discarded" message creates misleading visual groupings:
> > 
> > ```
> > probe at 30.000,270.000 is z=-0.659781
> > Probe sample discarded
> > probe at 30.000,270.000 is z=-0.655406
> > probe at 30.000,270.000 is z=-0.656031
> > probe at 30.000,270.000 is z=-0.656031
> > probe at 270.000,270.000 is z=-0.664781
> > Probe sample discarded
> > probe at 270.000,270.000 is z=-0.659156
> > probe at 270.000,270.000 is z=-0.659156
> > probe at 270.000,270.000 is z=-0.658531
> > probe at 270.000,30.000 is z=-0.670406
> > Probe sample discarded
> > probe at 270.000,30.000 is z=-0.664781
> > probe at 270.000,30.000 is z=-0.661656
> > probe at 270.000,30.000 is z=-0.659781
> > probe at 270.000,30.000 is z=-0.659781
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The message follows the discarded samples, so the four probes in each "group" are actually from two different probing sets. Better, I think, to just mark the discards inline:
> > ```
> > probe at 30.000,270.000 is z=-0.659781 (discarded)
> > probe at 30.000,270.000 is z=-0.655406
> > probe at 30.000,270.000 is z=-0.656031
> > probe at 30.000,270.000 is z=-0.656031
> > probe at 270.000,270.000 is z=-0.664781 (discarded)
> > probe at 270.000,270.000 is z=-0.659156
> > probe at 270.000,270.000 is z=-0.659156
> > probe at 270.000,270.000 is z=-0.658531
> > probe at 270.000,30.000 is z=-0.670406 (discarded)
> > probe at 270.000,30.000 is z=-0.664781
> > probe at 270.000,30.000 is z=-0.661656
> > probe at 270.000,30.000 is z=-0.659781
> > probe at 270.000,30.000 is z=-0.659781
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There's a fork of @richardjm's code that fixes this [here](https://github.com/GarthSnyder/klipper/tree/probe-discard-first).
> 
> Want me to integrate your changes into my PR? Happy to cherry pick across if you want.

Just had a look and I can't cherry pick as the commit message isn't to Klipper standards

GarthSnyder:
> Just had a look and I can't cherry pick as the commit message isn't to Klipper standards

I updated the commit message and rebased onto your current clone. This is a new branch from the link above, now found [here](https://github.com/GarthSnyder/klipper/tree/probe-discard-first-revised).

docs/Config_Reference.md
```diff
@@ -1817,6 +1817,10 @@ z_offset:
 #   samples. If this tolerance is exceeded then either an error is
 #   reported or the attempt is restarted (see
 #   samples_tolerance_retries). The default is 0.100mm.
+#samples_discard_first: 0
```
voidtrance, line 1820: I am not sure what the status of this PR is but there are talks of resurrection.

Would 'samples_discard' be more appropriate?

docs/Config_Reference.md
```diff
@@ -1817,6 +1817,10 @@ z_offset:
...
+#   The number of samples to discard when starting to probe. This can
```
voidtrance, line 1821: To me this is a bit confusing as a description.

What a non-zero value would do is add N number of samples that are discarded. In other words, it's a bit confusion whether the value here is taken out of 'samples' or not.

Cheffress:
> Interesting. Though, if this was the case, I'd expect to see it only on the first probe in a QUAD_GANTRY_LEVEL (or BED_MESH_CALIBRATE or Z_TILT_ADJUST) sequence. However, I've seen the issue reported on each probe descent, even when previous probe descents occurred recently.
> 
> I agree that tests like you described would help track down root cause.
> 
> Cheers, -Kevin

First time poster here. I had a pretty stock Voron 2.4R2 and never had an issue with an inductive probe - was great every single time. I recently upgraded to umbilical, Tap and PEI bed and then started getting very regular probing/QGL errors.
If I do a probe accuracy test (100 samples) the standard deviation is 0.00811, but interestingly the first 3 probes are -1.38875, -1.32375, -1.320. All subsequent readings are between -1.31 and -1.32. Ignoring just the first reading alone, the reading are very consistent indicates that the issue isn’t mechanical.
When I turn the spring steel sheet over and probe on the metal side, the probing is consistent right from the start. Both drift slightly over time, but the std dev is very good.
My steel sheet is 0.5mm and the PEI is an additional 0.5mm.
I believe the root cause has to do with the “give” in the PEI layer on my bed, so the ability to ignore the first x readings would be invaluable.  Looking at the details of some of the other posters here, it could explain their issue also.  I have added a graph on the klipper discourse group (7084) that shows the probing results for both PEI and non-PEI.

Stroppel:
Any new Information on this thread? I also have the same Issue... So i am really interested if there is an hidden Issue. Also is this samples_discard_first functionality now in normal klipper included or do i somehow have to include that myself? Sorry, i am not that firm with git...

PR #5970, opened 2023-01-04
Printer config for Ender3 V2 with Sprite Extruder

hugosacramento:
There is a mix of config and BLtouch properties from Sprite Extruder.


Sineos:
As there are too many possible combinations, only stock printers in their original manufacturer setup are accepted as printer configs. You may post your config in the "Config" section on https://klipper.discourse.group

PR #5974, opened 2023-01-06
screws_tilt_adjust: adds `is_base` status property

pedrolamas:
Following up on my comment [here](https://github.com/Klipper3d/klipper/pull/5921#issuecomment-1373749185), this PR removes the " (base)" suffix from the screw name and instead adds a new `is_base` boolean property to indicate which screw is the base one.

Here's a sample of the status object:

image

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -390,6 +390,7 @@ object:
     and "MM" is the number of "minutes of a clock face" representing
     a partial screw turn. (E.g. "01:15" would mean to turn the screw
     one and a quarter revolutions.)
+  - `is_base`: Returns True if this is the base screw.
 
 ## servo
 
--- a/klippy/extras/screws_tilt_adjust.py
+++ b/klippy/extras/screws_tilt_adjust.py
@@ -92,8 +92,9 @@ class ScrewsTiltAdjust:
                 self.gcode.respond_info(
                     "%s : x=%.1f, y=%.1f, z=%.5f" %
                     (name + ' (base)', coord[0], coord[1], z))
-                self.results.append({'name': name + ' (base)', 'x': coord[0],
-                    'y': coord[1], 'z': z, 'sign': 'CW', 'adjust':'00:00'})
+                self.results.append({'name': name, 'x': coord[0],
+                    'y': coord[1], 'z': z, 'sign': 'CW', 'adjust':'00:00',
+                    'is_base': True})
             else:
                 # Calculate how knob must be adjusted for other positions
                 diff = z_base - z
@@ -116,7 +117,8 @@ class ScrewsTiltAdjust:
                     (name, coord[0], coord[1], z, sign, full_turns, minutes))
                 self.results.append({'name': name, 'x': coord[0], 'y': coord[1],
                     'z': z, 'sign': sign,
-                    'adjust':"%02d:%02d" % (full_turns, minutes)})
+                    'adjust':"%02d:%02d" % (full_turns, minutes),
+                    'is_base': False})
         if self.max_diff and any((d > self.max_diff) for d in screw_diff):
             self.max_diff_error = True
             raise self.gcode.error(


KevinOConnor:
Thanks.  It looks fine to me.  I'll give a few days to see if @theophile or anyone else has comments.  Otherwise I'll look to commit.

-Kevin

theophile:
That looks good to me.

Sineos:
Would be `reference` or `reference_screw` / `reference_point` be more clear?

pedrolamas:
> Would be `reference` or `reference_screw` / `reference_point` be more clear?

Main reason I used `is_base` is because the existing implementation was already suffixing it with "base", but seems the [docs](https://www.klipper3d.org/Config_Reference.html#screws_tilt_adjust) also call it base screw:

image

----

@KevinOConnor Unrelated to this point on this PR, but the result does seem to expose more data that it actually needs to: `name`, `x`, `y` are all values from the config file and not dependent on the actual `SCREWS_TILT_ADJUST`!

As far as I can see, no other status object exposes config file values (with [configfile](https://www.klipper3d.org/Status_Reference.html#configfile) being the obvious exception!)

KevinOConnor:
I don't have a preference to the content of the status information exported by screws_tilt_adjust.  If you want to prune unnecessary fields, that's fine - just be sure to do so in a separate commit and be sure to update Config_Changes.md .

-Kevin

theophile:
@pedrolamas, the result object contains the `name`, `x`, and `y` values to make it easier to match the measured adjustments with the relevant screw. For my purpose, I use this in a [franken-macro](https://klipper.discourse.group/t/bed-tramming-from-the-lcd/809) to make a character LCD GUI for tramming the bed. It parses each screw in the results dictionary and displays the adjustment direction and degree for each screw on a location on the screen relative to where on the bed the screw is located. The macro uses the X and Y values to determine where on the screen to print the result. I suppose it would be possible to omit the `x` and `y` values from the results dictionary and then the macro could take the `name` and use that to look up the x and y positions from the config file, but that's an extra step.

pedrolamas:
Thanks @theophile, I expected that to be the reasoning behind the exposure of config values in the status object, my comment was just pointing the fact that no other status object is using that approach - so all other macros and interfaces like Fluidd and Mainsail will correlate with the `configfile` objects.

Nevertheless, you don't need the `name` for the current array-based `result` object (neither should you use that as there is nothing forcing it to be unique), you could just use the array index instead!

However, if we went to the `results` dictionary approach, it would be preferable to use the `screwN` keys instead, so the status object would be something more like this:

```json
"results": {
  "screw1": {
    "z": 0.0
    "sign": "CW",
    "adjust": "00:00",
    "is_base": true
  },
  "screw2": {
    "z": 0.0
    "sign": "CW",
    "adjust": "00:00",
    "is_base": false
  },
  ...
}
```

theophile:
> Thanks @theophile, I expected that to be the reasoning behind the exposure of config values in the status object, my comment was just pointing the fact that no other status object is using that approach - so all other macros and interfaces like Fluidd and Mainsail will correlate with the `configfile` objects.
> 
> Nevertheless, you don't need the `name` for the current array-based `result` object (neither should you use that as there is nothing forcing it to be unique), you could just use the array index instead!
> 
> However, if we went to the `results` dictionary approach, it would be preferable to use the `screwN` keys instead, so the status object would be something more like this:
> 
> ```json
> "results": {
>   "screw1": {
>     "z": 0.0
>     "sign": "CW",
>     "adjust": "00:00",
>     "is_base": true
>   },
>   "screw2": {
>     "z": 0.0
>     "sign": "CW",
>     "adjust": "00:00",
>     "is_base": false
>   },
>   ...
> }
> ```

Yes, I think that would work. And I agree it's probably cleaner. As you probably gathered, my approach to the original `results` object was to simply store all the variables passed to the `respond_info()` method in a dictionary where macros can see them. It's handy to have the screw coordinates there, but not necessary.

KevinOConnor:
Okay.  Should I look to commit this PR or wait for other changes?

Cheers,
-Kevin

pedrolamas:
My personal opinion goes to convert the result object to use the dictionary approach (that is a breaking change!), and given @theophile comments above, I believe he agrees with that.

If you are ok with it @KevinOConnor, we can close this PR (no merge) and I will create a new one with the breaking change including a note on Config_Changes.md

KevinOConnor:
It's fine with me.  I don't have a strong preference on status layout.

-Kevin

pedrolamas:
Follow up PR created, so I am now closing this one.

PR #5976, opened 2023-01-08
Example config file for Lulzbot Mini 2

VerySpookee:
Config file contains pin mappings to suit Lulzbot Mini 2 using Lulzbot SingleExtruder(0.5mm) and Lulzbot flexible magnetic bed. Mini 2 uses Einsy Retro mainboard.


JamesH1978:
Good morning,

Can i first ask, if this is a config for a stock lulzbot mini 2? As in, if a person were to purchase this printer, they would get all the options you have configed for? We only accept configs that are for out of box configurations.

You also need to remove any whitespace from your config files, this is usually spaces at the end of the line. This is causing it not to pass tests.

Please can i also ask that you read https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and make sure your naming of the file conforms and that you add the test regression changes.

Then you also need to add a signed off line as per https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

Thanks
James

VerySpookee:
Hi James,

The only change between this config and a stock mini 2 is the z_offset amount in the probe section. I'll update the value to something more typical of the stock printer so it can be a true out-of-the-box config.

Thanks for providing a link to those resources, I'm still very new at using github so I'll review the documents and make the required changes once I learn how to drive github properly.

VerySpookee:
Hi James,

I believe I have implemented all of your suggestions and followed the guidelines in the linked documents.
Please let me know if you spot anything still outstanding that needs to be fixed.

Thanks,
Trent

JamesH1978:
Thanks Trent, 

I can only see a signed off by line on 2 on your commits?

I will start to review your code though.

Thanks

James

config/printer-lulzbot-mini2-2018.cfg
```diff
@@ -0,0 +1,138 @@
...
+step_pin: PC0
+dir_pin: !PL0
+enable_pin: !PA7
+rotation_distance: 32
+microsteps: 16
+endstop_pin: tmc2130_stepper_x:virtual_endstop
+position_endstop: -7
+position_min: -7
+position_max: 168
+homing_positive_dir: FALSE
```
JamesH1978, line 25: This is not really needed, Klipper knows which direction to home according to the position_endstop value

config/printer-lulzbot-mini2-2018.cfg
```diff
@@ -0,0 +1,138 @@
...
+step_pin: PC1
+dir_pin: !PL1
+enable_pin: !PA6
+rotation_distance: 32
+microsteps: 16
+endstop_pin: tmc2130_stepper_y:virtual_endstop
+position_endstop: -5
+position_min: -5
+position_max: 192
+homing_positive_dir: FALSE
```
JamesH1978, line 38: Same comment here as stepper_x

config/printer-lulzbot-mini2-2018.cfg
```diff
@@ -0,0 +1,138 @@
...
+sample_retract_dist: 1.0
+samples_tolerance: 0.200
+[bed_tilt]
+points: -5, 22
+        -5, 190
+        160, 190
+        160, 22
+speed: 30
+horizontal_move_z: 5
+[gcode_macro G29]
```
JamesH1978, line 96: Please do not include macros

JamesH1978 (changes requested):
Looks good apart from my comments, I still assume this is stock and sensorless homing is how it comes out of the box?

Thanks
James

VerySpookee:
Thanks James,

Sorry I missed the signed off by line on two of the commits, I'm not yet well practised with GitHub and didn't realise it needed to be on each commit, instead of the push. Is it possible to edit the previous commits to include the line?

Sensorless homing is enabled out of the box for the X and Y axes which have ninjaflex bumpers where the Mini1 would have endstop switches.

Thanks for the tip regarding the homing direction, I didn't realise the position value sorted it.

I've removed the gcode macro but I'm assuming it is still ok to leave the [bed_tilt] section as is.

Cheers,
Trent

JamesH1978 (commented):
Initially the code looks fine, and you dont have to sign off every commit unless you dont sign off in the main PR

Thanks
James

JamesH1978 (approved):
Looks fine to me,

Thanks
James

KevinOConnor:
Thanks.  Could you make this config look more similar to the other configs by adding newlines between config blocks and by wrapping the comments to 80 columns?  (Point 7 and point 7g at https://www.klipper3d.org/Example_Configs.html )

-Kevin

VerySpookee:
Hi Kevin,

I've adjusted the formatting as suggested, please let me know if there was anything else missing or needing adjustment.

Thanks,
Trent

KevinOConnor:
Thanks.

-Kevin

PR #5978, opened 2023-01-09
Update Pressure_Advance.md

speed4andy:
I think that my change is necessary because many MANY people face wrong calculation thay only copy and paste...


dewi-ny-je:
You may want to add a reminder, but the all caps is surely not a way to go.

PR #4931, opened 2021-11-16
adxl345: Added automatic accelerometer calibration

dmbutyugin:
This adds `ACCELEROMETER_CALIBRATE` command and a lightweight calibration test,
which runs some sanity checks, detects the accelerometer orientation, required axes scaling, etc.
The results of the calibrations can be persisted in printer.cfg instead of deducing axes_map.

Signed-off-by: Dmitry Butyugin <dmbutyugin@google.com>

PR #5979, opened 2023-01-09
Support arbitrary ADXL345 orientation

pottedplant:
Allow arbitrary ADXL345 orientation to not constraint printer designs.

This is facilitated by using a matrix to transform all input (sensor) data which can be specified in the config.

E.g. EBB36 with integrated ADXL345 mounted on a CW2 extruder:
```
[adxl345]
# ...
# Transform sensor data so axis are aligned with the printers reference frame
axes_transform: .7071068, .7071068, 0, 0, 0, 1, .7071068, -.7071068, 0
```
adxl1
adxl2

Results:
shaper_calibrate_x
shaper_calibrate_y

There are a lot of online resources that help deriving the transformation values, e.g.:
https://www.andre-gaschler.com/rotationconverter/

rotmat

### DIFF
--- a/klippy/extras/adxl345.py
+++ b/klippy/extras/adxl345.py
@@ -30,6 +30,22 @@ SCALE_Z  = 0.003906 * FREEFALL_ACCEL # 1 / 256 (at 3.3V) mg/LSB
 Accel_Measurement = collections.namedtuple(
     'Accel_Measurement', ('time', 'accel_x', 'accel_y', 'accel_z'))
 
+# To find transformation values these resources can be used:
+# - https://www.andre-gaschler.com/rotationconverter/
+# - https://ninja-calc.mbedded.ninja/calculators/mathematics/geometry/3d-rotations
+# - https://danceswithcode.net/engineeringnotes/rotations_in_3d/demo3D/rotations_in_3d_tool.html
+
+# Simple (aka slow) no-deps implementation.
+# Could be replaced with `np.matmul(np.array(transform_float_list).reshape(3,3),input_data)`.
+def mat33_vec3_mul(m,v):
+    r = [0,0,0]
+
+    for i in range(3):
+        for j in range(3):
+            r[i] += v[j]*m[i*3+j]
+
+    return r
+
 # Helper class to obtain measurements
 class AccelQueryHelper:
     def __init__(self, printer, cconn):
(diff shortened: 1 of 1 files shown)


ripinteer:
Looks very interesting to me. Any tests conducted?

Sineos:
Very interesting idea. Also ref to @dmbutyugin' work here https://github.com/Klipper3d/klipper/pull/4931

Overall, I'm not sure that this is something the regular user can deal with successfully:
* Requires good understanding for math and trigonometry even if you use an online calculator to do the transformation matrix
* Requires accurate and correct measuring of the current orientation 
* How does the user validate and verify that his input was correct?

Maybe I do not fully understand the concept but I fear that it might be very easy to introduce errors that go unnoticed and users wondering why input shaping is crap.

pottedplant:
@Sineos

Thanks for the feedback!

> Also ref to @dmbutyugin' work here https://github.com/Klipper3d/klipper/pull/4931

Uh, nice!

> Overall, I'm not sure that this is something the regular user can deal with successfully:
> ...

True, probably not something regular users will deal with. I imagine this values will be provided by mount designers / board manufactures. I see them like other settings like oscillator frequency, pin names/mappings, .. .

pottedplant:
> Requires accurate and correct measuring of the current orientation

If you design the mount / have the CAD you know / can easily measure these values.

> How does the user validate and verify that his input was correct?

For most orientations mistakes will be quite obvious in the frequency response graphs.

pottedplant:
After looking into https://github.com/Klipper3d/klipper/pull/4931 I realised this PR provides a subset of the functionality of that PR. I'm leaving this in draft for now (if no one objects) until either #4931 is merged (then this can be closed) or #4931 is abandoned in which case I think this should be continued.

dmbutyugin:
Hi, thanks! Per se, I do not mind if it is integrated even in a limited form (that is, just the support of transformation rather than calibration and offsets). However, I would prefer to understand:

1.  Why is it useful for a regular user? Note that IS calibration currently does not care about the specific orientation of the sensor (and I got a confirmation of it in during the testing of my PR).

2. Implementation should likely do a manual unroll of the matrix multiplication for performance reasons (processing of accelerometer data is already a bottleneck on low-end hosts).

FWIW, if there large demand for this feature, I could extract just that part from my original PR and submit just the implementation as well. So, before jumping to (2) I would like to better understand (1).

pottedplant:
@dmbutyugin 

> Note that IS calibration currently does not care about the specific orientation of the sensor

Ha, did not realise that! Thanks! Could have checked such a simple thing. I guess I was blinded by people designing extra mounts so sensor axis are aligned while there seems to be no need for it.

So no real need for this!

ripinteer:
@dmbutyugin wow, good to know that. It just saved me a lot of hours wasted on designing stuff for no reason)

KevinOConnor:
Thanks.  FWIW, I think it would be good to merge #4931 (not for input shaper calibration, but for generic "motion analysis").  In general, I agree with deprecating the current `axes_map` and replace it with a more general matrix transform (once we also have the code to generate that transform).  Alas, I'm very far behind on PRs.

-Kevin

PR #5985, opened 2023-01-13
Update update_mks_robin.py

3dprintpt:
We are no longer using Python2.
Using the script as is, gives an error.
Removing the 2 from the script header solves the issue.

Signed-off-by: Márcio Pereira <3dprintpt@gmail.com>

### DIFF
--- a/scripts/update_mks_robin.py
+++ b/scripts/update_mks_robin.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # Script to update firmware for MKS Robin bootloader
 #
 # Copyright (C) 2020  Kevin O'Connor <kevin@koconnor.net>


JamesH1978:
Please read https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and pay particular attention to the signed off by line.

Thanks
James

3dprintpt:
> Please read https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and pay particular attention to the signed off by line.
> 
> Thanks James

is it ok now?

KevinOConnor:
Yes, thanks.

-Kevin

3dprintpt:
Awesome.
Thanks

LordPeq (commented):
After making this change, the script does not work with STM32 clones like the Kingroon KP3 V1.1  You still have to manually rename the firmware as it stands.  The error message says that there is no such file or directory 'python'

LordPeq:
> still have to manually rename the firmware as it stands. The error message says that there is no such file or directory 'python'

Also, changing the script to specify 'python3' as it is listed on the drive, also does not work, it is LIKELY my mistake but the logic isn't adding up.

Sineos:
This is absolutely the wrong place to discuss / ask about the issue you are having with your printer. Revert to https://klipper.discourse.group

PR #5986, opened 2023-01-13
config: added stock location of bed_screws for neptune2(s)(d)

dhruvinsh:
This config change adds stock config of the screw location. It can be found [here](https://github.com/NARUTOfzr/ZNP-Robin-Nano-V1.2-V1.3/blob/master/Configuration%20files/Neptune%202/elegoo.txt#L34).

### DIFF
--- a/config/printer-elegoo-neptune2-2021.cfg
+++ b/config/printer-elegoo-neptune2-2021.cfg
@@ -103,10 +103,16 @@ serial: /dev/ttyUSB0
 restart_method: command
 
 [bed_screws]
-screw1: 32.5, 32.5
-screw2: 32.5, 202.5
-screw3: 202.5, 32.5
-screw4: 202.5, 202.5
+screw1: 50, 50
+screw1_name: left_front
+screw2: 185, 50
+screw2_name: right_front
+screw3: 185, 185
+screw3_name: right_rear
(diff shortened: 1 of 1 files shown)


JamesH1978:
Hi,

Please read https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and pay particular attention to the signed off by line.

Thanks
James

dhruvinsh:
Thanks @JamesH1978  for notifying it. I have force pushed the commit and updated the commit message.

KevinOConnor:
Thanks.  The change looks fine to me, but does this printer really have 5 screws under the bed?

@just-trey - do you have any comments?

-Kevin

dhruvinsh:
Thank you for looking into this contribution.

No the printer does not have 5 screws. It has 4 physical screws. 5th is to provide a kind of backward compatibility of stock firmware. It serves a purpose to validate the other four screws since its location is set to the middle of the bed. If it does not seem fine then calibration can be started over and in some cases (like minute concave bed) it's very helpful.

KevinOConnor:
Okay - seems a bit unusual to describe a fifth location, but I don't really have a particular preference.  (I don't own this printer.)

-Kevin

PR #5988, opened 2023-01-15
stm32: Fix CAN2 handling in can.c

KevinOConnor:
Using the CAN2 interface still requires that the CAN1 be enabled, and the filtering hardware is always on the CAN1 hardware block.

-Kevin

### DIFF
--- a/src/stm32/can.c
+++ b/src/stm32/can.c
@@ -49,6 +49,7 @@
 
 #if CONFIG_MACH_STM32F0
  #define SOC_CAN CAN
+ #define FILTER_CAN CAN
  #define CAN_RX0_IRQn  CEC_CAN_IRQn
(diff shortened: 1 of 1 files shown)


bondus:
That's a very unexpected behaviour. I never checked the details for the multi can bus MCUs when it was originally implemented.

KevinOConnor:
Yeah, the CAN2 support on stm32 is quite goofy.  Thankfully it's not too difficult to get it to work.

I merged this branch.

Cheers,
-Kevin

coyothG:
> Using the CAN2 interface still requires that the CAN1 be enabled, and the filtering hardware is always on the CAN1 hardware block.
> 
> -Kevin
e como faz essa modificação no arquivo?

PR #5990, opened 2023-01-16
Updated the Creality Sermoon V1 Pro configuration file:

gcoppex:
- Added a note about the SD card filesystem required for flashing
 - Added the configuration for a CR-Touch bed levelling sensor and a link to the thingiverse repository for the part
 - Added the support for the case LED
 - Improved the comments and the file revision

### DIFF
--- a/config/printer-creality-sermoonV1-2022.cfg
+++ b/config/printer-creality-sermoonV1-2022.cfg
@@ -1,6 +1,12 @@
 # This file contains pin mappings for the Creality Sermoon V1
 # with CR-FDM-v2.4.S1.200 motherboard.
 
+# Version: 1.2.0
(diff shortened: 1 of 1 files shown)


JamesH1978:
Good evening,

Thankyou for submitting the changes,

Please can you follow the guidelines for pull requests set out in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially the signed off by line

I am not sure linking to a model on thingiverse is something we would add into a config, due to the fact that link may change or become not available.

Does this printer come with a CR-Touch as standard equipment? If not, this is also not something we would include in a printer config.

Please see https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Thanks
James

PR #5992, opened 2023-01-18
screws_tilt_adjust: status result as dictionary

pedrolamas:
Changes the format of the `printer.screws_tilt_adjust` status to a dictionary of screws with the resulting measurements.

image

Remaining values can easily be retrieved from active printer configuration.

This follows up on https://github.com/Klipper3d/klipper/pull/5921#issuecomment-1373749185 and replaces the PR #5974

### DIFF
--- a/klippy/extras/screws_tilt_adjust.py
+++ b/klippy/extras/screws_tilt_adjust.py
@@ -64,7 +64,7 @@ class ScrewsTiltAdjust:
             'results': self.results}
 
     def probe_finalize(self, offsets, positions):
-        self.results = []
+        self.results = {}
         self.max_diff_error = False
         # Factors used for CW-M3, CCW-M3, CW-M4, CCW-M4, CW-M5 and CCW-M5
         threads_factor = {0: 0.5, 1: 0.5, 2: 0.7, 3: 0.7, 4: 0.8, 5: 0.8}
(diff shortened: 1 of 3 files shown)


klippy/extras/screws_tilt_adjust.py
```diff
@@ -92,8 +92,7 @@ def probe_finalize(self, offsets, positions):
                 self.gcode.respond_info(
                     "%s : x=%.1f, y=%.1f, z=%.5f" %
                     (name + ' (base)', coord[0], coord[1], z))
-                self.results.append({'name': name + ' (base)', 'x': coord[0],
-                    'y': coord[1], 'z': z, 'sign': 'CW', 'adjust':'00:00'})
+                self.results["screw%d" % (i + 1,)] = {'z': z, 'is_base': True}
```
KevinOConnor, line 95: Unless I'm missing something, this drops the 'sign' and 'adjust' items from the dict.  Having dicts with entries that may or may not be present is a pain to code to.

-Kevin

pedrolamas: Indeed, I removed those from the item with `is_base=true` as they are pointless, but happy to add those back with some default value on them!

pedrolamas: @KevinOConnor I've now added those back (so all dict entries have the same keys) and updated the image on the initial post of this PR to reflect the change.

KevinOConnor:
Thanks.

-Kevin

PR #5994, opened 2023-01-18
Note that "TEST_MACRO25; comment" fails with "Malformed command"

naikymen:
Note an undocumented issue with comments in macro names.

To reproduce, create a macro named "TEST_MACRO25" with any content. Run it bare, and then with any comment. The commented version will fail.

Signed off by Nico

### DIFF
--- a/docs/Command_Templates.md
+++ b/docs/Command_Templates.md
@@ -11,6 +11,10 @@ lower case. If any numbers are used in the macro name then they must
 all be at the end of the name (eg, TEST_MACRO25 is valid, but
 MACRO25_TEST3 is not).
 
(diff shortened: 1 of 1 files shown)


naikymen:
Actually, this seems to happen with many more name patterns.

"T0; some comment" seems to be the exception. Perhaps because it is recognized as a "traditional" GCODE.

JamesH1978:
you need a space
image
otherwise klipper reads it as part of the name

naikymen:
Thanks James! That will solve my issue, I'll close this as I'm not sure what to add to the docs.

Perhaps a heading on the G-Codes page?

PR #5995, opened 2023-01-18
Add support for thermal protection without temperature delta

guestisp:
Sometimes using the temperature delta (target temp - current temp) as increment for the error counter, trigger the thermal protection too fast, particurally with low-power heater or a not perfectly aligned fan duct.

This PR add a setting `use_temperature_delta` that disable the use of delta, thermal protection is then triggered when the temperature is lower than `target - hysteresis` for at least `max_errors` seconds. This is similiar to Marlin's thermal runaway protection.

Tested properly with
```
[verify_heater extruder]
  hysteresis: 10
  max_error: 40
```

unplugging the thermistor cable starts the counter and after 40 seconds the heater is killed.

### DIFF
--- a/klippy/extras/verify_heater.py
+++ b/klippy/extras/verify_heater.py
@@ -31,7 +31,8 @@ class HeaterCheck:
         self.last_target = self.goal_temp = self.error = 0.
         self.goal_systime = self.printer.get_reactor().NEVER
         self.check_timer = None
-        self.use_temperature_delta = config.getboolean('use_temperature_delta', True)
+        self.use_temperature_delta = config.getboolean('use_temperature_delta',
+                                                        True)
     def handle_connect(self):
         if self.printer.get_start_args().get('debugoutput') is not None:
             # Disable verify_heater if outputting to a debug file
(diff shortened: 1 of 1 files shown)


Sineos:
More or less the same behavior (i.e. tune for low power heaters etc) can already be established with the current [max_error](https://www.klipper3d.org/Config_Reference.html#verify_heater) setting, except that is does not take time but accumulating the deviation. 
If I understand this PR correctly, it would completely redefine the today's meaning of `max_error`

guestisp:
Yes, if you enable the flag, max_error will be used as cumulative time.
I know that you can have something similiar by tuinign the `max_error` config but as you are using the temperature delta to increase the error counter, you can't know how much second are you allowing. In example, I had a huge temp drop (15°) with a 40W heater (slow to heat up) and every second "15" are summed to the error counter. reaching the default max error (120) took less than 10 seconds, killing a 48hours print. With this PR that works in a Marlin similiar way, the print would have been successful with almost zero security risks

Using `max_errors` as workaround lead to 3 things:
1. you have to use big values for `max_errors` thus delaying too much the protection threashold (it's risky)
2. you can't know how many seconds you are allowing before triggering the protection
3. if you use small values to avoid the problem number 1, you are triggering the protection too fast

guestisp:
Any chance to have this merged ?

KevinOConnor:
Thanks, but I don't think this is a good fit for the main Klipper branch.  This code adds complexity to the verify_heater code, and it's not clear to me that additional complexity is necessary as one can already tune that module using the existing configuration settings.

-Kevin

guestisp:
well, no, you can't have this working mode by tuning settings as the error counter doesn't add a fixed amount, thus you can't cut heating after a defined amount of seconds

Issue #4712, opened 2021-09-21
ADXL345 on Raspberry Pi Pico stopped after specific commit


glabifrons:
Using an ADXL345 on a Raspberry Pi Pico via USB worked up until (and including) commit e03e0e9dac5694b35b81726bd2ed18f1a67b0a26
This combination fails specifically on commit 8cfaed441d1ec295c8c4be8e0733099fe3f3395a and all later commits that I tried (including the one from 9/16/2021 and the ones from 9/4/2021).

I triggered the failure by sending "ACCELEROMETER_QUERY".
The error in the klippy.log is "Invalid adxl345 id (got f2 vs e5)".

For each test, I rebuilt the firmware for and reflashed both the primary MCU and the Pi Pico, then rebooted the host Raspberry Pi to ensure everything was on the exact same version of firmware.

Please find attached klippy log.
[klippy.log](https://github.com/KevinOConnor/klipper/files/7200481/klippy.log)

If needed, I can revert to the functional commit and upload the working klippy.log file as well.
I can also repeat the same sequence of tests on a different printer (I have several running Klipper, all different models and styles).
Please let me know if you need any more information from me.

KevinOConnor:
That's unfortunate and I don't know why that would be.  If you're working to find out why you can use this issue to relay your status.  If you're looking for someone else to help with this then you should reach out to the Klipper community on either Klipper Discord or Klipper Discourse.

-Kevin

glabifrons:
I finally have it figured out.

Prior to commit 8cfaed441d1ec295c8c4be8e0733099fe3f3395a you could use ACCELEROMETER_QUERY by itself and it would find the ADXL345 chip (per https://github.com/KevinOConnor/klipper/blob/master/docs/Measuring_Resonances.md ).
After commit 8cfaed441d1ec295c8c4be8e0733099fe3f3395a you **must** add "CHIP=default" for ACCELEROMETER_QUERY to not fail.
After commit 404b64fd627e04f68f64db4fef605fae1a18a747 the parameter is changed to "CHIP=adxl345", but _appears to no longer be mandatory_.


However, even in current versions ACCELEROMETER_QUERY gives an error the **first time** you execute it after a firmware restart, but succeeds every time after (verified on multiple printers with completely different hardware, but the same versions of Klipper).

```
Send: ACCELEROMETER_QUERY
Recv: !! No adxl345 measurements found
Send: ACCELEROMETER_QUERY
[...]

PR #5996, opened 2023-01-18
Document how to connect ADXL345 to Pico then on to Pi.

yegct:
This is a documentation-only change, no code is modified.

This pull request describes how to hook up an ADXL345 accelerometer to a Raspberry Pi Pico and then on to the main Raspberry Pi.

The documentation already describes [how to wire up an MPU-6000/9000 with Pico](https://www.klipper3d.org/Measuring_Resonances.html#configure-mpu-60009000-series-with-pico), so I figured it would be useful to describe this with the ADXL345. I expect this is the more commonly used accelerometer. Additionally, it's helpful to have your accelerometer connect to the main Raspberry Pi via USB instead of GPIO as it's a lot more convenient to move between klipper setups.

This is based on documentation from [this discourse post](https://klipper.discourse.group/t/raspberry-pi-pico-adxl345-portable-resonance-measurement/1757), with my thanks.

Signed-off-by: Christopher Thompson <chris@hypocrite.org>


KevinOConnor:
Thanks.  I'd like to see if there are comments from @dingyifei , @dmbutyugin , @Sineos on this change.

-Kevin

thijstriemstra (approved):
Had to follow some youtube clips (nero3d) to get this working yesterday, would've been nice to see documentation like this. Looks good to me!

The image for `make menuconfig` could be updated for latest klipper though, it now also includes a bootloader option (that does not have to be changed though).

yegct:
> The image for `make menuconfig` could be updated for latest klipper though, it now also includes a bootloader option (that does not have to be changed though) that is not present in this screenshot.

Very happy to update the screenshot. I should be able to do so tomorrow! Thanks for the review.

docs/Measuring_Resonances.md
```diff
@@ -60,6 +58,32 @@ Fritzing wiring diagrams for some of the ADXL345 boards:
...
+###### Via Raspberry Pi Pico
+
+You may connect the ADXL345 to your Raspberry Pi Pico and then connect the
+Pico to your Raspberry Pi via USB. This makes it easy to reuse the
+accelerometer on other Klipper devices, as you can connect via USB instead
+of GPIO.
+
+Please be particularly careful with your ADXL345. You want one without a
+level shifter and without being forced into I2C mode. For more details,
+see [this discourse post](https://klipper.discourse.group/t/raspberry-pi-pico-adxl345-portable-resonance-measurement/1757).
```
Sineos, line 70: I think it should be reworded, as it is OK to use an ADXL with level shifter as long as you do not connect it to VCC 5V.

Maybe something like:

> In order to avoid damage to your RPi make sure to connect the ADXL345 to 3.3V only. Depending on the board's layout, a level shifter may be present, which makes 5V dangerous for your RPi.

docs/Measuring_Resonances.md
```diff
@@ -60,6 +58,32 @@ Fritzing wiring diagrams for some of the ADXL345 boards:
...
+
+Note that we do not hook up the INT1 or INT2 pins on the ADXL345.
```
Sineos, line 72: Needed?

docs/Measuring_Resonances.md
```diff
@@ -164,6 +188,66 @@ probe_points:
...
+Now, while holding down the `BOOTSEL` button on the Pico, connect the Pico to
+the Raspberry Pi via USB. A new block device should appear. If you have no
+other block devices, this will be `/dev/sda`. You should now copy the klipper
+firmware to the pico.
+
+```
+sudo mount /dev/sda1 /mnt
+sudo cp out/klipper.uf2 /mnt
+sudo umount /mnt
+```
```
Sineos, line 216: Meanwhile outdated. See https://github.com/Klipper3d/klipper/commit/7c0559c6e62506af73d0e8f22733615705664dd5

Sineos (changes requested):
The connection diagram is quite hard to read. I think a bit higher resolution would be better. See https://canada1.discourse-cdn.com/free1/uploads/klipper/original/1X/9c8eaa478031620380eda99f012b4046ddc96e2f.png (this is the original from discourse)

docs/Measuring_Resonances.md
thijstriemstra: it doesn't hurt or confuse me as a user to have it in there.

yegct:
Thanks for the feedback, much appreciated! Converting this PR to draft until I get a chance to make the updates.

yegct:
> The connection diagram is quite hard to read. I think a bit higher resolution would be better. See https://canada1.discourse-cdn.com/free1/uploads/klipper/original/1X/9c8eaa478031620380eda99f012b4046ddc96e2f.png (this is the original from discourse)

When I push my changes, it will have the higher resolution image.

yegct:
@Sineos , I believe I have addressed all of your concerns in your first review pass of this PR. Back to you!

thijstriemstra (approved):

dingyifei:
I think the section `Flash the Pico Firmware` document fits better under the `Device Specific Documents`, just like the `RPi microcontroller` document. It is also worth mentioning the SPI performance may degrade if the rp2040 is on a controller board in charge of more than the accelerometer. I remember someone posting their logic analyzer result showing fewer reads and different frequency results when they used an STM32 F4 on a controller board.

docs/Measuring_Resonances.md
```diff
@@ -164,6 +187,62 @@ probe_points:
...
+
+```
+cd ~/klipper
+make clean
+make menuconfig
+```
+![Pico menuconfig](img/klipper_pico_menuconfig.png)
+
+Now, while holding down the `BOOTSEL` button on the Pico, connect the Pico to
+the Raspberry Pi via USB. Compile and flash the firmware.
```
Sineos, line 204: I'm not sure, maybe @KevinOConnor can comment but I believe a 
`sudo apt-install pkg-config`
is needed

docs/Measuring_Resonances.md
```diff
@@ -228,7 +307,9 @@ Recv: // adxl345 values (x, y, z): 470.719200, 941.438400, 9728.196800
 ```
 
 If you get an error like `Invalid adxl345 id (got xx vs e5)`, where `xx`
-is some other ID, it is indicative of the connection problem with ADXL345,
+is some other ID, immediately try again. There's an issue with SPI
```
Sineos, line 310: AFAIK the first measurement with a Pico will always fail. Maybe it should be in the section of the Pico itself

Sineos (commented):
Thanks for working on it and sorry for the late reply.

Regarding @dingyifei comment:
Both locations have their pros and cons. I have no feelings towards either.

yegct:
Thanks for the feedback. I'll make the updates when I have a moment, sometime in the next few days.

docs/Measuring_Resonances.md
joaobarros: Indeed but it wasn't always the case:  https://github.com/Klipper3d/klipper/issues/4712

docs/Measuring_Resonances.md
```diff
@@ -60,6 +58,31 @@ Fritzing wiring diagrams for some of the ADXL345 boards:
 
 ![ADXL345-Rpi](img/adxl345-fritzing.png)
 
+###### Via Raspberry Pi Pico
```
thijstriemstra, line 61: Instead of `Via` -> `Using` (or leave it out all together)

docs/Measuring_Resonances.md
yegct: I believe it's a bug as pointed out by joaobarros. I don't believe it's Pico-specific. For that reason, I don't believe it should be in the Pico section. Strong opinion, loosely held.

yegct:
> Thanks for working on it and sorry for the late reply.
> 
> Regarding @dingyifei comment: Both locations have their pros and cons. I have no feelings towards either.

@dingyifei, I added a bit about only running the accelerometer. After careful consideration, I believe the content belongs here rather than in a device-specific document. This is because this document already contains similar information (e.g. other accelerometers such as the MPU-9250, e.g. installing software packages, e.g. hooking up an accelerometer directly to a raspberry pi, instead of via a pi pico).

yegct:
@thijstriemstra, @Sineos , others: I believe I have made all requested changes or justified why the changes should not be made. Back to you folks for another pass.

(Also brought back up to date with the main branch)

docs/Measuring_Resonances.md
Sineos: Well, if a bug or not: I only see this behavior with the Pico. Not on a RPi 3 and also not on a Orange Pi 3 LTS

docs/Measuring_Resonances.md
```diff
@@ -164,6 +191,62 @@ probe_points:
...
+[adxl345]
+cs_pin: adxl:gpio1
+spi_bus: spi0a
+axes_map: x,z,y
+
+[resonance_tester]
+accel_chip: adxl345
+probe_points:
+    # Somewhere slightly above the middle of your print bed
+    147,154, 20
```
Sineos, line 238: There is one thing I forgot to mention (also on the thread in discourse)
Tests have shown that the 3V3 line on the Pico is very noisy, which leads to higher noise values on the ADXL measurements as well. This behavior can be improved by setting

```
[output_pin power_mode]
pin: adxl:gpio23
```

This will force the Pico's power source to go into high-performance mode rather than power saving mode.

docs/Measuring_Resonances.md
Sineos: Fine for me. I always use vanilla Linux distributions and I had to install it manually.

Sineos (approved):
LGTM 👍 

If you want you can work in the comment about the Pico's power mode. It makes a difference when looking at the raw data. If it makes a difference for real life measurements I'm not sure.

docs/Measuring_Resonances.md
yegct: Fascinating. After this PR, I may take a stab at fixing the bug. I don't see how it would be Pico-specific or Pi 4 specific but maybe I'll find out.

docs/Measuring_Resonances.md
yegct: This (more or less) matches the configuration described [elsewhere in the document](https://www.klipper3d.org/Measuring_Resonances.html#configure-mpu-60009000-series-with-pico). I will go and test to see if the above works.

Sineos: True, it was added by an excellent documentation rework of @dingyifei but it applies to this as well

yegct: Tested and change pushed. Seems to at least do no harm and, as you point out, may meaningfully improve things.

KevinOConnor:
Thanks.  In general it looks fine to me.  However, I'm unsure about adding a link to the Discourse forums in the main documentation.  (Over time, forum posts tend to "ramble" as new people comment, and I think it may be confusing to a new user if they are directed to a potentially rambling forum post.)

-Kevin

docs/Measuring_Resonances.md
glabifrons: It's specifically a bug in adxl345.py. A function to initialize the chip was removed in commit https://github.com/Klipper3d/klipper/commit/8cfaed441d1ec295c8c4be8e0733099fe3f3395a
I don't pretend to understand the reasoning behind the change, but that is the change that introduced the "always fails the first time" bug.

yegct:
> Thanks. In general it looks fine to me. However, I'm unsure about adding a link to the Discourse forums in the main documentation. (Over time, forum posts tend to "ramble" as new people comment, and I think it may be confusing to a new user if they are directed to a potentially rambling forum post.)

@KevinOConnor, reasonable. I have removed the link and brought the branch back up to date with the `master` branch. If you prefer the entire paragraph be removed, let me know. It wasn't a concern for me, my ADXL345 was correct. I'm unsure if my experience would generalise, though.

docs/Measuring_Resonances.md
yegct: Thanks, glibifrons. That at least should make it easy to reproduce and examine, whether that's done by me or by someone else. :)

thijstriemstra (approved):
thanks! and thanks for sticking to it, addressing review comments.

docs/Measuring_Resonances.md
thijstriemstra: @glabifrons could you open a ticket for this? It's an annoying bug that should be fixed if possible, especially if it's simply reverting part of a commit.

glabifrons: @thijstriemstra I did, over a year ago: https://github.com/Klipper3d/klipper/issues/4712

yegct:
> 

@KevinOConnor , this got closed as inactive. I believe I've addressed all of the concerns and the ticket has two signoffs. Are you waiting for anything more from me?

KevinOConnor:
Sorry, it slipped on my side.  I committed this change now.

Thanks.
-Kevin

PR #5999, opened 2023-01-19
config: Add config for Fysetc S6 v2 with BTT TMC2226

igor-arkhipov:
Basically we needs one `JP1` jumper to close  for UART mode([Fysetc guide](https://github.com/FYSETC/FYSETC-S6/blob/main/README.md#uart-mode-config)).
But it works only for Fysetc drivers, not for BigTreeTech drivers because they have a little different [pinout](https://cln.sh/LdphTWg6).

So I found out, additional one jumper we needs for `JP6-3` & `JP6-11` ([photo](https://cln.sh/rhgv72KF)) in case TMC2226 v1.0 from BTT.

Actual UART pins([screenshot](https://cln.sh/3XP8wBKQ)), which different from TMC2208, I took from [Fysetc repo](https://github.com/FYSETC/FYSETC-S6/blob/main/hardware/V2.0/S6%20V2.0%20SCH%20.pdf).

And maybe we don't need to solder `PDN` jumper in case TMC2209 BTT ([Fysetc guide](https://raw.githubusercontent.com/FYSETC/FYSETC-S6/main/images/TMC2209.JPG) bottom right corner), can't test myself.

### DIFF
--- a/config/generic-fysetc-s6-v2.cfg
+++ b/config/generic-fysetc-s6-v2.cfg
@@ -100,7 +100,7 @@ max_z_accel: 100
 
 
 ########################################
-# TMC UART configuration
+# TMC2208 UART configuration
 ########################################
 
 #[tmc2208 stepper_x]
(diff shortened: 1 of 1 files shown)


Fisheiyy:
I dont think this is needed, the goal of klippers configs is to be templates, that you take and modify to be your own and work for your stuff, the normal template for this board is pretty empty for a reason, its not meant to be a plug and play config as its just for the board and not a full printer.

igor-arkhipov:
I saw various default configs where is either 2208 or 2209. But  in the case of  Fycetc v2 board there is a noticeable difference in working with 2208 and 2209(2226). 
Maybe someone would like to instant apply the correct settings instead of spending several hours for not-fun-debugging.

JamesH1978 (commented):
Thankyou for submitting some changes to this board config. Unfortunately I do think this would be rather confusing as an addition to this file, mainly for the fact that the 2226 uses the same config as 2209 and this could cause confusion. 

Thanks
James

igor-arkhipov:
@JamesH1978 Ok, no rush. Let's wait for testing BigTreeTech's TMC2209.

PR #5993, opened 2023-01-18
Update printer-creality-ender3-s1plus-2022.cfg

xblackbytesx:
Added [bed_screws] block with appropriate positions of the screws for the S1 Plus.

PR #6000, opened 2023-01-19
module: Printer-creality-ender3-s1plus-2022, Added [bed_screws] block…

xblackbytesx:
… with appropriate positions of the screws for the Ender 3 S1 Plus.

In order to ease the manual leveling process I've added the [bed_screws] coordinates to the Ender 3 S1 Plus example configuration. Using the precision of the stepper motors allows you to level the bed more easily but also more accurately.

Signed-off-by: Fabio Bier githubklipper@proxymail.nl

### DIFF
--- a/config/printer-creality-ender3-s1plus-2022.cfg
+++ b/config/printer-creality-ender3-s1plus-2022.cfg
@@ -129,3 +129,9 @@ runout_gcode: PAUSE
 
 [pause_resume]
 recover_velocity: 25
+
+[bed_screws]
+screw1: 27, 27
+screw2: 267, 27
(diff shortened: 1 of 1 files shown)


JamesH1978 (commented):
Looks fine to me, as long as you can confirm these are the stock positions when the nozzle is over the screw

Also as a note, if this printer has a probe as standard, then a [screws_tilt_adjust] section may be more appropriate for the manual levelling process

Thanks
James

Fisheiyy:
Why make a new PR instead of keeping the one you [already had](https://github.com/Klipper3d/klipper/pull/5993) open, open?

PR #6001, opened 2023-01-20
 Klipper host is compatible with python2 and python3 running environments. Signed-off-by: Yangyuanxin <morixinguan@vip.qq.com>

Yangyuanxin:


JamesH1978:
Please review https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md particularly the signed off by line

Thanks
James

Yangyuanxin:
module: console、klippy
Prevent operating system users from not installing python in the default/usr/bin path, which is better compatible with the use environment of python2 and python3.
Signed-off-by: Yangyuanxin <morixinguan@vip.qq.com>

PR #6003, opened 2023-01-20
Update graph_accelerometer.py

DrPacemaker:
Usage of undeclared local variable 'header' due to missing indentation

### DIFF
--- a/scripts/graph_accelerometer.py
+++ b/scripts/graph_accelerometer.py
@@ -19,9 +19,9 @@ def parse_log(logname, opts):
         for header in f:
             if not header.startswith('#'):
                 break
-        if not header.startswith('freq,psd_x,psd_y,psd_z,psd_xyz'):
+            if not header.startswith('freq,psd_x,psd_y,psd_z,psd_xyz'):
             # Raw accelerometer data
-            return np.loadtxt(logname, comments='#', delimiter=',')
+                return np.loadtxt(logname, comments='#', delimiter=',')
     # Power spectral density data or shaper calibration data
     opts.error("File %s does not contain raw accelerometer data and therefore "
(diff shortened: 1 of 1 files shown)


JamesH1978:
Please review https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md particularly the signed off by line

Thanks
James

KevinOConnor:
@dmbutyugin - any thoughts on this bug report?

-Kevin

dmbutyugin:
FWIW, the code is somewhat correct. That is, if a user passes a csv file produced by Klipper, it works, there is no mistake there. I suppose the error could arise if the user passes an empty file as an input. A more proper fix would be
```diff
 def parse_log(logname, opts):
     with open(logname) as f:
         for header in f:
-            if not header.startswith('#'):
-                break
+            if header.startswith('#'):
+                continue
-        if not header.startswith('freq,psd_x,psd_y,psd_z,psd_xyz'):
+            if not header.startswith('freq,psd_x,psd_y,psd_z,psd_xyz'):
+                break
             # Raw accelerometer data
             return np.loadtxt(logname, comments='#', delimiter=',')
     # Power spectral density data or shaper calibration data
     opts.error("File %s does not contain raw accelerometer data and therefore "
                "is not supported by graph_accelerometer.py script. Please use "
                "calibrate_shaper.py script to process it instead." % (logname,))
```

KevinOConnor:
Okay, thanks.  Maybe I'm missing something, but the alternative implementation seems to negate the `if not header.startswith('freq,psd_x,psd_y,psd_z,psd_xyz')` check.

-Kevin

dmbutyugin:
@KevinOConnor Yes, you are right. So, I sent another PR #6072 with this fix and a few other small improvements.

PR #6004, opened 2023-01-22
Added printer config for SUNLU T3

dahliamoth:
printer-sunlu-t3-2022: Initial base config for the SUNLU T3 board

Pin outs mined from Marlin source code provided by SUNLU. In the Marlin source, the BTT SKR Mini E3 V2 is referenced. Almost all pin outs were identical to the BTT board. Changes being the use of a GigaDevice chip rather than a STM and running the TMC2209s in standalone mode rather than UART. Testing was done on my own printer for multiple prints before submitting the commit. Only submitted base config required to get the printer running, extra macros and advanced config options removed.

signed-off-by: Zachary Welvaert [zwelvaert@gmail.com]

### DIFF
--- a/config/printer-sunlu-t3-2022.cfg
+++ b/config/printer-sunlu-t3-2022.cfg
@@ -1,0 +1,204 @@
+# This file contains common pin mappings for the SUNLU Terminator T3 board
+
+# To use this config, the firmware should be compiled for the
+# STM32F103 with a "28KiB bootloader" and USB communication. Also,
+# select "Enable extra low-level configuration options" and configure
+# "GPIO pins to set at micro-controller startup" to "!PA14".
+
+# The "make flash" command does not work on the SUNLU Terminator T3 board. Instead,
+# after running "make", copy the generated "out/klipper.bin" file to a
+# file named "firmware.bin" on an SD card and then restart the board with that SD card.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+# Rename the file to printer.cfg
+
+##################################################################
+# Printer
+##################################################################
+
+[mcu]
+#obtain your MCU id using ls /dev/serial/by-path/*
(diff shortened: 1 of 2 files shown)


JamesH1978:
Good morning,

Thankyou for submitting a config, please can you review https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and pay particular attention to the regression tests, these need to be added to your commit.

Can you also confirm this printer ships with a touch style probe device? Repo configs should only describe a stock printer.

Thanks
James

dahliamoth:
> pleasecan you review https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and pay particular attention to the regression tests, these need to be added to your commit.
> 

I had read that many times while editing my config for submission and I am pretty sure I have everything from there done correctly. Also do not see anything in there that specifically references "regression tests". Perhaps that is in another document or I may just be not seeing it.


> Can you also confirm this printer ships with a touch style probe device? Repo configs should only describe a stock printer.
> 

The printer does come shipped with a bltouch clone, this clone touch probe required "pin_up_touch_mode_reports_triggered: FALSE" to function properly and had reduced accuracy when attempting to use "probe_with_touch_mode"
Printer also comes shipped with a filament runout sensor, that pin is also referenced in the config submitted

JamesH1978:
It is point 4 in that document. Please other PR's for an example, such as https://github.com/Klipper3d/klipper/pull/6006

Thanks
James

dahliamoth:
> It is point 4 in that document. Please other PR's for an example, such as #6006
> 

Yeah, I definitely did miss that. That is on me. I will get it added to that list when I have some time this afternoon

dahliamoth:
Added the config to the printers.test. Unless I missed something else, that should be everything

JamesH1978 (approved):
Looks good to me.

Thanks
James

KevinOConnor:
Thanks.

-Kevin

PR #6008, opened 2023-01-23
Update printer-creality-ender3-s1plus-2022.cfg

gogusrl:
added [bed_screws] & [screws_tilt_adjust] positions for the S1 Plus.

### DIFF
--- a/config/printer-creality-ender3-s1plus-2022.cfg
+++ b/config/printer-creality-ender3-s1plus-2022.cfg
@@ -115,6 +115,25 @@ mesh_max: 283, 263
 probe_count: 4,4
 algorithm: bicubic
 
+[screws_tilt_adjust]
+screw1: 60,70
+screw1_name: front left screw
+screw2: 300,70
+screw2_name: front right screw
+screw3: 300,307
+screw3_name: rear right screw
+screw4: 60, 307
+screw4_name: rear left screw
+horizontal_move_z: 10
+speed: 100
+screw_thread: CW-M4 # Use CW for Clockwise and CCW for Counter Clockwise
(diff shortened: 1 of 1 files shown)


Fisheiyy:
Basically same thing as #6000

elf98:
Shouldn't the positions in [bed_screws] and [screws_tilt_adjust] correlate? For screw 3 and 4 the positions differ.

elf98:
Yes, but the distance between the screws should probably be constant, and they are not:

So:
1. The relative position for the leftmost corner of the builtplate is 30,40.
2. Then all positions in screws_tilt_adjust should be the the same coordinates in bed_screws + the offset for the leftmost position of the buildplate.
3. Now they differ by a couple of mm for screw 3 and 4. 

[screws_tilt_adjust]
screw1: 60,70 
screw1_name: front left screw
screw2: 300,70
screw2_name: front right screw
screw3: 300,307
screw3_name: rear right screw
screw4: 60, 307
screw4_name: rear left screw
horizontal_move_z: 10
speed: 100
screw_thread: CW-M4 # Use CW for Clockwise and CCW for Counter Clockwise

[bed_screws]
screw1: 30,30
screw2: 270,30
screw3: 270,270
screw4: 28,270

gogusrl:
Actually I did try to follow the contributing guidelines but as it's my first time using github I don't actually have a clue what I'm doing :)

I'll check again and hopefully I can turn this in a valid PR. Thanks !

elf98:
Good!

Double check that the bed screw position are 100% correct and apply the offset on screw_tilt_adjust.

PR #6009, opened 2023-01-23
atsamd: Add support for SAMC21

Wulfsta:
Quick and dirty addition of samc21, missing the `sdadc`. Whoever reviews this, please have a close look at the clocks - I don't know if they're up to Klipper's standards.

### DIFF
--- a/src/atsamd/Kconfig
+++ b/src/atsamd/Kconfig
@@ -9,7 +9,7 @@ config ATSAMD_SELECT
     select HAVE_GPIO_ADC
     select HAVE_GPIO_I2C
     select HAVE_GPIO_SPI
-    select HAVE_GPIO_HARD_PWM if MACH_SAMD21
+    select HAVE_GPIO_HARD_PWM if MACH_SAMX2
     select HAVE_GPIO_BITBANGING
     select HAVE_STRICT_TIMING
     select HAVE_CHIPID
@@ -27,6 +27,9 @@ config BOARD_DIRECTORY
 
 choice
     prompt "Processor model"
+    config MACH_SAMC21G18
+        bool "SAMC21G18 (Duet 3 Toolboard 1LC)"
+        select MACH_SAMC21
     config MACH_SAMD21G18
         bool "SAMD21G18 (Arduino Zero)"
         select MACH_SAMD21
@@ -56,8 +59,14 @@ choice
         select MACH_SAME54
 endchoice
 
+config MACH_SAMX2
+    bool
+config MACH_SAMC21
+    bool
+    select MACH_SAMX2
 config MACH_SAMD21
     bool
+    select MACH_SAMX2
 config MACH_SAMX5
     bool
 config MACH_SAMD51
@@ -71,10 +80,15 @@ config MACH_SAME54
     select MACH_SAMX5
 config HAVE_SAMD_CANBUS
     bool
-    default y if MACH_SAME51 || MACH_SAME54
+    default y if MACH_SAMC21 || MACH_SAME51 || MACH_SAME54
+config HAVE_SAMD_USB
+    bool
+    default n if MACH_SAMC21G18
+    default y
 
 config MCU
     string
+    default "samc21g18a" if MACH_SAMC21G18
     default "samd21g18a" if MACH_SAMD21G18
     default "samd21e18a" if MACH_SAMD21E18
     default "samd21e15a" if MACH_SAMD21E15
@@ -88,7 +102,7 @@ config MCU
 config FLASH_SIZE
     hex
     default 0x8000 if MACH_SAMD21E15
-    default 0x40000 if MACH_SAMD21G18 || MACH_SAMD21E18
+    default 0x40000 if MACH_SAMC21G18 || MACH_SAMD21G18 || MACH_SAMD21E18
     default 0x80000 if MACH_SAMD51G19 || MACH_SAMD51J19 || MACH_SAMD51N19 || MACH_SAME51J19
     default 0x100000 if MACH_SAMD51P20 || MACH_SAME54P20
 
@@ -103,7 +117,7 @@ config RAM_START
 config RAM_SIZE
     hex
     default 0x1000 if MACH_SAMD21E15
-    default 0x8000 if MACH_SAMD21G18 || MACH_SAMD21E18
+    default 0x8000 if MACH_SAMC21G18 ||  MACH_SAMD21G18 || MACH_SAMD21E18
     default 0x30000 if MACH_SAMD51G19 || MACH_SAMD51J19 || MACH_SAMD51N19 || MACH_SAME51J19
     default 0x40000 if MACH_SAMD51P20 || MACH_SAME54P20
 
@@ -122,7 +136,7 @@ choice
         depends on MACH_SAMD21
         bool "8KiB bootloader (Arduino Zero)"
     config SAMD_FLASH_START_4000
-        bool "16KiB bootloader (Arduino M0)"
+        bool "16KiB bootloader (Arduino M0, Duet 3 Bootloader)"
     config SAMD_FLASH_START_0000
         bool "No bootloader"
 endchoice
@@ -140,11 +154,13 @@ config FLASH_APPLICATION_ADDRESS
 choice
     prompt "Clock Reference"
     config CLOCK_REF_X32K
-        bool "32.768Khz crystal"
+        bool "32.768Khz crystal" if !MACH_SAMC21
+    config CLOCK_REF_X12M
+        bool "12Mhz crystal" if MACH_SAMC21
     config CLOCK_REF_X25M
-        bool "25Mhz crystal" if MACH_SAMX5
+        bool "25Mhz crystal" if MACH_SAMC21 || MACH_SAMX5
     config CLOCK_REF_INTERNAL
-        bool "Internal clock"
+        bool "Internal clock" if !MACH_SAMC21
 endchoice
 
 choice
@@ -162,7 +178,7 @@ endchoice
 
 config CLOCK_FREQ
     int
-    default 48000000 if MACH_SAMD21
+    default 48000000 if MACH_SAMX2
     default 150000000 if SAMD51_FREQ_150
     default 180000000 if SAMD51_FREQ_180
     default 200000000 if SAMD51_FREQ_200
@@ -177,9 +193,11 @@ choice
     prompt "Communication interface"
     config ATSAMD_USB
         bool "USB"
+        depends on HAVE_SAMD_USB
         select USBSERIAL
     config ATSAMD_SERIAL
         bool "Serial"
+        depends on !MACH_SAMC21
         select SERIAL
     config ATSAMD_MMENU_CANBUS_PA23_PA22
         bool "CAN bus (on PA23/PA22)"
@@ -189,9 +207,13 @@ choice
         bool "CAN bus (on PA25/PA24)"
         depends on HAVE_SAMD_CANBUS
         select CANSERIAL
+    config ATSAMD_MMENU_CANBUS_PB11_PB10
+        bool "CAN bus (on PB11/PB10)"
+        depends on HAVE_SAMD_CANBUS && MACH_SAMC21
+        select CANSERIAL
     config ATSAMD_MMENU_CANBUS_PB13_PB12
         bool "CAN bus (on PB13/PB12)"
-        depends on HAVE_SAMD_CANBUS
+        depends on HAVE_SAMD_CANBUS && !MACH_SAMC21
         select CANSERIAL
     config ATSAMD_MMENU_CANBUS_PB15_PB14
         bool "CAN bus (on PB15/PB14)"
@@ -199,7 +221,7 @@ choice
         select CANSERIAL
     config ATSAMD_USBCANBUS
         bool "USB to CAN bus bridge"
-        depends on HAVE_SAMD_CANBUS
+        depends on HAVE_SAMD_CANBUS && HAVE_SAMD_USB
         select USBCANBUS
 endchoice
 choice
@@ -218,6 +240,9 @@ config ATSAMD_CANBUS_PA23_PA22
 config ATSAMD_CANBUS_PA25_PA24
     bool
     default y if ATSAMD_MMENU_CANBUS_PA25_PA24
+config ATSAMD_CANBUS_PB11_PB10
+    bool
+    default y if ATSAMD_MMENU_CANBUS_PB11_PB10
 config ATSAMD_CANBUS_PB13_PB12
     bool
     default y if ATSAMD_MMENU_CANBUS_PB13_PB12 || ATSAMD_CMENU_CANBUS_PB13_PB12
--- a/src/atsamd/Makefile
+++ b/src/atsamd/Makefile
@@ -7,6 +7,7 @@ dirs-y += src/atsamd src/generic lib/fast-hash
 
 MCU := $(shell echo $(CONFIG_MCU) | tr a-z A-Z)
 
+CFLAGS-$(CONFIG_MACH_SAMC21) += -mcpu=cortex-m0plus -Ilib/samc21/samc21/include
 CFLAGS-$(CONFIG_MACH_SAMD21) += -mcpu=cortex-m0plus -Ilib/samd21/samd21a/include
 CFLAGS-$(CONFIG_MACH_SAMD51) += -Ilib/samd51/samd51a/include
 CFLAGS-$(CONFIG_MACH_SAME51) += -Ilib/same51/include
@@ -31,6 +32,8 @@ src-$(CONFIG_HAVE_GPIO_I2C) += atsamd/i2c.c
 src-$(CONFIG_HAVE_GPIO_SPI) += atsamd/spi.c
 src-$(CONFIG_HAVE_SERCOM) += atsamd/sercom.c
 src-$(CONFIG_HAVE_GPIO_HARD_PWM) += atsamd/hard_pwm.c
+src-$(CONFIG_MACH_SAMC21) += atsamd/samd51_watchdog.c
+src-$(CONFIG_MACH_SAMC21) += atsamd/samc21_clock.c atsamd/timer.c generic/timer_irq.c
 src-$(CONFIG_MACH_SAMD21) += atsamd/watchdog.c
 src-$(CONFIG_MACH_SAMD21) += atsamd/clock.c atsamd/timer.c generic/timer_irq.c
 src-$(CONFIG_MACH_SAMX5) += atsamd/samd51_watchdog.c
--- a/src/atsamd/adc.c
+++ b/src/atsamd/adc.c
@@ -12,7 +12,26 @@
 #define ADC_TEMPERATURE_PIN 0xfe
 DECL_ENUMERATION("pin", "ADC_TEMPERATURE", ADC_TEMPERATURE_PIN);
 
-#if CONFIG_MACH_SAMD21
+#if CONFIG_MACH_SAMC21
+
+#define ADC_INPUTCTRL_MUXNEG_GND 0x18
+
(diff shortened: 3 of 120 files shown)


src/atsamd/Kconfig
```diff
@@ -162,7 +173,7 @@ endchoice
 
 config CLOCK_FREQ
     int
-    default 48000000 if MACH_SAMD21
+    default 48000000 if MACH_SAMC21 || MACH_SAMD21
```
KevinOConnor, line 176: I think it would be a good idea to add a new `MACH_SAMX2` definition (similar to the existing `MACH_SAMX5`) to reduce the number of cases of `CONFIG_MACH_SAMC21 || CONFIG_MACH_SAMD21`.

src/atsamd/adc.c
```diff
@@ -12,7 +12,29 @@
...
+    GPIO('B', 0), GPIO('B', 1), GPIO('B', 2), GPIO('B', 3), GPIO('B', 8),
+    GPIO('B', 9), GPIO('B', 4), GPIO('B', 5), GPIO('B', 6), GPIO('B', 7),
+    GPIO('A', 8), GPIO('A', 9)
+};
+static const uint8_t sdadc_pins[] = {
+    /* AINN */
+    GPIO('A', 6), GPIO('B', 8), GPIO('B', 6),
+    /* AINP */
+    GPIO('A', 7), GPIO('B', 9), GPIO('B', 7)
+};
```
KevinOConnor, line 35: Dead code?

src/atsamd/adc.c
```diff
@@ -51,7 +74,10 @@ static struct gpio_adc gpio_adc_pin_to_struct(uint8_t pin)
         if (adc_pins[chan] == pin)
             break;
     }
-#if CONFIG_MACH_SAMD21
+#if CONFIG_MACH_SAMC21
+    Adc* reg = (chan < 12 ? ADC0 : ADC1);
+    chan %= 12;
+#elif CONFIG_MACH_SAMD21
```
KevinOConnor, line 80: A `%= 12` will result in a division in the underlying assembler, which will bloat the code.  Instead, pad the `adc_pins` array to 16 so that the existing code for CONFIG_MACH_SAMX5 can be reused.

src/atsamd/adc.c
```diff
@@ -69,7 +95,38 @@ adc_init(void)
...
+    ADC0->SAMPCTRL.reg = ADC_SAMPCTRL_SAMPLEN(63);
+    ADC0->CTRLA.reg = ADC_CTRLA_ENABLE;
+
+    // ADC1
+    ADC1->REFCTRL.reg = ADC_REFCTRL_REFSEL_INTVCC1;
+    ADC1->CTRLB.reg = ADC_CTRLB_PRESCALER_DIV128;
+    ADC1->SAMPCTRL.reg = ADC_SAMPCTRL_SAMPLEN(63);
+    ADC1->CTRLA.reg = ADC_CTRLA_ENABLE;
+
+#elif CONFIG_MACH_SAMD21
```
KevinOConnor, line 129: This code seems very similar to the SAMX5 code - can both the SAMX5 and SAMC21 code share the same code?

src/atsamd/fdcan.c
```diff
@@ -234,12 +234,14 @@ compute_btr(uint32_t pclock, uint32_t bitrate)
...
 can_init(void)
 {
+#if CONFIG_HAVE_SAMD_USB
     if (!CONFIG_USB) {
         // The FDCAN peripheral only seems to run if at least one
         // other peripheral is also enabled.
         enable_pclock(USB_GCLK_ID, ID_USB);
         USB->DEVICE.CTRLA.reg = USB_CTRLA_ENABLE;
     }
+#endif
```
KevinOConnor, line 244: It's only necessary to enable one other device - if ID_USB isn't defined, we can choose some other device.  (This whole thing is some weird quirk of the samd power enable code on sam5x.)

src/atsamd/samc21_clock.c
```diff
@@ -0,0 +1,109 @@
...
+void
+SystemInit(void)
+{
+    // Reset GCLK
+    GCLK->CTRLA.reg = GCLK_CTRLA_SWRST;
+    while (GCLK->SYNCBUSY.reg & GCLK_SYNCBUSY_SWRST)
+        ;
+
+    // Init clocks
+    //if (CONFIG_CLOCK_REF_X25M)
```
KevinOConnor, line 107: Dead code.

src/atsamd/Makefile
```diff
@@ -31,10 +32,9 @@ src-$(CONFIG_HAVE_GPIO_I2C) += atsamd/i2c.c
 src-$(CONFIG_HAVE_GPIO_SPI) += atsamd/spi.c
 src-$(CONFIG_HAVE_SERCOM) += atsamd/sercom.c
 src-$(CONFIG_HAVE_GPIO_HARD_PWM) += atsamd/hard_pwm.c
-src-$(CONFIG_MACH_SAMD21) += atsamd/watchdog.c
-src-$(CONFIG_MACH_SAMD21) += atsamd/clock.c atsamd/timer.c generic/timer_irq.c
-src-$(CONFIG_MACH_SAMX5) += atsamd/samd51_watchdog.c
-src-$(CONFIG_MACH_SAMX5) += atsamd/samd51_clock.c generic/armcm_timer.c
+src-$(CONFIG_MACH_SAMC21) += atsamd/watchdog.c atsamd/samc21_clock.c atsamd/timer.c generic/timer_irq.c
+src-$(CONFIG_MACH_SAMD21) += atsamd/watchdog.c atsamd/clock.c atsamd/timer.c generic/timer_irq.c
+src-$(CONFIG_MACH_SAMX5) += atsamd/watchdog.c atsamd/samd51_clock.c generic/armcm_timer.c
```
KevinOConnor, line 37: I'm not sure it is a good idea to merge watchdog.c and samd51_watchdog.c together.  Can samc21 just reuse samd51_watchdog.c?  (It's not necessary to rename the file - on many platforms the prefix is meant to be "an example of" not "exclusive to".)

KevinOConnor (commented):
Thanks.  In general it looks good to me.  I have a few minor comments below.

Separately, can you give a brief overview of what the "sdadc" is and how it is used?  I don't think it is a roadblock for this PR, but a highlevel description would help give me some context.

-Kevin

src/atsamd/adc.c
Wulfsta: Oh, whoops - thought I got rid of that for this PR.

src/atsamd/adc.c
Wulfsta: Clever, will do.

src/atsamd/adc.c
Wulfsta: It is very similar, but I broke it out like this since a section for the sdadc will need to be added.

src/atsamd/fdcan.c
Wulfsta: Do you know if this is true for the samc21? This chip does not have usb, so we’d have to enable something else.

src/atsamd/Makefile
Wulfsta: It needs a few different things than samd51, but I don’t really understand the point of separating this file. I can pretty much put it wherever, honestly.

Wulfsta:
> Separately, can you give a brief overview of what the "sdadc" is and how it is used? I don't think it is a roadblock for this PR, but a highlevel description would help give me some context.

The sdadc is a different data structure, so an additional field in `gpio_adc` will be needed to indicate if this is the sdadc or not. I also might be misunderstanding, but the [specification](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/DataSheets/SAM-C20-C21-Family-Data-Sheet-DS60001479J.pdf) seems to suggest that the input is over a pair of pins? I haven't worked with adcs before and am a bit confused about why [the circuit](https://raw.githubusercontent.com/Duet3D/Duet3-Toolboard-1LC/master/Toolboard_1LC_v1.3/Duet3_TB_1LC_Schematic_v1.3.pdf) looks like it does...

KevinOConnor:
I'm a little confused - what pins on that Duet board are using the sdadc?

-Kevin

Wulfsta:
`TEMP_0`, the intended thermistor input, uses the sdadc. The I don’t understand why there are two `VSSA_MON_n` though, unless maybe `VSSA_MON_1` is an input to the sdadc…

KevinOConnor:
Okay, thanks for the info.  FWIW, the duet3 use of the sdadc looks to me to be a "micro optimization" that requires "high complexity".  My suggestion would be to not add any sdadc code and just use the traditional adc measuring of PB8 for TEMP_0.

As background, the sigma-delta adc (sdadc) is capable of calculating the voltage difference between the two "negative" and "positive" input lines.  (As opposed to a traditional ADC which measures a voltage as a ratio between VSSA and VREF.)  The sdadc has 16 bit resolution, while the traditional adc has 12 bit resolution.  So, it seems to me that the duet3 is just trying to get a few more bits of resolution (though because the comparison is always positive, it's at best a 15 bit resolution, and it's not clear to me that the additional bits would be of much value due to noise in the system).

-Kevin

Wulfsta:
Okay, I understand it now. I'll think about how I can hack it in when I get a few minutes to clean up this other stuff; it seems like a shame to waste the hardware if it's there.

src/atsamd/adc.c
Wulfsta: Decided to keep this, it's different enough that the code looks nicer this way, and I may add the sdadc in a future PR.

Wulfsta:
@KevinOConnor Don't merge this just yet, I should do some more testing.

Wulfsta:
I think this is good, but I'm not around hardware to test it right now. I should be able to in a few days.

KevinOConnor:
What's the current status of this PR?

-Kevin

Wulfsta:
> What's the current status of this PR?
> 
> -Kevin

Untested, on vacation. I’ll test once I’m back and have a few hours.

Wulfsta:
So I've figured out how to flash boards greater than revision 1.0 from a Pi 4, but am not seeing the mcu from a waveshare hat. I didn't look too carefully at the `fdcan.c` file, but it probably needs some work done to get it correct.

src/atsamd/samc21_clock.c
```diff
@@ -0,0 +1,107 @@
...
+        uint32_t pm_port = pm_id / 32, pm_bit = 1 << (pm_id % 32);
+        (&MCLK->APBAMASK.reg)[pm_port] |= pm_bit;
+    }
+}
+
+// Return the frequency of the given peripheral clock
+uint32_t
+get_pclock_frequency(uint32_t pclk_id)
+{
+    return CLKGEN_MAIN;
```
Wulfsta, line 52: Well that's a bug.

src/atsamd/fdcan.c
Wulfsta: @KevinOConnor can you provide more details about this? I’m having issues with CAN, and am wondering where I might investigate this further for the samc21.

KevinOConnor:
> @KevinOConnor can you provide more details about this? I’m having issues with CAN, and am wondering where I might investigate this further for the samc21.

Unfortunately, I don't know why that limitation exists on the atsame chips.  I just know that when I was running Klipper in usb-to-can bridge mode everything worked and that it wouldn't work if regular canbus mode.  Ultimately I found that if I just enabled the usb peripheral clock it was enough to get canbus working in normal standalone mode.  I saw a brief blurb in the documentation about the can hw on atsame not requesting clocks (or some such) and so I just forced it to work by enabling some other clock.  My recollection is that enabling any other clock worked.

Separately, what is the state of this PR?

If there is an issue with fdcan then I think it would help if we could separate out the fdcan changes from the rest of the PR.  If atsamc is working in regular serial mode then we can get that merged.  That will make the review process for the fdcan parts a lot smaller (and thus simpler).

Cheers,
-Kevin

Wulfsta:
Ah, thank you for the info. I'm lacking 5v hardware to hook this up to UART right now, so I'm inclined to continue until I have this working with CAN, so as to not accidentally push any bugs upstream. I've been using gdb and hardcoded calls to validate various functionality so far. I can convert this to a draft if you aren't happy with it sitting open while I investigate why I can't connect over CAN.

KevinOConnor:
What was the final results of the work on the samc21?

-Kevin

Wulfsta:
@KevinOConnor Can you comment out the USB initialization in `fdcan.c`, hack `assigned_id` to a set value upon initialization, and tell me if the board you originally wrote this CAN file for transmits status messages? I’m wondering if this is a related problem, and the Rx code in particular is what breaks when it’s the first thing initialized…

Wulfsta:
@KevinOConnor I added 12MHz clocks for v1.0 of the Toolboard, dug a Jetson TX2 out of storage, and tested with those two pieces of hardware. It worked immediately, so the issue was the hardware I was testing with. You can go ahead and merge.

src/atsamd/fdcan.c
```diff
@@ -234,13 +249,18 @@ compute_btr(uint32_t pclock, uint32_t bitrate)
...
 
+#if CONFIG_MACH_SAMC21
+    MCLK->AHBMASK.reg |= MCLK_AHBMASK_CANx;
+#endif
```
KevinOConnor, line 263: Can you explain what this does?

src/atsamd/chipid.c
```diff
@@ -6,12 +6,16 @@
 
 #include "autoconf.h" // CONFIG_USB_SERIAL_NUMBER_CHIPID
 #include "generic/canserial.h" // canserial_set_uuid
-#include "generic/usb_cdc.h" // usb_fill_serial
-#include "generic/usbstd.h" // usb_string_descriptor
 #include "sched.h" // DECL_INIT
 
 #define CHIP_UID_LEN 16
 
+#if CONFIG_HAVE_SAMD_USB
```
KevinOConnor, line 13: Is this ifdef necessary?  CONFIG_USB_SERIAL_NUMBER_CHIPID should only be set if CONFIG_USB is set, so this should all get compiled out without these ifdefs..

src/atsamd/Makefile
```diff
@@ -31,10 +32,9 @@ src-$(CONFIG_HAVE_GPIO_I2C) += atsamd/i2c.c
 src-$(CONFIG_HAVE_GPIO_SPI) += atsamd/spi.c
 src-$(CONFIG_HAVE_SERCOM) += atsamd/sercom.c
 src-$(CONFIG_HAVE_GPIO_HARD_PWM) += atsamd/hard_pwm.c
-src-$(CONFIG_MACH_SAMD21) += atsamd/watchdog.c
-src-$(CONFIG_MACH_SAMD21) += atsamd/clock.c atsamd/timer.c generic/timer_irq.c
-src-$(CONFIG_MACH_SAMX5) += atsamd/samd51_watchdog.c
-src-$(CONFIG_MACH_SAMX5) += atsamd/samd51_clock.c generic/armcm_timer.c
+src-$(CONFIG_MACH_SAMC21) += atsamd/samd51_watchdog.c atsamd/samc21_clock.c atsamd/timer.c generic/timer_irq.c
+src-$(CONFIG_MACH_SAMD21) += atsamd/watchdog.c atsamd/clock.c atsamd/timer.c generic/timer_irq.c
+src-$(CONFIG_MACH_SAMX5) += atsamd/samd51_watchdog.c atsamd/samd51_clock.c generic/armcm_timer.c
```
KevinOConnor, line 37: It would be preferable not to mix whitespace changes with functional changes.  Can the code add the two CONFIG_MACH_SAMC21 lines without altering the existing lines?

src/atsamd/Kconfig
```diff
@@ -71,10 +80,16 @@ config MACH_SAME54
     select MACH_SAMX5
 config HAVE_SAMD_CANBUS
     bool
-    default y if MACH_SAME51 || MACH_SAME54
+    default y if MACH_SAMC21 || MACH_SAME51 || MACH_SAME54
+config HAVE_SAMD_USB
+    bool
+    default n if MACH_SAMC21G18
+    default y
+
```
KevinOConnor, line 88: Spurious blank line added.

src/atsamd/serial.c
```diff
@@ -10,62 +10,97 @@
...
 #include "sched.h" // DECL_INIT
 
+#if CONFIG_MACH_SAMC21
+
+#define SERCOMx SERCOM2
+#define SERCOMx_IRQn SERCOM2_IRQn
+#define SERCOMx_GCLK_ID_CORE SERCOM2_GCLK_ID_CORE
+#define ID_SERCOMx ID_SERCOM2
+
+//DECL_CONSTANT_STR("RESERVE_PINS_serial", "PA12,PA9");
```
KevinOConnor, line 20: This adds a line of dead code (and there are a couple of other lines of dead code added later).

More generally, though, I got the impression from off-line conversations that serial was not tested on samc21.  If so, I'm not sure it is a good idea to modify serial.c, if it has not been tested.  Might be better to disable CONFIG_SERIAL in Kconfig for samc21.

src/atsamd/timer.c
```diff
@@ -11,19 +11,40 @@
...
 // Return the current time (in absolute clock ticks).
 uint32_t
 timer_read_time(void)
 {
-    return TC4->COUNT32.COUNT.reg;
+#if CONFIG_MACH_SAMC21
+    TCp->COUNT32.CTRLBSET.reg |= TC_CTRLBSET_CMD(TC_CTRLBCLR_CMD_READSYNC_Val);
+    while (TCp->COUNT32.SYNCBUSY.reg & TC_SYNCBUSY_COUNT)
+        ;
+#endif
```
KevinOConnor, line 46: Can you explain what this does?  The `timer_read_time()` is a "fast path" function - overhead here will have notable impact on performance and benchmark results.  Is there a way to just read the counter value without having to perform a write (or at least without having to perform a read-modify-write and a loop)?

KevinOConnor (commented):
Thanks.  I have some review comments and questions.  See below.

-Kevin

src/atsamd/timer.c
Wulfsta: There is a note hidden in the data sheet under the 32 bit register summary for this register (section 35.7.3.13) that states this synchronization must happen prior to any read access.

src/atsamd/serial.c
Wulfsta: Agree, will disable for this chip.

src/atsamd/Makefile
Wulfsta: It can be changed back, I only did this because it seemed neater this way.

src/atsamd/fdcan.c
Wulfsta: This bus clock for can is not enabled by default on this chip (annoyingly, see section 17.8.6 of the data sheet). This turns it on. Notice that it's not the same as the one in the clock code.

src/atsamd/chipid.c
Wulfsta: Hmm, yes - it should get compiled out. I probably just added guards knowing that USB is not available on this chip. I am hesitant to remove it though, for code clarity reasons.

Wulfsta: Opted to remove.

src/atsamd/Kconfig
Wulfsta: Removed, not sure why this is still green...

KevinOConnor:
Thanks.  I merged this PR.

In regard to `timer_read_time()`, have you tried without the synchronization?  It may just be that one will get a duplicate value without the sync, and that isn't a huge issue (less of an issue than going dramatically slower).

Along those lines, it would be interesting to see the benchmark results (see docs/Benchmarks.md) for this chip.

Also, would probably be a good idea to update the regression build tests (`test/configs/`) to compile test this chip.

-Kevin

PR #6014, opened 2023-01-28
Add pinout diagram for UART/serial connection for Ender3-S1

grantland:
It wasn't super clear to me so hopefully this helps someone in the future. I also removed VCC since it shouldn't be connected when utilizing UART (correct me if I'm wrong).

Also for posterity I used the following instructions: https://github.com/FYSETC/FYSETC-SPIDER/blob/main/firmware/Klipper/Connect%20RPI%20uart.md

and modified my `[mcu]` entry in my `printer.cfg`:

```
[mcu]
serial: /dev/ttyAMA0
restart_method: command
```

### DIFF
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -7,13 +7,25 @@
 # For a direct serial connection, in "make menuconfig" select
 # "Enable extra low-level configuration options" and  Serial
 # (on USART2 PA3/PA2), which is on the 10 pin IDC cable used
-# for the LCD module as follows: 3: Tx, 4: Rx, 9: GND, 10: VCC
+# for the LCD module as follows: 3: Tx, 4: Rx, 9: GND
+#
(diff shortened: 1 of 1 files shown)


JamesH1978 (commented):
Thank you for submitting a change to this configuration.

First of all you have not cleaned up your white space so therefore this is failing build checks.

Secondly you you need to follow the guidance at https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md can at least provide a signed off by line.

Once you have performed these measures I will then review the code changes you have made.

Thanks,
James

Fisheiyy (changes requested):
This PR is not needed at all? 
1. the VCC pin should be there, it might need to be utilized by others and its that way for ALL creality printer configs
2. I dont understand the change of the cable pinout diagram, there is no need to add a mirrored version of it
3. its not a NEED that firmware.bin will only flash while in a folder, its a MIGHT, because it will work for some without the folder and others with it

PR #6018, opened 2023-01-30
Support for load cell homing/probing + hx711 adc

D4SK:
Just wanted to share my current implementation of load cell probing as there seems to be a lot of interest in the feature.
I know there is already a PR regarding this (#5623) however my implementation is different in that it uses one continuous approach instead of moving in small steps. This is quicker and also enables homing with just the load cell.

In its current state the implementation works well on my printer, however it hasn't been broadly tested, and my setup uses a very nonstandard version of klipper. So it's possible i have overlooked something when creating the PR.

**Before merging**

This is not ready to be merged yet, I'm not really sure how the adc interface should be implemented,
also there are probably some shortcuts that will need to be adressed. Some feedback on this would be great.
Also I will fix the python2 compatibility, the formatting, and split this into separate commits. I just didn't bother with this yet as there will probably be some other changes as well.

**Limitations**
* needs numpy (I think this makes these things a lot more readable, maybe it can be installed manually for now)
* doesn't support multi-mcu homing (not sure how easy this is to implement or if it is needed, fundamentally there is less room for overshoot/latency as the printhead is already under tension at the end of a homing move) 
* delta printers probably don't work

**How to test this**
* you need a printer with a hx711 load cell adc
* you need to install numpy
* you need to to use python3 (for now)
* use a different retract distance for probing/homing (i think there is a bug with manual_move when commanding a move to the current position)
* tune the direction, force_threshold, and homing/probing speeds. (my printer uses 3 load cells in parallel, so the 70000 threshold is probably wrong for other printers)

**How it works**

Generally the code acts as a probe and as an endstop. So for probing or homing a move is issued until a fixed force threshold is reached which stops the stepper (like an endstop). Then the time of contact is determined by fitting a linear function to the force and calculating the intersection with a baseline value. 
Large disturbances during probing, or e.g. a filament blob on the nozzle can affect the result. To mitigate this an automatic retry algorithm is implemented. This has some additional (undocumented) config parameters.
This pr also includes a function to calculate screw rotations from a mesh_bed calibration. Not sure if this is useful as there is a separate module for this already.

Screenshot 2023-01-27 031648

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1250,6 +1250,9 @@ This enables the use of a load cell for homing and probing (e.g. mesh leveling).
 A QUERY_LOAD_CELL_PROBE command can be used to print out measurements.
 This can be useful for setting the initial values for the gain, threshold, and 'invert'.
 Afterwards a QUERY_LOAD_CELL_PROBE_END command stops the sampling.
+All probing or homing moves should be issued with low acceleration and low square-corner-velocity
+to prevent false triggers. Use a command like this 
+"SET_VELOCITY_LIMIT ACCEL=100 SQUARE_CORNER_VELOCITY=0.1"
 
 To use the load cell as a z endstop set the following config option in the stepper_z section
 ```
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1244,6 +1244,53 @@ See also: [extended g-code commands](G-Codes.md#z_thermal_adjust).
 #   parameter.
 ```
 
+## Load cell probe
+
+This enables the use of a load cell for homing and probing (e.g. mesh leveling).
+A QUERY_LOAD_CELL_PROBE command can be used to print out measurements.
+This can be useful for setting the initial values for the gain, threshold, and 'invert'.
+Afterwards a QUERY_LOAD_CELL_PROBE_END command stops the sampling.
+
+To use the load cell as a z endstop set the following config option in the stepper_z section
+```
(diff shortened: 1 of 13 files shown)


KevinOConnor:
Thanks.  In general it seems interesting to me.  Some important notes though:
1. I think discussing this on Discourse would be worthwhile.  More people will see it there and there is a greater chance more people will test it.  You should also follow up on the existing Discourse threads and be sure to ping @mhier.
2. This implementation adds code to the low-level stepper, probe, and homing code.  This should not be necessary, as it should be possible to use the existing homing, probe, and trsync modules to define a custom homing process.

-Kevin

D4SK:
Thanks for taking a look at this.

I agree that the existing trsync code should probably be used. It just seemed more complicated. I will see if i can look into this further.

In the homing module a retry flag has been added. I'm not sure there is a cleaner way to do this that doesnt modify the homing code. The retrying could also be useful for sensorless homing procedures in the future.

The current code replaces the probe module with a load_cell_probe. Using the existing probe module could require a lot of modifications as the load cells have some special requirements:
* There should be a small wait time between probing moves to sample an accurate baseline value.
* The trigger shouldn't be active during retract or xy moves
* The calculated noise values give information about the accuracy of the sample. This could be leveraged by specific load cell probing code in the future
* Probe calibration is not needed

I added a discourse topic as well.

D4SK:
https://klipper.discourse.group/t/testing-requst-load-cell-homing-probing-hx711-adc-6018/6412

PR #6021, opened 2023-01-31
config: Add config for Kingroon KP3S

JamesH1978:
A config that describes a Kingroon KP3S with the STM32F103 Robin Mini/Nano clone

Signed-Off-By: James Hartley <james@hartleyns.com>

### DIFF
--- a/config/printer-kingroon-kp3s-2020.cfg
+++ b/config/printer-kingroon-kp3s-2020.cfg
@@ -53,28 +53,29 @@ home_xy_position: 90,90
 z_hop: 10
 
 [thermistor Kingroon_B3950]
-temperature1 = 25.0
-resistance1 = 103180.0
-temperature2 = 150.0
-resistance2 = 1366.2
-temperature3 = 250.0
-resistance3 = 168.6
+temperature1: 25.0
+resistance1: 103180.0
+temperature2: 150.0
+resistance2: 1366.2
+temperature3: 250.0
+resistance3: 168.6
 
 [extruder]
 step_pin: PD6
 dir_pin: !PD3
 enable_pin: !PB3
 microsteps: 32
-rotation_distance: 8.101
+gear_ratio: 3:1
+rotation_distance: 23.244
 nozzle_diameter: 0.400
 filament_diameter: 1.750
 heater_pin: PC3
 sensor_type: Kingroon_B3950
 sensor_pin: PC1
-control = pid
-pid_kp = 27.057
-pid_ki = 1.171
-pid_kd = 156.254
+control: pid
+pid_kp: 27.057
+pid_ki: 1.171
+pid_kd: 156.254
 min_temp: 0
 max_temp: 250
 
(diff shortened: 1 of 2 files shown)


LanducciD:
Great to see!

JamesH1978 (commented):
Seems weird reviewing my own PR :)

looks fine to me...

at least I can't approve it myself.

So over to you @KevinOConnor 

Thanks
James

KevinOConnor:
Thanks.  Could you replace the '=' with ':' syntax.  https://www.klipper3d.org/Example_Configs.html point 7d.  Also, not required, but if you have the printer on hand it is nice to list the actual gear ratio of the extruder and use a `gear_ratio` statement (point 7e).

-Kevin

JamesH1978:
Dang it, moved them up from the save config.

Ok, added gear_ratio for the titan clone, and re did RD tuning.

Thanks
James

KevinOConnor:
Thanks.

-Kevin

tildebyte:
On my v1.3 (as marked on the main board) which I received from Ali in early February, all three axis steppers' 'dir_pin's are the reverse of what's in here.

While comparing this commit with my own config, I inadvertently switched the Z 'dir_pin', and boy did I get a nasty surprise on my next `G28`...

For reference, mine are *working properly* (and have been since I first set up Klipper) with

```
[stepper_x]
dir_pin: PE2

[stepper_y]
dir_pin: PB9

[stepper_z]
dir_pin: !PB4
```

This is not so much a "you're wrong", as it is a warning to those who come after to be *very* careful, and make sure to thoroughly test any changes to 'dir_pin' settings before trying to print (or even home, for that matter)

JamesH1978:
Thanks for the comment, this printer config was built specifically on the v1.2, which mine shipped with, at the time the v1.3 was an upgrade board. I should have probably specified that in the preamble. I will make a pr to update with that information. Or I can add a section which lists the changes on a v1.3, if you can confirm those the only changes needed?

Thanks,
James

tildebyte:
First approximation - those are the only *truly important* deltas, but I'll check to be sure

tildebyte:
The above mentioned 'dir_pin', plus

possibly important

```
[stepper_x]
full_steps_per_rotation: 200

[stepper_y]
full_steps_per_rotation: 200

[stepper_z]
full_steps_per_rotation: 200

[extruder]
full_steps_per_rotation: 200
gear_ratio: 67:22
max_temp: 300

[printer]
max_accel: 1800  # Reasonable people can disagree :D
```

new?

```
[heater_fan nozzle_fan]
# attached to heater1 port to support pwm
pin: PB0
heater: extruder
off_below:0.15
heater_temp: 50.0
fan_speed: 0.8
shutdown_speed: 1
```

I *think* I added `full_steps_per_rotation`.

I *think* the other extruder diffs come from one of the GH repos for configuring the v1.3 with Klipper

I'm now wondering if my board was simply wired backwards... 🤣

P.S. I also have a working config for the LCD beeper, which is super convenient for CANCEL, PAUSE, or if you're me and you like to know exactly when a print is about to start so you can check for cat clearance

PR #6023, opened 2023-01-31
Update printer-creality-ender2pro-2021.cfg

darxfame:
module: Update pins for serial

Signed-off-by: Artem Kochegizov <metra52@ya.ru>

### DIFF
--- a/config/printer-creality-ender2pro-2021.cfg
+++ b/config/printer-creality-ender2pro-2021.cfg
@@ -5,7 +5,7 @@
 
 # If you prefer a direct serial connection, in "make menuconfig"
 # select "Enable extra low-level configuration options" and select
-# serial (on USART3 PB11/PB10), which is broken out on the 10 pin IDC
+# serial (on USART2 PA3/PA2), which is broken out on the 10 pin IDC
 # cable used for the LCD module as follows:
 # 3: Tx, 4: Rx, 9: GND, 10: VCC
 


JamesH1978 (commented):
Thank you for submitting a change to this existing config.

Generally creality boards use UART3 on their display ports. Can you confirm this is actually not the case on this board, and you have tested this functionality selecting USART2 in the `make menuconfig` 

Unfortunately there are no official pin outs available for me to check schematics to confirm this myself.

Thanks,
James

JamesH1978:
thanks for that, I can see from the marlin sources that indeed they map PA2/PA3 to the idc port on the V4.2.3 board. but i cannot find evidence this has been changed on the S1 boards.

Thanks
James

JamesH1978 (approved):
I can confirm from your evidence and marlin source, that this is indeed the case for this board, so I will approve this change.

Thanks
James

stas2z:
> but i cannot find evidence this has been changed on the S1 boards.

https://github.com/MarlinFirmware/Marlin/blob/b7c23631c5840e2338cae531b171459e9e9c5301/Marlin/src/inc/Conditionals_adv.h#L1090

JamesH1978:
Interesting, fair enough. Didn't think to check the conditionals

It would of course have to be tested by a user of that board before a change was made to that config though, for completions sake.

Thanks
James

stas2z:
> before a change was made to that config

S1 already have USART2 mention in the config
https://github.com/Klipper3d/klipper/blob/master/config/printer-creality-ender3-s1-2021.cfg#L9

so we just need to fix it for e2p by this pr)

JamesH1978:
Great, I have approved this, and a maintainer will merge it in due course

Thanks
James

KevinOConnor:
It's fine with me, but I don't see a signed-off-by line - https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

darxfame:
> It's fine with me, but I don't see a signed-off-by line - https://www.klipper3d.org/CONTRIBUTING.html .
> 
> -Kevin

Signed

KevinOConnor:
Thanks.

-Kevin

PR #6024, opened 2023-01-31
Create printer-flying_bear_reborn_fysetc_s6.cfg

darxfame:
Klipper Configuration for FB Reborn v1 and motherboard Fysetc S6 and TMC2209 UART

### DIFF
--- a/config/printer-flying_bear_reborn_fysetc_s6.cfg
+++ b/config/printer-flying_bear_reborn_fysetc_s6.cfg
@@ -1,24 +1,26 @@
 #Configuration file in Klipper https://github.com/Klipper3d/klipper/blob/master/config/generic-fysetc-s6-v2.cfg
 # ——————————————————————————————————————————————————————————————————————————————
-#  Klipper Firmware Configuration
+# Klipper Firmware Configuration
 # ——————————————————————————————————————————————————————————————————————————————
 #[*] Enable extra low-level configuration options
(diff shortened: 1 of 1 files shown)


JamesH1978 (commented):
Thank you for submitting your example config.

Please can you make sure that you follow the instructions in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md paying particular attention to points 1,2 and 4. 

We would suggest you do not use pin aliases, and state the actual pin in the config. Using aliases will only serve to confuse, not to mention it does not follow the standard way.

Please also refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially the signed off by line.

Thanks,
James

JamesH1978:
Can you also confirm the printer you have described comes with the Fysetc S6 and 2209 drivers out of the box? We do not allow custom printers into the repo, due to confusion from other users wishing to klipper a stock printer. 

Thanks
James

darxfame:
Hello. It's not stock motherboard for this printer

JamesH1978:
Then we cannot accept this config into the repo for reasons given above and point 5 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Configs describing a custom setup can be posted on the klipper discourse. https://klipper.discourse.group/

Thanks
James

PR #6026, opened 2023-02-01
stm32 i2c speed

findus4fun:
Added the option to choose between 100 kHz and 400 kHz for STM32
in lowlevel Kconfig.
 
I plan to use an mpu6050 with a STM32F1 (Blue-Pill).

Signed-off-by: Jan Schunke <findus4fun@gmail.com>

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -337,6 +337,14 @@ config STM32F0_TRIM
         Default is 16 (use factory default). Each increment increases
         the clock rate by ~240KHz.
 
+choice
+    prompt "I2C Clock Speed" if LOW_LEVEL_OPTIONS
+    config STM32_I2C_CLOCK_100kHz
+        bool "100 kHz"
+    config STM32_I2C_Clock_400kHz
+        bool "400 kHz"
+endchoice
+
 
 ######################################################################
 # Communication inteface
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  However, we don't use Kconfig to define individual parameters like that.  The `rate` parameter in that function already has the requested data rate.  That said, we'd need confirmation that the code actually works at 400Khz as there are often other settings that are needed in order to use higher speed i2c rates.

-Kevin

findus4fun:
Hi,
I have tested the code on an STM32F1 and verified with an oscilloscope that
it works.
I do not have other devices to test it.
So what would be the next action?
Regards
Ja

KevinOConnor ***@***.***> schrieb am Sa., 4. Feb. 2023, 00:55:

KevinOConnor:
It's not clear from your comments if the code was verified to work with an i2c device.  At a minimum, the code would need to use `rate` and not a Kconfig definition, it would need to be tested and verified working with at least one i2c device, and I'm pretty sure the stm32 `I2C_TRISE` register would need to be programmed.

-Kevin

findus4fun:
Hi, I tested it with an mpu 6050 and communication was working.
I also used, like I said before, an oscilloscope to verify that the I2c
Clock was 100 kHz with a 50% Duty Cycle.
If I find time I will dig into the rate command and try to change it so
that it will comply with Klipper code.
I think you can close this pull request and I will send another once I have
understood how the rate command works.
Regards
Jan
P.S.: Thanks for all the time you spent on Klipper.

KevinOConnor ***@***.***> schrieb am Do., 9. Feb. 2023, 18:32:

Issue #2650, opened 2020-03-28
Support for 20*04 LCD (HD44780) via I2C Connection


hraley:
Hi folks!

As per the title, I have tried in multiple ways to implement the connection of an I2C LCD display to my MCU (AVR Board - Atmega1280).

_I've recently returned to 3D printing after a hiatus of nearly 4 years. I have 4 machines all running the same MCU and LCD set up (with encoder) - although they all have different mechanical set ups.... Deltas, Cartesian, Core XY...etc...etc. 
Upgrading both legacy Marlin and Repetier Firmware to Klipper hasn't been too much of a task after I blew the cobwebs off my Github skills!
Awesome firmware, a true revolution, I'm damn impressed. I can bring my printers back into current spec for the price of a Raspberry Pi 3 A+ or a Pi Zero W.
I am using DWC2 as the web interface/control instead of Octoprint, after reading how streamlined it was in a cpu consumption sense!_

To run the LCD's previously, it was dead easy, Just connect the wires up to the SDA and SCL pins on the MCU, add the LiquidCrystalLibrary to Arduino IDE. enable in config! Job done!

> (This is the library I previously used - https://bitbucket.org/fmalpartida/new-liquidcrystal/src/integration/ )

Klipper already has support for the HD44780 (2004 LCD), just in a direct connected state. Could I do something to change the communication to I2C? From trawling through Gihub and other config files, i believe there is another type of display that uses I2C communication, so... it's there!!??

In the `[display]` section, I have tried defining `lcd_type: hd44780` with `spi_bus:` and `spi_speed:`.... to no avail.

Pictures....
![YWRobot-LCM1602-I2C](https://user-images.githubusercontent.com/6191853/77823479-5d208e80-70f3-11ea-9c86-e9d325626121.png)
Pictures.....
![I2C-LCD 2004](https://user-images.githubusercontent.com/6191853/77823480-5e51bb80-70f3-11ea-96da-a32707fd47e5.jpg)

I hope someone might be able to point me in the correct direction, or at least educate me to why and how this does or doesn't work.
Hope I've provided enough info.

Kind Regards
Harry
[klippy.log](https://github.com/KevinOConnor/klipper/files/4397165/klippy.log)

KevinOConnor:
Unfortunately, Klipper does not currently have support for this.  Adding support would require an interested developer to implement and submit it.  (Unfortunately, it would likely require both host Python code and micro-controller C code.)

[...]

PR #6027, opened 2023-02-01
klippy display driver:  Support hd44780 displays via i2c bus

jrlomas:
This commit implements request #2650 (https://github.com/Klipper3d/klipper/issues/2650)

### DIFF
--- a/klippy/extras/display/display.py
+++ b/klippy/extras/display/display.py
@@ -6,7 +6,7 @@
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import logging, os, ast
-from . import hd44780, hd44780_spi, st7920, uc1701, menu
+from . import hd44780, hd44780_spi, st7920, uc1701, menu, hd44780_i2c
 
 # Normal time between each screen redraw
 REDRAW_TIME = 0.500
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  However, there are already two copies of the hd44780 code in Klipper right now.  The hd44780 code will require a refactoring as I don't feel comfortable merging a 3rd copy of the code.

If you search the github PR history, there are a couple of other PRs similar to this one that were not merged for the same reason.

Also, commits need to have a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

jrlomas:
Ok, whenever I get the chance I'll refactor the hd44780.py and hd44780_spi.py and the new driver and merge them into a single entity.  I'll look into the signed-off-by line.  Thank you for the pointer.

thijstriemstra:
> Thanks. However, there are already two copies of the hd44780 code in Klipper right now. The hd44780 code will require a refactoring as I don't feel comfortable merging a 3rd copy of the code.

Is there are a reason the display support code doesn't use any libraries like luma.oled (disclaimer: im a co-maintainer)? Wouldn't it be much easier to maintain and add new devices if a separate library would be used? Or is the goal to have all Klipper code contained in one codebase with no, or as few as possible, 3rd-party libraries?

PR #6028, opened 2023-02-02
Pull/support for ar100

eliasbakken:
SUPPORT FOR AR100 CPU

Some Allwinner SoCs have a coprocessor that
can be used to add real time functionality to
a Linux based OS. The AR100 present in
A64, H3, H5 and H6 are examples of this.
This PR adds support for the AR100 in the A64
which is in use on Recore controller board.

Resubmitting for PR #4994, speed tests have been redone

Signed-off-by: Elias Bakken [elias@iagent.no](mailto:elias@iagent.no)


src/ar100/Kconfig
```diff
@@ -0,0 +1,21 @@
+# Kconfig settings for AR100
+
+if MACH_AR100
+
+config AR100_SELECT
+    bool
+    default y
+    select HAVE_ENDSTOPS
+    select HAVE_STEPPERS
```
KevinOConnor, line 9: Dead code?

src/ar100/Makefile
```diff
@@ -0,0 +1,41 @@
...
+CFLAGS		+= -fno-builtin
+CFLAGS		+= -fno-pie
+CFLAGS		+= -ffreestanding
+CFLAGS		+= -msfimm -mshftimm -msoft-div -msoft-mul
+CFLAGS		+= -Ilib/ar100
+CFLAGS_klipper.elf := $(CFLAGS) -Wl,-T,src/ar100/ar100.ld -Wl,-Map,out/klipper.map
+CFLAGS_klipper.elf += -Wl,--no-undefined  -Wl,--gc-sections -static
+CFLAGS_klipper.elf += -Wl,--no-dynamic-linker -Wl,--allow-multiple-definition
+
+SFLAGS		 = -nostdinc -Wa,--fatal-warnings -MMD -Wmissing-prototypes -Wall -Wextra
```
KevinOConnor, line 14: The extra warning flags (`-W...`) end up causing unnecessary code churn - users end up running a variety of different versions of gcc which often don't agree on what a valid warning is.

src/ar100/Makefile
```diff
@@ -0,0 +1,41 @@
...
+SFLAGS		+= -Ilib/ar100
+
+# Add source files
+src-y += ar100/main.c ar100/gpio.c ar100/serial.c
+src-y += ar100/util.c ar100/timer.c
+src-y += generic/crc16_ccitt.c generic/timer_irq.c
+src-y += buttons.c tmcuart.c spi_software.c
+src-y += spicmds.c sensor_adxl345.c
```
KevinOConnor, line 22: This looks a bit suspect - is there a reason not to enable HAVE_GPIO_SPI and/or HAVE_GPIO_BITBANGING?

src/ar100/serial.c
```diff
@@ -0,0 +1,92 @@
...
+
+uint32_t r_uart_fifo_rcv(void){
+  return read_reg(R_UART_RFL);
+}
+
+void uart_putc(char c){
+  while(!(read_reg(UART0_LSR) & 1<<5))
+    ;
+  write_reg(UART0_THR, c);
+}
```
KevinOConnor, line 47: Dead code?  (And similar comment about the other uart_xxx functions.)

KevinOConnor (commented):
Thanks.  In general, it looks okay to me.  I have some minor comments below.

-Kevin

src/ar100/Makefile
eliasbakken: There is not enough space for the default options. I've instead switched to removing the files that are not needed for this board, is that better?

src/ar100/Makefile
eliasbakken: I've cleaned up the Makefile a little.

src/ar100/serial.c
eliasbakken: Yes, this was for debugging purposes.

eliasbakken:
Sorry, I merged this the wrong way. I'll open a new PR.

src/ar100/Makefile
KevinOConnor: Your latest PR looks fine to me, but I'm curious what "I've instead switched to removing the files that are not needed for this board" means.  What was the change?

Cheers,
-Kevin

eliasbakken: I'm now using HAVE_GPIO_SPI and HAVE_GPIO_BITBANGING via Kconfig, but since those options are also pulling in some LCD-related files etc. I'm removing unneded files for AR100  instead for of adding using src-y. I've not seen this pattern elsewhere in the code, so it is perhaps not welcome. The third alternative would be to have more fine grained control in the Kconfigs that separates LCD-related files from SPI-realated files.

PR #6033, opened 2023-02-04
extruder: store currently sync'ed motion queue name

pedrolamas:
Adds a new `motion_queue` property to `ExtruderStepper` so we can store the current motion queue name and return on `get_status()`.

image

When no motion queue is set, this returns null:

image

This will allow frontends (Fluidd, Mainsail, ...) to show the current motion queue in sync on each extruder stepper, and for easy changes!

Follows up on [this Discourse topic](https://klipper.discourse.group/t/storing-the-current-extruder-from-sync-extruder-motion/6413)

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -126,6 +126,9 @@ The following information is available for extruder_stepper objects (as well as
 [extruder](Config_Reference.md#extruder) objects):
 - `pressure_advance`: The current [pressure advance](Pressure_Advance.md) value.
 - `smooth_time`: The current pressure advance smooth time.
+- `motion_queue`: The name of the extruder that this extruder stepper is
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  Seems fine to me.  The status reference description seems a bit vague to me though.  How about something like: `motion_queue: The name of the extruder that this extruder stepper is currently synchronized to.  This is reported as None if the extruder stepper is not currently associated with an extruder.`

-Kevin

pedrolamas:
Thanks @KevinOConnor, I've now updated the docs following your comments.

KevinOConnor:
Thanks.  As before, it looks fine to me.  I'll give this a few days to see if others have comments.  Otherwise I will look to commit.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6035, opened 2023-02-05
Fixed filament sensor on printer-sunlu-t3-2022.cfg

dahliamoth:
Discovered that the filament sensor included with the printer is not a switch, it is a motion sensor.

The filament sensor on the printer used for initial testing had been removed prior to flashing klipper to it. Only a basic test to see it if triggered or not was preformed. It was discovered by someone else doing some testing on another printer that it would frequently pause. This was a simple oversight, assuming it was a runout switch.

Removed section of config related to filament runout switch and replaced with basic config for filament motion so they printer will simply pause when no more motion is detected. Also notated to add CLEAR_PAUSE to start gcode if using this sensor as it will trigger a paused state when loading or unloading filament. If printer is in a paused state at the start of the print, the attempt to pause will cause the printer to rapidly pause and unpause itself.

signed-off-by: Zachary Welvaert zwelvaert@gmail.com

### DIFF
--- a/config/printer-sunlu-t3-2022.cfg
+++ b/config/printer-sunlu-t3-2022.cfg
@@ -1,0 +1,204 @@
+# This file contains common pin mappings for the SUNLU Terminator T3 board
+
+# To use this config, the firmware should be compiled for the
+# STM32F103 with a "28KiB bootloader" and USB communication. Also,
+# select "Enable extra low-level configuration options" and configure
+# "GPIO pins to set at micro-controller startup" to "!PA14".
+
+# The "make flash" command does not work on the SUNLU Terminator T3 board. Instead,
+# after running "make", copy the generated "out/klipper.bin" file to a
+# file named "firmware.bin" on an SD card and then restart the board with that SD card.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+# Rename the file to printer.cfg
+
+##################################################################
+# Printer
+##################################################################
+
+[mcu]
+#obtain your MCU id using ls /dev/serial/by-path/*
+serial: dev/serial/by-id/usb-Klipper_stm32f103xe_833E31383534300530343833-if00
+
+[printer]
+kinematics: cartesian
+max_velocity: 300
+max_accel: 3000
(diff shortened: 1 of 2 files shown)


config/printer-sunlu-t3-2022.cfg
```diff
@@ -0,0 +1,211 @@
...
+nozzle_diameter: 0.400
+filament_diameter: 1.750
+heater_pin: PC8
+sensor_type: EPCOS 100K B57560G104F
+sensor_pin: PA0
+control: pid
+pid_Kp: 19.479 #calibrate your own PID
+pid_Ki: 1.073
+pid_Kd: 88.385
+min_extrude_temp: 175
```
JamesH1978, line 143: Please do not override safety settings, min_extruder_temp and max_extrude_only_distance should be removed.

JamesH1978 (changes requested):
Thank you for submitting your printer config.

I have made one comment so far, change this and I'll do another review and see if it's good to approve

thanks James

config/printer-sunlu-t3-2022.cfg
dahliamoth: Sorry about that. I had those in my own config for use with filament load/unload macros and forgot to remove them when i removed those macros. Removed now.

JamesH1978:
I am not sure why, but your PR has not started the check process

dahliamoth:
> I am not sure why, but your PR has not started the check process

I maybe bc to fix this issue, i edited my old fork that was already merged. I probably needed to delete this fork and suggest the edit on a new fork. I can do that if needed

JamesH1978:
I would assume that is what has happened. I would delete this and use a new fork.

Thanks
James

dahliamoth:
> I would assume that is what has happened. I would delete this and use a new fork.

I closed this PR and opened a new one on a updated fork. It is now autostarting the build test, but apparently at the moment that I submitted it there is a scheduled brownout of Ubuntu 18.04, so the test canceled itself. Hopefully when the brownout if over it will rerun its test

PR #6036, opened 2023-02-06
Add Tenlog D3-Pro IDEX printer config

kiler129:
Caveat: the LCD config is missing as the printer's LCD isn't supported by anything but custom Tenlog code embedded in Marlin v1.


JamesH1978 (commented):
Thank you for submitting your example config.

Please can you make sure you read https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and make sure your file conforms to it, including the format of the filename, adding regression checks, use of P pins and not ar pins. We have deprectaed and removed ar pins, and generally, they should not be used in stock printer configs, even with an alias section. Please do not override any safety features or extra config like pressure_advance as stated in the document. 

Also please follow the instructions in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying particular attention to the signed-off-by line.

I will do a full review once these changes have been made.

Thanks

James

PR #6037, opened 2023-02-06
stepper_enable: report status

pedrolamas:
Adds status report to the `stepper_enable` module, so one can know what steppers are currently enabled and use that information in interfaces (Fluidd, Mainsail, ... ) or macros.

image

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -395,6 +395,12 @@ The following information is available in
 - `printer["servo <config_name>"].value`: The last setting of the PWM
   pin (a value between 0.0 and 1.0) associated with the servo.
 
+## stepper_enable
+
+The following information is available in the `stepper_enable` object (this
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  I'll have to think on this a little.  I'm fine with exporting the information, but I'm unsure about rebuilding the dict with every status update.  Many clients just subscribe to everything which means we'd be rebuilding the dict with every update, even though it would almost always be static.

Separately, per convention, the status reference should add something like `(this object is available when any stepper object is defined)`.

-Kevin

pedrolamas:
Thanks Kevin, I've updated the docs entry following your indication.

KevinOConnor:
Thanks.  I merged this PR.  We can always optimize at a later date if it is needed.

-Kevin

pedrolamas:
Awesome, thank you Kevin! 🙂

PR #6040, opened 2023-02-07
Update printer-sunlu-t3-2022.cfg

dahliamoth:
Fix filament sensor and remove unneeded config changes in extruder

The filament sensor on the printer used for initial testing had been removed prior to flashing klipper to it. Only a basic test to see it if triggered or not was preformed. It was discovered by someone else doing some testing on another printer that it would frequently pause. This was a simple oversight, assuming it was a runout switch.

Removed section of config related to filament runout switch and replaced with basic config for filament motion so they printer will simply pause when no more motion is detected. Also notated to add CLEAR_PAUSE to start gcode if using this sensor as it will trigger a paused state when loading or unloading filament. If printer is in a paused state at the start of the print, the attempt to pause will cause the printer to rapidly pause and unpause itself.

Removed min_extruder_temp and max_extrude_only_distance from the extruder section, as they were there only for macros on a test system that is not put into this sample config

signed-off-by: Zachary Welvaert zwelvaert@gmail.com

### DIFF
--- a/config/printer-sunlu-t3-2022.cfg
+++ b/config/printer-sunlu-t3-2022.cfg
@@ -20,7 +20,7 @@
 
 [mcu]
 #obtain your MCU id using ls /dev/serial/by-path/*
-serial: dev/serial/by-id/usb-Klipper_stm32f103xe_*
+serial: dev/serial/by-id/usb-Klipper_stm32f103xe_00000000000
 
 [printer]
 kinematics: cartesian
--- a/config/printer-sunlu-t3-2022.cfg
+++ b/config/printer-sunlu-t3-2022.cfg
@@ -20,7 +20,7 @@
 
 [mcu]
 #obtain your MCU id using ls /dev/serial/by-path/*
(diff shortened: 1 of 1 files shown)


dahliamoth:
Resubmitted PR under updated fork as the previous PR wasn't running the build test. Build test started, but was immediately canceled due to scheduled Ubuntu 18.02 brownout

signed-off-by: Zachary Welvaert zwelvaert@gmail.com

dahliamoth:
Oversight, my apologies. I had wanted to get the revisions posted asap once I found the mistake, as it would cause a printer to not work as it was. Got those parts removed and we should be good to go now

KevinOConnor:
Thanks.  Seems the most recent commit adds an `*` to the `serial:` definition.  I'm leery of putting an `*` in the documentation for fear some users will think they can actually specify `*` as a wildcard.  Feel free to replace the existing printer specific serial number with all zeros or something similar instead.

-Kevin

dahliamoth:
Fair enough. I had just noticed my own serial had been left there. Replaced the * with a string of zeros

KevinOConnor:
Thanks.

-Kevin

PR #6043, opened 2023-02-08
scripts: update_mks_robin.py python env version

JamesH1978:
The recent PR to change this to say `#!/usr/bin/env python` instead of `#!/usr/bin/env python2` can cause issues on non rpi based OS's where `python` is not mapped to `python3`.

`#!/usr/bin/env python3` should work in both situations.

Thanks
James


KevinOConnor:
Thanks.

-Kevin

PR #6044, opened 2023-02-09
Add support for Ender 3/CR-10 stock display

ofussr:


JamesH1978:
Good morning, 

Thank you for submitting a change. 

This is not a good fit for this file, a **generic** config is for a board not for a printer therefore it has no screen.

The printer example configs for Ender 3 and CR10 already contain screen sections.

Thanks
James

PR #4796, opened 2021-10-12
Add MAX_SLOPE capability to TEMPERATURE_WAIT

garethky:
TEMPERATURE_WAIT updated to support an optional MAX_SLOPE argument that allows a wait for the rate of change of temp to drop below that value.

To support this, both Heaters and PrinterSensorGeneric need a common way to provide the temperature slope to TEMPERATURE_WAIT. The code to do this was already in various places in Heaters.py. A new TemperatureSensorAdapter consolidates this functionality. This is used in both Heater and PrinterSensorGeneric so they now all have a known temperature slope which TEMPERATURE_WAIT can access via the adapter. This removes corresponding functionality from Heater and the control algorithms

Signed-off-by: Gareth Farrington <gareth@waves.ky>

This is an update for https://github.com/Klipper3d/klipper/pull/4273. I'm sorry I didn't have time to make the requested updates until now. Hopefully this version is closer to what you had in mind. It includes all of the changes you requested and tries to stay away from any unnecessary edits or changes to how Heaters or temperature sensors work now.

PR #423, opened 2018-06-26
add pid_settle_* settings

hg42:
see #67, which I closed because I switched to cruwallers fork.

The patch makes the fixed constants PID_SETTLE_* in heater.py accessible in the config file.

Motivation:
There are way too different heaters to make them constant or find good defaults.
The constants only work for
* small overshoot < 1° where the printing starts when reaching the maximum overshoot
* for fast cool down (low heat capacity), where a bigger overshoot settles quickly

For example, my heated bed (300mm x 300mm x 9mm aluminum with a 2000W sauna heater) has a very high heat capacity and lots of power. This is very different from a small 40W hotend.
For such a heated bed overshooting isn't handled nice if the accepted overshoot is beyond PID_SETTLE_DELTA (which is only 1°). The heat capacity leads to a very long cool down and stays above PID_SETTLE_SLOPE for a long time (about 30 min).
Actually, my current settings are
```
pid_settle_delta = 1
pid_settle_slope = 9999
```
This effectively disables `pid_settle_slope` and starts printing when the bed first reaches 1° below the target temperature.
The amount of overshoot is already handled by the PID parameters.

Signed-off-by: Harald Gutsche <hg42@gmx.net>

PR #67, opened 2017-12-19
add config options for pid settling parameters

hg42:
allow adjustments of these pid parameters, because my bed doesn't work convenient otherwise (fast heat up but very slow cool down, waiting for eternity).

PR #6047, opened 2023-02-09
Config options for temperature wait parameters

Tekagi01:
I added the possibility to change the previous constants PID_SETTLE_DELTA and PID_SETTLE_SLOPE through the config file.
This gives users with noisy temperature readings a possibility to use the "wait for temperature" functionality (M109 etc.) by changing those parameters.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -840,6 +840,20 @@ max_temp:
 #   heater and sensor hardware failures. Set this range just wide
 #   enough so that reasonable temperatures do not result in an error.
 #   These parameters must be provided.
+#pid_settle_delta: 1.0
+#   The maximum delta between the set and measured (smoothed) temperature.
+#   This setting is used in conjunction with 'pid_settle_slope' to
+#   signal the heater has reached target temperature.
+#   Changing this should only be needed as a workaround for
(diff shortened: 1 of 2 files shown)


KevinOConnor:
There was a discussion on this in #423 .  Basically, there isn't much desire to change the behavior of "old g-code commands".  One can use the `TEMPERATURE_WAIT` command if an alternate behavior is desired.

-Kevin

Tekagi01:
> There was a discussion on this in #423 . Basically, there isn't much desire to change the behavior of "old g-code commands". One can use the `TEMPERATURE_WAIT` command if an alternate behavior is desired.
> 
> -Kevin

I totally missed #423 and #67, sorry.
The thing is, that the behavior of the "legacy g-code commands" is the desired one which `TEMPERATURE_WAIT` just can't provide, because it can't wait for the temperature to settle. #4796 didn't seem to have led to anything...

But I don't want to start yet another discussion, so feel free to close.

-Henning

KevinOConnor:
I understand.  If you (or someone else) wants to extend TEMPERATURE_WAIT with additional parameters to support "settling" then I think that is fine.  I don't think we should change "old g-code commands" though.

Cheers,
-Kevin

PR #6048, opened 2023-02-10
x_twist_compensate: Add X twist compensation module

koonweee:
# Context
Printers with a probe that have a probe y_offset may give inaccurate probe results due to a twist in the probe's X axis. This is common in printers with designs like the Prusa MK3, Sovol 06 etc, as described under [probe location bias](https://www.klipper3d.org/Probe_Calibrate.html#location-bias-check) in the Klipper documentation. It may result in probe operations such as Bed Mesh, Screws Tilt Adjust, Z Tilt Adjust etc returning inaccurate representations of the bed.

The following diagram (by @MauroGil66) and video help to visualize the issue
image
[Video "Visualization of x-sled-rotation", by Elhanan Maayan](https://www.youtube.com/watch?v=rPU4ZAPJTNc)

This issue is also often [reported on Marlin](https://github.com/MarlinFirmware/Marlin/issues/22791), and some of the users on the Klipper discourse [[Post 1](https://klipper.discourse.group/t/qgl-with-bed-mesh-still-cant-get-gantry-level-to-bed/4584/9), [Post 2](https://klipper.discourse.group/t/impossible-bed-mesh-leveling-cr10spro-v2/204/75)]

# Fix
The fix implemented adapts math from the [fix implemented in Marlin](https://github.com/MarlinFirmware/Marlin/pull/23238). The user can manually calibrate the z-offset at a few points along the x axis, and linear interpolation is done to figure out the appropriate z-offset at all other points.

`probe.py` is modified such that the z-compensation is applied for all calls to the probe to determine a z-height. This allows modules that use the probe such as bed_mesh and z_tilt_adjust to benefit from the calibrated probe z-heights.

# Testing
Admittedly, real world testing is currently limited on this. I have personally tested it to work wonderfully on my own SV06, and have posted it in a few Discord channels/the Klipper discourse for more users to test it out.

All comments, criticism, and feedback are very welcome! Thank you.

Signed-off by: Jeremy Tan jeremytkw98@gmail.com

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1302,6 +1302,43 @@ print.
 #### SDCARD_RESET_FILE
 `SDCARD_RESET_FILE`: Unload file and clear SD state.
 
+### [x_twist_compensation]
+
+The following commands are available when the
+[x_twist_compensation config section](Config_Reference.md#x_twist_compensation)
+is enabled.
+
+#### X_TWIST_CALIBRATE
+`X_TWIST_CALIBRATE [N_POINTS=<value>]`: Initiates the X twist calibration
+wizard. `N_POINTS` specifies the number of points along the X axis to calibrate
+at and defaults to 3.
+
+#### X_TWIST_COMPENSATE_STATUS
+`X_TWIST_COMPENSATE_STATUS`: Returns the status of the `[x_twist_compensation]`
+module. If compensation is enabled, this prints the `z_compensations` values and
+`recommended_z_offset`.
+
+#### X_TWIST_COMPENSATE_MESH
+`X_TWIST_COMPENSATE_MESH MESH_NAME=<value> COMPENSATION_NAME=<value>`: Applies
+a stored compensation profile of name `COMPENSATION_NAME` to a stored mesh
+profile of name `MESH_NAME`. This is useful for correcting a mesh that was taken
+while compensation was turned off/not enabled.
+
+#### X_TWIST_PROFILE_LOAD
+`X_TWIST_PROFILE_LOAD NAME=<value>`: Loads the stored compensation profile of
+name `NAME` as the active profile.
+
+#### X_TWIST_PROFILE_CLEAR
+`X_TWIST_PROFILE_CLEAR`: Clears the active profile.
+
+#### X_TWIST_PROFILE_DELETE
+`X_TWIST_PROFILE_DELETE NAME=<value>`: Deletes the stored compensation profile
+of name `NAME`.
+
+#### X_TWIST_PROFILE_SAVE
+`X_TWIST_PROFILE_SAVE NAME=<value>`: Saves the active compensation profile under
+the name `NAME`
+
 ### [z_thermal_adjust]
 
 The following commands are available when the
--- a/klippy/extras/x_twist_compensation.py
+++ b/klippy/extras/x_twist_compensation.py
@@ -1,0 +1,633 @@
+# X Twist Compensation
+# Copyright (C) 2022  Jeremy Tan <jeremytkw98@gmail.com>
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+"""
+[x_twist_compensation]
+horizontal_move_z: 10
+speed: 50
+start_x: 0 ; nozzle's x coordinate at the start of the calibration ! required
+end_x: 200 ; nozzle's x coordinate at the end of the calibration ! required
+y: 100 ; nozzle's y coordinate during the calibration ! required
+"""
+
+import logging
+import math
+from . import manual_probe as ManualProbe, bed_mesh as BedMesh
+
+DEFAULT_N_POINTS = 3
+DEFAULT_PROFILE_NAME = 'default'
+BED_MESH_CONFIG_NAME = 'bed_mesh'
+
+
+class Config:
+    # values stored in printer.cfg when a profile is saved
+    DEFAULT_SPEED = 50.
+    DEFAULT_HORIZONTAL_MOVE_Z = 10.
+    REQUIRED = True
+    OPTIONAL = False
+    CONFIG_OPTIONS = {
+        'horizontal_move_z': (float, OPTIONAL, DEFAULT_HORIZONTAL_MOVE_Z),
+        'speed': (float, OPTIONAL, DEFAULT_SPEED),
+        'start_x': (float, REQUIRED, None),
+        'end_x': (float, REQUIRED, None),
+        'y': (float, REQUIRED, None)
+    }
+
+
+class XTwistCompensation:
+    def __init__(self, config):
+        # get printer
+        self.printer = config.get_printer()
+
+        # get values from [x_twist_compensation] section in printer .cfg
+        for config_key, \
+            (config_type, required, default) in Config.CONFIG_OPTIONS.items():
+            value = None
+            if config_type == float:
+                value = config.getfloat(config_key, default)
+            else:
+                value = config.get(config_key, default)
+            if required and value is None:
+                raise config.error(
+                    "Missing required config option for section [{}]: {}"
+                    .format(config.get_name(), config_key))
+            setattr(self, config_key, value)
+
+        # setup persistent storage
+        self.pmgr = ProfileManager(config, self)
+
+        # setup calibrater
+        calibrater_config = {
+            'horizontal_move_z': self.horizontal_move_z
+                if hasattr(self, 'horizontal_move_z') else None,
+            'speed': self.speed if hasattr(self, 'speed') else None,
+            'start_x': self.start_x if hasattr(self, 'start_x') else None,
+            'end_x': self.end_x if hasattr(self, 'end_x') else None,
+            'y': self.y if hasattr(self, 'y') else None
+        }
+        self.calibrater = Calibrater(
+            config, self.pmgr, calibrater_config)
+
+        self.enabled = False
+
+        # register gcode handlers
+        self._register_gcode_handlers()
+
+    def _register_gcode_handlers(self):
+        # register gcode handlers
+        self.gcode = self.printer.lookup_object('gcode')
+        self.gcode.register_command(
+            'X_TWIST_COMPENSATE_MESH',
+            self.cmd_X_TWIST_COMPENSATE_MESH,
+            desc=self.cmd_X_TWIST_COMPENSATE_MESH_help)
+        self.gcode.register_command(
+            'X_TWIST_COMPENSATE_STATUS',
+            self.cmd_X_TWIST_COMPENSATE_STATUS,
+            desc=self.cmd_X_TWIST_COMPENSATE_STATUS_help)
+
+    def get_z_compensation_value(self, x_coord, optional_profile_name=None):
+        # returns the (lineraly interpolated) z compensation value
+        # for the given x coordinate
+        # uses the current profile if optional_profile_name is not specified
+        enabled = self.pmgr.get_is_enabled()
+        if enabled or optional_profile_name is not None:
+            current_profile = \
+                self.pmgr.get_current_profile() \
+                if optional_profile_name is None \
+                else self.pmgr.get_profile(optional_profile_name)
+            z_compensations = current_profile.z_compensations
+            n_points = len(z_compensations)
+            spacing = (self.end_x - self.start_x) / (n_points - 1)
+            interpolate_t = (x_coord - self.start_x) / spacing
+            interpolate_i = math.floor(interpolate_t)
+            interpolate_i = BedMesh.constrain(interpolate_i, 0, n_points - 2)
+            interpolate_t -= interpolate_i
+            interpolated_z_compensation = BedMesh.lerp(
+                interpolate_t, z_compensations[interpolate_i],
+                z_compensations[interpolate_i + 1])
+            return interpolated_z_compensation
+        else:
+            return 0
+
+    cmd_X_TWIST_COMPENSATE_MESH_help = \
+    "Compensate a mesh by applying the x" \
+    "twist compensation to the given raw mesh"
+
+    def cmd_X_TWIST_COMPENSATE_MESH(self, gcmd):
+        # get the mesh name from the gcode command
+        raw_mesh_name = gcmd.get('MESH_NAME', None)
+        Helpers.check_non_empty_param(raw_mesh_name, gcmd, 'MESH_NAME')
+
+        # get the compensation profile name from the gcode command
+        compensation_name = gcmd.get('COMPENSATION_NAME', None)
+        Helpers.check_non_empty_param(
+            compensation_name, gcmd, 'COMPENSATION_NAME')
+
+        # get the bed_mesh object, then the bed_mesh profile manager
+        bed_mesh = self.printer.lookup_object('bed_mesh', None)
+        if not bed_mesh:
+            raise gcmd.error(
+                "[bed_mesh] is not specified in your printer configuration")
+        bed_mesh_pmgr = bed_mesh.pmgr
+        # load specified bed mesh as active bed mesh
+        bed_mesh_pmgr.load_profile(raw_mesh_name)
+        # get the active bed mesh
+        active_bed_mesh = bed_mesh.get_mesh()
+        # modify the probed matrix by applying the x twist compensation
+        modified_probed_matrix = self._modify_probed_matrix(
+            active_bed_mesh, compensation_name)
+
+        # update active mesh with modified probed matrix, save under new name
+        compensated_mesh_name = \
+            raw_mesh_name + '_compensated_' + compensation_name
+        active_bed_mesh.build_mesh(modified_probed_matrix)
+        bed_mesh_pmgr.save_profile(compensated_mesh_name)
+
+    def _modify_probed_matrix(self, bed_mesh, compensation_profile_name):
+        # do compensating, by modifying z values in probed matrix
+        # probed matrix is a list of rows of probed z values
+        # eg. probed_matrix[0][0] = bottom left corner of mesh, z value
+        probed_matrix = bed_mesh.get_probed_matrix()
+        compensated_matrix = []
+        for row_index in range(len(probed_matrix)):
+            compensated_row = []
+            row = probed_matrix[row_index]
+            for col_index in range(len(row)):
+                z = row[col_index]
+                x_coord = self._get_mesh_point_x_coord(col_index, bed_mesh)
+                compensated_z = z + \
+                    self.get_z_compensation_value(
+                        x_coord, compensation_profile_name)
+                compensated_row.append(compensated_z)
+            compensated_matrix.append(compensated_row)
+        # compensated_matrix is a list of list
+        # bed_mesh expects tuple of tuple hence convert
+        compensated_matrix = tuple(tuple(row) for row in compensated_matrix)
+        return compensated_matrix
+
+    def _get_mesh_point_x_coord(self, col_index, mesh):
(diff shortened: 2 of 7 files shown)


koonweee:
Installation instructions for interested testers:
```
# change directory to your klipper install folder on your RPI/other klipper host
cd klipper

# add my klipper fork as a remote, and fetch it
git remote add koonweee https://github.com/koonweee/klipper.git
git fetch koonweee

# checkout the new "x_twist_compensation" module and the modified "probe" module
git checkout remotes/koonweee/jt/x-twist-compensation-simple klippy/extras/x_twist_compensation.py klippy/extras/probe.py
```

heqstf:
Hi @koonweee,

 Is this X twist compensation module,rearranged to XY? i.e: cross xy could have a twisted Y too. So the given reads could be a misrepresentation of your xy plane twist.
 
 Cheers.

LHARD:
Hi @koonweee ,

I just tested your work.

I did everything as described, but for me it didn't work; at least not without a small modification:
I had to invert the values in the printer.cfg to make the mesh look similar to my manual mesh.
Didn't do a new test-print since then. But first test was just as off as the mesh looked like in mainsail.

I use a Geeetech A30M with a creality sprite pro extruder (and obviously klipper).

koonweee:
> Hi @koonweee,
> 
> Is this X twist compensation module, be rearranged to XY? i.e: cross xy could have a twisted Y too. So the given reads could be a misrepresentation of your xy plane twist.
> 
> Cheers.

Hey! It is possible to generalize the module for the y axis too. The compensation required would be a summation of both the x and y axis if both are twisted. I believe X twist issues are more common however

koonweee:
> 

Hey thanks for testing the module. Could you elaborate on your probe setup on your printer? (eg. what probe is used)

koonweee:
> Hi @koonweee ,
> 
> I just tested your work.
> 
> I did everything as described, but for me it didn't work; at least not without a small modification: I had to invert the values in the printer.cfg to make the mesh look similar to my manual mesh. Didn't do a new test-print since then. But first test was just as off as the mesh looked like in mainsail.
> 
> I use a Geeetech A30M with a creality sprite pro extruder (and obviously klipper).

Hey thanks for testing the module. Could you elaborate on your probe setup on your printer? (eg. what probe is used)

fauks:
I will be testing this solution this week/weekend. I have mesh issues that I have narrowed down to X extrusion twisting. Unfortunately I cannot manually level this bed so I'm hoping this will work and I will report back my findings and printer information.

LHARD:
Hey, I edited my previous post with the missing info.

Sent with Spark
Am 12. Feb. 2023, 13:45 +0100 schrieb Jeremy Tan ***@***.***>:

koonweee:
> Hi @koonweee ,
> 
> I just tested your work.
> 
> I did everything as described, but for me it didn't work; at least not without a small modification: I had to invert the values in the printer.cfg to make the mesh look similar to my manual mesh. Didn't do a new test-print since then. But first test was just as off as the mesh looked like in mainsail.
> 
> I use a Geeetech A30M with a creality sprite pro extruder (and obviously klipper). My Probe is a Geeetech 3D Touch V3.2 Pro, mounted at the stock position for the CRTouch. It is quite far away from the nozzle: x_offset: -33 y_offset: -46

Just a guess on my part, but does `PROBE_CALIBRATE` work correctly on your setup? eg. does it probe a point, and then corrctly move the nozzle over that point for manual z offset calibration?

koonweee:
> I will be testing this solution this week/weekend. I have mesh issues that I have narrowed down to X extrusion twisting. Unfortunately I cannot manually level this bed so I'm hoping this will work and I will report back my findings and printer information.

Thank you and please let me know! out of curioisty, why cant your bed be manually levelled?

fauks:
>>
> Thank you and please let me know! out of curiosity, why cant your bed be manually levelled?

The build plate is hard mounted. I can actually level it since I have modified the bed, but the screws are located under the flex plate which makes tramming via paper/feeler gauge difficult as I would have to move the nozzle every time I needed to make an adjustment. There are also 12 mounting points instead of your typical 4. I however can use screws_tilt_calculate but I think this was not working right because of possible twist in the X axis. 

I tried it last night, it was more simple than I thought it would be. I had to manually adjust the values myself across 3 points but it did indeed work. I didn't have any non-magnetic feeler gauges so I first did the paper test and got better results than what I was getting prior, and then manually adjusted the values myself until now the whole plate has a great first layer. 

What method would you use for a bed that cannot be trammed? Like, what order of leveling with only access to screws_tilt_adjust. Should I run that without x_twist_compensation enabled first, then preform x_twist after, and then re-level using screws_tilt_calculate? 

I'm thinking the order I did things may have made the results not perfect first try around, or maybe its because I was using the paper test which can be pretty inaccurate. 

Anyways, thank you so much for this! I haven't had a good first layer in almost a year and have been ready to trash this printer so I'm pretty stoked.

koonweee:
> > > 
> > 
> > 
> > Thank you and please let me know! out of curiosity, why cant your bed be manually levelled?
> 
> The build plate was hard mounted. I have modified the bed so I can manually adjust the bed, but the screws are located under the flex plate which makes tramming via paper/feeler gauge difficult as I would have to move the nozzle and flex plate every time I needed to make an adjustment. The underside "nut" is just tapped into the Y carriage so I can't use a normal wheel nut to adjust from below. There are also 12 mounting points instead of your typical 4. I however can use screws_tilt_calculate since I can make all 12 adjustments, then re-run it until "level" but I think this was not working right because of possible twist in the X axis. I was having issues where the far right was nearly touching the bed, and the far left was stringing. My BLTouch sensor offset is x: -41, y: +25 from the nozzle. So my probe is actually behind the nozzle which explains why I'm see the opposite pattern as people with a probe in front of the nozzle.
> 
> I tried it last night, it was more simple than I thought it would be. I had to manually adjust the values myself across 3 points but it did indeed work. I didn't have any non-magnetic feeler gauges so I first did the paper test and got better results than what I was getting prior, and then manually adjusted the values myself until now the whole plate has a great first layer.
> 
> What method would you use for a bed that cannot be trammed? Like, what order of leveling with only access to screws_tilt_adjust. Should I run that without x_twist_compensation enabled first, then preform x_twist after, and then re-level using screws_tilt_calculate?
> 
> I'm thinking the order I did things may have made the results not perfect first try around, or maybe its because I was using the paper test which can be pretty inaccurate.
> 
> Anyways, thank you so much for this! I haven't had a good first layer in almost a year and have been ready to trash this printer so I'm pretty stoked.
> 
> EDIT: I also used the value recommended for probe_offset which worked nearly perfect. I adjusted it a bit after but no more than 0.02mm. I have not tried to use probe_calibrate after doing any of the above. I saw you asked another user about that.

Great to hear! I actually have a similar setup with my sv06 where the 4 mounting points have been hardmounted with nylock nuts but I can use bed_screws_calculate to figure out how much to adjust them. As you noted, using this compensation allows you to use bed_screws_calculate, since the probe readings are now correct. 

Since the x_twist_compensation profile corrects your probe readings, I recommend having compensation profile loaded before you perform any bed_screws_calculate or bed_mesh.

I think the slight inaccuracy you saw was indeed from using the paper test. I also use the paper since I do not have any feeler gauges and also have to tweak the recommended probe_offset and generated compensation values ever so slightly to get a perfect first layer. With a feeler gauge, I'm quite sure the generated compensation would be even more accurate on the first try.

Do not worry about the probe_calibrate, I asked the other user to perform it just to make sure that their probe setup is correct since they were facing unexpected results (the x twist correction generated was inverted from what was expected)

fauks:
>>
> Great to hear! I actually have a similar setup with my sv06 where the 4 mounting points have been hardmounted with nylock nuts but I can use bed_screws_calculate to figure out how much to adjust them. As you noted, using this compensation allows you to use bed_screws_calculate, since the probe readings are now correct.
> 
> Since the x_twist_compensation profile corrects your probe readings, I recommend having compensation profile loaded before you perform any bed_screws_calculate or bed_mesh.
> 
> I think the slight inaccuracy you saw was indeed from using the paper test. I also use the paper since I do not have any feeler gauges and also have to tweak the recommended probe_offset and generated compensation values ever so slightly to get a perfect first layer. With a feeler gauge, I'm quite sure the generated compensation would be even more accurate on the first try.
> 
> Do not worry about the probe_calibrate, I asked the other user to perform it just to make sure that their probe setup is correct since they were facing unexpected results (the x twist correction generated was inverted from what was expected)

I had been meaning to pick up a set of plastic feeler gauges or at least some non magnetic. Maybe I have more of an excuse for it now!

I'm about to start a full plate of prints which I haven't been able to do in a very long time :)

LHARD:
> > Hi @koonweee ,
> > I just tested your work.
> > I did everything as described, but for me it didn't work; at least not without a small modification: I had to invert the values in the printer.cfg to make the mesh look similar to my manual mesh. Didn't do a new test-print since then. But first test was just as off as the mesh looked like in mainsail.
> > I use a Geeetech A30M with a creality sprite pro extruder (and obviously klipper). My Probe is a Geeetech 3D Touch V3.2 Pro, mounted at the stock position for the CRTouch. It is quite far away from the nozzle: x_offset: -33 y_offset: -46
> 
> Just a guess on my part, but does `PROBE_CALIBRATE` work correctly on your setup? eg. does it probe a point, and then corrctly move the nozzle over that point for manual z offset calibration?
> 
> edit: Let me know about the above, but I will also write some test cases soon which should catch any anomalies in the code

PROBE_CALIBRATE works normal incl. moving the nozzle.
Just some random info: my z-endstop is some mm **above** my z=0 point; so after homing the nozzle will go down

murdock62:
Hi @koonweee,

I performed several tests. Since the result was not yet perfect and my mesh is done on 9 points, 
I performed an X_TWIST_CALIBRATE N_POINTS= 9.

The calibration went well, but when I want to perform a BED_MESH_CALIBRATE it probes the first point 
well and when it goes to the second klipper crashes and sends me this message :

"list indices must be integers, not float"


By leaving default (so with 3 points) no worries. But the result is not perfect. 
(with a 350mm bed this is probably normal)

Thank you for your help.

murdock62:
Hi @koonweee,

I looked quickly.
I think by simply replacing:

`interpolate_i = BedMesh.constrain(interpolate_i, 0, n_points - 2)`
by
`interpolate_i = int(BedMesh.constrain(interpolate_i, 0, n_points - 2))`

Must do the job. But I'm not a python expert.

murdock62:
I created a small simulation program based on your function (with the addition of the int conversion of the index) 
when calling the "_probe" function of klipper and it looks good, it gives the following results:

Simple_X_Twist correction

The way "interpolate_t" is calculated is used in the calculation of the linear correction gives a small error but 
that is really not far we are talking in the worst case of an error of about 3 thousandths of mm.

Now all that remains is to test with 9 points in real on my machine.

Many thanks for your work.

koonweee:
> 

Hey! Did you get round to testing it on your machine?

murdock62:
Hi @Koonweee.

Yes I'm not yet at the end of everything I wanted to test but if all goes well 
I'll be on the weekend from this Thursday evening and I'll be able to finish my tests. 

However, I still have the first results which are rather very encouraging. It's not perfect, 
but it's definitely better. 

However, I ask myself two questions about the small lack to achieve perfection :

- The problem could come from my bed. Clearly on this machine the bed is not great.
A deviation of about 0.4mm (bed from a fysetc kit) and clearly bell-shaped.

- The calibration procedure can on this type of machine (Voron 2.4) induce an error.
I will try to explain clearly.

Let's assume that this is my first use of your function.
To be able to launch your function I must at least make a home, a quad_gantry_level (voron 2.4 obliges)
another home to take into account the correction of the QGLet then I launch the calibration of your function.
However, the measurement reading of your calibration function is based on a QGL which is made with a z-offset 
which is itself biased by the twist of the axis.

When I launch a print I start by activating the compensation then home then QGL (this time with the compensation 
activated) then home then probe of the bed_mesh_calibrate.


The problem is that this QGL carried out with the compensation certainly gives a better mechanical alignment of 
the 4 Z axes but this change of alignment of the Z axes can still modify the angle of the axis and therefore partially 
invalidates the values ​​of compensation.

We could perhaps improve the linearity of the correction by playing on the calculation of the "interpolate_t" variable, 
but frankly this already gives very good results.

Congratulations and thank you again for your work.

fauks:
@koonweee I finally got around to redoing it with plastic feeler gauges and it worked perfectly. I can print full plates of parts with no deviation from side to side. 

I think I might have a slight twist in my Y axis as well though, so if we could get this to do both that would be amazing! Deviation from front to back is pretty minor, but noticeable when printing a full plate. Probably less than 0.1mm.

koonweee:
> @koonweee I finally got around to redoing it with plastic feeler gauges and it worked perfectly. I can print full plates of parts with no deviation from side to side.
> 
> I think I might have a slight twist in my Y axis as well though, so if we could get this to do both that would be amazing! Deviation from front to back is pretty minor, but noticeable when printing a full plate. Probably less than 0.1mm.

Thanks for the update! Doing it for the Y axis should be possible as well with some refactoring of the code. I am hoping to do this soon (swamped with school at the moment 😓 )

fauks:
> > @koonweee I finally got around to redoing it with plastic feeler gauges and it worked perfectly. I can print full plates of parts with no deviation from side to side.
> > I think I might have a slight twist in my Y axis as well though, so if we could get this to do both that would be amazing! Deviation from front to back is pretty minor, but noticeable when printing a full plate. Probably less than 0.1mm.
> 
> Thanks for the update! Doing it for the Y axis should be possible as well with some refactoring of the code. I am hoping to do this soon (swamped with school at the moment 😓 )

That would be awesome! No rush, I appreciate what you have done already!

klippy/extras/x_twist_compensation.py
```diff
@@ -0,0 +1,633 @@
...
+        # calculate average of results
+        avg = sum(self.results) / len(self.results)
+        # subtract average from each result
+        # so that they are independent of z_offset
+        self.results = [avg - x for x in self.results]
+        # create a new profile using profile manager
+        self.pmgr.create_profile(profile_name, self.results, avg)
+        # recommend z offset to user
+        self.gcmd.respond_info(
+            "X_TWIST_CALIBRATE: Calibration complete, reccomended z_offset: %f"
```
mhaas, line 374:
```suggestion
            "X_TWIST_CALIBRATE: Calibration complete, recommended z_offset: %f"
```

docs/G-Codes.md
```diff
@@ -1302,6 +1302,43 @@ print.
...
 `SDCARD_RESET_FILE`: Unload file and clear SD state.
 
+### [x_twist_compensation]
+
+The following commands are available when the
+[x_twist_compensation config section](Config_Reference.md#x_twist_compensation)
+is enabled.
+
+#### X_TWIST_CALIBRATE
+`X_TWIST_CALIBRATE [N_POINTS=<value>]`: Initiates the X twist calibration
```
mhaas, line 1312: Include `PROFILE_NAME` here?

klippy/extras/x_twist_compensation.py
```diff
@@ -0,0 +1,633 @@
...
+        "Loads a saved mesh as the active mesh"
+
+    def cmd_X_TWIST_PROFILE_LOAD(self, gcmd):
+        # loads a saved mesh as the active mesh
+        profile_name = gcmd.get('NAME', None)
+        Helpers.check_non_empty_param(profile_name, gcmd, 'NAME')
+        self.load_profile(profile_name)
+
+    cmd_X_TWIST_PROFILE_CLEAR_help = \
+        "Clears the active mesh"
```
mhaas, line 571: Is mesh the correct wording here?
```suggestion
        "Clears the active X twist compensation"
```

klippy/extras/x_twist_compensation.py
```diff
@@ -0,0 +1,633 @@
...
+
+    def cmd_X_TWIST_PROFILE_CLEAR(self, gcmd):
+        # clears the active mesh
+        self.clear_profile()
+
+    cmd_X_TWIST_PROFILE_SAVE_help = \
+        "Saves the active mesh to the config file"
```
mhaas, line 578:
```suggestion
        "Saves the active X twist compensation profile to the config file"
```

mhaas (commented):
Drive-by review - currently testing the PR and figured I would check the code as well :)

murdock62:
> 

After more tests I can say that according to my results this function brings a significant improvement.

On a machine where the X axis is fixed like a cr-10 or other, we are almost at perfection.

On a machine like a voron 2.4 there is still a very small defect but I think that as mentioned above 
the calibration being done after a QGL which does not use twist compensation if the QGLs made during 
the prints use this compensation the QGL will be better but the offset difference between the nozzle 
and the probe will no longer be the same and therefore the compensation will not be perfect. 

I think best when printing and only activating the correction after the QGL. 
This QGL will be less precise but will relace the machine with the same twist as during calibration. 
The compensation of the mesh will therefore be better.

Good work !!!

Jolliness0995:
I have a SV06 and I want to let you know that this module has made a huge difference! It seems to be working great for me!

KevinOConnor:
Interesting, thanks.  As high-level feedback, compensating for probe location based inaccuracies seems like useful functionality.  When reviewing the code it seems to me like the feature is in an "experimental" phase though.  So, I'm not sure it is ready to merge to the master Klipper repository.

Some high-level comments:
1. It is preferable to add only a few commands with each new module.  I'd expect this module to add an `X_TWIST_COMPENSATE_CALIBRATE` command.  Most of the other commands don't immediately seem necessary to me.
2. It's not clear to me why there is a "profile" system.  Why would a user define multiple "twist profiles".  Seems it would be simpler to have a single twist compensation.
3. It seems odd that this module only applies to X.  I'd imagine some printers would have hardware where the Y axis would be more likely to twist.  (I'm not suggesting the module needs to have more capabilities, but from a module naming perspective it would seem unfortunate to end up in a position where we have multiple "twist compensation" modules just due to axis naming.)

Cheers,
-Kevin

blastrock:
Thanks a lot for porting this feature to Klipper! I also have a SV06 and was thinking I had to give up the automatic bed mesh feature when switching from Marlin to Klipper.

I have tested your code (rebased on v0.11.0, but it shouldn't matter much). I have found a couple bugs, you can look at my fixes here https://github.com/blastrock/klipper/commit/0fc4d9ff77bea95f0223b6aee7d30fb40092c238 .

The first bug is that all offsets were calculated against the same measured Z from the probe, at the launch of Klipper. This not only included the X twist offset, but also the offset from the bed not being leveled. It looks like your patch discards the measured Z on each point.

The second bug is that the results array was not reset upon rerunning the calibration. This made the array contain 6 values at the end of the calibration instead of 3.

Really looking forward to this being merged :)

Arksine:
Having looked at this PR I agree with Kevin's comments.  I don't think its a good idea to transform saved mesh profiles, instead we should encourage users recalibrate the mesh after twist calibration. 

It may be worth exploring whether or not twist is impacted by heat.  If there is data that shows a measurable impact based on temperature then profiles may be useful.  Should it turn out that profiles are useful I think it would be desirable for the commands to be consistent with `bed_mesh`, where there is a single gcode command responsible for profile management.

blastrock:
Hi there,

Since there isn't much activity on this PR, I took the liberty of fixing what was pointed out during this review: https://github.com/Klipper3d/klipper/pull/6149

Flol0:
Is there still someone watching?

Sineos:
@Flol0 
It might have made sense to read at least the second last post before posting ...

PR #6049, opened 2023-02-10
scripts: Update flash-linux.sh to restart new systemd service

JamesH1978:
This is an update to the script that looks for the new systemd service and restarts it if appropriate.

Thanks
James

### DIFF
--- a/scripts/flash-linux.sh
+++ b/scripts/flash-linux.sh
@@ -24,3 +24,9 @@ if [ -f /etc/init.d/klipper_mcu ]; then
     echo "Attempting host MCU restart..."
     service klipper_mcu restart
 fi
+
+if [ -f /etc/systemd/system/klipper-mcu.service ]; then
+    echo "Attempting host MCU restart..."
+    service klipper-mcu restart
+fi
+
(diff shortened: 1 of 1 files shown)


scripts/flash-linux.sh
```diff
@@ -24,3 +24,8 @@ if [ -f /etc/init.d/klipper_mcu ]; then
     echo "Attempting host MCU restart..."
     service klipper_mcu restart
 fi
+
+if [ -f /etc/systemd/system/klipper-mcu.service ]; then
+    echo "Attempting host MCU restart..."
+    service klipper-mcu restart
```
meteyou, line 30:
```suggestion
    systemctl restart klipper-mcu
```
I would use the systemctl command instead of service for systemd services.

JamesH1978:
You are of course right, I was just following the syntax of the file.

I will make that change :)

Thanks,
James

meteyou (approved):

thijstriemstra (commented):
Can you also add a comment on line 27 (like the other commands in this script have)?

PR #6051, opened 2023-02-11
Add better creality support to spi_flash

mnorrsken:
Replaces the text %%HASH%% in firmware_path with a somewhat unique string based on the firmware hash. This allows for the filename to be unique for new firmware versions, so that creality bootloader will read these. This required some refactoring of hashing code in spi_flash.

Also adds support for Creality CR-FDM-v2 board based on STM32F401

Creality boards needs to be power cycled after flashing, also if two firmware versions are uploaded without a power cycle in between the bootloader seems to ignore the upgrade.


JamesH1978:
Please clean up your whitespace and review https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying particular attention to the signed-off-by line.

Thanks
James

mnorrsken:
An addition, the SD card has to have a directory named STM32F4_UPDATE, i could not find any commands to create directories in the fat library, but if that exists, i can add a test and create the directory if it dont exist.

Arksine:
These changes look good to me.  Its possible that #5759 could result in merge conflicts for this if its merged first, so something to keep in mind.

With regard to creating directories, its possible with the fatfs library however this is not implemented in the cffi wrapper.

KevinOConnor:
Thanks.  Change is fine with me, but it does look like a `git rebase` will be needed.

-Kevin

mnorrsken:
Woops im sorry, I didnt understand I was to do anything. I will create a new PR from the latest source.

Dids:
> Woops im sorry, I didnt understand I was to do anything. I will create a new PR from the latest source.

Any update on that PR?

If you do happen to get it going again, please let us know here, as I'm sure many of us would appreciate this feature.

PR #6055, opened 2023-02-13
scripts: update_chitu.py python env version

JamesH1978:
Change script to use Python 3 instead of Python 2, in line with changes to update_mks_robin.py

Thanks
James

### DIFF
--- a/scripts/update_chitu.py
+++ b/scripts/update_chitu.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 # Encodes STM32 firmwares to be flashable from SD card by Chitu motherboards.
 # Relocate firmware to 0x08008800!
 


KevinOConnor:
Thanks.  I'm fine with the change, but have you (or someone else) confirmed it actually works on Python3?

-Kevin

JamesH1978:
I have confirmed the scripts runs and outputs on both a pi and a ubuntu server build, but I have no chitu board to test it
image

Is that enough?

Thanks
James

KevinOConnor:
Yes, thanks.

-Kevin

mattaw:
If it is any comfort I compiled with python2 and 3 on my setup, and the md5sum's of both files were identical. I ran into this problem but chickened out of creating a PR as I was in a hurry, and just installed python2. Since I had both installed:

```bash
pi@mainsail:~/klipper $ md5sum ./out/update.cbd
1ca3065cebe7611299d2cbf1c785bb52  ./out/update.cbd
pi@mainsail:~/klipper $ md5sum ./out/update_py3.cbd
1ca3065cebe7611299d2cbf1c785bb52  ./out/update_py3.cbd
```

Nice work @JamesH1978 & @KevinOConnor

PR #6057, opened 2023-02-15
homing: Support lazy homing in G28 command

jschuh:
This adds the `O` optional/lazy homing parameter from Marlin, which skips rehoming for any axis that's already homed. This is very helpful in things like `START_PRINT`, `PARK`, and similar macros, where you want to ensure the axes are all homed but don't want to unnecessarily rehome.

This can already be done today by wrapping `G28` with macros, but the problem with such an approach is that `G28` is special cased for things like homing overrides. So, configs tend to get ugly pretty quickly when you take this route, which has become a growing problem with the increasing popularity of docking probes that require fairly involved homing overrides.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -8,7 +8,7 @@ commands that one may enter into the OctoPrint terminal tab.
 Klipper supports the following standard G-Code commands:
 - Move (G0 or G1): `G1 [X<pos>] [Y<pos>] [Z<pos>] [E<pos>] [F<speed>]`
 - Dwell: `G4 P<milliseconds>`
-- Move to origin: `G28 [X] [Y] [Z]`
+- Move to origin: `G28 [O] [X] [Y] [Z]`
 - Turn off motors: `M18` or `M84`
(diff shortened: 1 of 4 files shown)


int-0x03:
I think this is good option, especially when you do multicolor print, manually.

KevinOConnor:
Thanks.  As high-level feedback, I'm leery of "adding new tricks to old g-code".

In my opinion the `G28` command is cryptic and poorly defined.  I fear adding new flags to it will make it worse.

I think it's fine if you or someone else wants to propose a new `HOME_AXIS` (or similar) that is more capable and better defined.  That is, I'm fine with adding more capable functionality (in particular, if it is also more transparent), but I suggest doing that with a new command.

Cheers,
-Kevin

jschuh:
Agreed regarding the weirdness of `G28`. I don't think this particular behavior alone would necessarily warrant a new command, since it's so easy to implement with a macro wrapper. I just wanted to address the conflict with `homing_override`. Although, if there's value in exposing this and other homing behaviors through a better defined command I'd be happy to take a crack at it.

PR #6058, opened 2023-02-16
Anycubic 4 max pro 2.0 fixed problem with underextrusion after unretract.

int-0x03:
Soorry Guys, but in spent about two years for searching this problem.  I tried every possible calibrations, and they doesn't help. When I could avoid that problem, I did not belive it.

When extruder do unretract, acceleration is too high by default. The acceleration is so great that the plastic simply does not have time to melt, and at the same time, kinematics starts moving. In result I get under extrusion after unretraction. See pictures below.

I have been trying for a long time to figure out how this parameter is calculated by default, and what it was equal to. I don't know. 

```
#max_extrude_only_velocity:
#max_extrude_only_accel:
#   Maximum velocity (in mm/s) and acceleration (in mm/s^2) of the
#   extruder motor for retractions and extrude-only moves. These
#   settings do not have any impact on normal printing moves. If not
#   specified then they are calculated to match the limit an XY
#   printing move with a cross section of 4.0*nozzle_diameter^2 would
#   have.
```

Parts printed with default (empty) `max_extrude_only_accel`:

photo_2023-02-16_11-51-44
photo_2023-02-16_11-52-00

Let's see, that would be if I set  `max_extrude_only_accel: 1500` (truncated model from previous foto):

photo_2023-02-16_13-18-20

Awesome! That's amazing, I can't beleve it. I spent about 2 years (not active searching, because sometimes I lost motivation to search), but now it's good.

Tell me someone, how I can calc default `max_extrude_only_accel` for 0.4 nozzle diameter?

### DIFF
--- a/config/printer-anycubic-4maxpro-2.0-2021.cfg
+++ b/config/printer-anycubic-4maxpro-2.0-2021.cfg
@@ -70,7 +70,9 @@ rotation_distance: 7.488
 nozzle_diameter: 0.400
 filament_diameter: 1.750
 instantaneous_corner_velocity: 1.5
-max_extrude_only_distance: 500
+max_extrude_only_distance: 100
+max_extrude_only_velocity: 50
+max_extrude_only_accel: 1500
 heater_pin: PB4
 sensor_type: ATC Semitec 104GT-2
 sensor_pin: PK5
--- a/config/printer-anycubic-4maxpro-2.0-2021.cfg
+++ b/config/printer-anycubic-4maxpro-2.0-2021.cfg
@@ -140,6 +140,16 @@ screw4: 265, 5
 [filament_switch_sensor filament_sensor]
 switch_pin: ^!PC4
 
+[output_pin buzz]
+pin: PC6
+pwm: True
+
+[output_pin AUTO_POWEROFF]
+pin: PD0
+pwm: True
+cycle_time: 0.02
+value: 1
+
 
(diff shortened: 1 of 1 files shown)


JamesH1978:
Can you confirm if this is filament specific or not? As we do not usually allow these particular modifiers to be set in a base config unless it can be proven that it is something that will affect all filaments?

as point 6 ii in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md states that these should not be stated.

Thanks
James

KevinOConnor:
> Tell me someone, how I can calc default max_extrude_only_accel for 0.4 nozzle diameter?

Extracted from klippy/extruder.py:
```
filament_area = math.pi * (filament_diameter * .5)**2
def_max_cross_section = 4. * nozzle_diameter**2
def_max_extrude_ratio = def_max_cross_section / filament_area
e_accel = max_accel * def_max_extrude_ratio
```

So, for typical 1.75mm filament with 0.4mm nozzle the default would be `max_accel * 0.26608` where `max_accel` is whatever you specified that to be in the `[printer]` config section.

If you were using `max_accel: 3000` then the default max_extrude_only_accel would be 798.24.

I'd guess your slicer retraction length, retraction speed, z hop settings, and the type of filament you print with are all heavily influencing your print quality.

Might be better to move this discussion to Klipper Discourse server.

-Kevin

PR #6060, opened 2023-02-19
rp2040: running without external flash

Sasasu:
Added a new option for rp2040, RP2040_FLASH_NOFLASH, to enable it to run without external flash. This option allows the rp2040 to boot from its internal bootloader and execute code directly from RAM.

With this build option, firmware must be loaded into rp2040 every time it is powered on or reset this can be done by using an external debugger via SWD or using USB to upload a U2F file into bootrom.

This commit also repurposes 6 IO pins that were originally reserved for QSPI to be used as GPIO pins.

### DIFF
--- a/src/rp2040/Kconfig
+++ b/src/rp2040/Kconfig
@@ -53,40 +53,51 @@ config STACK_SIZE
 # Bootloader options
 ######################################################################
 
-config RP2040_HAVE_STAGE2
-    bool
-config RP2040_HAVE_BOOTLOADER
-    bool
-    default y if !RP2040_HAVE_STAGE2
-
 choice
-    prompt "Bootloader offset"
-    config RP2040_FLASH_START_0100
-        bool "No bootloader"
-        select RP2040_HAVE_STAGE2
-    config RP2040_FLASH_START_4000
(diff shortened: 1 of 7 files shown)


Sineos:
https://github.com/Klipper3d/klipper/pull/5889 the same?

Sasasu:
> #5889 the same?

noop. this one runs without NOR flash connect to rp2040. #5889 boot from NOR flash and move code to SRAM.

JamesH1978:
Please review https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying particular attention to the signed-off-by line.

Thanks
James

Sasasu:
image

the USB connection is much more stable than before :>.

Arksine:
Someone with more experience with the rp2040 than I will need to review this, but one thing I am certain of is that reporting a UID that is not unique will break CAN.

KevinOConnor:
Thanks.  As high-level feedback, this seems like a "neat hack" of the rp2040.  However, it's not clear to me how many people would use this in practice.  So, I'm not sure it is a good match for the main Klipper repository.

-Kevin

PR #6061, opened 2023-02-19
stepstats.py: use python 3

thijstriemstra:
Signed-off-by: Thijs Triemstra <info@collab.nl>

### DIFF
--- a/scripts/stepstats.py
+++ b/scripts/stepstats.py
@@ -22,7 +22,9 @@ def main():
                 continue
             args = dict([p.split('=', 1) for p in parts[1:]])
             if parts[0] == 'config_stepper':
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, it doesn't look to me like this code would run successfully on python3 (seems to be missing `b""` decorations).  For any python3 conversion patches I'd really like to see an explicit statement that the new code has been tested on both python2 and python3.

Also, commits should be one topic per commit.  Mixing in a python3 change with a change to `with xxx:` makes it harder to review.  (The `with` syntax is not related to py2 vs py3 as both python versions support both file opening strategies.)  So, best to have a PR with just the required python3 changes, as that is easier to review.

-Kevin

PR #6063, opened 2023-02-19
Config: addition of slicer specific gcode for START_PRINT

JamesH1978:
It is confusing that this macro only specifies to use START_PRINT, when specific parameters are needed to pass the variables in most slicers.

Thanks
James

### DIFF
--- a/config/sample-macros.cfg
+++ b/config/sample-macros.cfg
@@ -10,7 +10,18 @@
 ######################################################################
 
 # Replace the slicer's custom start and end g-code scripts with
-# START_PRINT and END_PRINT.
+# Cura:
+# ; M190 S{material_bed_temperature_layer_0}
+# ; M109 S{material_print_temperature_layer_0}
+# START_PRINT BED_TEMP={material_bed_temperature_layer_0} EXTRUDER_TEMP={material_print_temperature_layer_0}
+# Slic3r variants:
+# M140 S0
+# M104 S0
+# START_PRINT BED_TEMP={first_layer_bed_temperature[0]} EXTRUDER_TEMP={first_layer_temperature[0]}
+#
+# the addition of extra heat gcodes to these commands will tell
+# the individual slicer not to send its own heat gcodes. Note that
+# Cura needs these commented while Slic3r variants do not
 
 [gcode_macro START_PRINT]
 gcode:


config/sample-macros.cfg
```diff
@@ -10,7 +10,18 @@
 ######################################################################
 
 # Replace the slicer's custom start and end g-code scripts with
-# START_PRINT and END_PRINT.
+# Cura:
+# ; M190 S{material_bed_temperature_layer_0}
+# ; M109 S{material_print_temperature_layer_0}
+# START_PRINT BED_TEMP={material_bed_temperature_layer_0} EXTRUDER_TEMP={material_print_temperature_layer_0}
+# Slic3r variants:
```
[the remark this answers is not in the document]
JamesH1978: I was deliberately trying to be generic :) but i get what you mean

Thanks
James

JamesH1978:
My main issue was that that sample macros file just tells you to use START_PRINT which is obviously wrong when people come asking why temperatures aren't working

Sineos:
Interesting. I'm using Cura a lot and I do not need these commented `M109` / `M190`. I just call my `PRINT_START` macro and nothing more.

Further I'm not sure about this at all:
* It is one of many ways to customize Klipper. It is also not mandatory to use. The "traditional" style start-gcode will work as well
* Some start macros I know do not even use the variables but do something like `START_PRINT_PLA`
* There are a lot more variables that could potentially be customized
   * Prusa: https://help.prusa3d.com/article/list-of-placeholders_205643
   * Cura: http://files.fieldofview.com/cura/Replacement_Patterns.html

With macros it is highly difficult where to start and where to end documentation. In doubt I would just place a note to choose the correct Slicer variables acc. to the slicer documentation.

pedrolamas:
@Sineos if you call `PRINT_START` with nothing more, that means Cura is setting the temperatures WHEN it cares (default behavior).

What James is pointing here is that if you want to make sure that the temperature are set via `PRINT_START` macro, then one will need to pass those values to the macro AND make a few more changes to avoid Cura adding the `M109` / `M190`

Sineos:
> Note that Cura needs these commented while Slic3r variants do not

I'm referring to this sentence. Maybe I'm misunderstanding it.

JamesH1978:
basically you have to comment them in cura, and do not comment in the slic3r variants, as cura needs to see commented heater commands with the variables to not add its own, and slic3r variants just needs them there, commented works. SS actually has a little "custom gcode only" tick box that does the same thing. The point of this PR, is that the example macro is incomplete, as in it takes parameters for heating, but gives no example on how to pass them in to the macro.

Thanks
James

Sineos:
> basically you have to comment them in cura, and do not comment in the slic3r variants, as cura needs to see commented heater commands with the variables to not add its own

I cannot confirm this behavior in my Cura 5.2.2

pedrolamas:
@JamesH1978 I think we are mixing things here... all we need is to ensure that the `{material_bed_temperature_layer_0}` and `{material_print_temperature_layer_0}` are passed to the macro, nothing else.

The only reason we add the `M109` and `M190` as comments is for Moonraker to pick that as metadata (so that Fluidd and Mainsail can show the "Preheat" menu option, nothing else)

Sineos:
Interesting. It seems Cura checks if either `material_print_temperature_layer_0` or `M109` or `M190` is used and if not it adds it. Leaving away bed temperature does not change the behavior.

JamesH1978:
This is the crux of the PR, if a user was to use just START_PRINT on its own as the macro suggests, then the slicer will add its own heating codes. This is true in Cura and PS/SS. Which means the parameters set in that macro won't do much at all becausr the G code will already be heating before the macro is run.

pedrolamas:
> So from the findings we know that cura just needs to see those variables to stop putting in the heating codes

Correct, but for Cura those 2 lines with commented `M109` and `M190` can be safely removed (all we need is to pass the temperatures in the line with the `START_PRINT` macro, nothing else)

config/sample-macros.cfg
```diff
@@ -10,7 +10,18 @@
 ######################################################################
 
 # Replace the slicer's custom start and end g-code scripts with
-# START_PRINT and END_PRINT.
+# Cura:
+# ; M190 S{material_bed_temperature_layer_0}
+# ; M109 S{material_print_temperature_layer_0}
```
pedrolamas, line 15: These are not required for Cura

```suggestion
```

StuSerious:
_Useless deep dive into Cura's behaviour since I work constantly with *.def.json files:_

@pedrolamas's findings are indeed default behavior in Cura, but with a catch.
Cura decides whether to add in temperature values when:

1. no explicit `M109`/`M190` code is found in `machine_start_gcode` **&&** no related JSON keys are called (e.g. `material_print_temperature_layer_0`)
2. the `material_print_temp_prepend` and `material_bed_temp_prepend` def.json keys are set to `false` (_default `true`_) in the printer definition file.

This means that any custom (or 3rd party) printer with that does _not_ prepend temp commands will not work if just `PRINT_START` is issued in Cura's `machine_start_gcode`, unless actions are taken to set temperatures klipper-side in the `PRINT_START` routine.

A quick search in Cura's definitions directory reveals [5 printers](https://github.com/Ultimaker/Cura/search?q=%22material_print_temp_prepend%22%3A+%7B+%22default_value%22%3A+false+%7D%2C) that set prepend to `false` in the config.

Most of my (and I would imagine everyone's) custom *.def.json printers also do not prepend temperatures; this is useful for a bunch of reasons but mainly because of the usefulness of handling temperatures firmware-side.

I hope this was of use,
_~S²_

pedrolamas (approved):

KevinOConnor:
Thanks.  As high-level feedback, though, I didn't really understand the new sample-macros.cfg comments when I read them.  Thus I fear new users may get confused.  (I understand the high-level goal and I understand that slicers have quirky requirements - but it wasn't immediately clear to me that the intent of the new comments are to explain that.)

Maybe as an alternative, a couple of paragraphs could be added to docs/Slicers.md and perhaps add a simple `See docs/Slicers.md for information on configuring a START_PRINT macro.` to sample-macros.cfg .

-Kevin

JamesH1978:
Good idea, that was one the other idea bandied around earlier in the comments, it would be easier to explain in the slicer document than a # comment in the sample macro. I will write something up and probably close this and recommit

Thanks
James

StuSerious:
> Good idea, that was one the other idea bandied around earlier in the comments, it would be easier to explain in the slicer document than a # comment in the sample macro. I will write something up and probably close this and recommit

I can do a better write-up of Slicer behavior if that is deemed useful. I have access to a working profile for basically every slicer so I could easily test start gCodes.

JamesH1978:
@StuSerious that reminded me to send my changes :) feel free to commit other slicers to Slicers.md if this PR is merged.

Thanks
James

docs/Slicers.md
```diff
@@ -87,3 +87,33 @@ Klipper's maximum extrusion cross-section check.
...
+
+In Cura, to pass through temperatures, the following start gcode
+would be used:
+
+START_PRINT BED_TEMP={material_bed_temperature_layer_0} EXTRUDER_TEMP={material_print_temperature_layer_0}
+
+In slic3r derivatives such as PrusaSlicer and SuperSlicer, the
+following would be used:
+
+START_PRINT EXTRUDER_TEMP=[first_layer_temperature] BED_TEMP=[first_layer_bed_temperature]
```
pedrolamas, line 104:
````suggestion
```
START_PRINT EXTRUDER_TEMP=[first_layer_temperature] BED_TEMP=[first_layer_bed_temperature]
```
````

**Review by pedrolamas — CHANGES_REQUESTED** [review]

File: docs/Slicers.md [hunk 4]
```diff
@@ -87,3 +87,33 @@ Klipper's maximum extrusion cross-section check.
...
+
+## START_PRINT macros
+
+When using a START_PRINT macro or similar, it is useful to sometimes
+pass through parameters from the slicer variables to the macro.
+
+In Cura, to pass through temperatures, the following start gcode
+would be used:
+
+START_PRINT BED_TEMP={material_bed_temperature_layer_0} EXTRUDER_TEMP={material_print_temperature_layer_0}
```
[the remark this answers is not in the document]
JamesH1978: ah, did not think about that, not really updated docs before :)

Thanks for the catch
James

docs/Slicers.md
```diff
@@ -87,3 +87,37 @@ Klipper's maximum extrusion cross-section check.
...
+```
+
+In slic3r derivatives such as PrusaSlicer and SuperSlicer, the
+following would be used:
+
+START_PRINT EXTRUDER_TEMP=[first_layer_temperature] BED_TEMP=[first_layer_bed_temperature]
+
+Also note that these slicers will insert their own heating codes when
+certain conditions are not met. in Cura, the existence of the
+material_bed_temperature_layer_0 and material_print_temperature_layer_0
```
pedrolamas, line 110:
```suggestion
`{material_bed_temperature_layer_0}` and `{material_print_temperature_layer_0}`
```

config/sample-macros.cfg
```diff
@@ -10,7 +10,9 @@
 ######################################################################
 
 # Replace the slicer's custom start and end g-code scripts with
-# START_PRINT and END_PRINT.
+# START_PRINT and END_PRINT. Examples for using the variable to
+# parameter pass through in various slicers is documented in
+# docs/Slicers.md
```
KevinOConnor, line 16: FWIW, I'm not sure I understand the new comment.  How about: `See docs/Slicers.md for more information on using these macros.`

docs/Slicers.md
```diff
@@ -87,3 +87,37 @@ Klipper's maximum extrusion cross-section check.
...
+START_PRINT BED_TEMP={material_bed_temperature_layer_0} EXTRUDER_TEMP={material_print_temperature_layer_0}
+```
+
+In slic3r derivatives such as PrusaSlicer and SuperSlicer, the
+following would be used:
+
+START_PRINT EXTRUDER_TEMP=[first_layer_temperature] BED_TEMP=[first_layer_bed_temperature]
+
+Also note that these slicers will insert their own heating codes when
+certain conditions are not met. in Cura, the existence of the
```
KevinOConnor, line 109: "in Cura," -> "In Cura,"

strayr:
I came across this because I was about to add similar for prusaslicer/superslicer

It appears you need a dummy M107 to satisfy prusa/superslicer's start gcode checks, so start gcode looks like 

```
M104 S0
M140 S0
M107
START_PRINT BED_TEMP=[first_layer_bed_temperature] EXTRUDER_TEMP=[first_layer_temperature]
```

I field this issue a LOT on reddit , so it would be cool if it was in the docs.

pedrolamas:
> It appears you need a dummy M107 to satisfy prusa/superslicer's start gcode checks, so start gcode looks like

I have to admit, an empty `M107` looks rather odd... what happens if you don't add it?

JamesH1978:
I have never had to add a dummy M107 to PS or SS, and i have never seen anyone need to add it @strayr , as @pedrolamas says, what happens when you don't add it?

Thanks
James

StuSerious:
Chiming in again, PS

StuSerious:
Chiming in again. 
`M107` should not reside in the start gCode, and arguably not even in the end gcode.
Even then, if for some reason the fan is on at 100% say during a purge line, the next `M106` (that Slic3r derivatives inject at `layer_0`) will overwrite it if `cooling` is enabled and the correct conditions are configured.

Also, PS will go ahead and just prepend and append an `M107` to your gCode file, so there's that. (can't remember for SS, will test and get back)

JamesH1978:
Agree @StuSerious but until we know what issue he thinks he is solving with that code, not sure why you would want any M107 command anywhere other than in the gcode itself where the slicer might turn it off itself

strayr:
Sorry, I've been through and double checked with PS 2.5, 2.6a5 and SS 2.5.59 and can't reproduce the behaviour that this _used to fix_

Sorry for timewasting, I should have checked it was still an issue before jumping on this ticket with enthusiasm.

KevinOConnor:
Thanks.

-Kevin

PR #6030, opened 2023-02-03
Update Measuring_Resonances.md

tyler-cushing:
Typo correction

PR #6064, opened 2023-02-20
doc: use correct mpu9250 i2c_bus on pico

thijstriemstra:
refs https://github.com/Klipper3d/klipper/pull/6030#issuecomment-1416707228 by @Sineos


KevinOConnor:
Thanks.

-Kevin

PR #6065, opened 2023-02-20
logextract: use python3

thijstriemstra:


zellneralex:
You need to provide a sign off line as you did here https://github.com/Klipper3d/klipper/pull/6061

thijstriemstra:
@KevinOConnor are there unit tests for these scripts? If not, can I add one or is the infrastructure to test scripts not there yet?

PR #6068, opened 2023-02-21
Update Config_Reference.md

Thesinofthetin:
Added to clarify that you need to put [multi_pin] config above any other configs you wish to use the pin aliases on.  This stumped me earlier today while setting up multi-pin.


JamesH1978:
please refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially the signed-off-by line.

Thanks

James

PR #6075, opened 2023-02-23
Better running environment compatible with python3.

Yangyuanxin:


JamesH1978:
Unfortunately while this change will work on a pi, where `python` will automatically map to `python3`, this is not the case on other distributions, where `python3` is implicitly needed. That aside, making any change to this code may have consequences with the stock klipper code which is still geared to run on Python2.

Thanks
James

tildebyte:
This is kinda tangential to this specific PR, but I'm running Klipper under elementary OS 6.1 (Jólnir) in a Python 3.8.10 venv, and I've seen zero issues of any kind. Note that I'm a software engineer working primarily with Python, so this is maybe not something which the average end user should try :)

PR #6077, opened 2023-02-25
gcode: Add support for RS-274-D style comments

andysan:
RS-274-D specifies that comments can be enclosed in parentheses. Klipper currently only supports end-of-line comments indicated by a semicolon. Add support for RS-274-D style comments to support more gcode-generating software.

### DIFF
--- a/klippy/gcode.py
+++ b/klippy/gcode.py
@@ -172,13 +172,15 @@ class GCodeDispatch:
         self._respond_state("Ready")
     # Parse input into commands
     args_r = re.compile('([A-Z_]+|[A-Z*/])')
+    comment_r = re.compile('\([^)]*\)')
+    def _strip_comments(self, command):
+        command = command.split(";", 1)[0]
+        return self.comment_r.sub("", command)
     def _process_commands(self, commands, need_ack=True):
         for line in commands:
             # Ignore comments and leading/trailing spaces
             line = origline = line.strip()
-            cpos = line.find(';')
-            if cpos >= 0:
-                line = line[:cpos]
+            line = self._strip_comments(line)
             # Break line into parts and determine command
             parts = self.args_r.split(line.upper())
             numparts = len(parts)


KevinOConnor:
Thanks.  As high-level feedback, I'm not sure I understand the benefit of this change.  The goal for Klipper is to support G-Code such that Klipper can be used with common 3rd party applications (when those applications are in their common configurations).  There's a short blurb on this at https://www.klipper3d.org/G-Codes.html .

So, if we make this change, is there some applications that now works with Klipper, and if so what are those applications?

Cheers,
-Kevin

andysan:
The reason I implemented this patch was to enable support for gcodetools that get shipped with Inkscape. I have been using that to control a plotter that is running Klipper. That's the only specific software I'm aware of at the moment.

KevinOConnor:
Thanks.  As high-level feedback, if making this change and #6078 allows Klipper to work with "inkscape gcodetools" then I think it is okay to go forward with the change.

Can you confirm that those are the only two things blocking Klipper from working with that package?

I'm a little concerned about the performance implications of the change in this PR.  It adds an additional python function call and additional regex on every incoming command.  That adds overhead to the Klipper "fast path".  I wonder if there is an more cpu efficient way to find these comments - do you have some examples of g-code commands using these comments in practice?

-Kevin

andysan:
Yes, I can confirm that this change together with https://github.com/Klipper3d/klipper/pull/6078 (or https://github.com/Klipper3d/klipper/pull/6151) are the only changes needed to use gcodetools out of the box (that's how I'm driving my plotter). It's still necessary to define the right macros to control the pen (M3-M5), but that's just configuration.

When revisiting this change, I realised that there might be a problem if comments are nested. For example how is "foo(bar(baz))" handled? LinuxCNC  seems to treat this as an error. This change would treat it as the command "foo)". This is clearly not ideal.

I don't expect this to be particularly heavy in terms of performance. The regular expression is a true regular expression which means that it should be compiled into a DFA which can be evaluated in O(n) time (assuming a good regex library). This is likely the fastest way to match these comments given that Python's regular expression library seems pretty decent (it's a native module). The only alternative I can think of would be to search for an opening parenthesis and then take a slow path that strips out the comment. This would have the benefit of being able to spot nested comments, but I don't expect it to be faster.

andysan:
I have created an alternative version of this patch here: https://github.com/andysan/klipper/tree/rs274-comments-v2

This version adds some more stringent format checks and doesn't use the re library. I would probably prefer this version since it is more robust when handling malformed gcode.

PR #6078, opened 2023-02-25
RFC: gcode_aliases: Add support for specifying g-code aliases

andysan:
Some g-code dialects use a leading zero in G and M commands. For example, such dialects use G01 instead of G1. Add a mechanism to specify G-Code aliases in configuration files to provide a mechanism to support such dialects. For example, a user could add the following configuration snippet to support such a dialect:

    [gcode_aliases G0X]
    aliases:
        G00=G0, G01=G1, G02=G2, G03=G3, G04=G4

I'm posting this as an RFC. An alternative solution would be to teach Klipper's gcode parser to strip the leading zero from these commands if the specific command hasn't been defined. The aliases mechanism might still be useful though.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1346,6 +1346,19 @@ information.
 
 ## G-Code macros and events
 
+### [gcode_aliases]
+G-Code aliases (one may define any number of sections with a
+"gcode_aliases" prefix). Use this to define aliases for g-code
+commands.
+
+```
+[gcode_aliases my_aliases]
+aliases:
+aliases_<name>:
+#   A comma separated list of "name=value" G-Code aliases to create.
+#   Any number of options starting with "aliases_" may be specified.
+```
+
 ### [gcode_macro]
 
 G-Code macros (one may define any number of sections with a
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  I'm not sure we need an explicit "alias" mechanism as it should be possible to do it already using gcode macros.  Something like:
```
[gcode_macro G01]
    gcode: G1 {rawparams}
```

Stripping leading zeros is also a possibility, but I think we would only do that if it meant we could support a popular 3rd party application that requires it.

Cheers,
-Kevin

andysan:
I was using the macro mechanism, but it got pretty old when multiple G and M commands needed aliases. It's definitely possible, but a bit clunky. The software that I'm using that generates commands with a leading zero is gcodetools in Inkscape. Given that there isn't much use besides stripping leading zeros from commands, I'm tempted to agree that this patch isn't the right solution.

I'd be happy to implement a normalisation step that "normalises" traditional g-code commands if there isn't already a handler installed if we think this would be useful.

KevinOConnor:
Yeah - probably simpler to strip leading zeros.  I'd guess it would be okay to modify `gcode.py:cmd_default()` to check for a legacy g-code command with leading zeros, where there is a registered g-code command without the leading zeros, and then call that command.

Cheers,
-Kevin

PR #6079, opened 2023-02-25
probe: allow overriding horizontal_move_z on gcode

dalegaard:
This adds the ability to override `horizontal_move_z` in gcode commands using the `ProbePointsHelper` utility. This includes Z tilt, QGL, Bed mesh, Bed tilt, Screws tilt adjust, and Delta calibration.

These currently always use a static `horizontal_move_z`. This patch allows overriding this per-call with a `MOVE_Z` gcode parameter. The configured value is always used unless `MOVE_Z` is given for a specific command(the value does not "stick around").

The main use case is the ability to perform a "two stage" QGL or Z tilt adjust. One can (e.g. from a macro) perform a high-z-move high-tolerance run first(to guard against a heavily misaligned bed), and then a low-move, low-tolerance run after. This can significantly speed up the process, especially when using a probe like Beacon.

Best regards,
Lasse

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -146,14 +146,15 @@ The following commands are available when the
 (also see the [bed mesh guide](Bed_Mesh.md)).
 
 #### BED_MESH_CALIBRATE
-`BED_MESH_CALIBRATE [METHOD=manual] [<probe_parameter>=<value>]
-[<mesh_parameter>=<value>]`: This command probes the bed using
-generated points specified by the parameters in the config. After
-probing, a mesh is generated and z-movement is adjusted according to
-the mesh. See the PROBE command for details on the optional probe
-parameters. If METHOD=manual is specified then the manual probing tool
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I think it is fine to support overriding horizontal_move_z on these commands.  For what it is worth though, I'm not sure about using `MOVE_Z` as the parameter name.  Seems like it could be confused with a request to move a distance in the Z.  Perhaps the more verbose `HORIZONTAL_MOVE_Z` would be more clear.

-Kevin

dalegaard:
Hi @KevinOConnor 

I've force pushed with that change.

Best regards,
Lasse

kmobs:
Can confirm this is work well. A few other folks are running it without difficulty as well. 

Created a z_tilt_scan macro to override the default z tilt. 

https://youtu.be/S3t_dQoVY8w

KevinOConnor:
Thanks.

-Kevin

PR #6076, opened 2023-02-23
TMC2240 support

leptun:
This PR is based on top of #6074. As such, many unrelated commits can be seen in the diff. Only the tmc2240 tagged commits (the last one) are important. I have marked the PR as "draft" until the prerequisite PRs are merged.

The aim of this PR is to add support for the tmc2240 drivers. The aim was to keep the code as configurable as possible considering there isn't a lot of electronics available for the time being and things are still in development. As for already released electronics that can make use of this PR, there is the BTT SB2240. My testing setup uses TMC2240-EVAL boards.

PR #6080, opened 2023-03-01
tmc2240 driver

KevinOConnor:
This is a modification of PR #6076 (from @leptun).  The main change in this PR is that it is rebased to the master branch and is not dependent on other TMC changes.

As with that PR, this PR adds basic support for the Trinamic TMC2240 driver.

Various changes in this PR:
1. Rebase'd to the master branch without coolstep and other changes not related to tmc2240.
2. I reworked the RMS current calculation code.  It is no longer necessary to specify a `current_range` as that value can be calculated from the requested `run_current`.
3. I set the default PWM_OFS to 29 (instead of 30) as the tmc2240 defaults to 29.
4. I added the tmc2240 to the test/klippy/tmc.test regression test case.

I have tested basic functionality on a BTT EBB SB2240 board.  I also tested basic sensorless homing functionality on that board.

Comments?
-Kevin


KevinOConnor:
I should add that this PR (along with the PR from #6076) only adds support for the tmc2240 via SPI.  It seems this driver could be wired for "TMC UART" support.  That is, if these chips become popular, it is possible that some boards will use tmcuart while others use spi.  I don't think it would be difficult to update the driver to support both modes should we find boards wired for tmcuart.

(Just for reference, the tmc5160 chips also theoretically support both spi and tmcuart, but those chips don't support "step+dir" mode when using tmcuart and are thus useless in tmcuart mode.)

-Kevin

leptun:
👍 I will give the PR a try this evening on my printer. I will check that the basic homing still works, but I'm sure it will be worse than before.

About current_range, I'm not 100% sure that's a good idea. I noticed that the endstop phase would be different if I increased current_range from 1 to 3. This happened even though I had the same homing_current. So with your code, homing might be unintentionally influenced by the run_current, even when it shouldn't be. This won't be a problem if changing the current also recalculates current_range. I didn't check the new code carefully enough to come to a conclusion on that part.

Good catch with PWM_OFS. Didn't see that it changed.
Thank you also for implementing the tests.

As for uart mode, I didn't have a reason to implement it so far since I only used multiple drivers via SPI. The UART on this driver is implemented differently than on tmc220x. It still has address pins, but it also has the chaning support. It needs to be seen what needs to be implemented and how on klipper's side. It seems like that requires a bit of initialization to be done in order for each driver to get an address.

klippy/extras/tmc2240.py
```diff
@@ -0,0 +1,399 @@
...
+    "s2vsb":            (lambda v: "1(ShortToSupply_B!)" if v else ""),
+    "adc_temp":         (lambda v: "0x%04x(%.1fC)" % (v, ((v - 2038) / 7.7))),
+})
+
+
+######################################################################
+# TMC stepper current config helper
+######################################################################
+
+MAX_CURRENT = 3.000
```
leptun, line 269: I don't think this is correct. This is peak max current, while klipper uses RMS max current in all other places. The datasheet says the max RMS current is 2.1A, which is true when using Rref=12K and current_range=3, but a higher Rref will reduce the possible max current. So I think this value should also be calculated dynamically.

KevinOConnor:
Thanks.  I updated the code to recalculate the max_current based on the rref setting.  (And I squashed and rebase'd this branch with that change.)

I'm not sure I understand the issue with endstop phases - the only requirement is that the same phase is found with each homing attempt when using identical driver settings.  If the driver settings are changed I would not be surprised that one would find a different endstop_phase - it is only an issue if one obtains an unreliable endstop_phase with the same driver settings.

Cheers,
-Kevin

leptun:
> the only requirement is that the same phase is found with each homing attempt when using identical driver settings. If the driver settings are changed I would not be surprised that one would find a different endstop_phase

Ok, I thought that is going to be an issue. The endstop phase is still reliable with a different current_range, but it does result in a different value. I just found it weird that I had to recalibrate it, but I guess it's fine.

leptun (approved):
Switched to this branch, commented out the unimplemented features in my config from the other branch, reduced the SGT value a bit because of the lack of homing_current and printed an XYZ calibration cube. So the code still works on X and Y on my printer.

klippy/extras/tmc2240.py
```diff
@@ -0,0 +1,397 @@
...
+import math, logging
+from . import bus, tmc, tmc2130
+
+TMC_FREQUENCY=12500000.
+
+Registers = {
+    "GCONF":            0x00,
+    "GSTAT":            0x01,
+    "IFCNT":            0x02,
+    "SLAVECONF":        0x03,
```
leptun, line 16: This register was recently renamed by Trinamic in the rev1 tmc2240 datasheet to `NODECONF`. It seems they renamed all "slave" terms in the datasheet to "node". Should we update this register name as well or not? The other TMC drivers still call it `SLAVECONF` in the datasheet, but that might just be because they haven't updated the documents yet.

KevinOConnor:
> This register was recently renamed by Trinamic in the rev1 tmc2240 datasheet to NODECONF

Interesting.  The spec I downloaded just a few weeks ago still had SLAVECONF.  In any case, no harm in changing it to the new name.  (The name should only be used in DUMP_TMC.)  I updated this PR accordingly.

> I hope this can be merged soon so I can rebase the TMC other changes on top of the latest master branch.

Thanks.  I'll give a couple more days to see if there are other comments.  Otherwise I will look to commit.

-Kevin

PR #6082, opened 2023-03-03
probe: expose name on status report

pedrolamas:
Adds the probe name to the probe status object.

I know one can infer this from the config file (check for "bltouch", "smart_effector" or "probe" sections), but seems to me easier to just expose it here.


dalegaard:
This seems like a nice quality of life change for frontends. 👍 

Best regards,
Lasse

KevinOConnor:
Thanks.

-Kevin

PR #6084, opened 2023-03-04
config: Add Anycubic Kobra Plus

dkaser:
This adds support for the Anycubic Kobra Plus, based on the HC32F460 microcontroller.


KevinOConnor:
Thanks.  In general it seems fine to me.  The config file should use `:` instead of `=` syntax though - see https://www.klipper3d.org/Example_Configs.html point 7d.

Also, I'm a little leery of listing manufacturer product names in the test scripts.  Perhaps use a filename of `test/configs/hc32f460-serial-PA3PA2.config` (or something similar).

-Kevin

JamesH1978 (commented):
unfortunately I do not think a configuration that recommends a hard modification to the print board is a good fit for klipper, we like to make sure people have the choice to return to stock, so i am not sure this modification is one the average user will want to perform or have to reverse. @KevinOConnor can you weigh in on this one?

Thanks
James

dkaser:
I made the requested updates to the files.

As for the hard modification, I agree that it's not ideal, but as long as users know the modification and its consequences (which are described in the comments to the config file) I think it's OK to give them the choice if they want to do it or not.

JamesH1978:
fair enough then :) i will review

Thanks
James

JamesH1978 (approved):
Looks fine to me,

Thanks
James

KevinOConnor:
Thanks.

-Kevin

PR #6085, opened 2023-03-04
tmc: SET_TMC_FIELD set VELOCITY

leptun:
This PR introduces the argument `VELOCITY` for `SET_TMC_FIELD`. It can be used instead of `VALUE` so that the velocity is automatically converted to the TSTEP based value of that velocity.

The aim of this change is to allow macros to change fields such as `TPWMTHRS`, `TCOOLTHRS` and `THIGH` at runtime without needing to hardcode values which might differ based on microstepping resolution and steps/mm of the axis.
To implement this change, I ported the TSTEP helper from my other PR (#6074).

Tested on TMC2240 on another branch of mine. TMC2660 needs to be tested since it's the odd one out.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1236,13 +1236,16 @@ if StealthChop2 is used, the stepper must be held at standstill for >130ms so
 that the driver executes the AT#1 calibration.
 
 #### SET_TMC_FIELD
-`SET_TMC_FIELD STEPPER=<name> FIELD=<field> VALUE=<value>`: This will
-alter the value of the specified register field of the TMC driver.
+`SET_TMC_FIELD STEPPER=<name> FIELD=<field> VALUE=<value> VELOCITY=<value>`:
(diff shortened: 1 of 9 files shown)


leptun:
@KevinOConnor I've attempted to rewrite the code as requested. Is it ok now?

KevinOConnor:
Thanks.  Looks okay to me.  I'll give a few more days for comments and otherwise look to commit.

FWIW, separately, a nice refactor may be to add a `get_tmc_frequency()` method to `MCU_TMC2660_SPI`, `MCU_TMC_uart`, and `MCU_TMC_SPI` classes.  Thus avoiding the need to manually propagate that core value.

-Kevin

leptun:
@KevinOConnor Rebased onto the latest master branch and refactored the code as suggested. Please recheck. I tested this on my printer and it seems to still be working fine with TMC2130.

KevinOConnor:
Thanks.

-Kevin

PR #6086, opened 2023-03-04
tmc: Fix s2vs field formatters

leptun:
Fixed the s2vs field formatters. So instead of the misleading `LowSideShort_A`, it's now `ShortToSupply_A` (and the same for the B coil). Also implemented this field formatter for TMC5160 since it was missing there.

Based on PR #6074.

### DIFF
--- a/docs/TMC_Drivers.md
+++ b/docs/TMC_Drivers.md
@@ -544,7 +544,7 @@ hot. Typical solutions are to decrease the stepper motor current,
 increase cooling on the stepper motor driver, and/or increase cooling
 on the stepper motor.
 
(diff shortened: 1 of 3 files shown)


Sineos:
What does it effectively mean? 0 Ohm on the given coil?

leptun:
It means that an overcurrent condition was triggered when the driver tried to pull the coil to GND with the mosfets. That is usually an indicator that the coil is shorted to the 12/24V rail.

Sineos:
Thanks for the explanation 👍

KevinOConnor:
Thanks.

-Kevin

PR #6088, opened 2023-03-04
tmc: Beautify initialization code

leptun:
Purely esthetic. Make all tmc driver init code similar to one another. The various fields were grouped based on the register which they affect or the function they perform.

Based on PR https://github.com/Klipper3d/klipper/pull/6074.


KevinOConnor:
Thanks.

-Kevin

PR #6091, opened 2023-03-05
WIP: Support broken TMC boards on the Anycubic Kobra

Jookia:
**This is not intended for merging but for discussion for now.**

The Trigorilla v1.0.4 used in the Kobra series (Anycubic Kobra, Anycubic Kobra Plus, Anycubic Kobra Max) has a serious hardware bug: It looks like last minute Anycubic swapped the extruder TMC2209 for a TMC2208 but didn't update the address pins on the board. This means the extruder and X axis both use address 0. This creates an interesting situation where we can write to the controller registers but reading will cause a bus conflict. 

Currently Klipper insists on reading registers from the controller to confirm writing was successful and to check status when there's an error driving it. Both of these aren't entirely necessary for the printer to work, which is why Marlin works on this board.

The controllers will always stop on error or host overload on error, and writing to registers seems fairly reliable in practice. Furthermore the only time we actually need to write to registers is for sensorless homing which happens once during the printer's on cycle. All other activity is controlled using pins not conflicting.

The only way to currently run Klipper on these boards is with microsoldering, which isn't accessible to most people, and as these boards are in active production and used in actual printing it seems like a good idea to add some kind of workaround so people can use Klipper.

This PR introduces the concept of a 'tmchax' controller. This controller does two things: Refuses to read from registers, and controls two controllers at once as if it's one controller with multiple outputs. This 'we pretend two controllers are one' ensures that coherency between registers is enforced.

I have tested this on my printer and aided someone setting it up on theirs and from what I can see this works fine in practice, aside from the code breaking due to use of a global variable. I wrote this code in a few hours so it's not intended to be a solution, but a proof of concept. I have included my printer config as 'kobra_xhack.cfg'. If you want to test this, see my hack branch named 'tmc2209f' here with a README and full details: https://github.com/Jookia/klipper/tree/tmc2209f

---

Code wise, I'm not sure where to go with this:

- It doesn't seem like Klipper supports the idea of 'one controller two motors' which I believe is necessary to enforce some type of coherency
- This solution only works because the TMC2209 and TMC2208 are register compatible
- If two TMC2209s are ever on the same bus address the same fix could be used but locking registers would be required for the sensorless homing
- I'm duplicating a bit of code and hand selecting which registers are usable
- I've removed the SET_TMC_REGISTER g-code but that should be added back for calibrating sensorless homing
- I try writing the register 5 times. This is 4 more times than Marlin and seems to work fine
- The code crashes on Klipper reset due to the global variable or something

I also haven't investigated what actually happens on a bus conflict and if we can gain any knowledge from this. Probably nothing: If both controllers are driving the bus and giving opposing answers then it will get decoded as a wrong value that we can't check.

### DIFF
--- a/klippy/extras/tmchax.py
+++ b/klippy/extras/tmchax.py
@@ -1,0 +1,193 @@
+# TMCHAX frankensteined configuration
+#
+# Copyright (C) 2019  Stephan Oelze <stephan.oelze@gmail.com>
+# Copyright (C) 2023  Jookia <contact@jookia.org>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+from . import tmc2208, tmc2130, tmc, tmc_uart
+import logging
+
+TMC_FREQUENCY=12000000.
+
+Registers = {
+    "GCONF": 0x00, "IFCNT": 0x02, "SLAVECONF": 0x03, "FACTORY_CONF": 0x07,
+    "IHOLD_IRUN": 0x10, "TPOWERDOWN": 0x11, "TSTEP": 0x12, "TPWMTHRS": 0x13,
+    "TCOOLTHRS": 0x14, "VACTUAL": 0x22, "COOLCONF": 0x42, "SGTHRS": 0x40,
+    "CHOPCONF": 0x6c, "PWMCONF": 0x70
+}
+
+Fields = {}
+
+Fields["GCONF"] = {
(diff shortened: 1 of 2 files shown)


Jookia:
I'm closing this PR for now and instead maintaining my fork. I'll come back to this if there's more than just me interested in it. :)

PR #6092, opened 2023-03-06
stm32h7: Restore the RCC set in the bootloader to the default value

bigtreetech:
For example: the Octopus Max EZ bootloader has enabled SPI4


KevinOConnor:
Thanks.

-Kevin

PR #6093, opened 2023-03-06
tmc: calculate max_current on all drivers

leptun:
Just like with tmc2240, compute the maximum RMS current from the sense resistor and limit run_current and hold_current to that value. This way the user gets an error if their config has too high of a current instead of just maxing out IRUN and IHOLD.

Tested on tmc2130 with `sense_resistor=0.22` and `run_current: 2.0`: `Option 'run_current' in section 'tmc2130 stepper_z' must have maximum of 0.957540432857`

### DIFF
--- a/klippy/extras/tmc2130.py
+++ b/klippy/extras/tmc2130.py
@@ -116,20 +116,20 @@ FieldFormatters = {
 # TMC stepper current config helper
 ######################################################################
 
-MAX_CURRENT = 2.000
-
 class TMCCurrentHelper:
-    def __init__(self, config, mcu_tmc):
+    def __init__(self, config, mcu_tmc, sense_resistor_offset):
         self.printer = config.get_printer()
         self.name = config.get_name().split()[-1]
         self.mcu_tmc = mcu_tmc
         self.fields = mcu_tmc.get_fields()
-        run_current = config.getfloat('run_current',
-                                      above=0., maxval=MAX_CURRENT)
-        hold_current = config.getfloat('hold_current', MAX_CURRENT,
-                                       above=0., maxval=MAX_CURRENT)
-        self.req_hold_current = hold_current
         self.sense_resistor = config.getfloat('sense_resistor', 0.110, above=0.)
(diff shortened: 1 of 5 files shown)


KevinOConnor:
Thanks.  I'm not sure what the high-level benefit of this change is though.  The main reason for specifying a maximum current was to alert users if they used milliamps (eg, `2000`) instead of amps.  Otherwise, I don't see a harm in trying to meet the user request as best as we can.

Cheers,
-Kevin

leptun:
Ok. In that case I can drop the PR. The reasons for this were both for alerting the user when using too high of a current and for moving the change from tmc2240 to the other drivers as well.
Imo, there should at least be some warning if the current is clamped, both for the config and for `SET_TMC_CURRENT`

kmobs:
I like this change as it ensures the user's requested current is what's actually set rather than just having a hard cap defined.

leptun:
@KevinOConnor Is there anything I can do in this PR to get it accepted?

vkhodygo:
@KevinOConnor I hope you're doing well. Could you please re-open this PR and assign someone to review the changes?

leptun:
@Arksine I like this idea. Don't have energy to implement it at the moment though.

KevinOConnor:
One of the challenges is that the current code only ever finds an estimate for the requested current.  So, for example, the user may request a run_current of 1.2 amps, but the code may select 1.196642 amps.  If the user select 1.75 amps, they may get 1.740571 amps.

It seems to me it would be very awkward if a user requested a current of 1.75 amps and they got a warning or error stating - "sorry, but 1.740571 is the highest value".

So, I think we should raise an error on a clearly bogus value, but I'm not sure a simple check for maximum will catch a "clearly bogus value" from a reasonable request "for the closest value to what was requested".

The current code does check for clear outliers - anything over 2 amps is rejected by the tmc2130/tmc22xx chips, 2.4 for tmc2660, and 3 amps for tmc5160.  So, there are already some checks for "clearly bogus values".

Cheers,
-Kevin

PR #6094, opened 2023-03-06
stm32f7: add support for stm32f7 and remram board

fredericmorin:
Hi klipper community, 
Hi Kevin,

He's a PR to add support for STM32F7 uC and [remram](https://github.com/hasenbanck/remram) controller board to klipper.

I've exercise all of the remram board IOs:
- 4 tmc2130 with stepper motor
- fan1, fan2 with multimeter
- temp1, temp2, temp3 with NTC 100k
- heater with multimeter
- bed out with multimeter

Had to add a patch to console.py for OSX terminal support. 

I've left a bunch of code comments in case thats usefull. Please mark as resolved if theres no action to take.

Looking forward to code review

Fred.

---

Old PR content: Issues with DWT counter stuck at 0. Now resolved.


KevinOConnor:
I'm not sure what the issue is.  It looks like the DWT counter isn't being enabled.  It's not a USB or timing issue, as you would not get through the connection handshake process if those areas were not working correctly.

The console.py code is complaining because of a negative integer in a field expecting a positive integer.  You can likely work around it temporarily by something like:
```diff
--- a/klippy/serialhdl.py
+++ b/klippy/serialhdl.py
@@ -203,7 +203,7 @@ class SerialReader:
             self.ffi_lib.serialqueue_free)
     def set_clock_est(self, freq, conv_time, conv_clock, last_clock):
         self.ffi_lib.serialqueue_set_clock_est(
-            self.serialqueue, freq, conv_time, conv_clock, last_clock)
+            self.serialqueue, freq, conv_time, max(0, conv_clock), last_clock)
     def disconnect(self):
         if self.serialqueue is not None:
             self.ffi_lib.serialqueue_exit(self.serialqueue)
```

As for the DWT counter - I'm not sure why it would not be enabled.  You might want to double check the low-level power and initialization sequence on the F7.  If the board has a bootloader, it's also possible that bootloader code may be altering some low-level registers that later interfere with Klipper.  Finally, if you are connecting via SWD or JTAG, it's possible those debugger interfaces are altering the DWT handling.

FWIW, it seems you've come quite far along in board bring up, as it seems only the DWT counter is not functioning properly.

-Kevin

fredericmorin:
Ah! I think I found it. Seems the stm32f7 has its DWT register locked by default.

Here's what Marlin is doing to enable it: https://github.com/MarlinFirmware/Marlin/blob/2d162e0639f74b3a4e66e8c80081622e3215fe9d/Marlin/src/HAL/shared/Delay.cpp#L90-L92

Cross confirmation: https://stackoverflow.com/a/37345912

I'll try that tonight. Thanks for the pointers!

fredericmorin:
allright, got it to boot up without any hack. Took me longer than it should have cuz I was unlocking the wrong subsystem. 
`DWT->LAR = 0xC5ACCE55;` was all that was missing. 

Got a stepper to home and move with stealthchop enabled. will continue exercising all IOs before submitting final PR.

src/stm32/stm32f0_i2c.c
```diff
@@ -27,6 +27,9 @@ struct i2c_info {
   DECL_CONSTANT_STR("BUS_PINS_i2c1", "PB6,PB7");
   DECL_ENUMERATION("i2c_bus", "i2c1a", 1);
   DECL_CONSTANT_STR("BUS_PINS_i2c1a", "PF1,PF0");
+#elif CONFIG_MACH_STM32F7
+  DECL_ENUMERATION("i2c_bus", "i2c1_PB6_PB7", 0);
```
fredericmorin, line 31: did not test i2c to work on the hardware I got.
only added minmum config to match remram pinout

KevinOConnor:
What's the state of this PR?  Is it ready for a review (and ready to be merged into Klipper)?

I did notice that this PR has changes to gitignore and build rule changes.  It would be important for those types of changes not be included in functional changes (so that we can use tools like `git revert` and `git bisect`).  In general, though, we would not modify the main build or gitignore rules though - typically users would be expected to modify their own local gitignore rules for their particular editor of choice.

-Kevin

scripts/flash_usb.py
```diff
@@ -340,13 +340,22 @@ def flash_rp2040(options, binfile):
...
+    'same70': flash_atsam4,
+    'stm32f042': flash_stm32f4,
+    'stm32f072': flash_stm32f4,
+    'stm32f103': flash_stm32f1,
+    'stm32f4': flash_stm32f4,
+    'stm32f7': flash_stm32f4,
+    'stm32g0b1': flash_stm32f4,
+    'stm32g4': flash_stm32f4,
+    'stm32h7': flash_stm32f4,
+    'stm32l4': flash_stm32f4,
```
KevinOConnor, line 358: Please don't mix whitespace changes with functional changes.  It makes review harder and complicates merging and use of tools like `git revert`.

src/stm32/Kconfig
```diff
@@ -314,6 +322,8 @@ choice
         bool "16 MHz crystal"
     config STM32_CLOCK_REF_20M
         bool "20 MHz crystal"
+    config STM32_CLOCK_REF_24M
+        bool "24 MHz crystal"
```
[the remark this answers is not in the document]
KevinOConnor: Seems fine to me.

src/stm32/dfu_reboot.c
```diff
@@ -36,7 +36,7 @@ dfu_reboot(void)
     irq_disable();
     uint64_t *bflag = (void*)USB_BOOT_FLAG_ADDR;
     *bflag = USB_BOOT_FLAG;
-#if CONFIG_MACH_STM32H7
+#if __CORTEX_M >= 7
```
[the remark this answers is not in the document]
KevinOConnor: Seems fine to me.

KevinOConnor (commented):
Thanks.  In general the stm32f7 support looks fine to me.  However, I think the non-stm32f7 changes should not be separated out from this PR.  See my comments below.

-Kevin

fredericmorin:
Allright, removed all changes to the .dockerfile, .vscode, .gitignore, and console.py files and cleaned up flash_usb.py.

Are any of those changes interesting enough to your project to be worth a few more PRs to include? They made working on your project from other platform much more pleasing. It's ok if not too. Ill keep those in a branch of mine in case I have to seriously work on klipper code again.

config/generic-remram.cfg
```diff
@@ -0,0 +1,118 @@
...
+heater_pin: PA3  # HEATER
+sensor_type: Generic 3950
+sensor_pin: PC1  # TEMP2
+control: pid
+pid_Kp: 22.2
+pid_Ki: 1.08
+pid_Kd: 114
+min_temp: 0
+max_temp: 250
+min_extrude_temp: 0
```
KevinOConnor, line 87: Don't disable default safety systems in example configs - see https://www.klipper3d.org/Example_Configs.html (item 9)

src/generic/armcm_timer.c
```diff
@@ -95,6 +95,9 @@ timer_init(void)
 {
     // Enable Debug Watchpoint and Trace (DWT) for its 32bit timer
     CoreDebug->DEMCR |= CoreDebug_DEMCR_TRCENA_Msk;
+#if CONFIG_MACH_STM32F7
+    DWT->LAR = 0xC5ACCE55;
+#endif
```
KevinOConnor, line 100: Can this be moved into stm32f7.c instead?  That is, add something like the following to `armcm_main()`:
```c
   CoreDebug->DEMCR |= CoreDebug_DEMCR_TRCENA_Msk;
   DWT->LAR = 0xC5ACCE55;
```

I'm a little leery of adding architecture specific code to the generic code.

src/stm32/gpioperiph.c
```diff
@@ -31,6 +31,7 @@ gpio_peripheral(uint32_t gpio, uint32_t mode, int pullup)
     // stm32f0 is ~10Mhz at 50pF
     // stm32f2 is ~25Mhz at 40pF
     // stm32f4 is ~50Mhz at 40pF
+    // stm32f7 is ~XXMhz at XXpF
```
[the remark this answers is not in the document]
KevinOConnor: It's taken from the datasheet for the device - see section 6.3.20.  An OSPEEDR=2 setting supports up to 50Mhz (at 40pF) - same as the stm32f4.

KevinOConnor (commented):
It seems github lost some of my previous review comments - sorry about that.  There are a few minor changes that I think would be worth looking at before merging.  Otherwise, it looks fine to me.

Thanks,
-Kevin

src/generic/armcm_timer.c
fredericmorin: I would argue this is better as-is. The DWT register unlock is needed for DWT->CTRL access one line below. Setting that register out of context would make this harder than needed to track the problem.

Marlin does that too: https://github.com/MarlinFirmware/Marlin/blob/bugfix-2.1.x/Marlin/src/HAL/shared/Delay.cpp#L90-L92
Betaflight (drone flight controller) does that as well: https://github.com/betaflight/betaflight/blob/master/src/main/drivers/system.c#L79-L89

KevinOConnor: The convention is to not put chip specific code in the `generic/` directory (`grep CONFIG_MACH_ src/generic/*`).  I'd prefer to keep that convention as I fear the alternative would quickly devolve into a mess of ifdefs in the "generic" code.

Possible alternatives are:
1. Place the two line override in stm32f7.c (as proposed earlier).
2. Copy generic/armcm_timer.c to a new stm32/stm32f7_timer.c .  That's a lot of dup'ed code though and seems not worth it.
3. Use an architecture specific ifdef in armcm_timer.c (eg, `if (__CORTEX_M >= 7)`).  Unfortunately, this risks introducing a regression on other chips, and if a regression is found a revert would require reverting all of stm32f7.c .  So, seems like putting the code in stm32f7.c is less risky (even if only for the short-term).

-Kevin

fredericmorin: Allright, moved it to stm32f7.c

KevinOConnor:
Thanks.  I went to commit this, but I don't see a signed-off-by line.  See https://www.klipper3d.org/CONTRIBUTING.html .  Contributions to Klipper require the Signed-off-by line - it indicates agreement with https://github.com/Klipper3d/klipper/blob/master/docs/developer-certificate-of-origin

-Kevin

fredericmorin:
Allright, took a little bit to figure out but got it verified. 1password git signing integration needed a little convincing to get going. 
Hope everything is in order now!

fredericmorin:
Ah, I mixed up "signed-off by" and "verfied". 
Well... got both now.

KevinOConnor:
Thanks!

-Kevin

PR #6095, opened 2023-03-06
Richardjm/probe discard first

richardjm:
Supports the ability to discard the first n probes. Allowing a faster first probe before accurate subsequent measurements.

There are many users implementing this mechanism via manually replacing the probe.py file with a version that supports discarding the first probe.

- Previous PR was marked as inactive even though no decision had been made to exclude it - https://github.com/Klipper3d/klipper/pull/5969

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1817,6 +1817,10 @@ z_offset:
 #   samples. If this tolerance is exceeded then either an error is
 #   reported or the attempt is restarted (see
 #   samples_tolerance_retries). The default is 0.100mm.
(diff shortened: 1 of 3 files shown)


jakep82:
I thought Kevin's original response was pretty unambiguous. This seems like a workaround, not a solution to the actual issue.

KevinOConnor:
I think it would be better to move this to a thread on Klipper Discourse.  As mentioned previously, I'm not comfortable making code changes without an understanding of the root cause.  So, I think the forum is a better place for discussions.

-Kevin

PR #6096, opened 2023-03-07
Docs: prefer Marlin (legacy) over Marlin 2 where available

strayr:
Marlin 2 option in prusaslicer produces gcode with an M204 that Klipper doesn't interpret correctly

### DIFF
--- a/docs/Slicers.md
+++ b/docs/Slicers.md
@@ -7,8 +7,10 @@ Slic3r, Cura, Simplify3D, etc.
 ## Set the G-Code flavor to Marlin
 
 Many slicers have an option to configure the "G-Code flavor". The
-default is frequently "Marlin" and that works well with Klipper. The
-"Smoothieware" setting also works well with Klipper.
+default is frequently "Marlin" and that works well with Klipper. Note that
+some slicers that make optimisations for Marlin 2 may work better with the
+"Marlin (legacy)" option. The "Smoothieware" setting also works well with
+Klipper.
 
 ## Klipper gcode_macro
 


JamesH1978:

Thanks
James

Sineos:
https://github.com/prusa3d/PrusaSlicer/issues/3452#issuecomment-1453305678

There is hope 😄

KevinOConnor:
Thanks.  I don't really have an opinion on this change.  Maybe @JamesH1978 or @Sineos can provide their feedback.

For what it is worth, it seems kinda confusing to say "use marlin or smoothieware" - wouldn't just saying "use marlin" be simpler for users?

-Kevin

strayr:
I don't know the history of why that was originally chosen. I think it's there because smoothieware turns off machine limits sections in prusa slicer, at which point if there's a good use case for it remaining then we should explain why rather than give users a blind choice.

More invasive to as it changes a few lines of toolhead.py but how would you feel @KevinOConnor about handling prusaslicer's marlin 2 output with lone `M204 P` without an `M204 T` by setting the P value as acceleration? It would take a slightly cryptic step out of setting up slicer profiles for Klipper and make klipper a little more slicer agnostic. Yes, that could be done in a macro, but that would be a much bigger user task than setting up a slicer profile and this behaviour is most painful getting up and running. I don't know if that would break output from another slicer or if there's another reason why we don't do this?

I'm all for further simplifying the docs and going with the following, particularly as there's a chance prusaslicer will have some level of klipper support soon.

> Many slicers have an option to configure the "G-Code flavor". Choose Klipper if available, otherwise  choose Marlin, preferring Marlin (legacy) over Marlin 2

or even 

> If you can set the G-Code flavour, choose Klipper or Marlin (legacy) 

and hope users are smart enough to figure out that Marlin is a good enough match if not given the choice.

I went for the change proposed as it was the least invasive change that could make a positive difference. I'm happy to do more if its a BETTER solution.

strayr:
https://github.com/prusa3d/PrusaSlicer/releases/tag/version_2.6.0-alpha6

Yeah, this changes quite a bit.

Piezoid:
I accidentally made `M204 P/T` work while moving the acceleration state from `toolhead.py` to `gcode_move.py`. This makes acceleration behave more like the gcode velocity.
It's a [major change](https://github.com/Piezoid/klipper/commit/514cefe94368316c388dfce78976ad481943bff2#diff-7208b76dbffe63e7222ee0bc885f5b69fc9c82a709c81f9e9927ff4cf0cde52fR139-R191) that will have implications for numerous extra modules, but I can tidy it up for upstream if you're keen on this approach.

PR #6097, opened 2023-03-07
src/linux/console.c: fix arguments as read-only

guillaume-pais:
Some arguments are not marked as read-only (const) in ```console.c``` and ```internal.h``` for the functions ```report_errno()```, ```console_setup()``` and ```console_sleep()```

### DIFF
--- a/src/linux/console.c
+++ b/src/linux/console.c
@@ -25,7 +25,7 @@ static struct pollfd main_pfd[1];
 
 // Report 'errno' in a message written to stderr
 void
-report_errno(char *where, int rc)
+report_errno(const char *where, int rc)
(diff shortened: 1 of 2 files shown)


JamesH1978:
thanks for submitting this PR, please refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and particular attention to the signed-off-by line.

Thanks
James

KevinOConnor:
Thanks.  As high-level feedback, I'm not sure code "constification" is a worthwhile process.  I fear it leads to "code churn" and doesn't prevent many bugs in practice.

-Kevin

PR #6099, opened 2023-03-08
Add support for Nation N32G45x mcu

stas2z:
N32G452/G455 are mostly compatible with STM32F103 but have M4 core and different ADC.
So it compile as f103 and working except adc readings. This PR adds N32 ADC support as an option for STM32F103.
Also it twice faster than f103 as it can run at 144Mhz but without modifying common klipper stm32 rcc init code it can safely run at 128MHz only (seems like it requires flash latency 4, but not tested).
This PR uses some code from N32G45x BSP (seems like it's a stdperiph with renamed fields, vars, functions etc, so it's not conflict with stm32 defs), all copyrights given.

Also this PR adds config for Voxelab Aquila N32 board 1.0.1 - first printer shipped with this MCU, but there are some other printers with it like different Aquila models and some other Voxelab/FlashForge devices.

Signed-off-by: Alexey Golyshin <stas2z@gmail.com>


lib/README
```diff
@@ -160,3 +160,7 @@ revision 177b0073fe6f19281ee7f7fdbe9599e32d1b4b8b.
 The Huada HC32F460 directory contains code from:
   https://www.hdsc.com.cn/Category83-1490
 version 2.2 DDL minus example directory, empty/extra files
+
+The n32g45x directory contains parts of code from:
+  https://github.com/RT-Thread/rt-thread/tree/master/bsp/n32g452xx/Libraries/N32_Std_Driver
+version v1.0.1
```
KevinOConnor, line 166: For sources from github, please include the full revision (as seen in other github references in this README file).

src/stm32/Kconfig
```diff
@@ -244,6 +245,15 @@ config STM32_DFU_ROM_ADDRESS
...
 
 
+config N32G452
+    bool "Build for Nation N32G45x MCU"
+    depends on MACH_STM32F103 && LOW_LEVEL_OPTIONS
+    default n
+    help
+        Some Voxelab Aquila motherboards have N32G452 chip instead of
+        GD32F103. This cortex-M4 chip is very similar to F103 but
+        have different ADC. Enable this option to make ADC working.
```
KevinOConnor, line 255: I think it may be preferable to add a new MACH_N32G452 definition to the main "Processor model" menu.  If there are places in Kconfig or the code that are using CONFIG_MACH_STM32F103 where they should be using CONFIG_MACH_STM32F1 then I think we should just update those references.  That way, the MACH_N32G452 could be a MACH_STM32F1 but not a MACH_STM32F103.

src/stm32/Makefile
```diff
@@ -17,7 +17,12 @@ MCU := $(shell echo $(CONFIG_MCU))
 MCU_UPPER := $(shell echo $(CONFIG_MCU) | tr a-z A-Z | tr X x)
 
 CFLAGS-$(CONFIG_MACH_STM32F0) += -mcpu=cortex-m0 -Ilib/stm32f0/include
-CFLAGS-$(CONFIG_MACH_STM32F1) += -mcpu=cortex-m3 -Ilib/stm32f1/include
+ifeq ($(CONFIG_N32G452), y)
+	dirs-$(CONFIG_MACH_STM32F1) += lib/n32g45x
+	CFLAGS-$(CONFIG_MACH_STM32F1) += -mcpu=cortex-m4 -Ilib/stm32f1/include -Ilib/n32g45x/include
+else
+	CFLAGS-$(CONFIG_MACH_STM32F1) += -mcpu=cortex-m3 -Ilib/stm32f1/include
+endif
```
KevinOConnor, line 25: I think this should be turned into lists.  Ifdefs in the makefile rapidly become unmaintainable.  Something like:
```
dirs-$(CONFIG_MACH_N32G452) += lib/n32g45x
```

src/stm32/Makefile
```diff
@@ -40,7 +45,12 @@ src-$(CONFIG_MACH_STM32F0) += stm32/stm32f0.c stm32/gpioperiph.c
 src-$(CONFIG_MACH_STM32F0) += stm32/stm32f0_adc.c stm32/stm32f0_i2c.c
 src-$(CONFIG_MACH_STM32F1) += ../lib/stm32f1/system_stm32f1xx.c
 src-$(CONFIG_MACH_STM32F1) += stm32/stm32f1.c generic/armcm_timer.c
-src-$(CONFIG_MACH_STM32F1) += stm32/adc.c stm32/i2c.c
+ifeq ($(CONFIG_N32G452), y)
+	src-$(CONFIG_MACH_STM32F1) += ../lib/n32g45x/n32g45x_adc.c stm32/n32g45x_adc.c
+else
+	src-$(CONFIG_MACH_STM32F1) += stm32/adc.c
+endif
+src-$(CONFIG_MACH_STM32F1) += stm32/i2c.c
```
KevinOConnor, line 53: Same comment here.  Something like:
```
src-$(CONFIG_MACH_STM32F103) += stm32/adc.c
src-$(CONFIG_MACH_N32G452) += ../lib/n32g45x/n32g45x_adc.c stm32/n32g45x_adc.c
```

src/stm32/n32g45x_adc.c
```diff
@@ -0,0 +1,179 @@
+// ADC functions on N32G45x
+//
+// Copyright (C) 2019-2020  Kevin O'Connor <kevin@koconnor.net>
+// Copyright (C) 2022-2023  Alexey Golyshin <stas2z@gmail.com>
+//
+// Based on Kevin's stm32f1x code, but modified for N32G45x specific ADC
```
KevinOConnor, line 6: It's preferable not to put file history in the code comments - the git commit history can be used for that.  It's not necessary to credit me here.

Also, if you do have code comments, it's preferable to keep them separate from the initial license block, as we want to keep the license info succinct.

KevinOConnor (commented):
Thanks.  In general it looks good to me.  I have some minor code comments.  See below.

-Kevin

PR #6100, opened 2023-03-08
src/linux/main: Lock Memory and Increase Priority

mattaw:
Real-time programming best practice is to lock real-time processes memory to prevent paging under memory pressure.
As The Linux MCU process has well bounded memory and a small RAM footprint locking the entire process' current and
future memory has no downsides. (See bootlin training and Linux Foundation documentation linked below.) Additionally, RT 
process priority ranges from 0-99 (although POSIX only requires 32), boost MCU process priority to half the max/2 to improve behavior when the system is considering real-time kernel processes or user processes.

Reference links:
bootlin: https://bootlin.com/doc/training/preempt-rt/preempt-rt-slides.pdf Linux Foundation: https://wiki.linuxfoundation.org/realtime/documentation/howto/applications/application_base#howto_build_a_simple_rt_application

Signed-off-by: Matthew Swabey <matthew@swabey.org>

### DIFF
--- a/src/linux/main.c
+++ b/src/linux/main.c
@@ -4,10 +4,11 @@
 //
 // This file may be distributed under the terms of the GNU GPLv3 license.
 
-#include </usr/include/sched.h> // sched_setscheduler
+#include </usr/include/sched.h> // sched_setscheduler sched_get_priority_max
 #include <stdio.h> // fprintf
 #include <string.h> // memset
 #include <unistd.h> // getopt
+#include <sys/mman.h> // mlockall MCL_CURRENT MCL_FUTURE
 #include "board/misc.h" // console_sendf
 #include "command.h" // DECL_CONSTANT
 #include "internal.h" // console_setup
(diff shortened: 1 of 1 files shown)


mattaw:
I have an improvement to the MPU-* accelerometers that I will be submitting after more testing on more platforms which builds on this improvement, but this seems a nice standalone PR with no downsides so I am submitting it separately.

Jookia:
It might be better to consider changing from SCHED_FIFO to SCHED_DEADLINE which is meant for real-time use cases

mattaw:
I believe both SCHED_FIFO, SCHED_RR & DEADLINE are all real time schedulers. https://wiki.linuxfoundation.org/realtime/documentation/technical_basics/sched_policy_prio/start

Jookia:
SCHED_FIFO and SCHED_RR just give the tasks much higher priority, they don't actually guarantee you get time to run. See https://man7.org/linux/man-pages/man7/sched.7.html

mattaw:
@Jookia your suggestion works a lot better than what I was doing, and is likely to work properly for the widest range of klipper hosts. I'll submit a deadline scheduled version tomorrow.

Jookia:
That's super cool to hear! If I had a Linux board I'd test it. Good luck :)

PR #6102, opened 2023-03-09
filament_switch_sensor: Abort print if no filament is present

Jookia:
Currently the filament sensor only does anything if the filament runs out during print. It will happily let your printer print without any filament available.

Change this by checking at print start and erroring if there's no filament. Running G-code to pause and insert filmanet here isn't possible as the printer may not be homed or ready to do that.

### DIFF
--- a/klippy/extras/filament_switch_sensor.py
+++ b/klippy/extras/filament_switch_sensor.py
@@ -31,6 +31,7 @@ class RunoutHelper:
         self.sensor_enabled = True
         # Register commands and event handlers
         self.printer.register_event_handler("klippy:ready", self._handle_ready)
+        self.printer.register_event_handler("idle_timeout:printing", self._handle_print)
         self.gcode.register_mux_command(
             "QUERY_FILAMENT_SENSOR", "SENSOR", self.name,
             self.cmd_QUERY_FILAMENT_SENSOR,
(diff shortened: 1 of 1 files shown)


Jookia:
As a heads up, this patch will break configs that have a filament sensor listed but not connected.

jordan-woyak:
Just noting that this change would break my "workflow". I have a configured filament run out sensor but sometimes choose to not pass the filament through it.

Arksine:
I concur that this change will break existing workflows.  It would result in a raised exception even if a user simply moved to the tool to better position it for loading.

It should be possible for users to implement this kind of check themselves in a `PRINT_START` macro.  A snippet like the following should abort the print.  
```
{% if not printer["filament_switch_sensor my_sensor"].filament_detected %}
  {action_raise_error("No filament present at print start")}
{% endif %}
```

Gilabite:
what about printers with tool changers and multiple run out sensors? there may not need to be filament in some of them for some prints.

Jookia:
I'm confused- is this expected behaviour to start prints when your runout sensor is empty? The documentation doesn't say this. Maybe having a way to specify that it is required_for_print in config would work?

Jookia:
Okay I'm trashing this code because it doesn't even work right. Klipper doesn't really know when a print 'starts' so this also means you can't head up the extruder if there's no filament in it to insert it. Maybe the macro should be documented?

PR #6104, opened 2023-03-11
tmc: Do not override tcoolthrs if it is configured

leptun:
If tcoolthrs is configured (not the default 0), then do not force the value of tcoolthrs=0xfffff during homing. This way, tcoolthrs can be set to a custom value during homing.

`tpwmthrs` and `en_pwm_mode`/`en_spreadcycle` are now also correctly restored if they were changed after startup.

Without this change, it is not possible to use a value different than tcoolthrs=0xfffff during homing, which makes tmc homing less reliable than it can be. After this change, a delayed_gcode can be used at startup to set tcoolthrs to a specific velocity (eg 15mm/s) and make homing more reliable. Once I have finalized my homing macros, I'll make them available as well.

### DIFF
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -452,13 +452,6 @@ class TMCVirtualPinHelper:
         if self.diag_pin is None:
             raise ppins.error("tmc virtual endstop requires diag pin config")
         # Setup for sensorless homing
-        reg = self.fields.lookup_register("en_pwm_mode", None)
-        if reg is None:
-            self.en_pwm = not self.fields.get_field("en_spreadcycle")
-            self.pwmthrs = self.fields.get_field("tpwmthrs")
-        else:
-            self.en_pwm = self.fields.get_field("en_pwm_mode")
-            self.pwmthrs = 0
         self.printer.register_event_handler("homing:homing_move_begin",
                                             self.handle_homing_move_begin)
         self.printer.register_event_handler("homing:homing_move_end",
@@ -468,19 +461,24 @@ class TMCVirtualPinHelper:
     def handle_homing_move_begin(self, hmove):
         if self.mcu_endstop not in hmove.get_mcu_endstops():
             return
+        self.pwmthrs = self.fields.get_field("tpwmthrs")
+        self.coolthrs = self.fields.get_field("tcoolthrs")
         reg = self.fields.lookup_register("en_pwm_mode", None)
         if reg is None:
             # On "stallguard4" drivers, "stealthchop" must be enabled
+            self.en_pwm = not self.fields.get_field("en_spreadcycle")
             tp_val = self.fields.set_field("tpwmthrs", 0)
             self.mcu_tmc.set_register("TPWMTHRS", tp_val)
             val = self.fields.set_field("en_spreadcycle", 0)
         else:
             # On earlier drivers, "stealthchop" must be disabled
+            self.en_pwm = self.fields.get_field("en_pwm_mode")
             self.fields.set_field("en_pwm_mode", 0)
             val = self.fields.set_field(self.diag_pin_field, 1)
         self.mcu_tmc.set_register("GCONF", val)
-        tc_val = self.fields.set_field("tcoolthrs", 0xfffff)
-        self.mcu_tmc.set_register("TCOOLTHRS", tc_val)
+        if self.coolthrs == 0:
+            tc_val = self.fields.set_field("tcoolthrs", 0xfffff)
+            self.mcu_tmc.set_register("TCOOLTHRS", tc_val)
     def handle_homing_move_end(self, hmove):
         if self.mcu_endstop not in hmove.get_mcu_endstops():
             return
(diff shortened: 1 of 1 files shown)


leptun:
This is the macro I am using at the moment. It's purpose is to improve the homing on a Prusa MK3S printer. It should be possible to apply something like this for any cartesian printer. CoreXY can also be supported with some minor changes:
```
[delayed_gcode homing_setup]
gcode:
  SET_TMC_FIELD STEPPER=stepper_x FIELD=tcoolthrs VALUE=260
  SET_TMC_FIELD STEPPER=stepper_y FIELD=tcoolthrs VALUE=260
  # SET_TMC_FIELD STEPPER=stepper_x FIELD=tcoolthrs VELOCITY=15
  # SET_TMC_FIELD STEPPER=stepper_y FIELD=tcoolthrs VELOCITY=15


  # Debug pins for logic analyzer
  #SET_TMC_FIELD STEPPER=stepper_x FIELD=diag1_stall VALUE=1
  #SET_TMC_FIELD STEPPER=stepper_x FIELD=diag1_pushpull VALUE=1
  #SET_TMC_FIELD STEPPER=stepper_y FIELD=diag1_stall VALUE=1
  #SET_TMC_FIELD STEPPER=stepper_y FIELD=diag1_pushpull VALUE=1
initial_duration: 0.5

[gcode_macro G28]
rename_existing: G28.1
description: Fancy homing move
gcode:
  {% if 'X' in params or params|length == 1 %}
    SET_TMC_CURRENT STEPPER=stepper_x CURRENT=0.6
  {% endif %}
  {% if 'Y' in params or params|length == 1 %}
    SET_TMC_CURRENT STEPPER=stepper_y CURRENT=0.6
  {% endif %}
  {% if 'Z' in params or params|length == 1 %}
    SET_TMC_CURRENT STEPPER=stepper_z CURRENT=0.3
  {% endif %}

  G28.1 {rawparams}

  {% if 'X' in params or params|length == 1 %}
    SET_TMC_CURRENT STEPPER=stepper_x CURRENT={printer['tmc2240 stepper_x'].run_current} HOLDCURRENT={printer['tmc2240 stepper_x'].hold_current}
  {% endif %}
  {% if 'Y' in params or params|length == 1 %}
    SET_TMC_CURRENT STEPPER=stepper_y CURRENT={printer['tmc2240 stepper_y'].run_current} HOLDCURRENT={printer['tmc2240 stepper_y'].hold_current}
  {% endif %}
  {% if 'Z' in params or params|length == 1 %}
    SET_TMC_CURRENT STEPPER=stepper_z CURRENT={printer['tmc2130 stepper_z'].run_current} HOLDCURRENT={printer['tmc2130 stepper_z'].hold_current}
  {% endif %}
```

On startup it configures tcoolthrs for all drivers that are used during homing. I intend to make this value configurable in the config file in a PR later (the CoolStep PR). It also enables some debugging pins for me. That part can be ignored by other people.
When G28 is called, extra actions are performed. All axes that are involved in the homing move get their current reduced, the homing is performed and then the currents are restored. This avoids the need of specifying a homing_current in the config file (a change from an older PR that didn't get merged). Setting currents like this is so slow that I don't actually need to wait at all for the AT#1 autotune in the drivers to kick in.

With this macro, the "second_home" feature can be enabled in klipper and it now works properly since the second move doesn't instantly trigger a stall condition as it used to before configuring the current and tcoolthrs. This is my stepper config for X/Y:
```
homing_speed: 75
second_homing_speed: 25
homing_retract_dist: 8
```

This is how it looks like in practice:

https://user-images.githubusercontent.com/17808203/224506119-259398bd-0d7d-42be-9d1b-9b71b4ee6d80.mp4

KevinOConnor:
Thanks.  In general, it looks good to me.  It seems to me this change is fairly safe and is unlikely to introduce a regression.  I'd like to get confirmation that you agree there is little chance of a regression.

-Kevin

leptun:
I think there are few chances of regressions.

Moving the `self.en_pwm` and `self.pwmthrs` initialization from `setup_pin()` to `handle_homing_move_begin()` fixes the issue where if these fields are changed at runtime with the `SET_TMC_FIELD` command, after a home, the config value is restored instead of the correct "previous" value. I don't see how this could cause regressions, except for the cases where a user was specifically relying on this broken behaviour (?), but I doubt anyone would actually do that.

As for the `tcoolthrs` changes, I save the old value when the homing begins and restore it when it ends. This field defaults to 0 since it is never configured from the config file. So the old behaviour is to have the field to 0xfffff during homing and 0x0 at any other time. But this register is critical for CoolStep operation and for good homing, so having the homing functions always override this register is problematic. In terms of breaking changes, I don't think there are any. The old behaviour is preserved if the user doesn't configure the register with `SET_TMC_FIELD`. If the user does indeed configure it to something besides 0, then he knows what he is doing and it's probably fine. So I don't think it can cause any breaking change to anyone's config, unless they were trying to do something with CoolStep, but even then they already had other issues to fix such as having to reconfigure this register after every homing move.

KevinOConnor:
Thanks.  I committed this change with commit c54d83c9 on top.

-Kevin

leptun:
@KevinOConnor Would there be interest in getting the finalized homing macro into sample-macros.cfg? I think it could be helpful for people using sensorless homing on their printers.

Sineos:
The "Macro" Category on https://klipper.discourse.group might be a good place to start

masterxq:
> @KevinOConnor Would there be interest in getting the finalized homing macro into sample-macros.cfg? I think it could be helpful for people using sensorless homing on their printers.

Hi @leptun,

Did you post it somewhere?

Best regards

leptun:
Yes, see this comment. I updated it over time by editing the comment.

https://github.com/Klipper3d/klipper/pull/6104#issuecomment-1464980909

PR #6105, opened 2023-03-11
Support for USART5 on STM32G0B1

Florin-Popescu:
Hello!

I'm using an SKR mini E3 V3 board and I needed to access the USART5 pins which are on PD2, PD3.
Thought it would be as simple as extending that configuration in Kconfig with my combination of pins, but that didn't work.

Many headaches later, turns out that the bits to enable/reset USART5 & 6 do not follow the rule from `stm32g0.c` regarding the offset from APBPERIPH_BASE, but needed to be added as exceptions similar to the USB and others.

While I've fixed the RCC bits for USART6 as well, I was only able to test the configuration I needed, USART5 on PD2, PD3. That's why I only included this in the Kconfig.

Hoping this saves other people some time.

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -385,6 +385,10 @@ choice
         bool "Serial (on UART4 PA0/PA1)"
         depends on MACH_STM32H7
         select SERIAL
+    config STM32_SERIAL_USART5
+        bool "Serial (on USART5 PD2/PD3)" if LOW_LEVEL_OPTIONS
+        depends on MACH_STM32G0Bx
+        select SERIAL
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  Adding support for USART5 seems useful.  Commits need to have a signed-off-by line though - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

Sineos:
Any updates @Florin-Popescu?

Florin-Popescu:
Fixed! Sorry about that.

KevinOConnor:
Thanks.

-Kevin

PR #6108, opened 2023-03-12
linux/timer: Use volatile sig_atomic_t in signal handler to avoid undefined behavior

mattaw:
Convert flag variable to `volatile sig_atomic_t` for simplicity and to avoid potential undefined behavior, ref SIG31-C [1] and `man 2 signal` on Linux. This has the additional side benefit of eliminating the ARM specific code for readl() and writel() functions and will be portable.

Tested extensively by inserting timing output messages into timed events, no performance difference noted.

[1] https://wiki.sei.cmu.edu/confluence/display/c/sig31-c.+do+not+access+shared+objects+in+signal+handlers

Signed-off-by: Matthew Swabey matthew@swabey.org

### DIFF
--- a/src/linux/timer.c
+++ b/src/linux/timer.c
@@ -5,6 +5,7 @@
 // This file may be distributed under the terms of the GNU GPLv3 license.
 
 #include <time.h> // struct timespec
+#include <signal.h> // sig_atomic_t
 #include "autoconf.h" // CONFIG_CLOCK_FREQ
 #include "board/io.h" // readl
 #include "board/irq.h" // irq_disable
--- a/src/linux/timer.c
+++ b/src/linux/timer.c
@@ -19,8 +19,8 @@ static struct {
     uint32_t last_read_time;
     // Fields for converting from a systime to ticks
     time_t start_sec;
-    // Flags for tracking irq_enable()/irq_disable()
-    uint32_t must_wake_timers;
+    // Flags for tracking irq_enable()/irq_disable() ref SIG31-C
+    volatile sig_atomic_t must_wake_timers; 
     // Time of next software timer (also used to convert from ticks to systime)
     uint32_t next_wake_counter;
     struct timespec next_wake;
@@ -268,7 +268,7 @@ void
 irq_wait(void)
 {
     // Must atomically sleep until signaled
-    if (!readl(&TimerInfo.must_wake_timers)) {
+    if (!TimerInfo.must_wake_timers) {
         timer_disable_signals();
         if (!TimerInfo.must_wake_timers)
             console_sleep(&TimerInfo.ss_sleep);
@@ -280,6 +280,6 @@ irq_wait(void)
 void
 irq_poll(void)
 {
-    if (readl(&TimerInfo.must_wake_timers))
+    if (TimerInfo.must_wake_timers)
         timer_dispatch();
 }
--- a/src/linux/timer.c
+++ b/src/linux/timer.c
@@ -20,7 +20,7 @@ static struct {
     // Fields for converting from a systime to ticks
     time_t start_sec;
     // Flags for tracking irq_enable()/irq_disable() ref SIG31-C
-    volatile sig_atomic_t must_wake_timers; 
+    volatile sig_atomic_t must_wake_timers;
     // Time of next software timer (also used to convert from ticks to systime)
     uint32_t next_wake_counter;
     struct timespec next_wake;


mattaw:
Hum, FYI I can only find a reference to `volatile sig_atomic_t` and its use in `man 2 select_tut` and `man aio` on my MainsailOS install on my RPi, both of which are example programs and it is in the signal handler code.  I wonder why it is missing from the man page for signal.

KevinOConnor:
Thanks.  As a high-level question, is this a "code cleanup" change or a "bug fix" change?

The convention I've seen with atomic accesses is to prefer access functions over "volatile" decorations.  (It was a notable recommendation from the Linux devs a few years back.)  The issue with volatile is that it isn't well defined by the C specs, it's not well defined what happens when it is declared on a field in a struct, and it can cause undesirable pessimizations for accesses that don't require atomicity.  The readl() and writel() functions also perform a full compiler `barrier()` which gcc will treat differently from just a basic volatile access.  (`volatile` in gcc typically means strictly ordered wrt other volatile, while a `barrier` will strictly order wrt all memory accesses.)

-Kevin

KevinOConnor:
Interesting.  I looked a little bit further into this.  The standard libc headers define sig_atomic_t as:

```c
/* C99: An integer type that can be accessed as an atomic entity,
   even in the presence of asynchronous interrupts.
   It is not currently necessary for this to be machine-specific.  */
typedef int __sig_atomic_t;
```

So, in the end, the only thing this type conversion does, in practice, is change `uint32_t` to `int`.  In practice, on all platforms of interest, there is no distinction between these two types.

We can certainly make the change (for semantic purposes), but I don't think it is a good idea to move the `volatile` declaration into the field definition.  As I understand it, it could introduce a regression because a gcc "memory barrier" is no longer present on the `writel()`.  As per the convention mentioned above, I think it is preferable to use the `readl()` and `writel()` wrappers to better document that an access needs atomicity, and to instruct gcc that a full "barrier" is needed.

I understand why the use of `volatile sig_atomic_t` is listed as the only valid access from a signal handler - the goal is to make it clear that most memory accesses are not atomic and that the compiler is free to create arbitrary accesses from most instructions.  However, I believe the current code takes this into account and is careful to issue code that gcc will, in practice, produce atomic safe accesses.

Cheers,
-Kevin

mattaw:
Taking on board your feedback I think I will close this PR. 

Firstly, I plan to probe more into the Linux MCU, and I want to disable SA_RESTART and stress the heck out of it to see if it is restarting i2c_dev read/write/ioctls silently. It might also be doing that to the SPI read/writes. That might explain the strange FIFO size readings I get vs. actual MCU's, but it may not.

I may also try to scribble a version of the Linux MCU that uses signalfd instead of a signal handler to avoid the whole SA_RESTART/EINTR, sig_atomic_t and barrier() complexity by passing around an extra fd for the alarm and poll'ing both it and the console fd. Since the MCU never uses the async signal for any of its functionality, and never forks it will avoid the main downside of signalfd. I think it would only be a few lines in timer.c and console.c too.

KevinOConnor:
What do you think about making just the following change:

```diff
--- a/src/linux/timer.c
+++ b/src/linux/timer.c
@@ -20,7 +20,7 @@ static struct {
     // Fields for converting from a systime to ticks
     time_t start_sec;
     // Flags for tracking irq_enable()/irq_disable()
-    uint32_t must_wake_timers;
+    sig_atomic_t must_wake_timers;
     // Time of next software timer (also used to convert from ticks to systime)
     uint32_t next_wake_counter;
     struct timespec next_wake;
```

> I may also try to scribble a version of the Linux MCU that uses signalfd instead of a signal handler

FYI, that sounds similar to the code prior to 913d0992.  The main issue with a pure polling implementation is that the polling itself can become a cpu burden (in particular, prior to 913d0992 the code would call `clock_gettime()` in every `irq_poll()` call).

Cheers,
-Kevin

mattaw:
Oh, in this case I think we can block on the singalfd and the console fd using select() or poll() which should put us to sleep, waking either on a timer signal or traffic coming from Klippy vs. poll as in "constantly check a condition / variable" which as you say would consume a lot of CPU time.

guillaume-pais:
Here a link to a discussion we had on this topic: [https://klipper.discourse.group/t/is-linux-still-working-after-commit-913d099/7063](https://klipper.discourse.group/t/is-linux-still-working-after-commit-913d099/7063).
At the end we prefered an implementation with ```atomic_int```.

mattaw:
Thank you for all your inputs, @KevinOConnor and @guillaume-pais , I have not been able to demonstrate advantages of this PR despite trying several different approaches to break things so I am going to close it.

PR #6111, opened 2023-03-13
avr/i2c: Allow I2C bus speed to be set by Klipper

mattaw:
Allows Klipper to set the I2C bus speed. Previously the bus was fixed to 100kHz regardless of configured parameters. Tested at 100kHz and 400kHz.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4508,20 +4508,20 @@ SPI bus.
 The following parameters are generally available for devices using an
 I2C bus.
 
-Note that Klipper's current micro-controller support for i2c is
-generally not tolerant to line noise. Unexpected errors on the i2c
+Note that Klipper's current micro-controller support for I2C is
+generally not tolerant to line noise. Unexpected errors on the I2C
 wires may result in Klipper raising a run-time error. Klipper's
 support for error recovery varies between each micro-controller type.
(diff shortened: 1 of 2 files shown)


thijstriemstra (changes requested):
Can you also add this to the documentation?

mattaw:
Great comment - I never thought of that. Is there a way to get better visibility into the compiled code size etc. when building with Klipper so I can spot these in future? It has been years since I had to program AVRs and I am forgetting all the 8 bit tricks. I am creating a commit to merge i2c_init() into i2c_setup() as it is only ever called from i2c_setup() anyway with your suggestion. The last patch ran for 23 hours without error, so if this passes only a few minor tests I will commit it anyway.

KevinOConnor:

As for size increases, you can use the unix `size` program - for example: `size out/klipper.elf`.  Or, you can just look at the size of the final `out/klipper.bin`.  It's also possible to use objdump (eg, `avr-objdump -x out/klipper.elf | less`).  The automated build tests also report the file size of each binary - see: https://github.com/Klipper3d/klipper/actions

Separately, it's fine to move `i2c_init()` into `i2c_setup()`, but note that it does not improve performance or change code layout.  The Klipper build uses gcc optimizations `-O2` and `-flto`, which enables function inlining, even between c files.

-Kevin

PR #6112, opened 2023-03-14
scripts: add install script for Ubuntu 22.04 LTS 

Jesssullivan:
The python-dev package is not available on 22.04 Jammy, specifies the python3-dev package and python3 when creating a new virtualenv.

Signed-off-by: Jess Sullivan [jess@sulliwood.org](mailto:jess@sulliwood.org)

### DIFF
--- a/scripts/install-ubuntu-22.04.sh
+++ b/scripts/install-ubuntu-22.04.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+# This script installs Klipper on an Ubuntu 22.04 ("Jammy") machine
+
+PYTHONDIR="${HOME}/klippy-env"
+SYSTEMDDIR="/etc/systemd/system"
+KLIPPER_USER=$USER
+KLIPPER_GROUP=$KLIPPER_USER
+
+# Step 1: Install system packages
+install_packages()
+{
+    # Packages for python cffi
+    PKGLIST="virtualenv python3-dev libffi-dev build-essential"
+    # kconfig requirements
+    PKGLIST="${PKGLIST} libncurses-dev"
+    # hub-ctrl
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.

-Kevin

scripts/install-ubuntu-22.04.sh
```diff
@@ -0,0 +1,102 @@
...
+PYTHONDIR="${HOME}/klippy-env"
+SYSTEMDDIR="/etc/systemd/system"
+KLIPPER_USER=$USER
+KLIPPER_GROUP=$KLIPPER_USER
+
+# Step 1: Install system packages
+install_packages()
+{
+    # Packages for python cffi
+    PKGLIST="virtualenv python3-dev libffi-dev build-essential"
```
tomgwilliams, line 13: **python3-dev** needs to be replaced with **python-dev-is-python3**. Errors out during installation. I replaced manually and it seems to fix this.

"
Package python-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python2-dev:i386 python2:i386 python2-dev python2 python-dev-is-python3
"

tomgwilliams: /
/ #6112

Sineos:
I think you are mixing up things:
* `python3-dev` exists in Jammy --> https://packages.ubuntu.com/jammy/python3-dev
* `python-dev-is-python3` does not exists
* `python-is-python3` is just a symlink from python to python3 so that it is easier for the shebangs --> https://packages.ubuntu.com/jammy/python-is-python3

PR #6115, opened 2023-03-15
 sensor_mpu9250: Fix or improve reliability 

mattaw:
Fixes common MPU-9250 accelerometer issues for RPi Linux MCU and
improves reliability on all other architectures by adjusting the MPU-*
reading algorithm to only read whole Klipper messages' worth of
data from MPU-* and eliminating many unnecessary checks of the MPU FIFO
fill-level that consumed bus bandwidth needed for data transfer. Fixes
intermittent "Lost communication with MCU 'rpi'" due to "Timer too
close" and transposed / corrupted data due to FIFO overrun/data loss
when using MPU-* accelerometers. In addition FIFO overrun checks are
performed by testing the MPU interrupt flag vs. inferring from the FIFO
fill level.

Stress tested for 13hrs with two MPU-6500 attached to one I2C bus on RPi
    and one on a PR2040
Stress tested for 23hrs with two MPU-6500 attached to one I2C bus on RPi
    and one on a ATmega328P (Seeduino Nano)

### DIFF
--- a/src/sensor_mpu9250.c
+++ b/src/sensor_mpu9250.c
@@ -64,7 +64,7 @@ get_fifo_status (struct mpu9250 *mp)
 {
     uint8_t reg[] = {AR_FIFO_COUNT_H};
     uint8_t msg[2];
-    i2c_read(mp->i2c->i2c_config, sizeof(reg), reg, sizeof(2), msg);
+    i2c_read(mp->i2c->i2c_config, sizeof(reg), reg, sizeof(msg), msg);
     msg[0] = 0x1F & msg[0]; // discard 3 MSB per datasheet
     uint16_t bytes_to_read = ((uint16_t)msg[0]) << 8 | msg[1];
     if (bytes_to_read > mp->fifo_max) mp->fifo_max = bytes_to_read;
@@ -176,8 +176,9 @@ mp9250_start(struct mpu9250 *mp, uint8_t oid)
     msg[1] = SET_USER_FIFO_EN; // enable FIFO buffer access
     i2c_write(mp->i2c->i2c_config, sizeof(msg), msg);
 
-    uint8_t int_reg[] = {AR_INT_STATUS}; // clear FIFO overflow flag
-    i2c_read(mp->i2c->i2c_config, sizeof(int_reg), int_reg, 1, msg);
+    msg[0] = AR_INT_STATUS; // clear FIFO overflow flag. Use msg[] to save stack
+    i2c_read(mp->i2c->i2c_config, sizeof(msg[0]), &msg[0],
+                sizeof(msg[1]), &msg[1]);
 
     msg[0] = AR_FIFO_EN;
     msg[1] = SET_ENABLE_FIFO; // enable accel output to FIFO
@@ -200,12 +201,11 @@ mp9250_stop(struct mpu9250 *mp, uint8_t oid)
     i2c_write(mp->i2c->i2c_config, sizeof(msg), msg);
     uint32_t end2_time = timer_read_time();
 
-    // Detect if a FIFO overrun occured
-    uint8_t int_reg[] = {AR_INT_STATUS};
-    uint8_t int_msg;
-    i2c_read(mp->i2c->i2c_config, sizeof(int_reg), int_reg, sizeof(int_msg),
-                &int_msg);
-    if (int_msg & FIFO_OVERFLOW_INT)
+    // Detect if a FIFO overrun occured. Use msg[] to save stack
+    msg[0] = AR_INT_STATUS;
+    i2c_read(mp->i2c->i2c_config, sizeof(msg[0]), &msg[0],
+                sizeof(msg[1]), &msg[1]);
+    if (msg[1] & FIFO_OVERFLOW_INT)
         mp->limit_count++;
 
     // Report final data
--- a/src/sensor_mpu9250.c
+++ b/src/sensor_mpu9250.c
@@ -1,5 +1,6 @@
 // Support for gathering acceleration data from mpu9250 chip
 //
+// Copyright (C) 2023 Matthew Swabey <matthew@swabey.org>
 // Copyright (C) 2022 Harry Beyel <harry3b9@gmail.com>
 // Copyright (C) 2020-2021 Kevin O'Connor <kevin@koconnor.net>
 //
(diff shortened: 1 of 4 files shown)


mattaw:
Oh, I have left in one output() line in sensor_mpu9250.c called on mpu9250_stop(), which reports the max fill the fifo experienced while running and whether a FIFO overrun occurred during the run at one point. I can patch that out after review. The behavior is slightly different than for the ADXL345, as it does not update limit_count every time a FIFO overrun occurs due to the accelerometer not being read often enough, this code only notes if one or more did happen at some point during the run. All my testing shows no overruns so hopefully this is acceptable. I may rework accelerometers completely at some point to accommodate more chips more easily and retry if an overrun happened vs. accepting junk or failing.

src/sensor_mpu9250.c
```diff
@@ -55,14 +59,16 @@ enum {
...
 // Reads the fifo byte count from the device.
-uint16_t
+static uint16_t
 get_fifo_status (struct mpu9250 *mp)
 {
-    uint8_t regs[] = {AR_FIFO_COUNT_H};
+    uint8_t reg[] = {AR_FIFO_COUNT_H};
     uint8_t msg[2];
-    i2c_read(mp->i2c->i2c_config, sizeof(regs), regs, 2, msg);
+    i2c_read(mp->i2c->i2c_config, sizeof(reg), reg, sizeof(2), msg);
```
KevinOConnor, line 67: `sizeof(2)` is surely a bug.

src/sensor_mpu9250.c
```diff
@@ -182,6 +176,9 @@ mp9250_start(struct mpu9250 *mp, uint8_t oid)
     msg[1] = SET_USER_FIFO_EN; // enable FIFO buffer access
     i2c_write(mp->i2c->i2c_config, sizeof(msg), msg);
 
+    uint8_t int_reg[] = {AR_INT_STATUS}; // clear FIFO overflow flag
+    i2c_read(mp->i2c->i2c_config, sizeof(int_reg), int_reg, 1, msg);
```
KevinOConnor, line 180: gcc wont know that `int_reg[]` has limited lifetime and thus will allocate stack space for it.  It's better to just reuse the existing `msg[]` allocation.

src/sensor_mpu9250.c
```diff
@@ -203,18 +200,22 @@ mp9250_stop(struct mpu9250 *mp, uint8_t oid)
...
     uint32_t end2_time = timer_read_time();
 
-    // Drain any measurements still in fifo
-    uint16_t fifo_bytes = get_fifo_status(mp);
-    while (fifo_bytes >= BYTES_PER_FIFO_ENTRY) {
-        mp9250_query(mp, oid);
-        fifo_bytes = get_fifo_status(mp);
-    }
+    // Detect if a FIFO overrun occured
+    uint8_t int_reg[] = {AR_INT_STATUS};
```
KevinOConnor, line 204: Same observation on variable lifetimes.

src/sensor_mpu9250.c
```diff
@@ -203,18 +200,22 @@ mp9250_stop(struct mpu9250 *mp, uint8_t oid)
...
 
     // Report final data
     if (mp->data_count > 0)
         mp9250_report(mp, oid);
+    uint16_t bytes_to_read = get_fifo_status(mp);
     mp9250_status(mp, oid, end1_time, end2_time,
-                    fifo_bytes / BYTES_PER_FIFO_ENTRY);
+                    bytes_to_read / BYTES_PER_FIFO_ENTRY);
+    output("mp9250_stop limit_count=%hu fifo_max=%hu", mp->limit_count,
+            mp->fifo_max);
```
KevinOConnor, line 218: As you've indicated, we'd want to remove the debugging code (and all references to `fifo_max`) before merging.

KevinOConnor (commented):
Thanks.  As high-level feedback, I think this should be separated into three PRs - one for the linux change (which seems fine to me), one for the mpu changes (which seem fine to me, but see my comments below), and one for the doc changes.  In particular the doc will be reviewed by different people than the other changes.

Also, it would be preferable to keep the md files to a line length of 80 characters.

@bluesforte - do you have any comments on the mpu changes?

-Kevin

src/sensor_mpu9250.c
mattaw: ... how embarrassing. I was trying to standardize on sizeof() vs a mix of sizeof() vs. hardcoded values and ended up with something worse. Fixing now.

src/sensor_mpu9250.c
mattaw: Done!

mattaw:
Re your high-level feedback the change to mpu9250.py and sensor_mpu9250.c won't work without each other as they are interrelated. I could split it in two and add notes saying "you need PR####" in each if that fits the workflow better? The tight timing in mpu9250.py needs to be relaxed to avoid triggering a Linux MCU shutdown with "Timer too close", and the more efficient code in sensor_mpu9250.c is needed to prevent the occasional FIFO overflows inside the MPU that were happening after the timing relaxation.

After this is merged I would like to take a look at the ADXL345 and MPU code to detect problems and retry during the test. Perhaps try to make the code more pluggable/generic to support future accelerometers.

mattaw:
Merged fixes for comments in case someone would like to test this PR while I get it split in two.

KevinOConnor:
> the change to mpu9250.py and sensor_mpu9250.c won't work without each other as they are interrelated

All the mpu changes should be in a single PR, and likely a single commit (as we want the repo to be functional at each commit so tools like `git bisect` work properly).

The Linux change (schedule priority and memory locking) seems orthogonal to me.  The docs changes (mpu wiring suggestions) also seem orthogonal.  Thus, unless I'm missing something, those changes seem to add additional complexity to this PR.

Cheers,
-Kevin

Sineos:
Any updates @mattaw?

mattaw:
Yes! I am letting perfect be the enemy of good - the improvements work well enough but I can't get the latency of the Linux MCU code well controlled. Even with RT scheduling it seems. I'll do a PR for the changes I have now to the serial mpu 9250 driver and submit the PR.

On the back burner is recompiling the Pi kernel to support eBPF and watch the totality of the program run including the kernel internals to figure out just what it is doing that is taking so long.

mattaw:
Last sub-PR to complete this giant PR #6169 submitted.

PR #6116, opened 2023-03-15
Add support for Nation N32G45x mcus

stas2z:
N32G452/G455 are mostly compatible with STM32F103 but have M4 core and different ADC.
So it compile as f103 and working except adc readings..
Also it twice faster than f103 as it can run at 144Mhz but without modifying common klipper stm32 rcc init code it can safely run at 128MHz only (seems like it requires flash latency 4, but not tested).
This PR uses some code from N32G45x BSP (seems like it's a stdperiph with renamed fields, vars, functions etc, so it's not conflict with stm32 defs), all copyrights given.

Also this PR adds config for Voxelab Aquila STM32/G32/N32 board FFP0173 1.0.1 - first printer shipped with this MCU, but there are some other printers with it like different Aquila models and some other Voxelab/FlashForge devices.

This is a fixed version of #6099 where i deleted branch instead of rebase.
Hope all problems mentioned by @KevinOConnor are solved here. Sorry for my mistake.

Signed-off-by: Alexey Golyshin [stas2z@gmail.com](mailto:stas2z@gmail.com)


KevinOConnor:
Thanks.  As high-level feedback, it seems okay to me.  I do have a couple of comments:

1. Can the `MACH_N32G452` and `MACH_N32G455` do `select MACH_STM32F1` so that we don't have to make as many `#if` changes to the code?
2. The example printer config file would need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html .  At a minimum the filename needs to have a year in it.  It might be simpler to commit the code changes in one PR and add the example config in a separate commit.

-Kevin

stas2z:
> 1. Can the `MACH_N32G452` and `MACH_N32G455` do `select MACH_STM32F1` so that we don't have to make as many `#if` changes to the code?
Unfortunately no, or it will require makefile ifdefs again




> 2\. The example printer config file would need to follow the guidelines at https://www.klipper3d.org/Example_Configs.html .  At a minimum the filename needs to have a year in it.  It might be simpler to commit the code changes in one PR and add the example config in a separate commit.

Fixed.

stas2z:
> I think it can be done with lists

Ok, i got you,  i thought not only f103 supported in F1.
Will back with changes soon.

KevinOConnor:
Thanks.  I committed this with commit ecc23fc6 on top.

-Kevin

PR #6117, opened 2023-03-15
tmc2209: Remove duplicate `pdn_disable` initialization

leptun:
I don't think this can break anything. `pdn_disable` was already initialized to 1 a few lines higher. Only tmc2209 had this duplicate code.


KevinOConnor:
Thanks.

-Kevin

PR #6118, opened 2023-03-15
tmc: Enable `multistep_filt` for drivers that support it

leptun:
`multistep_filt` was not configured at all on tmc2240 and tmc5160. As such, it would default to the value of 0 when the GCONF field was sent to the driver. However, that field has a default value of 1 in the driver, so klipper was overriding the defaults by accident. This field improves the stability of velocity based thresholds (it's effectively TSTEP filtering) and it also improves the current regulation stability in stealthchop2 (according to the datasheet). This field was already set to 1 correctly in the tmc2208/tmc2209 driver code.

It should not cause any breaking change. It should just improve StealthChop2 operation.

### DIFF
--- a/klippy/extras/tmc2240.py
+++ b/klippy/extras/tmc2240.py
@@ -353,6 +353,7 @@ class TMC2240:
         self.get_phase_offset = cmdhelper.get_phase_offset
         self.get_status = cmdhelper.get_status
         # Setup basic register values
+        self.fields.set_field("multistep_filt", True)
         tmc.TMCWaveTableHelper(config, self.mcu_tmc)
         self.fields.set_config_field(config, "offset_sin90", 0)
         tmc.TMCStealthchopHelper(config, self.mcu_tmc, TMC_FREQUENCY)
--- a/klippy/extras/tmc5160.py
+++ b/klippy/extras/tmc5160.py
@@ -326,6 +326,7 @@ class TMC5160:
         self.get_phase_offset = cmdhelper.get_phase_offset
         self.get_status = cmdhelper.get_status
         # Setup basic register values
+        self.fields.set_field("multistep_filt", True)
         tmc.TMCWaveTableHelper(config, self.mcu_tmc)
         tmc.TMCStealthchopHelper(config, self.mcu_tmc, TMC_FREQUENCY)
         set_config_field = self.fields.set_config_field


KevinOConnor:
Thanks.

-Kevin

Virtualchronos:
Hello !
Sorry to bother but this commit broke my two printers, now they skip a lot of steps at high speed.
I'm using fysect spider motherboard on both and also fysetc tmc 5160 pro @ 40V
I hope this will help to find the issue cause and fix it !

leptun:
Hmmm. That's super weird. It's not the expected result of enabling this field. Can you please share the config file that you are using? Maybe there's something specific to your setup that's causing issues.

Virtualchronos:
[klipper_config Darkstar.zip](https://github.com/Klipper3d/klipper/files/11045509/klipper_config.Darkstar.zip)
[Voron config.zip](https://github.com/Klipper3d/klipper/files/11045514/Voron.config.zip)

Here's my two printers config

leptun:
When does it exactly skip steps? During the print or immediately after doing a quick move after homing? If it's right after homing, it's possible that the driver didn't autocalibrate the StealthChop2 parameters correctly, so it's using the wrong currents during the acceleration phase. You may try to help the driver by setting `driver_PWM_OFS` and `driver_PWM_GRAD` to the values of the registers PWM_OFS_AUTO and PWM_GRAD_AUTO after doing some medium velocity moves. Maybe this helps.
You may also try playing with `PWM_REG` to see if a higher value helps.

One other thing you could test is if disabling the `multistep_filt` option helps using the `SET_TMC_FIELD` command for X and Y steppers.

Virtualchronos:
it skips steps after a rapid move at 15k acceleration wich it handle perfectly well (until +-40k)
I could tune it if only i knew how to do without messing the entire thing... All those settings are totaly unknown from me

leptun:
@Virtualchronos I'll try to explain how to use those fields tomorrow (need to sleep now). You are using a very high acceleration for StealthChop. The extra filtering added by multistep_filt might make the current regulation more stable, but it also makes it slower, something that might pose problems at high acceleration rates. With your setup, you may be better off if you don't use multistep_filt.

@KevinOConnor I propose that we make this field settable from the config file and we default it to 1 (since most people run lower acceleration on their printer). Does this sound ok?

Virtualchronos:
if the setting can be turn off in the cfg file it would be a great improvement and i'm going to sleep too (1am in france) i gladly take your tutorial of tuning to try on my printers. I know they are awfully fast compared to regular printers that's why i often have some troubles others don't.

leptun:
Here is the best explanation I can come up with for tuning `driver_PWM_OFS` and `driver_PWM_GRAD`. These fields are the main parameters that control how much current the steppers get in StealthChop2 mode. This is different than run_current. Usually, these fields are autocalibrated by the driver at runtime. The autocalibration value is used automatically, however, you can still help the autocalibration by providing a good starting value for it. The best starting value that you can pick is usually whatever values the autocalibration last time successfully generated. To get the autocalibration result, you can DUMP_TMC the stepper after doing some moves at different speeds and looking at what the `PWM_OFS_AUTO` and `PWM_GRAD_AUTO` fields report back. You just take these values and put them back into `driver_PWM_OFS` and `driver_PWM_GRAD` in the config file. Now, I'm not entirely sure how much this will end up helping in your situation, but it's worth giving it a try,

The next field that you can tune is `driver_PWM_REG`. This value controls how quickly the StealthChop2 adapts to velocity changes. It has a range of 1-15 and it defaults to 4. You can try using a higher value to see if it helps in your case, but it might make the current regulation a bit more unstable. If it doesn't help, just leave it at the default value.

As for disabling the multistep_filt without modifying klipper itself, you can just add this to your config for now until I make it a PR to have it configurable:
```
[delayed_gcode disable_multistep_filt]
initial_duration: 0.5
gcode:
  SET_TMC_FIELD STEPPER=stepper_x FIELD=multistep_filt VALUE=0
  SET_TMC_FIELD STEPPER=stepper_y FIELD=multistep_filt VALUE=0
```

zztopper:
Guys, I've just got what I think is directly the consequence of this PR. 
Nothing changed in config, just updated Klipper on my Annex K3 printer. XY motion is running on TMC5160@48V, got this (see attached image).
After commenting the line `self.fields.set_field("multistep_filt", False)` everything's back to normal. 
As it seems to me, this PR takes more that gives. I'd remove it from the mainline, or at least documented that field and its new default and also I'd make it configurable as a tmc parameter.

leptun:
@zztopper See https://github.com/Klipper3d/klipper/pull/6134 for making it configurable.

PR #6119, opened 2023-03-16
config: Add Fysetc Cheetah V2.0

bbro44:
This adds support for the Cheetah v2.0 mainboard from Fysetc.


JamesH1978 (approved):
All looks good to me

Thanks
James

lbibass:
Just a note, this board has to have an obnoxiously long smooth time on the extruder thermistor due to a very noisy thermistor circuit. If I recall correctly, I had to set it to 5 seconds in order for it to be usable. Otherwise, prints just wouldn't start.

PR #6120, opened 2023-03-16
config: Add Ratrig V-Minion

bbro44:
This adds support for the stock configuration of the V-Minion printer from Ratrig, based on the Octopus v1.1 (STM32F446)

### DIFF
--- a/config/printer-ratrig-v-minion-2021.cfg
+++ b/config/printer-ratrig-v-minion-2021.cfg
@@ -75,6 +75,8 @@ nozzle_diameter: 0.4
 heater_pin: PA2
 sensor_type: Generic 3950
 sensor_pin: PF4
+control: pid
+
 min_temp: 0
 max_temp: 290
 
--- a/config/printer-ratrig-v-minion-2021.cfg
+++ b/config/printer-ratrig-v-minion-2021.cfg
@@ -76,6 +76,9 @@ heater_pin: PA2
 sensor_type: Generic 3950
 sensor_pin: PF4
 control: pid
+pid_Kp: 22.2
+pid_Ki: 1.08
+pid_Kd: 114
 
 min_temp: 0
 max_temp: 290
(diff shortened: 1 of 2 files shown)


config/printer-ratrig-v-minion-2021.cfg
```diff
@@ -0,0 +1,178 @@
+# This file contains pin mappings for the Ratrig V-Minion
+# with an Octopus Pro v1.1 board.
+#
+# To use this config, during "make menuconfig" select the STM32F446
+# with a "32KiB bootloader" and USB (on PA11/PA12)
+# communication.
```
JamesH1978, line 6: Please state that this F446 board needs a 12MHZ crystal(stock crystal setting causes many timing issues

Thanks
James

config/printer-ratrig-v-minion-2021.cfg
```diff
@@ -0,0 +1,178 @@
+# This file contains pin mappings for the Ratrig V-Minion
+# with an Octopus Pro v1.1 board.
```
JamesH1978, line 2: Please state that this is for the "full kit" version, as configs should only describe stock hardware that an end user would receive, we do not accept DIY printers due to the massive amount of variation that can be involved. If this config describes one of those DIY variants then we would be inclined not to accept this addition.

You may also want to state that this config will not have the full out of the box functionality of the stock ratos version of klipper.

Thanks
James

JamesH1978 (changes requested):
Thank you for submitting this config, I have made some comments.

Thanks,
James

JamesH1978 (approved):
Looks fine to me.

Thanks,
James

KevinOConnor:
Thanks.

-Kevin

dewi-ny-je:
20000 as standard acceleration? from the videos I saw online, that can be achieved only AFTER input shaping. The standard config file should be less extreme. 
For example, the Voron is set to 3000 as standard, which is a good starting point if no tuning is performed.

Can the V-minion use such high acceleration without input shaping?

bbro44:
Good point, I always forget that its default instead of max. I can update that once I get back from my trip. Thanks!

PR #6126, opened 2023-03-21
Adding support for Tina2S printer

securelyfitz:
The Tina2S made by Weefun/Weedo is mostly a generic stm32 printer, except for DTR being connected to reset with a resistor instead of with a capacitor circuit. Pyserial holds it low by default, keeping the printer in reset.
The printer works with no code changes if you disassemble it, remove the resistor, and use command restart_method, but as I have several printers I tried to get a software fix working. I modeled it after how the cheetah uses RTS.
I included a config file that has the screen and led working, plus instructions for building and flashing.


securelyfitz:
Note: The tina2s is the only klipper printer I have. I haven't tested that my changes don't break other printers

Deathproof76:
Hejo, just wanted to say thank you! I'm using your PR, got everything working fully and even my 3dbenchy looks better! 

It's my first time using Klipper and also my first 3D-Printer so I'm still figuring everything out. Got just one question: There was a quiet mode setting using the original firmware which most likely used "stealthChop" mode of the TMC2208. I've seen that there are specific TMC2208 configs for the Kipper printer.cfg. Is it possible to enable them with the "as is" pin configuration of the Tina2S? I've looked a bit around in the [firmware github](https://github.com/weedo3d/TINA2Sfirmware/blob/12bf76c10caa969325ad130d9108e179535b0dec/Marlin/src/pins/pins_WEEDO_72H2.h#L26) but I'm not sure which pin would correspond to the uart-pin, or if I'm completely off in my understanding of the possibilities.

JamesH1978:

Also you need to add the regression tests as per point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Thanks
James

PR #6128, opened 2023-03-21
tmc2240: overheat_threshold

leptun:
Configurable overtemperature prewarning of the tmc2240 driver using `overtempprewarning_vth`.
This is used as an alternative to constant polling of `adc_temp`.


KevinOConnor:
Thanks.  As high-level feedback, I'm not sure it's a good idea to turn a warning into an error.  If the driver errors, then I agree Klipper needs to report an error, but I suspect almost all users would want some kind of a "pause" system on a heat warning and routing it to an error seems confusing to me.

Also, for what it is worth, I'm not sure the `TMC2240TemperatureHelper` abstraction makes the code more readable.

-Kevin

PR #6130, opened 2023-03-23
stm32: Enable I2C on stm32h7

bigtreetech:

* `i2c1_PB8_PB9` has tested on Octopus Pro H723 version.


KevinOConnor:
Thanks.

-Kevin

PR #6135, opened 2023-03-24
Add Creality Ender 3 S1 board to spi_flash

xand3r40r93:


sammcj:
Oo nice - I was wondering why the board on my E5S1 didn't work with the flash command and I suspect this is likely the fix as they use the same board.

sammcj:
Was there something wrong with this PR? It looked correct to me.

KevinOConnor:
Thanks.  Commits need to have a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html

Other than that, it just needs a contributor familiar with spi_flash.py to do a quick review and verify everything looks okay.

-Kevin

PR #6136, opened 2023-03-25
tmc: QUERY_TMC_REGISTER

leptun:
Add command to dump the fields of a single TMC register. This is a more lightweight version of the `DUMP_TMC` command. This command comes useful when performing manual calibration of TMC parameters. `DUMP_TMC` always displays more fields than are needed during a calibration. With this command, the response spam can be reduced.
Example response for `QUERY_TMC_REGISTER STEPPER=stepper_x REGISTER=DRV_STATUS` while doing a simple X axis move:
```
DRV_STATUS: 001d00f2 sg_result=242 cs_actual=29
```

This type of calibration where the user queries a register/field while moving the axis is needed for setting up CoolStep or for fine tuning StealthChop parameters by hand.


leptun:
Random thought, this change could also be implemented on top of the DUMP_TMC command with a REGISTER parameter instead of having a separate query command. Since I already implemented the query command, I'll leave it like this for now until I get some feedback.

KevinOConnor:
Thanks.  As high-level feedback, I think it is fine to limit the information reported by DUMP_TMC.  As mentioned though, I do think it would be better to limit the number of new commands.  So, I think it would be preferable to enhance the existing DUMP_TMC command.

Separately, I'm not sure why the new code skips the `read_translate()` step.

-Kevin

PR #6137, opened 2023-03-25
extras: Employ numpy for arc generation

vkhodygo:
A small improvement in the case someone needs to work with arcs of extreme size.

The code passes all the tests aside from one:
```
FileNotFoundError: [Errno 2] No such file or directory: 'dict/hc32f460-serial-PA3PA2.dict'
```
That doesn't seem to be related to the changes I've introduced.

### DIFF
--- a/klippy/extras/gcode_arcs.py
+++ b/klippy/extras/gcode_arcs.py
@@ -7,6 +7,7 @@
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import math
+import numpy as np
 
 # Coordinates created by this are converted into G1 commands.
 #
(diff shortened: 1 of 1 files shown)


thijstriemstra:
build is failing

flowerysong:
numpy is not a hard requirement for klipper, so the import should fail gracefully and return a useful error message. See https://github.com/Klipper3d/klipper/blob/fec7ddd8aefd9561fffd7f1ccecd7231002dc884/klippy/extras/angle.py#L26-L29 and https://github.com/Klipper3d/klipper/blob/fec7ddd8aefd9561fffd7f1ccecd7231002dc884/klippy/extras/shaper_calibrate.py#L64-L70 for the approaches taken by existing uses of numpy in extras. The requirement should also be noted in `docs/Config_Reference.md` and something like "`gcode_arcs` now uses numpy for calculations. Users that enable this feature will need to ensure that the numpy module is available." added to `docs/Config_Changes.md`.

It's unclear what precise improvement this change brings. At what point does the existing code fail? Is the failure to accommodate extreme arc sizes something that an average user might actually experience, or does it require special circumstances like an enlarged build volume?

PR #6138, opened 2023-03-25
Added AHT10 I2C temperature sensor support

scottmudge:
Added support for AHT10 ambient temperature sensor. Confirmed to work on BTT GTR 1.0 MCU, i2c port `i2c3`. 

Followed general implementation of other temperature sensors, being sure the sensor is not in busy state before reporting temperature/humidity.

Note, the AHT10 should come pre-calibrated and has limited I2C commands, so the interface is fairly simple. No RH/temp compensation is noted in the datasheet. 

If I can get ahold of AHT15/AHT20, I will add support for these as well. They should all be pretty similar, only needing to substitute some command bytes.

In order for the RH to be reported in Fluidd/Mainsail, it will need to also be added to their 'compatible' device list for RH reporting. Manually adding it to the compiled Fluidd js shows it works fine, however.

---

Squashed and force-pushed whitespace/line-length fixes after initial CI tests got mad about that, so that it keeps this to 1 commit, for keeping base repo clean on merge.

### DIFF
--- a/klippy/extras/aht10.py
+++ b/klippy/extras/aht10.py
@@ -1,0 +1,160 @@
+# AHT10 I2c-based humiditure sensor support
+#
+# Copyright (C) 2023 Scott Mudge <mail@scottmudge.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging
+from . import bus
+
+######################################################################
+# Compatible Sensors:
+#       AHT10      -    Tested w/ BTT GTR 1.0 MCU on i2c3
+######################################################################
+
+AHT10_I2C_ADDR= 0x38
+
+AHT10_COMMANDS = {
+    'INIT'              :[0xE1, 0x08, 0x00],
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  Seems fine to me, but commits need to have a signed-off-by line containing your full real name - see: https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

-Kevin

scottmudge:
@KevinOConnor  - Ah, sounds good. I've updated the commit message and force-pushed over the original commit.

KevinOConnor:
Thanks.

-Kevin

KevinOConnor:
I realized after committing this change that the docs/Config_Reference.md file was not updated.  Would you be able to submit the corresponding documentation change for this new module?

-Kevin

scottmudge:
Sure, no problem. I'll be busy today, but I'll try to get something in ASAP. Configuration should be about the same as BME280, substituting `sensor_type`, so should be pretty quick.

PR #6140, opened 2023-03-27
Provide standard build environment using docker

fredericmorin:
To compile klipper for a specific target:
```
./scripts/docker-with-builder-env.sh make menuconfig
./scripts/docker-with-builder-env.sh make -j6
```

To run CI tests:
```
./scripts/docker-with-builder-env.sh ./scripts/ci-build.sh
```

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -1,3 +1,61 @@
+# Find the run target below for an example showing how it's possible to install Klipper in Docker.
+
+# But first, here is the docker builder environment target. Use this to build klipper from any
+# platform without installing locally the required packages and compilers. This is especially usefull
+# on old linux distribution and on non-linux OS'es.
+# IMPORTANT: The docker build must be run from the root of the repo. To use docker builder env
+# either prefix your command with ./scripts/with-docker-builder-env.sh like this:
+#       ./scripts/with-docker-builder-env.sh make menuconfig
+# or invoke it manually like so:
+#       docker build . -f scripts/Dockerfile --target builder -t klipper-builder
+#       docker run -it --rm -u `id -u`:`id -g` -v `pwd`:`pwd`:rw -w `pwd` klipper-builder make menuconfig
+FROM ubuntu:20.04 as builder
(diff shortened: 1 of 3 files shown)


scripts/Dockerfile
```diff
@@ -1,3 +1,61 @@
...
+        flex \
+        gcc-arm-none-eabi \
+        gcc-avr \
+        libffi-dev \
+        libgmp-dev \
+        libmpc-dev \
+        libmpfr-dev \
+        libnewlib-arm-none-eabi \
+        pv \
+        python2-dev \
```
thijstriemstra, line 33: python2?

scripts/Dockerfile
```diff
@@ -1,3 +1,61 @@
...
+# Install gnupru toolchain
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        curl \
+        git \
+        wget \
+    && rm -rf /var/lib/apt/lists/*
+RUN cd /usr/local/src/ && \
+    git config --global user.email "you@example.com" && \
+    git config --global user.name "Your Name" && \
+    git clone https://github.com/dinuxbg/gnupru -b 2023.01 --depth 1 && \
```
thijstriemstra, line 46: does this need to be updated with every releaes?

fredericmorin (commented):
If I may add a bit of colour to my original PR description, this is a pattern we use ourselves at work, using docker to standardize the build environment of a project across dev machines and ci environment.

scripts/Dockerfile
thijstriemstra: > I would much prefer see this move to python3 instead

What is needed for that?

fredericmorin: Happy to look into that. Will give that a try.

KevinOConnor:
Thanks.  As high-level feedback, I'm not really sure what the intended audience of this change is.  That is, who will use it and what tasks will they accomplish with it.  As such, I'm not sure I'm able to review this PR.

Separately, the changes to .gitignore don't look correct.

-Kevin

PR #6141, opened 2023-03-27
i2c_software: Implementation of software i2c

bigtreetech:
Follow up https://github.com/Klipper3d/klipper/pull/6130
Software I2C tested on `STM32H7` with both `BME280` and `sh1106`, The actual test results are normal.

The following are some results captured by the logic analyzer.
* 100KHz i2c read
a3d2b996336782eda56e5e4d4d3e12cb
* 100KHz i2c write
e9e9c51ef7d78c9b4b3c7f0e34123bb4
* ~400KHz i2c read~
dd6d0a3ea9c1503cbac3354e6f4c6617

* ~400KHz i2c write~
ff0a17ebe4b980061b658321ed68b08a

# Note
* ~The maximum rate in the .c source file is limited to 1MHz, and the actual rate captured by the logic analyzer is between 800KHz-1MHz~ Speed fixed to 100KHz
* Tested on Mega2560, it can control i2c devices normally, but even without the `i2c_delay` function, the occupancy rate of the mcu is still 100%, so it is not recommended to use software i2c on MCU like Mega2560


thijstriemstra:
> Tested on Mega2560, it can control i2c devices normally, but even without the i2c_delay function, the occupancy rate of the mcu is still 100%, so it is not recommended to use software i2c on MCU like Mega2560

Any other test results? e.g an Octopus board.

src/i2c_software.c
```diff
@@ -0,0 +1,164 @@
...
+static void
+i2c_software_send_byte(struct i2c_software *is, uint8_t b)
+{
+    for (uint_fast8_t i = 0; i < 8; i++) {
+        gpio_out_write(is->sda, b & 0x80);
+        b <<= 1;
+        i2c_delay(is->ticks);
+        gpio_out_write(is->scl, 1);
+        i2c_delay(is->ticks);
+        gpio_out_write(is->scl, 0);
```
KevinOConnor, line 86: To communicate over an I2C bus, the transmitter should transition between GND and high-Z states.  This code is transitioning between GND and HIGH states.  If the transmitter transitions to a HIGH state while the receiver is in a GND state it would result in a "short" on the wires that would consume power, generate heat, and possibly destabilize the board voltage.

I have written code to communicate with I2C devices using just GND/HIGH transitions (eg, `klippy/extras/mcp4018.py`).  This change, though, may confuse users into thinking Klipper supports I2C on any device.  I'm not sure that is a good idea.

src/i2c_software.c
```diff
@@ -0,0 +1,164 @@
...
+    gpio_out_write(is->scl, 1);
+    i2c_delay(is->ticks);
+    gpio_out_write(is->scl, 0);
+}
+
+static uint8_t
+i2c_software_read_ack(struct i2c_software *is)
+{
+    uint8_t nack = 0;
+    struct gpio_in sda_in = gpio_in_setup(is->sda_pin, 1);
```
KevinOConnor, line 67: It is not valid to call `gpio_in_setup()` nor `gpio_out_setup()` multiple times.  (It may work on some architectures, but will not work on others.)  Use `gpio_out_reset()` and `gpio_in_reset()` instead.  See `src/tmcuart.c` as an example.

src/i2c_software.c
```diff
@@ -0,0 +1,164 @@
...
+};
+
+void
+command_i2c_set_software_bus(uint32_t *args)
+{
+    struct i2cdev_s *i2c = i2cdev_oid_lookup(args[0]);
+    struct i2c_software *is = alloc_chunk(sizeof(*is));
+    // Max speed = 1MHz
+    unsigned int speed = args[3] > 1000000 ? 1000000 : args[3];
+    is->ticks = 1000000000 / speed / 2;
```
KevinOConnor, line 31: Many MCUs don't implement a divide instruction.  This change would cause gcc to add software divide code to all MCUs that enable i2c support (eg, the AVR chips).  Is it necessary to support arbitrary speeds?  If only a speed of 100000 is needed then the code can be optimized so that the divide is done at compile time.

KevinOConnor (commented):
Thanks.  As high-level feedback, the functionality seems useful.  I have some comments on the implementation - see below.

Can you provide additional information on how this will be used?  Is there a device that needs this support and requires high speed I2C access?

-Kevin

src/i2c_software.c
bigtreetech: Fixed speed to 100Khz https://github.com/Klipper3d/klipper/pull/6141/commits/c80f4a1244e30845aa2a791adbcb679560aadd32, avoid division during runtime.

src/i2c_software.c
bigtreetech: Use `gpio_**_reset` instead of `gpio_**_setup`, just like in `src/tmcuart.c` .
https://github.com/Klipper3d/klipper/pull/6141/commits/55b5e928f03b6cd9d590c40cd11cea4d2e88c026

src/i2c_software.c
bigtreetech: Yeah, Theoretically this is a problem.
It would be perfect if there were a universal API to set GPIO as Open-Drain output, but some MCU (such as RP2040) do not seem to have the Open-Drain feature

KevinOConnor: Could the code transmit bits by switching between `gpio_in_reset(is->sda_in, 1)` and `gpio_out_reset(is->sda_out, 0)`?

I suspect the transfer rate would be slower, but it would hopefully not cause electrical problems.

KevinOConnor:
Thanks.  I guess my main question is - which users are expected to utilize this and what hardware is it expected that they will use it with?  Is this just general functionality that some users may find useful or is there a specific hardware setup that requires this functionality?

In general I prefer to only add code when we have a general idea of who will be using it and what they use it for.  (That way, I can get a better feel for the maintenance costs of the extra code relative to the functionality obtained by that new code.)

Cheers,
-Kevin

src/i2c_software.c
bigtreetech: Yeah, the transfer rate would be slower. Mainly affecting MCU with lower main frequencies such as STM32G0B1 64MHz. 
The measured results are as follows

1 | `gpio_out_write` | `gpio_in/out_reset`
-- | -- | --
STM32G0B1 | 70KHz  | 30-40KHz
RP2040 | 84KHz  | 70KHz

For MCUs with high main frequencies like `STM32H7`,  The impact is minimal.

bigtreetech:
> Thanks. I guess my main question is - which users are expected to utilize this and what hardware is it expected that they will use it with? Is this just general functionality that some users may find useful or is there a specific hardware setup that requires this functionality?
> 
> In general I prefer to only add code when we have a general idea of who will be using it and what they use it for. (That way, I can get a better feel for the maintenance costs of the extra code relative to the functionality obtained by that new code.)
> 
> Cheers, -Kevin

This requirement is raised by some of our customers who want to connect `BME280` temperature and humidity sensors to our motherboard (such as Octopus max ez). There is no hardware i2c port reserved on the motherboard, but there are many GPIO

KevinOConnor:
Okay, thanks.  In general, it looks fine to me.  I'll look to commit early next week.

-Kevin

KevinOConnor:
Thanks.

-Kevin

JamesH1978:
This looks to break smaller STM32 builds like the 042 with an error of

`  Linking out/klipper.elf
/usr/lib/gcc/arm-none-eabi/10.3.1/../../../arm-none-eabi/bin/ld: out/klipper.elf section `.text' will not fit in region `rom'
/usr/lib/gcc/arm-none-eabi/10.3.1/../../../arm-none-eabi/bin/ld: region `rom' overflowed by 904 bytes
collect2: error: ld returned 1 exit status` 

Seems to push it over the flash size

Thanks
James

@KevinOConnor @bigtreetech

JamesH1978:
seems changing line 10 of Kconfig from `select HAVE_GPIO_I2C if !MACH_STM32F031` to `select HAVE_GPIO_I2C if !MACH_STM32F031 && !MACH_STM32F042` fixes it at least for the 042, i havent tested other STM32F0XX mcu's

Lefuneste83:
> seems changing line 10 of Kconfig from `select HAVE_GPIO_I2C if !MACH_STM32F031` to `select HAVE_GPIO_I2C if !MACH_STM32F031 && !MACH_STM32F042` fixes it at least for the 042, i havent tested other STM32F0XX mcu's

This seems to allow compilation of FW for STM32F042. I have one in a [Voron 0 display board](https://github.com/VoronDesign/Voron-Hardware/tree/master/V0_Display). But the commit [645a1b](https://github.com/Klipper3d/klipper/commit/645a1b8364c3110f706db0f976ac5fa20b968c36) seems to break i2c communication nonetheless. I have made several attempts, including a complete reinstall of klipper, but they all failed and I had to rollback to previous commit, which works without a problem, both in compiling FW and communication with the display board.

```
Protocol error
  File "/home/pi/klipper/klippy/klippy.py", line 180, in _connect
    cb()
  File "/home/pi/klipper/klippy/mcu.py", line 753, in _connect
    self._send_config(None)
  File "/home/pi/klipper/klippy/mcu.py", line 683, in _send_config
    cb()
  File "/home/pi/klipper/klippy/extras/bus.py", line 177, in build_config
    self.i2c_write_cmd = self.mcu.lookup_command(
  File "/home/pi/klipper/klippy/mcu.py", line 873, in lookup_command
    return CommandWrapper(self._serial, msgformat, cq)
  File "/home/pi/klipper/klippy/mcu.py", line 84, in __init__
    self._cmd = msgparser.lookup_command(msgformat)
  File "/home/pi/klipper/klippy/msgproto.py", line 315, in lookup_command
    self._error("Unknown command: %s", msgname)
  File "/home/pi/klipper/klippy/msgproto.py", line 243, in _error
msgproto.error: mcu 'display': Unknown command: i2c_write
```

KevinOConnor:
See #6244 for code to optionally disable software i2c (and other features) on chips with small flash sizes.

-Kevin

JamesH1978:
This also seems to break some rp2040 boards, at least a couple of SB2040 mellow owners have tmc uart issues on this commit, that i have seen on the discord

Thanks
James

Sineos:
Also ref https://klipper.discourse.group/t/possible-bug-in-recent-commit/8921

PR #6144, opened 2023-03-28
tmc: DUMP_TMC optional REGISTER parameter

leptun:
Add an optional REGISTER parameter to DUMP_TMC so that the output is more filtered/cleaner for manual TMC calibration.
This is a follow-up of #6136.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1219,8 +1219,9 @@ The following commands are available when any of the
 are enabled.
 
 #### DUMP_TMC
-`DUMP_TMC STEPPER=<name>`: This command will read the TMC driver
-registers and report their values.
+`DUMP_TMC STEPPER=<name>` [REGISTER=<name>]: This command will read all TMC
+driver registers and report their values. If a REGISTER is provided, only
+the specified register will be dumped.
 
 #### INIT_TMC
 `INIT_TMC STEPPER=<name>`: This command will initialize the TMC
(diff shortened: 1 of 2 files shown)


masterxq:
Hi, I have mixed drivers, `tmc5160` and `tmc2209`. With the version including this patch I get the following error when I try to move the stepper with the tmc2209 driver or calling `dump_tmc stepper=stepper_z`:

```'Unable to read tmc uart 'stepper_z' register GCONF'```

Downgrading below this patch fixes the issue. So maybe it is related. (Did only check once)

Sadly this is just what I remember from the tests I did 2 days ago, but possibly it is easy to trace it down, and it helps to prevent other users from running into this issue.

If additional testing is needed I can try to reproduce this in some days.

leptun:
I don't see how the change could cause the driver to fail to read the GCONF register. That register exists and is readable, so it should not error out. Can you please try switching to before and after this change a few more times to confirm that the issue indeed only happens after this PR? To me this sounds a lot like an intermittent connection issue with the driver caused by hardware, not software.

masterxq:
> I don't see how the change could cause the driver to fail to read the GCONF register. That register exists and is readable, so it should not error out. Can you please try switching to before and after this change a few more times to confirm that the issue indeed only happens after this PR? To me this sounds a lot like an intermittent connection issue with the driver caused by hardware, not software.

Ok, this may have many other reasons. This patch was just one that seems likely, as on repo top level was no other patches touching TMC driver at all expect this one. But a Hardware issue is at least unlikely, but it still possible that my drivers was in a bad state or something similar.

You don't need to use more time on this, I will report back the next time I updated. If the issue remains I will **double**-check your patch (go just below it and include this patch only), and tell you what I found or not found ^^

Best regards

masterxq:
I updated to the latest version, and I could not reproduce the problem. So I think my previous comment can be ignored. Sorry for wasting your time.

PR #6149, opened 2023-04-02
Add X twist compensation module

blastrock:
Hi,

This is a fork of the PR #6048. This is a feature I'd very like to see merged, so I took the liberty of taking the commits of @koonweee and fixing the comments made on his PR.

I can squash my commits if you prefer a cleaner history.

The changes include:
- A few bug fixes
- Some typos pointed out in the review
- Removal of the profile system, the rationale is that the twist only depends on the printer, not the temperature etc
- Rename of the module to axis_twist_compensation so that if we need some day to calibrate for another axis, there will be less changes to make to the user's configuration

On another note, I kept the recommended z offset feature from the original PR, which is also present in the Marlin implementation. This z offset has always been off in my tests. I also don't understand how making the average of the z offset from left to right makes sense to calculate a single z offset. The printer will only probe the middle of the bed when homing anyway, so that's the only offset that we should apply.

Also, a small difference from the original PR is that since I removed the profile system, there is always a configuration loaded when klipper starts.

TODO:

- [x] documentation (still unmodified from original PR, waiting for a first review)

### DIFF
--- a/docs/Axis_Twist_Compensation.md
+++ b/docs/Axis_Twist_Compensation.md
@@ -25,8 +25,8 @@ first use mechanical means to fix it prior to applying software corrections.
 perform `AXIS_TWIST_COMPENSATION_CALIBRATE`
 * The calibration wizard will prompt you to measure the probe Z offset at a few
 points along the bed
-* The calibration defaults to 3 points but you can use the option `N_POINTS=` to
-use a different number.
+* The calibration defaults to 3 points but you can use the option
+`SAMPLE_COUNT=` to use a different number.
 2. [Adjust your Z offset](Probe_Calibrate.md#calibrating-probe-z-offset)
 3. Perform automatic/probe-based bed tramming operations, such as
 [Screws Tilt Adjust](G-Codes.md#screws_tilt_adjust),
--- a/docs/Axis_Twist_Compensation.md
+++ b/docs/Axis_Twist_Compensation.md
@@ -16,6 +16,9 @@ This module uses manual measurements by the user to correct the probe's results.
 Note that if your axis is significantly twisted it is strongly recommended to
 first use mechanical means to fix it prior to applying software corrections.
 
+**Warning**: This module is not compatible with dockable probes yet and will
+try to probe the bed without attaching the probe if you use it.
+
 ## Overview of compensation usage
 
 > **Tip:** Make sure the [probe X and Y offsets](Config_Reference.md#probe) are
--- a/docs/Axis_Twist_Compensation.md
+++ b/docs/Axis_Twist_Compensation.md
@@ -2,8 +2,8 @@
 
 This document describes the [axis_twist_compensation] module.
 
-Some printers exhibit different Z offsets depending on the X position of the
-probed point, and this is independent of the bed being not flat or not trammed.
+Some printers may have a small twist in their X rail which can skew the results
+of a probe attached to the X carriage.
 This is common in printers with designs like the Prusa MK3, Sovol SV06 etc and is
 further described under [probe location
 bias](Probe_Calibrate.md#location-bias-check). It may result in
@@ -20,9 +20,6 @@ first use mechanical means to fix it prior to applying software corrections.
 
 > **Tip:** Make sure the [probe X and Y offsets](Config_Reference.md#probe) are
 > correctly set as they greatly influence calibration.
-
-> **Tip:** A feeler gauge is recommended in place of paper for more accurate
-> measurements
 
 1. After setting up the [axis_twist_compensation] module,
 perform `AXIS_TWIST_COMPENSATION_CALIBRATE`
@@ -41,9 +38,9 @@ use a different number.
 > **Tip:** Bed temperature and nozzle temperature and size do not seem to have
 > an influence to the calibration process.
 
-> **Tip:** Disable the Axis twist compensation using `AXIS_TWIST_PROFILE_CLEAR`
-> and perform Screws Tilt Adjust/Z Tilt Adjust/Bed Mesh to compare the
-> difference
+> **Tip:** Disable the Axis twist compensation using
+> `AXIS_TWIST_COMPENSATION_CLEAR` and perform Screws Tilt Adjust/Z Tilt
+> Adjust/Bed Mesh to compare the difference
 
 ## [axis_twist_compensation] setup and commands
 
@@ -69,25 +66,3 @@ interpolate linearly between each pair of points.
 
 > **Tip:** It is not necessary to redo a calibration when switching between
 > linear and multilinear.
-
-### Fine-tuning
-
-If required, the user may fine tune the `z_compensations` values.
-
-For example, the value of `0.12` represents the `z_compensation` applied
-when the printer is probing at `start_x`. This means that if the non-compensated
-z-height measured at `start_x` was `2.0`, the compensated z-height is now `2.0 +
-0.12 = 2.12`, resulting in a higher point on the recorded mesh.
-
-Should the squish on the side of the bed represented by `start_x` be
-insufficient, the user may specify a lower `z_compensation` value in place of
-`0.12`, such as `0.02`. This would result in the mesh being lowered by
-`0.12 - 0.02 = 0.1mm` at points along `start_x`.
-
-## Reference
-
-The calculations for this module are largely based off those found in
-[Marlin PR #23238](https://github.com/MarlinFirmware/Marlin/pull/23238).
-
-A visualization of the issue can be found in
-[this discussion thread](https://github.com/MarlinFirmware/Marlin/issues/22791).
--- a/docs/Axis_Twist_Compensation.md
+++ b/docs/Axis_Twist_Compensation.md
@@ -45,20 +45,3 @@ Configuration options for [axis_twist_compensation] can be found in the
 
 Commands for [axis_twist_compensation] can be found in the
 [G-Codes Reference](G-Codes.md#axis_twist_compensation)
-
-### Type of compensation
-
-This module supports two interpolation methods for the Z offset compensation:
-linear and multilinear.
-
-Some printers which have a slight twist only require a linear compensation, so
-you can use the 'linear' mode to do a linear regression to fit a line through
-the measured compensation.
-
-It is not uncommon for printers to have multiple combined non-linear twists
-(e.g. the tool head leaning forward in the middle of the axis due to its own
-weight). For that, it is recommended to use the 'multilinear' mode which will
-interpolate linearly between each pair of points.
-
-> **Tip:** It is not necessary to redo a calibration when switching between
-> linear and multilinear.
--- a/docs/Axis_Twist_Compensation.md
+++ b/docs/Axis_Twist_Compensation.md
@@ -38,10 +38,6 @@ points along the bed
 > **Tip:** Bed temperature and nozzle temperature and size do not seem to have
 > an influence to the calibration process.
 
-> **Tip:** Disable the Axis twist compensation using
-> `AXIS_TWIST_COMPENSATION_CLEAR` and perform Screws Tilt Adjust/Z Tilt
-> Adjust/Bed Mesh to compare the difference
-
 ## [axis_twist_compensation] setup and commands
 
 Configuration options for [axis_twist_compensation] can be found in the
--- a/docs/Axis_Twist_Compensation.md
+++ b/docs/Axis_Twist_Compensation.md
@@ -1,0 +1,93 @@
+# Axis Twist Compensation
+
+This document describes the [axis_twist_compensation] module.
+
+Some printers exhibit different Z offsets depending on the X position of the
+probed point, and this is independent of the bed being not flat or not trammed.
+This is common in printers with designs like the Prusa MK3, Sovol SV06 etc and is
+further described under [probe location
+bias](Probe_Calibrate.md#location-bias-check). It may result in
+probe operations such as [Bed Mesh](Bed_Mesh.md),
+[Screws Tilt Adjust](G-Codes.md#screws_tilt_adjust),
+[Z Tilt Adjust](G-Codes.md#z_tilt_adjust) etc returning inaccurate
+representations of the bed.
+
+This module uses manual measurements by the user to correct the probe's results.
+Note that if your axis is significantly twisted it is strongly recommended to
+first use mechanical means to fix it prior to applying software corrections.
+
+## Overview of compensation usage
+
+> **Tip:** Make sure the [probe X and Y offsets](Config_Reference.md#probe) are
+> correctly set as they greatly influence calibration.
+
+> **Tip:** A feeler gauge is recommended in place of paper for more accurate
+> measurements
+
+1. After setting up the [axis_twist_compensation] module,
+perform `AXIS_TWIST_COMPENSATION_CALIBRATE`
+* The calibration wizard will prompt you to measure the probe Z offset at a few
+points along the bed
+* The calibration defaults to 3 points but you can use the option `N_POINTS=` to
+use a different number.
+2. [Adjust your Z offset](Probe_Calibrate.md#calibrating-probe-z-offset)
+3. Perform automatic/probe-based bed tramming operations, such as
+[Screws Tilt Adjust](G-Codes.md#screws_tilt_adjust),
+[Z Tilt Adjust](G-Codes.md#z_tilt_adjust) etc
+4. Home all axis, then perform a [Bed Mesh](Bed_Mesh.md) if required
+5. Perform a test print, followed by any
+[fine-tuning](Axis_Twist_Compensation.md#fine-tuning) as desired
+
+> **Tip:** Bed temperature and nozzle temperature and size do not seem to have
+> an influence to the calibration process.
+
+> **Tip:** Disable the Axis twist compensation using `AXIS_TWIST_PROFILE_CLEAR`
+> and perform Screws Tilt Adjust/Z Tilt Adjust/Bed Mesh to compare the
+> difference
+
+## [axis_twist_compensation] setup and commands
+
+Configuration options for [axis_twist_compensation] can be found in the
+[Configuration Reference](Config_Reference.md#axis_twist_compensation).
+
+Commands for [axis_twist_compensation] can be found in the
+[G-Codes Reference](G-Codes.md#axis_twist_compensation)
+
+### Type of compensation
+
+This module supports two interpolation methods for the Z offset compensation:
+linear and multilinear.
+
+Some printers which have a slight twist only require a linear compensation, so
+you can use the 'linear' mode to do a linear regression to fit a line through
+the measured compensation.
+
+It is not uncommon for printers to have multiple combined non-linear twists
+(e.g. the tool head leaning forward in the middle of the axis due to its own
+weight). For that, it is recommended to use the 'multilinear' mode which will
+interpolate linearly between each pair of points.
+
+> **Tip:** It is not necessary to redo a calibration when switching between
+> linear and multilinear.
+
+### Fine-tuning
+
+If required, the user may fine tune the `z_compensations` values.
+
+For example, the value of `0.12` represents the `z_compensation` applied
+when the printer is probing at `start_x`. This means that if the non-compensated
+z-height measured at `start_x` was `2.0`, the compensated z-height is now `2.0 +
+0.12 = 2.12`, resulting in a higher point on the recorded mesh.
+
+Should the squish on the side of the bed represented by `start_x` be
+insufficient, the user may specify a lower `z_compensation` value in place of
+`0.12`, such as `0.02`. This would result in the mesh being lowered by
+`0.12 - 0.02 = 0.1mm` at points along `start_x`.
+
+## Reference
+
+The calculations for this module are largely based off those found in
+[Marlin PR #23238](https://github.com/MarlinFirmware/Marlin/pull/23238).
+
+A visualization of the issue can be found in
+[this discussion thread](https://github.com/MarlinFirmware/Marlin/issues/22791).
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1326,9 +1326,6 @@ section](Config_Reference.md#axis_twist_compensation) is enabled.
 twist calibration wizard. `SAMPLE_COUNT` specifies the number of points along
 the X axis to calibrate at and defaults to 3.
 
-#### AXIS_TWIST_COMPENSATION_CLEAR
-`AXIS_TWIST_COMPENSATION_CLEAR`: Clears the current compensation values.
-
 ### [z_thermal_adjust]
 
 The following commands are available when the
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1322,9 +1322,9 @@ The following commands are available when the
 section](Config_Reference.md#axis_twist_compensation) is enabled.
 
 #### AXIS_TWIST_COMPENSATION_CALIBRATE
-`AXIS_TWIST_COMPENSATION_CALIBRATE [N_POINTS=<value>]`: Initiates the X twist calibration
-wizard. `N_POINTS` specifies the number of points along the X axis to calibrate
-at and defaults to 3.
+`AXIS_TWIST_COMPENSATION_CALIBRATE [SAMPLE_COUNT=<value>]`: Initiates the X
+twist calibration wizard. `SAMPLE_COUNT` specifies the number of points along
+the X axis to calibrate at and defaults to 3.
 
 #### AXIS_TWIST_COMPENSATION_CLEAR
 `AXIS_TWIST_COMPENSATION_CLEAR`: Clears the current compensation values.
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1315,6 +1315,20 @@ print.
 #### SDCARD_RESET_FILE
 `SDCARD_RESET_FILE`: Unload file and clear SD state.
 
+### [axis_twist_compensation]
+
+The following commands are available when the
+[axis_twist_compensation config
+section](Config_Reference.md#axis_twist_compensation) is enabled.
+
+#### AXIS_TWIST_COMPENSATION_CALIBRATE
+`AXIS_TWIST_COMPENSATION_CALIBRATE [N_POINTS=<value>]`: Initiates the X twist calibration
+wizard. `N_POINTS` specifies the number of points along the X axis to calibrate
+at and defaults to 3.
+
+#### AXIS_TWIST_COMPENSATION_CLEAR
+`AXIS_TWIST_COMPENSATION_CLEAR`: Clears the current compensation values.
+
 ### [z_thermal_adjust]
 
 The following commands are available when the
--- a/klippy/extras/axis_twist_compensation.py
+++ b/klippy/extras/axis_twist_compensation.py
@@ -1,0 +1,347 @@
+# Axis Twist Compensation
+
+# Copyright (C) 2022  Jeremy Tan <jeremytkw98@gmail.com>
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+"""
+[axis_twist_compensation]
+horizontal_move_z: 5
+speed: 50
+start_x: 0 ; nozzle's x coordinate at the start of the calibration ! required
+end_x: 200 ; nozzle's x coordinate at the end of the calibration ! required
+y: 100 ; nozzle's y coordinate during the calibration ! required
+type: multilinear
+"""
+
+import logging
+import math
+from . import manual_probe as ManualProbe, bed_mesh as BedMesh
+
+
+DEFAULT_N_POINTS = 3
+
+
+class Config:
+    DEFAULT_SPEED = 50.
+    DEFAULT_HORIZONTAL_MOVE_Z = 5.
+    REQUIRED = True
+    OPTIONAL = False
+    CONFIG_OPTIONS = {
+        'horizontal_move_z': (float, OPTIONAL, DEFAULT_HORIZONTAL_MOVE_Z),
+        'speed': (float, OPTIONAL, DEFAULT_SPEED),
+        'start_x': (float, REQUIRED, None),
+        'end_x': (float, REQUIRED, None),
+        'y': (float, REQUIRED, None),
+        'type': (str, OPTIONAL, 'multilinear'),
+        'z_compensations': (str, OPTIONAL, None),
+    }
+
+    @staticmethod
+    def save_to_config(name, z_compensations, configfile):
+        values_as_str = ', '.join([Helpers.format_float_to_n_decimals(x)
+                                   for x in z_compensations])
+        configfile.set(name, 'z_compensations', values_as_str)
+
+
+class AxisTwistCompensation:
+    def __init__(self, config):
+        # get printer
+        self.config = config
+        self.printer = config.get_printer()
+        self.gcode = self.printer.lookup_object('gcode')
+
+        # get values from [axis_twist_compensation] section in printer .cfg
+        for config_key, \
+            (config_type, required, default) in Config.CONFIG_OPTIONS.items():
+            value = None
+            if config_type == float:
+                value = config.getfloat(config_key, default)
+            elif config_key == 'z_compensations':
+                value = Helpers.parse_comma_separated_floats(
+                    config.get('z_compensations', default=""))
+            else:
+                value = config.get(config_key, default)
+            if required and value is None:
+                raise config.error(
+                    "Missing required config option for section [{}]: {}"
+                    .format(config.get_name(), config_key))
+            setattr(self, config_key, value)
+
+        if self.type not in ['linear', 'multilinear']:
+                raise config.error(
+                    "Invalid axis_twist_compensation.type value: {}"
+                    .format(self.type))
+
+        self.m = None
+        self.b = None
+
+        # setup calibrater
+        self.calibrater = Calibrater(self)
+
+        self._register_gcode_handlers()
+
+    def get_z_compensation_value(self, x_coord):
+        if not self.z_compensations:
+            return 0
+        self._ensure_init()
+
+        if self.type == 'linear':
+            return self._get_z_compensation_value_linear(x_coord)
+        elif self.type == 'multilinear':
+            return self._get_z_compensation_value_multilinear(x_coord)
+
+    def _get_z_compensation_value_linear(self, x_coord):
+        return self.m * x_coord + self.b
+
+
+    def _get_z_compensation_value_multilinear(self, x_coord):
+        z_compensations = self.z_compensations
+        n_points = len(z_compensations)
+        spacing = (self.end_x - self.start_x) / (n_points - 1)
+        interpolate_t = (x_coord - self.start_x) / spacing
+        interpolate_i = int(math.floor(interpolate_t))
+        interpolate_i = BedMesh.constrain(interpolate_i, 0, n_points - 2)
+        interpolate_t -= interpolate_i
+        interpolated_z_compensation = BedMesh.lerp(
+            interpolate_t, z_compensations[interpolate_i],
+            z_compensations[interpolate_i + 1])
+        return interpolated_z_compensation
+
+    def _ensure_init(self):
+        if self.type == 'linear' and self.m is None:
+            z_compensations = self.z_compensations
+            n_points = len(z_compensations)
+            interval_dist = \
+                (self.end_x - self.start_x) / (n_points - 1)
+            indexes = [self.start_x + i*interval_dist
+                       for i in range(0, n_points)]
+
+            # Calculate the mean of x and y values
+            mean_indexes = sum(indexes) / n_points
+            mean_z_compensations = sum(z_compensations) / n_points
+
+            # Calculate the covariance and variance
+            covar = sum((indexes[i] - mean_indexes) *
+                        (z_compensations[i] - mean_z_compensations)
+                        for i in range(n_points))
+            var = sum((indexes[i] - mean_indexes)**2 for i in range(n_points))
+
+            # Compute the slope (m) and intercept (b) of the best-fit line
+            self.m = covar / var
+            self.b = mean_z_compensations - self.m * mean_indexes
+
+    def clear_compensations(self):
+        self.z_compensations = []
+        self.m = None
+        self.b = None
+
+    def _register_gcode_handlers(self):
+        self.gcode.register_command(
+            'AXIS_TWIST_COMPENSATION_CLEAR',
+            self.cmd_AXIS_TWIST_COMPENSATION_CLEAR,
+            desc=self.cmd_AXIS_TWIST_COMPENSATION_CLEAR_help)
+
+    cmd_AXIS_TWIST_COMPENSATION_CLEAR_help = \
+        "Clears the active axis twist compensation"
+
+    def cmd_AXIS_TWIST_COMPENSATION_CLEAR(self, gcmd):
+        self.clear_compensations()
+
+
+class Calibrater:
+    def __init__(self, compensation):
+        # setup self attributes
+        self.compensation = compensation
+        self.printer = compensation.printer
+        self.config = compensation.config
+        self.gcode = self.printer.lookup_object('gcode')
+        self.probe = None
+        # probe settings are set to none, until they are available
+        self.lift_speed, self.probe_x_offset, self.probe_y_offset, _ = \
+            None, None, None, None
+        self.printer.register_event_handler("klippy:connect",
+                                            self._handle_connect())
+        self.speed = compensation.speed
+        self.horizontal_move_z = compensation.horizontal_move_z
+        self.start_point = (compensation.start_x, compensation.y)
+        self.end_point = (compensation.end_x, compensation.y)
+        self.results = None
+        self.current_point_index = None
+        self.gcmd = None
+
+        # register gcode handlers
+        self._register_gcode_handlers()
+
+    def _handle_connect(self):
+        # gets probe settings when they are available
+        def callback():
+            self.probe = self.printer.lookup_object('probe', None)
+            if (self.probe is None):
+                raise self.config.error(
+                    "AXIS_TWIST_COMPENSATION requires [probe] to be defined")
+            self.lift_speed = self.probe.get_lift_speed()
+            self.probe_x_offset, self.probe_y_offset, _ = \
+                self.probe.get_offsets()
+        return callback
+
+    def _register_gcode_handlers(self):
+        # register gcode handlers
+        self.gcode = self.printer.lookup_object('gcode')
+        self.gcode.register_command(
+            'AXIS_TWIST_COMPENSATION_CALIBRATE',
+            self.cmd_AXIS_TWIST_COMPENSATION_CALIBRATE,
+            desc=self.cmd_AXIS_TWIST_COMPENSATION_CALIBRATE_help)
+
+    cmd_AXIS_TWIST_COMPENSATION_CALIBRATE_help = """
+    Performs the x twist calibration wizard
+    Measure z probe offset at n points along the x axis,
+    and calculate x twist compensation
+    """
+
+    def cmd_AXIS_TWIST_COMPENSATION_CALIBRATE(self, gcmd):
+        self.gcmd = gcmd
+        n_points = gcmd.get_int('N_POINTS', DEFAULT_N_POINTS)
+
+        # check for valid n_points
+        if n_points is None or n_points < 2:
+            raise self.gcmd.error(
+                "N_POINTS to probe must be at least 2")
+
+        # clear the current config
+        self.compensation.clear_compensations()
+
+        # calculate some values
+        x_range = self.end_point[0] - self.start_point[0]
+        interval_dist = x_range / (n_points - 1)
+        nozzle_points = self._calculate_nozzle_points(n_points, interval_dist)
+        probe_points = self._calculate_probe_points(
+            nozzle_points, self.probe_x_offset, self.probe_y_offset)
+
+        # verify no other manual probe is in progress
+        ManualProbe.verify_no_manual_probe(self.printer)
+
+        # begin calibration
+        self.current_point_index = 0
+        self.results = []
+        self._calibration(probe_points, nozzle_points, interval_dist)
+
+    def _calculate_nozzle_points(self, n_points, interval_dist):
+        # calculate the points to put the probe at, returned as a list of tuples
+        nozzle_points = []
+        for i in range(n_points):
+            x = self.start_point[0] + i * interval_dist
+            y = self.start_point[1]
+            nozzle_points.append((x, y))
+        return nozzle_points
+
+    def _calculate_probe_points(self, nozzle_points,
+        probe_x_offset, probe_y_offset):
+        # calculate the points to put the nozzle at
+        # returned as a list of tuples
+        probe_points = []
+        for point in nozzle_points:
+            x = point[0] - probe_x_offset
+            y = point[1] - probe_y_offset
+            probe_points.append((x, y))
+        return probe_points
+
+    def _move_helper(self, target_coordinates, override_speed=None):
+        # pad target coordinates
+        target_coordinates = \
+            (target_coordinates[0], target_coordinates[1], None) \
+            if len(target_coordinates) == 2 else target_coordinates
+        toolhead = self.printer.lookup_object('toolhead')
+        speed = self.speed if target_coordinates[2] == None else self.lift_speed
+        speed = override_speed if override_speed is not None else speed
+        toolhead.manual_move(target_coordinates, speed)
+
+    def _calibration(self, probe_points, nozzle_points, interval):
+        # begin the calibration process
+        self.gcmd.respond_info("AXIS_TWIST_COMPENSATION_CALIBRATE: "
+                               "Probing point %d of %d" % (
+                                   self.current_point_index + 1,
+                                   len(probe_points)))
+
+        # horizontal_move_z (to prevent probe trigger or hitting bed)
+        self._move_helper((None, None, self.horizontal_move_z))
+
+        # move to point to probe
+        self._move_helper((probe_points[self.current_point_index][0],
+                           probe_points[self.current_point_index][1], None))
+
+        # probe the point
+        self.current_measured_z = self.probe.run_probe(self.gcmd)[2]
+
+        # horizontal_move_z (to prevent probe trigger or hitting bed)
+        self._move_helper((None, None, self.horizontal_move_z))
+
+        # move the nozzle over the probe point
+        self._move_helper((nozzle_points[self.current_point_index]))
+
+        # start the manual (nozzle) probe
+        ManualProbe.ManualProbeHelper(
+            self.printer, self.gcmd,
+            self._manual_probe_callback_factory(
+                probe_points, nozzle_points, interval))
+
+    def _manual_probe_callback_factory(self, probe_points,
+        nozzle_points, interval):
+        # returns a callback function for the manual probe
+        is_end = self.current_point_index == len(probe_points) - 1
+
+        def callback(kin_pos):
+            if kin_pos is None:
+                # probe was cancelled
+                self.gcmd.respond_info(
+                    "AXIS_TWIST_COMPENSATION_CALIBRATE: Probe cancelled, "
+                    "calibration aborted")
+                return
+            z_offset = self.current_measured_z - kin_pos[2]
+            self.results.append(z_offset)
+            if is_end:
+                # end of calibration
+                self._finalize_calibration()
+            else:
+                # move to next point
+                self.current_point_index += 1
+                self._calibration(probe_points, nozzle_points, interval)
+        return callback
+
+    def _finalize_calibration(self):
+        # finalize the calibration process
+        # calculate average of results
+        avg = sum(self.results) / len(self.results)
+        # subtract average from each result
+        # so that they are independent of z_offset
+        self.results = [avg - x for x in self.results]
+        # save the config
+        configfile = self.printer.lookup_object('configfile')
+        Config.save_to_config(self.config.get_name(), self.results, configfile)
+        self.gcode.respond_info(
+            "AXIS_TWIST_COMPENSATION state has been saved "
+            "for the current session.  The SAVE_CONFIG command will "
+            "update the printer config file and restart the printer.")
+        # output result
+        self.gcmd.respond_info(
+            "AXIS_TWIST_COMPENSATION_CALIBRATE: Calibration complete, "
+            "offsets: %s, mean z_offset: %f"
+            % (self.results, avg))
+
+
+class Helpers:
+    @staticmethod
+    def format_float_to_n_decimals(raw_float, n=6):
+        # format float to n decimals, defaults to 6
+        return "{:.{}f}".format(raw_float, n)
+
+    @staticmethod
+    def parse_comma_separated_floats(comma_separated_floats):
+        if not comma_separated_floats:
+            return []
+        # parse comma separated floats into list of floats
+        return [float(value) for value in comma_separated_floats.split(', ')]
+
+# klipper's entry point using [axis_twist_compensation] section in printer.cfg
+
+def load_config(config):
+    return AxisTwistCompensation(config)
--- a/klippy/extras/axis_twist_compensation.py
+++ b/klippy/extras/axis_twist_compensation.py
@@ -14,29 +14,16 @@ y: 100 ; nozzle's y coordinate during the calibration ! required
 type: multilinear
 """
 
-import logging
 import math
 from . import manual_probe as ManualProbe, bed_mesh as BedMesh
 
 
 DEFAULT_N_POINTS = 3
+DEFAULT_SPEED = 50.
+DEFAULT_HORIZONTAL_MOVE_Z = 5.
 
 
 class Config:
-    DEFAULT_SPEED = 50.
-    DEFAULT_HORIZONTAL_MOVE_Z = 5.
-    REQUIRED = True
-    OPTIONAL = False
-    CONFIG_OPTIONS = {
-        'horizontal_move_z': (float, OPTIONAL, DEFAULT_HORIZONTAL_MOVE_Z),
-        'speed': (float, OPTIONAL, DEFAULT_SPEED),
-        'start_x': (float, REQUIRED, None),
-        'end_x': (float, REQUIRED, None),
-        'y': (float, REQUIRED, None),
-        'type': (str, OPTIONAL, 'multilinear'),
-        'z_compensations': (str, OPTIONAL, None),
-    }
-
     @staticmethod
     def save_to_config(name, z_compensations, configfile):
         values_as_str = ', '.join([Helpers.format_float_to_n_decimals(x)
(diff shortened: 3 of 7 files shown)


koonweee:
Hey! Thanks for continuing development of this. I have been using the original implementation on my sv06 and been super happy with it, but would love to see this polished up for klipper of course. I will not have time until after my exams in May to work on this, but will be glad to review and assist here where possible 👍

blastrock:
Thanks a lot for your work on this, I've been using this feature for a month on my sv06, it really helped! Good luck on your exams :)

HelloThisIsFlo:
Fantastic work @koonweee, and @blastrock thanks a lot for the updates!

I haven't tested it yet, but I will very soon.

I do have 2 questions:
1. Why the requirement to have at least 3 points?
    Reading the algorithm, it's only interpolating points 2by2. So, it would seem the more points, the more possibility of measurement error. With the current logic, 2 points would be sufficient, and actually preferable, no? Or am I missing something?
2. As a follow-up to `1.`, instead of interpolating on 2 points only, could we use linear regression to fit the best straight line?
    Again, I'm not sure I fully understand the concept of this z adjustment. So, just to make sure I follow: we are trying to fit a straight line, correct? That straight line is the representation of the x asis gantry, correct? Or did I misunderstand?
    
If I did misunderstand the logic and you don't feel like re-explaining everything (which is fair), I'd be happy if you could point me to some resources for further reading 🙂

Thanks again for your amazing work 👍

blastrock:
> Why the requirement to have at least 3 points?

I'd have to check, but I think it comes from the Marlin implementation. I'm pretty sure it would work with 2 points too, as you suggest.

> As a follow-up to 1., instead of interpolating on 2 points only, could we use linear regression to fit the best straight line?

You are assuming that the twist leads to a linear z offset correction, and I did too!

I studied the problem the best as I could, and noticed there are 3 twists combining on my SV06:

1. The print head is not the same orientation on the left and on the right of the printer and rotates around the X axis. That's the first intuition people have, and it's easy to notice it by holding a level on the print head and moving it. This one is probably linear.
2. When the print head is in the middle of the axis, it leans forward with its own weight. So it rotates around the X axis in one direction until the middle where it rotates back in the other direction. This one is definitely not linear.
3. This last one is hard to believe, and I'm even not sure I reached the right conclusion, but it looks like the print head rotates ever so slightly around the Y axis when traveling from left to right, kind of following a ◠ shape. So, the probe being to the right of the nozzle, it's offset too far from the bed when the print head is on the left, and too close when it's on the right. This last one might be linear? But it's around the Y axis, I'm not even sure how it combines with the others.

On my printer, I currently have the following correction offsets: `0.064583, -0.014167, -0.050417`, which makes it quite non-linear ^^

Jayksopp:
> > Why the requirement to have at least 3 points?
> 
> I'd have to check, but I think it comes from the Marlin implementation. I'm pretty sure it would work with 2 points too, as you suggest.
> 
> > As a follow-up to 1., instead of interpolating on 2 points only, could we use linear regression to fit the best straight line?
> 
> You are assuming that the twist leads to a linear z offset correction, and I did too!
> 
> I studied the problem the best as I could, and noticed there are 3 twists combining on my SV06:
> 
> 1. The print head is not the same orientation on the left and on the right of the printer and rotates around the X axis. That's the first intuition people have, and it's easy to notice it by holding a level on the print head and moving it. This one is probably linear.
> 2. When the print head is in the middle of the axis, it leans forward with its own weight. So it rotates around the X axis in one direction until the middle where it rotates back in the other direction. This one is definitely not linear.
> 3. This last one is hard to believe, and I'm even not sure I reached the right conclusion, but it looks like the print head rotates ever so slightly around the Y axis when traveling from left to right, kind of following a ◠ shape. So, the probe being to the right of the nozzle, it's offset too far from the bed when the print head is on the left, and too close when it's on the right. This last one might be linear? But it's around the Y axis, I'm not even sure how it combines with the others.
> 
> On my printer, I currently have the following correction offsets: `0.064583, -0.014167, -0.050417`, which makes it quite non-linear ^^

This is very interesting. On my SV06 the offset is pretty linear. I did some measurements with a daily gauge and a mount at the nozzle and at the probe position. 

If you look at the attached picture you can see the blue line where I have significant twist its pretty flat. On the red line there is very little twist and you can see something like you describe in number 2
Screenshot_2023-04-05-21-51-44-02_40deb401b9ffe8e1df2f1cc5ba480b12

blastrock:
That's interesting. I'm not sure I get the graph though. Is the blue line the distance between the probe and the bed and the red line the distance between the nozzle and the bed?

Jayksopp:
No both lines are the difference between the distance from probe to bed and nozzle to bed as measured with a dial gauge. It is the z offset tlyou have to correct for vs the x axis position. The blue line is in a situation w severe x twist. The red line is after I fixed it

blastrock:
Oh that's nice! In the end, the only way I found to make the offset linear is to [move the probe](https://www.printables.com/model/423815-sv06-z-probe-mount-with-insertless-adxl-support), but it blocks the extruder screw, which is also a bit incovenient.

Anyway, supporting non-linear twists will at least help people with a stock SV06 who don't want to tweak it ^^

blastrock:
Fixed @murdock62 and @hiddentr0ll's bug mentioned here https://github.com/Klipper3d/klipper/pull/6048#issuecomment-1435689190

HelloThisIsFlo:
Hey 👋
Just catching up now, busy week, as always. 😅 Thanks for your answer, the conversation is very interesting.

If I’m being honest I didn’t fully understand the problem when I posted my last message. I didn’t even realize we were talking about x axis *twist*. I had *tilt* in mind and none of the calculations made sense. I couldn’t figure out why the bed mesh couldn’t compensate for a slight tilt. But I tried and it worked 🤷‍♂️.

Then I realized that we were talking about **twist** and it made a lot more sense.

But I still felt it should possibly follow a sort of theoretical curve. My intuition made me guess it would be either a straight line or a sine wave. I tried to model the path in fusion, but got stuck with my lack of skills and ended up with components hard-struck together that wouldn’t move the moment I introduce some mis-alignment of the rods 😅

Your explanation would indeed make me less hopeful that there is a theoretical optimal curve. You have a very good point with twist 2), I didn’t consider it at all and @Jayksopp ’s graph helped me understand it. I’m not sure why twist 3) would happen, but I can believe it might.

However, I believe the impact of twist 2) and 3) is somewhat negligible, I still think we may be able to fit a straight line to the points. The reason I believe this is because, between then and now, I did some experiments with linear regression and they’ve proven very successful for my particular setup 😃 (also SV06). Also it seems to confirm @Jayksopp 's own results.

The process

1. I measured 5 points with a feeler gauge (blue)
2. Then I measured 9 points with a feeler gauge (red)
3. Finally I re-measured 9 points with a sheet of paper this time (yellow)

All 3 sets of measurements fit almost exactly the same line, with an average error of `0.005` (vs the 9 points with feeler gauge).

4. I then generated 3 interpolated points to put in the config file

The result is, in my case at least, near perfect first layer on all 4 corners of the bed

Screenshot 2023-04-08 at 12 37 52

Out of curiosity I fitted a line to your results as well, they fit quite nicely too. With a R² of 96%

Screenshot 2023-04-08 at 12 37 58

I’ll be implementing linear regression in my own fork in a few days, for personal use, but I understand if you or the community prefers to go with point to point interpolation. The reason I prefer to use linear regression is because it allows me to not worry about having perfect measurements, as we can see the 9 points measured with paper were taking in a fraction of the time it took me to do the measurements with a feeler gauge and yet yield the same trendline.

 If you’re curious, the algorithm is quite simple (I stole it from the internet).
```python
# Sample data points
indexes = [0, 1, 2, 3, 4]
z_offsets = [0.109125, 0.03725, -0.0315, -0.059, -0.055875]

# Calculate the mean of x and y values
mean_indexes = sum(indexes) / len(indexes)
mean_z_offsets = sum(z_offsets) / len(z_offsets)

# Calculate the covariance and variance
covar = sum((indexes[i] - mean_indexes) * (z_offsets[i] - mean_z_offsets) for i in range(len(indexes)))
var = sum((indexes[i] - mean_indexes)**2 for i in range(len(indexes)))

# Compute the slope (m) and intercept (b) of the best-fit line
m = covar / var
b = mean_z_offsets - m * mean_indexes

# Print the equation of the best-fit line
print(f"The equation of the best-fit line is: y = {m} * x + {b}")
```

Thanks a lot for taking the time to go in depth with the explanations, it’s very much appreciated 🙏

On a sidenote, @Jayksopp , how did you connect the z-rods to the top bar? Do you mean the lead screw (because the rods are already attached to the top)?

Jayksopp:
> Hey 👋 Just catching up now, busy week, as always. 😅 Thanks for your answer, the conversation is very interesting.
> 
> If I’m being honest I didn’t fully understand the problem when I posted my last message. I didn’t even realize we were talking about x axis _twist_. I had _tilt_ in mind and none of the calculations made sense. I couldn’t figure out why the bed mesh couldn’t compensate for a slight tilt. But I tried and it worked 🤷‍♂️.
> 
> Then I realized that we were talking about **twist** and it made a lot more sense.
> 
> But I still felt it should possibly follow a sort of theoretical curve. My intuition made me guess it would be either a straight line or a sine wave. I tried to model the path in fusion, but got stuck with my lack of skills and ended up with components hard-struck together that wouldn’t move the moment I introduce some mis-alignment of the rods 😅
> 
> Your explanation would indeed make me less hopeful that there is a theoretical optimal curve. You have a very good point with twist 2), I didn’t consider it at all and @Jayksopp ’s graph helped me understand it. I’m not sure why twist 3) would happen, but I can believe it might.
> 
> However, I believe the impact of twist 2) and 3) is somewhat negligible, I still think we may be able to fit a straight line to the points. The reason I believe this is because, between then and now, I did some experiments with linear regression and they’ve proven very successful for my particular setup 😃 (also SV06). Also it seems to confirm @Jayksopp 's own results.
> 
> The process
> 
> 1. I measured 5 points with a feeler gauge (blue)
> 2. Then I measured 9 points with a feeler gauge (red)
> 3. Finally I re-measured 9 points with a sheet of paper this time (yellow)
> 
> All 3 sets of measurements fit almost exactly the same line, with an average error of `0.005` (vs the 9 points with feeler gauge).
> 
> 4. I then generated 3 interpolated points to put in the config file
> 
> The result is, in my case at least, near perfect first layer on all 4 corners of the bed
> 
> Screenshot 2023-04-08 at 12 37 52
> 
> Out of curiosity I fitted a line to your results as well, they fit quite nicely too. With a R² of 96%
> 
> Screenshot 2023-04-08 at 12 37 58
> 
> I’ll be implementing linear regression in my own fork in a few days, for personal use, but I understand if you or the community prefers to go with point to point interpolation. The reason I prefer to use linear regression is because it allows me to not worry about having perfect measurements, as we can see the 9 points measured with paper were taking in a fraction of the time it took me to do the measurements with a feeler gauge and yet yield the same trendline.
> 
> If you’re curious, the algorithm is quite simple (I stole it from the internet).
> 
> ```python
> # Sample data points
> indexes = [0, 1, 2, 3, 4]
> z_offsets = [0.109125, 0.03725, -0.0315, -0.059, -0.055875]
> 
> # Calculate the mean of x and y values
> mean_indexes = sum(indexes) / len(indexes)
> mean_z_offsets = sum(z_offsets) / len(z_offsets)
> 
> # Calculate the covariance and variance
> covar = sum((indexes[i] - mean_indexes) * (z_offsets[i] - mean_z_offsets) for i in range(len(indexes)))
> var = sum((indexes[i] - mean_indexes)**2 for i in range(len(indexes)))
> 
> # Compute the slope (m) and intercept (b) of the best-fit line
> m = covar / var
> b = mean_z_offsets - m * mean_indexes
> 
> # Print the equation of the best-fit line
> print(f"The equation of the best-fit line is: y = {m} * x + {b}")
> ```
> 
> Thanks a lot for taking the time to go in depth with the explanations, it’s very much appreciated 🙏
> 
> On a sidenote, @Jayksopp , how did you connect the z-rods to the top bar? Do you mean the lead screw (because the rods are already attached to the top)?

Nice work! 

part of the text is missing. It said "x twist disappears when z rods are not connected to the top bar" . For a while I printed without top z blocks :D then I upgraded the top bar to a piece of 2020 extrusion with custom top z blocks and this completely fixed the issue.

blastrock:
@HelloThisIsFlo Thanks for the detailed answer!

Your measurements look very linear indeed, and your point about doing a linear regression to cancel measurement errors makes sense.

You convinced me to make a 9 point calibration, for science! So this is a stock SV06 with [only one mod](https://www.printables.com/model/413616-sv06-top-z-axis-rod-fixing-blocks-parametric) at the moment, the top right rod block is 2mm shorter so that the print head has the same inclination on the leftmost and rightmost positions. The measurements are done with a paper test, using a gauge is bothersome as you said. Here's the result:

image

It looks similar to @Jayksopp's blue curve in that post above, but opposite direction. It is linear on the left, then mostly flat. The red line is done manually with just my eyes, I don't know how to do linear regressions in libreoffice ><

So, on some configurations, the correction is linear, and sometimes it is not. It might make sense to support both with a config option. Thanks for the code snippet, I'll try to integrate it, maybe tomorrow if I have some time.

As a side note, the current code gives me satisfying results, but in theory the correction shouldn't be multiple linear interpolations, it should be something smoother, like a cubic interpolation I guess. I am not very mathematically inclined so I'm not sure a cubic interpolation is what would make most sense here. Please tell me if you think of another way of smoothing the correction.

Jayksopp:
> @HelloThisIsFlo Thanks for the detailed answer!
> 
> Your measurements look very linear indeed, and your point about doing a linear regression to cancel measurement errors makes sense.
> 
> You convinced me to make a 9 point calibration, for science! So this is a stock SV06 with 
img_1_1681053412946
[only one mod](https://www.printables.com/model/413616-sv06-top-z-axis-rod-fixing-blocks-parametric) at the moment, the top right rod block is 2mm shorter so that the print head has the same inclination on the leftmost and rightmost positions. The measurements are done with a paper test, using a gauge is bothersome as you said. Here's the result:
> 
> image
> 
> It looks similar to @Jayksopp's blue curve in that post above, but opposite direction. It is linear on the left, then mostly flat. The red line is done manually with just my eyes, I don't know how to do linear regressions in libreoffice ><
> 
> So, on some configurations, the correction is linear, and sometimes it is not. It might make sense to support both with a config option. Thanks for the code snippet, I'll try to integrate it, maybe tomorrow if I have some time.
> 
> As a side note, the current code gives me satisfying results, but in theory the correction shouldn't be multiple linear interpolations, it should be something smoother, like a cubic interpolation I guess. I am not very mathematically inclined so I'm not sure a cubic interpolation is what would make most sense here. Please tell me if you think of another way of smoothing the correction.

Hi, maybe this can help shed some light on the non linear offsets that we are seeing. A while back I did an experiment where I attached one z rod to a non square tower: 
img_1_1681053412946

Both side twisted linear offset, one side twisted linear and then flat ?

LHARD:
Hi there, I remember I did some manual measurements on my setup (Geeetech A30 with twisted gantry/x-axis), and my setup was defintively not linear. 

I also checked the axis_twist_compensation code, and as far as I can tell, doing more/enough points will account for non-linear twist, as the linear interpolation happens between the closest two compensation points.

Having a non-linear twist-profile (for whatever reason) means doing linear regression or any other approximation introduces errors to the actual offset.

Instead I'd suppose to do it like the BedMesh does: do multiple measurements on the same point to improve accuracy/eliminate errors.
To improve it, the script could also calculate the variance/std deviation of each point, and recommend further measurements or a redo of certain points. Or: let the user repeat the measurement, until a certain std deviation is matched, along with elimating outliers to speed things up (could be done by the automatic bedmesh probing, too).


One other thing I saw in the code: the compensation values are still based on the mean z_offset (from the compensation measurements). I thought this was not recommended? I think, it makes sense to use a fixed-point z_offset as base for the compensation, like i.e. the z_offset at the middle of your bed. Because the offset might change i.e. with a nozzle-swap, but the compensation would still be valid. With the "virtual" offset, you have to redo the compensation to get the new offset.

blastrock:
Some updates:

- integrated linear regression from @HelloThisIsFlo's code
- removed recommended z offset
- simplified the code a lot
- updated documentation
- squashed commits

@HelloThisIsFlo I tested the linear regression code, it seems to work, but I can't confirm it since it doesn't produce good results on my printer, could you test it?

blastrock:
@LHARD 

> Instead I'd suppose to do it like the BedMesh does: do multiple measurements on the same point to improve accuracy/eliminate errors. To improve it, the script could also calculate the variance/std deviation of each point, and recommend further measurements or a redo of certain points. Or: let the user repeat the measurement, until a certain std deviation is matched, along with elimating outliers to speed things up (could be done by the automatic bedmesh probing, too).

The probe measurement is already done 3 times as per Klipper's default behavior. I don't think repeating the manual measurement would help. Once you find the right distance, that distance should never change. If you home your printer, it should be able to get back to that exact position with a very small error. In my experience, the measurements have been very repeatable.

> One other thing I saw in the code: the compensation values are still based on the mean z_offset (from the compensation measurements). I thought this was not recommended? I think, it makes sense to use a fixed-point z_offset as base for the compensation, like i.e. the z_offset at the middle of your bed. Because the offset might change i.e. with a nozzle-swap, but the compensation would still be valid. With the "virtual" offset, you have to redo the compensation to get the new offset.

The problem with taking the z_offset of the middle of the bed is that when doing the axis twist calibration, there may be no point in the middle of the bed, for example if you use an even number of points, or if your min/max X are not symmetric with the middle of the bed.

But it turns out taking the average of the offsets (or any value actually) is not a problem at all. The addition of the average to the Z corrections is only there as a normalization. The exact value is not important because they are only used to calibrate the bed mesh which is a relative mesh (and is probably normalized with the average too). The Z correction is also applied when doing probe calibration, but that's not a problem either, you just need to do it *after* the axis twist calibration. The probe calibration will always add a constant offset (from the twist compensation), and it does not have any impact. What's important is that the probe calibration and bed mesh are done *after* the twist calibration.

LHARD:
@blastrock 
> @HelloThisIsFlo I tested the linear regression code, it seems to work, but I can't confirm it since it doesn't produce good results on my printer, could you test it?

If the results are bad, it might be the case that your axis_twist just is **not** linear. in that case the linear regression can't be applied. Looking at my printer, it can't be linear, as the gantry-mounting (parts connecting x-axis on both ends to the z-axis) is _inside_ the print-volume. And while these parts could carry the twisting force (like from the gantry in my case), they will also stiffen the x_axis near the ends. That would explain why for some of us the twist is like zero on one or both ends.
This could also explain the findings of @Jayksopp. Loosening the towers could remove the twisting forces.
BTW: I wouldn't do a linear regression on that values either.

> The probe measurement is already done 3 times as per Klipper's default behavior. I don't think repeating the manual measurement would help. Once you find the right distance, that distance should never change. If you home your printer, it should be able to get back to that exact position with a very small error. In my experience, the measurements have been very repeatable.

I totally agree on the the PROBE measurements! What I meant was the _manual_ measurements (like paper-test) (sorry for not beeing precise enough)
If you repeat them multiple times, they should get more precise (with a little math).
A good way could be following: after you did the manual adjustment for every point once, the script could go back to the first one with i.e. 0.2mm additional offset, and ask the user to redo the adjustment. to make it even better, the repeating measurements could be done without showing the current value, as that might influence the adjustment.

> The problem with taking the z_offset of the middle of the bed is that when doing the axis twist calibration, there may be no point in the middle of the bed, for example if you use an even number of points, or if your min/max X are not symmetric with the middle of the bed.

> But it turns out taking the average of the offsets (or any value actually) is not a problem at all. The addition of the average to the Z corrections is only there as a normalization. The exact value is not important because they are only used to calibrate the bed mesh which is a relative mesh (and is probably normalized with the average too). The Z correction is also applied when doing probe calibration, but that's not a problem either, you just need to do it after the axis twist calibration. The probe calibration will always add a constant offset (from the twist compensation), and it does not have any impact. What's important is that the probe calibration and bed mesh are done after the twist calibration.

Again sorry for unprecise description:
Ususaly you (I?) use PROBE_CALIBRATE for z_offset adjustment, which you usually do always in the same spot (like the middle of your bed). After changing your nozzle, you propably will redo z_offest adjustment again on that spot, leaving your bed_mesh untouched.
axis_twist could do the calibration just referenced/normalized to that offset (which was done _before_ axis_twist_calibration), whether that exact point was measured by the script makes no difference. Now, the next time you i.e. change your nozzle, you just redo PROBE_CALIBRATE (which of course must be done _without_ axis_twist_compensation).

Doing everything _after_ axis_twist_calibration should work too, and I if got it right (in my head ;-) ) doing a PROBE_CALIBRATE _after_ axis_twist_compensation, should give you the same value regardless of the position, right? And that would not be the actual offset from probe to nozzle, but the new mean offset, right?
I'm just thinking and asking to understand what all the values will mean/actually describe.

HelloThisIsFlo:
Hey there,

Thanks for your detailed responses too. It’s very interesting topic. I’ll try to catchup on everything, let me know if I forgot something.

@blastrock I can see you already implemented linear regression, that was quick! I’d be happy to test it out of course. 

What do you think is the best process to test it? 

- Run the calibration on my machine, and print a first layer test?
- Or put debug statements in the code to make sure it generates the same values I got with the excel sheet (which are proven to work on my setup)?

On a side-note, in the meantime I also tried to clearly mis-align the gantry (by manually twisting one of the Z rods a few millimeters up) and ran a test print _without_ re-calibrating `axis_twist_compensation`. I was surprised to see that the same values for `axist_twist_compensation` were still compensating perfectly (although my bed was showing as seriously tilted, which is expected). That’s why I didn’t end up implementing linear regression in my own fork, since it’s a set-and-forget measurement I felt it wasn’t worth it. But still very happy to help with the testing of your implementation.

However, looking at other people’s answer, it seems that I’m more the exception than the rule 🤔 Thanks to your detailed measurements, and @LHARD  and @Jayksopp  comments, I’m starting to wonder if my machine being linear is not more of a coincidence than anything else. So … event though it was my suggestion, I’m starting to back down on this linear regression idea. Because, as @LHARD  mentioned, if the twist progression is mathematically *not* a linear line, then by using linear regression we may be _introducing_ error more than anything else 😕 If that’s the case, it may just be that linear regression is wrong even in my setup, but the error it introduces is less than the measurement error I had before linear regression, so it works out.

The previous paragraph is a long way of saying: Maybe @LHARd suggestion of doing multiple Z-offset measurements at *each* point and averaging the results would be an even better way of protecting against measurement error than my idea of linear regression.

I can see you @blastrock had repeatable measurements, but as you can see on the graph of my original post …



… there is a measurement variance, in my measurements at least. Yes using a feeler gauge does help, but it’s much faster to do 2 or 3 measurements with paper, than it is doing 1 single feeler gauge measurement. The measurement error does not come from the printer being at an incorrect location, but from the user (me) having difficulty to perfectly decide when it has the “exact same friction” as the point before. That being said … the average difference between the min and the max measurement at each point is only `0.023` … so maybe it’s not worth the trouble?

Still I think offering the option to repeat measurement and take the average would remove all this potential for user error. As suggested by @LHARD, if we could do them all at once, and then do another pass it may be even better (as the user gets used to what friction “feels like” after doing the first pass). And requiring a “minimum difference between the results” between confirming would be the ultimate safety guard.

Finally, one point I’d like to mention is that I greatly appreciate the time you take to incorporate our feedback. But maybe we could go with a “good enough” version for now (and the current version is pretty great considering it fixed my printer 😃), get it merged, and then discuss about potential improvements. What do you think? Do we know what is blocking this from being merged?

**TL;DR:**
- I'm backing out of my idea of using Linear Regression, I believe repeated and averaged measurements would be more reliable way to prevent against user measurement errors (@LHARD suggestion)
  - Still happy to help with the testing of your implementation of the feature (see above I have 2 questions regarding testing)
- The repeated & averaged measurements method would be even better with the other @LHARD suggestions
  1. Doing 1 full pass, then another full pass
  2. Keep doing passes until the desired precision is achieved on all points
- Finally, this PR is already incredibly useful in its current state, so maybe we could focus on getting it merged, and open another PR later for potential improvements like the ones suggested above?

Thanks again for all your hard work,

Flo

Jayksopp:
/

blastrock:
> What do you think is the best process to test it?

I didn't think of that. I guess you can add a couple prints for the m and b values and check that they are close to the ones you found.

>  Because, as @LHARD mentioned, if the twist progression is mathematically not a linear line, then by using linear regression we may be introducing error more than anything else

That's the problem, you'd have to know beforehand that your twist is linear. Well now that it's done, I guess it can stay. If the reviewer doesn't want it, I'll remove it.

> That being said … the average difference between the min and the max measurement at each point is only 0.023 … so maybe it’s not worth the trouble?

That's my reasoning. I do my paper test with a 0.025 step anyway, so this is an acceptable error for me ^^

> Finally, one point I’d like to mention is that I greatly appreciate the time you take to incorporate our feedback

Thanks!

> we could go with a “good enough” version for now

That's also what I'm thinking, we can still improve on this later. I won't have much time to work more on this in the near future. I'm not sure what's the procedure to get this merged. I read somewhere that reviewers pick their PR themselves, I'll check again the instructions and maybe ping one.

koonweee:
> Hello, is there anyway I can get some installation instructions on how to install this? I believe I modified all of my Klipper files to match the "files changed" here, but the "AXIS_TWIST_COMPENSATION_CALIBRATE" command is unknown. I also tried to retrofit Koonweee's original installation instructions, but I still can't get Klipper to recognize the command. What I tried was this, `cd klipper` `git remote add blastrock https://github.com/blastrock/klipper.git` `git fetch blastrock` `git checkout remotes/blastrock/jt/x-twist-compensation-simple klippy/extras/axis_twist_compensation.py klippy/extras/probe.py` `git checkout remotes/blastrock/twist klippy/extras/axis_twist_compensation.py klippy/extras/probe.py`
> 
> Thanks! p.s. I hope this gets merged into Klipper soon! It's so useful!!!

Did you restart the klipper service after checking out the files?

blastrock:
Also did you add the axis_twist_compensation section to your printer.cfg?

Jolliness0995:
> Did you restart the klipper service after checking out the files?
Yes, I did.
> Also did you add the axis_twist_compensation section to your printer.cfg?
This was what I was missing! 

Thanks to both of you for your help!

Additionally, if anyone ends up in a similar situation to me and reads this, don't run this command `git checkout remotes/blastrock/twist klippy/extras/axis_twist_compensation.py klippy/extras/probe.py` it uses remotes/blastrock/twist when you need to be using remotes/blastrock/jt/x-twist-compensation-simple.

HelloThisIsFlo:
Hey @blastrock 

I just completed some tests and I can confirm that the implementation of linear regression is working 🎉

I plugged in the values I got from my original measurements. I logged the `m` and `b` values, and I get the exact same straight line and interpolated values.

Screenshot 2023-04-27 at 11 39 12
_(the blue dots don't show on the graph because they perfectly line up with the yellow ones)_

Just to make sure I then made a test print and it looked great!

Regarding the idea of doing multiple measurements during initial calibration, I still think it would be a nice improvement. The worst case error in my measurement was 0.03 and I usually do `PROBE_CALIBRATE` down to `0.01`. But again, it is absolutely not a blocker.

You did a great job 🎉, and I think if we can get this PR merged it will help so many people in the community. After it's merged, _maybe_ I could open a PR to add that multiple measurement feature at some point, but maybe I'll also decide it's not worth the hassle 😅. For now, LGTM 😃

Thanks again for your work and you too @koonweee . You made my printer super reliable 😃

Gitii:
Hi @blastrock,

I tested your changes, and I am very happy with the result (SV06). Big thanks for implementing!
The guide in the documentation had been great and it worked quite well!
The procedure had been easy to do, too 🙂
As a 3d printing novice, here is some feedback:
* It hasn't been obvious that it's possible to switch between 'linear' and 'multilinear' after calibration (and if it's not, I am still very happy with the result :) ). That means if the result doesn't change, just try the other one and see if that 'fixes' it. As I'd already done a 9 points run, I haven't been inclined to do everything again... I would mention that in the overview section as tip. It's described later but IMO it's very useful to mention sooner than later.
* The guide in the documentation doesn't mention how to change the number of points. The gcode spec does, but one might think oversee that.
* I think (and might be wrong here) that 'multilinear' works better with more points, doesn't it? If yes, than I would increase the default number of points to 9 and try both 'linear' and 'multilinear' and check which one works "better". Saves time & effort. But that's just a personal preference because I am lazy and z probing is annoying.

tomallo99:
Hello, I tried to apply this fix to my SV06, but I'm confused. Some of the [documentation](https://github.com/koonweee/klipper/blob/jt/x-twist-compensation-simple/docs/G-Codes.md#x_twist_compensate_status) references commands like X_TWIST_COMPENSATE_STATUS, or X_TWIST_PROFILE_LOAD NAME=<profile_name>.
Seems like now there's AXIS instead of X in the names, but I can only do "CALIBRATE". How do I make sure that the compensation is active? The "STATUS" command doesn't seem to be there, neither does the rest.
Apologies for my confusion, just installed Klipper for the first time today and I'm very new.

blastrock:
That's the old documentation, you should look at this pull request for the latest doc.

Once you have done the calibration and save_config, the values are saved to your printer.cfg. The compensation is always active unless your clear it or remove the line from printer.cfg.

blastrock:
@Gitii thanks for testing and for the feeedback.

I patched the documentation, it should help with the first two points you mentioned.

About defaulting to 9 points, I'm not sure, since it is annoying to do the calibration, and 3 points already gives very good results. Also it's the default value for Marlin too.

ohrn:
Hi everyone,

I tried using this with a Klicky probe on a Voron Trident but I got stuck trying to figure out how to get it to attach and dock the probe at the right time during calibration.

If one does a plain AXIS_TWIST_COMPENSATION_CALIBRATE command it immediately moves to the first calibration point without picking up the Klicky and fails with "Probe triggered prior to movement" which I guess is good (or it would crash into the plate).

Now if I wrap this in a naive macro that does ATTACH_PROBE first, then it picks up the Klicky, automatically probes the first point then goes into manual nozzle offset calibration. Here the Klicky needs to be detached somehow. If I manually remove it, then do TESTZ and ACCEPT it immediately moves on to the next calibration point without giving me a chance to reattach the Klicky so again it fails with "Probe triggered prior to movement".

Is there a way to write a macro that can do ATTACH_PROBE and DOCK_PROBE at the right time between each calibration point? Is it even desirable to do automatically or will tool head X-Y movement between the automatic and manual probing steps deteriorate the calibration accuracy?

If not automatic, can the AXIS_TWIST_COMPENSATION_CALIBRATE command be extended to make a pause to allow manual attach/detach of the Klicky between each point?

Finally, what is the correct way to write my own calibration values in the config if I want to play around with it? The docs mention z_compensation but doesn't describe the layout in detail. Is it as simple as a comma separated list of z values like so:

z_compensation: 0.05, 0.07, 0.08

Or does it also contain x and y coordinates?

docs/Axis_Twist_Compensation.md
```diff
@@ -0,0 +1,93 @@
...
+
+If required, the user may fine tune the `z_compensations` values.
+
+For example, the value of `0.116667` represents the `z_compensation` applied
+when the printer is probing at `start_x`. This means that if the non-compensated
+z-height measured at `start_x` was `2.0`, the compensated z-height is now `2.0 +
+0.116667 = 2.11667`, resulting in a higher point on the recorded mesh.
+
+Should the squish on the side of the bed represented by `start_x` be
+insufficient, the user may specify a lower `z_compensation` value in place of
```
ohrn, line 83: It says "a lower z_compensation value" but 0.016667 is larger than 0.116667! :dizzy:

Does it refer to lower as in closer to the bed or lower as in a numerically smaller value? 

Needs clarification imho.

blastrock:
I must admit I never thought about klicky probes. I am also very surprised that you would need a twist compensation on a Voron-class printer :o 

I don't own a Voron but I'm wondering whether the right course of action is to fix the printer hardware instead, since in your case it should be fixable. We use this feature on our printers because they are cheap and the twist is very hard to fix.

I guess you could enter the values manually as you suggest. It's just comma separated values, the X values are deduced from start_x and end_x from the config and the Y value is fixed, in the config too. Here are my settings:

    z_compensations = 0.083333, 0.054583, 0.017083, 0.002083, -0.039167, -0.037917, -0.026667, -0.031667, -0.021667

ohrn:
Hi @blastrock 

> I must admit I never thought about klicky probes. I am also very surprised that you would need a twist compensation on a Voron-class printer :o
> 

Well, fancy printer or not, I too have problems. :thinking: 

Now I don't expect this PR to be the right place for tech support but I'll give you a bit of background as to why I'm trying out your change.

I have a weird slant that causes increased squish from left to right on the plate along X that is not correctly compensated for by automatic bed mesh. There is no change in squish from front to back of the plate. The measured bed mesh shows no obvious leaning in any direction, but if I manually adjust it to make it increasingly higher on the right side I can make the squish problem go away.

(It's not the classic Trident problem of uneven left and right Y linear rails, that was visible at the back of the measured mesh on my machine and I've adjusted the rails.)

Another curiosity is that if I move the Z homing point (I'm using the Klicky for Z homing too) to the left I get a perfect first layer on the left side (no bed mesh). If I move the Z home point to the right I get a perfect first layer on the right, _without_ changing the probe Z offset. This makes me think I don't really have a twisted X axis as there does not seem to be a location bias in the Z offset, instead there is some other funny business going on.

But, I found this patch and still thought why not give it a try. Since it makes a correction along the X axis and my problem manifests along X too.

> I don't own a Voron but I'm wondering whether the right course of action is to fix the printer hardware instead, since in your case it should be fixable. We use this feature on our printers because they are cheap and the twist is very hard to fix.
> 

I've carefully adjusted everything I can think of on the hardware side and I've run out of ideas. If I probe a mesh on a flat piece of window glass I now have a variability of less then 0.04 mm (after Z tilt adjust), there are a few bumps but the pattern does not match any of the usual problems that show up in trouble shooting guides. (The real bed has worse variability of course with the magnet sheet and steel spring plate being far from perfect.)

The squish difference I get is more than 0.04 mm, I need to add more like 0.12 mm to the right edge when fixing up a bed mesh manually.

> I guess you could enter the values manually as you suggest. It's just comma separated values, the X values are deduced from start_x and end_x from the config and the Y value is fixed, in the config too. Here are my settings:
> 
> ```
> z_compensations = 0.083333, 0.054583, 0.017083, 0.002083, -0.039167, -0.037917, -0.026667, -0.031667, -0.021667
> ```

Thanks, I'll fill in some values and see if I can make my problem go away. It would be so much more convenient than having to tweak an entire bed mesh whenever I've messed with the steel spring plate and need to measure it again.

blastrock:
Thanks for the details, that's interesting. I am a bit curious about this issue...

> Another curiosity is that if I move the Z homing point (I'm using the Klicky for Z homing too) to the left I get a perfect first layer on the left side (no bed mesh). If I move the Z home point to the right I get a perfect first layer on the right, without changing the probe Z offset. This makes me think I don't really have a twisted X axis as there does not seem to be a location bias in the Z offset, instead there is some other funny business going on.

Indeed that's curious. Another thing that I thought about is that if I understand correctly how the klicky probe works, it probes right under the nozzle. The twist we have on our printers introduce errors in the measurements because the (induction or bltouch) probe is offset from the nozzle. So it looks like you're having another issue, but this patch may be able to compensate for it too.

> Thanks, I'll fill in some values and see if I can make my problem go away

Btw, you can put just two values in there to have a linear interpolation.

In case it helps, I've pushed a branch: https://github.com/blastrock/klipper/tree/twist-compensation-pause . This branch will wait 10s after moving the toolhead, just before each automatic probing. This can give you the time to insert the klicky probe manually. I didn't test it though ^^

I hope you figure out your offset issue. I'm very curious about its cause. When you figure it out, I'd appreciate you drop me an email (address is on my profile) :)

docs/Axis_Twist_Compensation.md
blastrock: >  0.016667 is larger than 0.116667

Hmm... 0.016667 *is* smaller than 0.116667 

I did the math again in my head, the formula for each bed mesh point is like `bed_mesh_z[x, y] = probed_z[x, y] + z_compensation[x]`. So if z_compensation is lower, the bed mesh point will be lower and klipper will compensate for that by the toolhead lower when it goes over that point, resulting in more squish. This paragraph looks correct to me.

ohrn:
> Indeed that's curious. Another thing that I thought about is that if I understand correctly how the klicky probe works, it probes right under the nozzle. The twist we have on our printers introduce errors in the measurements because the (induction or bltouch) probe is offset from the nozzle. So it looks like you're having another issue, but this patch may be able to compensate for it too.

No, the Klicky (as mounted on a Stealhtburner print head) sits approximately 20 mm behind the nozzle and 5 mm below. So there is opportunity for a lever effect if the head twists.

Even if a "Klicky style" solution was designed such that the mount allows the micro switch to end up right below the nozzle center the necessary Z offset would still give raise to a lever effect (though I guess a much smaller one).

> > Thanks, I'll fill in some values and see if I can make my problem go away
> 
> Btw, you can put just two values in there to have a linear interpolation.
> 

Yea, that I figured out. After a bit of trial and error I ended up with

start_x: 0
end_x: 300
type: linear
z_compensations: -0.05, 0.05

Together with a very small adjustment of the probe z offset and using automatic bed mesh I'm now getting an amazingly perfect first layer over the entire build plate. Success!

> In case it helps, I've pushed a branch: https://github.com/blastrock/klipper/tree/twist-compensation-pause . This branch will wait 10s after moving the toolhead, just before each automatic probing. This can give you the time to insert the klicky probe manually. I didn't test it though ^^
> 

I just gave it a try but the pause happens in the wrong place. After entering ACCEPT it sleeps 10 seconds _before_ raising the nozzle from the bed, it's not possible to attach the switch at this point. Then it rises up, moves to the next point and fails.

It needs to pause after reaching horizontal_move_z height. Now looking at your change it seems the code tries to do just that. It's not working though, maybe there is a flush command needed before the sleep to actually make the head move? I don't know enough about Klipper internals to say.

Rather than sleeping a fixed time it would be even better if raises the head and stops indefinitely. Then you attach the probe manually and enter some new "CONTINUE" g-code (in the same manner it waits for ACCEPT) to make it move on to the next calibration point.

> I hope you figure out your offset issue. I'm very curious about its cause. When you figure it out, I'd appreciate you drop me an email (address is on my profile) :)

I'm really really really happy that the x axis compensation worked out so well. Many thanks to you and everyone else involved in this patch. Of course I still want to figure out the root cause of that 0.1 mm slant, I'll be sure to send you a mail if I ever find it.

I think I'll revisit the z-offset once again by tedious manual measurements. To see if I can find a 0.1 bias across X or not.

docs/Axis_Twist_Compensation.md
blastrock: Indeed the math was wrong, and the numbers were hard to follow, fixed it.

blastrock:
> No, the Klicky (as mounted on a Stealhtburner print head) sits approximately 20 mm behind the nozzle and 5 mm below. 

Oh I didn't know that, it makes sense then.

> the necessary Z offset would still give raise to a lever effect (though I guess a much smaller one).

Agreed, I can't measure it, but my guess is that this one would be negligible.

> I'm now getting an amazingly perfect first layer over the entire build plate. Success!

Great! Glad this patch helped :)

> I don't know enough about Klipper internals to say.

Same here, this is my first contribution ^^

A proper solution would be what you suggested with the additional commands, or adding additional parameters like `BEFORE_PROBING=ATTACH_PROBE` to call your custom macro during the calibration. I don't have much time to work on this at the moment however, so I'd like to get this merged and improve on it later :)

Happy first-layering :D

dewi-ny-je:
What's blocking this from merging?
Is there anything we, non-coders, can do to speed up merge? I'd like to use it in mainline without having to switch to a different branch, and I'm sure many users (more than 100) will be interested in it as well.

KevinOConnor:
> What's blocking this from merging?

There's been a lot of traffic on this PR and I wasn't sure if it was fully ready for review.  If all is ready then I'll look to review it next week.  As high-level feedback I think the feature makes sense and would be useful to add to Klipper.

Cheers,
-Kevin

Zeanon:
I have used your module for about a week now and I have to say it's working great.
But I have encountered one Problem:
When you are using a dockable probe like a Euclid, Klicky or sideswipe it's not really usable since you have to manually remove the probe after it probed to adjust the offset and the put it back for the next probing point which is basically impossible since the next probing sequence is started as soon as you accept the values.

I have already written a fix for it which just adds a continue command between each probing sequence and also moves the toolhead to the horizontal_move_z height after the manual probe so you have time and space to add the probe back to the toolhead between each probing sequence.

I would do a PR for it if you are fine with that.

blastrock:
Thanks for testing!

This problem was indeed reported earlier in this thread, but I didn't want to add more features before this gets merged. However, that's definitely a good feature to have and it would be awesome if you opened a PR!

Zeanon:
Another thing I noticed: imo it would make sense to either remove the axis_twist_compensation_clear command or add a way to load the compensation back, otherwise you could clear it without having a way to enable it again without restarting.

Dustmuffins:
After a lot of struggling to get my bed level, including replacing my z guide rods, and all hardware, including X blocks and z motor mounts, this was the only thing that allowed me to use my full bed. Excellent work!

Sovol SV06 with CR touch probe.

klippy/extras/axis_twist_compensation.py
```diff
@@ -0,0 +1,347 @@
...
+            elif config_key == 'z_compensations':
+                value = Helpers.parse_comma_separated_floats(
+                    config.get('z_compensations', default=""))
+            else:
+                value = config.get(config_key, default)
+            if required and value is None:
+                raise config.error(
+                    "Missing required config option for section [{}]: {}"
+                    .format(config.get_name(), config_key))
+            setattr(self, config_key, value)
```
KevinOConnor, line 68: I find this code to be very confusing.  I think it would be much more readable if `config_getXXX()` was directly called the 7 times.

Please use `config.getlist()` (or `config.getlists()`) instead of implementing a custom `parse_comma_separated_floats()` function.

Don't keep a reference to the `config` object in the local class (doing so leads to confusion as all options must be read during the initialization phase).  If you need the `config.get_name()` value then read it during init and store it in a `self.name` member variable.

For what it is worth, I think the static `Config` and `Helpers` classes make the code harder to read.

klippy/extras/axis_twist_compensation.py
```diff
@@ -0,0 +1,347 @@
...
+        # gets probe settings when they are available
+        def callback():
+            self.probe = self.printer.lookup_object('probe', None)
+            if (self.probe is None):
+                raise self.config.error(
+                    "AXIS_TWIST_COMPENSATION requires [probe] to be defined")
+            self.lift_speed = self.probe.get_lift_speed()
+            self.probe_x_offset, self.probe_y_offset, _ = \
+                self.probe.get_offsets()
+        return callback
```
KevinOConnor, line 185: It is confusing for `_handle_connect()` to be a factory function - just pass `self._handle_connect` to `self.printer.register_event_handler()` and directly perform the initialization code.  (As is common in all the other Klippy modules.)

docs/G-Codes.md
```diff
@@ -1313,6 +1313,20 @@ print.
...
 `SDCARD_RESET_FILE`: Unload file and clear SD state.
 
+### [axis_twist_compensation]
+
+The following commands are available when the
+[axis_twist_compensation config
+section](Config_Reference.md#axis_twist_compensation) is enabled.
+
+#### AXIS_TWIST_COMPENSATION_CALIBRATE
+`AXIS_TWIST_COMPENSATION_CALIBRATE [N_POINTS=<value>]`: Initiates the X twist calibration
```
KevinOConnor, line 1323: I'd prefer to avoid abbreviations in the parameter names, so instead of `N_POINTS` maybe something like `POINTS` or `SAMPLE_COUNT` or `SAMPLES` or `PROBE_COUNT` or similar.

klippy/extras/probe.py
```diff
@@ -127,6 +127,13 @@ def _probe(self, speed):
             if "Timeout during endstop homing" in reason:
                 reason += HINT_TIMEOUT
             raise self.printer.command_error(reason)
+        # get z compensation from axis_twist_compensation
+        axis_twist_compensation = self.printer.lookup_object(
+            'axis_twist_compensation', None)
+        z_compensation = 0 if not axis_twist_compensation \
+            else axis_twist_compensation.get_z_compensation_value(pos[0])
```
KevinOConnor, line 134: In keeping with the style of the existing probe.py code, please break this out into a regular `if` statement - for example:
```
z_comp = 0.
if axis_twist_compensation is not None:
    z_comp = axis_twist_compensation.get_z_compensation_value(pos)
```

Please check for `None` instead of using a python truth test.  Please pass the full position (not just the x coordinate) to the axis_twist_compensation module.

KevinOConnor (commented):
Thanks.  Sorry for the delay in reviewing.  In general, it looks fine to me.

I have a few minor comments - see below.

-Kevin

docs/G-Codes.md
qwewer0: I would prefer `SAMPLE_COUNT`, but the others are fine too.

docs/Axis_Twist_Compensation.md
```diff
@@ -0,0 +1,93 @@
...
+Note that if your axis is significantly twisted it is strongly recommended to
+first use mechanical means to fix it prior to applying software corrections.
+
+## Overview of compensation usage
+
+> **Tip:** Make sure the [probe X and Y offsets](Config_Reference.md#probe) are
+> correctly set as they greatly influence calibration.
+
+> **Tip:** A feeler gauge is recommended in place of paper for more accurate
+> measurements
```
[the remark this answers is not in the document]
HelloThisIsFlo: I can chime in to confirm that I found paper gave much quicker results with very similar accuracy, if not better.

It's easier to gauge friction with paper than with a feeler gauge. And, considering consistency is key for this measurement, not the absolute height, as you mentioned, I would even go as far as _recommending_ to use paper for the measurement, over a feeler gauge

docs/Axis_Twist_Compensation.md
```diff
@@ -0,0 +1,93 @@
...
+`0.12`, such as `0.02`. This would result in the mesh being lowered by
+`0.12 - 0.02 = 0.1mm` at points along `start_x`.
+
+## Reference
+
+The calculations for this module are largely based off those found in
+[Marlin PR #23238](https://github.com/MarlinFirmware/Marlin/pull/23238).
+
+A visualization of the issue can be found in
+[this discussion thread](https://github.com/MarlinFirmware/Marlin/issues/22791).
```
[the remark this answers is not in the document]
HelloThisIsFlo: Again, chiming in (not PR owner). I personally found it useful to have the resources when I was trying to understand how it worked. But on the other hand, if it ... just works, like all other features of Klipper, maybe there is no need to understand where it came from. 

Is there a place where we collect the research made into the feature that ended up in Klipper? I just think it'd be a shame to lose this valuable context (but not the end of the world either 😌)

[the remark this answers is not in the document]
Sineos: Just use the GH search function and, e.g., search for `AXIS_TWIST_COMPENSATION_CALIBRATE`, `twist` etc
This way you will get this and potentially all future PRs related to this

docs/Axis_Twist_Compensation.md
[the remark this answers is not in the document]
blastrock: Paper worked very well for me too, I thought other's experiences differed. I'll just remove this tip :+1:

docs/Axis_Twist_Compensation.md
```diff
@@ -0,0 +1,93 @@
...
+
+For example, the value of `0.12` represents the `z_compensation` applied
+when the printer is probing at `start_x`. This means that if the non-compensated
+z-height measured at `start_x` was `2.0`, the compensated z-height is now `2.0 +
+0.12 = 2.12`, resulting in a higher point on the recorded mesh.
+
+Should the squish on the side of the bed represented by `start_x` be
+insufficient, the user may specify a lower `z_compensation` value in place of
+`0.12`, such as `0.02`. This would result in the mesh being lowered by
+`0.12 - 0.02 = 0.1mm` at points along `start_x`.
```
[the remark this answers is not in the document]
blastrock: I didn't need it for my printer, not sure it helped other people. I can remove it.

docs/Axis_Twist_Compensation.md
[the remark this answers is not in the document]
blastrock: I also think this information is not lost, just following the commits, the links in the PRs etc, you'll be able to reach it.

> Is there a place where we collect the research made into the feature that ended up in Klipper?

I wasn't here since the beginning, on my part all the research was Marlin's code and the previous work from @koonweee, nothing very hard to find

docs/G-Codes.md
```diff
@@ -1313,6 +1313,20 @@ print.
...
+wizard. `N_POINTS` specifies the number of points along the X axis to calibrate
+at and defaults to 3.
+
+#### AXIS_TWIST_COMPENSATION_CLEAR
+`AXIS_TWIST_COMPENSATION_CLEAR`: Clears the current compensation values.
```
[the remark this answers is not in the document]
blastrock: This is the equivalent of BED_MESH_CLEAR, except there are no profiles on this feature.

> how do they get the compensation back afterwards?

Restarting klipper would reload the compensation from the config (unless you used SAVE_CONFIG).

I can remove this command and let the user delete the appropriate section themselves from the klipper config if they want to clear the compensation. Your call :)

blastrock:
Thanks a lot for the review! I answered to some comments, I will be fixing the rest very soon.

docs/G-Codes.md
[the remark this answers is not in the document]
jamesarthur: Correct me if I’m wrong, but would you not want to clear the data before using a paper test to adjust bed screws (not using the probe)? Assuming that the probe is more offset than the nozzle, that is.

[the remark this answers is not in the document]
blastrock: I don't have bed screws, so I can't run the BED_SCREWS_CALIBRATION myself. But from what I understand, this calibration will only move the nozzle to the 4 corners at a height of 0, and let you adjust the screws to that the bed plane is parallel to the xy plane of the extruder. Does this calibration use the probe at all? If it does not, it won't be affected by this PR.

klippy/extras/axis_twist_compensation.py
blastrock: All fixed, but I have one question left. I now use config.getlist instead of parsing the line, but I couldn't find an equivalent to save the list. In bed_mesh.py the formatting is done by hand too: https://github.com/Klipper3d/klipper/blob/a96608add40e316f25f15d9c9d1c1fbd86dbbebe/klippy/extras/bed_mesh.py#L1239-L1245 

Is there a method to do that or should I leave the manual formatting code here?

docs/G-Codes.md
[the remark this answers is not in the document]
jamesarthur: There is BED_SCREWS_ADJUST which is as you say, and also SCREWS_TILT_CALCULATE which does indeed use the probe. In either case, wouldn’t adjusting the screws immediately invalidate the x-twist calibration?

My uninformed instinct is that for printers with severe x twist like mine, you should clear the x-twist calibration, then level the bed using BED_SCREWS_ADJUST, then do the x-twist calibration, in that order. And if you do have a lot of twist, and your probe is strongly offset, don’t use SCREWS_TILT_CALCULATE.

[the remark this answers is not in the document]
blastrock: Oh that's a nice feature!

Disclaimer: all this is just my understanding of things since I can't test it.

It shouldn't matter if you have twist compensation enabled when doing BED_SCREWS_ADJUST since it will not use the probe. As for SCREWS_TILT_CALCULATE, it's a feature very similar to the bed mesh, but instead of a height map, you get a number of turns to adjust your screws. As such, this SCREWS_TILT_CALCULATE *must* be done after applying the twist compensation, otherwise the result will be tilted, same as the bed mesh.

> In either case, wouldn’t adjusting the screws immediately invalidate the x-twist calibration?

It shouldn't, the twist calibration measures the difference between the height measured from the probe and the real height of the nozzle, same as probe_calibrate, but all along the X axis. Unless your bed was strongly tilted when doing the twist calibration, the introduced error should be negligible.

So maybe this AXIS_TWIST_COMPENSATION_CLEAR is useless after all...

blastrock:
@KevinOConnor I addressed most of your comments. There are still a couple up there where I need your feedback. I pushed my current work, but I haven't tested it yet, so you might want to wait before reading the new commits.

I have a few more questions too.

As you saw, this implementation supports two types of interpolation for the compensation, multilinear which is linear for each pair of points and linear which does a linear regression on all the points. @HelloThisIsFlo asked for this linear regression, but then changed his mind (https://github.com/Klipper3d/klipper/pull/6149#issuecomment-1512922147). I implemented it and left it here, but I'm starting to change my mind too.

Since we are trying to simplify the code, maybe I should just remove this mode. There might not be that much demand for it, and if there is, it's not too hard to add back. It looks like people (and me) are having pretty good results with just the multilinear approach, on both this PR and on Marlin. What do you think?

I have made the fixes for your comments in multiple commits, what's your policy on this? Should I squash them? Or will we merge them as they are now?

Also, this PR is becoming a bit old, should I rebase on master, or should I merge master into this branch? Or just do nothing?

Thanks again for your time on this, this is much appreciated!

klippy/extras/axis_twist_compensation.py
```diff
@@ -0,0 +1,306 @@
...
+
+"""
+[axis_twist_compensation]
+horizontal_move_z: 5
+speed: 50
+calibrate_start_x: 0 ; nozzle's x coordinate at the start of the calibration ! required
+calibrate_end_x: 200 ; nozzle's x coordinate at the end of the calibration ! required
+calibrate_y: 100 ; nozzle's y coordinate during the calibration ! required
+compensation_type: multilinear
+"""
```
blastrock, line 15: I can't find such a section in other modules, should I remove it from here?

HelloThisIsFlo:
> As you saw, this implementation supports two types of interpolation for the compensation, multilinear which is linear for each pair of points and linear which does a linear regression on all the points. @HelloThisIsFlo asked for this linear regression, but then changed his mind ([#6149 (comment)](https://github.com/Klipper3d/klipper/pull/6149#issuecomment-1512922147)). I implemented it and left it here, but I'm starting to change my mind too.
> 
> Since we are trying to simplify the code, maybe I should just remove this mode. There might not be that much demand for it, and if there is, it's not too hard to add back. It looks like people (and me) are having pretty good results with just the multilinear approach, on both this PR and on Marlin. What do you think?

As the person who came up with the original idea... I would agree that we can remove it (if it's not too much work). It spawned a very interesting discussion, and I loved collaborating with you, but it turns out my suggestion was based on a wrong assumption. I'm even using multilinear myself now 😅

Amazing job for making this PR mergeable @blastrock. I struggle to maintain 2 small open-source projects, and sometimes it takes me weeks to even just _investigate_ an issue. In contrast, I'm very impressed by your response time! Your work, as well as @koonweee 's, is absolutely essential in getting perfect first layers on the SV06. 

Thank youuuuu 🙏🙏🙏🙏

Zeanon:
I have made the PR for the optional CONTINUE command, since in the current state the module is unfortunately not usable with a dockable probe
Should I redo my changes on the improved version of the code?

klippy/extras/axis_twist_compensation.py
KevinOConnor: To save an updated list, just format it manually (eg, `",".join(["%.3f" % (v,) for v in my_values])`).

-Kevin

klippy/extras/axis_twist_compensation.py
KevinOConnor: Up to you.  I understood this section to be code comments.  If you feel it isn't helpful it's fine to remove it.

Cheers,
-Kevin

KevinOConnor:
> It looks like people (and me) are having pretty good results with just the multilinear approach, on both this PR and on Marlin. What do you think?

In general, I prefer to present less options in the config file.  I regularly get feedback that users are overwhelmed by the large number of options in Klipper.  So, any time we add a config option I think we (as developers) should have a firm feeling on what the option does, who is using it, and what tasks they need to set it for.  It sounds like we don't have a good understanding of how this option would be used (or why), so I'd suggest not including it - at least until we do have that understanding.

Cheers,
-Kevin

KevinOConnor:
> Also, this PR is becoming a bit old, should I rebase on master, or should I merge master into this branch? Or just do nothing?

As long as github allows me to "squash and merge" then everything is fine.  (And Github still offers that option.)  So no need to rebase at this time.

-Kevin

blastrock:
All comments should now be addressed! I rebased on master to test. I calibrated the compensation and did a bed mesh, it looks fine, I don't have the time to do a print though, maybe tomorrow. Or if anyone wants to test the branch again, feel free to do so :)

Anyway, this code is ready for another review.

@Zeanon Sorry, I forgot to answer to your question. The branch should be stable now, you can rebase your changes on it and open a pull request as soon as this gets merged :)

Zeanon:
Why not implement the continue feature before it gets merged as an the module is not usable if you have a dockable probe?

blastrock:
Well I have nothing against it, I just want to make sure this gets merged since the original PR dates back to February, so I'd rather avoid adding new features.

I suggest that you rebase your work over mine and open a new PR. Then if @KevinOConnor is ok, he can review it and merge both PRs at once.

Zeanon:
I get where you are coming from, but I think unfinished code has a lesser chance of getting merged, since RN this feature is not usable for a lot of users(since dockable probes are quite popular)
And also it is not really an added feature and more of a fix imo
Ultimately it's up to you and Kevin

KevinOConnor:
I'd prefer to review this PR as it is and handle dockable probes separately.

Cheers,
-Kevin

Zeanon:
Fair enough, but in that case I would add a warning to the docs that this module is not compatible with dockable probes yet just so nobody crashes their nozzle(which should not happen when the probe is nc but in case it is no you will have a problem)

blastrock:
Good point, I added a note in the doc.

kyleobrien91:
Has anyone else received an error `Internal error during connect: ':' expected after dictionary key (axis_twist_compensation.py, line 81)`?

I've had a look at the code at that location and it looks fine to me. 


I followed @Jolliness0995 's steps [here](https://github.com/Klipper3d/klipper/pull/6149#issuecomment-1520815293)

CleanShot 2023-07-11 at 08 27 02@2x

bassamanator:
Originally, I had completely solved my X-twist issue (before I knew what x-twist was) by painstakingly editing the mesh manually.

I pulled this code last night after someone on my repo told me about it, and it's a game changer. It does not compensate for x-twist completely, as is expected, however, I am no longer concerned about first layers. No longer is the nozzle digging into the bed on the left side, and laying down loose layers on the right side. First layers are more than acceptable over the entire build plate. I think with a bit of physical adjustment, which I was afraid to do before because I would have to manually adjust the mesh again, and this x-twist compensation feature, Andrew Ellis himself will give my first layer squish two thumbs up!

I checked my Klipper print history and I have over hundred of these prints (I'm not exaggerating):



No more!

@blastrock @koonweee Thanks!

KevinOConnor:
Thanks.

-Kevin

Flol0:
Finally! Thank you very much guys!

koonweee:
Apologies for abandoning the original PR, I was in the middle of moving across the world and dropped off the 3D printing world for a bit. Still, glad to see this merged, thanks @blastrock and @KevinOConnor!

HelloThisIsFlo:
@koonweee No worries, you started it all, and they finished it. Thanks to all of you ♥️

PR #6153, opened 2023-04-05
Config: SKR3, update for second chip choice

JamesH1978:
As found on the discord by user Qtin, this board now ships with one of STM32H743 or STM32H723 Example config updated to reflect this. This has been tested on the users board. To note, the H743 bin will flash to the H723 chip, but will not work.

Thanks
James

signed-off-by: James Hartley <james@hartleyns.com>


KevinOConnor:
Thanks.

-Kevin

PR #6155, opened 2023-04-07
Function for retraction and dwell after homing

fredriknk:
Adding parameters you can use in printer.cfg to retract toolhead by a specific distance after homing, this is quite importent in sensorless corexy devices because they will experience lockup if you home the y axis with the x axis in connection with the frame (or the other way around), It allso includes a dwell time you can set to prevent retriggering of the other axis which can happen if the next axis starts homing just as the first axis retract is settling. This gives you an easy way to retract the axis before it homes the next one without making a custom homing script.

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1219,7 +1219,7 @@ The following commands are available when any of the
 are enabled.
 
 #### DUMP_TMC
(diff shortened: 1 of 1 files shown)


JamesH1978:

Thanks
James

PR #4983, opened 2021-12-01
add  the reverse mode of temperature_fan

CF3B5:
add the reverse working mode of temperature_fan to some internal circulation heating fans.

Signed-off-by: Locke chen <cf3b5bb@gmail.com>

PR #6156, opened 2023-04-07
Reverse temperature fan

dingyifei:
Replace #4983 with PID reverse temperature fan implemented as well

Waiting for testers

High-level description:

It flips the sign of temperature difference to reverse the fan on/off

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2708,6 +2708,11 @@ information.
 #gcode_id:
 #   If set, the temperature will be reported in M105 queries using the
 #   given id. The default is to not report the temperature via M105.
+reverse: False
+#   If True, the working mode of the fan is reversed. If the temperature
+#   is lower than the target temperature, the fan speed increases;
+#   otherwise, the fan speed decreases.
+#   The default is False
 ```
 
 ### [fan_generic]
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  As high-level feedback, it's not clear to me how this is used (and thus how many people might find it useful).  If this is to heat a chamber, wouldn't it be simpler to just turn the fan on during chamber heating?

-Kevin

dewi-ny-je:
Wouldn't changing sign of pid_Kp do the same?

jschuh:
The use case is where the bed_heater doubles as a weak chamber heater by recirculating air under it, typically by mounting radial fans with the intake facing up in front of the bed and the exhaust just underneath. [It's a relatively popular Voron mod](https://github.com/VoronDesign/VoronUsers/tree/master/printer_mods/Ellis/Bed_Fans).

FWIW, I briefly had a printer set up like this, but I controlled it by configuring the fan as a heater_generic with verify_heater settings that ensured it couldn't inadvertently throw a heater_fault. It's a bit hacky, but as I recall it worked fine.

KevinOConnor:
Thanks.  I think I understand.  I'm not sure about this change though.  I'm not sure how many users would find it in the config and utilize it.

One of the things I was looking at was extending the "display_template" system to fan_generic - and thus allowing users to set an arbitrary math formula for the fan speed.  I wonder if this may be more flexible than adding config entries to temperature_fan.

-Kevin

dingyifei:
Sounds good. I'll close this now.

PR #6158, opened 2023-04-08
tmc: external clock support

leptun:
Allow the clock frequency of the tmc driver to be specified in case the user uses an external clock oscillator for the driver so that the velocity based thresholds are calculated correctly.
If the external clock frequency is not specified, it is assumed that the internal clock is the one that is used, so the old behaviour is preserved. The valid range for the clock frequency was set according to the datasheet electrical characteristics values.

In terms of hardware support, the Prusa MK3.5, MK3.9, MK4 and XL need this change since they all use a shared 16MHz external clock for the TMC drivers. Without this change, all velocity based thresholds are calculated incorrectly. There is also support for an external clock with the SilentStepStick TMC5160 v1.3, but I doubt anyone is using that hardware configuration.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3052,6 +3052,9 @@ cs_pin:
 #   define the stepper position in the chain and the total chain length.
 #   Position 1 corresponds to the stepper that connects to the MOSI signal.
 #   The default is to not use an SPI daisy chain.
+#external_clock_frequency: 13200000
+#   Configure only if an external clock oscillator is connected to the TMC
+#   driver clock input.
 #interpolate: True
 #   If true, enable step interpolation (the driver will internally
 #   step at a rate of 256 micro-steps). This interpolation does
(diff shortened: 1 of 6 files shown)


KevinOConnor:
Thanks.  I guess I'm a little unsure on this as it isn't clear to me that there would be very many people that would set this low-level field.  I wasn't aware that prusa had printers that synchronized the tmc clocks.  Do you have a link to the hardware with this support?

-Kevin

leptun:
@KevinOConnor Afaik, the schematics for the XL and MK4 electronics are not public yet. I asked for permission to post some snippets from the datasheet. Hope I'll get an answer soon. And I just noticed that in both printers only XYZ are synchronised, the extruder using the internal oscillator instead.

andreacampanella:
Hi, MK4's electronics schematics are now avaiable: https://www.prusa3d.com/page/open-source-at-prusa-research_236812/

PR #6159, opened 2023-04-09
Improve json encoding error handling

jschuh:
Json encoding errors currently force a shutdown with no context as to why. This can be verified by adding the following to a config, which triggers a silent TypeError:

```
[gcode_macro crasher]
variable_crash: {(1, 1): 1}
gcode:
    # empty
```

This change just logs the error (including a console message) so it's easier to diagnose what went wrong. In the case of the above error, the message is:

```
json encoding error: keys must be str, int, float, bool or None, not tuple
```

_**Edit:** Also added code to prevent setting bad variables inside `gcode_macro.py`._

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2018-2021  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import traceback, logging, ast, copy
+import traceback, logging, ast, copy, json
 import jinja2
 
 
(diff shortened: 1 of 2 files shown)


jschuh:
@Arksine, would you mind taking a look at this when you have a chance? It's a small logging addition that would have saved me quite a few hours of macro debugging (that turned into Klipper webhooks debugging).

Arksine:
In general the change looks ok to me.  I suspect we would want to return in the exception handler as well since there is no additional data to send.  I'm also not sure if we want to invoke a shutdown, that is something Kevin could provide better feedback on.

An alternative option may be to verify that a variable can be encoded to json in `gcode_macro.py`.  If it fails we could raise a config error which would provide more details on exactly what went wrong.

jschuh:
After thinking about it I decided that catching early in gcode_macro is the best solution for the problem I hit, so I added a second commit to this PR.

I also kept the first commit with the logging in webhooks because there are other paths that can get there, so it still seems useful for tracking down errors.

KevinOConnor:
Thanks.

-Kevin

PR #6162, opened 2023-04-11
Function and parameters for retraction and dwell after homing

fredriknk:
Now with "signed off by" commit messages

Adding parameters you can use in printer.cfg to retract toolhead by a specific distance after homing, this is quite importent in sensorless corexy devices because they will experience lockup if you home the y axis with the x axis in connection with the frame (or the other way around), It allso includes a dwell time you can set to prevent retriggering of the other axis which can happen if the next axis starts homing just as the first axis retract is settling. This gives you an easy way to retract the axis before it homes the next one without making a custom homing script.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -188,6 +188,16 @@ position_max:
 #second_homing_speed:
 #   Velocity (in mm/s) of the stepper when performing the second home.
 #   The default is homing_speed/2.
+#homing_retract_after: 0.
+#   Distance to backoff (in mm) after homing to prevent carriage lock
+#   in corexy machines. The home retract after variable uses the same
+#   homing_retract_speed as homing_retract_dist variable adjust speed.
+#   Set this to zero to disable the retract.
+#   The default is 0mm.
+#homing_retract_dwell: 0.
+#   Seconds dwell time after retract to prevent triggering when homing
+#   multiple axes at the same time in sensorless in corexy machines
+#   The default is 0 seconds
 #homing_positive_dir:
 #   If true, homing will cause the stepper to move in a positive
 #   direction (away from zero); if false, home towards zero. It is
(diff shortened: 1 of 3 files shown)


ld3300:
Hi @fredriknk,
I don't disagree that this is helpful, however this functionality can already be defined through the use of homing_override

for example:
```[HOMING_OVERRIDE]
gcode:
  G28 X
  G91
  G1 X10  # moves x 10 mm
  G90
  G28 Y
  G28 Z```
  This could be made even smarter by checking if a specific parameter (x or y) is called and just doing the home and adjust on that.  You could also exclude the gcodes from executing when just z is called.

fredriknk:
I know about the homing override and me and my brother spent two days trying in gcode before i changed the klipper source on our machines. We couldnt Get the homing to behave properly when splitting it up in a homing override snippet. Have you tried this on a sensorless corexy? If you search for "voron homing issues" you Get 10-20 threads with people who cant get the homing_override to work properly with sensorless homing. So most of them seem to just accept a skipped step when the Machine jams in the corner

jakep82:
Its unclear to me what this accomplishes that isn't already covered by the sample macros in the Klipper documentation. I use sensorless homing on 2.4, and I simply copy pasted the sample macro for homing X, made an identical macro for Y, and call them both from my homing_override. It works 100% reliably for me. 

https://www.klipper3d.org/TMC_Drivers.html#using-macros-when-homing

KevinOConnor:
Thanks.  However, I'm not sure this is a good fit for the Klipper master branch at this time.  The main concern I have is that the complexity of understanding these new config items and properly assigning them is equal to or greater than the complexity of adding a macro.

Ultimately, my suspicion is that having the list of steps in a "homing macro" will be easier for designers than having to search the entire config reference to find all the config items which may impact homing.

Cheers,
-Kevin

PR #6163, opened 2023-04-11
Add modules Recore and Thermocouple, add config files for Recore A6 and A7

eliasbakken:
This is the final PR for adding full upstream support for Recore boards revision A6 and A7. 
The PR adds a module called recore that enables the high power input section on the board and sets up the temperature sections for either thermistor (gain 1) or thermocouple (gain 100). There is also support for programming the pull-up and adding a voltage offset. These settings are necessary in order to allow both a thermistor and a thermocouple to be used on the same inputs. The thermocouple modules adds the necessary calculations for type K thermocouples.

### DIFF
--- a/config/generic-recore-a6.cfg
+++ b/config/generic-recore-a6.cfg
@@ -1,0 +1,234 @@
+# Generic Recore A6 config
+
+#[include mainsail.cfg]
+
+[recore]
+revision: A6
+gain_t0: 1
+gain_t1: 1
+gain_t2: 1
+gain_t3: 1
+pullup_t0: 1
+pullup_t1: 1
+pullup_t2: 1
+pullup_t3: 1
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Thanks.  Some high-level comments:
1. The configs will need to follow the example config guidelines at https://www.klipper3d.org/Example_Configs.html (at a minimum point 4 and 7a).
2. The two new modules will need to be added to docs/Config_Reference.md .
3. I'm not really sure I understand what thermocouple.py does - having the config reference may help with review.
4. The name thermocouple.py could cause confusion with the existing src/thermocouple.c .  Though it may be better to rename src/thermocouple.c to something more consistent (like src/sensor_spitemp.c) - not sure.

Cheers,
-Kevin

PR #6166, opened 2023-04-12
Add printer configuration for Anycubic Kobra Go

xiaopeng12138:
This adds support for the Anycubic Kobra Go, based on the Huada HC32f460KCT6 microcontroller.
The configuration is mostly based on the [official Marlin firmware](https://github.com/ANYCUBIC-3D/Kobra_Go). 
You can find more information about this printer on [KobraGoNeoInsights](https://1coderookie.github.io/KobraGoNeoInsights/).
I posted this confi originally on this Reddit post [Install Klipper on Kobra Go or Neo](https://www.reddit.com/r/anycubic/comments/10cwm16/install_klipper_on_kobra_go_or_neo/).
There is a variant of this printer. It is Kobra Neo. They share the same mainboard model, so many of their configs are the same. But I only have Kobra Go home and it looks like the variant of a printer is treated in Klipprer for a new .cfg file. That's why this config is only labeled for Kobra Go, not Neo.

### DIFF
--- a/config/printer-anycubic-kobra-go-2022.cfg
+++ b/config/printer-anycubic-kobra-go-2022.cfg
@@ -9,7 +9,7 @@
 # Installation:
 #  1. Rename the klipper bin to `firmware.bin` and copy it to an SD Card.
 #  2. Power off the Printer, insert the SD Card and power it on.
-#  3. The the LCD will be stuck on the Firmware-update screen. 
+#  3. The the LCD will be stuck on the Firmware-update screen.
 #     Just Wait for 3-5 minutes to ensure the firmware is flashed.
 #  4. After waiting, shutdown the printer and remove the SD Card.
 
@@ -138,9 +138,9 @@ pin: PB13
 
 [fan]
 pin: PB5
-#cycle_time: 0.0000318471 #This may be needed for the part fan. 
+#cycle_time: 0.0000318471 #This may be needed for the part fan.
                           #According to some users, the default value of klipper may kill the fan.
 
-[output_pin enable_pin] 
+[output_pin enable_pin]
 pin: PB6 #This pin enables the bed, hotend, extruder fan, part fan.
 static_value: 1
--- a/config/printer-anycubic-kobra-go-2022.cfg
+++ b/config/printer-anycubic-kobra-go-2022.cfg
@@ -108,16 +108,15 @@ z_hop_speed: 15
 
 [controller_fan controller_fan]
 pin: PB12
+cycle_time: 0.00005 #20kHz
 
 [heater_fan extruder_fan]
 pin: PB13
+cycle_time: 0.00005
 
 [fan]
 pin: PB5
-cycle_time: 0.0000318471
-    #This cycle_time may be needed for the part fan.
-    #According to some users,
-    #the default value of klipper may kill the fan.
+cycle_time: 0.00005
 
 [output_pin enable_pin]
 pin: PB6
--- a/config/printer-anycubic-kobra-go-2022.cfg
+++ b/config/printer-anycubic-kobra-go-2022.cfg
@@ -108,15 +108,13 @@ z_hop_speed: 15
 
 [controller_fan controller_fan]
 pin: PB12
-cycle_time: 0.00005 #20kHz
 
 [heater_fan extruder_fan]
 pin: PB13
-cycle_time: 0.00005
 
 [fan]
 pin: PB5
-cycle_time: 0.00005
+cycle_time: 0.00005 #20kHz
 
 [output_pin enable_pin]
 pin: PB6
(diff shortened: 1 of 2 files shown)


JamesH1978:
good morning,

thank you for submitting this PR Please can you make sure you read https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying particular attention to the signed off by line. and also https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md making sure to include the regression tests as set out in section 4.

Thanks
James

xiaopeng12138:
> good morning,
> 
> thank you for submitting this PR Please can you make sure you read https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying particular attention to the signed off by line. and also https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md making sure to include the regression tests as set out in section 4.
> 
> Thanks James

I'm sorry that I didn't pay attention to the commit format. Should I close this pr or use --force to update my commit?

JamesH1978:
that is entirely up to you.

thanks,
James

xiaopeng12138:
I fixed the commit format issue and made some small changes to the config file. If there has any further problem, please let me know.

test/klippy/printers.test
```diff
@@ -258,6 +258,7 @@ CONFIG ../../config/generic-bigtreetech-skr-pico-v1.0.cfg
 # Anycubic Printers using trigorilla board with the hc32f460
 DICTIONARY hc32f460-serial-PA3PA2.dict
 CONFIG ../../config/printer-anycubic-kobra-plus-2022.cfg
+CONFIG ../../config/printer-anycubic-kobra-go-2022.cfg
```
JamesH1978, line 261: This should be in alphabetical order.

Thanks

JamesH1978 (changes requested):
I cant see anything wrong with the config, as long as this describes a stock out-of-the-box Kobra Go. Only the alphabetical issue in the printers.test needs changing.

Thanks
James

test/klippy/printers.test
xiaopeng12138: It should be good now.
Thank you

JamesH1978 (approved):
Looks good to me

Thanks
James

config/printer-anycubic-kobra-go-2022.cfg
```diff
@@ -0,0 +1,136 @@
...
+
+[heater_fan extruder_fan]
+pin: PB13
+
+[fan]
+pin: PB5
+cycle_time: 0.0000318471
+    #This cycle_time may be needed for the part fan.
+    #According to some users,
+    #the default value of klipper may kill the fan.
```
KevinOConnor, line 120: Thanks.  I'd prefer if we didn't include statements like "the default value of klipper may kill the fan" in the example config files.  I fear it causes confusion among users.  If there is a known problem then I think we should document what the problem is.  If it is not known if there is a problem, then I'd prefer not to speculate on what problems could theoretically occur.

Otherwise, the config looks fine to me.

Cheers,
-Kevin

xiaopeng12138: I support you. The reason I added that is because I don't have any issue on my printer when using default pwm frequency of Klipper, but some other user reported that they have. I did some more digging into official marlin firmware and found the correct hard-coded pwm frequency. It's 20KHz.
https://github.com/ANYCUBIC-3D/Kobra_Go/blob/e675180de25a622936be74ddf364f59715ec3943/source/board/bsp_pwm.cpp#L82
I apologize for my mistake. I will remove the comment and change it to the correct value.

KevinOConnor:

Separately, if you want to use 20Khz switching for the fan, you may want to check if enabling `hardware_pwm: True` works, as software pwm at 20Khz is a little high.

-Kevin

1coderookie:
@KevinOConnor unfortunately `hardware_pwm: True` somehow doesn't work in this case (with the Go/Neo), don't know why tho.

KevinOConnor:
Okay.  Not a big deal.

Separately, I'm not sure how well tested hardware pwm is on the HC32F460.  Maybe @SteveGotthardt knows.

-Kevin

nickdollimount:
Hi there, using this configuration, the size of the bed appears to be set too large at 238. It should be 222 X and Y. With that change, the bed_mesh should also be updated. With the current 238, in Mainsail, pausing a print causes the print head to try move too far out and hits the limit and skips the motor.

Setting position_max to 222 in both stepper_x and stepper_y. Also bed_mesh update below appears to be correct based off the offsets of the probe. After making these changes, the pause print no longer hits the limit and the bed mesh calibration works.

[bed_mesh]
speed: 200
horizontal_move_z: 2.5
mesh_min: 5, 5
mesh_max: 201.2, 212
probe_count: 6, 6

JamesH1978:
position_max is not bed size, but the maximum allowed travel on an axis, which can be more. Usually to allow probe to move out of bed to allow a bigger bed mesh. It should be set to the point where toolhead hits an obstruction. These configs are based on the contributors printer, which should be stock, so we have to rely on the contributor to supply sane numbers. 

@xiaopeng12138 can you confirm these numbers are sane and that you can reach them on your printer?

Thanks
James

xiaopeng12138:
I can confirm that the x-axis value is the physical max position. For y-axis I need to do a test

Issue #5706, opened 2022-08-11
`make menuconfig` not works if the parent directory contains spaces


sl1pkn07:
if the parent directory contains spaces, `make menuconfig` fails

~~~
┌─┤[$]|[sl1pkn07]|[sL1pKn07]|[/media/COSAS/aplicaciones/Creality Ender 3/klipper]|
└───╼  make menuconfig
No change to configuration in '/media/COSAS/aplicaciones/Creality Ender 3/klipper/.config'
Loaded configuration '/media/COSAS/aplicaciones/Creality Ender 3/klipper/.config'
---snip----
~~~

but if move the main directory to one without spaces:

~~~
┌─┤[$]|[sl1pkn07]|[sL1pKn07]|[/media/COSAS/aplicaciones/klipper]|
└───╼  make menuconfig
(Top)
                                                                                                                                                                 Klipper Firmware Configuration
[ ] Enable extra low-level configuration options
    Micro-controller Architecture (Atmega AVR)  --->
    Processor model (atmega2560)  --->







[Space/Enter] Toggle/enter      [?] Help            [/] Search
[Q] Quit (prompts for save)     [ESC] Leave menu
~~~

start works

greetings

sl1pkn07:
ping

sl1pkn07:
ping

Neko-vecter:
The best solution probably use “-” to replace space.

Sineos:
Send a PR?

sl1pkn07:
seems involved python scripts, so no, i cant. sorry

greetings

sl1pkn07:
ping

Neko-vecter:
I think this error is caused by menuconfig instead of klipper. 

More info in https://wiki.st.com/stm32mpu/wiki/Menuconfig_or_how_to_configure_kernel

djessup:
Make and menuconfig both struggle with spaces. This is not a Klipper problem IMO.

sl1pkn07:
ping

Neko-vecter:
I checked the section on `make menuconfig` configuration. This is part of [Kconfiglib](https://github.com/ulfalizer/Kconfiglib) library. 
Which means you need to go to [Kconfiglib](https://github.com/ulfalizer/Kconfiglib) and open an issue.

sl1pkn07:
https://github.com/ulfalizer/Kconfiglib/issues/125

done

sl1pkn07:
ping again

PR #6168, opened 2023-04-13
makefile: Handle path with name with spaces

sl1pkn07:
Fix https://github.com/Klipper3d/klipper/issues/5706

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,10 @@ OUT=out/
 
 # Kconfig includes
 export KCONFIG_CONFIG     := $(CURDIR)/.config
--include $(KCONFIG_CONFIG)
+
+# Handle path with spaces
+ESC_KCONFIG_CONFIG = $(subst $() $(),$()\ $(),$(KCONFIG_CONFIG))
+-include $(ESC_KCONFIG_CONFIG)
 
 # Common command definitions
 CC=$(CROSS_PREFIX)gcc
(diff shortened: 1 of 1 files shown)


thijstriemstra (approved):

thijstriemstra:
Nice. Make sure your commits follow(ed) this: https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#format-of-commit-messages

sl1pkn07:
Hi. something like now?

greetings

sl1pkn07:
i just follow this https://stackoverflow.com/a/30691137 but adapted to change from space to scaped space

i have tested quoted all KCONFIG_CONFIG in the makefile rules. but none of them works, in special https://github.com/Klipper3d/klipper/blob/40b4b34998c0708b2050c394feb383f901b76a0b/Makefile#L99 and https://github.com/Klipper3d/klipper/blob/40b4b34998c0708b2050c394feb383f901b76a0b/Makefile#L107

i'm not coder, so im certainly can make other alternative. but this PR simply works as spected

greetings

sl1pkn07:
https://stackoverflow.com/questions/10571658/gnu-make-convert-spaces-to-colons

seems this also works

~~~patch
diff --git a/Makefile b/Makefile
index c112a832..8b037a1e 100644
--- a/Makefile
+++ b/Makefile
@@ -11,10 +11,7 @@ OUT=out/
 export KCONFIG_CONFIG     := $(CURDIR)/.config
 
 # Handle path with spaces
-null      :=
-SPACE     := $(null) $(null)
-ESC_SPACE := $(null)\ $(null)
-ESC_KCONFIG_CONFIG = $(subst $(SPACE),$(ESC_SPACE),$(KCONFIG_CONFIG))
+ESC_KCONFIG_CONFIG = $(subst $() $(),$()\ $(),$(KCONFIG_CONFIG))
 -include $(ESC_KCONFIG_CONFIG)
 
 # Common command definitions
~~~

is better?

greetings

KevinOConnor:
Thanks.  I guess I'm still not sure on this.  It seems simpler to me to just install Klipper in a directory without spaces.  I'm also concerned about other tools and functionality that may break even if we fix this part of the build.

-Kevin

eamaclean:
It seems unreasonable to me to expect users to not use valid paths on their filesystem.

sl1pkn07:
with "other tools" you mean external tools? well, imo, then should be fixed in their respective tools

greetings

sl1pkn07:
sad :(

PR #6171, opened 2023-04-14
added dynamic pa macro

SphaeroX:
I had this created using Chat GPT and made a small bug fix the values are really amazing

### DIFF
--- a/config/sample-macros.cfg
+++ b/config/sample-macros.cfg
@@ -308,3 +308,24 @@ gcode:
     {% set b_axis_pot_scale = printer.configfile.config["mcp4018 b_axis_pot"].scale|float %}
     SET_DIGIPOT DIGIPOT=b_axis_pot WIPER={ b_axis_pot_scale * (b_value / 127.0)}
   {% endif %}
+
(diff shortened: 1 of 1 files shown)


SphaeroX:
BOWDEN_LENGTH in centimeters
LAYER_HEIGHT and NOZZLE_SIZE in millimeters
PRINT_SPEED in millimeters per secound

jordan-woyak:
At the risk of falling for a punking by commenting here..
The math doesn't make any sense and the resulting values are unsurprisingly completely off the wall, so I guess "really amazing" is accurate.
April 1st was 2 weeks ago.

JamesH1978:
Please follow guideline set out in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially the signed-off-by line and naming conventions. But i am in agreement, anything Chat GPT outputs has to be taken with a grain of salt, and has no place in real code, unless thoroughly tested and proved to be correct, and as we have already had one reply to that end, i am not sure this will be accepted as good code for klipper.

Thanks
James

PR #6173, opened 2023-04-17
fix install issue on ubuntu 22.04

dhardy35:
Hi,
There are dependencies to python2 that were not taken into account in the install-ubuntu-22.04.sh script.
This patch manage them and is working at least in simulation mode and inside a docker image.
Signed-off-by: Damien Hardy damien.hardy@irisa.fr

### DIFF
--- a/scripts/install-ubuntu-22.04.sh
+++ b/scripts/install-ubuntu-22.04.sh
@@ -10,7 +10,7 @@ KLIPPER_GROUP=$KLIPPER_USER
 install_packages()
 {
     # Packages for python cffi
-    PKGLIST="virtualenv python3-dev libffi-dev build-essential"
+    PKGLIST="virtualenv python2-dev libffi-dev build-essential python2-pip-whl python2-setuptools-whl"
     # kconfig requirements
     PKGLIST="${PKGLIST} libncurses-dev"
     # hub-ctrl
@@ -18,7 +18,7 @@ install_packages()
     # AVR chip installation and building
     PKGLIST="${PKGLIST} avrdude gcc-avr binutils-avr avr-libc"
     # ARM chip installation and building
-    PKGLIST="${PKGLIST} stm32flash dfu-util libnewlib-arm-none-eabi"
+    PKGLIST="${PKGLIST} stm32flash libnewlib-arm-none-eabi"
     PKGLIST="${PKGLIST} gcc-arm-none-eabi binutils-arm-none-eabi libusb-1.0"
 
     # Update system package info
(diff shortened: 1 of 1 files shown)


JamesH1978:
thank you for submitting a PR, please refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying particular attention to the signed-off-by line and PR naming conventions.

Thanks
James

Sineos:
Seems to contradict https://github.com/Klipper3d/klipper/pull/6112

dhardy35:
Signed-off-by is added in my previous comment.
That's exactly the merge #6112 where I propose a patch because it does not work.

Sineos:
Well, as much as I understand the intention of #6112 was to use python3 and your change would bring it back to python2.

@Jesssullivan can you check and comment, please

Jesssullivan:
Thanks for the ping @Sineos-

@JamesH1978, I am unsure why you'd like to use python2 here, it has been deprecated since January of 2020 and not a default package for 22.04 lts for this reason.  Is there a specific reason you are keen to enforce the use of python2 here?  The two wheel packages you've added here  `python2-pip-whl python2-setuptools-whl` are no longer maintained.  I am also unsure why you've removed the `dfu-util` package, this is an important package used to flash our MCUs.

dhardy35:
@Jesssullivan I do not want to use python2 stuff but I have to. Some bugs occur when using the script with python3 :-/
BTW I wrote that script when yours was not online and I started from the 18.04 version.
For the dfu-util package I can add it. The reason why this one is not present is that it was not present in the 18.04 version.
I am still investigating why it failed with python3, I am going to put an engineer on that point but for sure it will take some times. If you prefer to wait for it, I understand.
Best

flowerysong:
"Some bugs occur" with no indication of what those bugs are is not a persuasive argument for moving to a deprecated Python version. What bugs? Did they occur after you used this installation script, or did you use a different installation method?

dhardy35:
Hi,
Finally catch up with this problem. The behavior of python2 and python3 is different when parsing a config file, at least a dedicated section in it. when starting an information with ';' in python2 no problem while with python3 the info is just ignore and the result is an empty string.
Example:
[my_section]
field = ; value

flowerysong:
Use of `;` as an inline comment delimiter is deliberately enabled under Python 3: https://github.com/Klipper3d/klipper/blob/645a1b8364c3110f706db0f976ac5fa20b968c36/klippy/configfile.py#L264-L265

Between that and the fact that [Python 2's ConfigParser](https://docs.python.org/2/library/configparser.html) is supposed to treat `;` as an inline comment delimiter, you appear to have found a bug that only occurs under Python 2 and that you were relying on continuing to be buggy, rather than a bug that occurs under Python 3. While making adjustments to your config might be a slight inconvenience for you, it's better to do that than to rely on an obsolete version of Python.

PR #6174, opened 2023-04-18
Fix stm32 usbotg irq wakeup notification

KevinOConnor:
The stm32 usb code on stm32f4, stm32f2, and stm32h7 devices will unnecessarily wake up the receive handlers.  This can cause confusingly high "awake time" statistics.

In particular, when using "usb to canbus bridge" mode it could result in effectively continuous wakeups leading to very high "awake time" reports.

These spurious wakeups are unlikely to directly cause a problem, but they make it harder to diagnose other issues.

-Kevin

### DIFF
--- a/src/stm32/usbotg.c
+++ b/src/stm32/usbotg.c
@@ -397,11 +397,11 @@ OTG_FS_IRQHandler(void)
     }
     if (sts & USB_OTG_GINTSTS_IEPINT) {
         // Can transmit data - disable irq and notify endpoint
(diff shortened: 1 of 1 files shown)


JamesH1978:
Kevin, will this trigger a mandatory reflash/firmware out of date error?

Thanks
James

KevinOConnor:
> will this trigger a mandatory reflash/firmware out of date error?

No.

-Kevin

PR #6176, opened 2023-04-18
New parsecandump.py utility to parse Klipper messages in a candump capture

KevinOConnor:
This adds a new `scripts/parsecandump.py` tool to help parse Klipper mcu commands found in the output from `candump -t z -Ddex can0,#FFFFFFFF`.  It works similarly to the existing `klippy/parsedump.py` tool.

-Kevin

### DIFF
--- a/scripts/parsecandump.py
+++ b/scripts/parsecandump.py
@@ -1,0 +1,125 @@
+#!/usr/bin/python3
+# Check for out of order timestamps in the output of candump
+#
+# Copyright (C) 2023  Kevin O'Connor <kevin@koconnor.net>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import sys, os, optparse
+
+def import_msgproto():
+    global msgproto
+    # Load msgproto.py module
+    kdir = os.path.join(os.path.dirname(__file__), '..', 'klippy')
+    sys.path.append(kdir)
+    import msgproto
+
+def read_dictionary(filename):
+    dfile = open(filename, 'rb')
+    dictionary = dfile.read()
+    dfile.close()
+    return dictionary
+
+def report(msg, line_info, name="", is_err=False):
+    line_number, line_time = line_info
+    warn = ""
+    if is_err:
+        warn = " WARN"
+    sys.stdout.write("%04d:%010.6f:%s%s %s\n"
+                     % (line_number, line_time, name, warn, msg))
+
+class canscan:
+    def __init__(self, name, mp):
+        self.name = name
+        self.mp = mp
+        self.data = bytearray()
+        self.need_scan = False
+    def handle_data(self, line_info, line, newdata):
+        data = self.data
+        data += bytearray(newdata)
+        while 1:
+            if self.need_scan:
+                drop = len(data)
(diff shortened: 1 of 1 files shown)


scripts/parsecandump.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/python3
+# Check for out of order timestamps in the output of candump
+#
+# Copyright (C) 2023  Kevin O'Connor <kevin@koconnor.net>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import sys, os, optparse
```
thijstriemstra, line 7: from the python docs:

```
Deprecated since version 3.2: The [optparse](https://docs.python.org/3/library/optparse.html#module-optparse) module is deprecated and will not be developed further; development will continue with the [argparse](https://docs.python.org/3/library/argparse.html#module-argparse) module.
```

Please use argparse instead.

scripts/parsecandump.py
```diff
@@ -0,0 +1,125 @@
...
+    # Load msgproto.py module
+    kdir = os.path.join(os.path.dirname(__file__), '..', 'klippy')
+    sys.path.append(kdir)
+    import msgproto
+
+def read_dictionary(filename):
+    dfile = open(filename, 'rb')
+    dictionary = dfile.read()
+    dfile.close()
+    return dictionary
```
thijstriemstra, line 20: this can be simplified using a `with` statement:

```python
with open(filename, 'rb') as f:
    return fdile.read()
```

scripts/parsecandump.py
```diff
@@ -0,0 +1,125 @@
...
+        opts.error("Incorrect number of arguments")
+    canfilename, canid, dictfilename = args
+    canid = int(canid, 16)
+
+    import_msgproto()
+    dictionary = read_dictionary(dictfilename)
+
+    canfile = open(canfilename, "r")
+    read_candump(canfile, canid, dictionary)
+    canfile.close()
```
thijstriemstra, line 122: this can also be simplied with an `with` statement.

thijstriemstra (changes requested):

Wulfsta:
Perhaps this is out of scope for what you hope to achieve with this script, but have you considered using `subprocess.Popen` to call candump and actively process the output?

KevinOConnor:
> Perhaps this is out of scope for what you hope to achieve with this script, but have you considered using subprocess.Popen to call candump and actively process the output?

We've seen a number of cases with "mysterious canbus failures".  The goal here was to allow users to run `candump -t z -Ddex can0,#FFFFFFFF > mycanlog` at the start of a test, then run a print, and then grab the resulting canbus log if a fault occurs.  That log could then be analyzed by developers with the new `parsecandump.py` tool here.  Specifically, this tool helps provide details on Klipper retransmits, lost messages, and command/response timing.

I've placed this PR on hold though, pending documentation for the tool.  Without documentation I'm not sure there is much value in adding the tool to the repo.

Cheers,
-Kevin

KevinOConnor:
I'm closing this PR - the tool (along with documentation) is now part of PR #6215 .

-Kevin

PR #6178, opened 2023-04-21
aht10: update comments and docs for AHT20/AHT21 support

scottmudge:
The control bytes for the AHT20 and AHT21 are identical to the AHT10, but I had not been able to test just yet to ensure the sensor code supported the other sensors. I've wrapped up testing and updated the comments/docs to reflect this additional support.


KevinOConnor:
Thanks.

-Kevin

PR #6179, opened 2023-04-23
Per-tool Z probe.

viesturz:
Hey, Klipper devs. This mod got a bit involved and maybe there is a chance to get this upstreamed.

I have a Toolchanger setup where each toolhead has it's own bed probe and there is no independant bed probe.
So this adds a new virtual probe Z endstop that can be reassigned between multiple probes.

The code here is very much an initial concept, happy to hear your feedback and make changes as needed.

### DIFF
--- a/klippy/extras/tool_probe.py
+++ b/klippy/extras/tool_probe.py
@@ -1,0 +1,54 @@
+# Per-tool Z-Probe support
+#
+# Copyright (C) 2023 Viesturs Zarins <viesturz@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+from . import probe
+from .probe import ProbeEndstopWrapper
+
+
+class ToolProbe(probe.PrinterProbe):
+    def __init__(self, config, mcu_probe):
+        self.tool = config.getint('tool')
+        # The below is a partial copy from PrinterProbe.__init__ except a bunch
+        # of registrations moved to the endstop.
+        self.printer = config.get_printer()
+        self.name = config.get_name()
+        self.mcu_probe = mcu_probe
+        self.speed = config.getfloat('speed', 5.0, above=0.)
+        self.lift_speed = config.getfloat('lift_speed', self.speed, above=0.)
+        self.x_offset = config.getfloat('x_offset', 0.)
+        self.y_offset = config.getfloat('y_offset', 0.)
+        self.z_offset = config.getfloat('z_offset')
+        self.probe_calibrate_z = 0.
+        self.multi_probe_pending = False
+        self.last_state = False
+        self.last_z_result = 0.
+        self.gcode_move = self.printer.load_object(config, "gcode_move")
+        # Infer Z position to move to during a probe
+        if config.has_section('stepper_z'):
+            zconfig = config.getsection('stepper_z')
+            self.z_position = zconfig.getfloat('position_min', 0.,
+                                               note_valid=False)
+        else:
+            pconfig = config.getsection('printer')
+            self.z_position = pconfig.getfloat('minimum_z_position', 0.,
+                                               note_valid=False)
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I'm not really sure what this code does.  I don't have this hardware, so it's hard for me to review or give feedback.  Perhaps you could explain what goes wrong with the current code, which users will make use of this new functionality, and how they would go about using it.

Separately, this code seems to be making changes to stepper.py which seems odd for probing code.

Also, note (as James indicated above) that new contributions would need a signed-off-by line.

Cheers,
-Kevin

viesturz:
Hey, Updated the pull request with a better description. Including the meaning of stepper.py changes.
I agree that this is pretty niche feature, hence I'm looking for early feedback if this makes sense in broader Klipper environment.

The long story:
Toolchanger printers have multiple tools and a automated tool changing sequence to remove a tool and install another during print.

Generally some hotend equipend is shared and remains on the shuttle while the rest is docked and replaced with the other tool.
In my case, the shuttle has no equipment at all, which **removes the need to run any wiring to it**. See [this video](https://www.youtube.com/watch?v=4ZeR229iMq8&list=PLqU7kX5nUJDRDw5z0NLwJ22OkV6fbjnSW&index=6)

Which means there is no dedicated Z endstop on the toolhead anymore.
But each hotend has it's own nozzle operated bed probe, which turns out has quite a few advantages:

- automated tool offset adjustments after nozzle changes by comparing Z probe height
- use the probe as docking switch, allowing to auto-detect which hotends are docked and which one is mounted.

But leaves with a Z homing problem, with few options:

- use a dedicated Z enstop - need to add an extra endstop that is not on the gantry, tricky for self bed leveling setups, like Voron Trident or 2.4
- designate one tool to be the Z endstop, which means you need to swap into that tool before you can home. Which might be impossible if Z homing is required to change tools.
- **this patch** dynamically assign a probe as the Z endstop 

You configure it with 
```
[tool_probe_endstop] 
# registers virtual endstop pin

[tool_probe tool0]
tool: 0
z_offset: -1.68
# regular probe params...

[tool_probe tool1]
tool: 1
z_offset: -1.68
# regular probe params...
```

And control with
```
SET_ACTIVE_TOOL_PROBE T=0
```
or autodetect active tool with
```
DETECT_ACTIVE_TOOL_PROBE
```
and access active tool with
```
printer.tool_probe_endstop.active_tool_number
```
This is all very preliminary. 

I can see 3 ways forward:

- incorporate this patch in full, Yay!
- incorporate only the stepper.py changes, allowing the rest to exist as an Klipper extension script
- keep it separate, which is quite cumbersome to maintain.

Cheers!

KevinOConnor:
Okay, thanks for the description.  As high-level feedback I don't think the approach taken in this PR is ideal.

I think we want to keep the stepper code related just to movement and I don't think it should be changed to facilitate different types of probes.  The PrinterRail `position_endstop` is intended to be a position on the physical rail that a carriage can not be moved past - that position on the rail does not change with different tools.

For what it is worth, I think enabling/disabling different rails (see the dual_carriage code for an example) or selecting between multiple probes may be something to look at as an alternative.

-Kevin

viesturz:
Thanks for the input, I will see what I can do.
Can def play with non kinematic moves...

One nit - isn't the min/max positions meant to be the physical limits?

KevinOConnor:
> One nit - isn't the min/max positions meant to be the physical limits?

Sorry, yes you are correct - the position_min/max are the maximum travel for the rail.  I guess I better understand the issue you are having (the tool position when hitting the endstop is different with different tools).  However, I'm still not sure it is a good idea to modify stepper.py.  Alternatives may be to swap rails for different tools (eg, dual_carriage) or to apply a gcode offset with each new tool.

-Kevin

PR #6182, opened 2023-04-25
Handle USB resets on rp2040

KevinOConnor:
On a USB bus reset the controller address needs to be reset to zero. Otherwise, the device becomes unresponsive after a USB reset and will remain unresponsive until a power cycle.  This often shows up in the Linux system logs as "device descriptor read/64, error -32" messages.

Make sure USB_INTE_BUS_RESET_BITS is enabled and clear the USB address on a bus reset.

### DIFF
--- a/src/rp2040/usbserial.c
+++ b/src/rp2040/usbserial.c
@@ -96,7 +96,8 @@ int_fast8_t
 usb_read_ep0_setup(void *data, uint_fast8_t max_len)
 {
     if (!(usb_hw->intr & USB_INTR_SETUP_REQ_BITS)) {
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I merged this change.

-Kevin

hubertsmits:
It's now January 2024, and I have the same problem with a Pi 4 and Pico. I'm running Klipper v0.12.0-100 and the Pi (and other subsystems) are all up-to-date. Problem occurs both with UART and USB. Am I overlooking something?

hubertsmits:
Never mind, after reflashing the Pico (which I did many times before) it now boots the Pi 4, and shows the Pico connected. Without using the reset button. The only thing bugging me is that I don't understand what changed. But I have my documentation for future challenges.

PR #6183, opened 2023-04-26
Bug fix: Changing dual-carriage carriage will inadvertently set axis …

Frans-Willem:
…to homed, even if it wasn't.

in hybrid_corexy and hybrid_corexz, the update_limits function is (only) called by the DualCarriage implementation, whenever the carriage changes. Unfortunately, these limits also keep track of homing status, when they're unhomed they are set to 1 to -1 (invalid range). As a fix, if the limit was set to "unhomed", we keep it that way, and only update it with the new rail limits if it was already properly homed before.


KevinOConnor:
Maybe @Tircown has some thoughts on this?

-Kevin

Tircown:
It looks fine to me. Thanks for patching this issue.

Frans-Willem:
Squashed both commits, and changed commit message according to "Contributing to Klipper" documentation. Let me know if I've missed anything else :)

KevinOConnor:
Thanks.

-Kevin

PR #6185, opened 2023-04-27
New printer configuration : Longer LK4 X

bistory:
New configuration file for the Longer LK4 X stock printer (with direct drive and bltouch).
This configuration is based on the one provided by Longer with some tweaks.

Tested and approved by myself, it just needs a small review in case I forgot something :-)

### DIFF
--- a/config/printer-longer-lk4x-2022.cfg
+++ b/config/printer-longer-lk4x-2022.cfg
@@ -1,0 +1,164 @@
+# This file contains pin mappings for the stock 2022 LONGER3D LK4 X
+# with the 32-bit LGT_KIT_V2_X board. To use this config, during
+# "make menuconfig" select the STM32F103 with a "32KiB bootloader" and
+# serial (on USART1 PA10/PA9) communication.
+
+# Flash this firmware by copying "out/klipper.bin" to a SD card, 
+# then rename it to "firmware.bin"and turning on the printer with 
+# the card inserted.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+[printer]
+kinematics: cartesian
+max_velocity: 300
+max_accel: 4000
+max_z_velocity: 5
+max_z_accel: 100
+
+[stepper_x]
+step_pin: PD3
+dir_pin: PD2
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thank you for submitting your config, please refer to point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md to add the regression tests.

I will do a full review once this is done, but from a cursory glance, please do not put pressure advance configuration in [extruder] this is a tuneable option that may be different for the end user depending on filament used. And please do not add any macros that do not directly relate to base functionality of the printer, see point 6 iv in the linked doc. start and end macros are a choice made by the user.

Thanks
James

bistory:
Thanks for your feedback @JamesH1978 !
I have fixed those and I have reviewed the configuration, I think it is good now.

Thomas

config/printer-longer-lk4x-2022.cfg
```diff
@@ -0,0 +1,118 @@
...
+z_offset: 0 # You need to adjust this for your printer
+
+[bed_mesh]
+speed: 120
+mesh_min: 10, 10
+mesh_max: 210, 210
+probe_count: 4, 4
+
+[safe_z_home]
+home_xy_position: 110, 110
```
JamesH1978, line 115: This should usually be set to put the probe central rather than nozzle when using a probe/bltouch setup, so with your stated offsets `home_xy_position: 161, 118` this will allow `PROBE_CALIBRATE` to move the nozzle central for the paper test.

Thanks
James

JamesH1978 (changes requested):
Apart from my 1 extra comment, this is near enough ready to be approved.

Please make that change.

Thanks
James

config/printer-longer-lk4x-2022.cfg
bistory: Thanks ! I have made the change 👍

JamesH1978 (approved):
Looks fine to me.

Thanks
James

KevinOConnor:
Thanks.

-Kevin

PR #6188, opened 2023-04-28
linux: Add options for setting group and mode

Laikulo:
This adds support for providing a group name and permission bits to the linux process "mcu". This is useful in situations where the process connecting to klipper is running as a different user, or if you want a more restrictive mode than provided by default (such as where all users have a common primary group)

This will only allow changes to a group that the user is already a member of (either supplemental or primary)


src/linux/main.c
```diff
@@ -81,8 +87,28 @@ main(int argc, char **argv)
...
+        case 'g':
+            pty_group_name = optarg;
+            break;
+        case 'm':
+            errno = 0; // see man 2 strtoul
+            unsigned long int input_pty_mode = strtoul(optarg, NULL, 8);
+            if (errno) {
+                fprintf(stderr, "unable to parse pty mode: %s", optarg);
+                return -1;
+            }
```
KevinOConnor, line 99: FWIW, despite the man page recommendation I'm not sure it is a great idea to inspect `errno` without first checking the return value.  FWIW, since `strtoul()` will return `ULONG_MAX` and that will already report a "mode invalid" in your code, I'm not sure the `errno` checks are necessary.

KevinOConnor (commented):
Thanks.  Looks fine to me.  Since @mattaw was recently working on this code, maybe he has some comments.

I have a minor comment (not a blocker) on the code - see below.

-Kevin

src/linux/main.c
Laikulo: Looking at the man page again, it appears that the ULONG_MAX and ULONG_MIN clamping behavior is only for out of range failures
```
       The  strtoul() function returns either the result of the conversion or, if there was a leading minus sign, the
       negation of the result of the conversion represented as an unsigned value, unless  the  original  (nonnegated)
       value would overflow; in the latter case, strtoul() returns ULONG_MAX and sets errno to ERANGE.  Precisely the
       same holds for strtoull() (with ULLONG_MAX instead of ULONG_MAX).
```
but failed conversions return 0, and errno is the only indication that something failed.
```
       The implementation may also set errno to EINVAL in case no conversion was performed (no digits seen, and 0 re-
       turned).
```
Though it does say that
```
       Since  strtoul()  can legitimately return 0 or ULONG_MAX (ULLONG_MAX for strtoull()) on both success and fail-
       ure, the calling program should set errno to 0 before the call, and then determine if  an  error  occurred  by
       checking whether errno has a nonzero value after the call.
```

Laikulo: Though, on second thought, I probably should stick that through errstr, or at least give the error

KevinOConnor: > Looking at the man page again, it appears that the ULONG_MAX and ULONG_MIN clamping behavior is only for out of range failures

Oops.  My mistake.

-Kevin

Laikulo: NP. I won't admit how many times I had to read that manpage the first time round.

mattaw:
Looking this over, this request is changing the ownership of the console pty, but not the actual MCU process. Is this the intended behavior, or do you want to run the whole thing as another user and group?

The simplest way is to edit the systemd file, `/etc/systemd/system/.....service` to add in the desired user and group: https://www.golinuxcloud.com/run-systemd-service-specific-user-group-linux/

I also have a strategy for changing the user and group of a program, by detecting that the current process is not right and literally re-executing the MCU after running it with the right user and group. Feels hacky but worked around some limitations.

Laikulo:
@mattaw That is the intention. In my setup the klipper_mcu runs as it's own user, separate from the klipper user. The both the klipper_mcu user and the klippy user have klipper_mcu_tty as a supplemental group, this allows klippy to access the mcu's console without giving it access gpio, or anything else that I gave the MCU, or anything that the klipper_mcu group has access to. (I have a udev rule with GROUP=klipper_mcu)

I initially wrote this with just the chmod functionality, after realizing that it was setting the mode explicitly, and not just picking from `umask`, at the time, I was running the services as a single user, and wanted to set 0600, but was unable to.

As far as switching groups for the process goes. I am pretty sure that process can switch to any of it's supplemental groups just by calling setgid, though there aren't many cases where you would want to do this.

mattaw:
Yeah, switching user is typically through invoking a setuid helper/SHIM, or similar strategies. The re-exec strategy I mentioned was the only one that allowed switching with a non-root user, and it felt wrong.

In this case, can you help me understand better what the real-world problem you are trying to solve is, which is not addressed by setting a user and group through systemd, after creating a group shared with klippy? Is a security concern and you want minimum overlap between klippy and the linux MCU process?

Laikulo:
Minimum overlap/privilege is the main reason for the setgid behavior.

In my current setup:
klipper_mcu group has unrestricted access to gpio.
klippy is running under a different uid,gid, but is a member of klippy_mcu_tty
If I add klippy to klipper_mcu, then it gets unrestricted gpio, which I do not want.

I'm currently managing the klippy api domain socket with a combination of umask and the setgid bit on a directory in /var/run managed by systemd-tmpfilesd, this results in the domain socket getting 0660 with klippy:klippy_tty without any special behavior from klippy.

It occurs to me that that I've not yet checked to see if klippy respects umask when creating its pty, which means that it might have greater access than I really want it to (since I pretty much only use it with the virtual SD card and UDS/API commands via moonraker, I would prefer it to be inaccessible to everything except human troubleshooting).

SkyKnightSKS:
This is closely related to the PR (https://github.com/Klipper3d/klipper/pull/6203) I created for klippy.
At first I also added arguments to change the permissions on the socket. But it was argued to have as little settings as possible. (https://github.com/Klipper3d/klipper/pull/6194)

So as an alternative systemd socket activation was suggested which works perfectly fine.
I think we should align these 2 to have the same behaviour. Either we go for both the setting permissions way.
Or we go for both the systemd sockets way.

KevinOConnor:
Thanks.  What is the status of this PR?  It seems like this functionality isn't needed because systemd can already perform the required functionality?

-Kevin

Laikulo:
> Thanks.  What is the status of this PR?  It seems like this functionality isn't needed because systemd can already perform the required functionality?
> 
> -Kevin

There are two features in here:
 1) setting the mode of the pty, which systemd (or whatever init) has no control over, because the current implementation ignores the umask, and just sets it to a static value.

2) setting the group of the pty to a supplemental group. Systemd also can't do this, but setting the group of the entire process probably covers most use cases.

Laikulo:
Rebased on currrent master

PR #6190, opened 2023-04-30
klippy: report git branch, remote, and tracking URL

Arksine:
This Pull Request builds on #6184, logging additional repo info useful for determining if a user is running modified code.

### DIFF
--- a/klippy/klippy.py
+++ b/klippy/klippy.py
@@ -343,7 +343,7 @@ def main():
     else:
         logging.getLogger().setLevel(debuglevel)
     logging.info("Starting Klippy...")
-    git_vers, git_status = util.get_git_version()
+    git_vers, git_status, repo_info = util.get_git_version()
     extra_files = [fname for code, fname in git_status
                    if (code in ('??', '!!') and fname.endswith('.py')
                        and (fname.startswith('klippy/kinematics/')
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  Looks useful.  I'll give another day or so on #6184 and then look to merge.

As a minor note (not a blocker), it might make sense to change `get_git_version()` to return a dictionary with "version", "file_status", "branch", "remote", and "url" fields.

-Kevin

Arksine:
Hi Kevin, I rebased and added a couple of commits.  

The [d2123cb](https://github.com/Klipper3d/klipper/pull/6190/commits/d2123cbcc2a10cc71270ff554dcae15191851d89) commit returns the result as a `dict` as you suggested.  One additional tweak in this commit is to return the dict as long `git describe` returns success.

The [a3f1469](https://github.com/Klipper3d/klipper/pull/6190/commits/a3f14697d09f6194609e2d727805701322169b91) commit outputs all of the git info as a json object in `make_version.py`.  I though this might be valuable, however I will drop the commit if you feel that it isn't necessary.

KevinOConnor:
Thanks!  First two commits look good to me.  I'm not sure on the third - I'm a little leery of making the file version look like the git version in the logs - for fear that it could confuse someone reviewing the log into thinking that the log was generated from a particular code version when we really don't actually know what code was running (the version file could be stale).

-Kevin

Arksine:
Make sense, I dropped the last commit.

KevinOConnor:
Thanks.

-Kevin

KevinOConnor:
> The information is apparently only added at Klipper start / restart but not on a firmware restart.

The information should be written to the log after every log file rollover.  So, it should always be at the very beginning of the log file.

-Kevin

PR #6192, opened 2023-05-05
Adding a config for the anycubic-chiron

gbit-is:
This is a fully working config for the anycubic chiron, from my research anycubic stopped shipping them with a bootloader installed sometime between 2018 and 2021.

I don't see those kind of details included in other configs but it may be relevant ?
Here is my write-up on getting a bootloader on the trigorilla v.0.0.2
https://github.com/gbit-is/Trigorilla-v0.0.2-burn-bootloader/

There is one thing I wasn't sure about, that is if I should keep the extruder rotation_distance value I measured and tuned or round it up to the nearest whole number

[extruder]
rotation_distance: 7.71

vs.

[extruder]
rotation_distance: 8

### DIFF
--- a/config/printer-anycubic-chiron-2021.cfg
+++ b/config/printer-anycubic-chiron-2021.cfg
@@ -51,7 +51,7 @@ step_pin: PA4
 dir_pin: PA6
 enable_pin: !PA2
 microsteps: 16
-rotation_distance: 7.71 
+rotation_distance: 7.71
 nozzle_diameter: 0.400
 filament_diameter: 1.750
 heater_pin: PB4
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thankyou for submitting this config, please could you refer to point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and other information in that doc and include the regression test, and also https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md paying particular attention to the signed-of-by line

In terms of the rotation distance, end users are expected to tune this themselves, so it doesnt really matter.

Thanks
James

PR #6193, opened 2023-05-05
Reading steppers' mcu_position from macros

Piezoid:
**tl;dr**: This PR introduces a `get_status()` function for stepper objects, allowing macros and frontends to access the internal step counter `mcu_position`. This feature is especially useful for detecting skipped steps during calibration routines.

*edit 110823:* I shortened the OP message to better reflect the reduced scope of the PR after discussion, also replacing the "zero dwells" recommendation with `M400`. To see the original proposal, please lookup the Github edits.

Some macros aim to identify skipped steps, like [Ellis's macro](https://ellis3dp.com/Print-Tuning-Guide/articles/determining_max_speeds_accels.html#determining-if-skipping-occured) for determining a printer's absolute maximum speeds and accelerations. This relies on users comparing the "mcu" positions of two GET_POSITION outputs. Unfortunately, there is currently no way to read a stepper's step position from a macro, which hinders the automated detection of skipped steps.

While GET_POSITION resides in `gcode_move`, I decided against adding overhead to its `get_status()` method with multiple FFI calls per stepper, as this method is frequently queried or subscribed by front-ends. Instead, this PR adds a `get_status()` method to `MCU_stepper` objects. Additionally, this PR registers newly created MCU_stepper objects in PrinterStepper using their config section name, except for the extruder stepper, which would collide with the `extruder` printer object. The ad-hoc rule is that the section name must contain the sub-string "stepper".

*Important Note:* To ensure accurate results when using this status, a pipeline flush is required, as discussed in the Klipper3d documentation. Executing M400 before evaluating a macro that reads mcu_position is recommended to fully flush the look-ahead and step generation code. The same applies to GET_POSITION's "mcu" and "stepper" lines.

<details>
<summary>Gcode macros demonstrating the effect of pipeline flushes:</summary>

```
[gcode_macro TEST_REPORT_X_POSITION]
gcode:
    REPORT_X_POSITION MSG="Before first move"
    G1 X100
    REPORT_X_POSITION MSG="After first move"
    M400
    REPORT_X_POSITION MSG="After first wait_moves"
    M400
    REPORT_X_POSITION MSG="After second move"
    M400
    REPORT_X_POSITION MSG="After second wait_moves"

[gcode_macro REPORT_X_POSITION]
gcode:
    M118 {params.MSG} stepper_x:{printer["stepper_x"]["mcu_position"]}
```

The lines indicating "After * move" report  old positions, whereas the lines with "After * wait_moves" are current. Additionally, it is important to keep in mind that the flush should be placed in a separate macro from the one used to evaluate the status.
</details>

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -399,6 +399,23 @@ The following information is available in
 - `printer["servo <config_name>"].value`: The last setting of the PWM
   pin (a value between 0.0 and 1.0) associated with the servo.
 
+## stepper_*
+
+The following information is available in the `stepper_*`, and
+`manual_stepper some_name`:
+- `mcu_position`: The total number of steps the micro-controller has issued
+  in a positive direction minus the number of steps issued in a negative
+  direction since the micro-controller was last reset. This is the "mcu"
+  position reported by `GET_POSITION`. See the developer documentation of
+  [GET_POSITION output](Code_Overview.md#coordinate-systems) for more
+  information.
+
+If the robot is in motion when the query is issued then the reported value
+includes moves buffered on the micro-controller, but does not include moves
+on the look-ahead queue. One may execute `M400` before evaluating a macro
+that reads `mcu_position` to fully flush the look-ahead and step generation
+code.
+
 ## stepper_enable
 
 The following information is available in the `stepper_enable` object (this
--- a/klippy/stepper.py
+++ b/klippy/stepper.py
@@ -250,7 +250,7 @@ def PrinterStepper(config, units_in_radians=False):
     mcu_stepper = MCU_stepper(name, step_pin_params, dir_pin_params,
                               rotation_dist, steps_per_rotation,
                               step_pulse_duration, units_in_radians)
-    if 'stepper' in name:
+    if not name.startswith('stepper'):
         printer.add_object(name, mcu_stepper)
     # Register with helper modules
     for mname in ['stepper_enable', 'force_move', 'motion_report']:
--- a/klippy/stepper.py
+++ b/klippy/stepper.py
@@ -228,6 +228,10 @@ class MCU_stepper:
         ffi_main, ffi_lib = chelper.get_ffi()
         a = axis.encode()
         return ffi_lib.itersolve_is_active_axis(self._stepper_kinematics, a)
+    def get_status(self, eventtime):
+        return {
+            'mcu_position': self.get_mcu_position(),
+         }
 
 # Helper code to build a stepper object from a config section
 def PrinterStepper(config, units_in_radians=False):
@@ -246,6 +250,8 @@ def PrinterStepper(config, units_in_radians=False):
     mcu_stepper = MCU_stepper(name, step_pin_params, dir_pin_params,
                               rotation_dist, steps_per_rotation,
                               step_pulse_duration, units_in_radians)
+    if 'stepper' in name:
+        printer.add_object(name, mcu_stepper)
     # Register with helper modules
     for mname in ['stepper_enable', 'force_move', 'motion_report']:
         m = printer.load_object(config, mname)


Fragmon:
i used it with a macro and tested it. i approve of inserting it

jschuh:
What about caching the details from the last `GET_POSITION` call into something like a `last_get_position` field in the `gcode_move` status dictionary? It's the same pattern that `QUERY_ENDSTOPS` uses (I assume for similar performance reasons).

Then the use pattern is just `G4 P0` followed by `GET_POSITION` and finally a call to the inner macro that evaluates the contents of `printer.gcode_move.last_get_position`.

pedrolamas:
I see considerable value in this proposal, but I tend to agree with @jschuh.

I don't think there is any need for real-time value changes from this in macros, so an on-demand only approach like `QUERY_ENDSTOPS` seems quite reasonable to me.

KevinOConnor:
Thanks.  Interesting.  I can understand the value in exporting this information.

Some high-level comments:
1. We haven't historically exported the `stepper_x` sections to the status reference.  But I guess there is no harm in doing that.  I'll give it some more thought.
2. I'm not sure what the `po_name` in the code is for - wouldn't we always want to use the config name for the status entry?
3. I'm not sure about reporting `mcu_position_offset` as it is an internal code detail.  It might be better to report  the "stepper position" (`get_commanded_position()`).  That does raise the question of whether or not we should report both the "mcu position" and the "stepper position", or if some synthetic value should be provided that relates the two.  I'm not sure.  It may be preferable to export values in "mm" (or "radians") instead of internal counters though.
4. If there is interest in the step distance then I think the external field name should be `step_distance`.
5. I'd be leery of documenting that `G4 P0` flushes the kinematics queue as we may change that behavior in the future.  Documenting `M400` (ie, `toolhead.wait_moves()`) seems safer to me.
6. This would increase bandwidth for api clients that subscribe to everything (as the new "mcu_position" is likely to change frequently during a print).  However, I guess that is okay as clients that don't want that bandwidth could just not subscribe to everything..

-Kevin

Piezoid:
>  2. I'm not sure what the `po_name` in the code is for - wouldn't we always want to use the config name for the status entry?

This is because the `MCU_stepper` for the extruder collides with the `extruder` printer object. The renaming is generic, but currently this is the only case where the status entry is named differently than the config section.
`'stepper_extruder' if name == 'extruder' else name` would be as effective.

>  3. I'm not sure about reporting `mcu_position_offset` as it is an internal code detail.  It might be better to report  the "stepper position" (`get_commanded_position()`).  

I wanted to avoid calling twice the same C function when using both `get_commanded_position()` and `get_mcu_position()`. Although, I can reorganize the code to prevent this.

> That does raise the question of whether or not we should report both the "mcu position" and the "stepper position", or if some synthetic value should be provided that relates the two.  I'm not sure.  It may be preferable to export values in "mm" (or "radians") instead of internal counters though.

For my specific use case, I aim to verify whether the step count discrepancy between two homing events is less than a full steps (`< microsteps`). However, the "stepper position" cannot be utilized for this purpose because it has the homing offset deducted from it.
One option could be using a synthetic value like `mcu_position * step_dist`, but I'm uncertain about the availability of the "full step distance" for comparison.
It's not entirely clear to me what could be other applications of this, this is why I attempted to expose everything. 

>  6. This would increase bandwidth for api clients that subscribe to everything (as the new "mcu_position" is likely to change frequently during a print).  However, I guess that is okay as clients that don't want that bandwidth could just not subscribe to everything..

The problem can be resolved by following the proposed approach of jschuh and pedrolamas, which involves using `GET_POSITION` to update cached values in the gcode_move status. However, this approach doesn't support reading the step position of manual and extruder steppers, and it also adds more messages to the console.
What is your opinion on this?

KevinOConnor:
> This is because the MCU_stepper for the extruder collides with the extruder printer object. The renaming is generic, but currently this is the only case where the status entry is named differently than the config section.
'stepper_extruder' if name == 'extruder' else name would be as effective.

I think this should be done differently.  I think it is too confusing to have new names invented in the "status" information.

As alternatives, the main ExtruderStepper object could call the `MCU_stepper.get_status()` it its own `get_status()` method; the extruder positions could not be exported; or perhaps some other module could export a dictionary of stepper positions.

> It's not entirely clear to me what could be other applications of this, this is why I attempted to expose everything.

For what it is worth, I'd prefer not to add code just because "someone may find it useful" as that approach becomes a long-term maintenance burden - we'll never know if we can change the code to facilitate future functionality because we don't know what the use case is for the existing code.

If the use case is only for "mcu_position" then perhaps we should only export that for now.

> The problem can be resolved by following the proposed approach of jschuh and pedrolamas, which involves using GET_POSITION to update cached values in the gcode_move status. However, this approach doesn't support reading the step position of manual and extruder steppers, and it also adds more messages to the console.
What is your opinion on this?

Seems a bit complex to me.  If API users subscribe to the stepper object then probably simpler to just send them the information they requested.

Cheers,
-Kevin

Zeanon:
I have found one problem with this:
the multi extruder test fails since you just check for
```if name != 'extruder'```
I think it should be
```if not name.startswith('extruder')``` to catch all extruders and not just the default one

Piezoid:
> `if not name.startswith('extruder')` to catch all extruders and not just the default one

Thanks, fixed! 
The first version handled this, but I missed it in the slimmed down version.

Fragmon:
Its working for me. I hope this pull will be accepted.

Surion79:
I do hope as well that this is being accepted :) it is not that big as previously implementedd afaiu :)

Fragmon:
Whats with this PR?

Fragmon:
What currently stands in the way of a merge?

SNX100:
Hello everyone. 
I'm also interested in knowing the chances of this being adopted. I'm working on a macro that depends on this essential feature. To make my macro available to the community, it would be great if they didn't have to work with a compromised Klipper file. Best regards.

Fragmon:
Why is this pull not accepted?

Surion79:
Could we get this please accepted? This would really help in automated speed calibration of printers using klipper.

SNX100:
is this pull forgotten? it would be very nice (and necessary for me & a lot more persons) to get this accepted.

whatsalinux:
Ditto...a google search just brought me here because I too  was trying to write a macro that needs this information.  Automation > Eyeball comparisons/hand jams

Piezoid:
@KevinOConnor sorry for the ping. Is there anything I can do to help this PR? I regularly get asked why this PR stalled.

Thanks

KevinOConnor:
Sorry for the delay.  The only reason for the delay is time constraints on my side.

Unless I'm missing something, this PR would break the `[extruder_stepper]` code - as it would result in both the PrinterStepper and ExtruderStepper classes trying to register the same name.

Separately, can the existing `[endstop_phase]` module be used to accomplish the goal here?  That module already exports the mcu_position after each home attempt.  https://www.klipper3d.org/Status_Reference.html#endstop_phase .  So, if one wants to detect if there were skipped steps, one could add a bare `[endstop_phase]` declaration to printer.cfg, and then inspect `printer.endstop_phase.last_home.stepper_x.mcu_position` after each home.  Of course, that functionality is different than what is proposed here, but I'd like to understand your requirements a little better.

-Kevin

PR #6194, opened 2023-05-06
Added socket permissions/script

SkyKnightSKS:
The problem being fixed here is the permissions on the socket file.
If moonraker is running as a different user additional manipulations are necessary to set permissions for to access the socket.

Two things here allow many different setups. The simplest is to use chmod to set certain permissions on the socket itself.
Using setgid on the parent directory you can automatically force the group on the socket and with chmod allow read/write access.

A second method that can be used is a script which can run a series of commands with or without sudo.

### DIFF
--- a/klippy/klippy.py
+++ b/klippy/klippy.py
@@ -305,6 +305,13 @@ def main():
                     help="input tty name (default is /tmp/printer)")
     opts.add_option("-a", "--api-server", dest="apiserver",
                     help="api server unix domain socket filename")
+    opts.add_option("--api-server_perm", dest="apiserver_perm",
+                    help="permissions for the api server unix domain socket")
+    opts.add_option("--api-server_script", dest="apiserver_script",
+                    help="script to execute on the socket  after creation")
+    opts.add_option("--api-server_script_sudo", dest="apiserver_script_sudo",
+                    default="False",
+                    help="use sudo for the api script")
     opts.add_option("-l", "--logfile", dest="logfile",
                     help="write log to file instead of stderr")
     opts.add_option("-v", action="store_true", dest="verbose",
@@ -321,7 +328,12 @@ def main():
         import_test()
     if len(args) != 1:
         opts.error("Incorrect number of arguments")
-    start_args = {'config_file': args[0], 'apiserver': options.apiserver,
+    start_args = {'config_file': args[0],
+                  'apiserver': options.apiserver,
+                  'apiserver_perm': options.apiserver_perm,
+                  'apiserver_script': options.apiserver_script,
+                  'apiserver_script_sudo': options.apiserver_script_sudo.lower()
+                                           in ['true', '1', 't', 'y', 'yes'],
                   'start_reason': 'startup'}
 
     debuglevel = logging.INFO
--- a/klippy/webhooks.py
+++ b/klippy/webhooks.py
@@ -5,6 +5,7 @@
 # This file may be distributed under the terms of the GNU GPLv3 license
 import logging, socket, os, sys, errno, json, collections
 import gcode
+import pwd, grp, subprocess
 
 REQUEST_LOG_SIZE = 20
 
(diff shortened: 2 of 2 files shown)


Sineos:
Is this change worth additional options?
I would guess that only very few users feel the need to do so. And if they feel this need they should have enough Linux admin knowledge to do so.

In any case also refer to: https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review
In particular no 2 and 3

SkyKnightSKS:
As for the target audience, this fix is related to a security enhancement. The idea is that this would be used as a common practice. Currently the common installation uses the same user for klipper and moonraker. This is against best practices both linux and security-wise.

To properly use the options you need to have the necessary linux admin knowledge. That is true. Typically a unix domain socket is being created the application creating it should also set the proper permissions on it.

All linux-workarounds to set these permissions are always running behind the fact when the socket was created. Having to monitor if the socket is there. Whereas it is normally the applications responsibility to do so.
If klipper decides to recreate the socket, it can use these options to set the permissions correctly as well.

Piezoid:
Can't this be achieved using [`systemd.socket`](https://www.freedesktop.org/software/systemd/man/systemd.socket.html)'s `SocketUser=` property?

SkyKnightSKS:
To my understanding that would be for socket-activated services. Which klipper is not as it is managing, creating and destroying the unix socket itself.

Sineos:
https://man7.org/linux/man-pages/man7/unix.7.html
>  Pathname socket ownership and permissions
>        In the Linux implementation, pathname sockets honor the
>        permissions of the directory they are in.  Creation of a new
>        socket fails if the process does not have write and search
>        (execute) permission on the directory in which the socket is
>        created.

According to this, it should be as simply as moving the socket out of tmp and using proper directory permissions. As such, it no longer matters that Klipper creates / destroys the sockets programmatically.
IMHO this is quite overdone but in the end it is @KevinOConnor's call.

Piezoid:
> To my understanding that would be for socket-activated services. Which klipper is not as it is managing, creating and destroying the unix socket itself.

You're absolutely right!
Adding socket activation support is actually pretty straightforward:
<details>

```diff
diff --git a/klippy/webhooks.py b/klippy/webhooks.py
index 9188a4f70..84c91243f 100644
--- a/klippy/webhooks.py
+++ b/klippy/webhooks.py
@@ -111,14 +111,21 @@ class ServerSocket:
         start_args = printer.get_start_args()
         server_address = start_args.get('apiserver')
         is_fileinput = (start_args.get('debuginput') is not None)
-        if not server_address or is_fileinput:
+        if os.environ.get('LISTEN_FDS') == '1':
+            # Valid systemd socket activation opens a single socket
+            self.sock = socket.fromfd(3, socket.AF_UNIX, socket.SOCK_STREAM)
+            logging.info('Started by socket activation on %r',
+                self.sock.getsockname())
+        elif server_address and not is_fileinput:
+            # Unix domain socket file managed by klipper
+            self._remove_socket_file(server_address)
+            self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+            self.sock.bind(server_address)
+            self.sock.listen(1)
+        else:
             # Do not enable server
             return
-        self._remove_socket_file(server_address)
-        self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         self.sock.setblocking(0)
-        self.sock.bind(server_address)
-        self.sock.listen(1)
         self.fd_handle = self.reactor.register_fd(
             self.sock.fileno(), self._handle_accept)
         printer.register_event_handler(
```

Example `klipper.socket` file:
```properties
[Socket]
ListenStream=/var/run/printers/ender3/klipper.socket
SocketMode=0660
SocketUser=ender3
SocketGroup=ender3

[Install]
WantedBy=sockets.target
```

We can ensure that systemd creates the socket by adding `Requires=klipper.socket` in `klipper.service`, then  `-a <uds_path>` can be removed from klippy's command line.
This example may not be your exact isolation scenario, but it gives a general sense of the idea.
</details>

SkyKnightSKS:
That is absolutely a nice and clean solution and overall a better approach.

I went the other route with the idea that klipper should be able to run without having a client connect on the socket.
But with some systemd settings you can achieve this as well.

And it has the added benefit that it does not introduce any new options.
Also no need to setup any sudo rights for klipper or ACL trickery and avoids setting up a socket proxy.

The code snippet you made does only support a single socket.
Not sure if this is something that needs to be considered. I assume not.

SkyKnightSKS:
If okay with you @Piezoid I opened a PR with your suggestion.

Arksine:
There is one issue with this from Moonraker's perspective.  Moonraker uses the socket's `PEERCRED` to identify Klipper's PID, this allows Moonraker to verify Klipper's unit name so it can manage the service.  Systemd socket activation breaks this.

PR #6195, opened 2023-05-07
Revert "atsamd: Add support for SAMC21"

Wulfsta:
This reverts commit 3b0729c94932f4df0298a7c992022f54a5454428.

It seems like there is an issue with the CAN Tx (that I must have done something to fix locally because it's an issue I noticed before I had Rx working, but I don't have the changes) on this chip. Issuing a revert in the meantime, will add this back once I figure out what happened here. Alternatively, I can work on figuring out what it is and push a fix once I do - probably not too much of an issue given the barrier to flashing the toolboard with this chip on it.


Wulfsta:
Leaving this open for now, but I've discovered the issue was that I had modified build flags to remove optimizations (`-O0`) when developing. With `-O2` on my machine I'm seeing errors after several CAN transmits...

PR #6198, opened 2023-05-12
atsamd: Fix compiler optimization and ADC bugs

Wulfsta:
The `O2` build flag was leading to broken CAN Tx code for the atsamc21.

### DIFF
--- a/src/atsamd/adc.c
+++ b/src/atsamd/adc.c
@@ -14,7 +14,9 @@ DECL_ENUMERATION("pin", "ADC_TEMPERATURE", ADC_TEMPERATURE_PIN);
 
 #if CONFIG_MACH_SAMC21
 
-#define ADC_INPUTCTRL_MUXNEG_GND 0x18
+DECL_CONSTANT_STR("RESERVE_PINS_adc", "PA3");
+
+#define ADC_INPUTCTRL_MUXNEG_GND ADC_INPUTCTRL_MUXNEG(0x18)
 
 #define SAMD51_ADC_SYNC(ADC, BIT) \
     while(ADC->SYNCBUSY.reg & ADC_SYNCBUSY_ ## BIT)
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  I don't think we should add the `ifdefs` as the delay is harmless on the other chips.

Separately, out of curiosity, instead of `readl(&txfifo->data[1])`, does a read of `CANx->TXBAR.reg` provide a sufficient delay?

-Kevin

Wulfsta:
I'm actually investigating other issues like this on this chip - I did all the manual testing on `O0` since my git tree was dirty and I hadn't realized, and am hitting other issues now. Today I found myself unable to run `QUERY_ENDSTOPS` on an  `O2` build, where it works fine on `O0`. Might need some help with this, because my experience dealing with compiler introduced errors at an embedded level is between extremely limited and nonexistent. Did you have any issues like these on `atsamd` chips in the past?

KevinOConnor:
> Did you have any issues like these on atsamd chips in the past?

Not that I can recall.  I don't typically develop with `-O0` though, so it's not likely I'd write code that is dependent on it.

Otherwise, what is the current status of this PR?

-Kevin

Wulfsta:
Still in progress; been meaning to make an update but haven’t had time. I’m observing an issue that I don’t think is samc21, and there was some chatter offline that seemed to hint that this is the case. I’ll follow up.

KevinOConnor:
Thanks.  I saw there were some recent updates on this PR.  Is it thought that it is now ready for merging?

-Kevin

Wulfsta:
I have no objections to merging. I sent some hardware to @eamaclean to have a look at the remaining bugs that I haven’t been able to find, and it would be nice if this fix made it into the code before then.

KevinOConnor:
Thanks.

-Kevin

PR #6199, opened 2023-05-12
configfile: adds meta options

pedrolamas:
This is a follow up on https://klipper.discourse.group/t/metadata-in-config-sections/8344

My intention here is to allow users to add `meta_` options to ANY section on the config with whatever value they would like.

These options will be completely ignored by Klipper and will only be available on the `configfile` status object, under `config.<section>.meta_<name>` property.

## Config example:

```ini
[mcu]
serial: /tmp/pseudoserial
restart_method: Arduino
meta_test: 1

[virtual_pins]
meta_other: whatever
```

## Moonraker output example:

image

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1764,6 +1764,16 @@ pins:
 #   provided.
 ```
 
+### "meta_" prefixed options
+
+```
+[<section>]
+#meta_<name>:
+#   One may specify any number of options with a "meta_" prefix.
+#   These options are not used anywhere in Klipper, but will be available
+#   on the `configfile.config.<section>.meta_<name>` status object
+```
+
 ## Bed probing hardware
 
 ### [probe]
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -300,7 +300,8 @@ class PrinterConfig:
                             % (section,))
             for option in fileconfig.options(section_name):
                 option = option.lower()
-                if (section, option) not in access_tracking:
+                if ((section, option) not in access_tracking and
+                        not option.startswith('meta_')):
                     raise error("Option '%s' is not valid in section '%s'"
                                 % (option, section))
         # Setup get_status()


ayufan:
@KevinOConnor @pedrolamas 

I would see the following use-case:

- Allow to predefine how macros are categorised and whether they are visible and when
- It makes things in UI more organised by grouping things together by the function
- Since those are pre-defines user could still overwrite them in UI if it wishes

Example usage pattern:

```
# ERCF specific GCodes
[gcode_macro ERCF_RESET]
meta_macro_category: ERCF

[gcode_macro T0]
meta_category: ERCF
meta_disabled_while_printing: true

# this is internal Gcode that we don't want to show in UI
[gcode_macro G32]
meta_visible: false

# other GCodes that help with printer calibaration/testing
[gcode_macro TEST_SPEED]
meta_category: Calibration
meta_disabled_while_printing: true
```

pedrolamas:
The specific use-case for macros management is a great example!

As this change allows metadata in any element, I would even take it even up a notch: one could use it to add names and descriptions to elements so that would show as captions and tooltips in the frontend, or add a color that would then be used in the frontend as foreground/background, etc.

ayufan:
I believe this `@github-actions` bot is anti contributor friendly. I don't think this issue become inactive, but it did not receive feedback. Per that I think it is harsh to close the issue.

cc @KevinOConnor

KevinOConnor:
Thanks for the feedback.  The purpose of the bot (at least in my mind) is to give an update to the contributor of the status of the PR - in particular that it is unlikely to be merged in its current state.  It is not the intent to discourage the contributor.  I certainly don't want to discourage contributors, but I also don't want to "leave them hanging" for years.

I feel there is a fundamental problem - I receive many more PRs than I can review.  So, I'm open to alternatives but I an alternative must really address that fundamental issue.

As for this PR, I didn't feel there was a consensus on it (there was also a discussion on Klipper Discourse).  So, I think it needs a consensus before merging.

Cheers,
-Kevin

ayufan:
> As for this PR, I didn't feel there was a consensus on it (there was also a discussion on Klipper Discourse). So, I think it needs a consensus before merging.

I'm unsure that this statement is fair here.

For example it was asked to provide concrete example, which was given: https://github.com/Klipper3d/klipper/pull/6199#issuecomment-1564009540. No one did say no to the given example, and also looking at prior discussion no one disagreed that such function would not be beneficial.

PR #6201, opened 2023-05-13
Add gcode_button to the Status Reference docs

rootiest:
The Status Reference documentation is missing a section for gcode_button.

This PR just adds that section.


KevinOConnor:
Thanks.

-Kevin

PR #6202, opened 2023-05-14
Pin aliases for Protoneer's CNC-shield (v3.0)

naikymen:
Pin aliases for the Arduino-UNO/CNC-shield (v3.0) by Protoneer.

I believe that this will be generally useful, and also specifically useful to those coming from GRBL, or just wanting to try out Klipper on their old board.


KevinOConnor:
Also see: https://www.klipper3d.org/Example_Configs.html

-Kevin

PR #6203, opened 2023-05-14
Added socket activation support

SkyKnightSKS:
This PR allows the usage of SystemD sockets.
The benefit would be that you can set permissions / user / group of the socket outside of klipper.
Allowing to isolate klipper more easily and moonraker to be run from different users.

https://github.com/Klipper3d/klipper/pull/6194

Committing the change suggested by @Piezoid
See the PR above.

### DIFF
--- a/klippy/webhooks.py
+++ b/klippy/webhooks.py
@@ -366,9 +366,14 @@ class WebHooks:
         state_message, state = self.printer.get_state_message()
         src_path = os.path.dirname(__file__)
         klipper_path = os.path.normpath(os.path.join(src_path, ".."))
-        response = {'state': state, 'state_message': state_message,
+        response = {'state': state,
+                    'state_message': state_message,
                     'hostname': socket.gethostname(),
-                    'klipper_path': klipper_path, 'python_path': sys.executable}
+                    'klipper_path': klipper_path,
+                    'python_path': sys.executable,
+                    'process_id': os.getpid(),
+                    'user_id': os.getuid(),
+                    'group_id': os.getgid()}
         start_args = self.printer.get_start_args()
         for sa in ['log_file', 'config_file', 'software_version', 'cpu_info']:
             response[sa] = start_args.get(sa)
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  As high-level feedback, it's not clear to me what this PR is intended to solve.  Wouldn't it be simpler to place the socket in a directory that had the desired access permissions?

The check for LISTEN_FDS seems a "hack" to me - I'm guessing it is based on info from `man sd_listen_fds`, but the steps described there are much more involved then just checking LISTEN_FDS.

-Kevin

SkyKnightSKS:
Thanks for the feedback.

As for the why, consider the scenario where klipper is running on a separate user & group than moonraker.
Creating a directory and putting the socket in there, even if moonraker has the necessary permissions on this will not allow moonraker to access the socket. Using extended acl on the filesystem you are able to make the group of the socket moonraker, but you cannot make the default permission writable for the group as this is a socket. It requires special behaviour compared to regular files.

However it is true, there's a whole gamma of workarounds and tricks that can be done.

However the simplest and cleanest solution would be to use a systemd-socket. For example:
`[Unit]
BindsTo=klipper.service
Before=klipper.service

[Socket]
ListenStream=/tmp/klippy.sock
SocketUser=moonraker
SocketGroup=moonraker
SocketMode=0600

[Install]
WantedBy=sockets.target`
This particular socket file in systemd would allow you to create and set any uid/gid mode you like for the socket.

All in all as far as I know, this is the cleanest and simplest solution and only requires minor changes to klipper.

I have added the additional checks which are typically implemented for socket activation. Which is to compare the LISTEN_PID to the running pid of klipper.
It is implemented this way as not to create a dependancy on the systemd python module. Which is not necessary for such a small thing.

SkyKnightSKS:
Indeed I've tested it, it still works when using provider=cli.
I can expand the info endpoint to include this information there.

KevinOConnor:
Thanks.  I'm not really sure this is a good fit for the master Klipper repo - or at least, I'm not able to meaningfully review this PR.  It's not clear to me how a typical user would use this or what real world tasks they would accomplish with it.

-Kevin

PR #6205, opened 2023-05-16
tmc2240: register as temperature sensor

pedrolamas:
Following up on #6145, this registers the `tmc2240` as a temperature sensor so that Moonraker can cache the temperature readings automatically.

### DIFF
--- a/klippy/extras/tmc2240.py
+++ b/klippy/extras/tmc2240.py
@@ -396,6 +396,9 @@ class TMC2240:
         set_config_field(config, "tpowerdown", 10)
         #   SG4_THRS
         set_config_field(config, "sg4_angle_offset", 1)
+        # Register temperature sensor
+        pheaters = config.get_printer().load_object(config, 'heaters')
+        pheaters.register_sensor(config, self)
 
 def load_config_prefix(config):
     return TMC2240(config)


pedrolamas:
@Arksine the above PR is causing issues on existing Moonraker code!

Here's a small bit of the log showing the problem:

```log
2023-05-18 12:41:11,058 [data_store.py:_init_sensors()] - Configuring available sensors: ['tmc2240 stepper_x', 'temperature_sensor mcu', 'temperature_sensor cb1']
2023-05-18 12:41:11,059 [eventloop.py:_wrapper()] - Error Running Callback
  File "/home/biqu/moonraker/moonraker/eventloop.py", line 73, in _wrapper
    await ret
  File "/home/biqu/moonraker/moonraker/server.py", line 334, in _process_event
    await asyncio.gather(*coroutines)
  File "/home/biqu/moonraker/moonraker/components/data_store.py", line 106, in _init_sensors
    self._set_current_temps(status)
  File "/home/biqu/moonraker/moonraker/components/data_store.py", line 119, in _set_current_temps
    round(data[sensor].get('temperature', last_val[0]), 2),
TypeError: type NoneType doesn't define __round__ method
```

TMC2240 will only report temperatures if the stepper is enabled (which does make sense to me) and otherwise it will be null, so the above code will throw an error as it tries to apply round to null.

I see two solutions here: either allow for null temperatures (which I think it is preferable) or ensure we do a null check here and just set to 0 in that case.

For the first option, I did the following test and it seems to work fine:

```diff
  for sensor in self.temperature_store:
     if sensor in data:
       last_val = self.last_temps[sensor]
+      temperature = data[sensor].get('temperature', last_val[0])
       self.last_temps[sensor] = (
+        round(data[sensor].get('temperature', last_val[0]), 2),
-        round(temperature, 2) if temperature else None,
         data[sensor].get('target', last_val[1]),
         data[sensor].get('power', last_val[2]),
         data[sensor].get('speed', last_val[3]))
```

Sineos:
Would there be a difference to the host temperature sensors? Do they allow for any methods?

Arksine:
The Host and MCU sensors implement all of the required temperature sensor methods.  I'm not sure it makes sense to extend the TMC2240 considering it only reports temperature when the steppers are enabled.

I lean toward reporting a third category, such as `available_monitors`, in the `get_status()` method of `heaters.py`.  We could consider a `monitor` a device that reports temperature but can't be used as a control for other peripherals.

pedrolamas:
Thank you for your comments, Kevin. I've now taken a closer look at the code around `heaters.register_sensor()` method and think I have a better understanding of your points.

> I lean toward reporting a third category, such as `available_monitors`, in the `get_status()` method of `heaters.py`. We could consider a `monitor` a device that reports temperature but can't be used as a control for other peripherals.

Personally, I like that!

Another option I was just now looking at was creating a new temperature sensor called `PrinterTemperatureDriver` in "temperature_driver.py", which would take a config like this:

```ini
[temperature_sensor stepper_x]
sensor_type: temperature_driver
sensor_mcu: stepper_x
...
```

Main advantage is that we can then use the temperature from the drivers directly on Klipper!

pedrolamas:
I've opened a new PR #6210 with my new proposed Temperature Sensor `temperature_driver` solution, so am closing this now.

PR #6206, opened 2023-05-17
stm32: add PB5/PB6 as CAN pins - required for BTT U2C v2.x

meteyou:
This pull request adds support for the CAN interface PB5/PB6 on the STM32G0B1. This MCU with these pins for CAN is used in the BigTreeTech U2C v2.x.

Here is a guide to use the U2C v2.x with Klipper CAN-bridge: https://docs.meteyou.wtf/btt-u2c-v2.x/klipper-usb-to-can/


KevinOConnor:
Thanks.

-Kevin

PR #6209, opened 2023-05-19
Add USART6 FOR stm32f401vxx

KimRay2015:
1.Add stm32f401ve usart6
2.modified for usart6 gpio_function


KevinOConnor:
Thanks.  What board is this needed on?  In general it seems fine but commits need to have a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

PR #6210, opened 2023-05-19
temperature_driver: adds driver temperature sensor

pedrolamas:
(This is a follow up on #6145 and discussion on PR #6205)

The following PR adds a new Temperature Sensor called `temperature_driver` that allows monitoring and reporting of the drivers temperature.

Note: right now, only TMC2240 supports temperature reporting, but I've added no validation on the type of driver, so users will still be able to use any driver, however temperature reported will always be 0 for non-TMC2240!

With this change there is no change required either on Moonraker, Fluidd, Mainsail or any other UI, as it will be up to the user to add the temperature sensor.

Here's a sample config:

```ini
[tmc2240 stepper_x]
cs_pin: PD9
spi_bus: spi2
run_current: 0.800
stealthchop_threshold: 999999

[temperature_sensor stepper_x]
sensor_type: temperature_driver
sensor_driver: tmc2240 stepper_x
```

I took the above configuration and put it on a test system with a TMC2240 and this is the output in Fluidd (code changes only the ones on this Klipper PR and nothing more)

image

Signed-off-by: Pedro Lamas pedrolamas@gmail.com

### DIFF
--- a/klippy/extras/temperature_driver.py
+++ b/klippy/extras/temperature_driver.py
@@ -1,0 +1,58 @@
+# Support for micro-controller chip based temperature sensors
+#
+# Copyright (C) 2020  Kevin O'Connor <kevin@koconnor.net>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+DRIVER_REPORT_TIME = 1.0
+
+class PrinterTemperatureDriver:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+
+        driver_name = config.get('sensor_driver')
+        self.driver = self.printer.lookup_object(driver_name)
+
+        self.temp = self.min_temp = self.max_temp = 0.0
+
+        self.reactor = self.printer.get_reactor()
+        self.sample_timer = self.reactor.register_timer(
+            self._sample_driver_temperature)
+
+        self.printer.register_event_handler("klippy:connect",
+                                            self.handle_connect)
+
+    def handle_connect(self):
+        self.reactor.update_timer(self.sample_timer, self.reactor.NOW)
+
+    def setup_callback(self, temperature_callback):
+        self.temperature_callback = temperature_callback
+
+    def setup_minmax(self, min_temp, max_temp):
+        self.min_temp = min_temp
+        self.max_temp = max_temp
+
+    def get_report_time_delta(self):
+        return DRIVER_REPORT_TIME
+
+    def _sample_driver_temperature(self, eventtime):
+        self.temp = self.driver.get_temperature()
+
+        if self.temp is not None:
+            if self.temp < self.min_temp or self.temp > self.max_temp:
+                self.printer.invoke_shutdown(
+                    "DRIVER temperature %0.1f outside range of %0.1f:%.01f"
+                    % (self.temp, self.min_temp, self.max_temp))
+
+        measured_time = self.reactor.monotonic()
+
+        if self.temp is not None:
+            mcu = self.driver.get_mcu()
+            self.temperature_callback(
+                mcu.estimated_print_time(measured_time),
+                self.temp)
+
+        return measured_time + DRIVER_REPORT_TIME
+
+def load_config(config):
+    pheaters = config.get_printer().load_object(config, "heaters")
+    pheaters.add_sensor_factory("temperature_driver", PrinterTemperatureDriver)
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2506,6 +2506,17 @@ sensor_type: temperature_mcu
 #   micro-controller specification.
 ```
 
+### Builtin stepper driver temperature sensor
+
+The TMC2240 stepper driver contain an internal temperature sensor. One can
+use the "temperature_driver" sensor to monitor this temperatures.
+
+```
+sensor_type: temperature_driver
+sensor_driver:
+#   The stepper driver to read from.
+```
+
 ### Host temperature sensor
 
 Temperature from the machine (eg Raspberry Pi) running the host software.
--- a/klippy/extras/temperature_sensors.cfg
+++ b/klippy/extras/temperature_sensors.cfg
@@ -33,6 +33,9 @@
 # Load "temperature_mcu" sensor
 [temperature_mcu]
 
+# Load "temperature_driver" sensor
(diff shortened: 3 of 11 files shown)


klippy/extras/temperature_driver.py
```diff
@@ -0,0 +1,63 @@
...
+                    % (self.temp, self.min_temp, self.max_temp))
+        else:
+            self.temp = 0.
+
+        mcu = self.driver.mcu_tmc.tmc_spi.spi.mcu
+        measured_time = self.reactor.monotonic()
+        self.temperature_callback(
+            mcu.estimated_print_time(measured_time),
+            self.temp)
+        return measured_time + DRIVER_REPORT_TIME
```
pedrolamas, line 54: @KevinOConnor though the code actually works as expected, I'm not entirely sure of this whole block and would really appreciate your input, especially on how best to retrieve the `mcu` here (if that is even necessary)

Sineos:
Would this allow for using the temperature to control a stepper driver fan? IMO this would be a nice use case instead unconditionally turning it on via `controller_fan` or just reporting the temperature.

klippy/extras/temperature_driver.py
pedrolamas: I've been looking closely at the MCU bit and will refactor this today to use a chain of `get_mcu()` methods to avoid accessing directly

klippy/extras/temperature_driver.py
```diff
@@ -0,0 +1,58 @@
...
+                    % (self.temp, self.min_temp, self.max_temp))
+        else:
+            self.temp = 0.0
+
+        mcu = self.driver.get_mcu()
+        measured_time = self.reactor.monotonic()
+        self.temperature_callback(
+            mcu.estimated_print_time(measured_time),
+            self.temp)
+        return measured_time + DRIVER_REPORT_TIME
```
pedrolamas, line 54: I'm not entirely sure of this whole block, so would really apreciate if anyone can validate this is correct!

KevinOConnor:
Thanks.  As high-level feedback, I'm leery of introducing a "dummy temperature" to indicate "the temperature is not known".  That is, I'm leery of using 0.0 when the driver temperature is not known.  (In particular, when the dummy value itself could plausibly be valid.)  I fear it could lead to confusion, weird looking graphs, and could propagate further downstream hacks to account for it.

-Kevin

pedrolamas:
Thanks Kevin, that's fair and makes complete sense.

I guess the alternative is returning `None` when the driver is disabled, but I believe Moonraker will then require changes to handle that (there's a `round` operation without any null check on it) - and I think same goes for Fluidd and Mainsail.

I will make the changes and test, and report back my findings.

pedrolamas:
Following up on the last comments, I've changed the code so that:
- `PrinterTemperatureDriver.temp` stores whatever the driver temperature read returns (so either a float or None)
- the callback is only called when we actually have a temperature read from the driver

Code performs as expected and looking at all the places in Klipper where `temperature_driver` can be used as a sensor, it seems ok that the callback is only called for non-None values.

Arksine:
I think the primary issue is that the TMC drivers don't consistently report temperature, thus they may not be a good candidate for a complete `temperature_sensor` implementation. I touched on this in the last PR which was why I recommended registering them as a monitor. 

The challenge is determining the appropriate course of action when the driver is disabled.  When the sensor is being relied upon by a peripheral this can introduce undesirable behavior, such as a continuously running fan.

pedrolamas:
That's a very good point, and I honestly didn't consider that possibility...

I admit I'm a bit out of ideas on how best to proceed on this PR, but @Arksine `available_monitors` proposal is starting to make a lot more sense as a way forward.

brotherdust:
@pedrolamas 
The tmc2240 datasheet indicates that the driver also has a way to estimate the motor temperature (see bottom of pg. 65). Does your PR expose that function? I’m keen to use it, so if not or if you don’t plan to, I’ll contribute it after your change is accepted.

Thanks!

pedrolamas:
Hi @brotherdust, no this PR is just to expose the existing temperature read as a temperature sensor in Klipper, although it has some issues as reported above.

Though out of scope for this PR, the estimation does sound very interesting! I'll keep an eye out for a contribution PR from you for it!

brotherdust:
@pedrolamas I think I will wait until the concerns raised above are addressed before submitting a PR (but will start working on it privately soon because I have some HOT steppers!)

eldeeb91:
can I somehow try this feature before it is rolled out?

eldeeb91:
I tried the configuration but I got the following error:

**Unknown temperature sensor 'temperature_driver'**

my config:

[stepper_x]
step_pin: STEP_1
dir_pin: !DIR_1
enable_pin: !EN_1
endstop_pin: !EndStop_1
microsteps: 64
rotation_distance: 40
# position_endstop: -12
position_endstop: -7
position_max: 300
# position_min: -12
position_min: -7
homing_speed: 50

[tmc2240 stepper_x]
cs_pin: UART_1
spi_software_sclk_pin: SCLK
spi_software_mosi_pin: MOSI
spi_software_miso_pin: MISO
interpolate = True
run_current: 0.8
# sense_resistor: 0.110
stealthchop_threshold: 500
# driver_IHOLDDELAY: 8
# driver_TPOWERDOWN: 20
# driver_TBL: 2
# driver_TOFF: 3
# driver_HEND: 0
# driver_HSTRT: 5
# driver_PWM_AUTOGRAD: True
# driver_PWM_AUTOSCALE: True
# driver_PWM_LIM: 12
# driver_PWM_REG: 8
# driver_PWM_FREQ: 1
# driver_PWM_GRAD: 14
# driver_PWM_OFS: 36
# driver_SGTHRS: 0

[temperature_sensor tmc2240_stepper_x]
sensor_type: temperature_driver
sensor_driver: tmc2240 stepper_x
# min_temp: 10
# max_temp: 100

pedrolamas:
@eldeeb91 this PR has not been merged, so the functionality here described is NOT available.

Having said that, #6292 was indeed merged, and together with changes that were also done in Moonraker, we have just released Fluidd 1.25 that shows the temperature of TMC2240 (screenshots here: https://github.com/fluidd-core/fluidd/pull/1133)

PR #6211, opened 2023-05-19
resonance_tester: Support CHIPS= parameter in SHAPER_CALIBRATE command

dmbutyugin:


dmbutyugin:
This may be useful for idex configuration and similar where a user may wish to run a calibration test with different accelerometer chips for different toolheads.

KevinOConnor:
Thanks.

-Kevin

PR #6212, opened 2023-05-19
resonance_tester: Apply input shaper params on SHAPER_CALIBRATE

dmbutyugin:

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,9 @@ All dates in this document are approximate.
 
 ## Changes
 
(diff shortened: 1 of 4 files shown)


dmbutyugin:
While `SHAPER_CALIBRATE` command is not meant to be used before every print in a manner similar to `BED_MESH_CALIBRATE`, it is still worthwhile to apply the suggested input shaper parameters right away so that the user could test them before saving. It might also be beneficial for idex configurations where it is not as straightforward to save input shaper configuration as just executing `SAVE_CONFIG` command, so a user may want to test the input shaper parameters before saving them.

KevinOConnor:
Thanks.

-Kevin

PR #6213, opened 2023-05-19
heater_fan: added hysteresis 

chestwood96:
I implemented some rudimentary hysteresis for heater_fan because the oscillations on my v0 got a bit annoying.

Tested on my enclosed voron v0.1 for the last couple weeks.

As far as I understand the code it should not change existing behaviorism unless enabled.

Signed-off-by: Adrian Joachim [adi.joachim12@gmail.com](mailto:adi.joachim12@gmail.com)

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2621,6 +2621,10 @@ a shutdown_speed equal to max_power.
 #heater_temp: 50.0
 #   A temperature (in Celsius) that the heater must drop below before
 #   the fan is disabled. The default is 50 Celsius.
+#heater_temp_off_offset: 0
+#   How many degrees (Celsius/Kelvin) the heater temperature has to 
+#   drop below "heater_temp" before the fan turns off 
+#   (to prevent a fan oscilating on and off). The default is 0K.
 #fan_speed: 1.0
 #   The fan speed (expressed as a value from 0.0 to 1.0) that the fan
 #   will be set to when its associated heater is enabled. The default
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2622,8 +2622,8 @@ a shutdown_speed equal to max_power.
 #   A temperature (in Celsius) that the heater must drop below before
 #   the fan is disabled. The default is 50 Celsius.
 #heater_temp_off_offset: 0
-#   How many degrees (Celsius/Kelvin) the heater temperature has to 
-#   drop below "heater_temp" before the fan turns off 
+#   How many degrees (Celsius/Kelvin) the heater temperature has to
+#   drop below "heater_temp" before the fan turns off
 #   (to prevent a fan oscilating on and off). The default is 0K.
 #fan_speed: 1.0
 #   The fan speed (expressed as a value from 0.0 to 1.0) that the fan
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I think it is fine to add hysteresis support to heater_fan.  However, I'm not sure it is worthwhile to add a new config setting for it, as I can't see where anyone would not want to use hysteresis or that anyone would need to supply a custom hysteresis.

Separately, the indentation of the code change is difficult to understand.  I think the code should probably be broken up into multiple if statements or otherwise clarified.

Cheers,
-Kevin

chestwood96:
> Thanks. As high-level feedback, I think it is fine to add hysteresis support to heater_fan. However, I'm not sure it is worthwhile to add a new config setting for it, as I can't see where anyone would not want to use hysteresis or that anyone would need to supply a custom hysteresis.

I didn't want it to change exiting behavior in case it causes side effects. Also not entirely sure what a good fixed value would be, 5C works on my v0 but that's a sample size of 1.

> Separately, the indentation of the code change is difficult to understand. I think the code should probably be broken up into multiple if statements or otherwise clarified.

Tell me about it, the 80 character limit kicked my ass there, looked a lot more readable before. Breaking it down into 2 if statements would yield a similarly ugly squished thing but I can do it if you prefer it.

KevinOConnor:
I would have thought a hysteresis of 1 degree would be fine.  It's just to avoid noise on the thermistor right?

-Kevin

chestwood96:
Thermistor noise definitely is part of it but in my case it looks like where is some residual heat in other parts of the block that leak into the thermistor after the fan stops, I do get a slow rise of 2-3 degrees over a few minutes after the fan turns off.

KevinOConnor:
Okay, thanks.  That is interesting, and I've not seen that report before.  All of my extruders will cool at a lower rate when the fan is turned off, but do not go up in temperature.

-Kevin

chestwood96:
I would still like to get this merged, can this be reopened or do I have to make anew request?

chestwood96:
@KevinOConnor Sorry to bother you but what do you want me to change to get this merged?

chestwood96:
@KevinOConnor I still would like to get this merged, please let me know what I can do there.

KevinOConnor:
Thanks.  Alas, I haven't been able to get time to look at this further.

One high-level comment I have is that I wonder if a more generic fan speed system may be more appropriate for your (a bit uncommon) case.  In previous PRs I've floated the idea of extending the "LED template" system to work with fans and PWM pins.  (That system allows one to configure a "formula" based on exported "printer" settings - the result of the "formula" is automatically recalculated and applied.)  That may be a more generic way of controlling a fan.

I'll be travelling for a couple of weeks, but will hopefully be able to look further at this after I return.

Cheers,
-Kevin

chestwood96:
@KevinOConnor Can we keep this open until the alternative implementation becomes a thing?

chestwood96:
@KevinOConnor Once again, can we keep this open until the alternative is merged/you tell me what to change to get it merged?

chestwood96:
@KevinOConnor keep open please

chestwood96:
@KevinOConnor keep open please

dewi-ny-je:
> #6695 Template support is merged.
> 
> I think this PR can be safely closed now.
> 
> Thanks.

how to use the template to achieve hysteresis?

Sineos:
> how to use the template to achieve hysteresis?

this is not the right place for such a discussion. Please use the contact options defined in https://github.com/Klipper3d/klipper/blob/master/docs/Contact.md

PR #6216, opened 2023-05-21
Fix heater with low sample rate temperature sensor, specifically BME680

01halibut:
Two separate issues fixed:
* `get_report_time_delta` returns the wrong thing for BME680 sensors. It returns the time for BME280, but the BME680 callback is set to be longer.
* In `heater.py`, for any temp sensor with a report cycle longer than `MAX_HEAT_TIME * 0.25` seconds, `next_pwm_time` can exceed `MAX_HEAT_TIME`. This means even if a sensor report is received within `MAX_HEAT_TIME`, it may be ignored by the PWM value update suppression. This looks like it's a bug/typo/oversight in how `next_pwm_time` is being set, it's unclear why a max delay from now of `0.75 * MAX_HEAT_TIME + self.pwm_delay` would make sense, but `0.75 * MAX_HEAT_TIME - self.pwm_delay` makes sense as preventing time to the next report from exceeding `MAX_HEAT_TIME`, plus some margin.
The second issue is the likely cause of #5189

### DIFF
--- a/klippy/extras/bme280.py
+++ b/klippy/extras/bme280.py
@@ -7,6 +7,7 @@ import logging
 from . import bus
 
 REPORT_TIME = .8
+REPORT_TIME_BME680 = .8 * 4
 BME280_CHIP_ADDR = 0x76
 BME280_REGS = {
     'RESET': 0xE0, 'CTRL_HUM': 0xF2,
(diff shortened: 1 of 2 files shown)


01halibut:
Reopening with a typo fix and in a non-main branch.

PR #6218, opened 2023-05-21
Led command improvements

Zeanon:
Changed the handling of INDEX in the SET_LED and SET_LED_TEMPLATE command so you can use INDEX=a,b,c to set multiple LEDs at the same time and also use INDEX=a-b to set all LEDs between a and b and also mix and match those two


Zeanon:
Forgot the signed off

PR #6221, opened 2023-05-21
bme280:  assign report time delta on chip discovery

Arksine:
This fixes a bug identified in the following discourse thread:

https://klipper.discourse.group/t/bug-bme680-get-report-time-delta-incorrect/8522/2

I don't have a BME680 to test with, however I don't believe there should be any issues as this is a relatively simple change.

### DIFF
--- a/klippy/extras/bme280.py
+++ b/klippy/extras/bme280.py
@@ -6,7 +6,7 @@
 import logging
 from . import bus
 
-REPORT_TIME = .8
+BME280_REPORT_TIME = .8
 BME280_CHIP_ADDR = 0x76
 BME280_REGS = {
     'RESET': 0xE0, 'CTRL_HUM': 0xF2,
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Maybe I'm missing something, but wouldn't it be simpler to report the bme680 data at the same rate as a bme280?

-Kevin

1am3r:
I think it was to calm down the gas sensor, but a quick test with the 0.8 report time from the bme280 seems to work. Could be i didn't really understand the values of the gas senor back then, as the 0.8 report time seems to give more realistic values for me now. 
I think we should just use the same value for bme280 and bme680.

Arksine:
Thanks, I'll go ahead and close this.  Changing the report time seems to be the best way to resolve the issue, so I'll go ahead and close this.  Kevin, would you prefer me to PR the change or make it directly?

KevinOConnor:
Thanks.  I'd prefer if someone could open a PR with the change, as I don't have either the bme280 or bme680 to run even basic tests.

-Kevin

PR #6189, opened 2023-04-29
docs: Fix GitHub case

MrBrain295:

PR #6223, opened 2023-05-21
Fix GitHub case

MrBrain295:
Fix up version of #6189.


dingyifei (approved):
Changes LGTM, but you might want to add signed off lines.

MrBrain295:
I signed off on the second one so when it is squashed it will be signed.

KevinOConnor:
Alas, I still don't see a signed-off-by line.  The signed-off-by line is a text statement that indicates you agree to the terms at https://github.com/Klipper3d/klipper/blob/master/docs/developer-certificate-of-origin .  See the documentation on it at https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

MrBrain295:
Oh, I thought I had to sign off the commit.

PR #6224, opened 2023-05-22
Led command improvements

Zeanon:
Changed the handling of INDEX in the SET_LED and SET_LED_TEMPLATE command so you can use INDEX=a,b,c to set multiple LEDs at the same time and also use INDEX=a-b to set all LEDs between a and b and also mix and match those two
If you put |n behind a range, it will only update every n'th LED in that range(e.g. INDEX=1-5|2 would update LED1, LED3 and LED5)


KevinOConnor:
Thanks.  As high-level feedback, I'm not sure this is a necessary change for the Klipper main branch.  It does make the command a little easier to type in by hand, but I suspect most users don't type in these commands by hand anyway.  On the flip side, I fear it increases the documentation and code complexity.

-Kevin

Zeanon:
I maybe added a bit much funktionality
But considering there are people with over 50 neopixels on their printers, it's way easier to set up templates on that many LEDs with the new system than with the old.
To be fair, my main goal was to be able to set a template to multiple defined LEDs and then I kind of got carried away with trying to perfect the command with as much functionality as possible

PR #6225, opened 2023-05-22
Max x, y velocity

your-friend-alice:
My Y axis is a bit clunky, I'd like to limit its velocity without holding back the X axis. This is probably common for bed slingers, but also in general there's often more velocity headroom with diagonal moves than horizontal ones, since a 45 degree move will only have each axis moving at `1/sqrt(2)` the speed.

Side note: i don't have experience with coreXY but I could see something similar being more subtly beneficial; horizontal moves have reduced stress on the motors, and diagonal moves have reduced stress on the rails, so it's possible an experienced user could characterize this and take advantage of this if there's more headroom in one of those two things.


jakep82:
This is a dup of #4661.

your-friend-alice:
oh yes good catch. any advice on getting this farther? I'd be happy to add some caveats to the documentation; the purpose of this feature isn't quality improvement, it's for eking out some extra speed from non-optimal machines. right now I'm using this to speed-print some storage bins on my Prusa Mini at 200mm/s, which would otherwise be physically impossible as the Y axis doesn't like going over 160. I'd be happy to direct this effort at the other PR if preferred, although at a glance I think this one is slightly more readable

your-friend-alice:
I've added a warning to the docs about possible quality impact, I also forgot to sign off on some commits so those have been remade

dmbutyugin:
If I read correctly, the main feedback on that other PR was to test this change on a variety of printers and demonstrate that the change offers noticeable improvements in print quality or allows to maintain the same level of quality while printing faster. Naturally, you don't have to run the tests on various different printers yourself, since a single person is not likely to have a park of machines. However, you may want to promote your changes (e.g. on Klipper discourse and/or discord) so that more people could test them and share their feedback. Ultimately, such a change isn't difficult to implement, but it requires testing to determine whether it makes sense or not.

A simple test that could be telling is printing a cylinder (in vase mode, or with a few wall, maybe with some light infill - depending on the cooling capabilities of the printer and the filament in use) at the speed and acceleration maximum for two axes and showing its quality. Then maybe benchmarking a few prints, i.e. by setting the velocity and acceleration to a minimum of two axes and configuring them separate for x and y, comparing the print times and showing that the quality is the same. Something along those lines, but other ideas are also welcome.

your-friend-alice:
that makes sense. i'm new to this community so one thing i'm missing is: is there a policy of features needing to not have a negative impact on print quality? or is it a documentation policy, like "good or bad, its quality impact must be well characterized"? or is this more of a general preference than a specific policy?

Sineos:
See https://www.klipper3d.org/CONTRIBUTING.html

dmbutyugin:
@your-friend-alice 

> is there a policy of features needing to not have a negative impact on print quality?

If I understand the concerns, they are that this option is a potential 'landmine'. Basically, the concern is that this option may produce unexpected and surprising defects when enabled that the users may fail to attribute to enabling that option. Which would lead them into a long and fruitless search of the defects of their printers that simply don't exist. Thus, for an option that's potentially this problematic it would be ideal to demonstrate that either:

- its dangers are overrated and most of the time the option is harmless even if incorrectly set;
- or, even if it _is dangerous_, its benefits outweigh the risks (e.g. on some printers you could easily reduce printing time by 30-50% or even more vs the case where you'd set the maximum acceleration and velocity to a minimum permissible for X and Y).

FWIW, my personal take is that there are only a handful of printers that would considerably benefit from this. Probably only very large bed slingers with very heavy bed could benefit from this - then one could set a maximum acceleration for Y around 700-1000-1500 mm/sec^2, and for the other axis - 3000-5000 mm/sec^2 depending on the machine, and speed up at least some moves. Most other printers, Ender 3 and MK3/4 including, would get single digit percentage improvements in the total print time (that's at least my suspicion). But I'd be happy to be proven wrong.

Separately, I'm not sure what the deal with
> to speed-print some storage bins on my Prusa Mini at 200mm/s, which would otherwise be physically impossible as the Y axis doesn't like going over 160.

Do you know why is it like that? I don't have that printer, but my naive assessment of it is that both X and Y axes are fairly symmetric (e.g. use the same 16T pulley with GT2 belts on the similar 1.8 degrees motors). Could this be an issue with StealthChop mode? Though in this case I'd imagine X motor (or rather, the driver) would be also operating at its limits.

your-friend-alice:
the landmine thing makes sense. i hoped adding a caveat to the docs would help but i can def see how if a user is having a specific problem, and they have a hundred different settings to change, it would be helpful to have a specific description of what symptoms to look out for.

As for my printer, my Y axis is fairly heavy, the motor is fine but the bearings start making noise at high acceleration. the best word i can find for it is like a squeak, but lower (somewhere around 300-400Hz).

I just started using klipper recently so my printer is not very well tuned yet, once i have it better tuned i can try running some tests. my assumption is that this won't introduce any new quality issues, but it will exacerbate pressure advance issues since going around a corner will change your speed. so if Y is set slower, maybe some over-extrusion near the edges of Y-oriented walls, and under-extrusion at the edges of X-oriented walls. but typically there's a million other reasons for a printer to speed up or slow down while extruding (say, slowing down for a tight corner) so i don't think it would add anything new.

I can snooze on this until I get the chance to test this more thoroughly, i'm guessing the results will be "kinda useful as a niche optimization"

kdb424:
Other thing to consider with per axis anything, is pressure advance in it's current state will get harder to tune if the machine is not treated the same on each axis. I'm running a modified kinematic based on the one from the other mentioned closed PR, and it has issues with that as well. There's testing and work into fixing these problems, but as things are with pressure advance, there are some print quality issues if you treat the axis differently on acceleration or max velocity in my experience. Once more work gets laid out to handle that, I believe that treating axis differently will be an improvement for most people. For now I'm only treating them differently on "invisible" features (like infill, supports, ect) to cope with the issue, but that's not intuitive and probably not a good idea to merge without more consideration.

BarsMonster:
I wish this option was available, not only for slow printers - but also for fast ones. 

There are more and more printers where movement speed is limited by reduced torque of stepper motor at high speed rather than inertia of print head. On CoreXY printers it would allow to immediately allow 1.4x speed for movements among X and Y axis, where 2 motors are working at the same time. I.e. right now I am at casual 400 mm/s speed, and with this change I can do 550 mm/s for ~50% of moves.

your-friend-alice:
note, this specific PR is just for cartesian kinematics, not CoreXY. In theory it wouldn't be that different to implement on CoreXY, might require some very basic additional math to get A and B axis movements out of `move.axes_d`

Sineos:
As already pointed out by @dmbutyugin and @kdb424:
A feature that has more ifs and whens than benefits, should not be added, regardless how much documentation is added.
* Experience shows that reading (and especially understanding) documentation is not terribly popular
* It is also more than evident that most folks live by "the more the better" philosophy: Terrible mechanics that not even run without binding is "remedied" with software optimization and tuning obscure TMC registers ("It has to be software! Completely unlikely that my highly sophisticated 200 USD printer would benefit from a die makers square and some lubrication")
* Solving a handful of edge cases but running the danger to create issues in a majority of careless deployments is not the way to go

Piezoid:
@BarsMonster I have a Core XY implementation [here](https://github.com/Piezoid/klipper/blob/work-peraxis/klippy/kinematics/limited_corexy.py)

On print quality: issues can be entirely avoided if the print accelerations are kept bellow the acceleration of slowest axis, same for velocities. This enables faster travels while keeping the accelerations isotropic during extrusion moves. Of course, this necessitates the "dynamic accelerations" settings in slicers.

@kdb424:  What kind of extruder are you using? Have you tried reducing the PA smooth time?

gaby64:
I use https://klipper.discourse.group/t/independant-acceleration-limits-for-x-and-y-axes/3831 on my Anycubic Chiron, the bed tends to make bad noises and skip steps if it tries to follow the accel on X.
Really which it was just included in klipper.

ganzgustav22:
> Solving a handful of edge cases but running the danger to create issues in a majority of careless deployments is not the way to go


>Terrible mechanics 


It's not an edge-case or terrible mechanics, it's just how 99% of printers are built and designed. X and Y are not the same. Not on a bedslinger, not on a classic cartesian, not on a CoreXY, not on a H-Bot.

Only printers I can think of that wouldn't need this feature because their axis are (nearly) the same are Cross-Gantry printers. Which are not very common. But even on those, X and Y is not exactly the same when looking at shaper graphs. They're close, but never the same.

your-friend-alice:
for what it's worth, these changes are rather small, it shouldn't be hard to keep this rebased against the upstream; I just ran a rebase now and there were no conflicts. I've made some changes to my hardware that make this feature less necessary so I probably won't keep up with this, but it's as simple as:

```
git clone https://github.com/your-friend-alice/klipper.git
git remote add upstream git@github.com:Klipper3d/klipper.git
git fetch upstream
git rebase upstream/master
```

I definitely did a bit of unnecessary refactoring in this patch  for example 97e5b7e7), but I have the changes split out into separate commits, which makes dealing with rebase conflicts a breeze (especially compared to merge conflicts, where all the changes are thrown at you in one go!) so if someone really wants to maintain it long term they can replay the commits and reconstruct something that puts more emphasis on cleanly patching against a moving target

chrisheib:
Just came across this, is there any change here? Still dormant / denied, until eventually picked up?

your-friend-alice:
if you can do some testing that validates that this produces beneficial results, you can definitely revive this PR! I just don't have the time to do that myself

chrisheib:
Ah, while the speed limits work awesomely, it seems the pressure adcanve doesn't scale. I would need to be able to set around 0.12 for my X axis at 12k accel, and 0.6 for my Y axis at 3k accel. Otherwise unfortunately the lines start to look reeeally different and print quality suffers a lot.

PR #6227, opened 2023-05-26
Kconfig: change default clock to 168MHz for HC32F460

SteveGotthardt:
Kconfig: The HC32F460 has a recommended clock of 168MHz. Make that the default.

Investigation with ICE shows the Ender Pro 2 has clock of 168MHz (was previously probed to be 200MHz)

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,8 +8,8 @@ All dates in this document are approximate.
 
 ## Changes
 
-20230519: `SHAPER_CALIBRATE` command immediately applies input shaper parameters
-if `[input_shaper]` was enabled already.
+20230525: `SHAPER_CALIBRATE` command immediately applies input shaper
+parameters if `[input_shaper]` was enabled already.
 
 20230407: The `stalled_bytes` counter in the log and in the
 `printer.mcu.last_stats` field has been renamed to `upcoming_bytes`.


JamesH1978:
Hi Steve, thanks for the PR, please can you sign it off with the signed-off-by line? 

Thanks
James

KevinOConnor:
Thanks.  I'm a little confused by this PR though - wouldn't other chips that run at 200Mhz now have a timing problem?  (Or, are all hc32f460 running at 168Mhz, and if so what does the comment in Kconfig about "anycube" mean?)

-Kevin

SteveGotthardt:
@KevinOConnor 

> Thanks. I'm a little confused by this PR though - wouldn't other chips that run at 200Mhz now have a timing problem? (Or, are all hc32f460 running at 168Mhz, and if so what does the comment in Kconfig about "anycube" mean?)
> 
> -Kevin


yes - anycube seems to run at 200MHz - how can i make the Kconfig adapt to that?   I had done this before by adding mainboard numbers in the Kconfig that I knew would use the hc32, but I think you (or someone) didn't like that?   Maybe it should be a low-level feature change? or?    thanks

KevinOConnor:
Sounds like a menu item should be added to allow the user to configure it - for example (totally untested):
```diff
--- a/src/hc32f460/Kconfig
+++ b/src/hc32f460/Kconfig
@@ -66,11 +66,17 @@ config STACK_SIZE
     int
     default 1024
 
-
+choice
+    prompt "Clock Speed"
+    config HC32F460_CLOCK_SPEED_168M
+        bool "168 MHz"
+    config HC32F460_CLOCK_SPEED_200M
+        bool "200 MHz"
+endchoice
 config CLOCK_FREQ
     int
-    default 200000000   # Voxelab uses 168000000
-
+    default 168000000 if HC32F460_CLOCK_SPEED_168M
+    default 200000000 if HC32F460_CLOCK_SPEED_200M
 
 config MCU
     string
```

-Kevin

SteveGotthardt:
@KevinOConnor 
Thanks for the help - 'choice' will be useful for the alternate app address as well.

KevinOConnor:
Thanks.  In general the changes look fine at a high-level, but I think this should probably be rebased as three separate commits: a change to allow frequency selection, a change to allow additional serial ports, and a change to allow selecting the boot address.

Separately, I'd prefer to avoid placing company names or product names in Kconfig.  (I know there are already a few in Kconfig, but over time I'd like to remove them.)  Manufacturers feel slighted if their products aren't in Kconfig and it leads to code churn.  If a particular product needs a setting then I think we should state that in their respective config/xxx.cfg file and not in "make menuconfig".

Cheers,
-Kevin

SteveGotthardt:
i don't know how to rebase as three separate commits

PR #6230, opened 2023-05-29
Add a temperature_combined class that can combine several other temperature sensors

EmJay276:
~I added a long desired feature to the temperature_sensor module to combine several sensors to a new one.~
~I have tested `temperature_sensor`, `extruder`, `heater_bed` with it.~

- ~The `sensor_type` must be set to `combined`~
- ~The sensors need to be provided as list in `sensors` matching the `self.printer.lookup_object('...') naming`~
- ~The combination type is privded in `type`, available options are `min`, `max`, `mean`~
- ~The value is updated each second(?)  `UPDATE_TEMPERATURE_INTERVALL = 1` with a register_timer~

UPDATE: 
I implemented a long desired feature to combine several sensors to a new temperature sensor called `temperature_combined` 
I have tested `temperature_sensor`, `extruder`, `heater_bed` with it.

- The `sensor_type` must be set to `temperature_combined`
- The sensors need to be provided as list in `sensors` matching the `self.printer.lookup_object('...') naming`
- The combination type is privded in `type`, available options are `min`, `max`, `mean`
- The value is updated each 0.3 seconds  `REPORT_TIME = 0.300` with a register_timer
- It can be used in heaters and extruders too

Signed-off-by: Michael Jäger [michael@mjaeger.eu](mailto:michael@mjaeger.eu)

UPDATE 2: New example config and picture

Example config
```
[temperature_sensor silicon_front]
sensor_type: NTC 100K MGB18-104F39050L32
sensor_pin: PC0
min_temp: 0
max_temp: 200

[temperature_sensor silicon_back]
sensor_type: NTC 100K MGB18-104F39050L32
sensor_pin: PC3
min_temp: 0
max_temp: 200

[heater_bed]
heater_pin: PD12
sensor_type: temperature_combined
sensor_list: temperature_sensor silicon_front,temperature_sensor silicon_back
combination_method: max
maximum_deviation: 30.0
max_power: 1.0
control = pid
pid_kp = 71.039
pid_ki = 2.223
pid_kd = 567.421
min_temp: 0
max_temp: 200
```
grafik


Sineos:
Can you elaborate on the use cases?

EmJay276:
@Sineos
I have two silicon heaters on my rectangular shaped build plate controlled by one SSR Relais.
I want to use the max temperature of both to control the heating to avoid damage by one of them overheating.

EmJay276:
Other usecases could be several sensors in one are that can be combined, like this frame sensors here
https://klipper.discourse.group/t/adding-a-virtual-frame-temperature-sensor/3399
or several heat chamber sensors to activate a fan if one of them is above the target temperature

Markus-1984:
@EmJay276 Thank you very much for your work. I´ll try it immediately

docs/Config_Reference.md
```diff
@@ -2519,6 +2519,21 @@ serial_no:
 #   The micro-controller to read from. Must be the host_mcu
 ```
 
+### Combined temperature sensor
+
+Combined temperature sensor is a virtual temperature sensor based on several other sensors. This sensor can be useed with extruders and heater beds.
```
kdb424, line 2524: Typo. "useed" should be "used"

kdb424:
Great change. Allows us to have several chamber temperature sensors and get an average of temperature as well. Working as expected.

JamesH1978:
Thankyou for submitting a PR, please refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially part 3 and the signed off by line.

Thanks
James

docs/Config_Reference.md
EmJay276: 👍 is fixed

EmJay276:
@JamesH1978 I tried to follow the contribution guidelines, if I missed anything please let me know 👌😀

EmJay276:
> ... signed off by line.

Signed-off-by: Michael Jäger [michael@mjaeger.eu](mailto:michael@mjaeger.eu)

livinhack:
> > ... signed off by line.
> 
> Signed-off-by: Michael Jäger [michael@mjaeger.eu](mailto:michael@mjaeger.eu)

I have relatively little knowledge "here", but don't you have to edit this line into the start post instead?

greetings

EmJay276:
> I have relatively little knowledge "here", but don't you have to edit this line into the start post instead?

ok thank's I've added it

klippy/extras/temperature_combined.py
```diff
@@ -0,0 +1,96 @@
+# -*- coding: utf-8 -*-
+# Support for combination of temperature sensors
+#
+# Copyright (C) 2023  Michael Jäger <michael@mjaeger.eu>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+UPDATE_TEMPERATURE_INTERVALL = 1.0
```
bryanboettcher, line 8: FYI, there's a typo here, which may prevent merging:  It should be spelled `UPDATE_TEMPERATURE_INTERVAL`.

docs/Config_Reference.md
```diff
@@ -2519,6 +2519,21 @@ serial_no:
 #   The micro-controller to read from. Must be the host_mcu
 ```
 
+### Combined temperature sensor
+
+Combined temperature sensor is a virtual temperature sensor based on several other sensors. This sensor can be used with extruders and heater beds.
+
```
bryanboettcher, line 2525: If this works with `[heater_generic]` sections too, that would be a good item to add.

bryanboettcher (changes requested):

klippy/extras/temperature_combined.py
EmJay276: Fixed via 5013943

docs/Config_Reference.md
EmJay276: Fixed via https://github.com/Klipper3d/klipper/commit/501394345675eb30dcd66b055d386dc34ca4e27b

EmJay276:
I found a problem I am not able to fix 😢 

Using the combined_sensor for a heater does report the correct values, but the heater_pin does not get triggered/activated when changing the target temperature (pwm get's set to 100%)...
I tried to debug it and find the responsible lines in the code, but I didn't manage to get work.

I implemented everything similar to the `PrinterADCtoTemperature` class, except the `self.mcu_adc` and `self.adc_convert,` because I am time triggered not analog triggered.
Perhapse the missing `query_analog_in` in `_build_config` in `class MCU_adc` does the trick, but I don't get whats happening there.

I haven't changed anyting at the output PWM pin in `class Heater`
```python
        # Setup output heater pin
        heater_pin = config.get('heater_pin')
        ppins = self.printer.lookup_object('pins')
        self.mcu_pwm = ppins.setup_pin('pwm', heater_pin)
        pwm_cycle_time = config.getfloat('pwm_cycle_time', 0.100, above=0.,
                                         maxval=self.pwm_delay)
        self.mcu_pwm.setup_cycle_time(pwm_cycle_time)
        self.mcu_pwm.setup_max_duration(MAX_HEAT_TIME)
```

Perhaps @KevinOConnor can help me out, you wrote most of the code I am dealing with.

kdb424 (approved):
Running smooth here. No requested changes from me.

KevinOConnor:
Thanks.  As high-level feedback, I'm not sure I fully understand the use case here.

I understand that you want to avoid damage to the bed heater, but it would seem to be simpler to declare a `[temperature sensor]` with a `max_temp` setting.

My concern with "merging" of temperatures is that it increases the risk of a serious failure - if either of the sensors fails the code would effectively ignore the warning signs and keep on sending power to the heater.

Separately, on the code implementation, it's not valid to "peek" into other class member variables - see the note at: https://www.klipper3d.org/Code_Overview.html#adding-a-host-module

-Kevin

bryanboettcher:
@KevinOConnor there are use-cases beyond just a sanity check for a temperature sensor.  For instance, the Slice Magnum+ hotend has dual thermistor sockets, and they are positioned on either side of the nozzle.  Because the thermistor cannot be directly inline with the nozzle, having a pair of them averaged together will give the best possible reading compensating for external effects like asymmetric cooling ducts.

In another example, a pair of heaters on opposite sides of an enclosure can be averaged together.  In this case, it makes sense to have the individual thermistors monitored to control their localized heater, and either thermistor should be monitored to ensure it's close to its partner.

I agree with your assessment of "one failing leads to a serious problem", and I have a set of changes I can PR that triggers a failure if the difference between the two values is above a configurable delta.  I think that would address your concerns.

EmJay276:
@KevinOConnor thank you for your feedback!
To clearify the high-level aspect: I use two heatbeds (with ONE Relais to controll them) and therefore want to use both heatbed sensors for the PID. This should be much more accurate than just using one of the sensors. As mentioned by @bryanboettcher this is a common usecase for several other hardware like current extruders.

> My concern with "merging" of temperatures is that it increases the risk of a serious failure - if either of the sensors fails the code would effectively ignore the warning signs and keep on sending power to the heater.

Having a `min_temp` and `max_temp` for each invidual sensor set completly avoids this risk. The `temperature_combined` can only use existing sensors.
If one of the sensors fails an error will be triggered like it's already the case.

> Separately, on the code implementation, it's not valid to "peek" into other class member variables - see the note at: https://www.klipper3d.org/Code_Overview.html#adding-a-host-module

Can you specify the exact part you are concered with?
I am only accessing "public" methods of the sensors, like `sensor.temperature_callback`, `sensor.setup_callback(...)`, `sensor.get_temp(...)`

klippy/extras/temperature_combined.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _handle_connect(self):
+        for sensor_name in self.sensor_names:
+            sensor = self.printer.lookup_object(sensor_name)
+            # if temperature sensor
+            if hasattr(sensor, 'get_temp'):
+                self.sensors.append(sensor)
+            # if heater
+            elif hasattr(sensor, 'heater'):
+                self.sensors.append(sensor.heater)
```
KevinOConnor, line 53: Unless I'm missing something, `heater` is a private member of the sensor class.

-Kevin

EmJay276: @KevinOConnor 
I have changed it to use `get_status` and updating being time triggered.
Now no private member is accessed anymore, and the callback functions don't need to be modified.

EmJay276:
I have changed the sensor class to use `get_status` of the underlying sensors and the update being now time triggered.
Now no more private member is accessed, and the callback functions don't need to be accessed and modified.

This PR should now meet all requirements mentioned here https://www.klipper3d.org/Code_Overview.html#adding-a-host-module.

KevinOConnor:
Thanks.  In general it seems fine, but I'm still a little leery about the emergency case.  Perhaps, add a `maximum_deviation` config parameter and call `self.printer.invoke_shutdown()` if the difference between the min and max temperatures exceed that value.

Also, I don't think it is a good idea to call `get_status()` with `self.reactor.NOW` - call `self.reactor.monotonic()` and pass that value to get_status.

Also, be sure to update the `temperature sensors` section of Status_Reference.md if a new `get_status()` export is added.

-Kevin

KevinOConnor:
Also, this module defines `min_temp` and `max_temp` values, but doesn't do anything with them.  These safety checks would need to be enforced.

-Kevin

bryanboettcher:
I have this started as a PR. Do you know if the author is on Discord?

EmJay276:
Everything should be fixed now
- I've added `maximum_deviation` and the safety checks for `min_temp` and `max_temp` and updated the References.
- I also added the `sensors` and the `type` used for combining them to the `get_status()` method.
- `self.reactor.NOW` was replaced by `self.reactor.monotonic()` (a relict from testing)

KevinOConnor:
Thanks.  In general it looks fine.  I do have some additional comments (sorry for the back and forth):
1. The code still seems to be passing `self.reactor.NOW` to `get_status()` in `_handle_connect()`.
2. Please set a reasonable default for `maximum_deviation`.  Perhaps 5 degrees?
3. It occurred to me that the config settings for this sensor are actually set in the `extruder` or `heated_bed` config section that uses them.  As such, I think more verbose config names may be useful - perhaps `sensor_list`, `sensor_combination_method`, and `sensor_maximum_deviation`?  If the naming is changed the status reference variables should likely be updated as well.

@bryanboettcher - I didn't understand your comment above.  Do you have a concern with merging this?

-Kevin

EmJay276:
> Thanks. In general it looks fine. I do have some additional comments (sorry for the back and forth):

No problem :)

> 1. The code still seems to be passing `self.reactor.NOW` to `get_status()` in `_handle_connect()`.

I changed it at the wrong line, it is fixed now.

> 2. Please set a reasonable default for `maximum_deviation`.  Perhaps 5 degrees?

I think we won't aggree here, having the `maximum_deviation` mandatory would be a possible solution for me.

In my oppinion deactivating it by default (using a large value) is reasonable. If one of the sensors fails, it is handled by the min/max of this sensor.
If you want the default to be "activated", I would suggest 20 degrees. Running into a emergency shutdown during a print because two chamber sensors differ by only 5 degrees would be very annoying for me.

> 3. It occurred to me that the config settings for this sensor are actually set in the `extruder` or `heated_bed` config 
> section that uses them.  As such, I think more verbose config names may be useful - perhaps `sensor_list`, `sensor_combination_method`, and `sensor_maximum_deviation`?  If the naming is changed the status reference variables should likely be updated as well.

The sensor can not only be used in extruder/heater (see below), therefore I would not suggest to name it `sensor_`..., but I like the `sensor_list` and `combination_method`. I renamed `type` to `combination_method` and `sensors` to `sensor_list`
```
[temperature_sensor silicon_front]
sensor_type: NTC 100K MGB18-104F39050L32
sensor_pin: PC0
min_temp: 0
max_temp: 200

[temperature_sensor silicon_back]
sensor_type: NTC 100K MGB18-104F39050L32
sensor_pin: PC3
min_temp: 0
max_temp: 200

[temperature_sensor sensor3]
sensor_type: temperature_combined 
sensor_list: temperature_sensor silicon_front,temperature_sensor silicon_back
combination_method: max
maximum_deviation: 20.0
min_temp: 0
max_temp: 200
```

Sineos:
> If you want the default to be "activated", I would suggest 20 degrees. Running into a emergency shutdown during a print because two chamber sensors differ by only 5 degrees would be very annoying for me.

My opinion is to have a lower minimum error in the sense of "better safe than sorry". 
It is quite a special feature that highly depends on the use-case, i.e. chamber sensors is a totally different story compared to a hotend. 
Thus, it is reasonable that a user sets a `maximum_deviation` that matches their use-case. In case it is forgotten, then the "better safe than sorry" philosophy kicks in.

Sineos:
I might have missed it: How is the combined sensor reported in the logging? Does it follow the known style, e.g.
```
heater_bed: target=0 temp=56.0 pwm=0.000 cold_junction: temp=35.9 voltage: temp=24.4 current: temp=0.1 fan_current: temp=0.0
```
Are the individual values, as well as the calculated value, identifiable for the user?

EmJay276:
> I might have missed it: How is the combined sensor reported in the logging? Does it follow the known style, e.g.
> Are the individual values, as well as the calculated value, identifiable for the user?

It was reporting by his `name` which was in case of the `heater_bed` also `heater_bed`
I changed it to `..._combined_sensor`
I also added the `sensor_list`, so the individual values are identifiable.

```
silicon_front: temp=23.4 silicon_back: temp=23.6 heater_bed_combined_sensor: temp=23.6 sensor_list=temperature_sensor silicon_front, temperature_sensor silicon_back heater_bed: target=0 temp=23.6 pwm=0.000
```

I don't think the combination method needs to be reported, do you?

Sineos:
> I don't think the combination method needs to be reported, do you?

IMO not needed, since it is a static value and can easily be taken from the cfg.

EmJay276:
Well, it is the same for the list of sensors.

KevinOConnor:
Thanks.  In general the code looks fine to me.  I don't think the code should have a `stats()` method though (I don't think it is needed and I fear it will break existing log parsers) and I think the only information exported by `get_status()` should be `temperature` (as the sensor_list and combination_method are already available from the config part of the status).  You can just add the `temperature_combined` to the list of sensors mentioned at https://github.com/Klipper3d/klipper/blob/master/docs/Status_Reference.md#temperature-sensors instead of adding a new section to the reference.

-Kevin

EmJay276:
@KevinOConnor `stats()` is removed, `get_status()` only reports `temperature` and the `Status_Reference.md` is changed too. :)

klippy/extras/temperature_combined.py
```diff
@@ -0,0 +1,128 @@
...
+                "COMBINED SENSOR temperature %0.1f "
+                "above maximum temperature of %0.1f."
+                % (self.last_temp, self.max_temp,))
+
+        # this is copied from temperature_host to enable time triggered updates
+        # get mcu and measured / current(?) time
+        mcu = self.printer.lookup_object('mcu')
+        measured_time = self.reactor.monotonic()
+        # convert to print time?! for the callback???
+        self.temperature_callback(mcu.estimated_print_time(measured_time),
```
bryanboettcher, line 114: ...?

KevinOConnor:
Thanks.

-Kevin

PR #6233, opened 2023-05-31
bme280: use consistent report time for BME680

Arksine:
This PR supersedes #6221, fixing the BME680 issue by reporting every .8 seconds.  As before I have tested this change against a BME280, however it would be useful if someone would like to make sure the BME680 works as intended.


pedrolamas:
FWIW, I've also tested this against the BME280 and it works perfectly fine.

I had a BME680 that seems to be dead, so I've ordered a new one, but that will take some time to get delivered...

KevinOConnor:
Thanks.

-Kevin

PR #6235, opened 2023-06-01
Kconfig: update for HC32F460 - clock, flash add, serial

SteveGotthardt:
Kconfig: adding serial port selection, 'choice' for flash app address and clock speed.

### DIFF
--- a/src/hc32f460/Kconfig
+++ b/src/hc32f460/Kconfig
@@ -46,8 +46,18 @@ config FLASH_SIZE
     hex
     default 0x40000
 
+choice "Application Address"
+    prompt "Application Address"
+    config HC32F460_FLASH_APPLICATION_ADDRESS_0x8000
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  I did a "squash and merge" on these three commits.  In the future I can pull all three independent commits if each commit message is formatted according to the guide lines described at: https://www.klipper3d.org/CONTRIBUTING.html#format-of-commit-messages .  (Alas, github doesn't make it easy for me to modify the commit messages unless I also squash them all into one commit.)

Cheers,
-Kevin

remizca:
hi, how do I use the clock speed selection? i'm fairly new to this and i;m trying to solve my issue on using bltouch on and ender 2 pro with a 2.4.s4 board and one reply i got was that he used 168MHz when h flashed klipper

PR #6240, opened 2023-06-05
Always set TCOOLTHRS to 0xfffff during homing.

andrewmcgr:
"tmc: Do not override tcoolthrs if it is configured" allows a configured tcoolthrs to be used during homing.

Unfortunately, it appears that the datasheet for the TMC2240 is incorrect. It states that if the step rate is below tcoolthrs, stall detection should be enabled. This does not seem to be true, and having this register set to 0xfffff appears to be necessary for sensorless homing to work. I have seen the status register record a stall but the diagnostic pin not register with tcoolthrs set to other values, even values greater than tstep at the time.

Since homing moves should be slow anyway, it seems unnecessary to provide configuration for this, and setting 0xfffff for all chips that support CoolStep should be adequate.

This does retain the other functionality of resetting tcoolthrs to its previous value when the homing move finishes.

### DIFF
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -516,9 +516,8 @@ class TMCVirtualPinHelper:
             self.fields.set_field("en_pwm_mode", 0)
             val = self.fields.set_field(self.diag_pin_field, 1)
         self.mcu_tmc.set_register("GCONF", val)
-        if self.coolthrs == 0:
-            tc_val = self.fields.set_field("tcoolthrs", 0xfffff)
-            self.mcu_tmc.set_register("TCOOLTHRS", tc_val)
+        tc_val = self.fields.set_field("tcoolthrs", 0xfffff)
+        self.mcu_tmc.set_register("TCOOLTHRS", tc_val)
     def handle_homing_move_end(self, hmove):
         if self.mcu_endstop not in hmove.get_mcu_endstops():
             return


KevinOConnor:
@leptun - do you have any comments on this?

-Kevin

leptun:
I don't think this is a good idea. I implemented that feature specifically to preserve backwards compatibility with old configs. So if no coolstep velocity is configured, then tcoolthrs would be 0 by default, which would then be changed to 0xfffff in `handle_homing_move_begin()`. If tcoolthrs is configured for any reason, then handle_homing_move_begin() shouldn't change it. tcoolthrs is really annoying since it is shared between both the stall detection feature and the coolstep feature, which are used in very different ways on printers.

> Unfortunately, it appears that the datasheet for the TMC2240 is incorrect. It states that if the step rate is below tcoolthrs, stall detection should be enabled. This does not seem to be true, and having this register set to 0xfffff appears to be necessary for sensorless homing to work.

I don't think that is true. In my testing, that was not the case at all, tcoolthrs working just fine. I actually use it in my config for the printer since without it, false stalls happen during the acceleration ramp.

> I have seen the status register record a stall but the diagnostic pin not register with tcoolthrs set to other values, even values greater than tstep at the time.

How did you get the values of the registers? Klipper is quite slow when it comes to fetching the tmc register data.

> Since homing moves should be slow anyway, it seems unnecessary to provide configuration for this, and setting 0xfffff for all chips that support CoolStep should be adequate.

I disagree. It's not really about speed, but about reliability. Stallguard performs incredibly poorly at low velocities. Low velocities are not really avoidable in an acceleration ramp, so the stall events at those velocities must be filtered out. This is why TCOOLTHRS is necessary to be configurable. Using the default value of 0xfffff allows stall events even at low velocities, but in my case, that produced mediocre results at best.


@andrewmcgr Can you please post your config and also a register dump right before homing and after homing?

andrewmcgr:
> I don't think this is a good idea. I implemented that feature specifically to preserve backwards compatibility with old configs. So if no coolstep velocity is configured, then tcoolthrs would be 0 by default, which would then be changed to 0xfffff in `handle_homing_move_begin()`. If tcoolthrs is configured for any reason, then handle_homing_move_begin() shouldn't change it. tcoolthrs is really annoying since it is shared between both the stall detection feature and the coolstep feature, which are used in very different ways on printers.

I can't disagree that it's annoying that the register is overloaded like that.

> > Unfortunately, it appears that the datasheet for the TMC2240 is incorrect. It states that if the step rate is below tcoolthrs, stall detection should be enabled. This does not seem to be true, and having this register set to 0xfffff appears to be necessary for sensorless homing to work.
> 
> I don't think that is true. In my testing, that was not the case at all, tcoolthrs working just fine. I actually use it in my config for the printer since without it, false stalls happen during the acceleration ramp.

Sounds like I have some more investigation to do then.

> > I have seen the status register record a stall but the diagnostic pin not register with tcoolthrs set to other values, even values greater than tstep at the time.
> 
> How did you get the values of the registers? Klipper is quite slow when it comes to fetching the tmc register data.

I don't recall exactly, I will have to attempt to reproduce that. It may be that I had not quite understood the diag pin registers at that time.

> 
> > Since homing moves should be slow anyway, it seems unnecessary to provide configuration for this, and setting 0xfffff for all chips that support CoolStep should be adequate.
> 
> I disagree. It's not really about speed, but about reliability. Stallguard performs incredibly poorly at low velocities. Low velocities are not really avoidable in an acceleration ramp, so the stall events at those velocities must be filtered out. This is why TCOOLTHRS is necessary to be configurable. Using the default value of 0xfffff allows stall events even at low velocities, but in my case, that produced mediocre results at best.

I see. I've had no trouble with that using the 2240s, though it may explain occasional homing issues with the 2209s. This does not seem to be well known amongst the community or well documented anywhere but the datasheets; I was aware of it from reading the datasheets, but it seemed to be more theoretical provided that the homing speed was adequate.

For context: I have been working on a method for autotuning everything possible on a TMC driver, given information that is already in the printer config file and the datasheets for the motors. Current state of that code is in [this repository](https://github.com/andrewmcgr/klipper_tmc_autotune) (as a Klipper extra, though I'm not against putting it in the form of a PR once I have a chance to write full documentation).

One of the issues with this (and coolthrs in general) is that it often results in a rather high speed for coolthrs, since it must be faster than pwmthrs and the desired effect is to retain stealthchop to the highest reasonable speed. The configuration I have on my printer right now, with 5160s, results in a coolthrs of 103 mm/s, and I have concerns about the mechanical consequences of homing at 105 mm/s or more, especially if it fails.

So, I take your point about setting that specific value for coolthrs during homing. But maybe we should be providing the ability to set it to a different configurable value during homing (for example, the configured homing speed less 1/8 to clear the small_hysteresis) and returning it to the run value afterward? This would be rather difficult in a macro, though I think I could make the autotuning code set the configured value so macros can look it up.

I'm still not sure that would work on my 2240 setup, but I'll test and see what happens once I get a chance to put those drivers back in the printer.

> 
> @andrewmcgr Can you please post your config and also a register dump right before homing and after homing?

I can, but not immediately. The examples in that repository above generate the configuration I was running, however. Printer is a Voron 2.4 350mm, BTT Octopus, all BTT drivers (either 2240, or 5160 Pro V1.1). XY motors are the common LDO 2004mah.

leptun:
> One of the issues with this (and coolthrs in general) is that it often results in a rather high speed for coolthrs, since it must be faster than pwmthrs

Does it have to be like this? I don't recall any restriction of this kind. In my setup I have the coolthrs velocity to something like 15mm/s and pwmthrs velocity to something like 200mm/s. On drivers with StealthChop2, coolstep and stall detection operates in the 15-200mm/s velocity range, while on drivers like tmc2240, it operates in a larger speed range (15-VHIGH mm/s).

Also, please beware there is some contradicting information in the TMC2240 datasheet. Since this is the first driver that has both StallGuard2 and StallGuard4, some parts of the datasheet still explain registers as if only one or the other is available. However, I think the StallGuard4 sections take precedence since those are newer and specifically written for this chip. Here is a snippet where they explain that in order to use StallGuard4 (and by extension coolstep) during StealthChop2 operation, you do indeed set coolthrs to a low velocity and pwmthrs to a higher velocity, so it is a valid combination:

```
DIAG is pulsed by StallGuard, when SG4_RESULT falls below SG4_THRS. It is only enabled in StealthChop2 mode,
and when TCOOLTHRS ≥ TSTEP > TPWMTHRS.
```

However, since you mention that you have issues with the diag output not working, please consider the issue I discovered with the driver (tmc2240 only) where the SG4_THRS register is implemented incorrectly in this chip, so the comparison range in StallGuard4 for the DIAG output is wrong. This could explain your issue if you were attempting to calibrate Stall detection for homing with StallGuard4. If you want more info about this issue, I will try to find where I wrote about it.
Please note that while SG4_THRS is broken for the DIAG output, StallGuard4 itself still works, so Coolstep with StallGuard4 works just fine, and so is the sg_angle_offset calibration. StallGuard2 during SpreadCycle is also completely unaffected by this.

andrewmcgr:
> Does it have to be like this? I don't recall any restriction of this kind.

According to the datasheets, yes. Page 76 of the 5160 datasheet is all about this, including requiring VPWMTHRS < VCOOLTHRS. The equivalent part of the 2240 datasheet, pp 49 and 50, seems to say that on that chip, if VCOOLTHRS < VPWMTHRS, they're treated as equal and you get CoolStep over VCOOLTHRS, while VPWMTHRS will be ignored.

> However, since you mention that you have issues with the diag output not working, please consider the issue I discovered with the driver (tmc2240 only) where the SG4_THRS register is implemented incorrectly in this chip, so the comparison range in StallGuard4 for the DIAG output is wrong.

I was homing the 2240 in SpreadCycle, because I was unable to get any sense out of SC4_THRS, probably because of the issue you refer to.

I should add: the autotuning works very well indeed on the 2240, producing amazingly good print quality and very low noise, without meaningfully compromising the speed of the motion system. Still experimenting with the 5160, I get the low noise but unfortunately also layer shifts, which I don't yet understand.

leptun:
The VPWMTHRS < VCOOLTHRS restriction only makes sense on the drivers with StallGuard2 only. So having such a restriction on tmc2240 with both StallGuard2 and StallGuard4 doesn't make much sense to me.
Pages 49-50 from the tmc2240 manual seem to be more or less a copy paste form older datasheets, having no mention of StallGuard4. I really think any claims about VPWMTHRS and VCOOLTHRS relationship in that chapter should be ignored.
If what they claim at pages 49-50 is true, then the newly added chapter with the title `CoolStep Load Adaptive Current Scaling` is entirely wrong and CoolStep can't be used with both StallGuard2 and StallGuard4 if you can't set VCOOLTHRS below VPWMTHRS. That's why I'm saying there is conflicting information in the tmc2240 datasheet.

As for other drivers with StallGuard2 only, what you say is most likely correct. Those drivers being tmc2130 and tmc5160. For those vcoolthrs must indeed be greater than vpwmthrs.
For tmc2208, there is no vcoolthrs, so we'll ignore that one.
For tmc2209, there is StallGuard4 only. With that one, VCOOLTHRS must be smaller than VPWMTHRS. So with this driver the condition is inverted since StallGuard4 is made to operate in StealthChop2 mode only.

andrewmcgr:
> Please see https://github.com/Klipper3d/klipper/pull/6134. Maybe that's why you get layer shifts. If you can confirm that's the issue, maybe that will convince @KevinOConnor to reopen my PR and merge it.

I have tested, and can confirm that enabling multistep_filt does indeed fix my layer shifts. Will comment over there too.

andrewmcgr:
So, this PR is definitely not necessary for the 5160 provided the thresholds are such that homing speed > vcoolthrs.

Next, test the 2240.

andrewmcgr:
Also not necessary for the 2240 now that I understand what the conditions for sensorless homing actually are on that chip.

Thank you all for the discussion, closing the PR.

PR #6241, opened 2023-06-05
Add saved gcode states to printer object

rootiest:
This allows referencing the `saved_states` created by the `SAVE_GCODE_STATE` command.

The `saved_states` object was already part of the `gcode_move` module, but wasn't included in the status reference.

Including it will allow macros/etc to test if any or a specific saved_state exists. They can also reference properties of that state such as the saved coordinates, absolute/relative positioning, etc.

An example where this could be useful would be a macro where you'd like to check if a saved_state already exists and restore it if so, or do something different if not.

Signed-off-by: Chris Laprade <chris@rootiest.com>


rootiest:
Here's a screenshot to show the output with this added:

output

KevinOConnor:
Thanks.  As high-level feedback I think we can export the saved_states, but I'm not sure the code here is correct.

The `get_status()` call needs to return objects that are not mutable (there is a blurb on this at https://www.klipper3d.org/CONTRIBUTING.html ).  So, just exporting `saved_states` would result in the api server not providing updates as expected.

Also, we typically export coordinates using the `Coord()` wrapper for easier access from macros - exporting the raw state seems like it would be less uniform.

I'm sure the above could be addressed, but it would likely make the change quite a bit more complex.  Can you go into more detail on the use-case you have in mind for this change?

-Kevin

rootiest:
Ah, I didn't realize it was not that simple.

The main goal I had in mind was just to check if a given saved_state exists by name. So that an if statement could determine whether a profile exists to restore or if another path should be taken.

The other stats like last_position and the absolute/relative mode were just side benefits that seemed like they could be useful as well in some cases. Perhaps if you wanted to move closer to the restore position before executing a RESTORE_GCODE_STATE?

KevinOConnor:
I understand.  I'm not sure the best way to proceed.  An alternative to changing the code may be to use macros to explicitly set a variable that the other macros could use to determine if some action is taken.  Another option might be to change the code to just export the list of saved state names.

-Kevin

rootiest:
So just returning the profile names would work?

I guess I just didn't quite understand what qualifies as immutable.

So having each profile be an array is a no-no, but we can have an array of profile names?

That would work for me if that fits the requirements. My main goal here was having a way to see the names of the save_states available.

KevinOConnor:
By directly exporting a reference to the Python dictionary, if the other code alters the content in that reference then the API code will not notice the change.  To make it immutable, one typically must alter the code to make a copy of the dictionary on each update (so that the API code can observe that there is a change in state and can relay that change), or alternatively to change the `get_status()` code to copy the information (so that it returns a new reference on each call).

Cheers,
-Kevin

PR #6242, opened 2023-06-07
bed_mesh: improve relative reference implementation

Arksine:
The current implementation of the `relative_reference_index` feature was initially added when the mesh configuration was static and could not be changed during runtime.  When runtime parameters were introduced it was also possible to change the `relative_reference_index` to accommodate a new mesh, however this puts restrictive requirements on where the new mesh is located as the new index must reference a coordinate near the original index.

With the forthcoming addition of "adaptive" mesh leveling it is necessary to rework how the reference coordinate is handled.  This pull request makes the following changes:

- The original points are no longer cached.  Points are generated each time `BED_MESH_CALIRBATE` is run to guarantee that
all attributes that depend on point generation are set.

-  A new `zero_reference_position` option has been added.  This behaves like the relative reference index, however it takes an 2D coordinate rather than an index.  If the coordinate is located within the mesh then the mesh will be offset to make this location apply zero adjustment.  If the coordinate is outside of the mesh then its location will be probed after calibration with the resulting Z value will be used as the z_offset.

-  To avoid breaking existing configurations the `relative_reference_index` option has not been removed.  If the RRI is specified then its location will be determined at startup, with the result set as the `zero_reference_position`.  Its behavior will mirror that outlined above.  The `RELATIVE_REFERENCE_INDEX` parameter for `BED_MESH_CALIBRATE` has been removed as this value cannot be changed after initialization.

These changes shouldn't affect existing configurations with the possible exception that some users may update the RRI via gcode command.  When it is determined that this is ready to merge I will note this change in Config_Changes.md.

### DIFF
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -261,13 +261,18 @@ fade_target: 0
   mesh is used, however it may be desirable to manually adjust the fade target
   if one wants to print on a specific portion of the bed.
 
-### The Relative Reference Index
+### Configuring the zero reference position
 
-Most probes are susceptible to drift, ie: inaccuracies in probing introduced by
-heat or interference.  This can make calculating the probe's z-offset
-challenging, particularly at different bed temperatures.  As such, some printers
-use an endstop for homing the Z axis, and a probe for calibrating the mesh.
-These printers can benefit from configuring the relative reference index.
+Many probes are susceptible to "drift", ie: inaccuracies in probing introduced
+by heat or interference.  This can make calculating the probe's z-offset
+challenging, particularly at different bed temperatures.  As such, some
+printers use an endstop for homing the Z axis and a probe for calibrating the
+mesh. In this configuration it is possible offset the mesh so that the (X, Y)
+`reference position` applies zero adjustment.  The `reference postion` should
+be the location on the bed where a
+[Z_ENDSTOP_CALIBRATE](./Manual_Level#calibrating-a-z-endstop)
+paper test is performed.  The bed_mesh module provides the
+`zero_reference_position` option for specifying this coordinate:
 
 ```
 [bed_mesh]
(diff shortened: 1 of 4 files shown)


Arksine:
@JamesH1978 The developers of KAMP are in the process of adapting their implementation to Python.  I was asked for feedback and specifically to handle the RRI.  Currently the best they can do is estimate the nearest point in the adapted mesh which would likely result in an incorrect offset.

voidtrance:
Yes, we are testing with a version which picks the point closest to the "pre-defined" RRI from the adapted mesh. However, if this gets accepted, all that will be removed.

Arksine:
Thanks Kevin.   I agree, it is best to deprecate the `relative_reference_index` to reduce long term confusion.  I pushed that change along with an update to the docs as separate commits so the changes are easy to review and will squash them before the merge.

I'm going to let some folks know that this change is coming as it is likely to generate substantial activity on support discords and forums.  Perhaps we can merge on Monday giving them time to spread the word.

kdb424:
I've been actively letting people know of this as well, and most agree that it's not only better, but much more easy to understand. RRI seemed to be a hard to grasp concept for many, and this addresses not only function, but ease of understanding.

Arksine:
I have squashed the deprecation commits and updated `config_changes.md`.  I believe that this can be merged when you are ready Kevin.

KevinOConnor:
Thanks.  Sorry I missed this yesterday.

-Kevin

MICHELOSCO:
i have installed the update but i am experiencing this strange issue: looks like the zero_reference_position offset does not apply to fade_end but only at the print start height so my mesh compension stretches the print.
i use z endstop fot nozzle end probe only for mesh

Arksine:
@MICHELOSCO I see the problem. I'll have a PR for a fix shortly.  In the short term if you add `fade_target: 0` to the `[bed_mesh]` section it will likely correct the issue.

Frix-x:
Hello,

I'm writing here since I got some problems since this recent change when using my adaptive bed mesh macros (this is also used by a lot of people since more than two years now), so I need to find a solution to fix it. I can agree that previously, the RRI system was a bit complex and lacked flexibility, but when used correctly, it worked pretty well.

I appreciate the recent changes because they have made it much easier for everyone to understand and utilize it. However, there's been a setback when it comes to adaptive mesh macros. Indeed, the issue is the inability to dynamically adjust the zero reference position as a BED_MESH_CALIBRATE parameter.

To explain, if the mesh is created on just one corner of the bed (not centered), the zero reference position will still be at the bed's center (or where it is statically configured). It remains there because it cannot be dynamically moved currently.

In the past, when using a virtual probe, I was able to home the Z axis directly on the mesh center (which was set as the RRI). This was pretty cool as the zero of the axis was done directly in the mesh center position and it was working pretty well. Now, with the new changes, this cannot be done anymore and it add a new probing point every time.

Could we consider reintroducing the ability to dynamically adjust the zero reference position as a parameter of BED_MESH_CALIBRATE? I think this will allow best of both world and avoid breaking all the adaptive mesh macros that was using this mechanism previously. If necessary, I'd be happy to make a minor PR to implement this change as it seems like a very small adjustment that could fix this issue.

PR #6244, opened 2023-06-08
New build menu to select optional features on chips with small flash

KevinOConnor:
Add a new "optional features" menu to "make menuconfig" for chips with limited code size (the atmega168, ar100, stm32f031, stm32f042, and pru).  This menu allows a user to select optional code features (bitbanging support, display support, sensor support, software spi, and software i2c).

This fixes the build for the stm32f042, in that it allows users to disable software i2c support (or some other feature) to make the build fit within the chip flash again.

-Kevin


Zemistr:
Tested with stm32f042x6 in Voron V0 display and works great.
To build the firmware I disabled software based features.

Zemistr:
@KevinOConnor Can you add also atmega328p and similar?

Fisheiyy:
This could also be possibly applicable to the RP2040 depending on its hardware configured flash size

PR #6245, opened 2023-06-09
Config: remove move_to_previous from Ender 3 S1 configs

JamesH1978:
This PR is to remove `move_to_previous: true` from the Ender 3 S1 and S1 plus configs. This is a confusing option, it is not really useful on these type of printers and can cause issue to the end user

Thanks
James

Signed_off_by: James Hartley <james@hartleyns.com>

### DIFF
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -120,7 +120,6 @@ home_xy_position: 147, 154
 speed: 75
 z_hop: 10
 z_hop_speed: 5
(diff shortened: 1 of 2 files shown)


Fisheiyy:
This is fine, but I added that so it will home like an OG E3 and also because when homed in the middle and the idle timeout activates, the gantry can sag and rehoming could possibly cause the nozzle to scrape across the bed.
I dont see how it could cause issues, but I agree its not very useful.

KevinOConnor:
Okay, thanks.

-Kevin

PR #6246, opened 2023-06-10
Simultaneous activation of steppers

shavrin777:
Adding the functionality of simultaneous activation or deactivation of specified motors.
It is needed in printers with multiple steppers per axis to synchronize steppers positions without parking.

Signed-off-by: Evgenii Shavrin [shavrin0591@gmail.com](mailto:shavrin0591@gmail.com)

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20230611: The `stepper` parameter in command SET_STEPPER_ENABLE has
+been renamed to `steppers`. Now it accepts multiple steppers.
+The `stepper` parameter is deprecated and will be removed in the near future.
+
 20230530: The default canbus frequency in "make menuconfig" is
 now 1000000. If using canbus and using canbus with some other
 frequency is required, then be sure to select "Enable extra low-level
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1193,8 +1193,10 @@ settings. Requires `control_pin` to be provided in the config section.
 The stepper_enable module is automatically loaded.
 
 #### SET_STEPPER_ENABLE
-`SET_STEPPER_ENABLE STEPPER=<config_name> ENABLE=[0|1]`: Enable or
-disable only the given stepper. This is a diagnostic and debugging
+`SET_STEPPER_ENABLE STEPPERS=<config_name> [ENABLE=[0|1]]`: Enable or
+disable only the given steppers. `<config_name>` can be one or
+more configured steppers, delimited with comma, for example
+`STEPPERS=stepper_x,stepper_y`. This is a diagnostic and debugging
 tool and must be used with care. Disabling an axis motor does not
 reset the homing information. Manually moving a disabled stepper may
 cause the machine to operate the motor outside of safe limits. This
--- a/klippy/extras/stepper_enable.py
+++ b/klippy/extras/stepper_enable.py
@@ -106,7 +106,7 @@ class PrinterStepperEnable:
                 el.motor_enable(print_time)
             else:
                 el.motor_disable(print_time)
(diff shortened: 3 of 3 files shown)


condottab:
I've been testing it for two weeks now without any problems. It has been a great help in my struggle to synchronise the awd corexy motors. Without it, you could enable all four motors by reversing the behaviour of the endstops, which is at best scuffling. SET_STEPPER_ENABLE allows you to enable the steppers one at a time, which caused desynchronisation when printing. My guess is that this is due to the fact that the next full step is different as all the steppers are enabled at the same time during the homing process.

ripinteer:
QuadXY (aka AWD CoreXY) requires manual motor step synchronization to avoid movement artifacts such as excessive ringing and reduced overall speed.  To perform this procedure all 4 motors had to be activated simultaneously* (One-by-one activation leads to unsatisfying results). Seems like there is no straight way to do it without employing tricks and config manipulations. So I want to see this feature in the mainline Klipper. This is similar to Marlin's M17 command, where you can specify multiple motors to be activated at once.

KevinOConnor:
Thanks.  I'm confused why this is needed though - a macro that calls SET_STEPPER_ENABLE for all steppers should be functionally identical.  Enabling and disabling a stepper doesn't alter it's internal or external position, so it's not clear to me how looping through a list of steppers in the python code is functionally different from looping in a macro.

-Kevin

condottab:
> Enabling and disabling a stepper doesn't alter it's internal or external position, so it's not clear to me how looping through a list of steppers in the python code is functionally different from looping in a macro.

The position changes, the stepper jumps to the next full step when it is activated, although this angle of rotation is not very large, it's still enough to cause an uneven tension on the belts. I'm not quite sure why the simultaneous activation leads to a different position, but this way the synchronisation stays, whereas when I synchronised the steppers with a macro and activated them one by one, the synchronisation was unstable. I tested this with m84, when you activate and deactivate the motors you can hear a sound, and the sound of the m84 is louder when the steppers are activated one by one, which means that you have released more tension.

joaomajesus:
> Thanks. I'm confused why this is needed though - a macro that calls SET_STEPPER_ENABLE for all steppers should be functionally identical. Enabling and disabling a stepper doesn't alter it's internal or external position, so it's not clear to me how looping through a list of steppers in the python code is functionally different from looping in a macro.
> 
> -Kevin

This is just a guess but is it possible that iterating in python is quicker than in the macros so that it appears that the motors enable simultaneously? In that case, it would still be better than using the macro for the purpose of syncing.

On the other side, it was mentioned that homing activates the motors simultaneously. Does the python code also iterate the motors or is there a different mechanism? May be worth investigating.

shavrin777:
> This is just a guess but is it possible that iterating in python is quicker than in the macros so that it appears that the motors enable simultaneously? In that case, it would still be better than using the macro for the purpose of syncing.

It is quicker, macro will trigger delay on each activation as `toolhead.dwell(DISABLE_STALL_TIME)`, what i avoided. Also I think setting enable pin with `print_time` will result at simultaneous activation.

About activation of motors during homing, I haven't been able to find how it is implemented. If Kevin points out what I missed, that would be nice.

KevinOConnor:
> About activation of motors during homing, I haven't been able to find how it is implemented. If Kevin points out what I missed, that would be nice.

Motors are automatically enabled during any movement.  The code path is `stepper_enable.py:EnableStepper` calls `stepper.add_active_callback(self.motor_enable)`.  Stepper movement is made by calling `stepper.generate_steps()` which will check for any registered "active callbacks" and thus will call the registered `motor_enable()` callback, which goes on to enable the stepper motor.

As to this PR, I guess I don't understand the real-world tasks that this change facilitates.  I'm leery of altering the parameters of an existing command and I'm leery of adding complexity to the code.

-Kevin

PR #6247, opened 2023-06-12
dockable_probe module

cloakedcode:
This is a redesign of the work done in #4328 to create a `[dockable_probe]` module to support Klicky-style probes that are physically detached most of the time from the toolhead.

## How we got here
The original PR has been sitting for a long time without attention and I took it upon myself to get it integrated into the main branch. Along the way I ended up simplifying the code and distilling it down to what is basically a handful of glorified movement commands (see the `MOVE_TO_*_PROBE` g-codes).

## Prior art
As mentioned, this is based on Mental's work but is significantly different. Rather than try to accommodate many different types of probe hardware and configurations, I targeted Klicky-style probes with fixed X/Y and fixed Z configurations. Other hardware and setups can override the movement g-codes where necessary to support their probes. See the `docs/Dockable_Probe.md` docs for setup examples.

## Testing
I've been running this code for about half a year on a bed slinger without issues (probe dock mounted Z 0). I put the code on a CoreXY today and successfully got it running there as well (probe dock mounted at fixed X, Y, used as virtual endstop for Z).

Back in Fall '22, I [posted on the Klipper Discourse about this code](https://klipper.discourse.group/t/reviving-the-dockable-probe/4256) and got help with wider tests and ironed out a couple bugs. There's been little chatter there until recently (a user with a very unique printer design) so I'm thinking it's been in the wild long enough to prove stable.

## Commits
I broke this down into three commits (rebased and squashed commits):

1. Everything Mental did on the original module (with potentially minor tweaks from me)
2. Everything I did in the rewrite, development followed the Discourse thread
3. A refactor commit to clarify/simplify the attachment verification methods

Let me know if this should be broken down in other ways.

## Finally

Thanks for your time and consideration! I know myself and others will be happy to have this feature native in Klipper.

### DIFF
--- a/klippy/extras/dockable_probe.py
+++ b/klippy/extras/dockable_probe.py
@@ -1,0 +1,600 @@
+# Dockable Probe
+#   This provides support for probes that are magnetically coupled
+#   to the toolhead and stowed in a dock when not in use and
+#
+# Copyright (C) 2018-2021  Kevin O'Connor <kevin@koconnor.net>
+# Copyright (C) 2021       Paul McGowan <mental405@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging
+from . import probe
+from math import sin, cos, atan2, pi, radians, degrees, sqrt
+
+PROBE_DETACH_DELAY         = .005
+PROBE_VERIFY_DELAY         = .1
+
+PROBE_UNKNOWN   = 0
+PROBE_ATTACHED  = 1
+PROBE_DOCKED    = 2
+
+MULTI_OFF       = 0
+MULTI_FIRST     = 1
+MULTI_ON        = 2
+
+HINT_VERIFICATION_ERROR = """
+{0}: A probe state verification method
+was not provided. A method to verify the probes attachment
+state must be specified to prevent unintended behavior.
+Please see {0}.md or config_reference.md for
+valid state verification methods and examples.
+"""
+
+# Helper class to handle polling pins for probe attachment states
+class PinPollingHelper:
+    def __init__(self, config, endstop):
+        self.printer = config.get_printer()
+        self.printer.register_event_handler('klippy:connect',
+                                        self._handle_connect)
+        self.query_endstop = endstop
+        self.last_verify_time  = 0
+        self.last_verify_state = None
+    def _handle_connect(self):
+        self.toolhead = self.printer.lookup_object('toolhead')
+    def query_pin(self, curtime):
+        if (curtime > (self.last_verify_time + PROBE_VERIFY_DELAY)
+            or self.last_verify_state is None):
+            self.last_verify_time = curtime
+            query_time = self.toolhead.get_last_move_time()
+            self.last_verify_state = not not self.query_endstop(query_time)
+        return self.last_verify_state
+    def query_pin_inv(self, curtime):
+        return not self.query_pin(curtime)
+
+# Helper class for manually set probe attachment states
+class ManualStateHelper:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        gcode = self.printer.lookup_object('gcode')
+        gcode.register_command('SET_PROBE_STATUS',
+                                self.cmd_SET_PROBE_STATUS,
+                                desc=self.cmd_SET_PROBE_STATUS_help)
+
+        self.probe_state = PROBE_UNKNOWN
+    cmd_SET_PROBE_STATUS_help = ("Manually specify probe status, valid probe " +
+                                 "states are UNKNOWN, ATTACHED, and DOCKED")
+    def cmd_SET_PROBE_STATUS(self, gcmd):
+        probe_states = ['UNKNOWN', 'ATTACHED', 'DOCKED']
+        state = gcmd.get('STATE').upper()
+        if state in probe_states:
+            self.probe_state = next(k for k, v in enumerate(probe_states)
+                                     if state == v)
+        else:
+            e = "Invalid probe state: {0}. Valid probe states are [{1}]"
+            raise gcmd.error(e.format(state, probe_states))
+    def query_pin(self, curtime):
+        if self.probe_state == PROBE_ATTACHED:
+            return True
+    def query_pin_inv(self, curtime):
+        if self.probe_state == PROBE_DOCKED:
+            return True
+
+# Helper class to verify probe attachment status
+class ProbeState:
+    def __init__(self, config, aProbe):
+        self.printer = config.get_printer()
+        self.check_open_attach = config.getboolean('check_open_attach', False)
+        self.probe_sense_pin = config.get('probe_sense_pin', None)
+        self.dock_sense_pin = config.get('dock_sense_pin', None)
+        self.printer.register_event_handler('klippy:ready',
+                                            self._handle_ready)
+
+        ppins = self.printer.lookup_object('pins')
+        gcode = self.printer.lookup_object('gcode')
+        gcode.register_command('GET_PROBE_STATUS',
+                                self.cmd_GET_PROBE_STATUS,
+                                self.cmd_GET_PROBE_STATUS_help)
+
+        if not any([self.check_open_attach,
+                    self.probe_sense_pin,
+                    self.dock_sense_pin]):
+            raise self.printer.config_error(HINT_VERIFICATION_ERROR.format(
+                                            aProbe.name))
+        else:
+            ehelper = PinPollingHelper(config, aProbe.query_endstop)
+
+            # Configure sense pins as endstops so they
+            # can be polled at specific times
+            def configEndstop(pin):
+                pin_params = ppins.lookup_pin(pin,
+                                              can_invert=True,
+                                              can_pullup=True)
+                mcu = pin_params['chip']
+                mcu_endstop = mcu.setup_pin('endstop', pin_params)
+                helper = PinPollingHelper(config, mcu_endstop.query_endstop)
+                return helper
+
+            probe_sense_helper = None
+            dock_sense_helper = None
+
+            if self.probe_sense_pin:
+                probe_sense_helper = configEndstop(self.probe_sense_pin)
+                self.probe_sense_pin = probe_sense_helper.query_pin
+            else:
+                self.probe_sense_pin = ehelper.query_pin_inv
+
+            if self.dock_sense_pin:
+                dock_sense_helper = configEndstop(self.dock_sense_pin)
+                self.dock_sense_pin = dock_sense_helper.query_pin
+            else:
+                if probe_sense_helper is not None:
+                    self.dock_sense_pin = probe_sense_helper.query_pin_inv
+                else:
+                    self.dock_sense_pin = ehelper.query_pin
+    cmd_GET_PROBE_STATUS_help = ("Prints the current probe state, valid probe" +
+                                 " states are UNKNOWN, ATTACHED, and DOCKED")
+    def cmd_GET_PROBE_STATUS(self, gcmd):
+        s = self.get_probe_state()
+        state = next(v for k, v
+                     in enumerate(['UNKNOWN', 'ATTACHED', 'DOCKED'])
+                     if s == k)
+        gcmd.respond_info('Probe Status: %s' % (state))
+    def _handle_ready(self):
+        self.last_verify_time = 0
+        self.last_verify_state = PROBE_UNKNOWN
+    def get_probe_state(self):
+        curtime = self.printer.get_reactor().monotonic()
+        last = self.last_verify_time
+        if (self.last_verify_state == PROBE_UNKNOWN
+            or curtime > self.last_verify_time + PROBE_VERIFY_DELAY):
+            self.last_verify_time = curtime
+            a = self.probe_sense_pin(curtime)
+            d = self.dock_sense_pin(curtime)
+            if a and not d:
+                self.last_verify_state = PROBE_ATTACHED
+            elif d and not a:
+                self.last_verify_state = PROBE_DOCKED
+            else:
+                self.last_verify_state = PROBE_UNKNOWN
+        return self.last_verify_state
+
+class DockableProbe:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.gcode = self.printer.lookup_object('gcode')
+        self.name = config.get_name()
+
+        # Configuration Options
+        self.position_endstop = config.getfloat('z_offset')
+        self.x_offset = config.getfloat('x_offset', 0.)
+        self.y_offset = config.getfloat('y_offset', 0.)
+        self.speed = config.getfloat('speed', 5.0, above=0.)
+        self.lift_speed = config.getfloat('lift_speed', self.speed, above=0.)
+        self.dock_retries = config.getint('dock_retries', 0) + 1
+        self.travel_speed = config.getfloat('travel_speed',
+                                             self.speed, above=0.)
+        self.attach_speed = config.getfloat('attach_speed',
+                                             self.travel_speed, above=0.)
+        self.detach_speed = config.getfloat('detach_speed',
+                                             self.travel_speed, above=0.)
+        self.dock_fixed_z = config.getboolean('dock_fixed_z', False)
+        self.allow_delayed_detach = config.getboolean('allow_delayed_detach',
+                                                       False)
+        self.sample_retract_dist = config.getfloat('sample_retract_dist', 2.,
+                                                   above=0.)
+
+        def parseCoord(name, expected_dims=3):
+            val = config.get(name, None)
+            error_msg = "Unable to parse {0} in {1}: {2}"
+            if not val:
+                return None
+            try:
+                vals = [float(x.strip()) for x in val.split(',')]
+            except Exception as e:
+                raise config.error(error_msg.format(name, self.name, str(e)))
+            supplied_dims = len(vals)
+            if not 2 <= supplied_dims <= expected_dims:
+                raise config.error(error_msg.format(name, self.name,
+                                   "Invalid number of coordinates"))
+            p = [None] * 3
+            p[:supplied_dims] = vals
+            return p
+
+        self.approach_position = parseCoord('approach_position')
+        self.detach_position   = parseCoord('detach_position')
+        self.dock_position     = parseCoord('dock_position')
+        self.safe_z_position   = parseCoord('safe_z_position', 2)
+
+        self.dock_angle, self.approach_distance = self._get_vector(
+                                                        self.dock_position,
+                                                        self.approach_position)
+        self.detach_angle, self.detach_distance = self._get_vector(
+                                                        self.dock_position,
+                                                        self.detach_position)
+
+        #Pins
+        ppins = self.printer.lookup_object('pins')
+        pin = config.get('pin')
+        pin_params = ppins.lookup_pin(pin, can_invert=True, can_pullup=True)
+        mcu = pin_params['chip']
+        mcu.register_config_callback(self._build_config)
+        self.mcu_endstop = mcu.setup_pin('endstop', pin_params)
+
+        # Wrappers
+        self.get_mcu              = self.mcu_endstop.get_mcu
+        self.add_stepper          = self.mcu_endstop.add_stepper
+        self.get_steppers         = self.mcu_endstop.get_steppers
+        self.home_wait            = self.mcu_endstop.home_wait
+        self.query_endstop        = self.mcu_endstop.query_endstop
+        self.finish_home_complete = self.wait_trigger_complete = None
+
+        pstate = ProbeState(config, self)
+        self.get_probe_state = pstate.get_probe_state
+
+        #Custom User Macros
+        gcode_macro = self.printer.load_object(config, 'gcode_macro')
+        self.pre_attach_gcode = gcode_macro.load_template(
+            config, 'pre_attach_gcode', '')
+        self.attach_gcode = gcode_macro.load_template(
+            config, 'attach_gcode', '')
+        self.post_attach_gcode = gcode_macro.load_template(
+            config, 'post_attach_gcode', '')
+        self.pre_detach_gcode = gcode_macro.load_template(
+            config, 'pre_detach_gcode', '')
+        self.detach_gcode = gcode_macro.load_template(
+            config, 'detach_gcode', '')
+        self.post_detach_gcode = gcode_macro.load_template(
+            config, 'post_detach_gcode', '')
+
+        #Gcode Commands
+        self.gcode.register_command('ATTACH_PROBE',
+                                    self.cmd_ATTACH_PROBE)
+        self.gcode.register_command('DETACH_PROBE',
+                                    self.cmd_DETACH_PROBE)
+
+        #Event Handlers
+        self.printer.register_event_handler('klippy:connect',
+                                            self._handle_connect)
+        self.printer.register_event_handler('klippy:ready',
+                                            self._handle_ready)
+        self.printer.register_event_handler('homing:home_rails_begin',
+                                            self._handle_homing_rails_begin)
+        self.printer.register_event_handler('homing:home_rails_end',
+                                            self._handle_homing_rails_end)
+
+    def _build_config(self):
+        kin = self.printer.lookup_object('toolhead').get_kinematics()
+        for stepper in kin.get_steppers():
+            if stepper.is_active_axis('z'):
+                self.add_stepper(stepper)
+        endstops = [es for rail in kin.rails
+                    for es, name in rail.get_endstops()]
+        if self in endstops and self.printer.lookup_object('safe_z_home', None):
+            e = ("{0}: cannot be used with safe_z_home if " +
+                "probe is used as z endstop")
+            raise self.printer.config_error(e.format(self.name))
+    def _handle_connect(self):
+        self.toolhead = self.printer.lookup_object('toolhead')
+    def _handle_ready(self):
+        self.last_z = -9999
+        self.is_z_endstop = False
+        self.multi = MULTI_OFF
+        self._last_homed = None
+        self.init_pos = None
+        kin = self.printer.lookup_object('toolhead').get_kinematics()
+        if not self.safe_z_position:
+            home_xy = [0,0]
+            for rail in kin.rails:
+                hi = rail.get_homing_info()
+                rmin, rmax = rail.get_range()
+                if hi.positive_dir:
+                    mid = (hi.position_endstop - rmin) *.5
+                else:
+                    mid = rmax * .5
+                for s in rail.get_steppers():
+                    a = s.get_name(short=True)
+                    if a == 'x':
+                        home_xy[0] = mid - self.x_offset
+                    if a == 'x':
+                        home_xy[1] = mid - self.y_offset
+                    if a == 'z':
+                        if hi.positive_dir:
+                            self.z_homes_positive = True
+                        else:
+                            self.z_homes_positive = False
+            self.safe_z_position = home_xy
+    def cmd_ATTACH_PROBE(self, gcmd):
+        return_pos = self.toolhead.get_position()
+        self.attach_probe(return_pos)
+    def cmd_DETACH_PROBE(self, gcmd):
+        return_pos = self.toolhead.get_position()
+        self.detach_probe(return_pos)
+
+    #######################################################################
+    # Functions for calculating points and moving the toolhead
+    #######################################################################
+
+    # Move the toolhead to minimum safe distance aligned with angle
+    def _move_to_vector(self, angle):
+        x, y = self._get_point_on_vector(self.dock_position[:2],
+                                         angle,
+                                         self.approach_distance)
+        self.toolhead.manual_move([x,y,None], self.travel_speed)
+
+    # Move the toolhead to angle within minimium safe distance
+    def _align_to_vector(self, angle):
+        approach = self._get_intercept(self.toolhead.get_position(),
+                                       angle + (pi/2),
+                                       self.dock_position,
+                                       angle)
+        self.toolhead.manual_move([approach[0], approach[1], None],
+                                  self.attach_speed)
+
+    # Determine toolhead distance to dock coordinates
+    def _check_distance(self, pos=None, dist=None):
+        if not pos:
+            pos = self.toolhead.get_position()
+        dock = self.dock_position
+
+        if dist > sqrt((pos[0]-dock[0])**2 +
+                       (pos[1]-dock[1])**2 ):
+            return True
+        else:
+            return False
+
+    # Find a point on a vector line at a specific distance
+    def _get_point_on_vector(self, point, angle, magnitude=1):
+        x = point[0] - magnitude * cos(angle)
+        y = point[1] - magnitude * sin(angle)
+        return (x, y)
+
+    # Locate the intersection of two vectors
+    def _get_intercept(self, point1, angle1, point2, angle2):
+        x1, y1 = point1[:2]
+        x2, y2 = self._get_point_on_vector(point1, angle1, 10.0)
+        x3, y3 = point2[:2]
+        x4, y4 = self._get_point_on_vector(point2, angle2, 10.0)
+        det1 = ((x1 * y2) - (y1 * x2))
+        det2 = ((x3 * y4) - (y3 * x4))
+        d = ((x1 - x2) * (y3 - y4)) - ((y1 - y2) * (x3 - x4))
+        x = float((det1 * (x3 - x4)) - ((x1 - x2) * det2)) / d
+        y = float((det1 * (y3 - y4)) - ((y1 - y2) * det2)) / d
+        return (x, y)
+
+    # Determine the vector of two points
+    def _get_vector(self, point1, point2):
+        x1, y1 = point1[:2]
+        x2, y2 = point2[:2]
+        magnitude = sqrt((x2-x1)**2 + (y2-y1)**2 )
+        angle = atan2(y2-y1, x2-x1) + pi
+
+        return angle, magnitude
+
+    # Align z axis to prevent crashes
+    def _align_z(self):
+        if not self._last_homed:
+            curtime = self.printer.get_reactor().monotonic()
+            homed_axes = self.toolhead.get_status(curtime)['homed_axes']
+            self._last_homed = homed_axes
+        if self.dock_fixed_z:
+            dock_z = (self.dock_position[2] +
+                      self.position_endstop +
+                      self.sample_retract_dist)
+            tpos = self.toolhead.get_position()
+            if 'z' in self._last_homed:
+                if tpos[2] < dock_z:
+                    self.toolhead.set_position(tpos)
+                    self.toolhead.manual_move(
+                        [None, None, dock_z],
+                        self.lift_speed)
+            else:
+                self._force_z_hop()
+        else:
+            if 'z' in self._last_homed:
+                self.toolhead.manual_move(
+                                        [None, None, self.dock_position[2]],
+                                        self.lift_speed)
+            else:
+                raise self.printer.command_error(
+                    "Cannot attach/detach probe, must home Z axis first")
+
+    # Hop z and return to un-homed state
+    def _force_z_hop(self):
+        this_z = self.toolhead.get_position()[2]
+        if self.last_z != this_z:
+            tpos = self.toolhead.get_position()
+            dock_z = self.dock_position[2] + self.position_endstop
+            self.toolhead.set_position([tpos[0], tpos[1], 0.0, tpos[3]],
+                                homing_axes=[2])
+            self.toolhead.manual_move(
+                [None, None, dock_z],
+                self.lift_speed)
+            kin = self.toolhead.get_kinematics()
+            kin.note_z_not_homed()
+            self.last_z = self.toolhead.get_position()[2]
+
+    # Attaching actions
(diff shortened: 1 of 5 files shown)


cloakedcode:
To make changes as a result of this PR process visible (e.g. fixing whitespace), I won't rewrite the git commits unless requested. If I should be doing something different, please let me know.

kdb424:
I've been running this for 4 days, and I've found it much easier to configure, more concise, better to control, and more reliable than the previous offering. I will continue to test any changes added and report back if I have any issues, but this is a wonderful addition for my printer.

kdb424:
I've run into a snag with this module. It doesn't fully act like a probe, but an endstop. There is no "probe" in the endstop checks, and when trying to mix this with a sexbolt or other nozzle Z endstop, this seems to not be triggering no matter what I try. Looks like there was an oversight here. Using the probe as the virtual Z endstop is working fine however.

kdb424:
There is some error in this module that needs updated. Auto Z works with Mental's module, but produces an error with this one. https://github.com/protoloft/klipper_z_calibration/issues/105

BelgarionNL:
feel free to ask me questions. I can reproduce the error every single time. once I paste in the newer dockable_probe.py

kdb424:
Captured a partial log to help make this easier to debug https://pastebin.com/TPwAd247. I haven't pinpointed what the change was in this module as opposed to Mental's, but that module works correctly, this module causes a firmware crash from that, which points at the error being here.

cloakedcode:
I'm afraid the cause of this bug isn't obvious and I'm away from home and printers for a week to debug.

I suspect it's something to do with the probe state query/changes occurring but that shouldn't interact with the stepper commands.

Does the issue appear with both modules loaded and running a normal 'PROBE' command? What about more complex probing like 'BED_MESH_CALIBRATE'? I'm trying to determine if it's only triggered by the exact series of commands in the Z calibration gcode or if having both modules loaded can trigger the bug.

I need more logs. 😋 Please include the full log and complete repro steps, I'll be back in front of a printer eventually to repro locally.

kdb424:
I'll get you a full log tonight. I can run bed mesh, QGL, Z tilt, ect all fine. It triggers when you touch the probe to the bed after measuring the nozzle and the probe body. I'm not getting any errors outside of that exact situation. Repro steps is simply to run a `CALIBRATE_Z` with the module loaded. It will cause a panic shown in the previous log the moment the probe touches the bed.

cloakedcode:
While you're in there troubleshooting, can you try two things for me?

1. Does the same issue occur when running `PROBE_Z_ACCURACY`?
2. Try changing line 121 ish in `dockable_probe.py` to not use the event reactor time in querying the endstop but last movement time, like this:

```
curtime = self.printer.lookup_object('toolhead').get_last_move_time()
```

Restart the Klipper service and try the Z calibration again.

I guess that would create two log files and both might be helpful in shedding light on where the bug is hiding.

Let me know how that goes! Thanks for your help.

kdb424:
>     1. Does the same issue occur when running `PROBE_Z_ACCURACY`?
No, it probes the nozzle as the nozzle endstop is the Z endstop, not the probe.
 
>     2. Try changing line 121 ish in `dockable_probe.py` to not use the event reactor time in querying the endstop but last movement time, like this:
> 
> 
> ```
> curtime = self.printer.lookup_object('toolhead').get_last_move_time()
> ```
This had no effect on logs as best as I can tell. They look nearly identical.

As promised, full log here.
https://gist.github.com/kdb424/cd9bd3bb45a1e62aedb753cbaca32854

cloakedcode:
I'm not sure but this might be caused by dockable_probe.py lines 587-593. Try deleting them entirely and test whether the probe behaves normally (regular probe movement and bed mesh). If all is well try Z calibration.

My suspicion is there's an ordering conflict with the endstop triggering and the attempt to query the endstop state which creates this opaque "invalid sequence" error (that doesn't point at what's wrong with the sequence).

kdb424:
That won't even let it start probing. No Z movement before panic. No usable log here because that isn't relevant to auto-z as even trying to QGL or anything probe related dies.

cloakedcode:
@kdb424 Can you put this code through its paces? Hopefully it works with and without auto Z calibration.

https://github.com/cloakedcode/klipper/commit/19139f1d10b08fabbacd0b7bb163f5ae519d23a7

churls5495:
@cloakedcode, could you elaborate on how this works with a bed slinger? I'm currently using a [custom module](https://github.com/churls5495/Annex-Engineering_User_Mods/blob/main/Extruders/Sherpa_Mini/Toolheads/Churls-Stiffy_E3/xz_dockable_probe.py) that requires an endstop to first locate the dock before attaching probe. I'm not seeing how your module homes Z before picking up the probe. More info [here](https://github.com/churls5495/Annex-Engineering_User_Mods/tree/main/Extruders/Sherpa_Mini/Toolheads/Churls-Stiffy_E3).

cloakedcode:
tl;dr - You have there a setup that is just different enough that you'll need to do a little extra configuration to make it work but that's by design.

(A note about probe support: Not all probe setups are meant to be fully supported out of the box, there are simply too many variations and new ones come along regularly. If this gets merged, I foresee probe designers documenting the necessary Klipper config to make their designs work, similar to below, rather than custom Klipper modules/extensions.)

That looks to be a QuickDraw probe/dock mounted to an Ender 3. Klicky and QuickDraw probes operate with the same attaching/detaching procedures. For attach: move near the dock, move to the probe, move away from the dock with the probe. For detach: move near the dock, move to the dock, move away from the dock leaving the probe behind.

The difference with your setup vs a CoreXY machine is that yours has the dock mounted at a fixed Z height (i.e. Z = 0). That's the [first position example in the docs](https://github.com/Klipper3d/klipper/blob/82e3db7547d0a81e6dd9e9fe8fa8d6ad0c6f04a6/docs/Dockable_Probe.md#position-examples). Also read the explanation of the [position config options](https://github.com/Klipper3d/klipper/blob/82e3db7547d0a81e6dd9e9fe8fa8d6ad0c6f04a6/docs/Dockable_Probe.md#attaching-and-detaching-positions). If Z is required for the dock, Z homing (`G28`) must happen before attaching the probe (`ATTACH_PROBE`) or an error will be raised. If the correct options are specified, a typical usage would not require any probe-specific gcode commands to be run (e.g bed mesh: `G28` followed immediately by `BED_MESH_CALIBRATE`) as the probe will be auto-attached if needed. In other words, if you're overriding built-in Klipper gcodes to wrap them in "attach probe" and "detach probe", that's unnecessary with this module.

The subtlety with your setup is the dock is mounted backwards compared to a Klicky dock where the probe is pulled out of the dock towards the bed. This complicates things slightly but easily doable as the [movement commands are meant to be overridden](https://github.com/Klipper3d/klipper/blob/82e3db7547d0a81e6dd9e9fe8fa8d6ad0c6f04a6/docs/Dockable_Probe.md#individual-movements) in this scenario ([similarly for Euclid](https://klipper.discourse.group/t/reviving-the-dockable-probe/4256/16)).

The attach movements for your probe can be, looking at the printer from the front:
```
1/5 > 2a
^     |
4a  < 3
      ^ dock sits here
```
detach:
```
1
|
2d  >  3  >  4d
       ^ dock sits here
```

* 1 and 5 are your `approach_position`
* 3 is your `dock_position`
* 4d is your `detach_position`
* 4a and 2d are identical
* 4a/2d and 2a don't have config options but are instead specified by overriding the gcode movements

This is similar to a Euclid probe [as seen in the last positioning example](https://github.com/Klipper3d/klipper/blob/82e3db7547d0a81e6dd9e9fe8fa8d6ad0c6f04a6/docs/Dockable_Probe.md#position-examples).

All that's needed now is to implement the additional movements:

```
[gcode_macro MOVE_TO_APPROACH_PROBE]
rename_existing: MOVE_TO_APPROACH_PROBE_BASE
gcode:
    MOVE_TO_APPROACH_PROBE_BASE
    G1 <2a>

[gcode_macro MOVE_TO_EXTRACT_PROBE]
gcode:
    G1 <4a>
    MOVE_TO_APPROACH_PROBE_BASE

[gcode_macro MOVE_TO_INSERT_PROBE]
gcode:
    G1 <2d>
```

churls5495:
> If Z is required for the dock, Z homing (`G28`) must happen before attaching the probe (`ATTACH_PROBE`) or an error will be raised. 

This is the part I'm trying to understand. How is Z homed _before_ the probe is attached? I assume it's coded in your module; I'd just like to understand the details.

cloakedcode:
Z homing is not handled by this module. The standard mechanisms in Klipper for homing should be used. For example, if the printer setup requires the XYZ axises be homed in a specific order, the `homing_override` config section/module can be used.

IIUC for your setup the Z endstop is configured as a standard Z endstop (i.e. defined in a stepper_z section). This is compatible with the module but homing is not implemented in this module.

cloakedcode:
@kdb424 / @BelgarionNL Have you tried the special branch/fix linked above with the auto Z calibration? I'm afraid I cannot test it as I don't have a printer with a nozzle probe. AFAIK this is the last outstanding issue with the module.

kdb424:
I no longer have a printer with both a nozzle Z and a dockable probe, so I won't be of much use on that front unfortunately. Sorry that died before I helped you get it sorted.

Hobbss27:
> @kdb424 / @BelgarionNL Have you tried the special branch/fix linked above with the auto Z calibration? I'm afraid I cannot test it as I don't have a printer with a nozzle probe. AFAIK this is the last outstanding issue with the module.

im running both currently if you need imput. im using the base klicky config with autoz from protoloft.

cloakedcode:
@Hobbss27 Yes please! Please run this [special branch code](https://github.com/cloakedcode/klipper/blob/trigger-rework/klippy/extras/dockable_probe.py) everywhere you can (i.e. anywhere you've got a klicky/dockable probe). It should work with and without protoloft's auto Z module.

Thank you for volunteering!

kdb424 (approved):

cloakedcode:
I have merged in the changes from the separate branch to make it compatible with the auto Z calibration module. In addition, I've added a check for dockable_probe/virtual Z endstop misconfiguration and removed mention of missing options from the docs.

cloakedcode:
At this point I'm not aware of any outstanding issues or incompatibilities with other modules or hardware setups. I have reviewed this PR to the best of my abilities and think it meets the contributing guidelines/review process.

KevinOConnor:
Thanks.  Alas, I haven't been able to get time to look at this.  I'll be travelling for a couple of weeks, but will hopefully get a chance to comment when I return.

Cheers,
-Kevin

vospascal:
i have also tested it it works fine 👍🏻 i got a corexy tripple z with still gantry where the bed moves down and gantry moves only x an y

pbsuper:
config
Error
@cloakedcode a couple users have gotten this issue with the latest version. it says to not use Z coordinates if a virtual endstop is given but when you remove the Z coordinate it still gives the error. previous version didn't have the issue.

pbsuper:
also might wanna add in the documentation that Homing Override is mandatory if not using a Z endstop. i was trying to use Safe_Z_Home and it went through the homing sequence and then once the probe was put back in the dock it threw a error. I tried if with G28 first and then i did all the axis seperately that also didn't work. Then i switch over from Safe_Z_Home to Homing_Overwrite and it worked.
image

cloakedcode:
For the virtual probe issue, please do a full Klipper service restart, not a firmware restart.

If this issue persists, and for the homing_override issue, please upload your klippy.log for me to examine.

cloakedcode:
@pbsuper / @vospascal Thank you for the reports! I have done more testing of dockable_probe + virtual Z endstop + safe_z_home (+ z-hop) and committed two changes (12c99a2 and ce73f72) that resolve the issues, in my testing. Please test the latest code and let me know how you fare.

vospascal:
@cloakedcode 
i think this options should be added looks like it has been removed i know there lot of users also using this feature

```
pre_attach_gcode: SET_SERVO servo=probe_dock Angle=90
post_attach_gcode:SET_SERVO servo=probe_dock Angle=0
pre_detach_gcode: SET_SERVO servo=probe_dock Angle=90
post_detach_gcode: SET_SERVO servo=probe_dock Angle=0
```

cloakedcode:
I removed those as they became redundant when the movements got g-codes that could be overridden. The trade off of moving that config option functionality to overriding a macro in exchange for removing the option seemed like a good one as the config was originally quite complex. See https://github.com/Klipper3d/klipper/pull/6247/files#diff-2930d37d8143c91ff5f6a8e550831e09400c2107d71767e809d9fbe535edb2e5R283

For example, overriding MOVE_TO_APPROACH_PROBE could move the servo for deployment, while MOVE_TO_DETACH_PROBE could have a servo retraction command.

cloakedcode:
@KevinOConnor thanks for reopening! With the above issue having been fixed as indicated by the "thumbs up" reactions and lack of further reports, I believe we're ready to proceed with review and discussion. Let me know if you have questions!

cloakedcode:
Is Kevin the only reviewer qualified to look this over or is there someone else available? Hopefully this responsibility doesn't solely rest on Kevin, that's a lot of work!

yantz74:
Transition from mental's version was straightforward. Removed some options, and it worked as intended.

Custom printer
Cross gantry
Triple belted z
Z virtual endstop

Yantz

cloakedcode:
This PR is now among the oldest open PRs. What can I do to get this merged? IIUC it needs a reviewer and those are all volunteers.

1. Is there a recurring gathering to review PRs or purely on a as-available basis?
2. How do I solicit a reviewer? Is there a better tactic than posting on the Discourse with subject: "looking for a PR reviewer"?

Not sure what to do here.

Sineos:
It is also a matter close to my heart to see this feature merged. I do not use a dockable probe, but then these hundreds of lines of macros disappear from the configs 😇

novaplusplus:
I am surprised this hasn't been merged yet honestly! I've been running this for a couple months now on my own printer, it seems flexible enough to adapt to most designs, including to my custom probe (deployment arm + lower to Z=0.5 to retract, someday I'll publish the thing properly). During setup I had a few crashes with the deployer but that was 100% my fault. It's a lot better than gigantic bowl of macro soup like klicky etc requires.

Though I did have to do some unorthodox stuff to get mine to work... I do feel like being able to replace the entire dock/undock sequence with custom gcode (instead of doing it in the individual sections) could be a useful feature, but not anything that would stop it being merged really.

For reference this is what my config looks like for my setup (there is also a fair chance I'm talking nonsense and am missing something super obvious)

```
[dockable_probe]
pin: !head:PD4
x_offset: 0
y_offset: 36
speed: 10
lift_speed: 12
travel_speed: 180
attach_speed: 180
detach_speed: 180
check_open_attach: True
dock_position: 235, 202.5
approach_position: 180, 202.5
detach_position: 0, 0 # not used
z_hop: 15

# We stow the probe by dropping Z below the pickup threshold so some macros are just skipped
[gcode_macro _PROBING_VARS]
variable_dock_return_skip: 0
gcode:

[gcode_macro MOVE_TO_INSERT_PROBE]
rename_existing: _MOVE_TO_INSERT_PROBE
gcode:
    SET_GCODE_VARIABLE MACRO=_PROBING_VARS VARIABLE=dock_return_skip VALUE=1

[gcode_macro MOVE_TO_DOCK_PROBE]
rename_existing: _MOVE_TO_DOCK_PROBE
gcode:
    # Only do dock move for probe approach
    {% if printer["gcode_macro _PROBING_VARS"].dock_return_skip == 0 %}
        _MOVE_TO_DOCK_PROBE
    {% endif %}

[gcode_macro MOVE_TO_DETACH_PROBE]
rename_existing: _MOVE_TO_DETACH_PROBE
gcode:
    SET_GCODE_VARIABLE MACRO=_PROBING_VARS VARIABLE=dock_return_skip VALUE=0
    G1 Z0.5
    G1 Z10
```

BillyNate:
> Though I did have to do some unorthodox stuff to get mine to work... I do feel like being able to replace the entire dock/undock sequence with custom gcode (instead of doing it in the individual sections) could be a useful feature, but not anything that would stop it being merged really.

Your situation sounds quite familiar, so I made changes to the dockable probe code to solve it: https://github.com/cloakedcode/klipper/pull/1
Replacing the coordinates in the `_route` settings of my PR for gcode support would make it even more versatile, but probably harder to add some kind of processing to the coordinates if we would want that later on.

cloakedcode:
> deployment arm + lower to Z=0.5 to retract

> (there is also a fair chance I'm talking nonsense and am missing something super obvious)

To get a reassuring second set of eyes on your solution, can you explain more about the dock and probe mount? Photos would be ideal. I don't understand the process for this detaching, in particular why you only move Z during detaching and not on attaching. Without movement in the INSERT/DOCK g-codes, the toolhead is probably not near the dock but that doesn't seem to be an issue for you. I'm mystified. This is sounding more and more atypical the more I think about it. 😅

novaplusplus:
> > deployment arm + lower to Z=0.5 to retract
> 
> > (there is also a fair chance I'm talking nonsense and am missing something super obvious)
> 
> To get a reassuring second set of eyes on your solution, can you explain more about the dock and probe mount? Photos would be ideal. I don't understand the process for this detaching, in particular why you only move Z during detaching and not on attaching. Without movement in the INSERT/DOCK g-codes, the toolhead is probably not near the dock but that doesn't seem to be an issue for you. I'm mystified. This is sounding more and more atypical the more I think about it. 😅

Oh, it's not all that complicated - there's a little arm that helps release the probe pin from a magnet, then it simply lowers below a certain Z height to let the magnet tug it up and out of the way again. It's proven quite reliable so far. Like I said, I need to properly publish the design at some point...

JonasVerhofste:
The module/plugin triggers an error on my end when trying to quad_gantry_level a voron 2.4. The probe correctly attaches and detaches. (Though I am somehow missing the virtual endstop in the mainsail UI, but not sure if that's by design?)
The printer is fully homed (z-axis is homed through a separate endstop) before trying to do a QGL. Without the module, QGL runs fine.

I'm running the current HEAD of the master branch (18de421c4acbdb00946df35dfb412dedb159abca as of writing), flashed on toolhead, mainboard and rpi/linux mcu.

I'm not entirely sure what's going on, but it's kind of late in my timezone and I'm probably missing something. I'll try to have another look tomorrow with a fresh head.

```
WARNING:root:Internal error on command:"QUAD_GANTRY_LEVEL"
ERROR:root:Internal Error on WebRequest: gcode/script
  File "/opt/klipper/klippy/webhooks.py", line 253, in _process_request
    func(web_request)
  File "/opt/klipper/klippy/webhooks.py", line 433, in _handle_script
    self.gcode.run_script(web_request.get_str('script'))
  File "/opt/klipper/klippy/gcode.py", line 229, in run_script
    self._process_commands(script.split('\n'), need_ack=False)
  File "/opt/klipper/klippy/gcode.py", line 211, in _process_commands
    handler(gcmd)
  File "/opt/klipper/klippy/gcode.py", line 137, in <lambda>
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/klipper/klippy/extras/quad_gantry_level.py", line 53, in cmd_QUAD_GANTRY_LEVEL
  File "/opt/klipper/klippy/extras/probe.py", line 442, in start_probe
    pos = probe.run_probe(gcmd)
          ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/klipper/klippy/extras/probe.py", line 175, in run_probe
    pos = self._probe(speed)
          ^^^^^^^^^^^^^^^^^^
  File "/opt/klipper/klippy/extras/probe.py", line 124, in _probe
    epos = phoming.probing_move(self.mcu_probe, pos, speed)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/klipper/klippy/extras/homing.py", line 246, in probing_move
    epos = hmove.homing_move(pos, speed, probe_pos=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/klipper/klippy/extras/homing.py", line 85, in homing_move
    wait = mcu_endstop.home_start(print_time, ENDSTOP_SAMPLE_TIME,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/klipper/klippy/extras/dockable_probe.py", line 619, in home_start
    self.finish_home_complete = self.mcu_endstop.home_start(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/klipper/klippy/mcu.py", line 294, in home_start
    trsync.start(print_time, report_offset,
  File "/opt/klipper/klippy/mcu.py", line 202, in start
    ffi_lib.trdispatch_mcu_setup(self._trdispatch_mcu, clock, expire_clock,
TypeError: initializer for ctype 'struct trdispatch_mcu *' must be a cdata pointer, not NoneType
```

Relevant parts of my config:
```
[dockable_probe]
approach_position: -2,265
dock_position: -2,290
detach_position: 25,290
z_hop: 15

auto_attach_detach: True
check_open_attach: True

travel_speed: 400
attach_speed: 50
detach_speed: 50

pin: toolhead:PA2
x_offset: 1
y_offset: 35
z_offset: 3.980
speed: 8
samples: 4
samples_result: median # Median of even number means average of middle two
sample_retract_dist: 3.0
samples_tolerance: 0.003
samples_tolerance_retries: 3

[quad_gantry_level]
gantry_corners:
 -60,-10
 360,370
points:
 30,0
 30,240
 270,240
 270,0
speed: 200
horizontal_move_z: 15
retries: 5
retry_tolerance: 0.0075
max_adjust: 15
```

JonasVerhofste:
Hmm, same error when running from the HEAD of the PR-branch. So it's either something with my config, or something with my system. Using a "normal" `[probe]` with the dockable_probe module installed does work, it's only when the `PROBE` command is issued with a `[dockable_probe]` that the error is triggered.

ProfDrYoMan:
Works wonderfully for me. Hope it gets merged „soon“. 

One comment. On bed slingers it would be nice if there is even no need to set a y axis position since this only creates useless moves of the bed, @cloakedcode. For undock x,y it could be x,-.

novaplusplus:
As has been mentioned having more direct control over the dock/undock procedure and navigation would help the bedslinger situation too... in general being able to use the module as a tool to keep track of the probe deployment/dock/etc status, and having the option to directly write dock/undock macros ourselves (not using the navigation built in) would be really useful for more unusual/uncommon use cases. I don't consider that something stopping it from being merged though. so maybe in 2037 after it's integrated things like that can be added

ProfDrYoMan:
I just figured I can do what I want by overwriting the given macros properly. Will try tomorrow. Now I don’t see the need for for a config change although it would be nice.

ProfDrYoMan:
As mentioned above:

```
[gcode_macro MOVE_TO_APPROACH_PROBE]
rename_existing: MOVE_TO_APPROACH_PROBE_DEFAULT
description: This command will move the toolhead to the approach_position.
gcode:
 {% set x, y = printer.configfile.settings.dockable_probe.approach_position.split(',') %}
 G1 X{x} F{printer.configfile.settings.dockable_probe.travel_speed | float * 60}

[gcode_macro MOVE_TO_DOCK_PROBE]
description: This command will move the toolhead to the dock_position.
rename_existing: MOVE_TO_DOCK_PROBE_DEFAULT
gcode:
 {% set x, y = printer.configfile.settings.dockable_probe.dock_position.split(',') %}
 G1 X{x} F{printer.configfile.settings.dockable_probe.attach_speed | float * 60}

[gcode_macro MOVE_TO_DETACH_PROBE]
description: This command will move the toolhead to the detach_position.
rename_existing: MOVE_TO_DETACH_PROBE_DEFAULT
gcode:
 {% set x, y, z = printer.configfile.settings.dockable_probe.detach_position.split(',') %}
 G1 Z{z} F{printer.configfile.settings.dockable_probe.detach_speed | float * 60}
 G1 X{x} F{printer.configfile.settings.dockable_probe.detach_speed | float * 60}
```

This should do the trick to not move Y anymore because not needed.

ammmze:
Huge thanks to @cloakedcode for putting this together. I really hope we can finally get this thing merged soon.

I'm only just starting to test it and hit a snag...it does look like there was a [change](https://github.com/Klipper3d/klipper/commit/acdf8bb1080ab829c6e8092dadf6d488aa416f77) in the last few days that breaks it though.

```
  {omitted for brevity}
  File "/home/pi/klipper/klippy/extras/bed_mesh.py", line 747, in cmd_BED_MESH_CALIBRATE
  File "/home/pi/klipper/klippy/extras/probe.py", line 444, in start_probe
    pos = probe.run_probe(gcmd)
  File "/home/pi/klipper/klippy/extras/probe.py", line 174, in run_probe
    pos = self._probe(speed)
  File "/home/pi/klipper/klippy/extras/probe.py", line 123, in _probe
    epos = self.mcu_probe.probing_move(pos, speed)
AttributeError: 'DockableProbe' object has no attribute 'probing_move'
```

I also have a suggestion...

I was working through getting this integrated into my print start macro working from the example:

```
SET_DOCKABLE_PROBE AUTO_ATTACH_DETACH=0
G28
ATTACH_PROBE                             # Explicitly attach the probe
QUAD_GANTRY_LEVEL                        # Tram the gantry parallel to the bed
BED_MESH_CALIBRATE                       # Create a bed mesh
DETACH_PROBE                             # Manually detach the probe
SET_DOCKABLE_PROBE AUTO_ATTACH_DETACH=1  # Make sure the probe is attached in future
```

However, because I am using a virtual z endstop (i.e. the probe), that G28 can't home all axis. So instead I have to home just x and y, and then since i can't home z, I am having to manually move to my "safe_z_home" to ensure I start in a safe position before picking up the probe. In addition, I couldn't figure out how to get the safe z home configuration to re-use it, so instead i've just copied it. But if I ever have to change it, I have to remember to update both places now.

So what I would suggest is being able to turn off auto detach and auto attach separately, while still allowing setting both with `AUTO_ATTACH_DETACH`. For example:

```
SET_DOCKABLE_PROBE AUTO_DETACH=0         # Disable auto detach, so it remains attached after homing
G28                                      # Home all axis (this will also attach the probe, but won't detach)
QUAD_GANTRY_LEVEL                        # Tram the gantry parallel to the bed
BED_MESH_CALIBRATE                       # Create a bed mesh
DETACH_PROBE                             # Manually detach the probe
SET_DOCKABLE_PROBE AUTO_ATTACH_DETACH=1  # Make sure the probe is attached in future
```

klippy/extras/dockable_probe.py
```diff
@@ -0,0 +1,634 @@
...
+    def multi_probe_end(self):
+        self.multi = MULTI_OFF
+
+        return_pos = self.toolhead.get_position()
+        # Move away from the bed to ensure the probe isn't triggered,
+        # preventing detaching in the event there's no probe/dock sensor.
+        self.toolhead.manual_move([None, None, return_pos[2]+2],
+                                  self.travel_speed)
+        self.auto_detach_probe(return_pos)
+
```
dim1triy, line 599: please add for fix problem after last update https://github.com/Klipper3d/klipper/commit/acdf8bb1080ab829c6e8092dadf6d488aa416f77

`
def probing_move(self, pos, speed):
        phoming = self.printer.lookup_object('homing')
        return phoming.probing_move(self, pos, speed)
`

cloakedcode:
@ammmze thanks for the kinds words. I will address that breaking update when I've got some free time. Frustrating that it comes now while this PR has been open for almost a year.

If you read through the PR comments and the Discourse thread you'll see there are many ideas for tweaks to the process, each to alleviate tiny pain points (relative to this remaining not in mainline Klipper). I've resisted changing the code in hopes that its stability and simplicity would win it faster review and adoption into the code base. The jury's still out on if that plan is working.

For your specific issue, rather than changing the module to add more knobs and twiddles (that a hapless user could accidentally damage their machine with), what do you think about forgoing the extra time savings of that first detach/attach and instead putting the G28 call before disabling the auto-attach-detach? Alternatively, "homing_override" will let you control the order of the homing axes and allow for inserting an explicit "ATTACH_PROBE" before homing Z (the explicit call doesn't respect the "auto" setting).

ammmze:
> If you read through the PR comments and the Discourse thread you'll see there are many ideas for tweaks to the process, each to alleviate tiny pain points (relative to this remaining not in mainline Klipper). I've resisted changing the code in hopes that its stability and simplicity would win it faster review and adoption into the code base.

👍🏻 Very good point and makes perfect sense.

ProfDrYoMan:
Hi @cloakedcode, thanks for your effort. Using it since a while now. Excellent. Also flexible enough, I think. 

Is there any way to get movement into the PR merge. I was about to send a note to Kevin, but then realized that pushing will not help and did not. 

Other stuff like the new Eddy probes get in fast. :/

cloakedcode:
@ammmze / @dim1triy / @ProfDrYoMan Added a patch to address that recent Klipper change.

@ProfDrYoMan I don't know what the delay is here. The code has been stable for months and used widely.

If anyone out there has an in with Kevin or another maintainer, please let me know! Or point them to this PR. AFAIK the only hold up is having a maintainer to approve and merge.

mental405:
Lemme get out the WD40.

ProfDrYoMan:
I noticed a small issue.
Using manual z-probe calibration and then in the end accepting the measured height leads into an error since there is no [probe] section anymore. 
Also manually tuning z-offset and the saving throws the same type of error.

cloakedcode:
@ProfDrYoMan What web UI are you using?

I just double-checked on Fluidd v1.28.1 and pressing the save button after adjusting the Z offset gives me a menu and picking "Z_OFFSET_APPLY_PROBE" puts the new offset under a "dockable_probe" section. All that to say, "works for me".

ProfDrYoMan:
Oh. I take the AR to check again. It’s also fluidd „latest“.

ProfDrYoMan:
My problem. I did put the [dockable_probe] section completely into an include config file. I do not like to mess with the original printer.cfg to much. This prevented saving.

Having at least the section header and the z_offet config in printer.cfg solved the isssue.

=> All good. Please merge! :)

KevinOConnor:
Thanks for working on this.  I see how this functionality can be useful, but I would prefer to hold off on merging functionality like this for several more months.

I understand the functionality of having a module like this - it allows users a more standardized way to handle dockable probes and it generally improves the error handling for them.  Right now, users are accomplishing the same end-goal using a variety of macros (often customized) and those macros can be complex and error prone.

However, the core functionality of this new module can be accomplished by macros today.  Even if we did merge this, many users would continue using their existing macros.  To wit, the module isn't adding new fundamental capabilities, it is mostly attempting to unify an existing capability in a way that does not require as many complex macros.

My main concern is that I'd prefer to try to tackle the generic problem of "gcode macro scalability" before merging individual gcode macro scalability improvements.  I've written a little bit about this at https://klipper.discourse.group/t/possible-klipper-plugins-instead-of-macros/9819 .  In particular, I'm concerned the more we add "macro-able" functionality to core Klipper the larger the ongoing load will be on reviewers and I fear that load is already too high today.

Thanks again for working on this.  I do agree we ultimately want to improve support for dockable probes.
-Kevin

novaplusplus:
> Since starting on this module, I've noticed it comes up on Discord or other and is followed by "we'll never have that in Klipper" or some other discouraged and negative view on Klipper's evolution.

To be fully blunt, that's how I've felt for a while. Convenience/simplification/ease of use features in general just feel like such a low priority and I've seen a lot people elsewhere (whether or not involved with development) just immediately crapping on any suggestion to improve those areas.

Macros for this sort of thing, or anything complex, is such a hacky nightmare and feels far more potentially failure prone than having a proper module done. Went through this myself with the orbiter filament sensor, had its internal state machine stuff fully lock up my printer more than once until I just gave up. Things like KAMP could be made into features of the existing modules among many other examples, overriding built-in commands with macros just makes my skin crawl. "You don't need a new feature added, just use macros" feels like "You don't need a bicycle or a car, you can just walk".

I know I'm just a random shmuck who hasn't directly contributed to the project so I don't exactly have much of a leg to stand on but I just really dont agree with that way of thinking about it at all.

ammmze:
While I would really like to see this integrated into klipper, I suppose a middle ground could be making it easier to extend klipper functionality. Something similar to [HACS](https://hacs.xyz/) if you are familiar with home assistant or [external components](https://esphome.io/components/external_components.html?highlight=externals) in esphome. It provides a way to easily install external components, so that maintenance of those can be done separately by a subset of the overall community.

I would love to have a way to do this declaratively instead of having to install from the shell. So instead of having to manually clone a 3rd party repo, run the installer which then creates symlinks it would cool if we could declaratively do it:

```ini
[plugin dockable_probe]
source: https://github.com/cloakedcode/klipper
# optionally configure a branch other than the default branch
branch: work-annex-probe
 # optionally specify an order to control the order plugins are layered in
order: 0
```

And perhaps we could expect a plugin repo to contain a manifest file called `klipper-plugin.yaml` that looks something like the following. And klipper upon startup ensure those are cloned and at the requested branch and then setup the symlinks listed in the manifest (or perhaps even change how files are loaded to look in the files listed in the manifest, allowing the klipper working dir to remain clean).

```yaml
name: dockable_probe
files:
 - src: klippy/extras/dockable_probe.py
   dest: klippy/extras/dockable_probe.py # perhaps default dest == src?
```

kdb424:
I was one of the first to test this, and I'm still using it. While I would absolutely hate for this to not upstream, I will continue to track the progress and continue to run this. Macros are absolutely horrid to use, and feel like hacks. It's an absolute shame to not have this in klipper proper.

novaplusplus:
> Macros are absolutely horrid to use, and feel like hacks.

I really would like to see from all this is a more user friendly and smooth macro/scripting system of some kind if new "macro-able" modules aren't on the table. I avoid using remotely complicated macros if I can at all help it because it always feels like such a kludge to do so, and chasing bugs in macros sucks big time...

BillyNate:
[My PR](https://github.com/cloakedcode/klipper/pull/1) from September replaces the `approach_position`/`detach_position`/`dock_position` and gcode usage for two concise `route` settings. How about that?

ProfDrYoMan:
Hi @cloakedcode,

latest klipper generates

```
  File "/home/mks/klipper/klippy/klippy.py", line 175, in _connect
    self._read_config()
  File "/home/mks/klipper/klippy/klippy.py", line 141, in _read_config
    self.load_object(config, section_config.get_name(), None)
  File "/home/mks/klipper/klippy/klippy.py", line 130, in load_object
    self.objects[section] = init_func(config.getsection(section))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mks/klipper/klippy/extras/dockable_probe.py", line 637, in load_config
    config.get_printer().add_object('probe', probe.PrinterProbe(config, msp))
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: PrinterProbe.__init__() takes 2 positional arguments but 3 were given
```
with this PR.

ProfDrYoMan:
Last working Klipper is g6cd17420, if this helps to nail down the problem. Not in my capability range range to fix this.

vospascal:
https://github.com/Klipper3d/klipper/commit/d4bae4dffe8d3996ed1b1dd8f5230a062ae3d033 you can see here the sigiture fo PrinterProbe has changed..

vospascal:
> Last working Klipper is g6cd17420, if this helps to nail down the problem. Not in my capability range to fix this.

i think its 6cd17420 not g6cd17420 ;)

ProfDrYoMan:
Thanks @cloakedcode. Printer is in maintenance mode. Mode take a while. I hope somebody else is reporting positive feedback.

53Aries:
I just want to add in, I've been using various Klicky probes on various printers for years now. It's ALWAYS such a huge headache to get it setup and working right and usually involves reprinting parts that get broken due to Klicky macros causing collisions. I had been using dockable_probe.py from Danger Klipper, pasted into my mainline Klipper extras for awhile but after updating Klipper tonight the dockable_probe.py module broke. So back to ANNOYING and cumbersome Klicky macros. I'm going to try this branch out but I really really wish it was just integral to Klipper and I didn't have to mess around so much. Even if I could drop a single module into mainline, like I was doing with dockable_probe, I would be happy.

marbocub:
I have used this module for more than half a year on a CoreXY and a Cartesian. I report the results as they work.

* CoreXY: stable, works excellent, very easy to setup since in case of using a standard fixed XY position dock.
* Cartesian: stable, works great, but it took a bit time to setup since I used an fixed XYZ position dock, a hardware Z limit switch for approaching the dock, [homing_override] in config and override five MOVE_TO_*_PROBE macros. on the other hand, the names of macros that should be overridden are good and easy to understand.

f1nch87:
Hey,
i just revived my nearly two years not printing printer.
I have a dockable quickdraw with servo mount.
How can i implement the gcode for moving the servo before attaching/detaching?

Im sorry maybe im too stupid but atm i dont get it.

Thx

ProfDrYoMan:
> @ProfDrYoMan / @vospascal I've pushed an update. Let me know how it goes.

Hi. Seems it’s broken again. At least it seems not to merge into lastest.

ProfDrYoMan:
Hi again, @cloakedcode it seems one of the latest updates in klipper generates issues.

Internal error during connect: ProbeSessionHelper.__init__() missing 1 required positional argument: 'start_session_cb'

ProfDrYoMan:
Shall I wait for an update to fix the build check fail? 

It seems this is only cosmetic.

PR #6248, opened 2023-06-12
scripts/logextract.py: port to python3

clee:
Does what it says on the tin. Tested it with several klippy.log files from Voron help tickets, and it appears to behave the same as the old python2 version.

### DIFF
--- a/scripts/logextract.py
+++ b/scripts/logextract.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 # Script to extract config and shutdown information file a klippy.log file
 #
 # Copyright (C) 2017  Kevin O'Connor <kevin@koconnor.net>
@@ -23,7 +23,7 @@ class GatherConfig:
         self.config_lines = []
         self.comments = []
     def add_line(self, line_num, line):
-        if line != '=======================':
+        if line != b'=======================':
             self.config_lines.append(line)
             return True
         self.finalize()
@@ -41,7 +41,10 @@ class GatherConfig:
             self.comments.append(comment)
     def write_file(self):
         f = open(self.filename, 'wb')
-        f.write('\n'.join(self.comments + self.config_lines).strip() + '\n')
+        lines = map(lambda l: l if bytes == type(l) else l.encode('utf-8'),
+                    [l for l in self.comments + self.config_lines])
+
+        f.write(b'\n'.join(list(lines)) + b'\n')
         f.close()
 
 
@@ -166,7 +169,8 @@ class MCUSentStream:
         self.sent_stream = []
         self.send_count = count
     def parse_line(self, line_num, line):
-        m = sent_r.match(line)
+        sline = str(line)
+        m = sent_r.match(sline)
         if m is not None:
             shortseq = int(m.group('shortseq'), 16)
             seq = (self.mcu.shutdown_seq + int(m.group('count'))
@@ -192,7 +196,8 @@ class MCUReceiveStream:
         self.mcu = mcu
         self.receive_stream = []
     def parse_line(self, line_num, line):
-        m = receive_r.match(line)
+        sline = str(line)
+        m = receive_r.match(sline)
         if m is not None:
             shortseq = int(m.group('shortseq'), 16)
             ts = float(m.group('time'))
@@ -251,20 +256,21 @@ class MCUStream:
             line = "mcu '%s': %s" % (self.name, line)
         return line
     def parse_line(self, line_num, line):
-        m = clock_r.match(line)
+        sline = str(line)
+        m = clock_r.match(sline)
         if m is not None:
             self.mcu_freq = int(m.group('freq'))
             st = float(m.group('st'))
             sc = int(m.group('sc'))
             f = float(m.group('f'))
             self.clock_est = (st, sc, f)
-        m = serial_dump_r.match(line)
+        m = serial_dump_r.match(sline)
         if m is not None:
             self.shutdown_seq = int(m.group('rseq'))
-        m = send_dump_r.match(line)
+        m = send_dump_r.match(sline)
         if m is not None:
             return True, MCUSentStream(self, int(m.group('count')))
-        m = receive_dump_r.match(line)
+        m = receive_dump_r.match(sline)
         if m is not None:
             return True, MCUReceiveStream(self)
         return False, None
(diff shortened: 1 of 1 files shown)


Piezoid:
Why not simply open the files in text mode?
<details>
<summary>This is the patch I use.</summary>

```diff
diff --git a/scripts/logextract.py b/scripts/logextract.py
index acf75bcb..078ebc27 100755
--- a/scripts/logextract.py
+++ b/scripts/logextract.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 # Script to extract config and shutdown information file a klippy.log file
 #
 # Copyright (C) 2017  Kevin O'Connor <kevin@koconnor.net>
@@ -40,7 +40,7 @@ class GatherConfig:
         if comment is not None:
             self.comments.append(comment)
     def write_file(self):
-        f = open(self.filename, 'wb')
+        f = open(self.filename, 'w')
         f.write('\n'.join(self.comments + self.config_lines).strip() + '\n')
         f.close()
 
@@ -383,7 +383,7 @@ class GCodeStream:
         # Produce output gcode stream
         if self.gcode_stream:
             data = [ast.literal_eval(gc) for gc in self.gcode_commands]
-            f = open(self.gcode_filename, 'wb')
+            f = open(self.gcode_filename, 'w')
             f.write(self.gcode_state + ''.join(data))
             f.close()
         return self.gcode_stream
@@ -561,7 +561,7 @@ class GatherShutdown:
         out = [i for s in streams for i in s]
         out.sort()
         out = [i[2] for i in out]
-        f = open(self.filename, 'wb')
+        f = open(self.filename, 'w')
         f.write('\n'.join(self.comments + out))
         f.close()
 
@@ -577,7 +577,7 @@ def main():
     handler = None
     recent_lines = collections.deque([], 200)
     # Parse log file
-    f = open(logname, 'rb')
+    f = open(logname, 'r')
     for line_num, line in enumerate(f):
         line = line.rstrip()
         line_num += 1
```
</details>

Separately, you need to include a [Signed-off-by line](https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#format-of-commit-messages) in your commit message.

clee:
> Why not simply open the files in text mode?

Well, that probably would have been a lot easier, but I didn't want to change the actual behavior of the original program, which did attempt to do everything in binary mode instead of text mode.

If Kevin thinks that's the better approach, I'd be happy to rewrite it.

> Separately, you need to include a [Signed-off-by line](https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#format-of-commit-messages) in your commit message.

Right, thanks for the reminder.

clee:
Just noticed that this does *not* appear to be fully working correctly. The configs are getting extracted properly, I think, but the actual crash logs are not. Marking as draft until I figure out why.

Piezoid:
An issue unrelated to encoding happens on Python3:
```Traceback (most recent call last):
  File "/home/clee/src/klipper/scripts/logextract.py", line 610, in <module>
    main()
  File "/home/clee/src/klipper/scripts/logextract.py", line 586, in main
    ret = handler.add_line(line_num, line)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/clee/src/klipper/scripts/logextract.py", line 542, in add_line
    self.finalize()
  File "/home/clee/src/klipper/scripts/logextract.py", line 558, in finalize
    if s[i-1][0] > s[i][0]:
       ^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
```
`StatsStream.get_lines` returns the first lines with a `None` timestamp until the first `Stats <ts>:` line is reached. None is comparable with itself and floats in python 2 but not in python 3.

The solution is to either drop the lines before the first timestamp:
```diff
@@ -502,13 +502,18 @@ class StatsStream:
                 break
         # Improve accuracy of stats timestamps
         last_ts = self.stats_stream[0][0]
+        start_line_idx = 0
         for i, (ts, line_num, line) in enumerate(self.stats_stream):
             if ts is not None:
                 last_ts = self.check_stats_seq(ts, line)
             elif (line_num >= self.shutdown_line_num
                   and last_ts <= max_stream_ts):
                 last_ts = max_stream_ts + 0.00000001
+            elif last_ts is None:
+                start_line_idx = i + 1
+                continue
             self.stats_stream[i] = (last_ts, line_num, line)
+        del self.stats_stream[:start_line_idx]
         return self.stats_stream
 
 # Main handler for creating shutdown diagnostics file
 ```
 or to back propagate the first timestamp:
 ```diff
 @@ -500,8 +500,13 @@ class StatsStream:
             if info[0] is not None and info[0] >= min_stream_ts - 5.:
                 del self.stats_stream[:i]
                 break
+        # Find the first stats timestamp
+        last_ts = min_stream_ts
+        for ts, line_num, line in self.stats_stream:
+            if ts is not None:
+                last_ts = ts
+                break
         # Improve accuracy of stats timestamps
-        last_ts = self.stats_stream[0][0]
         for i, (ts, line_num, line) in enumerate(self.stats_stream):
             if ts is not None:
                 last_ts = self.check_stats_seq(ts, line)
```

scripts/logextract.py
```diff
@@ -41,7 +41,10 @@ def add_comment(self, comment):
             self.comments.append(comment)
     def write_file(self):
         f = open(self.filename, 'wb')
```
thijstriemstra, line 43: can you use `with open(self.filename, 'wb') as f:` instead, for a more modern approach.

scripts/logextract.py
```diff
@@ -562,7 +569,9 @@ def finalize(self):
         out.sort()
         out = [i[2] for i in out]
         f = open(self.filename, 'wb')
```
thijstriemstra, line 571: this could use a `with open(self.filename, 'wb') as f:` as well. (don't forget to remove the `f.close()` call at the bottom).

thijstriemstra (changes requested):

KevinOConnor:
Thanks for looking at this.  For what it is worth, my general preference for debugging tools like logextract is to open the files in text mode.  In general I prefer to keep the code both python2 and python3 compatible with the minimal amount of code wrappers.

The comparison with None looks like a bug that should probably be fixed regardless.

Cheers,
-Kevin

PR #6251, opened 2023-06-14
Adding an adaptive horizontal_move_z setting for QUAD_GANTRY_LEVEL

akramer:
Adaptive horizontal_move_z allows the first Z height when starting QGL to be large, and then reduced to a smaller number (within the bounds of the error margin) for subsequent passes. This allows faster QGL with less z-hop between travels.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1181,6 +1181,15 @@ Where x is the 0, 0 point on the bed
 #horizontal_move_z: 5
 #   The height (in mm) that the head should be commanded to move to
 #   just prior to starting a probe operation. The default is 5.
+#adaptive_horizontal_move_z: false
+#   Whether to use an adaptive algorithm for horizontal_move_z.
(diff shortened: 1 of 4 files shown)


kdb424:
This should likely be added to screws tilt adjust as well to keep them as close to parity as possible.

akramer:
I don't have a Trident to verify that the code functions if I do make edits. Do you have one and would you be willing to test?

kmobs:
Doesn't this have significant overlap with this commit? https://github.com/Klipper3d/klipper/pull/6079/commits/c925c6d8aa78c1c2c455697bfa03ac40e95202c9

The current code can be used in conjunction with a macro to decrease the move z after the initial coarse passes.

KevinOConnor:
Thanks.  As high-level feedback, I'm not sure optimizing the time it takes to perform calibration tools is worth the additional config complexity that goes along with it.  I regularly receive feedback that users are overwhelmed by the large number of options in the Klipper printer.cfg file.

-Kevin

PR #6252, opened 2023-06-14
console.py: don't force usage of python2 in the shebang

kdomanski:
It seems that by now all parts of Klipper run fine on Python 3. Many systems (like MainsailOS) don't ship Python 2 anymore. On those systems simply running ./console.py fails due to the specific shebang.

### DIFF
--- a/klippy/console.py
+++ b/klippy/console.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # Script to implement a test console with firmware over serial port
 #
 # Copyright (C) 2016-2021  Kevin O'Connor <kevin@koconnor.net>


JamesH1978:
Thank you for submitting a PR, please can you refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially point3 and the signed off by line.

and just to note, unfortunately while this change will work on a pi, where `python` will automatically map to `python3`, this is not the case on other distributions, where python3 is implicitly needed.

Thanks
James

thijstriemstra (approved):

KevinOConnor:
@thijstriemstra - thanks for reviewing.  It would help if you could follow the review steps documented at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review as part of your review.

-Kevin

kdomanski:
Hi, I added the missing signoff.

@JamesH1978 I'm not sure I understand the implication. Are you saying this should be changed to `python3` instead?

KevinOConnor:
Thanks.  As James mentioned earlier though, we've found distributions that some Linux don't work with `python` - they require either `python2` or `python3`.  I think console.py should use the same definition as klippy.py , so I think it should remain `python2` until we are ready to define it as `python3`.

In the interim, users can directly run console.py (and klippy.py) by explicitly calling their python of choice - for example: `~/klippy3-env/bin/python3 ~/klipper/klippy/klippy.py`

-Kevin

kdomanski:
Interesting. Wouldn't a virtual env contain a `python` symlink regardless of the distro?

@JamesH1978 if you could clarify which distro(s) you mean, I'd love to personally test that.

JamesH1978:
this is not running in the klipper virtual env, but in main global python. I believe we saw this issue on full fat ubuntu installs, where it was implicity python2 or python3 and no python symlink, my Ubuntu 22.04 build:

image


Thanks
James

Sineos:
There are typically two apt packages that would create this symlink:
`python-is-python3` and `python-is-python2`. Both do nothing more than providing a symlink between, e.g. python3 and python.

kdomanski:
Sounds to me like it would be better to switch the shebang to `python3` across the board.
@KevinOConnor if you disagree with that, then please just close the PR - it's not that important anyways.

PR #6254, opened 2023-06-15
Kconfig HC32 - add provision to invert LCD drive polarity

SteveGotthardt:
module: Kconfig for HC32F460
One of the boards with HC32 uses inverters to drive the LCD and needs this code change to work.

### DIFF
--- a/src/hc32f460/Kconfig
+++ b/src/hc32f460/Kconfig
@@ -92,6 +92,11 @@ config CLOCK_FREQ
     default 200000000 if HC32F460_CLOCK_SPEED_200M
     default 168000000 if HC32F460_CLOCK_SPEED_168M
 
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I think we should try to avoid adding board specific options to "make menuconfig".  It's generally preferable to put these options in the python code (such that a user can enable it from printer.cfg).

If the python can't directly manage the pin values, then I thin it would be preferable to add a new "invert" parameter to the `config_st7920` mcu command.

That said, these old lcd screens are goofy at best.  Would an alternative solution be to tell users to put a touchscreen on their rpi and use a more modern interface?

-Kevin

SteveGotthardt:
Thanks. I think the python solution would be much better.

PR #6257, opened 2023-06-16
COLD_EXTRUDE and M302 command

Zeanon:
After at least two people and me in the klipper discourse wished to have an implementation of Marlin's M302 command, I just did it.
I tried to keep the changes minimal and the behaviour as close to the marlin command as possible(it is fully compatible with the marlin syntax, so every marlin gcode will just work as expected)

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -25,6 +25,7 @@ Klipper supports the following standard G-Code commands:
 - Set extruder temperature and wait: `M109 [T<index>] S<temperature>`
   - Note: M109 always waits for temperature to settle at requested
     value
+- Enable Cold Extrusion: `M302 [T<index>] [P<enable>] [S<min_extrude_temp>]
 - Set bed temperature: `M140 [S<temperature>]`
 - Set bed temperature and wait: `M190 S<temperature>`
   - Note: M190 always waits for temperature to settle at requested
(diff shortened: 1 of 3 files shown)


Sineos:
Following Klipper's philosophy of not adding "traditional" gcode commands, I would suggest:
* To remove `M302` --> Can be defined via macro if wanted
* Change `COLD_EXTRUDE HEATER=<heater_name> [P=<enable>] [S=<min_extrude_temp>]` to `COLD_EXTRUDE HEATER=<heater_name> MINIMUM=<target>`
* IMO, the `[P=<enable>]` is not needed since once can control it via the `MINIMUM=<target>`

Zeanon:
You are most definitely right on that one.
There is just one thing I would disagree on:
I would keep the P parameter(however I would call it if I remove the M302 implementation), since it allows you to activate cold extrude without modifying the min_extrude_temp, so no config write or anything

Zeanon:
I have rewritten the code a bit so now COLD_EXTRUDE doesnt accept `P` & `S` but rather `ENABLE` and `MIN_EXTRUDE_TEMP`.

KevinOConnor:
Thanks.  As high-level feedback, I'm not sure this is needed as it would seem users can already accomplish the same goals using existing mechanisms.

One could either change the config file to disable `min_extrude_temp` or one can use `FORCE_MOVE`.  I'm not sure we need to add another method.

-Kevin

Zeanon:
I guess using force move is a way you could do it and I never thought about it.
I just wrote it since a bunch of people on the discourse seemed to want it.
Maybe adding a little bit of info in the docs would be sufficient then?

PR #6258, opened 2023-06-17
Minor correction in i2c docs

rootiest:
When the software i2c stuff was added, the description for the i2c_bus option ended up below the new software i2c options instead of below the i2c_bus option like it should be.

This PR just corrects this very minor mistake.

Sorry for the excess commits, I'm not particularly git-literate 🤪

Signed-off-by: Chris Laprade [chris@rootiest.com](mailto:chris@rootiest.com)


KevinOConnor:
Thanks.  I agree this should be fixed in the docs, but I think we should also document the two parameters.  I have done this in commit 2ae3dd2f.

-Kevin

PR #6259, opened 2023-06-17
hc32f460: Fix UART for pins PH2/PB10 and PA13/PA14

Bobo1239:
I've tested every available UART pin config for the HC32F460 and can confirm they all work now. (HC32F460 devboard + `console.py`; and PH2/PB10 on my Anycubic Kobra via the LCD connector)

cc @SteveGotthardt


SteveGotthardt (approved):
Thanks for tracking that down!

KevinOConnor:
Thanks.

-Kevin

PR #6260, opened 2023-06-17
probe: Use fixed speed for probe offset movement

Bobo1239:
Low calibration speed settings can be confusing to the user during a PROBE_CALIBRATE since the nozzle will move very slowly to the probe offset and any commands during that time are ignored. Introduce a setting for probe offset movement speed to circumvent this.

### DIFF
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -263,7 +263,7 @@ class PrinterProbe:
         # Move the nozzle over the probe point
         curpos[0] += self.x_offset
         curpos[1] += self.y_offset
-        self._move(curpos, self.speed)
+        self._move(curpos, 50)
         # Start manual probe
         manual_probe.ManualProbeHelper(self.printer, gcmd,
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I don't think we need to add a configuration option for an infrequently used tool like PROBE_CALIBRATE.  I regularly receive feedback that the large number of configuration options in Klipper is overwhelming for users.

-Kevin

Bobo1239:
Yeah, I can understand that as a relative new user myself :sweat_smile:. I've changed it to a fixed 50mm/s.

Bobo1239:
@KevinOConnor Ping since it seems like you may have missed my updated commit.

KevinOConnor:
Thanks.  However, I don't think we should change this part of the code.  I fear 50mm/s may not be a good choice for all printers.  I think it is okay to continue to use `self.speed` for this infrequently used calibration tool.

-Kevin

PR #6261, opened 2023-06-18
docs: Add multistep_filt warning to config reference

leptun:
See https://github.com/Klipper3d/klipper/pull/6134#issuecomment-1595640504

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3536,6 +3536,10 @@ run_current:
 #   (hex form is not supported). In order to compute the wave table fields,
 #   see the tmc2130 "Calculation Sheet" from the Trinamic website.
 #driver_MULTISTEP_FILT: True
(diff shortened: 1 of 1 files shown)


Sineos (approved):

KevinOConnor:
Thanks.  As high-level feedback, I fear this documentation change could be more confusing than helpful to the majority of users.  There are *a lot* of tmc driver settings that could cause subtle print failures.  I'm not sure we should point users to `microstep_filt` should they encounter an issue.

The microstep_filt change was made to tmc5160.py a little over 2.5 months ago and I've only seen a couple of regression reports.  (Though I admit that I may not have seen all reports.)  Perhaps we should wait to see if there are many more reports of issues and to see if we can discern a pattern to the failures, such that we could document what type of hardware/usages may need this config change.

Cheers,
-Kevin

PR #6262, opened 2023-06-18
Implement CLEAR for SET_KINEMATIC_POSITION 

twelho:
`SET_KINEMATIC_POSITION CLEAR=<[X][Y][Z]>` clears the homing status (resets the axis limits) without turning off motors. This is particularly useful when implementing safe Z homing in `[homing_override]` on printers with multiple independent Z steppers (where `FORCE_MOVE` can't be used):

```
# Forcibly move Z up (Z hop)
{% set homed = printer.toolhead.homed_axes | upper | list %}
{% set not_homed = ['X', 'Y', 'Z'] | reject('in', homed) %}
{% if not 'X' in homed or not 'Y' in homed %}
  SET_KINEMATIC_POSITION Z=0
  G0 Z{z_hop} F{bltouch_speed}
  SET_KINEMATIC_POSITION CLEAR={not_homed | join}
{% endif %}
```

This functionality was already available to the safe Z homing module via `note_z_not_homed`. As a bonus, the `_motor_off` implementations across various kinematics are also simplified.

I'm not 100% sure whether this should send the `toolhead:set_position` event since it only modifies the limits and not the position itself. The current consumer (singular, `GCodeMove`) of that event doesn't seem to need it, but maybe you have some other future plans?


twelho:
Rebased on top of master and implemented a fix for `KEEP=XYZ` not keeping anything (bug with falsy values)

KevinOConnor:
Thanks.  As high-level feedback the change seems fine to me.  I do have a couple of minor comments.

I'm not sure about adding a new CLEAR_KINEMATIC_POSITION command.  I wonder if adding some kind of `CLEAR=XYZ` type of parameter to the existing SET_KINEMATIC_POSITION may be more appropriate.  (Limiting builtin commands reduces possible conflicts with user macros.)

Is there a reason that `clear_position()` takes a list of axes numbers instead of a string of axes (eg, "xyz")?

If `clear_positions()` is to be a mandatory method of the kinematic classes it would be good to update docs/Code_Overview.md to include it in the "Adding new kinematics" section (bullet 5).

If you're rebasing this PR it would be nice to have the kinematic `clear_position()` change in a separate commit from the XXX_KINEMATIC_POSITION command.  Though if you're not super familiar with git then it's not a requirement.

-Kevin

KevinOConnor:
Also, to your earlier question - I don't think the code should call `printer.send_event("toolhead:set_position")` on a clearing of homing state.  That event is intended to inform the gcode parsing code that it has to resynchronize its position with the toolhead code.

In general, I don't think the toolhead needs to be involved in a "clear homing state" request - the low-level code can just directly instruct the kinematic class of the change in state (as was done for `note_z_home()`).  It is not necessary to flush the step generation either.

Along those same lines, maybe `clear_homing_state()` would be a better name than `clear_position()`?

-Kevin

Sineos:
@twelho any updates?

twelho:
> I'm not sure about adding a new CLEAR_KINEMATIC_POSITION command. I wonder if adding some kind of CLEAR=XYZ type of parameter to the existing SET_KINEMATIC_POSITION may be more appropriate. (Limiting builtin commands reduces possible conflicts with user macros.)

That works as well, I can expose the functionality through `CLEAR` for `SET_KINEMATIC_POSITION`.

> Is there a reason that clear_position() takes a list of axes numbers instead of a string of axes (eg, "xyz")?

`set_position` for kinematics takes in `homing_axes` as numerical values, so I copied that pattern for consistency.

> If clear_positions() is to be a mandatory method of the kinematic classes it would be good to update docs/Code_Overview.md to include it in the "Adding new kinematics" section (bullet 5).

Thanks for the pointer, I'll update the the the docs as well.

> If you're rebasing this PR it would be nice to have the kinematic clear_position() change in a separate commit from the XXX_KINEMATIC_POSITION command. Though if you're not super familiar with git then it's not a requirement.

Sure, I can do that.

twelho:
The changes are now implemented and I've tested that it works as intended on my hardware, so it's ready for review.

b00gie-dev:
@twelho Awesome! Thank you for your work on that!

twelho:
Not stale, could someone please take a look at this?

b00gie-dev:
Can we please reopen this? It‘s still a wanted feature 😊

austinrdennis:
I'm dying for this to be integrated! It's been tricky to work around this in my homing override and this addition would fix my only remaining problem with that setup.

Surion79:
i would love to get this implemented

b00gie-dev:
So what we need to do to get this feature? Someone has to accept it or what's the issue right now?

twelho:
I believe it is in need of a reviewer. Should there be more changes requested, I'm happy to work on them to get this PR merged.

pciavald:
When applying this PR to the recent klipper commit `6cd1742`, i'm getting the following error:

```
  File "/root/klippy/klippy.py", line 130, in _connect
    self._read_config()
  File "/root/klippy/klippy.py", line 125, in _read_config
    m.add_printer_objects(config)
  File "/root/klippy/toolhead.py", line 671, in add_printer_objects
    kinematics.extruder.add_printer_objects(config)
  File "/root/klippy/kinematics/extruder.py", line 313, in add_printer_objects
    pe = PrinterExtruder(config.getsection(section), i)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/klippy/kinematics/extruder.py", line 177, in __init__
    self.extruder_stepper = ExtruderStepper(config)
                            ^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/klippy/kinematics/extruder.py", line 21, in __init__
    ffi_lib.extruder_stepper_free)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/klippy/pyenv/versions/3.11.10/lib/python3.11/site-packages/cffi/api.py", line 912, in __getattr__
    make_accessor(name)
  File "/root/klippy/pyenv/versions/3.11.10/lib/python3.11/site-packages/cffi/api.py", line 907, in make_accessor
    raise AttributeError(name)
AttributeError: extruder_stepper_free
```

pciavald:
Hi, i've managed to apply this PR.

However i'm observing the following behavior:

```
Recv: // printer['toolhead'].homed_axes =

G28 X

Recv: // printer['toolhead'].homed_axes = x

SET_KINEMATIC_POSITION CLEAR=X

Recv: // printer['toolhead'].homed_axes = yz
```

klippy/extras/force_move.py
```diff
@@ -131,8 +131,16 @@ def cmd_SET_KINEMATIC_POSITION(self, gcmd):
         x = gcmd.get_float('X', curpos[0])
         y = gcmd.get_float('Y', curpos[1])
         z = gcmd.get_float('Z', curpos[2])
-        logging.info("SET_KINEMATIC_POSITION pos=%.3f,%.3f,%.3f", x, y, z)
-        toolhead.set_position([x, y, z, curpos[3]], homing_axes=(0, 1, 2))
+        clear = gcmd.get('CLEAR', '').upper()
+        axes = ['X', 'Y', 'Z']
```
pciavald, line 135: this should be based on printer['toolhead'].homed_axes

pciavald (changes requested):

klippy/extras/force_move.py
pciavald: this seems to work

```py
        clear = gcmd.get('CLEAR', '').upper()
        curtime = self.printer.get_reactor().monotonic()
        kin_status = toolhead.get_kinematics().get_status(curtime)
        axes = list(kin_status['homed_axes'].upper())
        clear_axes = [axes.index(a) for a in axes if a in clear]
        homing_axes = tuple(i for i, a in enumerate(['X', 'Y', 'Z']) if a in axes and a not in clear)
```

pciavald: the `clear` variable may need to be sorted alphabetically in case user inputs CLEAR=ZYX

twelho: Thanks for noticing this, I'll take a look this weekend and make sure it works in those cases as well.

twelho:
> Hi, i've managed to apply this PR.
> 
> However i'm observing the following behavior:
> 
> ```
> Recv: // printer['toolhead'].homed_axes =
> 
> G28 X
> 
> Recv: // printer['toolhead'].homed_axes = x
> 
> SET_KINEMATIC_POSITION CLEAR=X
> 
> Recv: // printer['toolhead'].homed_axes = yz
> ```

Now that I look at this, this is intended behavior. `SET_KINEMATIC_POSITION` will always set the full kinematic position (for backwards compatibility), the `CLEAR` option just allows you to discard specific axis after everything has been set. If you want to ensure that un-homed axes remain un-homed, you need to add them to `CLEAR`, see the `not_homed` variable in the example in the description.

`CLEAR` is implemented this way for two reasons: backwards compatibility and to retain the declarative and atomic nature of  `SET_KINEMATIC_POSITION` (i.e., after running the command, you will always end up in a predicable and repeatable kinematic state). This reduces the chances of catastrophic mistakes (crashing the gantry etc.). It also allows combining operations like `SET_KINEMATIC_POSITION X=150 Y=150 CLEAR=Z`. If you omit an axis, `SET_KINEMATIC_POSITION` will simply use the current position for it (which is the existing behavior).

klippy/extras/force_move.py
twelho: Upon closer inspection this is intended behavior, see response [below](https://github.com/Klipper3d/klipper/pull/6262#issuecomment-2548252460).

klippy/extras/safe_z_home.py
```diff
@@ -40,8 +40,8 @@ def cmd_G28(self, gcmd):
                 toolhead.set_position(pos, homing_axes=[2])
                 toolhead.manual_move([None, None, self.z_hop],
                                      self.z_hop_speed)
-                if hasattr(toolhead.get_kinematics(), "note_z_not_homed"):
-                    toolhead.get_kinematics().note_z_not_homed()
+                if hasattr(toolhead.get_kinematics(), "clear_homing_state"):
+                    toolhead.get_kinematics().clear_homing_state((2,))
```
KevinOConnor, line 44: If we're making `clear_homing_state()` required for all kinematics, then I think it should be called unconditionally here (that is, no `hasattr()` check).

klippy/toolhead.py
```diff
@@ -457,13 +457,14 @@ def _flush_handler(self, eventtime):
     # Movement commands
     def get_position(self):
         return list(self.commanded_pos)
-    def set_position(self, newpos, homing_axes=()):
+    def set_position(self, newpos, homing_axes=(), clear_axes=()):
```
KevinOConnor, line 460: I don't think we should add a `clear_axes` parameter to `set_position()`.  Instead, force_move can explicitly call `toolhead.get_kinematics().clear_homing_state()` (as save_z_home does).

KevinOConnor (commented):
Thanks for working on this.  In general it seems fine to me.  Sorry for the delay in responding.

I have a couple of minor comments - see below.

-Kevin

klippy/toolhead.py
twelho: That would mean the `toolhead:set_position` event is sent before instead of after the homing states of the given axes are cleared. If that's acceptable, I'll go ahead and follow your suggestion.

KevinOConnor: Changing the order of events seems okay to me.

-Kevin

klippy/extras/manual_stepper.py
```diff
@@ -109,7 +109,7 @@ def flush_step_generation(self):
         self.sync_print_time()
     def get_position(self):
         return [self.rail.get_commanded_position(), 0., 0., 0.]
-    def set_position(self, newpos, homing_axes=()):
+    def set_position(self, newpos):
```
KevinOConnor, line 112: This change doesn't look correct to me.

-Kevin

twelho: I was not sure whether this `set_position` complies to the same interface as the toolhead. From what I can see, it's essentially implementing the same interface as [`PrinterRail.set_position`](https://github.com/Klipper3d/klipper/blob/8a3d2afd796414d64b995ac753148484b77198dd/klippy/stepper.py#L419), and thus shouldn't have the homing info.

KevinOConnor: No - the manual_stepper class is passed to homing.py as an emulated "toolhead" and thus it should follow the toolhead `set_position()` signature.

-Kevin

twelho: All clear, fixed!

KevinOConnor:
Try rebasing to the lastest master branch - as I've updated to the latest "uploads-artifact" code there.

Thanks,
-Kevin

KevinOConnor:
Thanks.

-Kevin

b00gie-dev:
It finally happened 🎉
Big thanks to for this for you all!

PR #6263, opened 2023-06-18
virtual_sdcard: reset file position and size to integer zero

Arksine:
This is a fix to a minor bug which results in the `get_status()` reporting floating point values for `file_position` and `file_size` after a reset or cancel.   Some API clients expect the values in these fields to be integers, as reported in arksine/moonraker#673.


KevinOConnor:
Thanks.

-Kevin

PR #6264, opened 2023-06-18
Create printer-longer-lk1-2019

kenavru:
Tested config on Longer LK1


JamesH1978:
Thank you for submitting a PR, please can you refer to the documentation, https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially point 3 and the signed-off-by line and https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md especially point 4 and adding the regression tests.

Thanks
James

PR #6267, opened 2023-06-21
bed_mesh: fix fade regression

Arksine:
This resolves a regression introduced in #6242, where the average Z value of the mesh is calculated before the mesh is offset and never updated.  When fade is enabled the default value of `fade_target` is set to the average z, resulting in parts that are stretched vertically.


KevinOConnor:
Thanks.

-Kevin

PR #6273, opened 2023-06-28
Update Config_Reference.md - heater_fan wording

JamesH1978:
The reference to my_nozzle_fan is misleading and could cause people to skip over the [fan] definition and assume this is the parts/print cooling fan. Alias changed to reflect heatbreak_cooling_fan as used in most configs that have a controllable hotend fan.

Thanks
James


KevinOConnor:
Thanks.

-Kevin

PR #6274, opened 2023-06-28
Multiple config - change wording for heater fan

JamesH1978:
The reference to `my_nozzle_fan` in these configs could be confusing and misleading, leading users to think it is the parts/print cooling fan, changed to heatbreak_cooling_fan as used in other configs.

Thanks
James


KevinOConnor:
Thanks.

-Kevin

PR #6275, opened 2023-06-28
Add modules Recore and Thermocouple and add config files for Recore A6 and A7

eliasbakken:
This is the final PR for adding full upstream support for Recore boards revision A6 and A7. The PR adds a module called recore that enables the high power input section on the board and sets up the temperature sections for either thermistor (gain 1) or thermocouple (gain 100). There is also support for programming the pull-up and adding a voltage offset. These settings are necessary in order to allow both a thermistor and a thermocouple to be used on the same inputs. The thermocouple modules adds the necessary calculations for type K thermocouples.


config/generic-recore-a7.cfg
```diff
@@ -0,0 +1,284 @@
...
+voltage1: 0
+temperature2: 33
+voltage2: 3.3
+
+[temperature_sensor fan_current]
+adc_voltage: 3.3
+sensor_pin: PB0
+sensor_type: fan_current
+max_temp: 2.0
+gcode_id: FanCurrent
```
KevinOConnor, line 284: Out of curiosity, is there a reason for these gcode_id fields?  They only change the output of the `M105` command which, as far as I know, none of the modern interfaces use.

docs/Config_Reference.md
```diff
@@ -4410,6 +4410,62 @@ host_mcu:
...
+# pullup_t0: 1
+# pullup_t1: 1
+# pullup_t2: 1
+# pullup_t3: 1
+#   Programmable 4.7 K pull-up. 0 = off, 1 = on
+# offset_t0: 0
+# offset_t1: 0
+# offset_t2: 0
+# offset_t3: 0
+#   Programmable voltage offset on the input. 0 = off, 1 = on
```
KevinOConnor, line 4444: Please expand on the field comments here.  It's not clear to me what each field does.  It's also good to explicitly state what the default is (or to state the field is required).

klippy/extras/thermocouple.py
```diff
@@ -0,0 +1,85 @@
+# Temperature measurements with thermocouples. Only support for Type K for now.
+# Cold junction compensation
```
KevinOConnor, line 2: Please start the file with a one line comment, then an empty comment line, then the copyright statement, then an empty comment line, then the license statement.  Feel free to add any number of comments later in the file, but I'd prefer to keep the top copyright statement uniform across files.

KevinOConnor (commented):
Thanks.  In general it seems fine to me.  However, I didn't understand the new `thermocouple.py` module.  The name seems generic, but I couldn't figure out from the documentation what the module does.  If this is intended to be a generic module for thermocouple temperature sensors then I think the documentation will need to be expanded.  If it is intended to be a module specific to the recore then I think a different name should be chosen.  In either case though, can you elaborate on what the thermocouple.py code does?

I also have a few minor comments - see below.

Cheers,
-Kevin

PR #6277, opened 2023-06-29
toolchanger: Initial API documentation for review.

viesturz:
Setting up toolchangers with klipper is currently a free for all land of macros and plugins. This is an attempt to introduce some common structure that is generic enough to fit almost all use cases but still provide useful structure to make most configs look familiar.

https://github.com/TypQxQ/Klipper_ToolChanger is a capable implementation of something similar. This is trying to be significantly more minimalistic and less opinionated.

### DIFF
--- a/config/sample-idex.cfg
+++ b/config/sample-idex.cfg
@@ -34,22 +34,6 @@ pid_Kd: 114
 min_temp: 0
 max_temp: 250
 
-# Helper script to park the carriage (called from T0 and T1 macros)
-[gcode_macro PARK_extruder]
-gcode:
-    SAVE_GCODE_STATE NAME=park0
-    G90
-    G1 X0
-    RESTORE_GCODE_STATE NAME=park0
-
-# Activate the primary extruder
-[gcode_macro T0]
-gcode:
-    PARK_{printer.toolhead.extruder}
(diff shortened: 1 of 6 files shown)


JamesH1978:
I am not sure what you are trying to achieve here? there is no module called `toolchanger` or `tool` in klipper? is this tied to another PR that will introduce these modules? This might be better suited for the discourse server where it could be discussed further.

Thanks
James

viesturz:
Thanks for the input. This is to discuss an API conceptually. The API here
seems to be the more tricky part, implementation is relatively straight
forward.
Yes Discord seems to be the right place to discuss, I posted this
yesterday, looking forward for feedback.


Ready to do the implementation once the API is ironed out.

viesturz:
Pushed significant update. Getting rid of templates in favor of option to nest toolchangers.

Considering maybe adding per-toolchanger variables, would help with chaining multiple changers of same type, like AMS units, but seems too complex for first run.

KevinOConnor:
I'm not sure what the status of this PR is or was.  If there is interest in discussing future Klipper development, I suggest using Klipper Discourse for that.  Note that Discourse is not the same as Discord.

-Kevin

viesturz:
Hi, Discourse vs Discord is a good catch, I was wondering why Discord dev chat was not responsive.

PR #6278, opened 2023-07-01
Docs: Update Config_checks.md

JamesH1978:
I am endeavoring to rewrite some of the docs to be more front end agnostic to reflect the different front ends available since these docs were written. I am starting with this one, as it quite often used. Please can I have some feedback on the wording.

Thanks
James

### DIFF
--- a/docs/Config_checks.md
+++ b/docs/Config_checks.md
@@ -15,7 +15,7 @@ config file is successfully loaded.
 ## Verify temperature
 
 Start by verifying that temperatures are being properly reported.
-Navigate to the temperature graph section in your front end.
+Navigate to the temperature graph section in the user interface.
 Verify that the temperature of the nozzle and bed (if applicable) are
 present and not increasing. If it is increasing, remove power from the
 printer. If the temperatures are not accurate, review the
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  We can certainly change the wording of the Klipper documents to better correlate with newer interfaces.

I do think using "front end" and "terminal/console" may be a bit too "techy" for these documents though.  For what it's worth, I'm not sure we need to be explicit with "front end" and maybe we could choose a term (like "command console") and use it consistently throughout the document.

Cheers,
-Kevin

Sineos:
Coming from a (new / inexperienced) user perspective:
In typical setups today, we have:
* an intimidating Linux shell
* a GUI, i.e. a webinterface
* a console / command line in the GUI

The fact that there are shell-like command interfaces and some more exotic things, I would disregard because once a user starts thinking about these, they are no longer inexperienced.

I'd propose to use a terminology that "feels recognizable" or maybe add some small glossary already in the installation document that defines:

Shell:
Command line interface or terminal window of Linux that is displayed after logging in with SSH tools like Putty. Mainly used during the setup process of Klipper.

User Interface:
Typically, a webinterface like Octoprint, Mainsail, Fluidd etc. that allows controlling and working with the 3D printer

Console:
A text-base command interface that is found in the User Interfaces. It can be used to send command directly to the Printer.

JamesH1978:
Thanks for the feedback, I have taken a mix of @KevinOConnor and @Sineos comments and used User interface and Command console.

Thanks
James

KevinOConnor:
Thanks!

-Kevin

PR #6279, opened 2023-07-02
rp2040_flash: main.c: Fix the type when printing num_blocks

Haxk20:
image->num_blocks is of type unsigned size_t thus unsigned int.

%lu specifies a type of long unsigned int. Thus resulting in compiler warning about type mismatch. Correct this by printing the value with a correct type.

### DIFF
--- a/lib/rp2040_flash/main.c
+++ b/lib/rp2040_flash/main.c
@@ -174,7 +174,7 @@ int main(int argc, char *argv[]) {
         rc = 1;
         goto do_exit;
     }
-    fprintf(stderr, "Loaded UF2 image with %lu pages\n", image->num_blocks);
+    fprintf(stderr, "Loaded UF2 image with %zu pages\n", image->num_blocks);
 
     bool has_target = false;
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I think a size_t on some platforms is 64bit and I suspect that is the reason for the `%lu`.  Perhaps use `%u` and explicitly cast to `(int)`.  In either case, though, I'd defer to @dalegaard on this.

-Kevin

Haxk20:
> Thanks.  I think a size_t on some platforms is 64bit and I suspect that is the reason for the `%lu`.  Perhaps use `%u` and explicitly cast to `(int)`.  In either case, though, I'd defer to @dalegaard on this.
> 
> -Kevin

Actually the way to do this correctly is to use the length modifier z. It was but too late when i did this PR i apologize for the broken fix. I will push the proper fix once I'm home :)

Haxk20:
Proper fix pushed :)

KevinOConnor:
Thanks.

-Kevin

PR #6280, opened 2023-07-02
scripts: logextract.py Python 3 compatibility

Piezoid:
This is an alternative to PR #6248.

- Open log files in text mode,
- Normalize the encoding of tmcuart messages: logs from klippy on python3 contains `tmcuart_send oid=0 write=b'\xea[...]`.  With python2, there is no `b''` string marker. The regular expression removes the marker if present. Then, a `b` is prepended to the literal-encoded string before parsing it with `ast.literal_eval`. This ensures interpretation as a byte-string on Python3, independently of the source of the log. It has no effect when running `logextract.py` on Python 2.
- Fix a [bug](https://github.com/Klipper3d/klipper/pull/6248#issuecomment-1587135735) where StatsStream.get_lines() would return None timestamps for lines before the first Stats line. Timestamps are now back-propagated.

I compared the outputs with `cat klippy.log.* | md5sum`, observing no difference between running the current version with Python 2, and the new ported version with Python 2 and 3.

### DIFF
--- a/scripts/logextract.py
+++ b/scripts/logextract.py
@@ -4,7 +4,7 @@
 # Copyright (C) 2017  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import sys, re, collections, ast
+import sys, re, collections, ast, itertools
 
 def format_comment(line_num, line):
     return "# %6d: %s" % (line_num, line)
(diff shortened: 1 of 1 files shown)


Piezoid:
Added a second commit slightly modernizing the code, using:
- context managers (`with open` statements) as requested by @thijstriemstra,
- generators in places where it makes sense to avoid copies.

Behavior change: shutdown files now end with a new line.

clee:
This works much better than my draft PR. I'm happy to abandon mine :)

KevinOConnor:
Thanks.

-Kevin

PR #6282, opened 2023-07-05
fan: Support multiple part fans

viesturz:
Multi extruder setups can have separate part cooling fan for each extruder. This change allows runtime switching between fans using ACTIVATE_FAN command, similar to ACTIVATE_EXTRUDER.

The fix is to essentially make [fan] the same as [fan_generic] and use ACTIVATE_FAN to specifcy which fan is the printer fan. Inactive printer fans may be manually controlled using existing  SET_FAN_SPEED command.

This helps primarliy with PrusaSlicer gcode, that just emits M106 S110; T1; when changing tools. Making emulation with macros quite complex.

### DIFF
--- a/config/sample-idex.cfg
+++ b/config/sample-idex.cfg
@@ -34,6 +34,10 @@ pid_Kd: 114
 min_temp: 0
 max_temp: 250
 
+# The definition for the part cooling fan for primary extruder
+[fan]
+pin: PH6
+
 # Helper script to park the carriage (called from T0 and T1 macros)
 [gcode_macro PARK_extruder]
 gcode:
@@ -47,6 +51,7 @@ gcode:
 gcode:
     PARK_{printer.toolhead.extruder}
     ACTIVATE_EXTRUDER EXTRUDER=extruder
+    ACTIVATE_FAN FAN=fan
     SET_DUAL_CARRIAGE CARRIAGE=0
     SET_GCODE_OFFSET Y=0
 
@@ -81,6 +86,10 @@ pid_Kd: 114
 min_temp: 0
 max_temp: 250
 
+# The definition for the part cooling fan for secondary extruder
+[fan secondary]
+pin: PH7
+
 [gcode_macro PARK_extruder1]
 gcode:
     SAVE_GCODE_STATE NAME=park1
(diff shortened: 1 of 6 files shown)


KevinOConnor:
Thanks.  Sorry for the delay in responding.  As high-level feedback, I'm not sure we should add new commands that are used solely to control the behavior of old g-code commands.  That is, M106 and M107 are old poorly defined commands, and I suspect it may be more transparent for users to define their own macros if they need those commands to do something else.  For example, something like:

```
[gcode_macro M106]
    variable_active_fan: "fan_generic primary_fan"
    gcode: SET_FAN_SPEED FAN="${active_fan}" SPEED={params.S|float / 255.0}
```

-Kevin

viesturz:
Thanks for the feedback.
I can see this going roughly three directions:

**A Custom macros for each user**
+ No need to change klipper or slicers
- users need to tinker around to find a working slicer/macro combo
- no agreement what gcodes to use
- Slicers seem to be adding hacks for Klipper already, making it even more messy

**B Klipper provides better support for M106/M107 for existing slicers**
+ Known documented way how to setup multiple fans
+ minimal/no changes to slicers.
 - more code in Klipper

**C Slicers start using SET_FAN_SPEED / fan_generic for multi material setups**
+ Known documented way how to setup multiple fans
~need a Klipper gcode dialect, this seems to be happening already however
- SET_FAN_SPEED does not support the main fan, making migration painful - this should be easy-ish to fix however

Can you share some guidance where do you see Klipper going?

KevinOConnor:
Well, if you want my opinion, I'd prefer if slicers didn't mess with the fans at all.  That is, I suspect the overwhelming majority of users would get better results if the fan got turned on at an early layer (eg, layer 2) and the fan speed was never changed after that.  I suspect I am in the minority with that opinion though.  That is, I suspect that the slicers are trying to be way too clever, and I also suspect many users think "too clever" is a good thing.

That aside, I agree with your concerns on a lack of standardization.  I have the same comments mentioned at https://github.com/Klipper3d/klipper/pull/6294#issuecomment-1686772640 - I think we should have a consensus on a "vision" before making changes so that we don't inadvertently make the lack of standardization worse.

-Kevin

PR #6283, opened 2023-07-07
Config: Update Octopus Max EZ for wrong bltouch pins

JamesH1978:
It has been discovered on the discord by the user Pizn that the bltouch pins are swapped in this config and don't include the ^ pull on the sensor pin.

This has been tested and confirmed by the user.

Thanks
James


dmbutyugin (approved):
Thanks, the change seems to be in line with the published board pinout:
https://github.com/bigtreetech/Octopus-Max-EZ/blob/master/Hardware/BIGTREETECH%20Octopus%20MAX%20EZ%20V1.0%20PinOut.png
So, should be good.

KevinOConnor:
Thanks.

-Kevin

PR #6284, opened 2023-07-07
Add process id, user id and group id to the info endpoint.

SkyKnightSKS:
Adding these three pieces of information to the info endpoint allows the system that connects to the klippy.sock to simply ask klipper these details. This avoids having to resort to less accurate methods. 

For example moonraker is checking the process that generated the socket itself. This works for the most time. However when using a socket proxy this fails.


pedrolamas (approved):
Code wise these changes look fine to me and from my tests it is working as intended.

image

KevinOConnor:
Thanks.

-Kevin

PR #6285, opened 2023-07-09
Add printer config for Anycubic Kobra Neo

dDaYb:
Config: for Anycubic Kobra Neo.
It is almost the same as for Anycubic Kobra Go.


JamesH1978:
Good morning, thank you for submitting a PR, Please could you refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially the section about the signed-off-by line and point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and adding the regression tests,

Thanks
James

PR #6288, opened 2023-07-11
samc21: Fix CAN RAM base address

eamaclean:
Upstream header bug leads to clobbering of RXF0C and TXBC registers, causing severe memory corruption.

fdcan.c (and any other name applied to a Bosch M_CAN driver) on all targets should probably be modified to prevent such behaviour in future, but this will get SAM C21 up and running.


Wulfsta:
Tested, seems to fix the error case we've been hitting. Should be good to merge.

KevinOConnor:
Thanks.  I suspect that was very challenging to track down.

-Kevin

PR #6289, opened 2023-07-12
motan: Added norm2 and smooth analyzers

dmbutyugin:
These two analyzers are useful, in my opinion, for various analyses of the motion data, e.g. as

- `["norm2(derivative(stepq(stepper_x)),derivative(stepq(stepper_y)))"]` to find the commanded toolhead velocity in XY plane from its X and Y steppers velocities;
- `["smooth(adxl345(hotend,x))","smooth(adxl345(bed,y))"]` to plot the smoothed accelerometer readings to eliminate high-frequency noise from the data.

### DIFF
--- a/scripts/motan/analyzers.py
+++ b/scripts/motan/analyzers.py
@@ -105,6 +105,59 @@ class GenIntegral:
         return data
 AHandlers["integral"] = GenIntegral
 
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #6290, opened 2023-07-14
Sort SD card files by modification date

Matszwe02:
Sort SD card files by modification date descending to make selecting files convenient on displays.

I have a display that just pulls files from sd card and displays them as is. But when my printer has dozens of files, or even more, it gets more and more frustrating to try to select the proper one. The suggested solution in this PR sorts the sd card contents by modification date to make it easier to select the right file, because usually users want to print the freshly sliced one.

Without this feature I can't treat the `[display]` as a solution that replaces mainsail or fluidd. Having as few as 10 files makes it pretty hard, if not impossible to navigate the desired file, especially if I sliced it without looking on the exact gcode name

### DIFF
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -76,10 +76,15 @@ class VirtualSD:
             dname = self.sdcard_dirname
             try:
                 filenames = os.listdir(self.sdcard_dirname)
-                return [(fname, os.path.getsize(os.path.join(dname, fname)))
-                        for fname in sorted(filenames, key=lambda x: -os.path.getmtime(os.path.join(dname, x)))
+                return [
+                    (fname, os.path.getsize(os.path.join(dname, fname)))
+                        for fname in sorted(
+                    filenames,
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, there isn't much interest in changing the formatting or interaction of these older lcd screens.  Most of the recent focus has been on the newer interfaces that utilize Moonraker and the Klipper API server interface.

-Kevin

Matszwe02:
> Thanks. However, there isn't much interest in changing the formatting or interaction of these older lcd screens. Most of the recent focus has been on the newer interfaces that utilize Moonraker and the Klipper API server interface.
> 
> -Kevin

Right, but still Voron 0.2 comes with that display, not a newer one.

joaobarros:
> Thanks. However, there isn't much interest in changing the formatting or interaction of these older lcd screens. Most of the recent focus has been on the newer interfaces that utilize Moonraker and the Klipper API server interface.
> 
> -Kevin

This issue was one of the reasons I stopped using these screens (I still have them).
There are many people who still use them, the OP mentioned one example of recent printer that uses it.
It’s really a shame improvements like this get discarded.

PR #6236, opened 2023-06-01
Printer configs

SteveGotthardt:
Config: for Voxelab Aquia with v1.0.0 (X1, etc.) and v1.0.3 boards with HC32F460

These are updated with the company specific choices for the hc32f460 Kconfig

Signed-off-by: Steven Gotthardt <gotthardt@gmail.com>

PR #6291, opened 2023-07-14
hc32f460: Revert default clock back to 200MHz

Jookia:
Various build guides and documentation are based around the previous config default of 200MHz and don't instruct users to change this. The result of this is "timer to close" shutdown errors.

Change the default back to 200MHz to avoid nasty surprises.

### DIFF
--- a/src/hc32f460/Kconfig
+++ b/src/hc32f460/Kconfig
@@ -81,10 +81,10 @@ config STACK_SIZE
 
 choice "Clock Speed"
     prompt "Clock Speed"
-    config HC32F460_CLOCK_SPEED_168M
-        bool "168 MHz"
     config HC32F460_CLOCK_SPEED_200M
         bool "200 MHz"
+    config HC32F460_CLOCK_SPEED_168M
+        bool "168 MHz"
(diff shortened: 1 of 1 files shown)


KevinOConnor:
@SteveGotthardt - do you have any comments on this?

-Kevin

SteveGotthardt:
It doesn't matter to me. Most boards seem to use 168MHz. 
Once more printer configs are added that call out the speed will this be an issue?

KevinOConnor:
This only changes the default.  Ideally, the example printer config files would list the correct setting for each board.  If most boards use 168Mhz then I guess we should leave the default as 168Mhz.

-Kevin

Jookia:
I've already seen users follow instructions to install Klipper on these boards and have it break with weird errors because 200MHz is the assumed default.

KevinOConnor:
I understand.  But I think getting example configs committed to Klipper will be a better way to address the problem.  See #6236 .

-Kevin

Jookia:
The existing configs for the Kobra don't mention which configuration to use

PR #6292, opened 2023-07-17
heaters: adds temperature monitors

pedrolamas:
(This is a follow up on #6145 and discussion on PR #6205 and #6210)

This PR adds a new `available_monitors` entry to the `heaters` status object.

This allows the `TMC2240` to register itself as a temperature monitor so that Moonraker can then read the temperature and store the history data and provide it to the frontends.

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -245,6 +245,11 @@ object is available if any heater is defined):
   temperature sensors by their full config section names,
   e.g. `["extruder", "heater_bed", "heater_generic my_custom_heater",
   "temperature_sensor electronics_temp"]`.
+- `available_monitors`: Returns a list of all currently available
+  temperature monitors by their full config section names,
+  e.g. `["tmc2240 stepper_x"]`.  While a temperature sensor is always
+  available to read, a temperature monitor may not be available and
+  will return null in such case.
 
 ## idle_timeout
 
--- a/klippy/extras/heaters.py
+++ b/klippy/extras/heaters.py
@@ -230,6 +230,7 @@ class PrinterHeaters:
         self.gcode_id_to_sensor = {}
         self.available_heaters = []
         self.available_sensors = []
+        self.available_monitors = []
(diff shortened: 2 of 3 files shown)


pedrolamas:
@Arksine I believe this goes to what you proposed for minimum TMC2240 temperature support in Moonraker.

pedrolamas:
@thijstriemstra AFAIK, only the 2240 has temperature reporting capabilities, and looking at 5160 specification, I see no indication of such capability.

Nevertheless, if that is in fact possible, it will first need to be implemented on [tmc5160.py](https://github.com/Klipper3d/klipper/blob/master/klippy/extras/tmc5160.py)

thijstriemstra:
Yea I'm not sure if it has the ability either, guess this [tweet](https://twitter.com/BigTreeTech/status/1669663640914657282) implies there's a temperature sensor in there that controls the RGB light, thus it's also able to expose this sensor to klipper?

> Instant temperature status at a glance:
Built-in dual RGB lights and thermistor, EZ5160 RGB can monitor its own real-time temperature status through changes in the RGB lights. (6/8)

but I'm hijacking this PR, I'll shut up now :)

hcet14:
> @thijstriemstra AFAIK, only the 2240 has temperature reporting capabilities, and looking at 5160 specification, I see no indication of such capability.

Agree, only 2240 has true temperature reporting capabilities.

Arksine:
The changes look ok to me.  Perhaps we should document that the difference between a `monitor` and a `sensor` is that a `monitor` can have a null value?

pedrolamas:
> Perhaps we should document that the difference between a `monitor` and a `sensor` is that a `monitor` can have a null value?

Thanks @Arksine, I've now improved that section of the docs with a small note indicating the difference between `sensor` and `monitor` and the fact that a `monitor` will return null if it can't read the temperature.

KevinOConnor:
Thanks.

-Kevin

PR #6293, opened 2023-07-18
ci: update dependencies

thijstriemstra:
Signed-off-by: Thijs Triemstra <info@collab.nl>

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -6,10 +6,10 @@ jobs:
   build:
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Setup cache
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: ci_cache
         key: ${{ runner.os }}-build-${{ hashFiles('scripts/ci-install.sh') }}
@@ -21,7 +21,7 @@ jobs:
       run: ./scripts/ci-build.sh 2>&1
 
     - name: Upload micro-controller data dictionaries
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         name: data-dict
         path: ci_build/dict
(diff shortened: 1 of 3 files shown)


thijstriemstra:
There are a few dependencies that have newer versions but breaking backwards incompatible changes so I didn't update those but I can take a look at those as well if wanted.

hcet14:
> There are a few dependencies that have newer versions but breaking backwards incompatible changes so I didn't update those but I can take a look at those as well if wanted.

What could go wrong?

thijstriemstra:
> What could go wrong?

Alright.

> There are a few dependencies that have newer versions but breaking backwards incompatible changes so I didn't update those 

After another review I couldn't find any incompatible changes and all dependencies should now be using their latest version.

KevinOConnor:
Thanks.  We can certainly update the dependencies.  I'd like to understand if this is a proactive change or if there is a particular reason to make a change now?

-Kevin

thijstriemstra:
@KevinOConnor sorry for the late response.

>  I'd like to understand if this is a proactive change or if there is a particular reason to make a change now?

If you check a random ga build, e.g. https://github.com/Klipper3d/klipper/actions/runs/5828773956, it shows the following warning:

```
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/cache@v2, actions/upload-artifact@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```

So I try to keep these actions up to date for all my projects.

KevinOConnor:
Okay, thanks.

-Kevin

KevinOConnor:
This PR introduced a regression that resulted in actions/stale@v8 closing several PRs.  I committed c9aa7ac8 in an attempt to fix the regression.

-Kevin

thijstriemstra:
sorry @KevinOConnor, that's a pretty shitty regression. i hope it got solved with your additional changes.

PR #6295, opened 2023-07-19
screws_tilt_adjust: add max_diff to report status

meteyou:
This value is necessary to not show the helper dialog in the GUI when using MAX_DEVIATION.

Here is an example for it: https://github.com/mainsail-crew/mainsail/issues/1472


KevinOConnor:
Thanks.  In general it looks fine.  However, if the command parameter is `MAX_DEVIATION` could we call the status value `max_deviation`?

-Kevin

meteyou:
Yes of course! I will update this PR as soon as possible. I just use the same name like in the py file.

meteyou:
i have updated this PR and the PR on the Mainsail repo to `max_deviation`.

PR #6296, opened 2023-07-20
printer-anycubic-i3-mega-2017: drivers fan not turned on while moves without heating

skarasov:
Drivers/mainboard fan of Anycubic i3 Mega placed in [heater_fan ...] group. 
So it turned on only while heating but not while moves without heating.  
Mainboard/drivers fan must be in [controller_fan ...] group by default to prevent overheating.

### DIFF
--- a/config/printer-anycubic-i3-mega-2017.cfg
+++ b/config/printer-anycubic-i3-mega-2017.cfg
@@ -93,5 +93,5 @@ max_accel: 3000
 max_z_velocity: 10
 max_z_accel: 60
 
-[heater_fan stepstick_fan]
+[controller_fan drivers_fan]
 pin: PH4


JamesH1978:
Good morning, thank you for submitting a PR, Please could you refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md especially the section about the signed-off-by line

Thanks
James

skarasov:
printer-anycubic-i3-mega-2017: drivers fan not turned on while moves without heating

Drivers/mainboard fan of Anycubic i3 Mega placed in [heater_fan ...] group.
So it turned on only while heating but not while moves without heating.
Mainboard/drivers fan must be in [controller_fan ...] group by default to prevent overheating.

Signed-off-by: Sergiy Karasov <sergiy.karasov@gmail.com>

JamesH1978 (commented):
Seems fine to me, I can see where this could be a problem, if you do many moves without heaters on, but i cant see them heating up that much for non-print moves.

I would ask the original author for input, but this was first committed 5 years ago! but @mjedlinski if you are still around, do you concur with this change?

I will leave this comment before actual approval to see if anyone has any more input on the change.

Thanks
James

skarasov:
Printes also used for drawing, cutting... without heating.
Safety first!

PR #6297, opened 2023-07-20
idex_modes: COPY and MIRROR mode implementation

dmbutyugin:
This PR adds a new code for COPY and MIRROR modes for dual_carriage (IDEX). This is a bit of a rewrite of an existing dual_carriage code support. Instead of fiddling with different stepper_kinematics allocs to support dual carriage and their reallocation, the new code introduces a new stepper_kinematics class for dual (and primary) carriage, which performs the following coordinate transforms:
```
x_new = x * x_scale + x_offset;
y_new = y * y_scale + y_offset;
```
Different modes for the carriages are achieved by setting the parameters accordingly for the corresponding axis (the other axis has scale == 1.0 and offset == 0.0):
- PRIMARY: scale = 1.0, offset = 0.0;
- COPY: scale = 1.0, offset = <carriages_distance>;
- MIRROR: scale = -1.0, offset = <...>;

I believe this approach is simpler and a bit more extensible.

For regular step generation, the call sequence is approximately the following:
```
itersolve -> dual_carriage_calc_position -> *stepper*calc_position
                          |
                          V
                    move_get_coord
```
With input shaper enabled, it looks approximately as follows:
```
itersolve -> shaper*calc_position -> dual_carriage_calc_position -> *stepper*calc_position
                        |
                        V
                  move_get_coord
```

Separately, I added a couple of new commands (SAVE_/RESTORE_IDEX_STATE) and modified an existing SET_DUAL_CARRIAGE command in a backwards-compatible manner to support new IDEX modes, however I did change the exported status for the dual carriage. Any feedback here is welcome. [This](https://klipper.discourse.group/t/support-of-copy-and-mirror-modes-for-idex/9487) is a thread on Klipper Discourse having more details about this feature.


DrumClock:
Hi, I tested cartesian kinematics on IDEX COPY and MIRROR mods. 
Movement and guard collision carriages is fine.
I didn't find anything wrong with the kinematics.

dmbutyugin:
Given the amount of testing this code received, I am confident that pretty much all bugs have been fished out. So, it is now in the state ready to be reviewed and merged, if the review goes well.

KevinOConnor:
Thanks.  In general it seems fine to me.  I have a few minor comments.

1. It seems this PR has a few commits that should be squashed.  Did you want me to squash them all upon commit, or are you planning to do a `git rebase`?
2. The latest code defines `dual_carriage_alloc()` using K&R function syntax.  I can't tell if that is a typo or if you are doing that to accomplish something very subtle.  I think we should try to avoid K&R functions.  Can you elaborate on this part of the change?
3. I'd prefer if we didn't add a new `restore_velocity` config option.  I think config parameters to set default command parameters unnecessarily increase the config complexity and in general are a bit "goofy".  I think it is reasonable that if a user wants to use MOVE on RESTORE_IDEX_STATE that they provide an explicit MOVE_SPEED on that command.
4. How about `DUAL_CARRIAGE_SAVE_STATE` and `DUAL_CARRIAGE_RESTORE_STATE` instead of `SAVE_IDEX_STATE` and `RESTORE_IDEX_STATE`?

@Tircown - do you have any comments on this PR?

-Kevin

dmbutyugin:
Thank you Kevin for your feedback. So, here are my comments before I implement the changes for the context:



> 1. It seems this PR has a few commits that should be squashed.  Did you want me to squash them all upon commit, or are you planning to do a `git rebase`?

I'll push one or more commits with the suggested changes. So, if someone makes last-minute checks it would be better to have them as a separate commits in my branch, but please squash them into a single commit upon merging, if you're OK with that.

> 2. The latest code defines `dual_carriage_alloc()` using K&R function syntax. I can't tell if that is a typo or if you are doing that to accomplish something very subtle. I think we should try to avoid K&R functions. Can you elaborate on this part of the change?

Thanks for catching this, I didn't intend anything special, I guess I'm just writing too much C++ code and didn't notice this. I'll push a change to fix that.

> 3. I'd prefer if we didn't add a new restore_velocity config option. I think config parameters to set default command parameters unnecessarily increase the config complexity and in general are a bit "goofy". I think it is reasonable that if a user wants to use MOVE on RESTORE_IDEX_STATE that they provide an explicit MOVE_SPEED on that command.

The context here as follows. Restoring IDEX/DC state in COPY/MIRROR mode without restoring the position doesn't make a lot of sense, since the user will not be able to continue the print. This is because if the carriages have moved since the time the state was stored, they'll have a different offset against each other, and they'll continue then to move in a wrong manner. I see that most of the users could use these commands in custom homing routines (to properly restore COPY/MIRROR mode after homing), but these commands can be useful in various other scripts/macros. So, the default for `RESTORE_IDEX_STATE` is `MOVE=1`. And I didn't want different behavior of this command in case it restores COPY/MIRROR mode vs other modes. So, I needed some velocity to do that. I see a few options here:
1.  change the default to not move;
2. keep the default to move but do not introduce the velocity parameter, instead have a reasonable default fall back velocity;
3. keep the current implementation.

To be clear, I'm fine with either of the options. Maybe option (2) makes the most sense (and in fact today the code defaults to the rail homing speed if no other velocity is provided). Maybe you have another good suggestion for the default speed? And in general, please let me know your preference here, I'll implement it.



> 4\. How about `DUAL_CARRIAGE_SAVE_STATE` and `DUAL_CARRIAGE_RESTORE_STATE` instead of `SAVE_IDEX_STATE` and `RESTORE_IDEX_STATE`?

I don't have a strong opinion here, so I can rename these commands. Maybe just for consistency with `SET_DUAL_CARRIAGE` they could be named as `SAVE_DUAL_CARRIAGE_STATE` and `RESTORE_DUAL_CARRIAGE_STATE`, but I'm fine either way.

KevinOConnor:
> please squash them into a single commit upon merging, if you're OK with that.

That's fine.

> Maybe option (2) makes the most sense (and in fact today the code defaults to the rail homing speed if no other velocity is provided). 

That seems fine to me.  Using the rail homing speed (if an explicit speed is not provided in the command) seems fine to me.

> Maybe just for consistency with SET_DUAL_CARRIAGE they could be named as SAVE_DUAL_CARRIAGE_STATE and RESTORE_DUAL_CARRIAGE_STATE, but I'm fine either way.

I don't have a strong preference.  If I were to "flip a coin" I'd start these new commands with DUAL_CARRIAGE.

Thanks again.
-Kevin

DrumClock:
Hi, @KevinOConnor , @dmbutyugin 
for me (who doesn't know much English) it's more embarrassing:

SAVE_DUAL_CARRIAGE_STATE
RESTORE_DUAL_CARRIAGE_STATE

as is the case with:
SAVE_GCODE_STATE
RESTORE_GCODE_STATE

SAVE_VARIABLE

dmbutyugin:
OK, thanks all for the feedback. @KevinOConnor I've addressed all 4 points that you've raised. For the commands names, I went with `SAVE_DUAL_CARRIAGE_STATE` and `RESTORE_DUAL_CARRIAGE_STATE`, as this is more in line with the existing commands such as `SET_DUAL_CARRIAGE`, `SAVE_GCODE_STATE` and `RESTORE_GCODE_STATE`, and I don't think we'd want to rename those (I also don't think it is worthwhile renaming just `SET_DUAL_CARRIAGE`).

Tircown:
Hi, I haven't tested it yet but I reviewed the code. Overall all seems fine to me and I am quite happy if COPY and MIRROR modes are finally available natively.
`SAVE_DUAL_CARRIAGE_STATE`, `RESTORE_DUAL_CARRIAGE_STATE` are more in line with older commands and are totally clear to anyone who are familiar with Klipper.
Maybe the documentation or a config sample should mention the official RepRap G-Code command for dual carriages modes `M605 S[mode] X[offset]` with mode 0= full control, 1= auto-park, 2= Copy, 3= Mirror.
Thanks @dmbutyugin for your work.

KevinOConnor:
Thanks.

-Kevin

HelgeKeck:
looks like this allows me to set the primary toolhead, is this correct? I am using a V-Core IDEX and it has a inverted belt path

HelgeKeck:
very nice work, thank you guys. But i have some problems with it.

1. a RatRig V-Core IDEX has a inverted belt path, this means Carriage 0 and 1 needs to be swapped, I cant use GCode Offset and need to use set kinematic position to make it work. But it still works even if its not ideal

2. i also want to be able to query the current IDEX mode with `{% set idex_mode = printer["dual_carriage"].mode|lower %}`, this worked with the other PRs

i can live with 1, but i really would like to have point 2 fixed if this is possible

thank you

DrumClock:
> 2\. i also want to be able to query the current IDEX mode with `{% set idex_mode = printer["dual_carriage"].mode|lower %}`, regardless the active toolhead, this worked with the other PRs

HI,
point 2 should also be resolved as follows:

```
{% set idex_mode = 'copy' if printer['dual_carriage'].carriage_1 = 'COPY'
              else 'mirror' if printer['dual_carriage'].carriage_1 = 'MIRROR'
              else 'full_control' %}
```

HelgeKeck:
yes, i am awar eof that, was jsut much simpler and more userfriendly before

HelgeKeck:
this should be part of the get_status function and returning it as `mode`

DrumClock:
> this should be part of the get_status function and returning it as `mode`

Yes I know, but before it only worked for corexy and hybrid. So for those who have cartesian it was useless anyway. you have to be able to advise yourself and work with what is.

HelgeKeck:
ideally mode would  just return:

'copy'
'mirror'
'carriage_0'
'carraige_1'

DrumClock:
Yes I agree with you

dmbutyugin:
@HelgeKeck, @DrumClock, my 2 cents:

- I don't have a very strong preference how IDEX status should be exported in the `dual_carriage` object. I accept that there could be different PoV and desires to have the status in some specific format.
- The current Klipper IDEX code is not the same as the one from Tircown. So, _internally_ it must use the per-carriage modes that we see today and it cannot use the same bits as what Tircown's code uses.
- I just went with the straightforward option - export the internal status as-is. I also think it is sufficiently versatile and extensible in the future (for example, supporting truly independent carriages on two rails, something like dualgantry-corexy). It does not mean that the code could not perform some conversion internally and export a different status. But I did not know the ideal export status (I still don't know it - for example, the status information from Tircown's branch may be ideal in some scenarios but less ideal in other scenarios of its use).
- Since the status code can be converted also in the macro, I thought that if one desires a different status, they could perform such a conversion (or just implement an appropriate condition in `if` themselves). That is probably the easiest option. However, by popular demand the exporting status can be changed as well. So the question could also be put like this: how do you want and plan to use `printer.dual_carriage.mode` status? (and why just looking at `printer.dual_carriage.carriage_1` is not sufficient?)

DrumClock:
I wrote above. I can handle it in the macro.

The "mode" reference used to only work in corexy and I have cartesian.

So I didn't use it. I personally adapt to macros.

DrumClock:
For example, this was the reference in the last version I tested for @Tircown

`printer['toolhead'].dual_carriage = {'active_carriage': 'CARRIAGE_0', 'init_mode_positions': (None, None), 'mode': 'FULL_CONTROL'}`

However, I use my own variables, so I personally don't care what is in the reference. :)

HelgeKeck:
yes, i am aware that you can already use it, in fact i already do. i just found it a bit cleaner before. however, i understand the explanation from @dmbutyugin, makes totally sense. i can live with it

PR #6298, opened 2023-07-21
config: Add 1LC example configuration and accompanying flashing script

Wulfsta:
This adds an example config and a flashing script for the Duet3D Toolboard 1LC. 

@eamaclean please have a look at this when you get a chance, would like your thoughts.

### DIFF
--- a/config/sample-duet3-1lc.cfg
+++ b/config/sample-duet3-1lc.cfg
@@ -1,0 +1,81 @@
+# This file contains common pin mappings for the Duet3 1LC. To use
+# this config, the firmware should be compiled for the SAMC21G18 with:
+# Bootloader offset of "No Bootloader"
+# Clock Reference of "25 Mhz crystal" if the board version is v1.1 or later
+# Clock Reference of "12 Mhz crystal" if the board version is v1.0 or earlier
+# Communication interface of "CAN bus (on PA25/PA24)"
+
+# To flash the board use a debugger, or use a raspberry pi and follow
+# the instructions at docs/Bootloaders.md fot the SAMC21. You may
+# supply power to the 1LC by connecting the 3.3v rail on the Pi to the
+# 5v input of the SWD header on the 1LC.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+
+# Pins for reference, v1.3 board:
+# Driver Step Pin - PA27
+# Driver Dir Pin  - PA28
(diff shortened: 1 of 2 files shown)


config/generic-duet3-1lc.cfg
```diff
@@ -0,0 +1,80 @@
+# This file contains common pin mappings for the Duet3 1LC. To use
+# this config, the firmware should be compiled for the SAMC21G18 with:
+# Bootloader offset of "No Bootloader"
+# Clock Reference of "25 Mhz crystal"
```
eamaclean, line 4: Should mention the 12MHz crystal on 0.6 and 1.0 boards

config/generic-duet3-1lc.cfg
```diff
@@ -0,0 +1,80 @@
...
+# Communication interface of "CAN bus (on PA25/PA24)"
+
+# To flash the board use a debugger, or use a raspberry pi and follow
+# the instructions at klipper/scripts/flash-atsamc21.sh. You may
+# supply power to the 1LC by connecting the 5v rail on the Pi to the
```
eamaclean, line 9: I would suggest using the 3.3V supply out of an abundance of caution - Pi GPIO is fragile.

config/generic-duet3-1lc.cfg
Wulfsta: Good point.

config/generic-duet3-1lc.cfg
Wulfsta: I haven’t tried this, honestly not sure if it will work…

eamaclean: It works - the CAN transceiver won't operate at 3.3V but the MCU runs just fine.

KevinOConnor:
Thanks.  The convention is to add toolhead board configs using the "sample" prefix and not the "generic" prefix (as all "generic" configs would need to be added to the regression test cases, which a toolhead board can pass on its own) - https://www.klipper3d.org/Example_Configs.html

I'm not sure it makes sense to add a script for flashing this toolhead board, as it depends on installing/building openocd which I suspect most users wont be prepared to do.  Consider adding pertinent information to docs/Bootloaders.md instead (which already has information on installing and using openocd).

Cheers,
-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6299, opened 2023-07-22
atsamd: Optimize timer for samc21

Wulfsta:
This optimizes the timer code for samc21. Tested by running single stepper benchmark as follows:
```
allocate_oids count=1
config_stepper oid=0 step_pin=PA27 dir_pin=PA28 invert_step=-1 step_pulse_ticks=0
finalize_config crc=0
SET ticks 97
reset_step_clock oid=0 clock={clock+freq}
set_next_step_dir oid=0 dir=0
set_next_step_dir oid=0 dir=1
```
Prior to this change, the minimum value for `ticks` was `99`. This change is marginally faster with a minimum `ticks` value of `97`. I'm not sure this optimization was worth it, but at least we know these values now. Notably, this is slower than the recorded benchmark for samd21, but that chip does not require syncing this field before a read.

### DIFF
--- a/src/atsamd/timer.c
+++ b/src/atsamd/timer.c
@@ -41,8 +41,8 @@ timer_read_time(void)
 {
 #if CONFIG_MACH_SAMC21
     TCp->COUNT32.CTRLBSET.reg |= TC_CTRLBSET_CMD(TC_CTRLBCLR_CMD_READSYNC_Val);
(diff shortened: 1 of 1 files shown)


Wulfsta:
I take it back, rebuilt and reran after making the branch and it's back to a `ticks` of `99`; I must have built on a different compiler or something and didn't realize? I'm closing this PR because I don't see any reason to bother with making it deterministic. @KevinOConnor forgive the ping, I know you were interested in the numbers.

PR #6300, opened 2023-07-23
Add config for printer Creality CR10s Pro V2

flatline-84:
Created a base config for the Creality CR10s Pro v2 as described here: https://www.creality.com/products/cr-10s-pro-v2-3d-printer

This will allow users with this printer to more easily setup Klipper firmware.

### DIFF
--- a/config/printer-creality-cr10s-pro-v2-2020.cfg
+++ b/config/printer-creality-cr10s-pro-v2-2020.cfg
@@ -69,10 +69,10 @@ max_extrude_only_accel: 1000.0
 heater_pin: PB4 #ar10
 sensor_type: EPCOS 100K B57560G104F
 sensor_pin: PK5 #analog13
-control = pid #calibrated in Klipper, you will need to run this for your machine
-pid_kp = 28.359
-pid_ki = 1.616
-pid_kd = 124.426
+control: pid #calibrated in Klipper, you will need to run this for your machine
+pid_kp: 28.359
+pid_ki: 1.616
+pid_kd: 124.426
 min_extrude_temp: 170
 min_temp: 5
 max_temp: 275
--- a/config/printer-creality-cr10s-pro-v2-2020.cfg
+++ b/config/printer-creality-cr10s-pro-v2-2020.cfg
@@ -1,11 +1,7 @@
 # This file contains pin mappings for the Creality CR-10S Pro V2. To use
 # this config, the firmware should be compiled for the AVR atmega2560.
 
-# Based on this Desuuuu config here but modified:
-#   https://github.com/Desuuuu/DGUS-reloaded-Klipper-config/blob/master/base/creality-cr10sprov2.cfg
-
 # See docs/Config_Reference.md for a description of parameters.
-
 
 ## General Config
 [mcu]
@@ -103,7 +99,7 @@ max_temp: 140
 [bltouch]
 sensor_pin: ^PD2
 control_pin: PB5
-stow_on_each_sample: False # BE CAREFUL WITH THIS
+stow_on_each_sample: False # Be careful
 probe_with_touch_mode: True
 x_offset: -27
 y_offset: -2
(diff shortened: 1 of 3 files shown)


JamesH1978:
Thankyou for submitting a PR, please could you refer to point 3 in the "what to expect" section in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and add a signed-off-by line and point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md and include the regression tests.

Thanks
James

flatline-84:
Hi James, I have addressed all the points. Hope this is correct

config/printer-creality-cr10s-pro-v2-2020.cfg
```diff
@@ -0,0 +1,158 @@
...
+serial: /dev/serial/by-id/<YOUR_USB_ID>
+
+[printer]
+kinematics: cartesian
+max_velocity: 200
+max_accel: 1500
+max_z_velocity: 10
+max_z_accel: 100
+
+[gcode_macro G29]
```
JamesH1978, line 17: Please do not include macros in a config that do not pertain to base functionality of a printer, BED_MESH_CALIBRATE is perfectly valid gcode in its own right.

Thanks
James

config/printer-creality-cr10s-pro-v2-2020.cfg
```diff
@@ -0,0 +1,158 @@
...
+
+[filament_switch_sensor e0_sensor]
+switch_pin: PE4 #ar2
+pause_on_runout: False
+runout_gcode:
+  PAUSE_PARK
+
+## Calibrating the Bed
+# this is for the bltouch to calibrate the bed
+[screws_tilt_adjust]
```
JamesH1978, line 134: You have the same screw positions for this and bed_screws, this is not likely right, as one needs absolute coordinates when probe is over the screw(screws_tilt_adjust), and the other needs absolute coordinates when nozzle is over screw(bed_screws).

Thanks
James

config/printer-creality-cr10s-pro-v2-2020.cfg
```diff
@@ -0,0 +1,158 @@
...
+screw2_name: front right screw
+screw3: 295,280
+screw3_name: rear right screw
+screw4: 40,280
+screw4_name: rear left screw
+speed: 50
+horizontal_move_z: 10
+screw_thread: CW-M3
+
+[bed_screws]
```
JamesH1978, line 147: see above comment

JamesH1978 (changes requested):
Thank you for making the changes i asked, please see my comments.

Thanks
James

config/printer-creality-cr10s-pro-v2-2020.cfg
flatline-84: Done

config/printer-creality-cr10s-pro-v2-2020.cfg
flatline-84: Fixed

JamesH1978 (approved):
Looks fine to me.

Thanks
James

KevinOConnor:
Thanks.  Could you also update the pid lines to use `option: value` syntax instead of `option = value` syntax?  https://www.klipper3d.org/Example_Configs.html point 7d.

-Kevin

flatline-84:
Good catch Kevin. Thanks, updated now

JamesH1978 (approved):
Apologies for missing the syntax,

Still looks good

Thanks
James

KevinOConnor:
Alas, the `option = value` is still present in the heater_bed section.

Cheers,
-Kevin

flatline-84:
I could have sworn I did a find + replace but apparently not. Should be properly good to go now

KevinOConnor:
Thanks.

-Kevin

PR #6301, opened 2023-07-25
scripts: Support both H743 and H723 BTT SKR3 boards sdflash

dewhisna:
BigTreeTech is shipping the SKR-3 Controller Board now with the STM32H723 micro in addition to the existing STM32H743 version.

This adds the H723 target to the spi-flash script and renames the existing H743 variant to be consistent with other entries and disambiguate between the two.

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20230810: The flash-sdcard.sh script now supports both variants of the
+Bigtreetech SKR-3, STM32H743 and STM32H723. For this, the original tag
+of btt-skr-3 now has changed to be either btt-skr-3-h743 or btt-skr-3-h723.
+
 20230729: The exported status for `dual_carriage` is changed. Instead of
 exporting `mode` and `active_carriage`, the individual modes for each
 carriage are exported as `printer.dual_carriage.carriage_0` and
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -87,8 +87,15 @@ BOARD_DEFS = {
         'spi_bus': 'spi3a',
         'cs_pin': 'PA15'
     },
-    'btt-skr-3': {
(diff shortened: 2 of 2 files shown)


KevinOConnor:
Thanks.  It's hard for me to review changes to board_defs.py - maybe @JamesH1978 or @Arksine could provide some comments.

-Kevin

JamesH1978:
It's exactly the same pin out for both, so i would suggest using an alias to the original(renamed), rather than a whole new section.

Thanks
James

dewhisna:
But it's NOT an alias... It's a different microprocessor and has a different microprocessor value.  And no other entry has the H723.  So there's nothing it's an alias of.

The only way to be an alias would be if the microprocessor type is useless in this script, in which case, it should be removed from all of the entries and many more become aliases.

JamesH1978:
Apologies, i see what you mean, wasnt taking the different processors into account. @Arksine would have to comment on where the mcu is used, if anywhere on a SDIO connection

Arksine:
Overall the PR looks good to me with one sticking point...changing the name from `btt-skr-3` to `btt-skr-3-h743` will break update workflows for existing owners of the SKR3, they will get an error when they attempt to update using `btt-skr-3`.  If we make that change I think it needs to be noted in `config_changes.md`.

dewhisna:
Fair enough.  I added an entry in `Config_Changes.md` as requested, with text phrased similar to changes like this one.  And I rebased this PR to prevent a merge conflict on that file.

Thanks for clarifying how that should be documented.  I wasn't sure how that was handled and only knew that leaving the original entry unchanged was even more confusing to end-users as it required a priori knowledge that the original SKR3 was the H743.

Arksine (approved):

Arksine:
Looks good to me.  Thanks.

KevinOConnor:
Thanks.

-Kevin

locki-cz:
@dewhisna can you add also alias for Octopus board? https://github.com/Klipper3d/klipper/pull/6306

PR #6302, opened 2023-07-29
Extruder: Allow gaps in extruder numbering.

viesturz:
When an extruder is not available, IE, it's broken or otherwise, the strict incremental numbering makes it cumbersome to reconfigure the printer.

This allows gaps in the extruder numbering.
The only downside I can see is slight extra work on printer startup. That should amount to <1ms of real life latency.

### DIFF
--- a/klippy/extras/motion_report.py
+++ b/klippy/extras/motion_report.py
@@ -261,7 +261,7 @@ class PrinterMotionReport:
                 ename = "extruder"
             extruder = self.printer.lookup_object(ename, None)
             if extruder is None:
-                break
+                continue
             etrapq = extruder.get_trapq()
             self.trapqs[ename] = DumpTrapQ(self.printer, ename, etrapq)
         # Populate 'trapq' and 'steppers' in get_status result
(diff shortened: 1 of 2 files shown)


Arksine:
This change might break front ends and/or applications connected through the webhooks API.  As things stand now it wouldn't be unreasonable for them to make UI decisions based on the assumption that the primary tool will always be named `extruder`, tool 1 `extruder1`, etc.  I cant say for certain that this is the case, but it is something that likely needs thorough testing.

viesturz:
True. Fluidd so far seems totally fine with this.
But more testing would be useful indeed.

This still forces to have a primary extruder named extruder. Only the rest
can be arbitrary ordered.

This does not break any existing configurations, just allows new
configurations that might be incompatible.

Do you think this can be incremental process of Klipper relaxing the
restriction followed by frontends updating if/when this is an issue?

KevinOConnor:
Thanks.  As high-level feedback, I'd be leery of making this change for fear of it introducing a regression somewhere.  I think it's reasonable to ask users to number their extruders in ascending order.

-Kevin

PR #6303, opened 2023-07-29
Toolhead: Use dict for step generation flush times.

viesturz:
Makes the API between toolhead and extruder, input shaper more robust, avoiding the need to track the old delay.

### DIFF
--- a/klippy/extras/input_shaper.py
+++ b/klippy/extras/input_shaper.py
@@ -135,16 +135,13 @@ class InputShaper:
             is_sk = self._get_input_shaper_stepper_kinematics(s)
             if is_sk is None:
                 continue
-            old_delay = ffi_lib.input_shaper_get_step_generation_window(is_sk)
             for shaper in self.shapers:
                 if shaper in failed_shapers:
                     continue
                 if not shaper.set_shaper_kinematics(is_sk):
                     failed_shapers.append(shaper)
             new_delay = ffi_lib.input_shaper_get_step_generation_window(is_sk)
-            if old_delay != new_delay:
-                self.toolhead.note_step_generation_scan_time(new_delay,
-                                                             old_delay)
+            self.toolhead.note_step_generation_scan_time(self, new_delay)
         if failed_shapers:
             error = error or self.printer.command_error
             raise error("Failed to configure shaper(s) %s with given parameters"
--- a/klippy/kinematics/extruder.py
+++ b/klippy/kinematics/extruder.py
@@ -70,15 +70,11 @@ class ExtruderStepper:
         self.stepper.set_trapq(extruder.get_trapq())
         self.motion_queue = extruder_name
     def _set_pressure_advance(self, pressure_advance, smooth_time):
-        old_smooth_time = self.pressure_advance_smooth_time
-        if not self.pressure_advance:
-            old_smooth_time = 0.
         new_smooth_time = smooth_time
         if not pressure_advance:
             new_smooth_time = 0.
         toolhead = self.printer.lookup_object("toolhead")
-        toolhead.note_step_generation_scan_time(new_smooth_time * .5,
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Thanks.  Seems fine to me.  @dmbutyugin do  you have any comments?

-Kevin

KevinOConnor:
On a second look though, I'm not sure we can avoid the call to `flush_step_generation()` as the callers may be expecting this code to always flush the step queue.  Have you verified that all callers already flush the step queue?

-Kevin

dmbutyugin:
TBH, I do not have strong opinion either way. I think the current code isn't that complicated and it is not troublesome to get the old window value. And the new code makes it marginally better. Although, perhaps the author of this PR has stronger reasons to make this change.

Separately, I do agree that `flush_step_generation()` call is required, at least it seems that the extruder code misses that now, and issuing a command `SET_PRESSURE_ADVANCE SMOOTH_TIME=...` with a different (larger?) value than the current smooth_time would result in a stepcompress error.

viesturz:
Thanks, make the flush_step_generation unconditional.

I agree it's a minor improvement. From POV of someone that is fairy new to the codebase, this was certainly a head scratcher and a possible footgun unless you pay extra attention.

rednaxal:
After this update I've started getting a constant "**Internal error on command:"G1"**" on one of my printers. 
[bug.zip](https://github.com/Klipper3d/klipper/files/13060655/bug.zip)
I'm attaching the klippy.log and a minified version of a gcode that results on that error all the time (this is a part of a wipe tower generated by Prusa Slicer).

Sineos:
Possibly similar report here: https://klipper.discourse.group/t/internal-error-in-stepcompress-after-upgrading-to-v0-11-0-303-on-cr-10s-with-raspberry-pi-4-and-ubuntu-arm/11090

viesturz:
Hi, trying to repro this.

viesturz:
@rednaxal Working just fine here, if you can create a new bug for this, https://www.klipper3d.org/Contact.html#i-found-a-bug-in-the-klipper-software

rednaxal:
@viesturz Have tried your patch with no luck. The issue is still there. 
A rollback to 1 commit behind - fixes the issue. So there is definitely something wrong with this particular commit. Will create a new bug report now if necessary.

viesturz:
Thanks, please do.
And please document your setup as much as possible, there seems to be some
hardware/configuration specific trigger.

JamesH1978:
We have now seen this multiple times(at least 5 so far) causing stepcompress issues on G0/G1/G28 in the discord, fixed with either disabling IS or rolling back 1 commit to 302 so this one is not included.

Thanks
James

viesturz:
Ah, found the issue, creating a fix pull request in a few hours.

JamesH1978:
Thanks, the user i am testing with at the moment says his are both set to mzv though.

James

KevinOConnor:
I reverted this PR.

Looking a little closer at the code, the input_shaper.py code calls `note_step_generation_scan_time()` in a loop - once for each stepper.  Thus, using the `InputShaper` instance itself is not unique.

We can look again at this after the v0.12.0 release.  It might be best to just leave things as they are though, as the cleanup was relatively minor.

-Kevin

viesturz:
If someone can verify #6372 fixes it?

KevinOConnor:
FYI, as part of PR #6410 I took a look at changing the toolhead code to auto-detect the low-level itersolve scan windows.  Thus not requiring input_shaper nor pressure_advance to explicitly register/unregister the scan windows with the toolhead code.

If interested, the demo code is up at https://github.com/KevinOConnor/klipper-dev/tree/work-scanwin-20231128 .  I'm not sure it's a worthwhile change.  Might be a "win" if future code makes use of the low-level scanning windows.

-Kevin

PR #6305, opened 2023-07-29
tmc: Adding UART interface support to tmc2240

FrYakaTKoP:
Adding UART communication option for tmc2240 driver chips, which support both spi and uart. To achieve this a new config field has been added `interface` with the options `spi` (default/fallback) and `uart`.

The code is tested by atleast 3 users on a new upcoming board, which uses UART on the 2240 drivers to save mcu pins.

Signed-off-by:  Christoph Frei <fryakatkop@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3362,6 +3362,7 @@ the name of the corresponding stepper config section (for example,
 
 ```
 [tmc2240 stepper_x]
+interface: spi
 cs_pin:
 #   The pin corresponding to the TMC2240 chip select line. This pin
 #   will be set to low at the start of SPI messages and raised to high
@@ -3472,6 +3473,23 @@ run_current:
 #   "sensorless homing". (Be sure to also set driver_SGT to an
 #   appropriate sensitivity value.) The default is to not enable
 #   sensorless homing.
+```
+
+### [tmc2240] UART
+
+Configure a TMC2240 stepper motor driver via UART. To use this
+feature, define a config section with a "tmc2240" prefix followed by
+the name of the corresponding stepper config section (for example,
+"[tmc2240 stepper_x]").
+
+See the [TMC2240 config reference](Config_Reference.html#tmc2240) for additional
+configuration fields.
+
+```
+[tmc2240 stepper_x]
+interface: uart
+uart_pin: 
+#diag_pin:
 ```
 
 ### [tmc5160]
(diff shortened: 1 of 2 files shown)


cneshi:
I have also tested this PR on a Voron2.4 without any issues so far

KevinOConnor:
Thanks.  Looks like useful functionality!  I'd ask for a couple of minor changes though:
1. I think it would be more in keeping with other Klipper modules if the uart vs spi mode was auto-detected.  That is, if `uart_pin` is found then use uart mode, otherwise use spi mode (`if config.get('uart_pin', None) is not None: ...`).
2. I don't think we should add another config section to Config_Reference.md .  I think the existing `[tmc2240]` doc section should be updated to include information on `uart_pin` (and associated config options).

Cheers,
-Kevin

FrYakaTKoP:
Thanks for the helpful feedback. It was fairly simple to implement. I'm uncertain if the rather minimalistic config reference is enough, but config_helper should guide the user pretty well.

Please let me know if i could improve further.

-FrY

JNP-1:
Works for me as expected

KevinOConnor:
Thanks.  In general it seems fine.  Though there is still an `interface` reference in Config_Changes.md.  I'll be travelling for a couple of weeks but I'll look to commit when I return.

-Kevin

FrYakaTKoP:
I'm actually confused by your last message. I had never touched Config_Changes.md. Anyway we can figure this out when you get back. Enjoy your travel!

-FrY

Travis90x:
So now we can use uart_pin in tmc2240 section?
My experiece many weeks ago:
https://www.facebook.com/groups/395452440884132?multi_permalinks=1698169560612407


362928950_10227907699991553_5075929394662401082_n

docs/Config_Reference.md
```diff
@@ -3355,13 +3355,14 @@ run_current:
...
 
-Configure a TMC2240 stepper motor driver via SPI bus. To use this
+Configure a TMC2240 stepper motor driver via SPI bus or UART. To use this
 feature, define a config section with a "tmc2240" prefix followed by
 the name of the corresponding stepper config section (for example,
 "[tmc2240 stepper_x]").
 
 ```
 [tmc2240 stepper_x]
+interface: spi
```
KevinOConnor, line 3365: This line added to Config_Reference.md seems to no longer be accurate.

-Kevin

FrYakaTKoP:
good catch i missed that one., thanks!

KevinOConnor:
Thanks.

-Kevin

PR #3937, opened 2021-02-15
New fan configuration option for lower power bounds

towe96:
module: Modify fan behavior on lower power bounds

Adds two options to fans to change how they operate at lower requested speeds.
min_power is an alternative to off_below. Fans speeds above 0 and below min_power are raised to min_power to ensure the fan keeps running after the kick_start_time.

With power_scaling set to True, the requested fan speeds (e.g. from M106 S0 to M106 S255 for the part cooling fan) will be scaled from min_power to max_power instead.

Signed-off-by: Tobias Weiß <t.weiss@bk.ru>

PR #3208, opened 2020-08-19
fan: Scale fan speed between off_below and max_power

pfemiani:
New relative_power fan setting will allow the scaling of fan speed requests
between  off_below and max_power. For example, if off_below is .3 and
max_power is .9 and a fan speed of 20% is requested, then the fan power
will be set to 42%). 

With this setting enabled, once you have off_below and
max_power properly calibrated, it is much easier to request fan speeds
between 0% and 100% (or 0-255) via slicer for example and have the fan
behave as intended without having to be mindful of what speed the fan
will be under (ie not started) or over powered.

Signed-off-by: Peter Femiani <peter@femiani.com>

PR #6307, opened 2023-08-01
Normalising Fan PWM power

kubaracek:
What this does
---
It makes it possible to set 1% for a fan such as CPAP with a large off_below value and the fan will run at its lowest power at this setting

image

Why
---
Fans with quite high value for `off_below` (CPAP for example) are currently in klipper bit tricky to operate and people use several tricks either in slicer/gcode/macros to scale the values as requesting a 10% fan for cpap that might start at 16% or 30% gets ignored by klipper and fan stays off. When fan is properly configured in klipper and I request a fan to run at 1% of its speed I expect it to actually move, which is not the case currently. So the slicer/gcode needs to have a 'low level' knowledge of my hardware (when my fan actually will kick-in).

The value is actually currently partially scaled but between 0 and max_power. It's a bit odd.

How
---
This normalises the value between `off_below` and `max_power`.
When `off_below` would be set to `0.3` and max_power to `1.0`  a `10%` increment would have a value of `0.7` so requesting 10% of a fan will result in a value of `off_below + 0.7 = 0.37` and not `0.1` as it does currently which effectively has no effect.

Normalizing the value according to fan setting such that the CPAP and similar fans will kick in even at 1% so the whole range can be used in same fashion as with any other fan where `off_below=0`

Note
--

__These two notes bellow were addressed by renaming the parameter.__

~When testing this. Mainsail has a logic in place where they use `off_below` as a minimal setting for the fan using the fan slider, so setting the fan to say 2% will automatically set it to 30%
I will create a separate PR in Mainsail to remove this logic as it will no longer be needed.~

~After such update, users with `off_below > 0` who will not change their `M106` to lower the value will have their fans a bit louder, heh. They will generaly expect that their cpap might kick-in at around 16-30% so that's what they will have their lowest value set to in slicer and such. After the update it will be an actual 16-30% fan speed.~

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,20 @@ All dates in this document are approximate.
 
 ## Changes
 
+20230801: The setting `fan.off_bellow` has been changed to `fan.min_power`.
+However, this change will not affect users who do not utilize this setting.
+With this update, a PWM scaling has been introduced between `min_power` and
+`max_power`. Fans that require a higher `min_power` will now have access to
+their full "safe" power curve. By properly setting `min_power`, any fan
+(such as CPAP) should engage even at `M106 S1`. It's recommended to review
+your slicer/macros to adjust your fan speeds. Your previously designated 20%
+fan speed may no longer represent your lowest fan setting but will now
+correspond to an actual 20% fan speed.
+If you had previously set `max_power` to anything below 1.0 (default),
+it is advisable to recalibrate `min_power` and `kick_start_time` again, with
+the settings `min_power: 0` and `max_power: 1`.
+
+
 20230729: The exported status for `dual_carriage` is changed. Instead of
 exporting `mode` and `active_carriage`, the individual modes for each
 carriage are exported as `printer.dual_carriage.carriage_0` and
--- a/config/printer-artillery-sidewinder-x2-2022.cfg
+++ b/config/printer-artillery-sidewinder-x2-2022.cfg
@@ -76,15 +76,15 @@ max_temp: 130
 
 [fan]
 pin: PC8
-off_below: 0.1
+min_power: 0.1
 
 [heater_fan extruder]
 pin: PC7
-off_below: 0.2
+min_power: 0.2
 
 [controller_fan case]
 pin: PC6
-off_below: 0.3
+min_power: 0.3
 idle_speed: 0.0
 
 [temperature_sensor mainboard]
--- a/config/printer-geeetech-301-2019.cfg
+++ b/config/printer-geeetech-301-2019.cfg
@@ -127,7 +127,7 @@ heaters: extruder
 pin: multi_pin:extruder_fans
 heater: extruder
 max_power: 0.8
-off_below: 0.2
+min_power: 0.2
 shutdown_speed: 0
 
 [mcu]
--- a/config/printer-monoprice-select-mini-v2-2018.cfg
+++ b/config/printer-monoprice-select-mini-v2-2018.cfg
@@ -128,7 +128,7 @@ pin: PB8
 # chassis fan
 [controller_fan chassis_fan]
 pin: PB3
-off_below: 0.25
+min_power: 0.25
 
 [mcu]
 serial: /dev/ttyACM0
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2565,15 +2565,14 @@ Print cooling fan.
 pin:
 #   Output pin controlling the fan. This parameter must be provided.
 #max_power: 1.0
-#   The maximum power (expressed as a value from 0.0 to 1.0) that the
-#   pin may be set to. The value 1.0 allows the pin to be set fully
-#   enabled for extended periods, while a value of 0.5 would allow the
-#   pin to be enabled for no more than half the time. This setting may
-#   be used to limit the total power output (over extended periods) to
-#   the fan. If this value is less than 1.0 then fan speed requests
-#   will be scaled between zero and max_power (for example, if
-#   max_power is .9 and a fan speed of 80% is requested then the fan
-#   power will be set to 72%). The default is 1.0.
+#   The maximum power (0.0 to 1.0) that the pin may be set to. A value
+#   of 1.0 enables the pin fully for extended periods, while 0.5 allows
+#   it for no more than half the time. Use it to limit total power output
+#   (over extended periods) to the fan. This value is combined with
+#   min_power to scale fan speed. With `min_power` at 0.3 and
+#   `max_power` at 1.0, fan speed request scales between 0.3 (min_power)
+#   and 1.0 (max_power). Requesting 10% fan speed results in a value of
+#   0.37. Default is 1.0.
 #shutdown_speed: 0
 #   The desired fan speed (expressed as a value from 0.0 to 1.0) if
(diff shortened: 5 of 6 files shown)


JamesH1978:

Also you need to fix your whitespaces/length of line to pass build checks

Thanks
James

kubaracek:
@JamesH1978 It should be fine now. Let me just test everything once more and I think it's ready.

Sineos:
You might want to cross check with:
https://github.com/Klipper3d/klipper/pull/5360
https://github.com/Klipper3d/klipper/pull/4720
https://github.com/Klipper3d/klipper/pull/3937
https://github.com/Klipper3d/klipper/pull/3208
https://github.com/Klipper3d/klipper/pull/2165

kubaracek:
@Sineos Thanks. I looked into the implementation and they modify a lot the input parameter. This would cause troubles for say Mainsail and other UIs as if you set a slider to say 10% it would get translated to 50% if that makes sense. I'm only changing what actually goes to `self.mcu_fan.set_pwm` function and leave the input value unchanged.

What I see as a reoccurring pattern in the disscusions is that for some the `off_power` might be handy for a reason I don't fully understand honestly.

So what we could do is to add a new param `min_power` that would default to either 0 or `off_power`. Keep the logic of `off_power` and normalise the values between min and max. Though someone would need to explain to me, why he/she would desire to have these two values 🤔 I don't see the benefit just yet. Renaming the value might make sense though

kubaracek:
Actually I went though the discussions again and there seem not to be any objections to have both of the settings and actually preferable to have just one.

So do we agree to rename the off_power to min_power and scale the value between the two?

kdb424:
I would very much like to see a min/max power, and scale between them. Currently "50%" doesn't mean anything related to "50%" relevant to the fan, but relevant to pwm speed. For bare pins, that makes sense, but in the context of fans, I'd like to see 50% *mean* 50% of the available fan curve. Consider me onboard with this change in concept. Will actively be checking back to code review and do testing on my printers.

Sineos:
There is only one person you need to convince and this is @KevinOConnor 😉

kubaracek:
https://github.com/Klipper3d/klipper/assets/16261998/6bb516ee-1255-4496-9be9-3b5744271a25

docs/Config_Changes.md
```diff
@@ -8,6 +8,20 @@ All dates in this document are approximate.
...
+With this update, a PWM scaling has been introduced between `min_power` and
+`max_power`. Fans that require a higher `min_power` will now have access to
+their full "safe" power curve. By properly setting `min_power`, any fan
+(such as CPAP) should engage even at `M106 S1`. It's recommended to review
+your slicer/macros to adjust your fan speeds. Your previously designated 20%
+fan speed may no longer represent your lowest fan setting but will now
+correspond to an actual 20% fan speed.
+If you had previously set `max_power` to anything below 1.0 (default),
+it is advisable to recalibrate `min_power` again, with the settings
+`min_power=0` and `max_power=1`.
```
kdb424, line 22: Should probably be `min_power: 0` and `max_power: 1` to keep in line with klipper formatting and not confuse users.

kdb424 (changes requested):

dewi-ny-je:
Isn't this something that can be implemented as macro? why a native way?

kubaracek:
> Isn't this something that can be implemented as macro? why a native way?

I wrote in description of the PR. I can't see any reason why in the context of Fan values below certain treshold should be ignored while `M106 S255` being scaled to max_power. What I'm configuring in klipper is a safe power range for my fan. The question is what use do you have for this not to be scaled. Why would you like to to be requesting a number bellow `off_bellow` and why have you decided to scale the uper limit 🤷 

Does anybody have a good example of where the current behaviour is useful?

docs/Config_Changes.md
kubaracek: Ah thanks. Will update

kubaracek:
M106:


```
[M106 – Set Fan Speed](https://www.simplify3d.com/support/articles/3d-printing-gcode-tutorial/#M106)
This command allows you to set the speed of your printer’s part cooling fan. This is an external cooling fan that is pointed towards the part that you are printing. Keep in mind that your printer may also have an extruder fan that helps cool the extruder drive mechanism, so make sure you are looking at the correct fan first. While most printers have an external cooling fan, there are a few exceptions, so check your machine first to make sure it has an external cooling fan that you can control.

Arguments:
The S value sets the speed of the cooling fan in a range between 0 (off) and 255 (full power).

Example usage:
M106 S255 ; set the fan to full speed

M106 S127 ; set the fan to roughly 50% power

M106 S0 ; turn off the fan completely
```

Sineos:
IMO this is a good change:
* From a developer's perspective, it is clear that a PWM range is between 0 and 255, regardless of any capping at the upper or lower limits, and it is also clear to the developer that any capping will shift the resulting duty cycle. 
* From a user's perspective, this is not very intuitive as the user sees it from the fan's perspective and expects 50% to be 50% fan speed and not something in between.

freakydude:
... and in addition, 
- it removes (or reduces) the slicer setting dependencies to the installed fans or their configuration

kubaracek:
Indeed. Also searching for M106 the internet says `Set Fan Speed` not `Set fan PWM percentage`. It shouldn't be on the client to know what my fan hardware is. If I have my fan configured well in klipper, 1% fan speed should mean that it spins on its lowest power so this should be abstracted away from the client.

kdb424 (approved):
Looks good to me. Workin as intended.

trofen:
What if to leave both "off_below" and "min_power"? This will ensure backward compatibility and both parameters are still different in meaning and may be useful.

surikatec:
Is this implemented in newest klipper version? (Trying that but can't find it.)

JamesH1978:
@surikatec no this is an open PR, it is not implemented or merged

freakydude:
What tasks ar open here?

KoenVanduffel:
It looks like this thing has stalled. I was looking for exactly this and manually copied the fan.py into my RatOs setup. Works like a charm.
A quick check gave me this result:  with min_power: 0.2 and max power: 1 my 4 pin 4028 reports following rpm: 1% fan it runs 1200 rpm, 25% = 5600 rpm, 50% = 11000 rpm, 75% = 16100 rpm and 100% 24900 rpm. Great that is plenty linear (and probably the linearity of the fan itself).

kubaracek:
Thanks for testing, it stalled on merging. There isn't anybody assigned to this with the rights to merge this unfortunately.
I don't think there is anything more to discuss or code for this.

stablestud:
> people merging this themself to their self-built klipper
And yes I'm unfortunately doing it aswell, shame that this simple feature is not in klipper...

@KoenVanduffel [comment|https://github.com/Klipper3d/klipper/pull/6307#issuecomment-1890602631] regarding developer and user perspective is very good and its confusing for most people. Marlin does that quite well with FAN_MIN_PWM and FAN_MAX_PWM

@KevinOConnor can you please merge this or #4720?
Looking at the amount of issues and opened PRs this feature is quite in demand

willpuckett:
Chiming in to support this PR. It seems this would align the way fans work with most users expected behhavior, and would simplify/reduce the need to explicitly configure slicer settings for low fan speeds.

PR #6309, opened 2023-08-01
Updating Wanhao Duplicator i3 Plus 2017 config to add probe related details 

travmo:


JamesH1978:
Thankyou for submitting a PR, please could you refer to point 3 in the "what to expect" section in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and add a signed-off-by line.

If this printer did not come with a probe as stock, please comment out the probe section with #'s and any other sections that you have added as related to it, so as not to confuse those with a stock version of this printer.

You also still have whitespace issues you need to fix so the checks can pass.

I will review properly once you have addressed these issues.

Thanks
James

PR #6312, opened 2023-08-07
lis2dw12: Add support for lis2dw12 accelerometer

bigtreetech:
lis2dw12 is an accelerometer from [STMicroelectronics](https://www.st.com/resource/en/datasheet/lis2dw12.pdf)
With better performance than the ADXL345 according to the datasheet.


bigtreetech:
This `Build test / build (pull_request) ` error is `Klipper image is too large`. 
Using `Optional features (to reduce code size)` locally to disable other features and enable `Support external sensor devices` can successfully build

KevinOConnor:
Thanks.  In general it seems fine to me.  I have a few comments:
1. The build tests need to pass - try adding `WANT_DISPLAYS=n` and/or `WANT_SENSORS=n` to `test/configs/ar100.config` to see if that fixes the issue.  (Or perhaps @eliasbakken has a suggestion here.)
2. The new `klippy/extras/lis2dw.py` and `src/sensor_lis2dw.c` files need a copyright statement at the top of the file.  If you don't want to add your copywrite, you can copy the copyright statements from `klippy/extras/adxl345.py` and `src/sensor_adxl345.c` respectively.
3. The docs/Config_Reference.md file should be updated with `[lis2dw]`.

Also, @dmbutyugin do you have any comments on this PR?

-Kevin

eliasbakken:
I think `WANT_DISPLAYS` is already deselected. Deselecting `WANT_SENSORS` would also remove adxl345, which some people might be using with ar100. Is it an option to increase the granularity a bit? Either that or remove this new accelerometer by using the "HAVE_LIMITED_CODE_SIZE" setting to make the ar100 pass.

bigtreetech:
> Thanks. In general it seems fine to me. I have a few comments:
> 
> 1. The build tests need to pass - try adding `WANT_DISPLAYS=n` and/or `WANT_SENSORS=n` to `test/configs/ar100.config` to see if that fixes the issue.  (Or perhaps @eliasbakken has a suggestion here.)
> 2. The new `klippy/extras/lis2dw.py` and `src/sensor_lis2dw.c` files need a copyright statement at the top of the file.  If you don't want to add your copywrite, you can copy the copyright statements from `klippy/extras/adxl345.py` and `src/sensor_adxl345.c` respectively.
> 3. The docs/Config_Reference.md file should be updated with `[lis2dw]`.
> 
> Also, @dmbutyugin do you have any comments on this PR?
> 
> -Kevin

1. According to the suggestion of @eliasbakken, the configuration of `WANT_LIS2DW` has been added. And now all build tests can pass.
2. The copyright has been added as required.
3. The instructions of how to configure LIS2DW has been added in docs/Config_Reference.md.

KevinOConnor:
Thanks.  It looks fine to me.  I'll give a couple of days to see if there are additional comments - otherwise I'll look to commit.

Cheers,
-Kevin

klippy/extras/lis2dw.py
```diff
@@ -0,0 +1,269 @@
...
+                % (dev_id, LIS2DW_DEV_ID))
+        # Setup chip in requested query rate
+        # ODR/2 +-16g
+        dev_data =  0x30
+        self.set_reg(REG_LIS2DW_CTRL_REG6_ADDR, dev_data)
+        # Continuous mode: If the FIFO is full
+        # the new sample overwrites the older sample.
+        dev_data =  0xC0
+        self.set_reg(REG_LIS2DW_FIFO_CTRL, dev_data)
+        # High-Performance / Low-Power mode 800/200 Hz
```
dmbutyugin, line 203: The comment and the supplied value (0x84) contradict the set `self.data_rate = 1600` data rate. If I read the datasheet correctly, the correct value would be 0x94 (and the comment would need to be updated to reflect 1600 Hz data rate).

klippy/extras/lis2dw.py
```diff
@@ -0,0 +1,269 @@
...
+        dev_id = self.read_reg(REG_LIS2DW_WHO_AM_I_ADDR)
+        logging.info("lis2dw_dev_id: %x", dev_id)
+        if dev_id != LIS2DW_DEV_ID:
+            raise self.printer.command_error(
+                "Invalid lis2dw id (got %x vs %x).\n"
+                "This is generally indicative of connection problems\n"
+                "(e.g. faulty wiring) or a faulty lis2dw chip."
+                % (dev_id, LIS2DW_DEV_ID))
+        # Setup chip in requested query rate
+        # ODR/2 +-16g
```
dmbutyugin, line 196: You could also mention that you enable low-noise mode here.

dmbutyugin (commented):
Thanks. I had just a few comments. I didn't test the accelerometer though, so I cannot comment whether and how it works either in the current state or with the suggested mode change. Besides that, I don't have any major feedback for the support of this accelerometer.

klippy/extras/lis2dw.py
bigtreetech: Yeah. Thanks a lot for the review.
This is indeed a problem and has been fixed now.

klippy/extras/lis2dw.py
bigtreetech: Hi, 0x30 is diasble low-noise mode. Do I need to set it to 0x34?
image

dmbutyugin: Ah, yes, you are right, 0x34 enables the low-noise configuration. Well, I did not test this accelerometer, but just looking at the datasheet, I see no drawbacks of enabling it. It just makes the sensor consume ~30% more current, but it is still little, and if the powerchain is adequate, it should cause no problems.

bigtreetech: enabled now

KevinOConnor:
Thanks.

-Kevin

PR #6317, opened 2023-08-12
Default to no additional temperature smoothing

KevinOConnor:
The default for heater `smooth_time` used to be 2 seconds; it was reduced to 1 second on 20210903.  This PR would further change that default to not perform any extra temperature smoothing.

Klipper already has a relatively slow response time of 300ms for heater updates.  We've seen in the past that additional "smoothing" time can result in worse PID stability (which was why the default was originally reduced from 2 seconds).  I suspect that no additional smoothing may be a better overall default for most printers.  That is, I suspect most printers today don't need extra smoothing and could benefit from a more responsive pid.

In particular, many printers today have stable temperature circuits.  Also, regardless of the `smooth_time` setting, Klipper already averages 8 measurements on every reported sample.  So, in effect, there is already some "smoothing" even with no extra `smooth_time` set.

Comments?

@rext3d , @dans98 - fyi.

-Kevin

P.S.  This is a followup to PR #4646 .

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,11 @@ All dates in this document are approximate.
 
 ## Changes
 
+20230820: The heater `smooth_time` now defaults to not perform any
+additional temperature smoothing. To return to the previous behavior
+of one second smoothing, set `smooth_time: 1.0` in the heater config
+section.
+
 20230810: The flash-sdcard.sh script now supports both variants of the
 Bigtreetech SKR-3, STM32H743 and STM32H723. For this, the original tag
 of btt-skr-3 now has changed to be either btt-skr-3-h743 or btt-skr-3-h723.
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -798,10 +798,10 @@ sensor_pin:
 #   The resistance (in ohms) of the pullup attached to the thermistor.
 #   This parameter is only valid when the sensor is a thermistor. The
 #   default is 4700 ohms.
-#smooth_time: 1.0
+#smooth_time:
 #   A time value (in seconds) over which temperature measurements will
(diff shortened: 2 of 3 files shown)


ReXT3D:
I will have a look at the impact on my two printers in the next few days:

1. Cr-10S Pro with Duet3 Mini 5+, stock bed (~400W 24V DC heater), Dragon with 50W heater & PT1000 connected directly to the Duet3 Mini 5+.
2. Voron 2.4 with Duet3 Mini 5+, 6mm MIC bed with RTV attached 400W 120V AC Keenovo heater, Mosquito with 50W heater & PT1000 connected via Duet PT100 Daughter Board (with reference resistors modified for PT1000 RTD)

Based on my previous experiments I expect that the removal of extra temperature smoothing will work well in some cases but will result in very high levels of PWM "noise" in other cases.  I will run some comparisons and post the results.

KevinOConnor:
Thanks for testing and reporting results.  It seems we may need to get more tests on various printers before proceeding with this merge.  (In general, I think it should be okay for the PWM value to have greater variability as long as the resulting temperature stays relatively steady, but more tests is probably a good thing.)

-Kevin

freakydude:
Hello everyone,

Just a quick feedback from me.

I did a quick test on my Sidewinder X2 with Manta M8P board. 
I have Volcano Heatblock/Nozzle with 64W heater.

I switched to this patch. Recalibrated the PID with 240°C target. 

After calibration, heating to 210°C results in temperature fluctuation around 207-209°C.
Heating to the calibrated 240°C gives 241-243°C.

Finally, I switched back to the master branch using the same calibrated PID values, the temperature stays within +-0.2°C. Either at 210°C or 240°C

Best Regards
freakyDude

PR #6320, opened 2023-08-21
Limit maximum temperature in MAX31865.calc_adc() to melting point of platinum

DavidvtWout:
Limit the maximum temperature in MAX31865.calc_adc() to the melting point of platinum. Above this temperature the Callendar-Van Dusen formula does not make sense. The default value for max_temp is 9999999.9 and at this temperature the result of this formula is negative. This sets max_sample_value to 0 which causes the mcu to shutdown.

Discourse topic: https://klipper.discourse.group/t/max31865-min-value-and-max-value-not-set/10029

### DIFF
--- a/klippy/extras/spi_temperature.py
+++ b/klippy/extras/spi_temperature.py
@@ -320,6 +320,7 @@ class MAX31865(SensorBase):
     def calc_adc(self, temp):
         # Calculate relative resistance via Callendar-Van Dusen formula:
         #  resistance = rtd_nominal_r * (1 + CVD_A * temp + CVD_B * temp**2)
+        temp = min(temp, 1768.3)  # Melting point of platinum
         R_div_nominal = 1. + CVD_A * temp + CVD_B * temp * temp
         adc = int(R_div_nominal / self.adc_to_resist_div_nominal + 0.5)
         adc = max(0, min(MAX31865_ADC_MAX, adc))
--- a/klippy/extras/spi_temperature.py
+++ b/klippy/extras/spi_temperature.py
@@ -323,7 +323,7 @@ class MAX31865(SensorBase):
         temp = min(temp, 1768.3)  # Melting point of platinum
         R_div_nominal = 1. + CVD_A * temp + CVD_B * temp * temp
         adc = int(R_div_nominal / self.adc_to_resist_div_nominal + 0.5)
-        adc = max(0, min(MAX31865_ADC_MAX, adc))
+        adc = max(0, min(MAX31865_ADC_MAX - 1, adc))
         adc = adc << 1 # Add fault bit
         return adc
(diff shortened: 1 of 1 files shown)


JamesH1978:

Thanks
James

DavidvtWout:
Thanks @JamesH1978! I have fixed the commit messages.

KevinOConnor:
Thanks.  Seems fine to me.  I'll give a few days to see if anyone else has comments.

-Kevin

Sineos:
Would it make sense to update the documentation along with this change? E.g.:

```
[temperature_sensor my_sensor]
#sensor_type:
#sensor_pin:
#min_temp:
#max_temp:
#   See the "extruder" section for the definition of the above
#   parameters. Contrary to the "extruder" section, 
#   "min_temp" / "max_temp" are not mandatory.
#gcode_id:
#   See the "heater_generic" section for the definition of this
#   parameter.
```

Sineos:
> The `#` already indicates that the value is optional, right? But I agree that it may be helpful to emphasise this.

My understanding is that values without # have no default value and as such are mandatory (although it seems it is not followed through consequently). In this sense the documentation block should look like:

```
[temperature_sensor my_sensor]
sensor_type:                                           #<-- Needs to be mandatory
sensor_pin:                                            #<-- Needs to be mandatory
#min_temp: -273.15                                     #<-- Default value
#max_temp: 1768.3                                      #<-- Default value
```

Since this looks kind of weird, I'd propose:
```
[temperature_sensor my_sensor]
sensor_type:
sensor_pin:
#min_temp:
#max_temp:
#   See the "extruder" section for the definition of the above
#   parameters. Contrary to the "extruder" section, 
#   "min_temp" / "max_temp" are optional.
#gcode_id:
#   See the "heater_generic" section for the definition of this
#   parameter.
```

DavidvtWout:
I added the proposed error messages to this PR. If you don't like it I can drop the commit.

KevinOConnor:
Thanks.  The spi_temperature.py changes look fine, but I'm not sure about the changes to thermocouple.c - as some of the sensors report smaller values on higher temperatures so the error may be misleading.  In general, I think it is preferable for the python code to do most of the error reporting as the MCU code often doesn't have enough information to provide useful information.

-Kevin

DavidvtWout:
> Thanks. The spi_temperature.py changes look fine, but I'm not sure about the changes to thermocouple.c - as some of the sensors report smaller values on higher temperatures so the error may be misleading. In general, I think it is preferable for the python code to do most of the error reporting as the MCU code often doesn't have enough information to provide useful information.
> 
> -Kevin

I didn't know some sensors have the values reversed. I dropped the error message commit.

KevinOConnor:
Thanks.

-Kevin

PR #6321, opened 2023-08-23
make inverted belt path work with hybrid_corexy idex

HelgeKeck:
This PR tries to fix an issue with an inverted belt path for hybrid_corexy Idex setups.

Thats the v-core idex belt path. It seems that it isnt possible atm to properly configure this situation with the current hybrid_corexy and dualcarriage implementation

image

If i configure the Toolhead stepper direction so that they move in the right direction, then they move diagonaly when moving just Y. If i invert the Y stepper direction then toolheads do move correctly in y, but then Y is inverted. Also tried to use mtor B for the X-carraige, same result.

Only way to make it somehow work is to put the dual carriage on the left side and the X carriage to the right side, but then i cant use Set Gcode offset anymore for the copy and mirror mode, only set kinematics helps then which results in a not working bed mesh leveling for the main toolhead

it would work if my toolheads would point into this direction
image

i dont know a lot about kinematic coding in klipper, but this change seems to work very well from what i can say from my first tests.

would be nice if someone could have a look and tell me if there are any sideeffects i can face with this kind of change.

thanks

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2084,6 +2084,8 @@ configuration.
 axis:
 #   The axis this extra carriage is on (either x or y). This parameter
 #   must be provided.
+invert_axis:
+#   Set to True in case you have a inverted belt path.
 #safe_distance:
 #   The minimum distance (in mm) to enforce between the dual and the primary
 #   carriages. If a G-Code command is executed that will bring the carriages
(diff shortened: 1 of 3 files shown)


HelgeKeck:
just to be clear, i just sued set kinematic position to make some simple moving tests, no homing or even printing has been done yet. i assume there is mch more to that than just chaning these two signs

HelgeKeck:
ok, homing doesnt work anymore

HelgeKeck:
not sure if its helpful, but homing X alone works, but if y is already homed, then the x-carriage doesnt stop on the second homing attempt after the homing retraction move

PR #6322, opened 2023-08-25
Update generic-bigtreetech-skr-3.cfg

danieldefreitasleite:
Correcting the generic config to avoid problems when using regular T8 leadscrews on Z axis

### DIFF
--- a/config/generic-bigtreetech-skr-3.cfg
+++ b/config/generic-bigtreetech-skr-3.cfg
@@ -33,7 +33,8 @@ step_pin: PE2
 dir_pin: PE3
 enable_pin: !PE0
 microsteps: 16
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thankyou for submitting a PR, but this is a board configuration not a printer configuration, it is up to the end user to adapt this for their printer, which could be a wide array of values not just 40 or 8. That being said, a simple change of value to 8 with no comment would be fine, as not many printers would have a Z value of 40 anyway.

Also to submit a valid PR please refer to point 3 in the "what to expect" section in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and add a signed-off-by line.

Thanks
James

PR #6323, opened 2023-08-29
uc1701: x_offset and columns values added for uc1701 displays

VragVideo:
Changing x_offset and columns in the printer.cfg allows to fix artifacts like shown on image. x_offset parameter shifts picture forward and columns parameter set to 132 clear point at end of line.
image

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4030,6 +4030,11 @@ a0_pin:
 #contrast:
 #   The contrast to set. The value may range from 0 to 63 and the
 #   default is 40.
+#x_offset: 0
+#   Set the horizontal offset value on UC1701 displays. The value may
+#   range from 0 to 4 and the default is 0.
+#columns: 128
+#   Set the number of columns on UC1701 displays. The default is 128.
 ...
 ```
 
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  Can you elaborate on what hardware needs this customization?  It seems unusual that there are some uc1701 chips that are so different from other uc1701 chips.  Is it perhaps a different display chip entirely?

-Kevin

VragVideo:
> Can you elaborate on what hardware needs this customization? It seems unusual that there are some uc1701 chips that are so different from other uc1701 chips. Is it perhaps a different display chip entirely?
I use klipper on Alfawise Ex8 printer. This printer is version of the Anet A8. It has display with label on plate "WYH L12864 COG".
As I know this is ST7565 based LCD with a single 10-pin connector. Sorry for delay I was out of net.

KevinOConnor:
Okay, thanks.  If the chip is actually an ST7565 then I don't think we should add new options to the uc1701 config.  Consider adding a new ST7565 definition (both to the code and the documentation).  For example, you can check out how the SH1106 is implemented as a customization on the base uc1701 implementation.

-Kevin

PR #6325, opened 2023-08-29
Update Manual_Level.md

MightyMirko:
I have hard time understanding it from first read. This would clear things up and relax my mind a bit


docs/Manual_Level.md
```diff
@@ -179,6 +179,8 @@ Recv: // rear right screw : y=155.0, y=190.0, z=2.71500 : adjust CCW 00:50
 Recv: // read left screw : x=-5.0, y=190.0, z=2.47250 : adjust CW 00:02
 Recv: ok
 ```
+Counter clockwise:   tighten the screw so that the bed goes down. 
+clockwise:           loosen the screw so that the bed goes up.
```
kdb424, line 183:
```suggestion
Clockwise:           loosen the screw so that the bed goes up.
```

PR #6327, opened 2023-09-01
virtual_sdcard: add support subdirectories in menu and M'commands

pOmelchenko:
Added a configuration for working with subdirectories on a virtual sd card


JamesH1978:

Thanks
James

PR #6330, opened 2023-09-03
config: Improve Ultimaker default config

xhalo32:
Hi! I have some changes to `config/generic-ultimaker-ultimainboard-v2.cfg`, to at least have better support for Ultimaker 2+:
- Added display and encoder wheel and LED support
- Fixed extruder gear ratio
- Added heatbreak fan

I have tested these on an Ultimaker 2+ and the only warnings that I get are related to missing configuration:
-  virtual_sdcard is not defined in config.
-  pause_resume is not defined in config.
-  gcode_macro pause is not defined in config.
-  gcode_macro resume is not defined in config.
-  gcode_macro cancel_print is not defined in config. 

If these features should be moved to their own file specifically for Ultimaker 2+, I can move them.

### DIFF
--- a/config/generic-ultimaker-ultimainboard-v2.cfg
+++ b/config/generic-ultimaker-ultimainboard-v2.cfg
@@ -31,7 +31,7 @@ step_pin: PC2
 dir_pin: !PC1
 enable_pin: !PC3
 microsteps: 16
-rotation_distance: 40
+rotation_distance: 16
 endstop_pin: ^!PA7
 position_endstop: 215
 position_max: 215
@@ -39,10 +39,11 @@ homing_speed: 20.0
 
 [extruder]
 step_pin: PL7
-dir_pin: PL6
+dir_pin: !PL6
 enable_pin: !PC0
 microsteps: 16
 rotation_distance: 33.500
+gear_ratio: 36:11
 nozzle_diameter: 0.400
 filament_diameter: 2.850
 heater_pin: PE4
@@ -58,10 +59,11 @@ max_temp: 275
 # Dual extruder support.
 #[extruder1]
 #step_pin: PL0
-#dir_pin: PL2
+#dir_pin: !PL2
 #enable_pin: !PL1
 #microsteps: 16
 #rotation_distance: 33.500
+#gear_ratio: 36:11
 #nozzle_diameter: 0.400
 #filament_diameter: 2.850
 #heater_pin: PE5
@@ -73,6 +75,14 @@ max_temp: 275
 #pid_Kd: 114
 #min_temp: 0
 #max_temp: 275
+
+[heater_fan nozzle_fan]
+pin: PJ6
+max_power: 1
+shutdown_speed: 0
+heater: extruder
+heater_temp: 50.0
+fan_speed: 0.6
 
 [heater_bed]
 heater_pin: PG5
@@ -125,3 +135,17 @@ scale: 2.000
 cycle_time: .000030
 hardware_pwm: True
 static_value: 1.250
+
+[display]
+lcd_type: ssd1306
+reset_pin: PE3
+menu_timeout: 60
+click_pin: ^!PD2
+encoder_pins: ^PG1, ^PG0
+encoder_steps_per_detent: 2
+
+[pca9632 wheel_leds]
+i2c_address: 96
+initial_RED: 1.0
+initial_GREEN: 1.0
+initial_BLUE: 1.0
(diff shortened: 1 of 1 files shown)


JamesH1978:

Additionally  the warnings you see are specifically for Mainsail requirements, and not for klipper itself, therefore are not needed for inclusion in a generic config.

Thanks
James

xhalo32:
Hi @JamesH1978 

I have signed off my changes to certify they have been done according to the Developer Certificate of Origin.

Is there anything else you would like me to change?

config/generic-ultimaker-ultimainboard-v2.cfg
```diff
@@ -31,18 +31,19 @@ step_pin: PC2
 dir_pin: !PC1
 enable_pin: !PC3
 microsteps: 16
-rotation_distance: 40
+rotation_distance: 16
```
JamesH1978, line 34: Can i ask why you have changed this to 16?

Thanks
James

xhalo32: Setting it to the "correct" value of 8mm according to [this guide](https://www.klipper3d.org/Rotation_Distance.html#axes-with-a-lead-screw) leads to the Z axis moving at double speed on Ultimaker 2+. Therefore I set it to 16mm.

There's clearly something I don't understand about how to calculate that value correctly or how klipper handles these values. With `rotation_distance: 16` moving by 8mm causes one full rotation of the leadscrew.

JamesH1978: ok i have done some research on this, originally this file was commited with a step_distance of 0.005(these values being the precursor to rotation_distance) this was then updated later to be a rotation_distance of 8 and a microsteps of 16, which i think was wrong, as rotation_distance = <full_steps_per_rotation> * <microsteps> * <step_distance> so 200 * 16 * 0.005 actually equals 16 which works for you. but the tangle i found was that the ultimaker board actually might use 8 microsteps on Z, so in that case 200 * 8 * 0.005 = 8, so i would expect rotation_distance: 8 and microsteps: 8. I do not have this board so i cannot confirm what it comes from factory jumpered for on Z for microsteps. How this all became rotation_distance: 40 i am not sure, i cannot find the commit that did this in the history.

So... if you can confirm the microsteps set on the board, assuming its stock, we can get the right values in place :)

Thanks
James

Sineos: https://github.com/MarlinFirmware/Configurations/blob/abb24105b39f593a093baa5477310d5043682b1c/config/examples/Ultimaker/Ultimaker%202%2B/Configuration.h#L1239
https://github.com/MarlinFirmware/Configurations/blob/abb24105b39f593a093baa5477310d5043682b1c/config/examples/Ultimaker/Ultimaker%202%2B/Configuration_adv.h#L1393

Marlin uses 200/16

JamesH1978:
thanks @Sineos then that confirms it 16 microsteps set in configuration_adv.h, so (200*16)/200 does indeed equal RD of 16. I can continue to review now.

Thanks
James

config/generic-ultimaker-ultimainboard-v2.cfg
```diff
@@ -74,6 +76,14 @@ max_temp: 275
 #min_temp: 0
 #max_temp: 275
 
+[heater_fan nozzle_fan]
```
JamesH1978, line 79: Please consider this naming, as [fan] is the parts fan, and the naming you have chosen for the heater_fan can be confusing, we have settled on `[heater_fan heatbreak_cooling_fan]` as the common naming for a fan that cools the hot end. Unless this is some other type of fan that isnt the main parts fan or the heatbreak cooling fan?

Thanks
James

xhalo32: Updated.

JamesH1978 (approved):
Looks fine to me now,

Thanks
James

KevinOConnor:
Thanks.  It seems odd that a display is defined for a "generic" config file (point 6a at https://www.klipper3d.org/Example_Configs.html ).  Is the goal here to describe a particular printer instead of the control board?

-Kevin

KevinOConnor:
Maybe @averyct has comments on this ultimaker config PR?

-Kevin

KevinOConnor:
What is the status of this PR?

I think we can either add this file as something like `printer-ultimaker-2plus.cfg` or comment out the new display section in the update to `config/generic-ultimaker-ultimainboard-v2.cfg`.

-Kevin

JamesH1978:
Personally i think its more suited as its own printer config.

Thanks
James

PR #6332, opened 2023-09-04
config: Kingroon KP3S stepper will go to the right direction

chesti84:
With the default configuration, the home position was ser in the wrong direction. 

Please check the [Kingroon KP3S configuration file](https://github.com/nehilo/Klipper-KingRoon-Printers/blob/main/KP3S_Configuration/stepper.cfg) and [Kingroon KP3S howto page](https://kingroon.com/blogs/3d-print-101/how-to-prepare-klipper-firmware-for-kingroon-kp3s-using-fluiddpi) for more details

### DIFF
--- a/config/printer-kingroon-kp3s-2020.cfg
+++ b/config/printer-kingroon-kp3s-2020.cfg
@@ -18,7 +18,7 @@
 
 [stepper_x]
 step_pin: PE3
-dir_pin: !PE2
+dir_pin: PE2
 enable_pin: !PE4
 microsteps: 32
 rotation_distance: 40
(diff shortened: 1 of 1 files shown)


JamesH1978:

That being said, on my KP3S stock version 1 with the old board, the directions were correct, as this is my config. These are example configs, you are meant to be checking sanity of them, even if stock with https://github.com/Klipper3d/klipper/blob/master/docs/Config_checks.md

Thanks
James

rcarmo:
I had to fix this in my config, and can confirm this is needed.

JamesH1978:
As 2 people habe now confirmed it, my stepper wiring must be different on mine. Who knows, but i will approve the small change.

Thanks
James

PR #6334, opened 2023-09-07
Config: Add HC32F460 variant of Ender 2 Pro

JamesH1978:
PR to add the newer version of the Ender 2 Pro with the CR-FDM-v2.5.54.170 motherboard with the HC3232F460 chip.

Confirmed as working with discord member NyftHeart and using configs from Steve Gotthardt, with his permission to use and submit.

Thanks

James


KevinOConnor:
Thanks.

-Kevin

PR #6335, opened 2023-09-08
Add MKS Monster 8

Bomberski:
Just a generic config for the mks Monster 8


JamesH1978:

You have also failed the build checks, looks like you need to add a new line at the end of the file.

You also need to add to the printere.test for regression tests, see point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Thanks
James

PR #6337, opened 2023-09-08
build: Add support for building mcu code on MacOS

thatcomputerguy0101:
This change is mostly for development purposes so that experimental MCU code can be built and deployed without one of the fully supported Linux environments. I restored some of the deleted files from the bossac dependency and made some tweaks to the build scripts to adapt to the different environment. With the proper compiler installed corresponding to the MCU archetecture, this worked for me on both x86_64 and ARM64 Macs.


thatcomputerguy0101:
It may be possible to extend this into full MacOS support, especially with the majority of Klipper being in Python. Would this be useful?

KevinOConnor:
Thanks.

One question I have is - is it possible to run a docker container on macos that can successfully compile (and/or run) Klipper?

In general, I'm leery of adding if/else type of conditions to the build rules for individual build targets.  In past experience I found that doing so leads to excessive build rules that become difficult to maintain.  It seems far simpler to emulate one base build system (eg, via containers) than to maintain custom build rules for so many different possible build environments.

Separately, some changes (like adding bossac build files) seem fine, but would need to be squash-and-merged into a single commit.  That is, we don't want to mix changes to flash_usb.py, with Makefile changes, with bossac changes in a single commit - as that makes it difficult to use important tools like `bit bisect` and `git revert`.

-Kevin

thatcomputerguy0101:
As far as I can tell, it seems to be impossible to forward USB or serial ports in the default Docker engine in MacOS Docker Deskop without using some utility like `socat` to pipe the serial port over a network interface. I believe that is due to missing feature support in the underlying virtualization framework. There is apparently a Virtualbox-based Docker host that can share USB ports, which I haven't yet explored, but would make for a higher barrier of entry over the well-published Docker Desktop.

With one exception, the changes that I made so far are either user-invoked (the install script) or detect feature differences (the serial port detection). The one exception being the `cpp` compiler condition in the Makefile, which I now realize I could redo to detect the output of `cpp --version` since that reports using clang if that is the underlying implementation. Would strictly using feature detection instead of making environment assumptions if in "MacOS" make you less leery?

With respect to the commit organization, are you saying that the first commit should be split apart, or that all of these commits should be squashed into one? Not mixing changes to different files and squash-and-merge into a single commit seems to be conflicting information to me.

KevinOConnor:
Thanks.  And sorry for the delay in responding.

What I meant is, splitting the commits into separate atomic parts (not merging together, and not having many commits with merged functionality).

So, for example, if we need to place a space between `-I` and `$(OUT)` that's fine and if we need a `libclean` that's fine, but those types of changes ideally would not be in the same commit as a bossac change.  Similarly for flash_usb changes.

Also, I'm not sure it is a good idea to try to guess the CPP version in the makefiles - that type of logic could become a long-term maintenance burden.  For the unusual build platforms I think it is fine for users to know that they need to run `make CPP="clang -E"`.

-Kevin

PR #6340, opened 2023-09-16
usb_cdc: add support for bootloader requests over USB CDC

Laikulo:
This adds support for the "Request Serial Bootloader" psuedo-command to the virtual serial device.


Laikulo:
Retracting in favor of existing 1200/DTR functionality

PR #6341, opened 2023-09-16
fan: Support multiple printer fans

viesturz:
Allows to setup multiple printer fans.

Multi extruder setups can have separate part cooling fan for each extruder. This integrates support for that in core klipper, removing the need for every multi extruder config to override their macros.

But probably more importantly defines an endorsed way how multi fan setups should be controlled by the slicers.

### DIFF
--- a/config/sample-idex.cfg
+++ b/config/sample-idex.cfg
@@ -34,6 +34,10 @@ pid_Kd: 114
 min_temp: 0
 max_temp: 250
 
+# The definition for the part cooling fan for primary extruder
+[fan primary]
+pin: PH6
+
 # Helper script to park the carriage (called from T0 and T1 macros)
 [gcode_macro PARK_extruder]
 gcode:
@@ -47,6 +51,7 @@ gcode:
 gcode:
     PARK_{printer.toolhead.extruder}
     ACTIVATE_EXTRUDER EXTRUDER=extruder
+    ACTIVATE_FAN FAN=primary
     SET_DUAL_CARRIAGE CARRIAGE=0
     SET_GCODE_OFFSET Y=0
 
@@ -83,6 +88,10 @@ pid_Kd: 114
 min_temp: 0
 max_temp: 250
 
+# The definition for the part cooling fan for secondary extruder
+[fan secondary]
+pin: PH7
+
 [gcode_macro PARK_extruder1]
 gcode:
     SAVE_GCODE_STATE NAME=park1
@@ -94,6 +103,7 @@ gcode:
 gcode:
     PARK_{printer.toolhead.extruder}
     ACTIVATE_EXTRUDER EXTRUDER=extruder1
+    ACTIVATE_FAN FAN=secondary
     SET_DUAL_CARRIAGE CARRIAGE=1
     SET_GCODE_OFFSET Y=15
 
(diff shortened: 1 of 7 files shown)


viesturz:
This is another take on https://github.com/Klipper3d/klipper/pull/6282. 
Also see https://klipper.discourse.group/t/multiple-tools-gcode-standardization-possible/10127 for the broader context.

KevinOConnor:
Thanks.  In general it seems fine to me.  I'd prefer to see some comments from users of this functionality as well though (I don't own hardware that would utilize this functionality so it is hard for me to judge).

Also, `multi_fan` seems a little ambiguous to me.  Perhaps `fan_multiple_toolhead` might be more explicit.  Similarly, command names should ideally contain the module name in the command name (eg, `FAN_MULTIPLE_TOOLHEAD_ACTIVATE`).

Finally, what's the pros/cons of adding a new module as opposed to adding support for ACTIVATE_FAN to existing "fan" module?

-Kevin

viesturz:
Hi, thanks for the feedback.
I actually prefer extending fan. The multi_fan choice was to keep the implementation separate. Will update the PR to fan. I'm not seeing any other cons for using just fan.

One other part that I would like to add is a
Possibility to activate multiple fans - that would allow controlling side mounted additional fans for example. But not sure if the extra complexity is worth it.

viesturz:
Changed to use the [fan].


I am trying with the idea of adding  ACTIVATE_FAN FAN=fan1,fan2 bit personally do not have a use case for it.

myxxx323:
Good afternoon, everything works fine, but only in 2 extruder mode, in copy or mirror mode, the second fan does not work at all

myxxx323:
Your solution is the best that can be found on the Internet, I hope you will be able to make the second fan work

viesturz:
Back from vacation, updated to handle IDEX mirror mode and the like.

myxxx323:
> Back from vacation, updated to handle IDEX mirror mode and the like.

Thanks, it works :) Can you tell me where the tool change speed setting is, the heads are changed very slowly

viesturz:
Hmm, not sure what you are referring to. The index example config?
Something else?

In general the speed is controlled by Fxxx parameter in G0/G1 commands.
Falling back previously used values when absent.

Perhaps discord is a better place troubleshoot your particular config?

myxxx323:
Thank you, I figured it out, I just added what accelerations were needed to the parking macros

viesturz:
Starting a discourse thread: https://klipper.discourse.group/t/review-request-support-multiple-part-cooling-fans/11411

myxxx323:
> Starting a discourse thread: https://klipper.discourse.group/t/review-request-support-multiple-part-cooling-fans/11411

Good afternoon, tell me, are there any plans to add these fixes to the Klipper updates? After updating the Klipper, everything is deleted. When changing files, the Klipper becomes dirty.

eldeesmith:
interesting how IDEX support was added without the ability to add a second parts cooling fan. are IDEX people just using multipin and having the parked head blow the cooling fan in tandem with the active head? definitely need multiple fan support in klipper

brotherdust:
I need this for my IDEX printer as well. Please merge.

PR #6343, opened 2023-09-18
min_time setting in [fan] section

trofen:
This PR is aimed at solving the problem with Dynamic fan speed on overhangs feature in Prusa Slicer. 
When this function is enabled on complex models, there are situations when the fan speed changes several times per second (sometimes these changes can occur up to hundreds of times per second if you print at really high accelerations). In such cases, klipper begins to accumulate a queue of fan speed changes, changing the speed much later than it is required and ignoring (postponing in a long queue) new commands. I saw the problem in the FAN_MIN_TIME parameter in fan.py . This parameter limits the number of fan speed changes to 10 times per second. 
I am not sure that this parameter can be removed, but it is quite possible to try to allow the user to set this parameter in the configuration file. 
I will also attach an [example](https://github.com/Klipper3d/klipper/files/12648484/test.zip) of an gcode, which causes fan spinning after the end of printing and ignore control comands for several minutes or until firmware restart.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2661,6 +2661,12 @@ pin:
 #   Time (in seconds) to run the fan at full speed when either first
 #   enabling or increasing it by more than 50% (helps get the fan
 #   spinning). The default is 0.100 seconds.
+#min_time: 0.1
+#   The minimum time in seconds during which the fan will execute the
+#   previous set speed value and not start executing the next one.
+#   The default is 0.100 seconds.
+#   The value can be changed downwards if the gcode contains a lot of
+#   fast fan speed changes.
 #off_below: 0.0
 #   The minimum input speed which will power the fan (expressed as a
 #   value from 0.0 to 1.0). When a speed lower than off_below is
--- a/klippy/extras/fan.py
+++ b/klippy/extras/fan.py
@@ -5,8 +5,6 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 from . import pulse_counter
 
-FAN_MIN_TIME = 0.100
-
 class Fan:
     def __init__(self, config, default_shutdown_speed=0.):
         self.printer = config.get_printer()
@@ -18,6 +16,7 @@ class Fan:
                                                minval=0.)
         self.off_below = config.getfloat('off_below', default=0.,
                                          minval=0., maxval=1.)
+        self.min_time = config.getfloat('min_time', 0.1, above=0.)
         cycle_time = config.getfloat('cycle_time', 0.010, above=0.)
         hardware_pwm = config.getboolean('hardware_pwm', False)
         shutdown_speed = config.getfloat(
@@ -51,7 +50,7 @@ class Fan:
         value = max(0., min(self.max_power, value * self.max_power))
         if value == self.last_fan_value:
             return
-        print_time = max(self.last_fan_time + FAN_MIN_TIME, print_time)
+        print_time = max(self.last_fan_time + self.min_time, print_time)
         if self.enable_pin:
             if value > 0 and self.last_fan_value == 0:
                 self.enable_pin.set_digital(print_time, 1)


Sineos:
IMHO, it is not the firmware's responsibility to work around short-comings or bad design decisions of a slicer.
Apart from stressing Klipper, it will stress the hardware and actually is not even workable, since no fan can switch from 0 to 100% multiple times a second.

trofen:
> IMHO, it is not the firmware's responsibility to work around short-comings or bad design decisions of a slicer. Apart from stressing Klipper, it will stress the hardware and actually is not even workable, since no fan can switch from 0 to 100% multiple times a second.

Here's my point of view:

1. This PR does not affect the experience of the average user in any way. As it was 0.1s, so now this default value is 0.1s. This does not break mcu/config compatibility and does not make the codebase more confusing. Only those who face this problem will want to change this setting.
2. Those who face this problem cannot solve it in an elegant way. Only by disabling a feature in the slicer (honestly, a good feature) or making changes to the source files of the user's klipper. You can't fix it with any tricks with macros or configuration files (I mean in an elegant way).
3. This will load klipper, but if you choose between increased CPU, microcontroller or bus resources and completely incorrect execution of gcode, I chose the first one. Moreover, the use of these resources increases only if the user has configured his klipper in this way.
4. This is workable, and you can test it with my gcode example. No one wants to change 0-100-0-100 several times per second, but if slicer's gcode smoothly increases the fan speed from 20 to 50 in 1 second with 50 control points, why not? The current klipper will instead stretch the execution of these commands by 5 seconds!
5. Allow the user to change anything in the configuration if he wants to!

Sineos:
You are not going to tell me that gcode like
```
M106 S25
M106 S204
G1 X117.434 Y135.618 E.00739
M106 S25
G1 F6359.991
M106 S204
G1 X117.343 Y135.636 E.0027
M106 S25
M106 S204
G1 F6720.764
```
serves any practical purpose, do you?

trofen:
1. Although this gcode has no practical application, I don't see any contradiction in this gcode for firmware. Moreover, if the firmware received this gcode, I see no reason why the firmware should slow down the fan for 0.1 seconds and only then speed it up. 
2. I do not see any similarity between the attached issue with the Orca Slicer and my PR.

Sineos:
I recognize your opinion, but my opinion is just as clear:
* Even if Klipper would process these instructions, they do not have any practical meaning, since no fan on this planet is able to follow these change due to electrical and mechanical constraints
* Klipper spends processing cycles on it for no reason
* The attempt to make the firmware seemingly adhere to these impossible to fulfill instructions is the wrong approach
* A slicer that is creating such meaningless instructions and thus puts strain on the hardware should be fixed

KevinOConnor:
The `FAN_MIN_TIME` value is not a setting that users can reasonably configure.  It's not an attribute of the fan hardware, it's an upper bound of the worst case transmission latency from host to MCU that we feel the Klipper system can reasonably support.  If a user were to notably decrease this value then it would notably increase the risk of cryptic "timer too close" errors.

Support for many rapid fan speed changes would require adding support for queuing of those fan speed changes in the micro-controller (along with the corresponding memory management for that queue).  That is a notable amount of work.  There's no real interesting in doing that work (nor interest in maintaining the resulting code) because, as mentioned earlier in this thread, it's completely pointless to make rapid changes to fan speed.  The best solution is to disable these broken options in the slicer.

Separately, commits need to have a signed-off-by line in order for them to be merged - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

trofen:
>     * Even if Klipper would process these instructions, they do not have any practical meaning, since no fan on this planet is able to follow these change due to electrical and mechanical constraints
> 
>     * Klipper spends processing cycles on it for no reason
> 
>     * The attempt to make the firmware seemingly adhere to these impossible to fulfill instructions is the wrong approach

IMHO, the firmware has the right not to execute meaningless or physically impossible slicer commands, but it definitely does not have the right to execute them absolutely asynchronously at a near-random moment. This makes no sense with the whole essence of the sequential execution of gcode

>     * A slicer that is creating such meaningless instructions and thus puts strain on the hardware should be fixed

Let me explain. 
```
G1 X201.39 Y220.409 E.03883
M106 S226.95
M106 S119.85
G1 F3819.37
M106 S229.5
G1 X200.309 Y219.855 E.04172
M106 S221.85
G1 X199.503 Y219.414 E.03155
```
This real gcode is garbage. I see no reason to change the firmware just to handle this.
However, I don't see anything wrong with the code below.
```
M106 S206.55
G1 X202.418 Y220.893 E.00232
G1 F4054.361
M106 S209.1
G1 X201.39 Y220.409 E.03883
G1 F3819.37
M106 S229.5
G1 X200.309 Y219.855 E.04172
M106 S221.85
G1 X199.503 Y219.414 E.03155
```
But the current klipper firmware can still give out the same problem. It will be expressed 2-3 times less than on a *bad* slicer, but it will not disappear anywhere.
Even if the fan does not have time to accelerate to the specified speed, this is ok. It is much more important to do it synchronously. 
DynamicFanSpeedExample
The slicer draws a gradient-changing fan speed and it works quite well, although it may require specifying 10-20 reference speeds per second and klipper can not handle this.

> The `FAN_MIN_TIME` value is not a setting that users can reasonably configure. It's not an attribute of the fan hardware, it's an upper bound of the worst case transmission latency from host to MCU that we feel the Klipper system can reasonably support. If a user were to notably decrease this value then it would notably increase the risk of cryptic "timer too close" errors.
> 
> Support for many rapid fan speed changes would require adding support for queuing of those fan speed changes in the micro-controller (along with the corresponding memory management for that queue). That is a notable amount of work. There's no real interesting in doing that work (nor interest in maintaining the resulting code) because, as mentioned earlier in this thread, it's completely pointless to make rapid changes to fan speed. The best solution is to disable these broken options in the slicer.

I don't see anything in common between the transmission latency and the minimum interval between setpoints.
I agree that FAN_MIN_TIME is not an attribute of the fan hardware and should actually be specified in a more complex way.
Moreover, in my opinion, the FAN_MIN_TIME parameter should not exist at all, and in its place there should be a mechanism analyzing the nearest changes in the time interval and setting the total weight average of the used time of this speed on this gcode fragment. Such an algorithm could help speed up and slow down the fan, like pressure_advance. However, this is a colossal overcomplication of the code base, and let's be honest, no one really needs it. Therefore, I propose a method that affects the firmware in a minimal way. Perhaps someone would suggest a better way to solve the problem in addition to simply turning off the function that *works crookedly in our firmware*.

> Separately, commits need to have a signed-off-by line in order for them to be merged - see: https://www.klipper3d.org/CONTRIBUTING.html

My bad. Fixed it

KevinOConnor:
For what it is worth, this conversation may be more appropriate on Klipper Discourse - in particular see the recent conversation at https://klipper.discourse.group/t/clock-rollover-and-fan-commands-causes-intermittent-timer-too-close-issues/9665

-Kevin

PR #6344, opened 2023-09-18
docs: add info about requesting bootloaders

Laikulo:


docs/Bootloader_Entry.md
```diff
@@ -0,0 +1,119 @@
...
+is being used to connect to it), sending the string
+`<SPACE><FS><SPACE>Request Serial Bootloader!!<SPACE>~`.
+
+`<SPACE>` is an ASCII literal space, 0x20.\
+`<FS>` is the ASCII File Separator,
+0x1c.
+
+Note that this is not a valid message as per the
+[MCU Protocol](Protocol.md#micro-controller-interface), but sync characters(`~`)
+are still respected.\
```
KevinOConnor, line 54: I noticed several instances in this file where the line ends in a backslash (`\`).  Why is this?

KevinOConnor (commented):
Thanks.  Seems fine to me.  I'll give a few days to see if others have comments.

I do have one question - see below.

-Kevin

docs/Bootloader_Entry.md
Laikulo: This causes a line break to be inserted when rendering as HTML, otherwise it takes a blank line, or two spaces at the end of a line.

When I wrote the file, I had used double spaces, but running it through `mdformat` swapped them for backslashes.

KevinOConnor: Okay, thanks.  I'd recommend inserting blank lines instead.  Ideally the documentation will be generally readable in text, on github, and on the klipper3d.org site - using a backslash seems subtle and I fear it may not render well on all three locations.

-Kevin

Laikulo: Updated

KevinOConnor:
Thanks.

-Kevin

PR #6345, opened 2023-09-19
Rewrote some parts of bed_mesh.py

BillyNate:
Backstory can be found on [klipper.discourse](https://klipper.discourse.group/t/bed-mesh-first-probed-coordinate/10504)

Here's what I did:
- Removed the error message on L489
- Changed the check on L336 to accept negative distances
- Made `within()` on L25 more generic/robust: Now accepts "reversed" coordinates, needed to keep faulty regions working
- Rewrote creation of 2D list on L722: It is now based on the min/max positions & x_cnt/y_cnt, accepting points in any order, instead of relying on an expected zig-zag order only.
- Adjusted y-axis length check on L739 to handle changed 2D array
- Rewrote probed values extrapolation for round beds on L746:
  - Now handles the new 2D array setup
  - Instead of extrapolating by copying the previous x-axis z_offset it now works its way from the inside out and taking the average of two neighboring z_offsets
- Replaced x-axis length check on L764 for a final check if all 2D array mesh positions are set

This code has been tested with settings for
- a rectangular bed with low `mesh_min` / high `mesh_max`
- a rectangular bed with high `mesh_min` / low `mesh_max`
- a rectangular bed with a faulty region
- a round bed

This still definitely needs testing/reviewing, so if anyone wants give it a try, I'd be pleased.
Any feedback would be welcome!

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -22,8 +22,10 @@ def isclose(a, b, rel_tol=1e-09, abs_tol=0.0):
     return abs(a-b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
 
 # return true if a coordinate is within the region
-# specified by min_c and max_c
-def within(coord, min_c, max_c, tol=0.0):
+# specified by c_a and c_b
+def within(coord, c_a, c_b, tol=0.0):
+    min_c = [min(c_a[0], c_b[0]), min(c_a[1], c_b[1])]
+    max_c = [max(c_a[0], c_b[0]), max(c_a[1], c_b[1])]
     return (max_c[0] + tol) >= coord[0] >= (min_c[0] - tol) and \
         (max_c[1] + tol) >= coord[1] >= (min_c[1] - tol)
 
@@ -333,7 +335,7 @@ class BedMeshCalibrate:
         # floor distances down to next hundredth
         x_dist = math.floor(x_dist * 100) / 100
         y_dist = math.floor(y_dist * 100) / 100
-        if x_dist < 1. or y_dist < 1.:
+        if abs(x_dist) < 1. or abs(y_dist) < 1.:
             raise error("bed_mesh: min/max points too close together")
 
         if self.radius is not None:
@@ -486,8 +488,7 @@ class BedMeshCalibrate:
             x_cnt, y_cnt = parse_config_pair(config, 'probe_count', 3, minval=3)
             min_x, min_y = config.getfloatlist('mesh_min', count=2)
             max_x, max_y = config.getfloatlist('mesh_max', count=2)
-            if max_x <= min_x or max_y <= min_y:
-                raise config.error('bed_mesh: invalid min/max points')
+
         orig_cfg['x_count'] = mesh_cfg['x_count'] = x_cnt
         orig_cfg['y_count'] = mesh_cfg['y_count'] = y_cnt
         orig_cfg['mesh_min'] = self.mesh_min = (min_x, min_y)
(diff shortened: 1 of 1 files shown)


JamesH1978:

Thanks
James

BillyNate:
My apologies for the missing signed-off-by line 🙈 
I added it just now.

Sineos:
Ping @Arksine

Arksine:
Thanks.  I do have some high level observations and comments about these proposed changes. 

> - Removed the error message on L489
> - Changed the check on L336 to accept negative distances
> - Made within() on L25 more generic/robust: Now accepts "reversed" coordinates, needed to keep faulty regions working

The above changes seem to be centered around accepting invalid configuration.  I don't agree with this, I think its safer and easier to document when we are explicit about the coordinates accepted.  The `mesh_min`, `mesh_max`, and options related to faulty regions have documentation specifying how they should be configured relative to the origin.  That said, it may be useful to add a validation step to the faulty regions similar to that of the mesh, and be more explicit in the documentation about how they should be configured.

> - Rewrote creation of 2D list on L722: It is now based on the min/max positions & x_cnt/y_cnt, accepting points in any order, instead of relying on an expected zig-zag order only.
> - Adjusted y-axis length check on L739 to handle changed 2D array

Is there an immediate functional benefit to this change?  The code seems fine, however any change risks introducing an unforeseen regression.  The stated reasoning to to accept points in any order, however currently they will only be received according to the zigzag pattern.

> - Rewrote probed values extrapolation for round beds on L746:
>   - Now handles the new 2D array setup
>   - Instead of extrapolating by copying the previous x-axis z_offset it now works its way from the inside out and taking the average of two neighboring z_offsets
> - Replaced x-axis length check on L764 for a final check if all 2D array mesh positions are set

Part of the above changes seem to be related to the proposal to change the way the probed matrix is generated, which was addressed in the previous comment.  The primary new addition appears to be a change in how round beds extrapolate points outside of the probed area.  Is there any data to suggest that averaging the two nearest points provides a real world benefit over the current extrapolation?  I had considered this during the original implementation and decided that it likely wouldn't make a difference, however I am open to data showing otherwise.

Arksine:
> Theoretically a stepper in Klipper could be configured to have position_min: -300 and position_max: 0, right?
A mesh_min: -300, -300 isn't "nearest to the origin" and mesh_min: 0, 0 would trigger the error.

Yes, this could be done in theory, but I can think of no valid reason to deviate from recommended configurations, a user would be asking for nothing but trouble. The origin can be placed at any corner, it depends on your stepper configuration and which way moves toolhead the positive direction, so it isn't really necessary to operate in negative coordinates when dealing with cartesian-like kinematics.  That said, the documentation could be reworded to state "the lowest value on both the X and Y axes" or something similar.  The option name of `mesh_min` is somewhat self explanatory as well.

> I might add other patterns next to the zigzag pattern later on (primarily to influence the last point probed). Is this of interest to the community?

I'm willing consider additional patterns, but there would need to be a substantial audience and a valid functional reason for doing this.  While I empathize with the reason of "because I'd like to", the maintenance burden of introducing changes to `bed_mesh` requires a higher standard.

> Several Klipper interfaces show a visualization of the bed mesh. This of course includes the extrapolated points, causing confusion with users on round beds. How do you feel about not extrapolating at all, but only saving the actual probed points?

Making such a change would likely break the current visualizations and other functionality that relies on the points reported by `bed_mesh`.  It should possible for front-ends to better accommodate round beds without significant changes to `bed_mesh`.   We may need to add some additional reporting, but I don't think its necessary to refactor the extrapolation step.

BillyNate:
Thank you for elaborating on this.
I guess the pull request can be closed now?
If there's anything else I can do concerning the bed mesh implementation, please let me know.

PR #6346, opened 2023-09-21
pid_calibrate: allowing TUNE_PID_DELTA to pass through gcode

SHKinsem:
I have utilized the Bambulab hotend paired with a 70W ceramic heater cartridge. After extensive testing, it became evident that this setup exhibited an exceptionally rapid temperature response. This led to an issue where the original delta value was too small, causing the temperature sensor to be unable to collect data quickly enough, resulting in suboptimal PID calibration parameters. By adjusting the delta value, significant improvements can be achieved in the effectiveness of the automatic calibration process.

Signed-off-by: Jaccob Hui <783739330@qq.com>


SHKinsem:
With the change, new command would be 'PID_CALIBRATE HEATER= TARGET= TUNE_PID_DELTA='
if TUNE_PID_DELTA is not specified, default value would be 5.0

PR #6355, opened 2023-09-26
Linux: Add Using gpio with wiringPi

how2flow:
Addition of gpio control scheme.

in a wirringPi-enabled environment,
Add Using wirringPi Option.

Rather than requesting R/W through ioctl
Expect more performance improvement with direct access through wiringPi's memory mapping(mmap).

### DIFF
--- a/src/linux/Kconfig
+++ b/src/linux/Kconfig
@@ -20,4 +20,18 @@ config CLOCK_FREQ
     int
     default 50000000
 
+config WIRINGPI
+    bool "Use wiringPi"
+    default n
+    depends on HAVE_GPIO
+    help
+        Gpio control with wiringPi API.
+        Unlike conventional methods of opening files and requesting R/W operations,
+        it is a method of putting values directly into memory mapped in mmap.
+        It will improve speed.
+
+        Require some packages.
+        [libwiringpi2], [libwiringpi-dev] is required.
+        In other words, wiringPi must be supported for use it.
+
 endif
--- a/src/linux/Makefile
+++ b/src/linux/Makefile
@@ -4,8 +4,17 @@ dirs-y += src/linux src/generic
 
 src-y += linux/main.c linux/timer.c linux/console.c linux/watchdog.c
 src-y += linux/pca9685.c linux/spidev.c linux/analog.c linux/hard_pwm.c
-src-y += linux/i2c.c linux/gpio.c generic/crc16_ccitt.c generic/alloc.c
+src-y += linux/i2c.c generic/crc16_ccitt.c generic/alloc.c
 src-y += linux/sensor_ds18b20.c
+
+ifeq ($(CONFIG_WIRINGPI),y)
+	CFLAGS += -DWIRINGPI
+	CFLAGS += -I/usr/local/include -I/usr/include/wiringpi2
+	CFLAGS_klipper.elf += -lwiringPiDev -lwiringPi -lm -lcrypt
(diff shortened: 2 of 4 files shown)


KevinOConnor:
Thanks.  I don't know much about wiringPi, but when I did a web search it did not appear to have a GPL compatible license.  As such, I don't think it is a good fit for Klipper (which is GNU GPLv3).

Maybe I am missing something though.

-Kevin

Sineos:
> it did not appear to have a GPL compatible license

Isn't GNU Lesser Public License v3 compatible to GPL (thought it is more permissive even). See http://wiringpi.com/download-and-install/ at the end of the page and https://github.com/WiringPi/WiringPi

FWIW, I cannot much contribute to the usefulness of it in the context of Klipper, but it is IMO one of the best solutions for serious GPIO work with the RPi. I'm using it for some home automation stuff.

Unfortunately, it is no longer actively developed except some minor stuff under the above mention GH project. Sad to say, but I doubt that it is a sustainable solution.

how2flow:
> Unfortunately, it is no longer actively developed except some minor stuff under the above mention GH project. Sad to say, but I doubt that it is a sustainable solution.

I also take this part the most, but in controlling gpio, "wiringPi" uses mmap to read/write values directly to the register.
That's why there doesn't seem to be a need to update the code to improve performance or add functionality.

I think the packages that have been used for a long time(It's still used by many users), and wiringPi support products, like RPi/Odroids, will be updated with new revisions whenever new boards are released.
(In the case of Odroids, I'm updating it myself..   https://github.com/hardkernel/wiringPi)

+) I don't think there's a big problem with the license. LGPL3 is compatible with GPL3

But if you don't think it fits the direction of the klipper, just think that there is a method like this.
There was no error in the running, and I'm enjoying printing by applying this code in my local :)

KevinOConnor:
Thanks.  LGPL is fine.  I guess my confusion is that the first paragraph at wiringpi.com states that the source code is not publicly available (specifically, `The source code is not publicly available but may be made available to those who wish commercial support.`).

If someone can point me to the actual code for wiringpi (it would seem wiringpi.com is not the actual location for it) then I can take another look.

Cheers,
-Kevin

Sineos:
It is my understanding the source was once published under https://git.drogon.net/ but no longer is since the official maintainer stopped. The closest thing left is https://github.com/WiringPi/WiringPi

KevinOConnor:
Thanks.  If the upstream is now https://github.com/WiringPi/WiringPi then that is fine (as that repo appears to be LGPL-v3).  I'll continue with a review if the original submitter can confirm that I should continue.

-Kevin

how2flow:
I think this solution is very nice.

But since the official development has stopped,
Raspberry Pi's new RPi5 model may not be compatible.
(It's possible if the user ports it separately locally..)
Of course, I'm an Odroid wiringpi package manager, so I do keep updating this side.

I think add descript like it's only supported up to Raspberry Pie 4, Or,
If I have time, I can port wiringpi to RPi5 in my repo, and give you a guide.

All of this is possible because I gave the use of wiringpi as an option in this commit.

Thanks, Have a nice day

Sineos:
Interesting: https://github.com/WiringPi/WiringPi/issues/186

While I agree that an advanced GPIO manipulation library may have its merits, I think WiringPi is not a future-proof choice.

thijstriemstra:
>  I think WiringPi is not a future-proof choice.

Agreed. Someone will have to maintain this new code @how2flow..

PR #6356, opened 2023-09-27
Add manual installation instructions

Michal-Szczepaniak:
I found that in personal experience, when you're not running on debian or debian-based distros you have very hard time getting any instructions for installing klipper and best for now are the installation scripts. So I took whats inside those scripts and compiled them into manual installation instruction for anyone like me that is not running debian-based distro or even raspberry pi for that matter. I will adjust if any changes will be necessary.

### DIFF
--- a/docs/Manual_Installation.md
+++ b/docs/Manual_Installation.md
@@ -1,21 +1,21 @@
 # Manual Installation
 
-These instructions are for people who want to run klipper on 
+These instructions are for people who want to run klipper on
 custom system that's not debian-based. Flashing steps are same
-as in [Installation](Installation.md) instruction. 
+as in [Installation](Installation.md) instruction.
 Naturally this instruction is for more advanced users.
 
 # Preparing for installation
 
-First you need to install required packages. This step is different 
+First you need to install required packages. This step is different
 for every other package manager and distribution so provided are
-debian packages and fiding package names for your distribution are 
+debian packages and fiding package names for your distribution are
 up to you.
 
 ```
 virtualenv python-dev libffi-dev build-essential libncurses-dev
-libusb-dev avrdude gcc-avr binutils-avr avr-libc stm32flash 
-libnewlib-arm-none-eabi gcc-arm-none-eabi binutils-arm-none-eabi 
+libusb-dev avrdude gcc-avr binutils-avr avr-libc stm32flash
+libnewlib-arm-none-eabi gcc-arm-none-eabi binutils-arm-none-eabi
 libusb-1.0 pkg-config
 ```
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Can you elaborate on the advantage of adding this document as opposed to pointing users to the script?  I would think that users able to manually install would be able to understand the steps detailed in the basic installation script.

-Kevin

Michal-Szczepaniak:
I was told that the scripts would be deleted in favour of https://github.com/dw-0/kiauh. Additionally it's quite nicer to have description rather to read script directly.

docs/Manual_Installation.md
```diff
@@ -0,0 +1,109 @@
...
+These instructions are for people who want to run klipper on
+custom system that's not debian-based. Flashing steps are same
+as in [Installation](Installation.md) instruction.
+Naturally this instruction is for more advanced users.
+
+# Preparing for installation
+
+First you need to install required packages. This step is different
+for every other package manager and distribution so provided are
+debian packages and fiding package names for your distribution are
```
SteveGotthardt, line 12: 'fiding' should be finding

Michal-Szczepaniak: Will fix but i'll delay in case you find more. Thanks for help!

Michal-Szczepaniak:
@KevinOConnor I really think this should be merged, there's nothing wrong with adding helpful instructions isn't it? I myself would prefer to have it nicely in documentation rather than searching code for installation scripts and then deciphering what has to be done. It would really have potential to help people out

Sineos:
While your instructions seem to be correct, I'm not sure if they do bring more clarity to the users or rather the opposite.
* Someone who is experienced in Linux will not really need them
* Someone who lacks this experience will not be able to follow them, e.g. latest at stages like
    > As a root, create file in [...]

    or

    > Of course you need to adjust user, `venv_directory`, [...]

Michal-Szczepaniak:
@Sineos I disagree "Someone who is experienced in Linux will not really need them", then we can just as well remove the normal installation instructions as non-experienced people will also not need them they can just run script. It will take them both several more hours without instructions but they don't need them right? Even if you're experienced its much faster to have nicely documented instructions rather than spending hours trying to figure out what am i supposed to do. I don't even know why you're both opposed to making life easier for people

someone who's not as experienced can ask for help on discord or forums

naikymen:
I second Michal's view. I am an experienced Linux user, and I would have loved to follow these instructions instead of also investing time in reverse-engineering the dependencies.

This PR is helpful and should be merged! :)

PR #6357, opened 2023-09-27
probe: use the x coordinates from the correct toolhead after probing with dual carriage

HelgeKeck:
After probing with the dual_carriage activated, klipper sets the x coordinate to the x pos of the x carriage.

This pr fixes that and lets you probe with any of the toolheads

this is just for the `PROBE` g-code command, there is a similar issue after z-homing, but more complex

Signed-off-by: Helge Magnus Keck [HelgeKeck@hotmail.com](mailto:HelgeKeck@hotmail.com)

### DIFF
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -122,6 +122,10 @@ class PrinterProbe:
         pos[2] = self.z_position
         try:
             epos = phoming.probing_move(self.mcu_probe, pos, speed)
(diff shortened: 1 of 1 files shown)


HelgeKeck:
there are multiple places where this happens

HelgeKeck:
this is fixed by
https://github.com/Klipper3d/klipper/pull/6351

PR #6359, opened 2023-09-29
Apply 'strip' on the tool version string to remove any whitespace con…

pbleyer:
…trol characters that can generate erroneous code.

The 'tool_versions' function in buildcommands.py does not account for other whitespace control characters beside newline, ('\n'). In particular, compiling in Windows under the MSYS2 environment, the DOS-style line returns ('\r\n') output by the GNU tools break the correct generation of the compile_time_request.c module.


JamesH1978:

Thanks
James

PR #6219, opened 2023-05-21
Updated bed mesh configuration for Creality Ender 3 S1 and S1 Pro

Auxx:
Existing `mesh_min` values start probing too far away from the edge.

Additionally `mesh_max` Y value of 197 is out of bed movement range (197 + `bltouch.y_offset` = 237.5) and causes Y stepper to overhit, which can potentially damage the printer in the long run.

Signed-off-by: Aleksandr Ivanov <aux@hexmode.org>

PR #6360, opened 2023-10-03
Ender3S1 Stepper_Y limit

ThomasNyk:
Fixed the Ender3 S1 crashing on the high end of the Y Axis

This necessitated an update to the bed mesh limits as well

I also changed from a 4x4 bed mesh to 5x5 as that is what i believe the original Ender 3 S1 uses, but I am not sure

Signed-off-by: Thomas Nykjær nykjaerthomas+github@gmail.com

### DIFF
--- a/config/printer-creality-ender3-s1-2021.cfg
+++ b/config/printer-creality-ender3-s1-2021.cfg
@@ -38,8 +38,8 @@ microsteps: 16
 rotation_distance: 40
 endstop_pin: !PA6
 position_endstop: -8
-position_max: 238
-position_min: -13
+position_max: 223
+position_min: -10
 homing_speed: 50
 
 [stepper_z]
@@ -110,9 +110,9 @@ stow_on_each_sample: false
 
 [bed_mesh]
 speed: 120
-mesh_min: 10, 10
-mesh_max: 200, 194
-probe_count: 4,4
+mesh_min: 10, 20
+mesh_max: 200, 182
+probe_count: 5,5
 algorithm: bicubic
 
 [safe_z_home]


Fisheiyy:
Looks fine, I thought it was already set to a 5x5 mesh but I guess not.

JamesH1978 (approved):
Looks fine to me

Thanks
James

Luro02:
The official marlin firmware uses the same values for `position_min`: https://github.com/CrealityOfficial/Ender-3S1/blob/b77d06db5a57a8b7743c7ccc76bdf714e244cf6a/Marlin/Configuration.h#L1370

How has the maximum position been chosen? The marlin config sets it to `MAX_X = 230` and `MAX_Y = 220`

JamesH1978:
All example configurations are community submitted by users, the klipper project itself does not have access to every known printer to do full and complete testing, we have to rely on the person submitting to commit a sane config that others can build from. This is why they are considered examples.

Thanks
James

Luro02:
The values look a bit arbitrary, so I measured the true physical limits of my printer. I measured them, by increasing the limits in the config, homing, then moving with G0 to the max/min positions.

This is the result:
| Firmware               | Stock Firmware                                 | E3V2S1 FW       | Klipper Master  | Klipper PR      | Physical limits |
|------------------------|------------------------------------------------|-----------------|-----------------|-----------------|-----------------|
| Max Position (x, y, z) | (230, 220, 270)                                | (230, 220, 270) | (235, 238, 270) | (235, 223, 270) | (252, 240, 278) |
| Min Position (x, y, z) | (-10, -8, 0)                                   | (-10, -8, 0)    | (-15, -13, -4)  | (-15, -10, -4)  | (-10, -8, 0)    |

For the stock firmware there are comments for `y_min = -10` and `z_max = 250`.

The min positions seem to be wrong. I was unable to move the printer farther than the endstops.
I do not understand why `MIN_Z` is set to -4.

The configuration reference does not mention why Z should go below zero:

> ```ini
> #position_min: 0
> #   Minimum valid distance (in mm) the user may command the stepper to
> #   move to.  The default is 0mm.
> ```

The extruder fan is almost level with the nozzle, so even when moving outside of the bed (to prevent the nozzle from crashing into the bed), the fan would hit the bed at Z=-1.

At X = 230 Y = 230 the nozzle seems to be in the corner of the bed, but for Y ~2.8mm are used by the alignment screws of the bed, so the realistic maximum for printing would be X = 230 Y = 227.

----

I think the suggested values by this PR for `MAX_X` and `MAX_Y` are good default values. How about adding comments that mention the true physical limits?

```diff
[stepper_x]
step_pin: PC2
dir_pin: PB9
enable_pin: !PC3
microsteps: 16
rotation_distance: 40
endstop_pin: !PA5
position_endstop: -10
+ # Physical Limit: ~252
position_max: 235
- position_min: -15
+ position_min: -10
homing_speed: 50

[stepper_y]
step_pin: PB8
dir_pin: PB7
enable_pin: !PC3
microsteps: 16
rotation_distance: 40
endstop_pin: !PA6
position_endstop: -8
+ # Physical Limit: ~240
position_max: 223
- position_min: -13
+ position_min: -8
homing_speed: 50

[stepper_z]
step_pin: PB6
dir_pin: !PB5
enable_pin: !PC3
microsteps: 16
rotation_distance: 8
endstop_pin: probe:z_virtual_endstop
+ # Physical Limit: ~278
position_max: 270
- position_min: -4
+ position_min: 0
```

Should I create a separate PR for this (and for fixing the min positions)?

Sineos:
> I do not understand why `MIN_Z` is set to -4.
> The configuration reference does not mention why Z should go below zero:

See https://www.klipper3d.org/Bed_Level.html?h=negative#choose-the-appropriate-calibration-mechanism

JamesH1978:
This is not marlin, never was. The min position is set to allow the endstop position which is at a minus value, therefore needs to be allowed, so that 0,0 is set correctly. As default 0 in klipper the nozzle is not at perfect front left position.

This PR is to change a setting in an existing config, if you would like to propose your own changes, then please open a new PR.

Thanks
James

KevinOConnor:
Thanks.  I'm fine with making this change of course.  However, it seems we just changed the bed mesh range in #6219 - can @auxx confirm this change here also looks good?

-Kevin

Auxx:
I have never had any issues after my changes, so I'm not sure if these limits are needed. Will they change any positioning maths/calculations?

Auxx:
> I don't know about the calculations, all i know is that my printer along with at least two other people from the klipper discord had the same issue as me. When calibrating the bed mesh, the y axis would ram the end of the rail.

Can you double check that you have my changes in your config and still have axis crashing issue? Mesh settings should be:

```
mesh_min: 10, 10
mesh_max: 200, 194
```

Instead of

```
mesh_min: 20, 20
mesh_max: 200, 197
```

MrSpoocy:
I can agree with @ThomasNyk, I update my Ender3S1 (F4) and also crash on the y axis.

zackbcom:
Used these to stop my bed from crashing when doing a mesh. Thanks.

PR #6361, opened 2023-10-05
doc: tiny Py2 cleanup

unixcarbide:
super minor doc updates which referred to python2

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -135,8 +135,8 @@ packages (e.g. `sudo apt install python3-numpy python3-matplotlib`) to
 continue using these scripts. For more details, refer to
 [Software installation](Measuring_Resonances.md#software-installation).
 Alternatively, users can temporarily force the execution of these scripts
-under Python 2 by explicitly calling Python2 interpretor in the console:
-`python2 ~/klipper/scripts/calibrate_shaper.py ...`
+under Python 2 by explicitly calling Python3 interpretor in the console:
+`python3 ~/klipper/scripts/calibrate_shaper.py ...`
 
 20211110: The "NTC 100K beta 3950" temperature sensor is deprecated.
 This sensor will be removed in the near future.  Most users will find
--- a/docs/Packaging.md
+++ b/docs/Packaging.md
@@ -9,7 +9,7 @@ are as follows:
 Klipper uses a C module to handle some kinematics calculations more quickly.
 This module needs to be compiled at packaging time to avoid introducing a
 runtime dependency on a compiler. To compile the C module, run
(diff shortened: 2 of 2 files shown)


docs/Config_Changes.md
```diff
@@ -135,8 +135,8 @@ packages (e.g. `sudo apt install python3-numpy python3-matplotlib`) to
 continue using these scripts. For more details, refer to
 [Software installation](Measuring_Resonances.md#software-installation).
 Alternatively, users can temporarily force the execution of these scripts
-under Python 2 by explicitly calling Python2 interpretor in the console:
-`python2 ~/klipper/scripts/calibrate_shaper.py ...`
+under Python 2 by explicitly calling Python3 interpretor in the console:
+`python3 ~/klipper/scripts/calibrate_shaper.py ...`
```
unixcarbide, line 139: self-review:

**Problem this submission fixes:**
This python file being executed appears to have been updated to python3 since 2021, https://github.com/Klipper3d/klipper/commit/309b8e030f68329d04573042ffc53a2745ac088b

Most UNIX systems don't ship with python2 any more, (and it's getting harder to maintain the python2 ecosystem), so this line trips up anyone trying to follow the instructions.

This small submission is high impact to the audience of users who are following this documentation.

docs/Packaging.md
```diff
@@ -8,8 +8,8 @@ are as follows:
 
 Klipper uses a C module to handle some kinematics calculations more quickly.
 This module needs to be compiled at packaging time to avoid introducing a
-runtime dependency on a compiler. To compile the C module, run `python2
-klippy/chelper/__init__.py`.
+runtime dependency on a compiler. To compile the C module, run
+`~/klippy-env/bin/python klippy/chelper/__init__.py`.
```
unixcarbide, line 12: self-review:

Problem this submission fixes:
Instruction to use `python2`, creates numerous stumling blocks for users following this documentation, primarily, that most folks no longer have a working python2 interperter on their system.
The change here, to simply use the python which should have been set up already, makes this documentation run correctly regardless of which python the user has installed- as well as more likely to contain the libraries required by the code which is run.

Most UNIX systems don't ship with python2 any more, (and it's getting harder to maintain the python2 ecosystem), so this line trips up anyone trying to follow the instructions.

This small submission is high impact to the audience of users who are following this documentation.

unixcarbide (commented):
Self-review for small documentation changes.

KevinOConnor:
Thanks.  I'm not sure this change is an improvement.  In particular, the first change looks like an error (one can't force the use of python2 by using python3).  The second change looks fine, but seems unnecessary as anyone reading that document should know how to manually invoke python.

-Kevin

PR #6362, opened 2023-10-09
klippy/extras: Add boolean filter for gcode_macro.

Luro02:
This allows one to write
```
{% set FLAG = params.FLAG|boolean %}
```
instead of
```
{% set FLAG = (params.FLAG|string).lowercase() == "true" %}
```

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -72,6 +72,14 @@ class PrinterGCodeMacro:
     def __init__(self, config):
         self.printer = config.get_printer()
         self.env = jinja2.Environment('{%', '%}', '{', '}')
+
+        # jinja does not provide a boolean filter, so a filter is added here
+        def boolean(value):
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, I think we should try to limit the amount of Jinja2 customizations.  Adding these customizations would require associated documentation (so that users knew of its existence).  I agree that Jinja2 is cryptic, but I don't think we can realistically make Jinja2 "not cryptic".  I fear at best we can only introduce "many different kinds of cryptic".  Users today can use the standard Jinja2 documentation (which is thorough) to figure out appropriate syntax, and I think we should encourage that going forward.

Cheers,
-Kevin

flowerysong:
> ```
> % set FLAG = (params.FLAG|string).lowercase() == "true" %}
> ```
A more idiomatic way to do this is
```
{% set FLAG = params.FLAG | lower == "true" %}
```

PR #6364, opened 2023-10-09
heater_fan: Add Heater Fan Self Check Logic

cneshi:
Add a simple self-check functionality to heater_fans with tachometer signal.

### Why do we want this?

This functionality is especially useful for hotend fans, which when they fail can result in some quite catastophic consequences for the hotend and surrounding parts.

### Configuration

The user simply adds configurations related to fan tachometer (tachometer_pin, tachometer_ppr) and one additional configuration:
`min_rpm`, which specifies the minimum expected RPM of the fan when it is enabled. When fan speed is detected to be below the minimum RPM. then klipper shuts down with an error. By default, `min_rpm` is 0 and thus the self-test functionality is not enabled.

### Code description

- **Initialization**, the heater_fan module checks for `min_rpm` config, and sets it to the specified value. A config error is thrown if tachometer configuration is not specified by the user but `min_rpm` is.
- **Event handler**, a `fan_check()` event handler is created if min_rpm is enabled. This event handler is first called 3 sec after startup and every 1.5 sec afterwards.
- **Self-check logic**, the main logic for fan self-check is implemented in `fan_check()` which is called periodically. Each time `fan_check` is called the current fan speed and it's related heater temperature is read. If heater temp is above `heater_temp` and current fan speed is lower than `min_rpm`, the error counter is incremented; otherwise the error counter is reset to zero. An error is thrown if the error counter reaches 4 and klipper is shutdown.

### DIFF
--- a/klippy/extras/heater_fan.py
+++ b/klippy/extras/heater_fan.py
@@ -59,8 +59,8 @@ class PrinterHeaterFan:
         if has_err:
             self.num_err += 1
             if self.num_err > self.max_err:
-                msg = "'%s' spinning below minimum safe speed of %d rev/min"
-                    % (self.name, self.min_rpm)
+                msg = "'%s' spinning below minimum safe speed of %d rev/min" % (
+                    self.name, self.min_rpm)
                 logging.error(msg)
                 self.printer.invoke_shutdown(msg)
                 return self.printer.get_reactor().NEVER
--- a/klippy/extras/heater_fan.py
+++ b/klippy/extras/heater_fan.py
@@ -55,17 +55,16 @@ class PrinterHeaterFan:
             for heater in self.heaters:
                 current_temp, target_temp = heater.get_temp(eventtime)
                 if current_temp > self.heater_temp and rpm < self.min_rpm:
-                    has_err = True
-        if has_err:
-            self.num_err += 1
-            if self.num_err > self.max_err:
-                msg = "'%s' spinning below minimum safe speed of %d rev/min" % (
-                    self.name, self.min_rpm)
-                logging.error(msg)
-                self.printer.invoke_shutdown(msg)
-                return self.printer.get_reactor().NEVER
-        else:
-            self.num_err = 0
+                    self.num_err += 1
+                    break
+            else:
+                self.num_err = 0
+        if self.num_err > self.max_err:
+            msg = "'%s' spinning below minimum safe speed of %d rev/min" % (
+                self.name, self.min_rpm)
+            logging.error(msg)
+            self.printer.invoke_shutdown(msg)
+            return self.printer.get_reactor().NEVER
         return eventtime + 1.5
 
 def load_config_prefix(config):
--- a/klippy/extras/heater_fan.py
+++ b/klippy/extras/heater_fan.py
@@ -6,6 +6,7 @@
 from . import fan
 
 PIN_MIN_TIME = 0.100
+SAFETY_CHECK_INIT_TIME = 3.
 
 class PrinterHeaterFan:
     def __init__(self, config):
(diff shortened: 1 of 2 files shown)


rogerlz:
That's a very nice feature, but I'd love to be able to decide the fail behaviour, e.g. call a gcode, or shut down the heater + pause the print.

camerony:
Is the submission free of defects and is it ready to be widely deployed? The code has been tested and verified to work by multiple testers and LDO engineers

Regression:
whitespace check : PASSED
data dictionaries (klipper-dict-20230921.tar.gz): PASSED


image

bennierex:
Would be nice to be able to specify the threshold for the error counter to cause a shutdown. With the fixed value of 4, it will take at least 6 seconds to shutdown after the fan fails. 

You could also check if the previous x readings are below min_rpm. That way it will not trigger a shutdown on 4 misreadings on the tach pin during a long period of time.

cneshi:
> Would be nice to be able to specify the threshold for the error counter to cause a shutdown. With the fixed value of 4, it will take at least 6 seconds to shutdown after the fan fails.
> 
> You could also check if the previous x readings are below min_rpm. That way it will not trigger a shutdown on 4 misreadings on the tach pin during a long period of time.

I opted to make it a fixed value of 4 because I feel like that is a good general value for most situations. If we let the user specify the value, that is another config option they need to take time to read about, understand, and configure. Anyways that is just my opinion

The logic is already such that it must be 4 consecutive below `min_rpm` readings to trigger the shutdown - so misreadings do not accumulate over time.

Rudis1261:
Awesome stuff!

klippy/extras/heater_fan.py
```diff
@@ -37,6 +48,25 @@ def callback(self, eventtime):
...
+            self.num_err += 1
+            if self.num_err > self.max_err:
+                msg = "'%s' spinning below minimum safe speed of %d rev/min" % (
+                    self.name, self.min_rpm)
+                logging.error(msg)
+                self.printer.invoke_shutdown(msg)
+                return self.printer.get_reactor().NEVER
+        else:
+            self.num_err = 0
+        return eventtime + 1.5
```
EmJay276, line 69: I have some suggestions to simplify the code

```suggestion
    def fan_check(self, eventtime):
        rpm = self.fan.tachometer.get_status(eventtime)['rpm']
        has_err = False
        if rpm is not None:
            for heater in self.heaters:
                current_temp, target_temp = heater.get_temp(eventtime)
                if current_temp > self.heater_temp and rpm < self.min_rpm:
                    self.num_err += 1
                    break
            else:
                self.num_err = 0
        if self.num_err > self.max_err:
            msg = "'%s' spinning below minimum safe speed of %d rev/min" % (
                self.name, self.min_rpm)
            logging.error(msg)
            self.printer.invoke_shutdown(msg)
            return self.printer.get_reactor().NEVER
        return eventtime + 1.5
```

EmJay276 (commented):
I have a suggestions to simplify the code

klippy/extras/heater_fan.py
cneshi: Thank you for this suggestion. I think it also makes the coder easier to read.

EmJay276 (approved):

wuyoutech:
I did almost the same thing, but I chose to modify fan.py instead of heaterfan.py so that this feature can cover all fans. 

My friend's printer lost  heater_fan while printing, and large pieces of material spilled on to the bed, The toolhead eventually collided with them and alomst destory everything. 

I'm glad someone did the samething, I will commit my code for reference as soon as possible.

PS: English is not my native language, please ignore any grammar issues, and I would be very pleased if you point out my issue.

KevinOConnor:
Thanks.  It's fine with me if we want to check for tachometer errors.  It does seem odd to me that we would add this code only to heater_fan fans though.  It would seem that if we were to implement this code it would be preferable to implement it for all fans.

There is a PR #6429 that does something similar.

-Kevin

KevinOConnor:
Thanks.  As high-level feedback I think it is fine to raise an error on a malfunctioning fan.  However, it seems odd to me that this check is only made available to `heater_fan` instances instead of on all fans (as is done in #6429).  Also, the code here does not look correct to me as it seems like it would have the same issue described at https://github.com/Klipper3d/klipper/pull/6429#issuecomment-1905056206 .

-Kevin

PR #6366, opened 2023-10-11
same70: Add CAN

Wulfsta:
Add CAN to same70. Thanks to @eamaclean for the help.

Depends on [#6365](https://github.com/Klipper3d/klipper/pull/6365).

### DIFF
--- a/src/atsam/fdcan.c
+++ b/src/atsam/fdcan.c
@@ -1,0 +1,307 @@
+// CANbus support on atsame70 chips
+//
+// Copyright (C) 2021-2022  Kevin O'Connor <kevin@koconnor.net>
+// Copyright (C) 2019 Eug Krashtan <eug.krashtan@gmail.com>
+// Copyright (C) 2020 Pontus Borg <glpontus@gmail.com>
+//
+// This file may be distributed under the terms of the GNU GPLv3 license.
+
+#include "command.h" // DECL_CONSTANT_STR
+#include "generic/armcm_boot.h" // armcm_enable_irq
+#include "generic/canbus.h" // canbus_notify_tx
+#include "generic/canserial.h" // CANBUS_ID_ADMIN
+#include "internal.h" // enable_pclock
+#include "sched.h" // DECL_INIT
+
+
+/****************************************************************
+ * Pin configuration
+ ****************************************************************/
+
+#if CONFIG_ATSAM_CANBUS_PB3_PB2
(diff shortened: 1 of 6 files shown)


src/atsam/fdcan.c
```diff
@@ -0,0 +1,309 @@
...
+void
+can_init(void)
+{
+#if !CONFIG_USBCANBUS
+    // Configure UPLL for USB and CAN
+    PMC->CKGR_UCKR = CKGR_UCKR_UPLLCOUNT(3) | CKGR_UCKR_UPLLEN;
+    while ( !(PMC->PMC_SR & PMC_SR_LOCKU) )
+    {
+    }
+#endif
```
KevinOConnor, line 243: It is preferable to avoid `#ifdefs` where possible.  It would be preferable to write this as:
```
    if (!CONFIG_USBCANBUS) {
        ...
    }
```

Also, please follow the syntax conventions elsewhere in this file and use:
```
    while (...)
        ;
```

src/atsam/main.c
```diff
@@ -14,6 +14,10 @@
 #define FREQ_PERIPH_DIV (CONFIG_MACH_SAME70 ? 2 : 1)
 #define FREQ_PERIPH (CONFIG_CLOCK_FREQ / FREQ_PERIPH_DIV)
 
+#if CONFIG_MACH_SAME70
+ #define FREQ_CAN 80000000
+#endif
+
```
KevinOConnor, line 20: I think it would be preferable to always make this define - something like:
```
#define FREQ_SAME70_CAN 80000000
```

src/atsam/main.c
```diff
@@ -62,6 +66,10 @@ enable_pclock(uint32_t id)
 uint32_t
 get_pclock_frequency(uint32_t id)
 {
+#if CONFIG_MACH_SAME70
+    if (id == MCAN0_CLOCK_ID || id == MCAN1_CLOCK_ID)
+        return FREQ_CAN;
+#endif
```
KevinOConnor, line 72: As before, it would be preferable to use a regular if instead of an ifdef.

KevinOConnor (commented):
Thanks.  Seems fine to me.  I have a few very minor comments - see below.

-Kevin

src/atsam/main.c
Wulfsta: This isn't feasible here, unless we also define the clock ids for other chips, which would be a bit silly. Going to keep this one.

KevinOConnor:
Thanks.

-Kevin

PR #6367, opened 2023-10-11
docs: In Configuration Checks doc, missing ! in pin inversion example in Verify endstops

thomasdunn:
In the "Verify endstops" section there is an example of adding an exclamation point to the pin definition to invert its logic.  I believe the intention in the example is to keep the hardware pull-up ^ and follow it with ! to invert the pin logic, but the ! is missing.


KevinOConnor:
Thanks.

-Kevin

PR #6368, opened 2023-10-12
Update generic-fysetc-cheetah-v2.0.cfg

iura-iura:


JamesH1978:
Thank you for submitting a PR, please refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and make sure you have provided some context to your change and you have added signed off by line.

Also you have stated /dev/ttyS2, what device is this on? These are generic config's and likely ttyS2 is a device specific to your hardware

Thanks

James

Sineos:
`/dev/ttyS2` is a connection over direct UART wiring and only very rarely used. I do not think this go into a cfg as default.

PR #6369, opened 2023-10-14
New pwm_tool module that supports mcu queing of pwm updates

KevinOConnor:
This is an alternative implementation for #4128.  The goal is to support fine grained control of spindles, lasers, and similar tools that are controlled by PWM input.  The current `output_pin` module requires that each pin update have a minimum duration of 100ms, which is often not sufficient for these tools.  This PR introduces a new `pwm_tool` module capable of queuing PWM updates in the micro-controller and is therefore capable of much higher update rates.

This differs from #4128 in the implementation.  I'm leery that some of the low-level changes in #4128 may introduce a regression.  This PR makes only a few changes outside of the new pwm_tool.py code.  That is, the implementation here is mostly isolated to the new module and is therefore unlikely to introduce a regression to any users not using this module.

Unfortunately, I do not have a good setup to test this module.  Feedback is appreciated.

@Cirromulus - FYI.

-Kevin

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3127,6 +3127,26 @@ pin:
 #   parameter.
 ```
 
+### [pwm_tool]
+
+Pulse width modulation digital output pins capable of high speed
+updates (one may define any number of sections with an "output_pin"
+prefix). Pins configured here will be setup as output pins and one may
+modify them at run-time using "SET_PIN PIN=my_pin VALUE=.1" type
+extended [g-code commands](G-Codes.md#output_pin).
+
+```
+[pwm_tool my_tool]
+pin:
+#   The pin to configure as an output. This parameter must be provided.
+#value:
+#shutdown_value:
+#cycle_time: 0.100
+#hardware_pwm: False
+#scale:
+#   See the "output_pin" section for the definition of these parameters.
+```
+
 ### [static_digital_output]
 
 Statically configured digital output pins (one may define any number
--- a/klippy/extras/pwm_tool.py
+++ b/klippy/extras/pwm_tool.py
@@ -1,0 +1,157 @@
+# Queued PWM gpio output
+#
+# Copyright (C) 2017-2023  Kevin O'Connor <kevin@koconnor.net>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import chelper
+
+PIN_MIN_TIME = 0.100
+MAX_SCHEDULE_TIME = 5.0
+
+class error(Exception):
+    pass
+
+class MCU_queued_pwm:
+    def __init__(self, pin_params):
+        self._mcu = pin_params['chip']
+        self._hardware_pwm = False
+        self._cycle_time = 0.100
+        self._max_duration = 2.
+        self._oid = self._mcu.create_oid()
+        ffi_main, ffi_lib = chelper.get_ffi()
+        self._stepqueue = ffi_main.gc(ffi_lib.stepcompress_alloc(self._oid),
+                                      ffi_lib.stepcompress_free)
+        self._mcu.register_stepqueue(self._stepqueue)
+        self._stepcompress_queue_mq_msg = ffi_lib.stepcompress_queue_mq_msg
+        self._mcu.register_config_callback(self._build_config)
+        self._pin = pin_params['pin']
+        self._invert = pin_params['invert']
+        self._start_value = self._shutdown_value = float(self._invert)
+        self._last_clock = self._cycle_ticks = 0
+        self._pwm_max = 0.
+        self._set_cmd_tag = None
+    def get_mcu(self):
+        return self._mcu
+    def setup_max_duration(self, max_duration):
+        self._max_duration = max_duration
+    def setup_cycle_time(self, cycle_time, hardware_pwm=False):
+        self._cycle_time = cycle_time
+        self._hardware_pwm = hardware_pwm
+    def setup_start_value(self, start_value, shutdown_value):
+        if self._invert:
+            start_value = 1. - start_value
+            shutdown_value = 1. - shutdown_value
+        self._start_value = max(0., min(1., start_value))
+        self._shutdown_value = max(0., min(1., shutdown_value))
+    def _build_config(self):
+        if self._max_duration and self._start_value != self._shutdown_value:
+            raise pins.error("Pin with max duration must have start"
+                             " value equal to shutdown value")
+        cmd_queue = self._mcu.alloc_command_queue()
+        curtime = self._mcu.get_printer().get_reactor().monotonic()
+        printtime = self._mcu.estimated_print_time(curtime)
+        self._last_clock = self._mcu.print_time_to_clock(printtime + 0.200)
+        cycle_ticks = self._mcu.seconds_to_clock(self._cycle_time)
+        mdur_ticks = self._mcu.seconds_to_clock(self._max_duration)
+        if mdur_ticks >= 1<<31:
+            raise pins.error("PWM pin max duration too large")
+        if self._hardware_pwm:
+            self._pwm_max = self._mcu.get_constant_float("PWM_MAX")
+            self._mcu.add_config_cmd(
+                "config_pwm_out oid=%d pin=%s cycle_ticks=%d value=%d"
+                " default_value=%d max_duration=%d"
+                % (self._oid, self._pin, cycle_ticks,
+                   self._start_value * self._pwm_max,
+                   self._shutdown_value * self._pwm_max, mdur_ticks))
+            svalue = int(self._start_value * self._pwm_max + 0.5)
+            self._mcu.add_config_cmd("queue_pwm_out oid=%d clock=%d value=%d"
+                                     % (self._oid, self._last_clock, svalue),
+                                     on_restart=True)
+            self._set_cmd_tag = self._mcu.lookup_command(
+                "queue_pwm_out oid=%c clock=%u value=%hu",
+                cq=cmd_queue).get_command_tag()
+            return
+        # Software PWM
+        if self._shutdown_value not in [0., 1.]:
+            raise pins.error("shutdown value must be 0.0 or 1.0 on soft pwm")
+        if cycle_ticks >= 1<<31:
+            raise pins.error("PWM pin cycle time too large")
+        self._mcu.add_config_cmd(
+            "config_digital_out oid=%d pin=%s value=%d"
+            " default_value=%d max_duration=%d"
+            % (self._oid, self._pin, self._start_value >= 1.0,
+               self._shutdown_value >= 0.5, mdur_ticks))
+        self._mcu.add_config_cmd(
+            "set_digital_out_pwm_cycle oid=%d cycle_ticks=%d"
+            % (self._oid, cycle_ticks))
+        self._cycle_ticks = cycle_ticks
+        svalue = int(self._start_value * cycle_ticks + 0.5)
+        self._mcu.add_config_cmd(
+            "queue_digital_out oid=%d clock=%d on_ticks=%d"
+            % (self._oid, self._last_clock, svalue), is_init=True)
+        self._set_cmd_tag = self._mcu.lookup_command(
+            "queue_digital_out oid=%c clock=%u on_ticks=%u",
+            cq=cmd_queue).get_command_tag()
+    def set_pwm(self, print_time, value):
+        clock = self._mcu.print_time_to_clock(print_time)
+        minclock = self._last_clock
+        self._last_clock = clock
+        if self._invert:
+            value = 1. - value
+        max_count = self._cycle_ticks
+        if self._hardware_pwm:
+            max_count = self._pwm_max
+        v = int(max(0., min(1., value)) * max_count + 0.5)
+        data = (self._set_cmd_tag, self._oid, clock & 0xffffffff, v)
+        ret = self._stepcompress_queue_mq_msg(self._stepqueue, clock,
+                                              data, len(data))
+        if ret:
+            raise error("Internal error in stepcompress")
+
+class PrinterOutputPin:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        ppins = self.printer.lookup_object('pins')
+        # Determine pin type
+        pin_params = ppins.lookup_pin(config.get('pin'), can_invert=True)
+        self.mcu_pin = MCU_queued_pwm(pin_params)
+        cycle_time = config.getfloat('cycle_time', 0.100, above=0.,
+                                     maxval=MAX_SCHEDULE_TIME)
+        hardware_pwm = config.getboolean('hardware_pwm', False)
+        self.mcu_pin.setup_cycle_time(cycle_time, hardware_pwm)
+        self.scale = config.getfloat('scale', 1., above=0.)
+        self.last_print_time = 0.
+        self.mcu_pin.setup_max_duration(0.)
+        # Determine start and shutdown values
+        self.last_value = config.getfloat(
+            'value', 0., minval=0., maxval=self.scale) / self.scale
+        self.shutdown_value = config.getfloat(
+            'shutdown_value', 0., minval=0., maxval=self.scale) / self.scale
+        self.mcu_pin.setup_start_value(self.last_value, self.shutdown_value)
+        # Register commands
+        pin_name = config.get_name().split()[1]
+        gcode = self.printer.lookup_object('gcode')
+        gcode.register_mux_command("SET_PIN", "PIN", pin_name,
+                                   self.cmd_SET_PIN,
+                                   desc=self.cmd_SET_PIN_help)
+    def get_status(self, eventtime):
+        return {'value': self.last_value}
+    def _set_pin(self, print_time, value):
+        if value == self.last_value:
+            return
+        print_time = max(print_time, self.last_print_time)
+        self.mcu_pin.set_pwm(print_time, value)
+        self.last_value = value
+        self.last_print_time = print_time
+    cmd_SET_PIN_help = "Set the value of an output pin"
+    def cmd_SET_PIN(self, gcmd):
+        # Read requested value
+        value = gcmd.get_float('VALUE', minval=0., maxval=self.scale)
+        value /= self.scale
+        # Obtain print_time and apply requested settings
+        toolhead = self.printer.lookup_object('toolhead')
+        toolhead.register_lookahead_callback(
+            lambda print_time: self._set_pin(print_time, value))
+
+def load_config_prefix(config):
+    return PrinterOutputPin(config)
--- a/test/klippy/pwm.test
+++ b/test/klippy/pwm.test
@@ -28,3 +28,11 @@ SET_PIN PIN=soft_pwm_pin VALUE=0.5 CYCLE_TIME=0.5
 SET_PIN PIN=soft_pwm_pin VALUE=0.5 CYCLE_TIME=0.5
 SET_PIN PIN=soft_pwm_pin VALUE=0.75 CYCLE_TIME=0.5
 SET_PIN PIN=soft_pwm_pin VALUE=0.75 CYCLE_TIME=0.75
+
+# PWM tool
+# Basic test
+SET_PIN PIN=test_pwm_tool VALUE=0
+SET_PIN PIN=test_pwm_tool VALUE=0.5
+SET_PIN PIN=test_pwm_tool VALUE=0.5
+SET_PIN PIN=test_pwm_tool VALUE=0.25
+SET_PIN PIN=test_pwm_tool VALUE=1
--- a/config/sample-pwm-tool.cfg
+++ b/config/sample-pwm-tool.cfg
@@ -2,9 +2,8 @@
 # such as a laser or spindle.
 # See docs/Using_PWM_Tools.md for a more detailed description.
 
-[output_pin TOOL]
+[pwm_tool TOOL]
 pin: !ar9       # use your fan's pin number
-pwm: True
 hardware_pwm: True
 cycle_time: 0.001
 shutdown_value: 0
@@ -36,9 +35,9 @@ gcode:
 
 [menu __main __control __toolonoff]
 type: input
-enable: {'output_pin TOOL' in printer}
+enable: {'pwm_tool TOOL' in printer}
 name: Fan: {'ON ' if menu.input else 'OFF'}
-input: {printer['output_pin TOOL'].value}
+input: {printer['pwm_tool TOOL'].value}
 input_min: 0
 input_max: 1
 input_step: 1
@@ -47,9 +46,9 @@ gcode:
 
 [menu __main __control __toolspeed]
 type: input
-enable: {'output_pin TOOL' in printer}
+enable: {'pwm_tool TOOL' in printer}
 name: Tool speed: {'%3d' % (menu.input*100)}%
-input: {printer['output_pin TOOL'].value}
+input: {printer['pwm_tool TOOL'].value}
 input_min: 0
 input_max: 1
 input_step: 0.01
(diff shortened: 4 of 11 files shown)


klippy/extras/pwm_tool.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(self, pin_params):
+        self._mcu = pin_params['chip']
+        self._hardware_pwm = False
+        self._cycle_time = 0.100
+        self._max_duration = 2.
+        self._oid = self._mcu.create_oid()
+        ffi_main, ffi_lib = chelper.get_ffi()
+        self._stepqueue = ffi_main.gc(ffi_lib.stepcompress_alloc(self._oid),
+                                      ffi_lib.stepcompress_free)
+        self._mcu.register_stepqueue(self._stepqueue)
```
Cirromulus, line 25: Minor cosmetic: It is not really a *step*-queue.

In object-oriented C++ I would have suggested a general buffer management queue (e.g. `command_queue`), which the stepcompress system would inherit the queue from and add the compression part (i.e. step-dir-filter etc.)

Perhaps I am thinking too general, but the reason for this is that this might be a possible location for a hypothetical *proportionality* feature (i.e. *S*-Parameter in [G0 in RepRap documentation](https://reprap.org/wiki/G-code#G-commands)).
After a stepcompress-flush, a toolhead might insert `pwm_tool` messages into the queue according to `actual speed /  requested speed` and a gamma-curve.

But we might "cross the bridge once we get there".

test/klippy/pwm.test
```diff
@@ -28,3 +28,11 @@ SET_PIN PIN=soft_pwm_pin VALUE=0.5 CYCLE_TIME=0.5
...
 SET_PIN PIN=soft_pwm_pin VALUE=0.75 CYCLE_TIME=0.5
 SET_PIN PIN=soft_pwm_pin VALUE=0.75 CYCLE_TIME=0.75
+
+# PWM tool
+# Basic test
+SET_PIN PIN=test_pwm_tool VALUE=0
+SET_PIN PIN=test_pwm_tool VALUE=0.5
+SET_PIN PIN=test_pwm_tool VALUE=0.5
+SET_PIN PIN=test_pwm_tool VALUE=0.25
+SET_PIN PIN=test_pwm_tool VALUE=1
```
Cirromulus, line 38: Other things worthy to test: 
 - `maximum_mcu_duration` by activating the pin to some non-default value and starting a toolhead move that will take longer than this duration.
 - small rasterized image with some speed and reasonable resolution to test a "real" world example (perhaps the *Klipper3D* logo)

If wanted, I can provide these tests as I used something similar (but manually).

klippy/toolhead.py
```diff
@@ -543,6 +543,14 @@ def register_lookahead_callback(self, callback):
         last_move.timing_callbacks.append(callback)
     def note_kinematic_activity(self, kin_time):
         self.last_kin_move_time = max(self.last_kin_move_time, kin_time)
+    def note_move_queue_activity(self, mq_time):
+        if self.special_queuing_state == "Flushed":
```
Cirromulus, line 547: During homing and force move, the `special_queueing_state` is set to "Drip".
While it is not recommended to have some `pwm-tool` in its non-default state during homing, it might still happen (e.g. queued GCODE).

If it is not flushed in this mode, it will still have an active PWM-Tool during homing, but will fail either because of `maximum_mcu_duration` or, after homing, due to the delayed schedule (`timer_too_close`).

At least, it happened to me with only `update_move_time`. I am not sure whether updating the `last_kin_move_time` in "Drip"-mode will lead to updates, because `_update_move_time(kin_time)` did not.

Cirromulus (commented):
Overall, I like this MR.
It is concise in the stepcompress system and adds the `pwm_tool` as a klipper-extra which is nice (although that adds some redundancy to `output_pin.py`).


Real-world tests are still outstanding, but might get to this by this or the following weekend.

docs/Config_Reference.md
```diff
@@ -3127,6 +3127,27 @@ pin:
...
+### [pwm_tool]
+
+Pulse width modulation digital output pins capable of high speed
+updates (one may define any number of sections with an "output_pin"
+prefix). Pins configured here will be setup as output pins and one may
+modify them at run-time using "SET_PIN PIN=my_pin VALUE=.1" type
+extended [g-code commands](G-Codes.md#output_pin).
+
+```
+[output_pin my_pin]
```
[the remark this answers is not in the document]
KevinOConnor: Thanks.  Fixed.

klippy/extras/pwm_tool.py
KevinOConnor: Earlier this year I spent some time reworking the trapq, stepcompress, and steppersync code.  It's difficult to get all that code correct however, so I haven't gotten it to the point that it is ready to merge.  Largely, the pwm_tool code is largely orthogonal, so likely best not to introduce additional dependencies to it.

KevinOConnor:
Thanks for the detailed review.

I made some changes and rebased this branch.

Upon further thought, I'm not confident that the `maximum_mcu_duration` feature is correct on this PR.  I have now separated out the code for it into a separate commit for review purposes, but I suspect it is not a good plan to add support for `maximum_mcu_duration` as it is currently implemented here.

The Klipper code attempts to plan out stepper movement two seconds in advance of when it is needed.  That helps ensure that the micro-controllers have the necessary movement information fully buffered so that it can maintain precise timing.  As part of this, the host prioritizes the filling of the move_queue and the serial bandwidth up to two seconds into the future.

Unfortunately, the `maximum_mcu_duration` code on this PR may attempt to send the "time extension" pwm update with only 100ms of time before it is needed on the mcu.  The problem is, it's possible that the main kinematic planner may have filled the move_queue and queued up messages for the upcoming two seconds, such that an appended pwm update wont arrive within 100ms.  This could lead to "timer too close" and "Missed scheduling of next digital out event" errors.

There is also a related problem that an appended pwm update with a deadline just 100ms in the future does not necessarily wake up the host flushing mechanism in time.  (This is likely the reason you saw errors when in "Drip" mode.)

I'm not sure what a good way to address this issue is.  Some possibilities:
1. We could commit the code without `maximum_mcu_duration`.  That could notably limit the utility of this code though.
2. Some kind of mechanism separate from queued PWM updates could be used to ensure the host and mcu do not lose contact.  The code for this could get pretty complicated though.
3. We could alter the mcu move_queue flushing mechanism (`mcu.py:flush_moves()`) to check if a PWM time extension update is needed, and append that update just prior to flushing.  This would effectively allow the PWM updates to stay in sync with the two seconds of kinematic pre-planning.  However, some additional system would still be needed to ensure that the mcu continues to flush if a pwm_tool remains active while the main kinematic code is idle.

I'm not sure.

-Kevin

Cirromulus:
> 1.  We could commit the code without maximum_mcu_duration. That could notably limit the utility of this code though.  
> 2. Some kind of mechanism separate from queued PWM updates could be used to ensure the host and mcu do not lose contact. The code for this could get pretty complicated though.
> 3. We could alter the mcu move_queue flushing mechanism (mcu.py:flush_moves()) to check if a PWM time extension update is needed, and append that update just prior to flushing. This would effectively allow the PWM updates to stay in sync with the two seconds of kinematic pre-planning. However, some additional system would still be needed to ensure that the mcu continues to flush if a pwm_tool remains active while the main kinematic code is idle.


Currently, number 2 is most appealing to me.
I think of a keepalive / timeout system, where the shortest `maximum_mcu_duration` is taken in where the toolhead (or other appropriate location) keeps track of the time the last message was sent, and inserts a keepalive / null message upon a nearing deadline.
The MCU on the other hand also tracks the time since the last received message and goes into failsafe when the shortest `maximum_mcu_duration` was reached. This way, for x PWM pins, the serial link does not need to be used for x updates per `maximum_mcu_duration` for redundant information as it currently would be.

This, however, holds only if Klippy's fail-mode is *fail silent* (which I believe it is).

>There is also a related problem that an appended pwm update with a deadline just 100ms in the future does not necessarily wake up the host flushing mechanism in time. (This is likely the reason you saw errors when in "Drip" mode.)

This is probably the reason for it. I had a hunch, and thought of making a version of your suggestion Nr. 3, but I did not finish this idea because it felt hacky and added a lot of special-case code.

KevinOConnor:
FYI, I rebased this branch again.

> Currently, number 2 is most appealing to me.

It's possible, but the issue with that approach is that it likely requires mcu code changes.  That's a pain to deploy and may risk regressions.

A user could probably "hack" this today by declaring a dummy pin (or led or something innocuous) as an output_pin with its maximum_mcu_duration.  Then as long as the user commanded that pin on during print start and off during print end, then they should get general host/mcu disconnect protection.

In general, I'd probably prefer approach 3.  It's more work and likely requires some notable refactoring, but seems like less of a hack.

In any case, I suspect only option 1 is practical if we want to merge prior to the next Klipper release.  https://klipper.discourse.group/t/upcoming-v0-12-0-release/10948

> This, however, holds only if Klippy's fail-mode is fail silent 

I'm not sure what you mean by "fail silent".

-Kevin

Cirromulus:
> It's possible, but the issue with that approach is that it likely requires mcu code changes. That's a pain to deploy and may risk regressions.

This is correct. With this big of a user basis, I see the pain of deployment. I still find a timeout / keepalive system more elegant than having a separate "non-move" queue with a higher priority, but this discussion is probably more suited in the discourse development channel.
Until then, I agree with option 1 for this MR. My fork also did not support it correctly, so we would not loose anything.

>A user could probably "hack" this today by declaring a dummy pin (or led or something innocuous) as an output_pin with its maximum_mcu_duration. Then as long as the user commanded that pin on during print start and off during print end, then they should get general host/mcu disconnect protection.

As my [Howto](https://github.com/Klipper3d/klipper/compare/master...Cirromulus:klipper:pwm_sync_channel#diff-cab11ae01ea302890cda2d6a50971e1eca46c6195b9a9163380e891e4129f9cfR30) suggests a separate enable-pin for power supply to the Laser / Tool, this would still fit well. The non-high-throughput pin can have the safety timeout, so that the PWM-pin does not need it. I suggest adding this to the `sample-pwm-tool.cfg` and the `docs/Using_PWM_Tools.md`.

> I'm not sure what you mean by "fail silent".
Fail silent means that if something fails in klipper host, it does not continue to send (possibly bogus) messages to the MCU, but does not send anything at all.
A classic example is a live-lock, where the actual code is stuck but some other thread still sends (old) data.

KevinOConnor:
Okay, thanks.  I've rebased this branch to just include the basic `[pwmtool]` module without the `maximum_mcu_duration` support.  (The previous `maximum_mcu_duration`, known to not work fully correct, code is up at https://github.com/KevinOConnor/klipper-dev/tree/work-pwmtool-20231020 ).

It would help if others with actual pwm controlled tools can confirm that the PR here works on that hardware and provides some utility.

-Kevin

Cirromulus:
I will test it with my laser setup tomorrow.

I still suggest adapting the example configuration to include the (regular) power supply pin:

```patch
diff --git a/config/sample-pwm-tool.cfg b/config/sample-pwm-tool.cfg
index 48986061..e4cbd45e 100644
--- a/config/sample-pwm-tool.cfg
+++ b/config/sample-pwm-tool.cfg
@@ -3,46 +3,46 @@
 # See docs/Using_PWM_Tools.md for a more detailed description.
 
 [pwm_tool TOOL]
-pin: !ar9       # use your fan's pin number
+pin: !ar9       # use your pin number connected to PWM input
 hardware_pwm: True
 cycle_time: 0.001
 shutdown_value: 0
-maximum_mcu_duration: 5
-# Default: 0 (disabled)
-# Amount of time in which the host has to acknowledge
-# a non-shutdown output value.
-# Suggested value is around 5 seconds.
-# Use a value that does not burn up your stock.
-# Please note that during homing, your tool
-# needs to be in default speed.
+
+[output_pin TOOL_ENABLE]
+pin: !ar10      # use the pin number connected to power supply of Tool
+shutdown_value: 0
+maximum_mcu_duration: 2
 
 [gcode_macro M3]
 gcode:
+    SET_PIN PIN=TOOL_ENABLE VALUE=1
     {% set S = params.S|default(0.0)|float %}
     SET_PIN PIN=TOOL VALUE={S / 255.0}
 
 [gcode_macro M4]
 gcode:
+    SET_PIN PIN=TOOL_ENABLE VALUE=1
     {% set S = params.S|default(0.0)|float %}
     SET_PIN PIN=TOOL VALUE={S / 255.0}
 
 [gcode_macro M5]
 gcode:
     SET_PIN PIN=TOOL VALUE=0
+    SET_PIN PIN=TOOL_ENABLE VALUE=0
 
 
 # Optional: LCD Menu Control
 
 [menu __main __control __toolonoff]
 type: input
-enable: {'pwm_tool TOOL' in printer}
-name: Fan: {'ON ' if menu.input else 'OFF'}
+enable: {'output_pin TOOL_ENABLE' in printer}
+name: Tool: {'ON ' if menu.input else 'OFF'}
input: {printer['output_pin TOOL_ENABLE'].value}
 input_min: 0
 input_max: 1
 input_step: 1
 gcode:
-    M3 S{255 if menu.input else 0}
+    SET_PIN PIN=TOOL_ENABLE VALUE={1 if menu.input else 0}
 
 [menu __main __control __toolspeed]
 type: input
diff --git a/docs/Using_PWM_Tools.md b/docs/Using_PWM_Tools.md
index 108ae37a..98e9f91c 100644
--- a/docs/Using_PWM_Tools.md
+++ b/docs/Using_PWM_Tools.md
@@ -21,12 +21,14 @@ _fully on_ for the time it takes the MCU to start up again.
 For good measure, it is recommended to _always_ wear appropriate
 laser-goggles of the right wavelength if the laser is powered;
 and to disconnect the laser when it is not needed.
-Also, you should configure a safety timeout,
+An additional safety layer would be powering your tool through the hotend-heater pin.
+In the configuration, the additional `[output_pin]` would be activated by `M3/M4` and disabled by `M5`.
+In this pin you should configure a safety timeout,
 so that when your host or MCU encounters an error, the tool will stop.
 
 For an example configuration, see [config/sample-pwm-tool.cfg](/config/sample-pwm-tool.cfg).
 
-## Commands
+## Example Commands
 
 `M3/M4 S<value>` : Set PWM duty-cycle. Values between 0 and 255.
 `M5` : Stop PWM output to shutdown value.
@@ -36,26 +38,29 @@ For an example configuration, see [config/sample-pwm-tool.cfg](/config/sample-pw
 If you use Laserweb, a working configuration would be:
 
     GCODE START:
-        M5            ; Disable Laser
         G21           ; Set units to mm
         G90           ; Absolute positioning
         G0 Z0 F7000   ; Set Non-Cutting speed
 
     GCODE END:
         M5            ; Disable Laser
-        G91           ; relative
-        G0 Z+20 F4000 ;
-        G90           ; absolute
 
     GCODE HOMING:
         M5            ; Disable Laser
         G28           ; Home all axis
 
     TOOL ON:
-        M3 $INTENSITY
+        M3
 
     TOOL OFF:
-        M5            ; Disable Laser
+        M3 S0         ; No output, but powered
 
     LASER INTENSITY:
-        S
+        M3 S
+
+    PWM MIN S VALUE:
+        0
+
+    PWM MAX S VALUE:
+        255
```

tregnier:
Hi @Cirromulus and @KevinOConnor,

First of all thanks for your work on that upgrade!

I've been trying to use it with my setup, but I'm having some issue _**Internal error on command:"G1"**_ when trying to run Gcode from laserburn or snapmaker
I feel it's getting too many command queued, but I really have no knowledge to tell.

Below 2 klippy logs taken right after the crash happens, if it helps to review the feature.

[klippy(5).log](https://github.com/Klipper3d/klipper/files/13259169/klippy.5.log)
[klippy(4).log](https://github.com/Klipper3d/klipper/files/13259171/klippy.4.log)

I have a laser and can try to run some test if you want, to review the feature (not only related to my issue) so it can eventually be integrated with the main branch.

I can't get the laser working right now, but the printing setup works fine. I've created 2 config file, printer.cfg and laser.cfg, so I deactivate the laser config when printing normally.

Cirromulus:
The relevant portion of your log is the following:

```
  File "/home/biqu/klipper/klippy/extras/virtual_sdcard.py", line 264, in work_handler
    self.gcode.run_script(line)
  File "/home/biqu/klipper/klippy/gcode.py", line 216, in run_script
    self._process_commands(script.split('\n'), need_ack=False)
  File "/home/biqu/klipper/klippy/gcode.py", line 198, in _process_commands
    handler(gcmd)
  File "/home/biqu/klipper/klippy/extras/gcode_move.py", line 143, in cmd_G1
    self.move_with_transform(self.last_position, self.speed)
  File "/home/biqu/klipper/klippy/extras/bed_mesh.py", line 209, in move
    self.toolhead.move([x, y, z + self.fade_target, e], speed)
  File "/home/biqu/klipper/klippy/toolhead.py", line 427, in move
    self.move_queue.add_move(move)
  File "/home/biqu/klipper/klippy/toolhead.py", line 187, in add_move
    self.flush(lazy=True)
  File "/home/biqu/klipper/klippy/toolhead.py", line 176, in flush
    self.toolhead._process_moves(queue[:flush_count])
  File "/home/biqu/klipper/klippy/toolhead.py", line 331, in _process_moves
    cb(next_move_time)
  File "/home/biqu/klipper/klippy/extras/pwm_tool.py", line 154, in <lambda>
    lambda print_time: self._set_pin(print_time, value))
  File "/home/biqu/klipper/klippy/extras/pwm_tool.py", line 143, in _set_pin
    self.mcu_pin.set_pwm(print_time, value)
  File "/home/biqu/klipper/klippy/extras/pwm_tool.py", line 106, in set_pwm
    ret = self._stepcompress_queue_mq_msg(self._stepqueue, clock,
OverflowError: integer 29 622 610 503 does not fit 'uint32_t'
```
This does not seem to be the clock value, but the "setvalue" `v` should be clamped to 0-1. 
Does this also happen when using *software* PWM?

tregnier:
Just tried with software PWM (set hardware_pwm = False), and I got the same issue

I'm using Biqu CB1 instead of Rpi by the way, not sure it could have an effect?

KevinOConnor:
Thanks.  That looks like a bug in the new pwm_tool.py code.  I have updated the code and rebased this branch.  Hopefully this issue is now fixed.

-Kevin

Cirromulus:
I have successfully tested the newest version (b2a9308c7f78a4f0aa37382888cf0f92fc66853a) to work with my GT2560 Mainboard.
On higher workloads, my `maximum_mcu_duration` target of 3 seconds can't be always met in the (laser) power pin, but I do have a very weak MCU and CPU (orange pi zero). I saw a slight behavior in the fork, as it usually tended to fail with `timer too close` first.
But it does work well in all (my) scenarios that already worked in the fork.

tregnier:
looks like it  works for me as well with the new code
I have issue to have proper engraving, but this is due to my laser, I need to order a new one.
No more error at least

Thanks!

KevinOConnor:
Thanks.  I committed this branch.  @Cirromulus - if you'd like a different config example, feel free to send it along.

I'd like to try to get to better "trapq flushing" in the next month or so, which should hopefully allow us to reintroduce the "maximum duration" check.

-Kevin

A-Panic:
Hi everyone, I'm trying to set up a Melzi 1.1.4 board (8bit atmega1284p) to take advantage of this new feature.
At the moment the laser is not yet plugged to the mainboard, I'm using a pwm fan as a test device.
Upon manually sending `M3 S255` nothing seems to happen, then after sending another M3/M4/M5 command I get an `MCU 'mcu' shutdown: Timer too close`. Host side I'm using a BTT Pi v1.2 running only klipper, the error suggests that the problem could be on the host side, but as far as I can tell everything seems ok. I used these same host, cable and PSU with my printer for a while with no problem.
I'm attaching a klippy log taken after the error

[klippy(4).log](https://github.com/Klipper3d/klipper/files/13404797/klippy.4.log)

I don't really have the knowledge to understand where the problem is, so if there's need for any other information or test I will try to help as soon as possible.

Thank you for this awesome firmware and for this feature

willpuckett:
This is happening a good bit with my laser as well when sending commands manually, but it works great in the context of a fully printed gcode file. I'm not sure if the console commands are queued differently or this is an expected part of the max duration issue mentioned in the previous post…

A-Panic:
@willpuckett thank you for your feedback, later today I'll try to run a job instead of manual commands and see what happens

A-Panic:
Hi everyone,

As @willpuckett suggested (thank you!) running a gcode file do not give any error also for me.
Anyway I switched from the fan to the actual laser before running these tests and still there is no actual change in the output: while the file is "printing" the toolhead moves, but the laser stays off.
At this point the problem may be in the physical connection or in a bad/lacking configuration in `[pwm_tool TOOL]`
- The laser is connected to the part cooling fan pin (PB4), which shloud be capable of hardware-PWM (`hardware_pwm` set to true or false does not change the result);
- I'm using Fluidd and `[virtual_sdcard]` is enabled by default (and as far as I understand, mandatory). Does this interfere in the command queue process? Again, I'm just giving more infos and context as I'm not sure how this feature actually works;
- This is the connector on my laser module:
immagine
Do I need to set `cycle_time` or `scale` to a specific value to match that 1kHz on the PWM pin?

Sorry for all the questions and thank you in advance

KevinOConnor:
> Upon manually sending M3 S255 nothing seems to happen, then after sending another M3/M4/M5 command I get an MCU 'mcu' shutdown: Timer too close.

This is a defect in the new pwm_tool code.  Issuing a command when the printer is otherwise idle does not flush out updates.  (In contrast, if the printer is in an "active" state, it should flush the updates.)   This is similar to the problem identified earlier with maximum_mcu_duration.  Unfortunately, there does not seem to be a simple fix to this issue.

I'll have to think about a solution to this issue.

-Kevin

firatsezer:
MKS Robin Nano Mini v3 is not working with Master, before i used the repo from Cirromulus and the same config has not worked.
i adapted it a little bit to this but nothing work. the laser doesnt appear.

currently pwm.cfg, commented parts not included:

[pwm_tool LASER]
pin: PB0
hardware_pwm: True
shutdown_value: 0




with Cirromulus, i declared it as output_pin and everything has worked as expected.

I dont need enable the tool it get power from the mainline of print head. i have a TevoUp Hydra 2 in 1

Cirromulus:
> This is a defect in the new pwm_tool code. Issuing a command when the printer is otherwise idle does not flush out updates. (In contrast, if the printer is in an "active" state, it should flush the updates.) This is similar to the problem identified earlier with maximum_mcu_duration. Unfortunately, there does not seem to be a simple fix to this issue.
> 
> I'll have to think about a solution to this issue.
> 

In my branch, I check for this case (idle) based on the `kin_time` and flush it with ` self.note_kinematic_activity(time)`:
https://github.com/Klipper3d/klipper/pull/4128/files#diff-48322802b4775fb953d78d2928004fc887f50efead57b6d599e3045a05a90b45R557-R569

A-Panic:
Hi everyone,

Today I managed to make my laser module work. It turns out that the PWM input my module need is at 5V instead of 12V, so after setting a different pin the laser started to work.

Thanks to you all, if you need to run some tests count me in

KevinOConnor:
> In my branch, I check for this case (idle) based on the kin_time and flush it with  self.note_kinematic_activity(time):

Alas, calling `self.note_kinematic_activity()` is also insufficient to flush updates.  In addition to setting `self.last_kin_move_time` the toolhead would also need to make sure the flush timer is active (`self.flush_timer`).

-Kevin

KevinOConnor:
I believe commit #6410 should fix the "Timer too close" errors reported here.

-Kevin

KevinOConnor:
FYI, I have created a new development branch at https://github.com/KevinOConnor/klipper-dev/tree/work-pwmflush-20231130 with initial support for `maximum_mcu_duration` with pwm_tool.

That development branch is based on the code from #6410.  It's tricky covering all the "corner cases" during flushing.  However, I'm hopeful the above development branch is correct.

If all goes well with #6410 then I'll create a PR for the above development branch.

-Kevin

PR #6370, opened 2023-10-15
Add support for BMP180 sensor

vaxxi:
Add support for the older BMP180 sensor into the BMxx80 sensor category. Sensors supports oversampling for pressure only. No additional changes required to Mainsail, as it integrates into the existing display of temperature and humidity.

image

### DIFF
--- a/klippy/extras/bme280.py
+++ b/klippy/extras/bme280.py
@@ -400,7 +400,8 @@ class BME280:
         self.write_register('CTRL_MEAS', meas)
 
         try:
-            time.sleep(0.005)
+            for i in range(94000):
+                pass
             data = self.read_register('REG_MSB', 3)
             UP = ((data[0] << 16)|(data[1] << 8)|data[2]) >> (8 - self.os_pres)
         except Exception:
--- a/klippy/extras/bme280.py
+++ b/klippy/extras/bme280.py
@@ -400,7 +400,7 @@ class BME280:
         try:
             self.reactor.pause(self.reactor.monotonic() + .01)
             data = self.read_register('REG_MSB', 3)
-            pressure_raw = \ 
+            pressure_raw = \
                 ((data[0] << 16)|(data[1] << 8)|data[2]) >> (8 - self.os_pres)
         except Exception:
             logging.exception("BMP180: Error reading pressure")
--- a/klippy/extras/bme280.py
+++ b/klippy/extras/bme280.py
@@ -388,7 +388,7 @@ class BME280:
         try:
             self.reactor.pause(self.reactor.monotonic() + .01)
             data = self.read_register('REG_MSB', 2)
-            UT = (data[0] << 8) | data[1]
+            temp_raw = (data[0] << 8) | data[1]
         except Exception:
             logging.exception("BMP180: Error reading temperature")
             self.temp = self.pressure = .0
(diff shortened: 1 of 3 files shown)


JamesH1978:
Thank you for submitting a PR, please refer to https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and make sure you have added the signed off by line.

Thanks

James

vaxxi:
Done, hope everything's correct now.

KevinOConnor:
Thanks.  I took a quick look at this PR, and I noticed a couple of issues:
1. It's not valid to call `sleep()` in the host code - there is a little bit of info on this at https://www.klipper3d.org/Code_Overview.html#adding-a-host-module .
2. This PR amends the copyright block at the top of the file with an ambiguous statement.  For code attribution, use commit messages, to claim a copyright then explicitly state that there is a copyright.

I'm not an expert on the bme code (Arksine can likely review that).  I don't think this will be something we will look to merge prior to the upcoming Klipper release ( https://klipper.discourse.group/t/upcoming-v0-12-0-release/10948 )

Cheers,
-Kevin

vaxxi:
Hi,

Agreed on the `sleep()`, just that I've seen it used in `mpu9250` as well:

```
mpu9250.py:        time.sleep(20. / 1000) # wait for accelerometer chip wake up
```

and figured it's accepted for non-critical routines. Anyway, I've replaced it with an alternative (the sensor doesn't have any other method to check the measurement is complete other than waiting for the specified period of time).

Ammended the header note as well, no copyright claim (just wanted a way for people to get in touch if needed).

Thank you!

klippy/extras/bme280.py
```diff
@@ -8,6 +8,8 @@
 
 REPORT_TIME = .8
 BME280_CHIP_ADDR = 0x76
+BMP180_CHIP_ADDR = 0x77
```
Arksine, line 11: This constant doesn't appear to be used anywhere in the module.

Arksine:
Outside of the comments above the implementation looks ok to me.  I addition, several variables are declared in uppercase (`UT`, `UP`, `X1`, etc).  The general rule in Python is that identifiers declared in uppercase are constants.  Code style isn't typically something we would comment on, however I thought I would point this out as this specific rule makes it easier for others to read and review code.

klippy/extras/bme280.py
```diff
@@ -334,6 +382,44 @@ def data_ready(stat):
         self._callback(self.mcu.estimated_print_time(measured_time), self.temp)
         return measured_time + REPORT_TIME
 
+    def _sample_bmp180(self, eventtime):
+        meas = self.chip_registers['CRV_TEMP']
+        self.write_register('CTRL_MEAS', meas)
+
+        try:
+            for i in range(94000):
+                pass
```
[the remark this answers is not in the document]
vaxxi: Ah, that's perfect, thank you.

klippy/extras/bme280.py
vaxxi: Indeed, it was mostly a note for documenting. Removed.

klippy/extras/bme280.py
```diff
@@ -443,6 +526,37 @@ def _calculate_gas_heater_duration(self, duration_ms):
...
+        dig = self.dig
+        b5 = self.t_fine
+        b6 = b5 - 4000
+        x1 = (dig['B2'] * (b6 * b6 / 4096)) / 2048
+        x2 = dig['AC2'] * b6 / 2048
+        x3 = x1 + x2
+        b3 = ((int(dig['AC1'] * 4 + x3) << self.os_pres) + 2) / 4
+        x1 = dig['AC3'] * b6 / 8192
+        x2 = (dig['B1'] * (b6 * b6 / 4096)) / 65536
+        X3 = ((x1 + x2) + 2) / 4
```
Arksine, line 547: It appears that `X3` here should be changed to `x3`

Arksine:
I had a couple more minor observations.  When those are addressed I believe this PR will be ready.

vaxxi:
All addressed, thank you.

Arksine:
Changes look good to me.  Thanks.

KevinOConnor:
Thanks.  Out of an abundance of caution, I think we should look to merge this after the upcoming v0.12.0 release (in a couple of weeks).

-Kevin

PR #6372, opened 2023-10-21
toolhead: Fix input shaper crash

viesturz:
Fixes issues with https://github.com/Klipper3d/klipper/pull/6303
That is using the input shaper as the window owner, but there may be a separate window for each axis, so changing that to use the stepper as the window owner.


olivierberthomme:
For me you fixed it. Got multiple random shutdown with #6303 (at initial movement or bed probing) and with #6372 could print 2 parts without issue.
I'm with TMC2208 and 2 different input shapers for X and Y.

PR #6373, opened 2023-10-22
Update generic-printrboard.cfg

Mercalyn:
added useful notes to to combat sparse information

### DIFF
--- a/config/generic-printrboard.cfg
+++ b/config/generic-printrboard.cfg
@@ -1,6 +1,18 @@
 # This file contains common pin mappings for Printrboard boards (rev B
-# through D). To use this config the firmware should be compiled for
-# the AVR at90usb1286.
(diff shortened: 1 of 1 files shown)


Sineos:
Thanks for the proposal. In fact, we are currently evaluating means to make this information more clear in the config examples. The goal is:
* Have it more structured
* Avoid redundant information, e.g. not mentioning settings that are default anyway
* Avoid duplication of instructions as it brings a lot of maintenance burden should anything change

So, likely we will defer such PRs until we have a conclusion there.

In general, please refer to [master/docs/CONTRIBUTING.md](https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md?rgh-link-date=2023-10-15T11%3A42%3A55Z) and make sure you have added the signed off by line.

PR #6374, opened 2023-10-22
hall_filament_width_sensor: max filament diameter

ressu:
Add support for maximum filament diameter to hall filament width sensor. If the diameter of the filament diameter is larger than the limit, the virtual runout sensor will trigger. The default value is set to maximum flow adjustment threshold to prevent oversized filament from clogging.

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,12 @@ All dates in this document are approximate.
 
 ## Changes
 
+20231216: The `[hall_filament_width_sensor]` is changed to trigger filament runout
+when the thickness of the filament exceeds `max_diameter`. The maximum diameter
+defaults to `default_nominal_filament_diameter + max_difference`. See
+[[hall_filament_width_sensor] configuration
+reference](./Config_Reference.md#hall_filament_width_sensor) for more details.
+
 20231207: Several undocumented config parameters in the `[printer]`
 config section have been removed (the buffer_time_low,
 buffer_time_high, buffer_time_start, and move_flush_time parameters).
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4428,6 +4428,9 @@ adc2:
 #   command.
 #min_diameter: 1.0
 #   Minimal diameter for trigger virtual filament_switch_sensor.
+#max_diameter:
+#   Maximum diameter for triggering virtual filament_switch_sensor.
+#   The default is default_nominal_filament_diameter + max_difference.
 #use_current_dia_while_delay: False
 #   Use the current diameter instead of the nominal diameter while
 #   the measurement delay has not run through.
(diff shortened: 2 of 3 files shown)


dewi-ny-je:
@KevinOConnor  this is a straightforward to check and as far as I can tell it's done correctly.
We can of course discuss about the difference between "filament present/runout" and "min/max diameter" for performing the correction, but as long as the current overall design of the module is maintained, the changes proposed seem ok to me.

The feature is quite useful, since sometimes too wide filaments (blobs, for example with joined filaments or home made filaments) can get stuck in the reverse-Bowden tube,  or in the Bowden between extruder and hot end (for example EVA toolheads use this configuration).

ressu:
Just to clarify, this won't catch short blobs due to the way how measurement is done. The blob needs to extend at least over 10mm (or what ever is the configured measurement distance) and generally needs a bit longer than that. But it will detect out of spec filament reliably.

I agree that the internal variable naming should change a bit, but it I wanted to keep the changes to minimum in the PR and did my best to find suitable names for the variables within the constraints we had.

I've been running this locally for a few weeks now.

dewi-ny-je:
Then it's much less interesting, but it still makes sense to some degree.

It would be even nicer to use the Hall filament sensor module for both blobs and average compensation, for example by polling diameter every 1 mm just for filament run out/blob detection, and every 10 mm for diameter compensation.

Once I start working on the module to find a better way to apply the compensation (mid-2024 maybe?) I'll look into it. Ideally, I would prefer the compensation not to translate into flow% corrections, since those should be a user-facing correction which should not jump around. But it's not clear what to use instead.

KevinOConnor:
I don't know enough to really review this.  However, it seems fine to me.  If you confirm this PR is still ready to merge then I'll look to commit.

Cheers,
-Kevin

ressu:
Yup, ready to merge. I synced the branch to master just now to verify that it still cleanly applies.

KevinOConnor:
Thanks.  I just realized that this is a change in behavior.  So, I think we should document that change in docs/Config_Changes.md (or alternatively change the default to be high enough that it would not cause a runout trigger).

-Kevin

ressu:
Apologies for the late update. I went back and forth with setting a high default vs. sticking with the default and documenting it.

In the end I came to the conclusion that when you exceed the maximum adjustment, the behaviour is going to be unwanted and the whole purpose of the diameter sensor is to avoid those artifacts. So for (what I think are) the majority of users that use this feature having the print pause automatically when the automatic adjustment range is exceed is the desired behaviour. So instead of asking everyone to update their configuration individually, it makes sense to have those that prefer print quality issues over print pausing change their confiuration instead.

KevinOConnor:
Thanks.  Looks fine, but I need a signed-off-by line here or in the commit.  See https://www.klipper3d.org/CONTRIBUTING.html .

-Kevin

ressu:
Whoops, missed that when I rebased the changes. Fixed now.

KevinOConnor:
Thanks.

-Kevin

PR #6375, opened 2023-10-25
config: Allow config includes to use subfolder globs

bdistin:
This has been a minor change I have been using for a while to keep my config well organized/modular.
```
[include hardware/**/*.cfg]
[include macros/**/*.cfg]
```
You can see my use of it in action here: https://github.com/bdistin/v2.3231_config/


klippy/configfile.py
```diff
@@ -223,7 +223,7 @@ def _resolve_include(self, source_filename, include_spec, fileconfig,
         dirname = os.path.dirname(source_filename)
         include_spec = include_spec.strip()
         include_glob = os.path.join(dirname, include_spec)
-        include_filenames = glob.glob(include_glob)
+        include_filenames = glob.glob(include_glob, recursive=True)
```
Arksine, line 226: The `recursive` argument for `glob.glob()` was added in Python 3.5.  AFAIK Klipper still supports running on Python 2.7, so this would be a breaking change.

Arksine:
I added a review comment.  It would be possible check the interpreter version and enable recursive blobs for Python >= 3.5.

bdistin:
> I added a review comment to take care of the bot. It would be possible check the interpreter version and enable recursive blobs for Python >= 3.5. The final decision to merge would be up to Kevin.

Oh, apologies. I think I saw Kevin do a similar version check in another PR so I think I can update the PR to check the version. Thanks for the review. Didn't even think to check dep requirements.

Arksine:
Something like the following should work:

```python
if sys.version_info >= (3, 5):
    include_filenames = glob.glob(include_glob, recursive=True)
else:
    include_filenames = glob.glob(include_glob)
```

bdistin:
> Something like the following should work:
> 
> ```python
> if sys.version_info >= (3, 5):
>     include_filenames = glob.glob(include_glob, recursive=True)
> else:
>     include_filenames = glob.glob(include_glob)
> ```
> 
> In addition, we probably need to tweak the documentation to indicate that recursive globs only work on Python 3.

Apologies for the noob question. (python isn't really one of my languages) But in your example, you don't hoist include_filenames, but I see other places in the code where variables are hoisted. Does Python (or only some versions of Python) have block-level scoping? Or is my hoist unnecessary code?

PR #6378, opened 2023-10-28
Filament Sensors: Add switch_pin_state status variable

Gunner087:
Added a switch_pin_state status variable to the RunoutHelper class. Variable state is modified by _button_handler() in SwitchSensor and encoder_event() in EncoderSensor.

This allows a macro to read the logic state of the switch_pin for filament_switch_sensor and filament_motion_sensor. In my case, I wanted this functionality in order to write a macro that tracks the motion of the filament during a purge in the START_PRINT macro. That way I can check if the extruder is working properly before the print starts. Checking filament extrusion during filament load or unload macros is also possible.

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -152,7 +152,6 @@ objects:
 - `switch_pin_state`: Returns True/False if the switch_pin digital
   input is logic 1/0
 
-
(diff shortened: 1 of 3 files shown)


Gunner087:
Closed, found a better solution

PR #6380, opened 2023-10-30
config: Update generic-bigtreetech-octopus.cfg for Octopus PRO 1.1 pins

bigtreetech:


KevinOConnor:
Thanks.  I think this should be addressed now with #6382.

-Kevin

PR #6381, opened 2023-10-30
tmc5160: The maximum current of tmc5160 is determined by the sensor resistor

bigtreetech:

Some independent TMC5160 modules use high-efficiency MOSFETs and active heat sinks, so the current can be very high. The maximum current of TMC5160 is determined by the sampling resistance rather than a fixed 3A now.

### DIFF
--- a/klippy/extras/tmc5160.py
+++ b/klippy/extras/tmc5160.py
@@ -257,7 +257,6 @@ FieldFormatters.update({
 ######################################################################
 
 VREF = 0.325
-MAX_CURRENT = 3.000
 
 class TMC5160CurrentHelper:
     def __init__(self, config, mcu_tmc):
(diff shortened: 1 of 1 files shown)


thijstriemstra:
I'm about to install some TMC5160 RGB and TMC5160 Pro v1.1 drivers and read only yesterday about a hard-coded current value in klipper code. Glad to see it's getting fixed!

KevinOConnor:
Thanks.  This was also raised in #6093.  As with the comments in that PR, I'm concerned that the error check could become confusing to users.  The main reason for the error check was to avoid some users entering in a milliamps (eg, 2000) where amps were desired (eg, 2).  If 3 amps is not sufficient, can we just increase the maximum to 5 or 10 amps on tmc5160?

-Kevin

bigtreetech:
> If 3 amps is not sufficient, can we just increase the maximum to 5 or 10 amps on tmc5160?


It is definitely sufficient for stepper motors if it can be changed to 10A.

klippy/extras/tmc5160.py
```diff
@@ -257,24 +257,26 @@
...
         self.fields = mcu_tmc.get_fields()
+        self.sense_resistor = config.getfloat('sense_resistor', 0.075, above=0.)
+        max_current = self._calc_max_current()
         run_current = config.getfloat('run_current',
-                                      above=0., maxval=MAX_CURRENT)
-        hold_current = config.getfloat('hold_current', MAX_CURRENT,
-                                       above=0., maxval=MAX_CURRENT)
+                                      above=0., maxval=max_current)
+        hold_current = config.getfloat('hold_current', max_current,
+                                       above=0., maxval=max_current)
```
HumbleDeer, line 272: @bigtreetech 

May I ask out of curiosity: 
What is the purpose of changing `MAX_CURRENT` to `max_current`? 

Is it merely because MAX_CURRENT was a defined directive, compared to the lowercase one being a defined variable in the config, under `__INIT__`? As seen on line 268

bigtreetech:
close since https://github.com/Klipper3d/klipper/commit/bb4711c5d31e8159945f945c662e6668059a174f

PR #6383, opened 2023-10-30
Update the Contact document

KevinOConnor:
Try to make it more clear that Discord is not Discourse, as the two similarly sounding services are often confused by users unfamiliar with them.

The Klipper Discourse and Klipper Discord are available for many topics related to Klipper - reword some sections to make their use more open.

Avoid referring to the Klipper github repo in this Contact document as we no longer use Klipper github issues at all.

Remove the table of contents as it is largely redundant for the document.

-Kevin


Sineos:
While you are at it:
You might want to mention that the klippy.log also often is found in `~/printer_data/logs/klippy.log` or can be downloaded via the WebIfs

KevinOConnor:
Thanks.  I further updated the Contact.md document.

-Kevin

Sineos (approved):

PR #6384, opened 2023-10-30
Update contributing steps (use Discourse first)

KevinOConnor:
This changes the documentation (and "reviewer required" github script) to indicate that any new enhancements to Klipper should start with a discussion on Klipper Discourse.

This is in association with a discussion I raised a few months ago on Discourse: https://klipper.discourse.group/t/new-submission-process-for-feature-enhancements/9818

-Kevin


bdistin:
Are contributors expected to follow the updated guidelines if pull requests are already made?

I may be in the minority, but respectfully if I had to follow these updated guidelines, I probably wouldn't have made my pull request to begin with. I personally have far too much to keep track of that I don't need yet another place to read/interact with to try to improve the things I use. I have only made one tiny pull request, so I understand my opinion carries little-to-no weight. I also understand you and the reviewers likewise have limited time and attention, so I don't think you are necessarily wrong either. My only intent in sharing this opinion is to make sure you are conscious that added friction to contribute may result in fewer people willing to contribute.

neeamradia:
In reality, all this is going to do is reduce the amount of community contributions to klipper as nobody wants to deal with yet another place to keep track of messages and have another account.

KevinOConnor:
Thanks for the feedback.  There does seem to be general concern about trying to move development conversations to Klipper Discourse.  It seems this change needs more thought, and I suspect there will be further changes to this PR before merging.

-Kevin

KevinOConnor:
It does not seem that there is consensus on this change, so I am going to close this PR.  So, both Discourse and Github PRs seem fine for discussing Klipper code changes.  I have disabled the github bot that automatically closes PRs without a reviewer assigned.  

-Kevin

PR #6385, opened 2023-10-31
Replace deprecated logger.warn with logger.warning

thijstriemstra:
the `logger.warn` alias will be removed in upcoming Python 3.13. Changes are compatible with Python >= 2.7.


KevinOConnor:
Thanks.

-Kevin

PR #6387, opened 2023-11-01
bed_screws: Reset on move error

pedrolamas:
This is a follow up on [this issue](https://github.com/mainsail-crew/mainsail/issues/1624).

If the `stepper_x.position_max` or `stepper_y.position_max` values are smaller than those in the `bed_screws` entries, when one starts the `BED_SCREWS_ADJUST` tool, it will throw a "Move out of range" error, but the tool will be stuck in a running state, yet nothing works until we restart Klipper.

The fix here is to reset the state of the tool in case of error.


KevinOConnor:
Thanks.

-Kevin

PR #6388, opened 2023-11-02
Improve support for python3.12 in configfile.py

KevinOConnor:
It seems python3.12 has removed support for readfp() - use read_file() instead.

-Kevin


thijstriemstra (approved):
Looks good.

PR #6391, opened 2023-11-03
stm32: enable 64KiB bootloader offset for all F4

docgalaxyblock:
Enables the 64KiB bootloader offset for the whole F4 family.


KevinOConnor:
Thanks.

-Kevin

PR #6394, opened 2023-11-08
GD32 Support from Creality Klipper release on K1

sbtoonz:
Add GD32 SOC support for the new creality boards

solves:
https://klipper.discourse.group/t/support-for-new-creality-boards-4-2-2-with-gd32f303/3016

### DIFF
--- a/lib/gd32e23x/gd32e23x.h
+++ b/lib/gd32e23x/gd32e23x.h
@@ -1,0 +1,216 @@
+/*!
+    \file    gd32e23x.h
+    \brief   general definitions for GD32E23x
+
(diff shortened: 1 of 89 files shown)


JamesH1978:
Thankyou for submitting a PR, but i am not sure what issue this fixes? GD32F103 and GD32F303 flash fine with the stock STM32F103 firmware. They have been in the wild for at least a couple of years now and I have not come across one that doesn't flash yet. 

Anyone one with this issue should visit the Kipper Discord and discuss what is happening, so we can guide them through the flash process.

Also your code is causing many errors in the checks, i am not sure code from a clone of klipper would be a good fit for mainline anyway.

Thanks
James

PR #6395, opened 2023-11-08
tmc5160: Increase maximum current error check

KevinOConnor:
It's possible to build and configure tmc5160 drivers with external mosfets that support more than 3 amps.  The actual maximum for tmc5160 drivers is dependent on how the board is wired and the mosfets used. Increase the error check to 10 amps.  This error checking is primarily intended to catch "obvious misconfigurations" (eg, specifying milli-amps instead of amps), and the new value of 10 amps should suffice for this task.

This is an alternative to #6381 (and #6093).

@bigtreetech - FYI.

-Kevin

### DIFF
--- a/klippy/extras/tmc5160.py
+++ b/klippy/extras/tmc5160.py
@@ -257,7 +257,7 @@ FieldFormatters.update({
 ######################################################################
 
 VREF = 0.325
-MAX_CURRENT = 3.000
+MAX_CURRENT = 10.000 # Maximum dependent on board, but 10 is safe sanity check
 
 class TMC5160CurrentHelper:
     def __init__(self, config, mcu_tmc):


shaneshort:
I actually think that this just kicks the can down the road? If we're just arbitrarily bumping the limits why have it at all?

I personally much prefer the BTT implementation having the maximum current defined by the sense resistor value, which for all intents and purposes is the most correct way of defining it

thijstriemstra:
> I personally much prefer the BTT implementation having the maximum current defined by the sense resistor value, which for all intents and purposes is the most correct way of defining it?

> If your concern is that the error message isn't clear enough, can we adjust the error message to suit?

Agreed.

KevinOConnor:
Thanks for the feedback.  I decided to merge this.  Mainly because I don't think merging this change will hamper any alternative.  (That is, the choice of 3 was arbitrary and the new choice of 10 doesn't preclude alternate implementations.)

-Kevin

PR #6397, opened 2023-11-10
config: Add Sovol SV06 Plus

hjmcnew:
config: Add printer-sovol-sv06-plus-2023.config

Add a configuration for the Sovol SV06 Plus. 

Signed-off-by:  Herb McNew <herb.mcnew@gmail.com>


JamesH1978:
Thank you for submitting a PR, I have just one question why do you have a uart_address for each driver when they all use different uart pins, Is this a quirk of the board or does it work fine without this address being set?

Thanks
James

hjmcnew:
> Thank you for submitting a PR, I have just one question why do you have a uart_address for each driver when they all use different uart pins, Is this a quirk of the board or does it work fine without this address being set?

Quirk of the board (I guess). If one comments out the uart_address and attempts to home they get `!! Unable to read tmc uart 'stepper_x' register IFCNT`

JamesH1978 (approved):
Thank you for confirming that uart_address is needed

Looks good to me,

thanks,
James

KevinOConnor:
Thanks.

-Kevin

PR #6398, opened 2023-11-11
doc: omit rpi device version nrs

thijstriemstra:
rpi 5 is out and instead of perpetually updating this text, omit the version nrs (which can be mentioned in the FAQ instead).

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -2,7 +2,7 @@
 
 These instructions assume the software will run on a Raspberry Pi
 computer in conjunction with OctoPrint. It is recommended that a
-Raspberry Pi 2, 3, or 4 computer be used as the host machine (see the
+Raspberry Pi computer be used as the host machine (see the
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thanks for the PR, it might be best to list the versions that are not recommended like the original 2012 pi 1 and the zero 1, i would say the pi2 IMO iss getting a little underpowered these days too as functionality increases.

Also please refer to point 3 in "Common things a reviewer will look for:" section in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and provide a signed-of-by line

Thanks
James

KevinOConnor:
Thanks.  Maybe word it with something like "a Raspberry Pi 2 (or later)".  I would not recommend running Klipper on an rpi1 or rpi0.

-Kevin

thijstriemstra:
thanks for the review, I made the changes @KevinOConnor

KevinOConnor:
Thanks.

-Kevin

PR #6399, opened 2023-11-12
Resolve numpy errors w/ MEASURE_RESONANCES

HumbleDeer:
In lieu of issues like these: https://klipper.discourse.group/t/failed-to-import-numpy-module/4528/7

Supposedly, a recent version of Numpy changed the default implementation of BLAS from Atlas to OpenBLAS, causing the original documentation's installation of Atlas to be fruitless. The exact cause of what changed the default isn't clear to me as of now.

I added the `libopenblas-base` package which could be missing, causing the importing of numpy to fail when running MEASURE_RESONANCES or SHAPER_CALIBRATE commands. When it does fail on numpy because of this package, it was not very verbose about it.

I left the installation of Atlas in, because it doesn't seem to impact anything as of now. 

_For some reason, editing on Klipper3d/klipper submitted my patch branch on my fork of danger-klipper. Should be fine though, since there's no apparent other difference._

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -207,7 +207,7 @@ software dependencies not installed by default. First, run on your Raspberry Pi
 the following commands:
 ```
 sudo apt update
-sudo apt install python3-numpy python3-matplotlib libatlas-base-dev
+sudo apt install python3-numpy python3-matplotlib libatlas-base-dev libopenblas-base
 ```
 
 Next, in order to install NumPy in the Klipper environment, run the command:


HumbleDeer:
## More resources:

Choose whether NumPy uses OpenBLAS or Atlas:
https://stackoverflow.com/questions/12249089/how-to-use-numpy-with-openblas-instead-of-atlas-in-ubuntu

OpenBLAS vs Atlas:
https://stackoverflow.com/questions/17858104/what-is-the-relation-between-blas-lapack-and-atlas

cereal7802:
`root@voron2:~# apt list --installed|egrep -i "blas|atlas"

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

libblas3/oldstable,now 3.9.0-3+deb11u1 arm64 [installed,automatic]
root@voron2:~# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
`


Here is the output of my printer that has been setup for a little while and uses klipper with input shaper. I do not have the atlast package installed, although it is available for the distro. libopenblas-base is also available but not installed.

HumbleDeer:
The package `libopenblas-dev` could be a possible candidate, as it's a metapackage for itself, `libopenblas0`, as well as some other things.

My debian12 can't find a package called libopenblas-base, though.

HumbleDeer:
@shiftingtech I'm tagging you as this may interest you and/or be more your jam? 
You also wrote the fix post in Discord for the bugged udev package

PR #6402, opened 2023-11-15
Update printer-lulzbot-taz6-2017.cfg

fernplant:
Stock nozzle sizes for Lulzbot TAZ6 are 0.5mm


JamesH1978:
Thanks for the PR, please refer to point 3 in "Common things a reviewer will look for:" section in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and provide a signed-of-by line

Thanks
James

PR #6403, opened 2023-11-16
Configurable timeout

EmJay276:
Make trsync_timeout and trsync_single_mcu_timeout configurable

Some people (including me) have (sometimes) issues with a multi mcu setup with a timeout during z-homing.
The default timeout is 25ms, in my case it just don't workout every time. Increasing the limit to 50ms fixes this issue for me entierly, but needs to be done by modifying the mcu.px file after every klipper update.
This PR makes the timeout configurable via the config while keeping the default values as is.

Signed-off-by: Michael Jäger michael@mjaeger.eu

PS
Yes I know the massiv thread for fixing can issues, None of the suggestions helped for me, I have no lost or retransmitted packages, ...

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -58,6 +58,13 @@ serial:
 #   sending a Klipper command to the micro-controller so that it can
 #   reset itself. The default is 'arduino' if the micro-controller
 #   communicates over a serial port, 'command' otherwise.
+#trsync_timeout: 0.025
+#   Sync limit for multi-mcu homing in seconds. The default is 0.025.
+#   Should Klipper detect a communication issue between micro-controllers during
+#   multi-mcu homing then it will raise a "Communication timeout during homing"
+#   error.
+#trsync_single_mcu_timeout: 0.250
+#   Sync limit for single-mcu homing in seconds. The default is 0.250.
 ```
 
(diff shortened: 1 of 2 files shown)


Sineos:
As mentioned in the thread you referred to, I would caution against applying this temporary solution as it may only address certain uncommon scenarios without resolving the underlying problem. Users who use this "fix" carelessly could encounter issues that are incredibly difficult to diagnose.

EmJay276:
> As mentioned in the thread you referred to, I would caution against applying this temporary solution as it may only address certain uncommon scenarios without resolving the underlying problem. Users who use this "fix" carelessly could encounter issues that are incredibly difficult to diagnose.

I am aware of the issues caused by longer timeouts, but should the user not be able to set this in the config instead of the python code? The option already exists, why not make it able to be configered and therefore also show up in the log?

I think in my case it has something todo with the USB -> CAN adapter, cause USB is sometime bussy and therefore the CAN message will timeout.

silence48:
Was this included in 0.12

Sineos:
No and likely will never be as it is not considered a fix but a bandaid that might lead to more undesirable behavior.

rjacobs1969:
having the same issue with multi mcu z homing; better a bandaid than printers that don't work.

PR #6405, opened 2023-11-17
feat: Add configurable i2c timeout

EiNSTeiN-:
I am playing around with a custom sensor which I query using I2C. The sensor runs off a rp2040 which acts as the I2C target in this case. I wrote the code in CircuitPython for convenience, unfortunately my logic analyser shows the response time from the rp2040 to a read request is routinely in the 4ms to 6ms range which trips Klipper's hardcoded I2C timeout.

In this PR I made the timeout configurable in klippy and added the `i2c_timeout` config option which defaults to `5000` (the formerly hardcoded timeout value).

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4722,4 +4722,8 @@ via the `i2c_speed` parameter. All other Klipper micro-controllers use a
 #   The Klipper implementation on most micro-controllers is hard-coded
 #   to 100000 and changing this value has no effect. The default is
 #   100000. Linux, RP2040 and ATmega support 400000.
+#i2c_timeout:
+#   The time in microseconds after which a i2c timeout occurs. Defaults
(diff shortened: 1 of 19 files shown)


KevinOConnor:
Thanks.  As high-level feedback, I don't think this will work properly in common usage.  The issue is that i2c commands currently block all other task execution in the Klipper mcu code.  If a device of any kind were to block the task execution for more than a few milliseconds it can cause hard to debug problems in the entire system (eg, skewed clock estimation, spurious host retransmits, lost sensor readings, spurious watchdog reboots, etc.).  In that sense, the current 5ms hardcoded timeout is basically intended to catch the point where we need to resume task execution regardless of whether the device is working normally or not.

Cheers,
-Kevin

EiNSTeiN-:
That makes sense, this change was useful mainly during prototyping my own code and I've since rewritten it in C so the response times fall within the 5ms delay without changes to klipper. I'll close this PR

PR #6408, opened 2023-11-22
bed_mesh: Restore originally-saved profiles on BED_MESH_CLEAR

voidtrance:
BedMeshCalibrate.probe_finalize() updates the config file with the newly generated mesh data. That method is called during the completion of meshing. This means that every BED_MESH_CALIBRATE command updates the config file with new mesh data.

This can lead to mesh profiles being unintentionally overwritten. This can happen because BED_MESH_CLEAR does not restore the old mesh data in the config file. So, if a user runs the following commands:

   BED_MESH_CALIBRATE
   BED_MESH_CLEAR
   ...
   SAVE_CONFIG

the mesh will be saved despite the user having run BED_MESH_CLEAR.

Fortunately, the newly generated bedmesh is stored in a separate section of the config that holds pending changes. Therefore, all that is needed in order to avoid storing a bed mesh is to remove the section for the current profile, which removes it from the set of pending changes.

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -87,6 +87,7 @@ def parse_gcmd_coord(gcmd, name):
 class BedMesh:
     FADE_DISABLE = 0x7FFFFFFF
     def __init__(self, config):
+        self.name = config.get_name()
         self.printer = config.get_printer()
         self.printer.register_event_handler("klippy:connect",
                                             self.handle_connect)
@@ -264,6 +265,13 @@ class BedMesh:
             gcmd.respond_info("Bed has not been probed")
     cmd_BED_MESH_CLEAR_help = "Clear the Mesh so no z-adjustment is made"
     def cmd_BED_MESH_CLEAR(self, gcmd):
+        # Remove the current bed mesh profile from the config file.
+        # We do this because BedMeshCalibrate.probe_finalize() updates
+        # the config file, which can lead to the wrong mesh being saved.
+        profile_name = self.pmgr.get_current_profile()
+        configfile = self.printer.lookup_object('configfile')
+        cfg_name = self.name + " " + profile_name
+        configfile.remove_section(cfg_name)
         self.set_mesh(None)
     cmd_BED_MESH_OFFSET_help = "Add X/Y offsets to the mesh lookup"
     def cmd_BED_MESH_OFFSET(self, gcmd):


Arksine:
I'm not sure this is the correct behavior.  It could unintentionally remove a saved profile when the user runs `BED_MESH_CLEAR`.

For example:
```
BED_MESH_PROFILE LOAD=bed60
..do stuff
BED_MESH_CLEAR

BED_MESH_CALIBRATE PROFILE=bed80
SAVE_CONFIG
```
The above would remove `bed60`, however its unlikely that the user really intended to remove it.

voidtrance:
> I'm not sure this is the correct behavior. It could unintentionally remove a saved profile when the user runs `BED_MESH_CLEAR`.
> 
> For example:
> 
> ```
> BED_MESH_PROFILE LOAD=bed60
> ..do stuff
> BED_MESH_CLEAR
> 
> BED_MESH_CALIBRATE PROFILE=bed80
> SAVE_CONFIG
> ```
> 
> The above would remove `bed60`, however its unlikely that the user really intended to remove it.

I'd argue that they did. Isn't that the point of the `BED_MESH_CLEAR`? If I am misunderstanding how `BED_MESH_CLEAR` is supposed to work, I am sorry. But in that case, there is still the problem of the unintentional saving of a profile. How would you suggest that is fixed?

voidtrance:
> I'm not sure this is the correct behavior. It could unintentionally remove a saved profile when the user runs `BED_MESH_CLEAR`.
> 
> For example:
> 
> ```
> BED_MESH_PROFILE LOAD=bed60
> ..do stuff
> BED_MESH_CLEAR
> 
> BED_MESH_CALIBRATE PROFILE=bed80
> SAVE_CONFIG
> ```
> 
> The above would remove `bed60`, however its unlikely that the user really intended to remove it.

I just noticed that your example uses `BED_MESH_PROFILE LOAD=bed60`. As far as I can tell, BED_MESH_PROFILE LOAD` does not update the config file. Therefore, there would be no pending changes beyond the profile created by the `BED_MESH_CALIBRATE PROFILE=bed80`. Therefore, the `bed60` profile should not be removed as the config is not updated to remove it.

Arksine:
> I'd argue that they did. Isn't that the point of the BED_MESH_CLEAR? If I am misunderstanding how BED_MESH_CLEAR is supposed to work, I am sorry. But in that case, there is still the problem of the unintentional saving of a profile. How would you suggest that is fixed?

`BED_MESH_PROFILE` has an explicit `REMOVE` parameter that should be used when a user wants to remove a profile.  The `BED_MESH_CLEAR` method is intended to clear the internal (volatile) mesh, it is not intended to remove persistent data.

> I just noticed that your example uses BED_MESH_PROFILE LOAD=bed60. As far as I can tell, BED_MESH_PROFILE LOAD does not update the config file as there are no calls to ProfileManager.save_profile() in that call stack. Therefore, there would be no pending changes beyond the profile created by the BED_MESH_CALIBRATE PROFILE=bed80. Therefore, the bed60 profile should not be removed as the config is not updated to remove it.

The example presumes a previously saved profile named `bed60` exists in the autosave section.  It is intended to show that a user could load a profile, clear the volatile mesh with `BED_MESH_CLEAR`, then do something that requests `SAVE_CONFIG`.  It could be `BED_MESH_CALIBRATE`, `PROBE_CALIBRATE`, etc.  The `bed60` profile would be removed with no notification to the user.

With regard to existing behavior, the user is notified that the autosave section has been updated after `BED_MESH_CALIBRATE`. It behaves like all other calibration routines in Klipper that save persistent state, a restart will reverse it, `SAVE_CONFIG` will commit it.

Arksine:
One additional comment...I believe that your proposed changes wouldn't just revert the pending changes, it would remove that section entirely if it exists in the autosave section of `printer.cfg`.

voidtrance:
> `BED_MESH_PROFILE` has an explicit `REMOVE` parameter that should be used when a user wants to remove a profile. The `BED_MESH_CLEAR` method is intended to clear the internal (volatile) mesh, it is not intended to remove persistent data.

I don't want to remove a saved profile. I want to avoid saving the current mesh, potentially overriding a profile.

> With regard to existing behavior, the user is notified that the autosave section has been updated after `BED_MESH_CALIBRATE`. It behaves like all other calibration routines in Klipper that save persistent state, a restart will reverse it, `SAVE_CONFIG` will commit it.

The problem is that there is no way to "throw away" a mesh that is undesired, yet save something else that is needed. If I ran:

```gcode
BED_MESH_CALIBRATE
BED_MESH_CLEAR

PROBE_CALIBRATE
```
and I only wanted to save the probe settings, there is no way for me to do that without restarting and re-running `PROBE_CALIBRATE`.

> One additional comment...I believe that your proposed changes wouldn't just revert the pending changes to the profile, it would remove that section entirely if it exists in the autosave section of printer.cfg.

Yes, you are correct on this. Personally, I think this is a bug with how the autosave data is handled. Having said that, it appears that this change, as is, is not what I wanted to implement. So, I would appreciate some advice on how to approach this.

It seems to me that in addition to this change, the `PrinterConfig.set()`, `PrinterConfig.remove_section()`, and `PrinterConfig.cmd_SAVE_CONFIG()` will have to be changed as such:

1. `PrinterConfig.set()` does not update `self.autosave.fileconfig`.
2. `PrinterConfig.remove_section()` does not remove anything from `self.autosave.fileconfig`
3. `PrinterConfig.cmd_SAVE_CONFIG` updates `self.autosave` from `self.status_save_pending`.

Of course, I'll have to check all the places where `set` and `remove_section` are used...

Arksine:
> I don't want to remove a saved profile. I want to avoid saving the current mesh, potentially overriding a profile.

Understood.  The default behavior of `BED_MESH_CALIBRATE` is to save the latest result to the generic `default` profile.  When desired, the profile can be saved to a more descriptive name using `BED_MESH_PROFILE`.  This avoids unintentionally overwriting a profile.

At request, a `PROFILE` parameter was added to `BED_MESH_CALIBRATE` allowing the user to forego the second step.

> Yes, you are correct on this. Personally, I think this is a bug with how the autosave data is handled. Having said that, it appears that this change, as is, is not what I wanted to implement. So, I would appreciate some advice on how to approach this.

It is behaving as intended, so it is not a bug.  I do not think the default behavior of `BED_MESH_CLEAR` should change, however I would not be opposed to adding a parameter that allows the user to revert back to the original profile.  This would require some tracking in the profile manager to store the original state before a profile is saved, and a method to revert the changes.

If you desire a more generic method to revert changes to the autosave data, I think that is something Kevin would need to provide guidance on.

voidtrance:
> It is behaving as intended, so it is not a bug. I do not think the default behavior of `BED_MESH_CLEAR` should change, however I would not be opposed to adding a parameter that allows the user to revert back to the original profile. This would require some tracking in the profile manager to store the original state before a profile is saved, and a method to revert the changes.

I am not sure if that is enough to prevent the issue described, unless reverting to the old settings also updates the autosave state. Is that what you were suggesting/thinking?

PR #6410, opened 2023-11-28
Rework toolhead and move_queue flushing

KevinOConnor:
This PR fixes "Timer too close" errors in the pwm_tool module when SET_PIN commands are issued while the toolhead is idle.  (And also likely fixes similar errors that may occur with "MANUAL_STEPPER SYNC=0" commands.)

Previously, the toolhead code would only manage flushing of the lookahead, kinematic, and mcu move_queues while the toolhead was in an active state.  Once the lookahead queue becomes empty, the toolhead would disable its internal flushing timers.  This PR reworks the toolhead code so that the its flush timer will stay active any time there is data in any of the kinematic queues, regardless of the lookahead state.

@Cirromulus , @dmbutyugin - FYI.

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20231208: Several undocumented config parameters in the `[printer]`
+config section have been removed (the buffer_time_low,
+buffer_time_high, buffer_time_start, and move_flush_time parameters).
+
 20230826: If `safe_distance` is set or calculated to be 0 in `[dual_carriage]`,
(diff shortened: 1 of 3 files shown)


A-Panic:
I can confirm that with this branch I'm able to manually control the pwm output without errors.

Cirromulus:
While the changes look sensible, unfortunately I don't feel very confident in my knowledge about this part of the toolhead code and thus can't give a detailed review.

I would personally like an overview about the states and their implications in some sort of documentation, but this is perhaps for a different MR.

At least I could test it sucessfully.

PR #6413, opened 2023-12-01
kinematics: Fix "Unable to infer active extruder stepper" errors.

xtoddx:
# Background

x-in-1-out hotend configurations currently require naming the tool in the slicer and adding an extra `extruder=...` paramter to SET_PRESSURE_ADVANCE g-code. Failure to add the parameter results in *Unable to infer active extruder stepper* errors and aborted prints on the first tool change (at least with PrusaSlicer-like software).

As the error message states, there is an attempt to infer the value, but currently the value of printer.toolhead.extruder.extruder_stepper is _always_ the primary extruder_stepper (the one named "extruder").

This updates the sync_to_extruder method so the printer model is kept consistent with the currently active extruder_stepper by assigning it to the PrinterExtruder.

### DIFF
--- a/klippy/kinematics/extruder.py
+++ b/klippy/kinematics/extruder.py
@@ -69,6 +69,7 @@ class ExtruderStepper:
         self.stepper.set_position([extruder.last_position, 0., 0.])
         self.stepper.set_trapq(extruder.get_trapq())
         self.motion_queue = extruder_name
+        extruder.extruder_stepper = self
     def _set_pressure_advance(self, pressure_advance, smooth_time):
         old_smooth_time = self.pressure_advance_smooth_time
         if not self.pressure_advance:


JamesH1978:

Thanks
James

xtoddx:
Updated. Apologies for the oversight.

KevinOConnor:
Thanks.  However, this PR doesn't look like it would work correctly when multiple steppers are simultaneously assigned to the same extruder.  It doesn't look correct to change the main extruder's reference to its specific stepper.

-Kevin

bazarovdev:
Currently both [PrusaSlicer](https://github.com/prusa3d/PrusaSlicer/blob/5dc04b4e8f14f65bbcc5377d62cad3e86c2aea36/src/libslic3r/GCode/WipeTower.cpp#L110) and [Orca](https://github.com/SoftFever/OrcaSlicer/blob/31be67926e633c3e12970e026097c27032a94a0b/src/libslic3r/GCode/WipeTower2.cpp#L103) slicers in multiextruder configurations + `Wipe Tower` add 

```
        if (m_gcode_flavor == gcfKlipper)
            m_gcode += "SET_PRESSURE_ADVANCE ADVANCE=0\n";
        else
```
and it causes this error `Unable to infer active extruder stepper` (SuperSlicer did implement fix for it https://github.com/supermerill/SuperSlicer/issues/2934).

Is it feasible to detect currently used stepper/s and apply `SET_PRESSURE_ADVANCE ADVANCE=0` to them or should it be fixed on the slicers level?

bazarovdev:
@gabin1369 Thanks! Looks promising, I will try it.

> BUT you have to use SET_PRESSURE_ADVANCE_BASE macro when the pressure advance value is not zero
Can you explain why is that? Looking on macro, I don't see problem running it with non-zero values.

gabin1369:
Sorry, forget what I wrote. I updated my previous comment and changed the macro

PR #6414, opened 2023-12-01
stats: add fan speed stats to log

CODeRUS:
Signed-off-by: Andrei Kozhevnikov <coderusinbox@gmail.com>

### DIFF
--- a/klippy/extras/controller_fan.py
+++ b/klippy/extras/controller_fan.py
@@ -43,6 +43,8 @@ class ControllerFan:
     def handle_ready(self):
         reactor = self.printer.get_reactor()
         reactor.register_timer(self.callback, reactor.monotonic()+PIN_MIN_TIME)
+    def stats(self, eventtime):
+        return self.fan.stats(eventtime)
     def get_status(self, eventtime):
         return self.fan.get_status(eventtime)
     def callback(self, eventtime):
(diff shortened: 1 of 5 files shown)


Sineos:
Can you explain the rationale?
The log output would never reflect all changes and also does not align to any layer information or similar. Additionally, I do not see any diagnostic benefit, but I might be missing something.

CODeRUS:
@Sineos exactly for diagnostic purposes. Most common cases are ADC out of range errors after turning on part cooling fan. Can see if temperature goes down (probably) cause of turning on cooling.

CODeRUS:
then i can check logs to see if there any problems

firefox_2023-12-01_22-02-33
([via](https://coderus.openrepos.net/klipper_logs/2609c8da58a1d73fbf8e27f69845faa4))

KevinOConnor:
Thanks.  As feedback, the code change here doesn't seem correct to me.  The main diagnostic case (that I'm aware of) is reporting the "layer cooling fan" speed, but this change doesn't add a `stats()` callback to the `PrinterFan()` class.  Maybe I'm missing something.  If we do want to add more diagnostic info, I think we should restrict it to just the layer fan, as most other fans (eg, heater_fan) are unlikely to be needed for diagnostics.

-Kevin

PR #6420, opened 2023-12-07
Add maximum_mcu_duration support to pwm_tool

KevinOConnor:
This adds support for `maximum_mcu_duration` to the pwm_tool.py module.  This is intended to allow users to configure a safety check within the mcu to check for lost communication between host and mcu while a power tool is enabled.  With the merge of #6410 it should now be possible to implement this check.

I've run some basic tests on this code, but I don't have hardware to test real world cases.

Once this support is added, I will look to deprecate the existing `maximum_mcu_duration` in the `[output_pin]` module.  (That module has been getting complex, and I think the code may be more maintainable going forward if we can move users that need maximum_mcu_duration to the pwm_tool module.)

@Cirromulus - FYI.

-Kevin

### DIFF
--- a/klippy/extras/pwm_tool.py
+++ b/klippy/extras/pwm_tool.py
@@ -5,8 +5,8 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import chelper
 
-PIN_MIN_TIME = 0.100
 MAX_SCHEDULE_TIME = 5.0
+CLOCK_SYNC_EXTRA_TIME = 0.050
 
 class error(Exception):
     pass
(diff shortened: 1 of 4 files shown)


KevinOConnor:
FYI, I have reworked the code on this branch slightly.  The flush notification system is now, hopefully, a little easier to understand.

-Kevin

klippy/extras/pwm_tool.py
```diff
@@ -74,39 +86,50 @@ def _build_config(self):
...
         if ret:
             raise error("Internal error in stepcompress")
+        # Notify toolhead so that it will flush this update
+        wakeclock = clock
+        if self._last_value != self._default_value:
+            # Continue flushing to resend time
+            wakeclock += self._duration_ticks
+        wake_print_time = self._mcu.clock_to_print_time(wakeclock)
+        self._toolhead.note_kinematic_activity(wake_print_time
+                                               + CLOCK_SKEW_EXTRA_TIME)
```
Cirromulus, line 121: `SKEW_EXTRA_TIME` is the buffer time to allow latency between MCU and host?

KevinOConnor: It's to account for minor changes in clock synchronization.  If we need to flush to clock 800000, which translates to time 100.0, but when time 100.0 actually comes around if an updated clock synchronization translates that to 799998, then an update at 800000 wont get flushed.

Basically, there is no harm in leaving flushing enabled for longer than necessary, so it's easier to add a little bit of extra time than to risk going under.

-Kevin

Cirromulus: Ah, so to flush further into "the future", I understand now.

PR #6422, opened 2023-12-09
Makefile: Use C preprocessor from the target toolchain

### LINKED ISSUE #6421: Klipper fails to build with Apple's cpp
I'm building STM32 target for Klipper on MacOS 13.6. It fails with a cryptic message:

```
$ make
...
  Preprocessing out/src/generic/armcm_link.ld
clang: error: no input files
make: *** [out/src/generic/armcm_link.ld] Error 1
$
```
Tracing the build shows the following:
```
$ make V=1
...
cpp -Iout/ -P -MD -MT out/src/generic/armcm_link.ld src/generic/armcm_link.lds.S -o out/src/generic/armcm_link.ld
clang: error: no input files
make: *** [out/src/generic/armcm_link.ld] Error 1
$
```
And indeed running this command fails:
```
$ cpp -Iout/ -P -MD -MT out/src/generic/armcm_link.ld src/generic/armcm_link.lds.S -o out/src/generic/armcm_link.ld
clang: error: no input files
$
```
This preprocessor is from Xcode:
```
$ cpp --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
$
```

The preprocessor coming with GCC works:
```
$ make CPP=arm-none-eabi-cpp
...
  Creating hex file out/klipper.bin
$
```

`CPP` is set to `cpp` in Makefile, I guess changing it to `$(CROSS_PREFIX)cpp` is not going to break anything.

## Environment

- MacOS 13.6 (arm64)
- gcc is from Homebrew `gcc-arm-embedded`
- Native toolchain is from Xcode 15.0.1

dottedmag:
C preprocessor coming with the host toolchain may not understand the command-line syntax for preprocessing linker scripts.

Closes #6421

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ LD=$(CROSS_PREFIX)ld
 OBJCOPY=$(CROSS_PREFIX)objcopy
 OBJDUMP=$(CROSS_PREFIX)objdump
 STRIP=$(CROSS_PREFIX)strip
-CPP=cpp
+CPP=$(CROSS_PREFIX)cpp
 PYTHON=python3
 
 # Source files


bevanweiss:
Everything compiles fine for me with this change in.
The Makefile is currently inconsistent on this aspect
```diff
# Common command definitions
CC=$(CROSS_PREFIX)gcc
AS=$(CROSS_PREFIX)as
LD=$(CROSS_PREFIX)ld
OBJCOPY=$(CROSS_PREFIX)objcopy
OBJDUMP=$(CROSS_PREFIX)objdump
STRIP=$(CROSS_PREFIX)strip
- CPP=cpp
+ CPP=$(CROSS_PREFIX)cpp
PYTHON=python3
```

so it does seem like a simple fix.
@KevinOConnor any chance of getting this in to trim the PR backlog?

bevanweiss (approved):
Looks good.  Worked for me.

KevinOConnor:
Thanks.  It is not clear to me what real-world systems this change would allow Klipper to build on.  I originally set `CPP=cpp` because I ran into cross compiling toolchains that did not include cpp.  I'd prefer not to make build changes that could introduce a regression without first identifying the set of users that benefit from the change.

If someone is building on an unusual machine, it's possible to just override CPP directly using something like `make CPP=/somewhere/cpp` .

Cheers,
-Kevin

dottedmag:
> Thanks. It is not clear to me what real-world systems this change would allow Klipper to build on.

All Macs ever. XCode's cpp is not compatible with whatever is thrown at it.

KevinOConnor:
> All Macs ever. XCode's cpp is not compatible with whatever is thrown at it.

Thanks.  Is that the only change necessary to get Klipper to build on a mac?  How does one get all the other packages needed to build Klipper on a mac (the avr toolchain, the arm toolchain, the bash/gcc support, etc.)?

-Kevin

dottedmag:
That was the only change I had to do to build an AVR image for Klipper. AVR toolchain from Homebrew and bash from base system work fine. Compiler provided by XCode also works.

I haven't tested ARM toolchain, as I don't have any boards.

OmegaRelay:
> I haven't tested ARM toolchain, as I don't have any boards.

The change from this PR fixed my mac based build for a STM32 based znp robin board

PR #6426, opened 2023-12-12
Add rp2040 to MCUs with internal temp sensors

RazorSh4rk:
Small change, came up in a discord thread and I wanted to add it since the skr pico is so popular

Signed-off-by: Levi Szabo <iamrazorshark@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2559,7 +2559,7 @@ sensor_type: LM75
 
 ### Builtin micro-controller temperature sensor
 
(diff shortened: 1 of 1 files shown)


JamesH1978:

Thanks
James

RazorSh4rk:
Added, thanks

JamesH1978 (approved):
Can't see any issues in adding that to the documentation. I can confirm the temperature is readable from those mcu's.

Thanks
James

KevinOConnor:
Thanks.

-Kevin

PR #6427, opened 2023-12-14
Debian bookworm install

BlackStump:
Debian bookworm klipper install more inline with current klipper frontends
Includes a klipper.service file which maybe useful for other install files
Includes directories for current frontends most of the info is straight from Mainsails manual install page.
Includes klipper-mcu.service install 

I have next to no programming skills so hopefully some one can have a check on the quailty, I have tested on a armbian os

Trevor Wilson   altocoey@hotmail.com


JamesH1978:
Thank you for submitting a PR, pleas refer to point 3 in "What to expect in a review" in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and provide a proper signed off by line.

That being said I have looked at your files and I am not sure this will work. You have included the mcu service which will likely fail as the actual service has not been made and "flashed" therefore the underlying bin not available to run,  not to mention not working at all on non sbc based devices or systems with a user not called "pi", Klipper's install base is not just Pi and its variants/clones anymore. Also your service file assumes that Moonraker has been installed first, and that it runs first. Which would mean if Moonraker was installed then it would not run as it would have no klipper socket to connect to.

Apologies but I don't think this will be good for submission

Thanks
James.

BlackStump:
no apologies needed you are correct about the pi user I did not think that through and it is a deviation fron the original debian install .

BlackStump:
@JamesH1978  I have updated the script [here](https://github.com/BlackStump/KlipperUI/blob/main/scripts/klipper-install.sh)
it is part of a complete install package but will also work as a stand alone script, can you have a look to see if it is worth another pr 

> Also your service file assumes that Moonraker has been installed first, and that it runs first. Which would mean if Moonraker was installed then it would not run as it would have no klipper socket to connect to.

I dont understand that assumption the only consideration to moonraker is that if you are going to use Fluidd/Mainsail then of course you will install Moonraker but as the service file stands Klipper will work as intended without Moonraker? I would like to get my head around your assumption to further my knowledge.

PR #6428, opened 2023-12-14
tmc2240: Fixed the problem that the uart mode of tmc2240 cannot be enabled normally.

kluoyun:
tmc2240: Fix tmc2240 cannot enable uart mode.

Using double quotes to get the configuration may lead to incorrect results in some system environments, always being None. It is recommended to use single quotes to create strings that do not contain special characters.

Signed-off-by:  Xiaokui Zhao <xiaok@zxkxz.cn>


KevinOConnor:
I'm not sure what problem you are looking to solve, but python doesn't make a distinction between double quotes and single quotes, so this change doesn't alter the functionality.

-Kevin

FLYmaker:
> 我不确定您要解决什么问题，但 python 不区分双引号和单引号，因此此更改不会改变功能。
> 
> -凯文

It seems so, but when we actually tested the current klipper source code, it was unable to use the 2240 uart mode. I think you need to test it. This is the conclusion we drew from the test. It may be a problem with the python version. We are using python3 not 2.
After modification, the uart configured with 2240 will not report an error.

PR #6430, opened 2023-12-16
virtual_sdcard: Open file in utf-8 encoding

CODeRUS:
When printing file with utf-8 comments in it, pause/resume breaks printing process, it tries to resume at wrong position, caused by utf-8 (or utf-16) characters byte size.

```
00:12:07  // Unknown command:"F30000"
00:12:40  // Unknown command:"Y109.644"
00:13:04  // Unknown command:"ERIMETER"
00:13:33  // Unknown command:"Y117.633"
00:14:17  // Unknown command:"Y107.81"
```

This simple PR fixes issue

### DIFF
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -177,7 +177,7 @@ class VirtualSD:
             if fname not in flist:
                 fname = files_by_lower[fname.lower()]
             fname = os.path.join(self.sdcard_dirname, fname)
-            f = io.open(fname, 'r', newline='')
+            f = io.open(fname, 'r', newline='', encoding='utf-8')
             f.seek(0, os.SEEK_END)
             fsize = f.tell()
             f.seek(0)


thijstriemstra (approved):

thijstriemstra:
looks like a sensible change?

zxy16305:
The resume process begins at `self.current_file.seek(self.file_position)`, where `self.file_position` represents a byte-length offset. However, `file_position` is incremented by `len(line)`, which corresponds to the string length. If there are any UTF-8 characters preceding the 'PAUSE' command, the file position must be set before the occurrence of 'PAUSE'.

To address this issue, replacing the string length with the byte length is sufficient. Please find the modification below:

https://github.com/Klipper3d/klipper/pull/6469

```python
# Before
# next_file_position = self.file_position + len(line) + 1

# After
# next_file_position = self.file_position + len(line.encode()) + 1

KevinOConnor:
Hopefully this is fixed now with the merge of #6472 .

Thanks.
-Kevin

3vilson:
I have had this problem for years, can you give me a step by step explanation to fix this? 

I attached my gcode Fusion 360 names every  new Body "Körper" in German.

[cawca32_ASA_7m22s.zip](https://github.com/user-attachments/files/17607916/cawca32_ASA_7m22s.zip)

Sineos:
This is considered fixed as per @KevinOConnor's comment above already 9 months ago with no new reports that I'm aware of. 
If you are still experiencing the problem, please open a thread in the discourse and provide a test case along with the klippy.log showing the unwanted behavior.

dewi-ny-je:
@Sineos so why is the ticket under the "Issues" still open?

PR #6431, opened 2023-12-17
Update LIS2DW example to match current BTT documentation

buzztiaan:
As found while testing my own BTT S2DW board, the klipper doc example was using wrong pins (and method?).

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -314,8 +314,11 @@ Restart Klipper via the `RESTART` command.
 serial: /dev/serial/by-id/usb-Klipper_rp2040_<mySerial>
 
 [lis2dw]
-cs_pin: lis:gpio1
-spi_bus: spi0a
+cs_pin: lis:gpio9
+#spi_bus: spi0a
+spi_software_sclk_pin: lis:gpio10
+spi_software_mosi_pin: lis:gpio11
+spi_software_miso_pin: lis:gpio8
 axes_map: x,z,y
 
 [resonance_tester]


JamesH1978:

Thanks
James

Sineos:
I'm not sure if this is the right thing to do.

You have two ways of connecting a SPI IMU like the LIS2DW or ADXL3xx:

**Hardware SPI:**
```
[...]
cs_pin: lis:gpio1
spi_bus: spi0a
```

**Software SPI:**
```
[...]
cs_pin: lis:gpio9
spi_software_sclk_pin: lis:gpio10
spi_software_mosi_pin: lis:gpio11
spi_software_miso_pin: lis:gpio8
```

The S2DW is a combination of a RP2040 MCU and a LIS2DW and BTT (in this very product) choose to wire it in software SPI due to an issue with hardware SPI that is also mentioned here: https://www.klipper3d.org/Measuring_Resonances.html#measuring-the-resonances
grafik

Generally I'd prefer to use hardware SPI but with this combination it might be preferable to use software SPI for the casual user.
If we go by this, then the entire chapter should be reworked and software SPI be recommended for the RP2040 / SPI combination.

buzztiaan:
The 'SW instead of HW' is just copied from BTT's example. As the original config blurb came from them, it seemed wise to update it to mirror.

Beside the 'SW instead' , its also using different pins. The specified pins aren't connected to the S2DW on this specific product. Also i haven't seen any other products with the LIS2DW chip included.

Sineos:
The documentation is meant to explain the principle and not be tailored to a single product.

There is a similar product from [Fly](https://mellow.klipper.cn/?spm=a2g0o.detail.1000023.1.c5665Pp05Pp0He#/advanced/usb_lis2dw) which is using different pins. 

And still the hardware SPI is preferable (hoping the SPI issue can be resolved).

buzztiaan:
The linked fly docs say they DO use the same pins as my pr, plus also SW spi

Sineos:
Fly:
```
[lis2dw]
cs_pin: LIS:gpio9
spi_software_sclk_pin: LIS:gpio10
spi_software_mosi_pin: LIS:gpio11
spi_software_miso_pin: LIS:gpio12
```

BTT:
```
[lis2dw]
cs_pin: lis:gpio9
spi_software_sclk_pin: lis:gpio10
spi_software_mosi_pin: lis:gpio11
spi_software_miso_pin: lis:gpio8
```

¯\\\_(ツ)_/¯

buzztiaan:
Oh, missed the miso pin ;) either way, its a lot closer :D hahaha
Maybe its more suitable for a example-config then?

Sineos:
In principle, I agree with you, but then again, the entire documentation with respect to this would have to be reworked.

ADXL3xx Situation:
* Stand alone break-out board are existing
* For these boards the documentation is correct as is
* For the ADXL + RD2040 the same issue applies: SPI will error out on the first attempt but work for the subsequent ones
* For the commercial made boards that integrate an ADXL with a RD2040, e.g. https://github.com/bigtreetech/ADXL345/blob/master/Firmware/sample-bigtreetech-adxl345-v2.0.cfg, the documentation is as "wrong" as for the LIS2DW since the manufacturers work around the SPI issue by using software SPI

LIS2DW situation:
* Except https://wiki.dfrobot.com/Gravity_I2C_LIS2DW12_Triple_Axis_Accelerometer_SKU_SEN0409 I do not know any IMU break-out boards with this chip and the DFrobot one is I2C and not SPI
* The documentation is correct per se and for sure the manufacturers like BTT or Fly would have chosen hardware SPI if there was not this SPI issue
* The [KUSBA-Pro](https://github.com/xbst/KUSBA-PRO/blob/master/Firmware/adxlmcu.cfg) uses hardware SPI, but this board is based on a STM32F042 and as such not affected by the SPI issue

Unfortunately, there is no one-solution-fits-all and running behind the potential pin and wiring scenarios will not really take you anywhere.

PR #6432, opened 2023-12-17
Simplification of host bulk sensor code

KevinOConnor:
This PR introduces new klippy/extras/bulk_sensor.py code and is intended to simplify the existing adxl345.py, mpu9250.py, lis2dw.py, and angle.py code.  This PR is an internal rework and should not produce a user noticeable change.

The existing host "bulk sensor" modules have a lot of duplicate code.  This makes it harder to maintain these modules, and makes it harder to add new "bulk sensors".  The goal of this PR is to move much of the low-level message handling code to bulk_sensor.py and to keep the per-chip modules focused just on chip specific logic.  Ideally this will make it easier to add future "bulk sensors".

This PR will also facilitate refactoring of the mcu code.  MCU code changes will require users to reflash their MCUs, so I'll plan on introducing those changes in a separate PR.

@garethky - FYI.

-Kevin

### DIFF
--- a/klippy/extras/adxl345.py
+++ b/klippy/extras/adxl345.py
@@ -220,9 +220,9 @@ class ADXL345:
         self.api_dump = bulk_sensor.APIDumpHelper(
             self.printer, self._api_update, self._api_startstop, API_UPDATES)
         self.name = config.get_name().split()[-1]
-        wh = self.printer.lookup_object('webhooks')
-        wh.register_mux_endpoint("adxl345/dump_adxl345", "sensor", self.name,
-                                 self._handle_dump_adxl345)
+        hdr = ('time', 'x_acceleration', 'y_acceleration', 'z_acceleration')
+        self.api_dump.add_mux_endpoint("adxl345/dump_adxl345", "sensor",
+                                       self.name, {'header': hdr})
     def _build_config(self):
         cmdqueue = self.spi.get_command_queue()
         self.query_adxl345_cmd = self.mcu.lookup_command(
@@ -345,10 +345,6 @@ class ADXL345:
             self._start_measurements()
         else:
             self._finish_measurements()
-    def _handle_dump_adxl345(self, web_request):
-        self.api_dump.add_client(web_request)
-        hdr = ('time', 'x_acceleration', 'y_acceleration', 'z_acceleration')
-        web_request.send({'header': hdr})
     def start_internal_client(self):
         cconn = self.api_dump.add_internal_client()
         return AccelQueryHelper(self.printer, cconn)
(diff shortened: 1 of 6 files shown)


garethky:
Exciting! Looking forward to what the MCU code looks like.

Coincidentally, I re-based last night, and got the dreaded "Too many message ids" error when compiling the MCU code. Maybe this work will hold off having to deal with that. (demo branch: https://github.com/garethky/klipper/tree/adc-endstop-rebase)

KevinOConnor:
I have published the corresponding MCU cleanups to a dev branch at: https://github.com/KevinOConnor/klipper-dev/tree/work-bulksensor-20231217

That dev branch also includes changes from this PR and PR #6433.

Separately, I think it would be useful if we could support the hx71x and ads1263 as "bulk sensors" in the mainline Klipper repo today.  I understand that there is ongoing work with load cells, but I think the basic "bulk sensor" support (as exportable over webhooks/motan) has value on its own.

-Kevin

garethky:
Sounds good. I'm happy to refactor what I have to work with `sensor_bulk`.

I'll need a way to call the `adc_endstop` with each new sample [[ref](https://github.com/garethky/klipper/blob/42f6d6d7d7c9b919e58173c1c32c09fdd8f7a2dc/src/sensor_multiplex_adc.c#L88)]. Taking sensor_bulk as a pattern, I'd be doing something inside the `sensor_x` code. Is that the way you think it should work?

I don't personally have a strong reason to submit the ADS1263 code, I was planning on deleting it for an eventual PR. Its very much a prototyping thing and I don't think there will be any commercial interest in having it in klipper. But if you see the value for hobbyists and tinkering then I can keep it alive.

KevinOConnor:
If I understand your code correctly, you have a "multiplex_adc" which has a pointer to an "hx71x" and has a pointer to a "loadcell_endstop".  The "loadcell_endstop" has  a pointer to a "trsync".

For what it is worth, I suspect "multiplex_adc" will not scale well.  It'll be difficult to get different sensor chips to conform to a single interface.

An alternate approach would be an "hx71x" that can read its own measurements, populate its own "sensor_bulk" reports, and communicate them directly to a host "hx71x.py" module.  That would allow the most flexibility in implementing future chip quirks.  In this alternate scheme, I'd imagine that "hx71x" would have a pointer to an "adc_endstop" (which in turn would have a pointer to a trsync and would determine when to trigger it).  For a short-term merge, the "adc_endstop" could be omitted.

It's difficult to make judgements on code structure, however.  I'm not entirely certain on all the requirements of load cells.

If you don't think the ADS1263 has immediate value, then I wouldn't spend time on it.

Cheers,
-Kevin

garethky:
Ok, I can go down that path, structurally, and we can see how it looks

PR #6433, opened 2023-12-17
Fixes to mpu9250

KevinOConnor:
I noticed a couple of errors in the mpu9250 code.

The first is the use of `time.sleep()` in the host code, which isn't valid - it can causes glitches in the host code processing and doesn't actually guarantee a pause at the mcu.

The second is incorrect timing in the command_query_mpu9250_status() implementation.  To get good measurement timing, that function needs to report the number of samples in the mpu9250 chip at the time of the call - so it should not use a cached value of the fifo level.

I don't have an actual mpu9250 (nor similar) so I can't test these changes directly.

@mattaw , @bluesforte - fyi.

-Kevin

### DIFF
--- a/klippy/extras/mpu9250.py
+++ b/klippy/extras/mpu9250.py
@@ -4,7 +4,7 @@
 # Copyright (C) 2020-2021 Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import logging, time
+import logging
 from . import bus, adxl345, bulk_sensor
 
 MPU9250_ADDR =      0x68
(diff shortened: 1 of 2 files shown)


mattaw:
Great you looked into it. I'll check it out this week.

KevinOConnor:
That's strange - the changes don't increase the number of i2c messages - the command_query_mpu9250_status() code was always sending an i2c message, it just wasn't using the results.

A useful debugging test would be to see which of the three commits in this PR is introducing the regression.  Each commit makes a minor change.

-Kevin

mattaw:
Firstly, I take back what I said about the patch introducing more I2C bus transactions, I'll need to recheck on the OSC what is going on properly. It might be a newer kernel version or similar that is causing the issue.

To be honest, my fix to the code was somewhat band-aiding it, and I never fully understood what klippy was doing with the bus messages. Could you flag the dev docs for (or spare 15 mins to talk me through) the accelerometer code path from klippy.py thru the MCU, and its algo? Then I could have a better stab at properly addressing the issue, maybe adding in a way of detecting failures and recovering etc.

KevinOConnor:
FYI, I added an additional change proposal to this PR.  The current code queries the INT_STATUS register and reports an overflow condition only upon stopping the sensor.  Unfortunately, the host code currently ignores messages after initiating a stop, so an overflow detected at that point doesn't get propagated to the ultimate user reports.

The additional change is to query the INT_STATUS register (and check for overflows) on every `query_mpu9250_status` command.  The host calls that command once every 100ms, so this change does introduce some additional i2c messages, but it should be a trivial increase.

-Kevin

KevinOConnor:
FYI, I have rebased this branch on top of the latest changes in the master branch (which now includes the merged PR #6432).

-Kevin

KevinOConnor:
@mattaw - do you think you will be able to check the code on this branch (and branch #6437)?  If not, could you recommend someone that could?

Thanks,
-Kevin

dockterj:
@KevinOConnor I was working on this a while ago but never got it working and haven't picked it up since Mattaw made his fixes.  I have an MPU6050 attached to an arduino nano and it seems to be working reliably now (I added pull up resistors).  The axes noise is high (xy around 270 and z around 500).  This is on a makerbot replicator 2 (cartesian).  I was getting i2c errors without the pullup resistors (I should make a PR to change the Klipper documentation to reference them).

orig files are with the main branch.  new files are from dropping in the 2 files in this PR to my environment.

They look close to me.  Let me know if there is more that I can test or verify.

shaper_calibrate_y-new
shaper_calibrate_x-new
shaper_calibrate_y-orig
shaper_calibrate_x-orig

KevinOConnor:
Thanks for testing!  If there are no further comments I'll look to commit in a couple of days.

-Kevin

PR #6434, opened 2023-12-18
Fixes for STM32G4 devices

eamaclean:
The STM32G4 support never had USB or CAN tested, and changes to the STM32H7 ADC code in the meantime revealed a timing issue seen only as a small footnote in the reference manual.

Thanks to @mattthebaker for finding the USB issue first

### DIFF
--- a/src/stm32/fdcan.c
+++ b/src/stm32/fdcan.c
@@ -162,10 +162,10 @@ canhw_set_filter(uint32_t id)
     can_filter(1, id);
     can_filter(2, id + 1);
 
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Thanks.  Looks fine to me.  I'll give a few days to see if there are other comments.

-Kevin

mattthebaker:
I can confirm from hardware tests that both the USB and ADC changes are
effective and necessary.

Matt

KevinOConnor:
Thanks.

-Kevin

PR #6435, opened 2023-12-20
filament_switch_sensor: send insert and runout events

mdvorak:
Filament sensors are nowadays used in a way more places, than just for runout detection. MMU units are good example. 
Firing events would allow additional software to reuse this class, without resolving for hacks like https://github.com/moggieuk/Happy-Hare/blob/main/extras/mmu_sensors.py#L136

This PR does not change current behavior.

Additionally, I would consider adding a new config parameter, allowing `insert_gcode` and `runout_gcode` to be executed regardless of printing status.

### DIFF
--- a/klippy/extras/filament_switch_sensor.py
+++ b/klippy/extras/filament_switch_sensor.py
@@ -74,20 +74,33 @@ class RunoutHelper:
         is_printing = idle_timeout.get_status(eventtime)["state"] == "Printing"
         # Perform filament action associated with status change (if any)
         if is_filament_present:
+            # insert detected
+            self.min_event_systime = self.reactor.NEVER
+            logging.info(
+                "Filament Sensor %s: insert event detected, Time %.2f" %
+                (self.name, eventtime))
+            self.printer.send_event("filament:insert", eventtime, self.name)
+
             if not is_printing and self.insert_gcode is not None:
-                # insert detected
-                self.min_event_systime = self.reactor.NEVER
-                logging.info(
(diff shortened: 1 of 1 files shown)


JamesH1978:

Thanks
James

Arksine:
Hi.  After an initial review it appears to me that this PR can change behavior of sensors that depend on the RunoutHelper.  Currently, `self.min_event_systime` is only reset after runout/insert gcode is processed.  This PR resets it after each change in state.  This could cause an event to be skipped that would otherwise be detected.

> Firing events would allow additional software to reuse this class, without resolving for hacks like https://github.com/moggieuk/Happy-Hare/blob/main/extras/mmu_sensors.py#L136

Currently Klipper doesn't have official support for 3rd party extensions.  This is something Kevin is [looking to add](https://klipper.discourse.group/t/klipper-development-goals-for-2024/12808/1) in 2024.  IMO extensions really need a stable API with semantic versioning.  Using the example of the Happy Hare, even if this were to be merged the 3rd party extension would have to verify that the installed version of Klipper supports these events and fallback if it doesn't.  It would likely make more sense to exclusively use the workaround rather than support both.

> Additionally, I would consider adding a new config parameter, allowing insert_gcode and runout_gcode to be executed regardless of printing status.

I would be ok with this change.

mdvorak:
Ah, I missed that change in behavior. However if you'll agree with the alternative implementation, I'll prepare new PR. It is not so "clean", but it will fit in stable configuration API.

Thanks!  Looking forward to extension support.

PR #6436, opened 2023-12-21
Extend sermoon v1 configuration example

arddennis:
Repo has only core definition for the printer and missing filament sensors. This change adds both filament and optional door sensor. Idea is to stop printing when filament runs out or door being open.
LED pin added to the config to be able to turn LED light of the printer.

### DIFF
--- a/config/printer-creality-sermoonV1-2022.cfg
+++ b/config/printer-creality-sermoonV1-2022.cfg
@@ -94,6 +94,23 @@ pin: PC15
 # for more information.
 # pin: PB4
 
+[output_pin LED_pin]
+pin: PC14
+value: 0
+
+[filament_switch_sensor RunoutSensor]
+pause_on_runout: True
+runout_gcode: PAUSE
+insert_gcode: RESUME
+switch_pin: PA3
+
+# Sermoon V1 Pro has additional DoorSensor
+# [filament_switch_sensor DoorSensor]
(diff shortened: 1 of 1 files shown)


JamesH1978:

Thanks
James

JamesH1978:
@duchengyao Anything to add to this? as this was your original PR?

If no objections I will review and likely approve shortly

Thank
James

arddennis:
@JamesH1978 Sorry to ask an obvious question. 

I just saw another Sermoon V1 Pro printer using a different pin for a runout sensor. What is the best practice to explain these details? Should I add a comment with instructions on how to test it and possible other options? Alternative settings is: switch_pin: !PC6 instead of PA3 on mine.

Also, there is a cr/bl-touch sensor connection outlet on the printer available. The sensor is not bundled with the printer by-default, but I managed to connect it and it works. Should I note these details as well in this commit or send you another PR to include this documentation as a comment in the text. Such could save me some time when I searched for the info...

JamesH1978:
We do not include instructions to do testing in configurations. We only include options for a stock machine. 

If connecions are available on the board for probes with clearly defined pins then these can be included in the configuration but commented out.

This is the same for a run out sensor if it is stock and comes with a printer then the pin should not be different. This does not stop someone buying this printer and adding a 3rd party sensor and using a different pin but that is outside the scope of these configurations.

PR #6438, opened 2023-12-27
menu: Respect position_min for z axis.

bindertron:
With this change the move menu items for the Z axis behave the same as the ones for X and Y.


jakep82:
I don't agree with this change. Many printers use a negative position_min value on the Z axis for purposes of homing or generating a mesh. On both of my printers it's not safe to go below zero other than during homing or meshing. I think many printers are configured in a similar way.

jakep82:
One other note, for those that want their printer configured this way this can be done by adding these lines to their printer.cfg. The default doesn't need to be changed to achieve this functionality.

PR #6439, opened 2023-12-27
Keep stepcompress move history relative to current time

EiNSTeiN-:
Currently the stepcompress history is removed as soon as it's older than 30 seconds before the furthest move in the stepcompress queue. If the furthest move is long and will end more than 30 seconds in the future, then the stepcompress history will return a "future" stepper position rather than a past position.

I found this issue while trying to graph the output of `extruder.find_past_position(print_time)` with a 45mm move at 1.33mm/s, which puts it around 33.8 seconds, and it would give me a graph like this one:

image

Note the purple line (the expected extruder position) is constant for the first ~4 seconds of the move rather than providing the expected stepper position. With this fix applied, the graph is as expected:

image

This was reported, for example, in: https://klipper.discourse.group/t/filament-sensor-triggers-on-long-g0-e-command/3771. It's hard to track real world problems which can be caused by flushing the history too early but I've personally observed false positives with my BTT SFS v1 which might be related t this. I have a 350mm printer so long moves on top/bottom layers are to be expected when printing large parts at slow speeds, as we do with the first layer and the top infill.

### DIFF
--- a/klippy/chelper/stepcompress.c
+++ b/klippy/chelper/stepcompress.c
@@ -734,7 +734,7 @@ steppersync_set_time(struct steppersync *ss, double time_offset
 }
 
 // Expire the stepcompress history before the given clock time
-void __visible
+static void
 steppersync_history_expire(struct steppersync *ss, uint64_t end_clock)
 {
     int i;
--- a/klippy/chelper/stepcompress.c
+++ b/klippy/chelper/stepcompress.c
@@ -54,8 +54,6 @@ struct step_move {
     int16_t add;
 };
 
-#define HISTORY_EXPIRE (30.0)
-
 struct history_steps {
     struct list_node node;
     uint64_t first_clock, last_clock;
@@ -292,6 +290,13 @@ free_history(struct stepcompress *sc, uint64_t end_clock)
     }
 }
 
+// Expire the stepcompress history older than the given clock
+static void
+stepcompress_history_expire(struct stepcompress *sc, uint64_t end_clock)
+{
+    free_history(sc, end_clock);
+}
+
 // Free memory associated with a 'stepcompress' object
 void __visible
 stepcompress_free(struct stepcompress *sc)
@@ -322,9 +327,6 @@ calc_last_step_print_time(struct stepcompress *sc)
 {
     double lsc = sc->last_step_clock;
     sc->last_step_print_time = sc->mcu_time_offset + (lsc - .5) / sc->mcu_freq;
-
-    if (lsc > sc->mcu_freq * HISTORY_EXPIRE)
-        free_history(sc, lsc - sc->mcu_freq * HISTORY_EXPIRE);
 }
 
 // Set the conversion rate of 'print_time' to mcu clock
@@ -731,6 +733,18 @@ steppersync_set_time(struct steppersync *ss, double time_offset
     }
 }
 
+// Expire the stepcompress history before the given clock time
+void __visible
+steppersync_history_expire(struct steppersync *ss, uint64_t end_clock)
+{
+    int i;
+    for (i = 0; i < ss->sc_num; i++)
+    {
+        struct stepcompress *sc = ss->sc_list[i];
+        stepcompress_history_expire(sc, end_clock);
+    }
+}
+
 // Implement a binary heap algorithm to track when the next available
 // 'struct move' in the mcu will be available
 static void
@@ -758,7 +772,8 @@ heap_replace(struct steppersync *ss, uint64_t req_clock)
 
 // Find and transmit any scheduled steps prior to the given 'move_clock'
 int __visible
-steppersync_flush(struct steppersync *ss, uint64_t move_clock)
+steppersync_flush(struct steppersync *ss, uint64_t move_clock
+                  , uint64_t clear_history_clock)
 {
     // Flush each stepcompress to the specified move_clock
     int i;
@@ -806,5 +821,7 @@ steppersync_flush(struct steppersync *ss, uint64_t move_clock)
     // Transmit commands
     if (!list_empty(&msgs))
         serialqueue_send_batch(ss->sq, ss->cq, &msgs);
+
+    steppersync_history_expire(ss, clear_history_clock);
     return 0;
 }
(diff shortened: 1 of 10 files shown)


klippy/chelper/stepcompress.c
```diff
@@ -731,6 +733,18 @@ steppersync_set_time(struct steppersync *ss, double time_offset
     }
 }
 
+// Expire the stepcompress history before the given clock time
+void __visible
+steppersync_history_expire(struct steppersync *ss, uint64_t end_clock)
```
KevinOConnor, line 738: Should be `static`.

klippy/chelper/trapq.c
```diff
@@ -206,7 +204,7 @@ trapq_set_position(struct trapq *tq, double print_time
                    , double pos_x, double pos_y, double pos_z)
 {
     // Flush all moves from trapq
-    trapq_finalize_moves(tq, NEVER_TIME);
+    trapq_finalize_moves(tq, NEVER_TIME, 0);
```
KevinOConnor, line 207: This would introduce a subtle change in behavior.  All places that fully cleared the history should continue to fully clear the history - thus `trapq_finalize_moves(tq, NEVER_TIME, NEVER_TIME);`

klippy/extras/force_move.py
```diff
@@ -86,7 +86,7 @@ def manual_move(self, stepper, dist, speed, accel=0.):
                           0., 0., 0., axis_r, 0., 0., 0., cruise_v, accel)
         print_time = print_time + accel_t + cruise_t + accel_t
         stepper.generate_steps(print_time)
-        self.trapq_finalize_moves(self.trapq, print_time + 99999.9)
+        self.trapq_finalize_moves(self.trapq, print_time + 99999.9, 0)
```
KevinOConnor, line 89: Same here.

klippy/extras/manual_stepper.py
```diff
@@ -67,7 +67,7 @@ def do_move(self, movepos, speed, accel, sync=True):
                           0., cruise_v, accel)
         self.next_cmd_time = self.next_cmd_time + accel_t + cruise_t + accel_t
         self.rail.generate_steps(self.next_cmd_time)
-        self.trapq_finalize_moves(self.trapq, self.next_cmd_time + 99999.9)
+        self.trapq_finalize_moves(self.trapq, self.next_cmd_time + 99999.9, 0)
```
KevinOConnor, line 70: Same here.

klippy/toolhead.py
```diff
@@ -522,7 +524,9 @@ def drip_move(self, newpos, speed, drip_completion):
             self.move_queue.flush()
         except DripModeEndSignal as e:
             self.move_queue.reset()
-            self.trapq_finalize_moves(self.trapq, self.reactor.NEVER)
+            clear_history_time = self.reactor.monotonic() - MOVE_HISTORY_EXPIRE
+            self.trapq_finalize_moves(self.trapq, self.reactor.NEVER,
+                                      clear_history_time)
```
KevinOConnor, line 529: Same here.

KevinOConnor (commented):
Thanks for working on this!  In general, it looks good to me.  See some minor comments below.

-Kevin

klippy/toolhead.py
```diff
@@ -289,13 +290,14 @@ def _advance_flush_time(self, flush_time):
         for sg in self.step_generators:
             sg(sg_flush_time)
         self.last_sg_flush_time = sg_flush_time
+        clear_history_time = self.reactor.monotonic() - MOVE_HISTORY_EXPIRE
```
KevinOConnor, line 293: The trapq times are relative to the "print_time", not the "system time".  To go from a system time to a print_time use `self.mcu.estimated_print_time(systime)`.

EiNSTeiN-: Done, I made everything a print time including the time provided to `flush_moves` and convert using `print_time_to_clock` rather than `get_clock`.

KevinOConnor:
Okay, thanks.  However, the flushing in `DripModeEndSignal` was not corrected.  Also, it is confusing to use two different mechanisms to implement a full flush in force_move.py and manual_stepper.py (it would be preferable to consistently use either use "+ 99999.9" or "reactor.NEVER").

-Kevin

klippy/chelper/trapq.c
```diff
@@ -190,10 +189,9 @@ trapq_finalize_moves(struct trapq *tq, double print_time)
     if (list_empty(&tq->history))
         return;
     struct move *latest = list_first_entry(&tq->history, struct move, node);
-    double expire_time = latest->print_time + latest->move_t - HISTORY_EXPIRE;
```
EiNSTeiN-, line 193: Just for the record I think we're now clearing the history in places where we wouldn't have before which is a change in behavior. When `print_time` is `NEVER_TIME` (or a large value) all the moves would get transferred to the history, but this code would still only clear 30s prior to the last move because of the math on this line. This is why I had the new parameter as 0 initially (to keep the history until this is called from the toolhead code).

EiNSTeiN-:
I made the last changes, this should be good to go if you're OK with my other comment

klippy/chelper/trapq.c
KevinOConnor: Okay, I goofed.  You are correct and I misread the original code.  I think we do want to avoid a change in behavior here, because this is an area of the code where we really want to be able to use tools like `git bisect` if necessary.

If I'm reading the code correctly, the call in `trapq_finalize_moves()` doesn't need to flush the history (it can use `0`), and the code in `DripModeEndSignal` doesn't need to flush (it can also use `0`).  However, the code in manual_stepper.py and force_move.py do need to fully flush (using the current `+ 99999.9` method) as nothing else will free the history (and nothing uses the history of these modules).

Does that sound correct?

Thanks.
-Kevin

EiNSTeiN-: If you mean `trapq_set_position` and not `trapq_finalize_moves` that looks correct to me (as in, it preserves the current behavior), however I'm not super familiar with this code. I left the other 2 calls with `+ 99999.9` to clear the history.

KevinOConnor:
Thanks.  I committed this PR along with commit 9847b449 on top.

-Kevin

KevinOConnor:
I just realized that the new history flushing code would not do the right thing when in a debugging "batch mode" ( https://www.klipper3d.org/Debugging.html#translating-gcode-files-to-micro-controller-commands ).  I pushed commit 25bc649c to address this.

-Kevin

PR #6441, opened 2023-12-29
Make TRSYNC_TIMEOUT configurable

EiNSTeiN-:
I have a 5-toolhead printer, each toolhead is equipped with BTT SB2209 connected to the main mcu over canbus. Each has its own optical endstop (Voron Tap PCB) connected to the toolhead board. A little while ago I kept getting failed homing and following the solution in this post solved the issue for me: https://klipper.discourse.group/t/canbus-communication-timeout-while-homing-z/3741 (and similarly: https://github.com/bigtreetech/EBB/issues/3).

> Changed TRSYNC_TIMEOUT value in “/home/pi/klipper/klippy/mcu.py” file from 0.025 to 0.050, i.e. was “TRSYNC_TIMEOUT = 0.025”, and became “TRSYNC_TIMEOUT = 0.050”, and the error “Communication timeout during homing probe” disappeared

Seems like this timeout value is used only in the homing code in `home_start` which explains why there are no communication errors at any other time. I've been running a version of klipper with this local change for a while but I'd like to keep getting updates, so I'm opening this PR to make the timeout value configurable at the MCU level, and default to the old value since this seems to work fine for most boards.

I'm also open to writing alternative fixes, perhaps just doubling the timeout when the mcu is connected over canbus would be better?

### DIFF
--- a/klippy/mcu.py
+++ b/klippy/mcu.py
@@ -280,9 +280,9 @@ class MCU_endstop:
         self._rest_ticks = rest_ticks
         reactor = self._mcu.get_printer().get_reactor()
         self._trigger_completion = reactor.completion()
-        expire_timeout = TRSYNC_TIMEOUT
+        expire_timeout = self._mcu.get_homing_trsync_timeout()
         if len(self._trsyncs) == 1:
-            expire_timeout = TRSYNC_SINGLE_MCU_TIMEOUT
+            expire_timeout = self._mcu.get_homing_trsync_single_mcu_timeout()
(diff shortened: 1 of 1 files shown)


jakep82:
This a duplicate of an open PR (#6403), and 2 other rejected PRs.  

#5951
#5816

EiNSTeiN-:
Thanks and sorry for not searching first! I'll see if I can find the root cause and keep the conversation going on discourse.

PR #6449, opened 2024-01-04
Add  36KiB bootloader offset option

JakubPrzystasz:
This offset is used by Anycubic Kobra 2 Neo bootloader


JamesH1978:

Thanks
James

KevinOConnor:
Thanks.  I committed this change and committed 2e8b54ae on top.  Going forward I think it would be best if we did not list product names in the "make menuconfig" menus.

Cheers,
-Kevin

PR #6451, opened 2024-01-07
filament_switch_sensor: add option to always execute gcode handlers

mdvorak:
Added `run_always` config option as a follow-up to the #6435.

I'm open for better name for the variable.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4328,6 +4328,10 @@ more information.
 #   detected. See docs/Command_Templates.md for G-Code format. The
 #   default is not to run any G-Code commands, which disables insert
 #   detection.
+#run_always: False
+#  When false, runout_gcode is executed only during print, and
+#  insert_gcode only when out of print. Set to true to execute
+#  gcode handlers regardless of printing status. Default is False.
 #event_delay: 3.0
 #   The minimum amount of time in seconds to delay between events.
 #   Events triggered during this time period will be silently
--- a/klippy/extras/filament_switch_sensor.py
+++ b/klippy/extras/filament_switch_sensor.py
@@ -23,6 +23,7 @@ class RunoutHelper:
         if config.get('insert_gcode', None) is not None:
             self.insert_gcode = gcode_macro.load_template(
                 config, 'insert_gcode')
+        self.run_always = config.getboolean('run_always', False)
         self.pause_delay = config.getfloat('pause_delay', .5, above=.0)
         self.event_delay = config.getfloat('event_delay', 3., above=0.)
         # Internal state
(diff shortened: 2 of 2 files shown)


Zeanon:
maybe this is a dumb idea, but why not do something like this:
```
        if is_filament_present == self.filament_present and not force:
            return
        self.filament_present = is_filament_present
        if eventtime < self.min_event_systime:
            # do not process during the initialization time, duplicates,
            # during the event delay time, while an event is running, or
            # when the sensor is disabled
            return
        if is_filament_present:
            self.printer.send_event("filament:insert", eventtime, self.name)
        else:
            self.printer.send_event("filament:runout", eventtime, self.name)
        if not self.sensor_enabled:
            return
   ```
   That would fire the event (in this case even when the sensor is disabled(reason being, a lot of people prefer to keep the sensors disabled cause for some odd reason the runout gcode sometimes tends to fire even though the printer is not actually in a printing state)) without changing any behaviour at all, except for firing the event

Arksine (approved):

Arksine:
> That would fire the event (in this case even when the sensor is disabled(reason being, a lot of people prefer to keep the sensors disabled cause for some odd reason the runout gcode sometimes tends to fire even though the printer is not actually in a printing state)) without changing any behaviour at all, except for firing the event

The runout gcode cannot fire when `idle_timeout` reports `Printing`.  The `idle_timeout` will report `Printing` if the printer is in any way busy, such as a stepper moving.   Most likely some individuals confusing this with the `print_stats` printing state which is tied to the `virtual_sdcard`.  The sensor has to keep compatibility with print sources over the pty, such as Octoprint, so we cannot depend on the latter.  This is why its possible to enable/disable sensors via gcode.

Regardless, I don't think adding printer events should be a part of this PR, as they may block merging.  The PR as-is looks good to me.

KevinOConnor:
Thanks.  We can certainly commit this change.

For what it is worth, though, I did not understand the documentation for this new parameter.

As a user, when would I set this new config option, and what real world tasks can I now achieve after setting it?

-Kevin

mdvorak:
Most real-world examples are probably related to multi-material prints (generally, not only ERCF).

tphan26363:
The filament switch detection routine is sometime over detection.  This happen even the switch only open in a very short time(us I guest) depend on how fast the MCU or the execution is.  Vibration some time can open the switch even with the filament in in a very short time causing the false trigger.  This especially happen to the bad quality switches.   The routine should have to detect long pulse since the mechanical switch should open for a long time when the filament run out.  1 second long pulse has to be a minimum requirement for the detection to be valid.  Howe ever this can also be fix mechanically just by revert the state of the switch when filament is present to " Open" instead of "Close".  Since when filament in, any vibration can not close the switch no matter how violent it is.  This require to revert the state of the pin definition and I found this to be extremely reliable solution.
For the "Motion Sensor"  I can not do much but I believe a complex "insert Gcode" when trigger can change the way of motion detect more reliable.  The way motion detection right now with be not reliable.  Just by printing at very slow speed can generate false triggering.  I believe a more intelligent detection need to be implement.  One thing to make sure that when the extruder stop, the motion sensor should not doing detection during the time extruder not moving the filament and when the extruder moves the filament,  the detection should start working.  Retraction should be ignore since the movement is to short for the motion sensor to feel it.  Any way the current routine will work reliably if you are printing large object at a relative fast speed and that is the only case that motion sensor should be use for.

PR #6452, opened 2024-01-08
Added QuadBedLevel feature

Phil1988:
New function for moving beds with 4 independent moving z axes.
It's based/derived on quad_bed_level but its task is to flatten the level the bed by moving the individual axes.

I did several successfull tests and this feature is able to flatten/level beds of any shapes like ones with a hyperbolic parabola effect (potato chip).

Some examples:
Bed before:
image

Bed after 3 iterations of leveling (took ~6min)
image

Here is the console output during a quad_bed_level:
```
16:24:52
Retries: 3/5 Probed points range: 0.006000 tolerance: 0.010000
16:24:49
Making the following Z adjustments:
stepper_z = -0.003839
stepper_z1 = 0.001486
stepper_z2 = 0.003822
stepper_z3 = -0.001469
16:24:49
Adjustment positions:
z: -0.003839 z1: 0.001486 z2: 0.003822 z3: -0.001469
16:24:49
Average: 16.647960
16:24:49
Actuator Positions:
z: 16.651799 z1: 16.646474 z2: 16.644138 z3: 16.649429
16:24:49
Bed-relative probe points:
0: 16.651242 1: 16.646742 2: 16.645242 3: 16.649742
16:24:49
probe at 540.000,60.000 is z=3.350258
16:24:36
probe at 540.000,600.000 is z=3.354758
16:24:23
probe at 50.000,600.000 is z=3.353258
16:24:10
probe at 50.000,60.000 is z=3.348758
16:24:01
Retries: 2/5 Probed points range: 0.029000 tolerance: 0.010000
16:23:58
Making the following Z adjustments:
stepper_z = 0.027657
stepper_z1 = -0.018258
stepper_z2 = -0.000115
stepper_z3 = -0.009283
16:23:58
Adjustment positions:
z: 0.027657 z1: -0.018258 z2: -0.000115 z3: -0.009283
16:23:58
Average: 16.648247
16:23:57
Actuator Positions:
z: 16.620591 z1: 16.666506 z2: 16.648362 z3: 16.657531
16:23:57
Bed-relative probe points:
0: 16.628740 1: 16.657740 2: 16.651240 3: 16.651740
16:23:57
probe at 540.000,60.000 is z=3.348260
16:23:44
probe at 540.000,600.000 is z=3.348760
16:23:32
probe at 50.000,600.000 is z=3.342260
16:23:19
probe at 50.000,60.000 is z=3.371260
16:23:10
Retries: 1/5 Probed points range: 0.121000 tolerance: 0.010000
16:23:06
Making the following Z adjustments:
stepper_z = 0.089295
stepper_z1 = -0.095760
stepper_z2 = 0.010195
stepper_z3 = -0.003730
16:23:06
Adjustment positions:
z: 0.089295 z1: -0.095760 z2: 0.010195 z3: -0.003730
16:23:06
Average: 16.649476
16:23:06
Actuator Positions:
z: 16.560181 z1: 16.745237 z2: 16.639282 z3: 16.653206
16:23:06
Bed-relative probe points:
0: 16.581330 1: 16.702330 2: 16.653330 3: 16.638830
16:23:06
probe at 540.000,60.000 is z=3.361170
16:22:53
probe at 540.000,600.000 is z=3.346670
16:22:40
probe at 50.000,600.000 is z=3.297670
16:22:27
probe at 50.000,60.000 is z=3.418670
16:22:18
Retries: 0/5 Probed points range: 1.017500 tolerance: 0.010000
16:22:14
Making the following Z adjustments:
stepper_z = 0.269584
stepper_z1 = -1.088170
stepper_z2 = 0.395465
stepper_z3 = 0.423121
16:22:14
Adjustment positions:
z: 0.269584 z1: -1.088170 z2: 0.395465 z3: 0.423121
16:22:14
Average: 16.643972
16:22:14
Actuator Positions:
z: 16.374388 z1: 17.732142 z2: 16.248507 z3: 16.220851
16:22:14
Bed-relative probe points:
0: 16.354000 1: 17.265000 2: 16.447000 3: 16.247500
16:22:14
probe at 540.000,60.000 is z=3.752500
16:22:01
probe at 540.000,600.000 is z=3.553000
16:21:48
probe at 50.000,600.000 is z=2.735000
16:21:35
probe at 50.000,60.000 is z=3.646000
16:21:29
quad_bed_level
```

It started from `points range: 1.017500 `
after 3 iterations it went down to `points range: 0.006000`

Configuration is similar to "[quad_gantry_level]"

So the section would look like this:

```
[quad_bed_level]
#bed_corners:
# X/Y coordinates describing the location of each bed "pivot point".
# These points beeing measured from the coordinates of the nozzle
# This parameter must be provided
#points:
# X/Y coordinates of each probe point during a Quad_bed_level command.
# The location is measured from the nozzle coordinates, so make sure to adjust the coordinate according to the probe offset
#speed: 150
#   The speed (in mm/s) of non-probing moves during the calibration.
#   The default is 50.
#horizontal_move_z: 30
#   The height (in mm) that the head should be commanded to move to
#   just prior to starting a probe operation. The default is 5.
#max_adjust: 7
#   Safety limit if an adjustment greater than this value is requested
#   quad_gantry_level will abort.
#retries: 5
#   Number of times to retry if the probed points aren't within
#   tolerance.
#retry_tolerance: 0.010
#   If retries are enabled then retry if largest and smallest probed
#   points differ more than retry_tolerance.
```

### DIFF
--- a/klippy/extras/quad_bed_level.py
+++ b/klippy/extras/quad_bed_level.py
@@ -1,0 +1,136 @@
+# Mechanicaly conforms a moving bed to the bed with 4 Z steppers
+#
+# Copyright (C) 2018  Maks Zolin <mzolin@vorondesign.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging
+from . import probe, z_tilt
+
+# Leveling code for XY rails that are controlled by Z steppers as in:
+#
+# Z stepper1 ----> O                             O <---- Z stepper2
+#                  | * <-- probe1   probe2 --> * |
+#                  |                             |
+#                  |                             | <--- Y2 rail
+#   Y1 rail -----> |                             |
+#                  |                             |
+#                  |=============================|
+#                  |            ^                |
+#                  |            |                |
+#                  |   X rail --/                |
+#                  |                             |
+#                  | * <-- probe0   probe3 --> * |
+# Z stepper0 ----> O                             O <---- Z stepper3
+
+class QuadBedLevel:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.retry_helper = z_tilt.RetryHelper(config,
+            "Possibly Z motor numbering is wrong")
+        self.max_adjust = config.getfloat("max_adjust", 4, above=0)
+        self.horizontal_move_z = config.getfloat("horizontal_move_z", 5.0)
+        self.probe_helper = probe.ProbePointsHelper(config, self.probe_finalize)
+        if len(self.probe_helper.probe_points) != 4:
+            raise config.error(
+                "Need exactly 4 probe points for quad_bed_level")
+        self.z_status = z_tilt.ZAdjustStatus(self.printer)
+        self.z_helper = z_tilt.ZAdjustHelper(config, 4)
+        self.bed_corners = config.getlists('bed_corners', parser=float,
+                                              seps=(',', '\n'), count=2)
+        if len(self.bed_corners) < 2:
+            raise config.error(
+                "quad_bed_level requires at least two bed_corners")
+        # Register QUAD_BED_LEVEL command
+        self.gcode = self.printer.lookup_object('gcode')
+        self.gcode.register_command(
+            'QUAD_BED_LEVEL', self.cmd_QUAD_BED_LEVEL,
+            desc=self.cmd_QUAD_BED_LEVEL_help)
+    cmd_QUAD_BED_LEVEL_help = (
+        "Bend a bed to the shape of the gantry")
+    def cmd_QUAD_BED_LEVEL(self, gcmd):
+        self.z_status.reset()
+        self.retry_helper.start(gcmd)
+        self.probe_helper.start_probe(gcmd)
+    def probe_finalize(self, offsets, positions):
+        # Mirror our perspective so the adjustments make sense
+        # from the perspective of the bed
+        z_positions = [self.horizontal_move_z - p[2] for p in positions]
+        points_message = "Bed-relative probe points:\n%s\n" % (
+            " ".join(["%s: %.6f" % (z_id, z_positions[z_id])
+                for z_id in range(len(z_positions))]))
+        self.gcode.respond_info(points_message)
+        # Calculate slope along X axis between probe point 0 and 3
+        ppx0 = [positions[0][0] + offsets[0], z_positions[0]]
+        ppx3 = [positions[3][0] + offsets[0], z_positions[3]]
+        slope_x_pp03 = self.linefit(ppx0, ppx3)
+        # Calculate slope along X axis between probe point 1 and 2
+        ppx1 = [positions[1][0] + offsets[0], z_positions[1]]
+        ppx2 = [positions[2][0] + offsets[0], z_positions[2]]
+        slope_x_pp12 = self.linefit(ppx1, ppx2)
+        logging.info("quad_bed_level f1: %s, f2: %s"
+                     % (slope_x_pp03, slope_x_pp12))
+        # Calculate bed slope along Y axis between stepper 0 and 1
+        a1 = [positions[0][1] + offsets[1],
+              self.plot(slope_x_pp03, self.bed_corners[0][0])]
+        a2 = [positions[1][1] + offsets[1],
+              self.plot(slope_x_pp12, self.bed_corners[1][0])]
+        slope_y_s01 = self.linefit(a1, a2)
+        # Calculate bed slope along Y axis between stepper 2 and 3
+        b1 = [positions[3][1] + offsets[1],
+              self.plot(slope_x_pp03, self.bed_corners[3][0])]
+        b2 = [positions[2][1] + offsets[1],
+              self.plot(slope_x_pp12, self.bed_corners[2][0])]
+        slope_y_s32 = self.linefit(b1, b2)
+        logging.info("quad_bed_level af: %s, bf: %s"
+                     % (slope_y_s01, slope_y_s32))
+		# Calculate z height of each stepper
+        z_height = [0,0,0,0]
+        z_height[0] = (1) * self.plot(slope_y_s01, self.bed_corners[0][1])
+        z_height[1] = (1) * self.plot(slope_y_s01, self.bed_corners[1][1])
+        z_height[2] = (1) * self.plot(slope_y_s32, self.bed_corners[2][1])
+        z_height[3] = (1) * self.plot(slope_y_s32, self.bed_corners[3][1])
+
+        ainfo = zip(["z","z1","z2","z3"], z_height[0:4])
+        apos = " ".join(["%s: %06f" % (x) for x in ainfo])
+        self.gcode.respond_info("Actuator Positions:\n" + apos)
+
+        # take average height
+        z_ave = sum(z_height) / len(z_height)
+        self.gcode.respond_info("Average: %0.6f" % z_ave)
+        # adjust all steppers to the average height
+        z_adjust = []
+        for z in z_height:
+            z_adjust.append(z_ave - z)
+			
+        adjustinfo = zip(["z","z1","z2","z3"], z_adjust[0:4])
+        adjustpos = " ".join(["%s: %06f" % (x) for x in adjustinfo])
+
+        self.gcode.respond_info("Adjustment positions:\n" + adjustpos)
+
+        adjust_max = max(z_adjust)
+        if adjust_max > self.max_adjust:
+            raise self.gcode.error("Aborting quad_bed_level"
+                                   " required adjustment %0.6f"
+                                   " is greater than max_adjust %0.6f"
+                                   % (adjust_max, self.max_adjust))
+
+        speed = self.probe_helper.get_lift_speed()
+        self.z_helper.adjust_steppers(z_adjust, speed)
+        return self.z_status.check_retry_result(
+            self.retry_helper.check_retry(z_positions))
+
+    def linefit(self,p1,p2):
+        if p1[1] == p2[1]:
+            # Straight line
+            return 0,p1[1]
+        m = (p2[1] - p1[1])/(p2[0] - p1[0])
+        b = p1[1] - m * p1[0]
+        return m,b
+    def plot(self,f,x):
+        return f[0]*x + f[1]
+    def get_status(self, eventtime):
+        return self.z_status.get_status(eventtime)
+
+def load_config(config):
+    return QuadBedLevel(config)
+


JamesH1978:
Thank you for submitting a PR, please refer to point 3 in "What to expect in a review" in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and provide a signed off by line.

Thanks
James

Phil1988:
Thanks @JamesH1978 for your comment.
Added the Signed-off-by line.

Is it sufficient or am I missing something?

devond76180:
Looks like a useful feature,  will need test it out.  Can you add the files for command documentation.  Thanks.

Phil1988:
>  Can you add the files for command documentation. Thanks.

Hey @devond76180 , thanks for your interest.
What files do you exactly mean?

devond76180:
What are the parameters for the command?  Is it the same as quad_gantry_level?  The file would be Config_Reference.md with quadbedlevel added.Sent from my Galaxy
-------- Original message --------From: Phil1988 ***@***.***> Date: 1/25/24  1:49 PM  (GMT-05:00) To: Klipper3d/klipper ***@***.***> Cc: Devon Davis ***@***.***>, Mention ***@***.***> Subject: Re: [Klipper3d/klipper] Added QuadBedLevel feature (PR #6452) 

Can you add the files for command documentation. Thanks.

Hey @devond76180 , thanks for your interest.
What files do you exactly mean?

Phil1988:
@devond76180 : Please have a look at the config section in my first comment here.
Its all you need, but if you have questions, please reach out to me :)

madnezzm:
bed_corners parameters same as quad_gantry_level.gantry_corners? Can you show example?

Phil1988:
@madnezzm yes, its similar.
Here an example:

```
[quad_bed_level]
bed_corners:
# X/Y coordinates describing the location of each bed "pivot point".
# These points beeing measured from the coordinates of the nozzle
     -170, -5
     -170, 634
     604, 630
     604, -5
points:
# X/Y coordinates of each probe point during a Quad_bed_level command.
# The location is measured from the nozzle coordinates, so make sure to adjust the coordinate according to the probe offset
     50, 60
     50, 600
     540, 600
     540, 60
speed: 150
#   The speed (in mm/s) of non-probing moves during the calibration.
#   The default is 50.
horizontal_move_z: 20
#   The height (in mm) that the head should be commanded to move to
#   just prior to starting a probe operation. The default is 5.
max_adjust: 7
#   Safety limit if an adjustment greater than this value is requested
#   quad_gantry_level will abort.
retries: 5
#   Number of times to retry if the probed points aren't within
#   tolerance.
retry_tolerance: 0.010
#   If retries are enabled then retry if largest and smallest probed
#   points differ more than retry_tolerance.
```

clearchris:
I'm currently running this code on two machines, and I really like it.  In the coming weeks, I'll likely be depolying it to at least 2 more.  It's of particular use on larger beds as it's easier for them to get twisted out of plane.  Even on my machine, which has a 9.5mm thick bed, it's pretty easy for the z axis to twist the bed.

So far in testing, one machine went from 0.581mm bed range to 0.395mm range, big improvement IMHO.  But even more importantly, this will lead to increased consistency because any bed twist throws off the bed_mesh.  Having the probe points bent back into place before each print assures the mesh actually works for the bed.  The second machine didn't see much improvement, the bed was pretty flat to begin with.  However this module ensures that the bed will return to the same position prior to print.

What's really nice about this has not many changes from code that has been run by voron users for a very long time now, so I don't expect many (if any) strange bugs to pop up in this well tested code.

**Thank you Phil for doing the work on this!**

Note the "Range" in the before and after pics.
Before:
unnamed-1

After (please forgive the floating bed, someone messed with the z-offset, and the machine was turned off for the weekend):
unnamed

devond76180:
Tried the command.  got the following error.  Used the same config as quad_gantry_level.  Assume I did something bad in the config but not sure.
Send: quad_bed_level
Recv: // probe at 5.000,10.000 is z=1.391418
Recv: // probe at 5.000,580.000 is z=1.338918
Recv: // probe at 250.000,580.000 is z=1.361418
Recv: // probe at 250.000,10.000 is z=1.346418
Recv: // Bed-relative probe points:
Recv: // 0: 3.608582 1: 3.661082 2: 3.638582 3: 3.653582
Recv: // Klipper state: Shutdown
Recv: !! Internal error on command:"QUAD_BED_LEVEL"

Do you want the klippy.log and where do I send it?


Here is the short cut of the error from the log 

Bed-relative probe points:
0: 3.608582 1: 3.661082 2: 3.638582 3: 3.653582

quad_bed_level f1: (0.00018367346938663798, 3.601602818273141), f2: (-9.183673469394933e-05, 3.664572206028035)
Internal error on command:"QUAD_BED_LEVEL"
  File "/home/pi/klipper/klippy/gcode.py", line 198, in _process_commands
    handler(gcmd)
  File "/home/pi/klipper/klippy/gcode.py", line 135, in <lambda>
  File "/home/pi/klipper/klippy/extras/quad_bed_level.py", line 53, in cmd_QUAD_BED_LEVEL
  File "/home/pi/klipper/klippy/extras/probe.py", line 425, in start_probe
    done = self._move_next()
  File "/home/pi/klipper/klippy/extras/probe.py", line 394, in _move_next
    res = self.finalize_callback(self.probe_offsets, self.results)
  File "/home/pi/klipper/klippy/extras/quad_bed_level.py", line 80, in probe_finalize
    self.plot(slope_x_pp03, self.bed_corners[3][0])]
IndexError: tuple index out of range
Transition to shutdown state: Internal error on command:"QUAD_BED_LEVEL"

Phil1988:
Very little information on this one.
With these pieces I can not help.

devond76180:
The bed_corners and gantry_corners are very different.  Gantry is only z and z2 which should cover all needed information.  Bed_corners is z, z1, z2, z3.  If there are cases where Z1 and Z3 cannot be derived in the bed_level case, would this also be true for the gantry_level?

Phil1988:
Please show us your config :)

clearchris:
> The bed_corners and gantry_corners are very different. Gantry is only z and z2 which should cover all needed information. Bed_corners is z, z1, z2, z3. If there are cases where Z1 and Z3 cannot be derived in the bed_level case, would this also be true for the gantry_level?

You need to enter all 4 points, whether or not they can be derived.

Phil1988:
It would be great, if a reviewer can have a look and make this feature availible for others.
Already 3 different people showed interessed in this here and the results looking good to me.

Chris-Jayden:
I have a Anycubic Chiron into VORON Quadent conversion in the works, so this is a really exciting feature; I'd love to see it ported/supported in Klipper-main!~

nefelim4ag:
Sorry, maybe it is a silly question, but how does it differ from ztilt?

Phil1988:
> Sorry, maybe it is a silly question, but how does it differ from ztilt?

See my 2nd sentence:
> It is able to flatten/level beds of any shapes like ones with a hyperbolic parabola effect (potato chip) 

;)

ztilt only tilts. 
This moves the z position for all motors at the perfect height and does not only tilt, but level the bed.

Phil1988:
After 17month "pending" I guess this will never make it?
It is a usefull feature, but if klipper doesnt want to have it, it may be a better idea to delete this request instead of making it pending forever.
@KevinOConnor 

I am using it for 18month now without a single issue.

Wulfsta:
> After 17month "pending" I guess this will never make it? It is a usefull feature, but if klipper doesnt want to have it, it may be a better idea to delete this request instead of making it pending forever. @KevinOConnor
> 
> I am using it for 18month now without a single issue.

Please don’t delete the branch, it is still useful to have this code around - even if it does not make it into klipper.

devond76180:
I have been using it for a year now with no issues.

Andreas-strg:
Pherhaps we try is with Kalito?

Phil1988:
I think more users will benefit from it, if it makes it into mainline klipper.
Most users dont take a look into pull requests and implementing it on their own to test.
While many would actually use it, if the officiall option was given and integrates as a single QBL button in mainsail/fluidd.

Thanks @devond76180 for this feedback.
I hope it catches more attention if other users can confirm it does work and give value.

klippy/extras/quad_bed_level.py
```diff
@@ -0,0 +1,135 @@
...
+        b1 = [positions[3][1] + offsets[1],
+              self.plot(slope_x_pp03, self.bed_corners[3][0])]
+        b2 = [positions[2][1] + offsets[1],
+              self.plot(slope_x_pp12, self.bed_corners[2][0])]
+        slope_y_s32 = self.linefit(b1, b2)
+        logging.info("quad_bed_level af: %s, bf: %s"
+                     % (slope_y_s01, slope_y_s32))
+        # Calculate z height of each stepper
+        z_height = [0,0,0,0]
+        z_height[0] = (1) * self.plot(slope_y_s01, self.bed_corners[0][1])
```
Wulfsta, line 88: Why is there a multiply by one here as well as on the following lines? Seems like it was included by mistake?

Wulfsta (commented):
My high level understanding of this algorithmically is as follows:

1. Probe points `p0`, `p1`, `p2`, `p3`.
2. For pairs `(p0, p3)` and `(p1, p2)`, perform a slope fit (in z) on the x-axis, then linearly interpolate or extrapolate to the y-axis value of the steppers (or y-rail, here defined by `bed_corners`). Assume the ordering is the same and call these points `e0`, `e1`, `e2`, `e3`. 
3. For pairs `(e0, e1)` and `(e2, e3)`, perform a slope fit (in z) on the y-axis then linearly interpolate or extrapolate to the x-axis value of the steppers as above. Assume the ordering is the same and call these points `s0`, `s1`, `s2`, `s3`.
4. For each `s0`, `s1`, `s2`, `s3`, read the z-axis value, and adjust the corresponding stepper by that amount.
5. Probe the points again; if the z value differences are below a certain threshold, then quit, else repeat these steps.

The issue I see is that you define this as four probe points, here are some edge cases:
1. What if `p0` and `p3` are equivalent? This leads to a division by zero.
2. What if  `p2` and `p3` are very close and the "y rail" is very far away? Then your extrapolants cross each other in the x,y-plane. I think this leads to the values flipping and results in bending the bed more? I haven't actually thought about it in depth though...

I would recommend just keeping the probe points defined as corners, like QGL, as well as adding validation code to avoid this sort of thing.

klippy/extras/quad_bed_level.py
Phil1988: Yes that could be removed.
It is a leftover from some advance math tests I did to improve the accuracy and decrease needed iterations.
But thats nothing major I guess ;)

Phil1988:
> 1. What if `p0` and `p3` are equivalent? This leads to a division by zero.

Then its not "quad" bed level per definition.
Only three z motors is not supported. 

> 2. What if  `p2` and `p3` are very close and the "y rail" is very far away?

I think this edge case is rather fictional then something we can expect from a real printer - what do you think?
I mean we can include some checks and spend some time for printers that might never exist, but well... I have got more important  things to do at the moment 😅

Wulfsta:
> > 1. What if `p0` and `p3` are equivalent? This leads to a division by zero.
> 
> Then its not "quad" bed level per definition. Only three z motors is not supported.

My point is not that this is no longer quad bed leveling, it is that there is no validation code. A bad user input can cause klippy to crash during QBL, rather than failing at configuration parsing time and telling the user why. Validation code, _especially_ for inputs that control physical machines, is important.

> I think this edge case is rather fictional

I tend to agree, but I think that the use case of non-rectangular probe points also makes little sense for most machines. If it is not rectangular, then all the implications of that need to be considered.

Phil1988:
I understand your point, but with respect, I dont get why this feature must have validation checks, while other included features dont.

There are numerous examples, but to give you one thats 1:1 the same: https://github.com/Klipper3d/klipper/blob/master/klippy/extras/quad_gantry_level.py#L119

I mean I could include something, but honestly I dont want to spend more time on a feature that already does work perfectly.

I dont want to fight to make the devs see its value.
Take it or tell me that you dont like it and I will delete it.

Wulfsta:
> I understand your point, but with respect, I dont get why this feature must have validation checks, while other included features dont.

This actually appears like something that was missed, in my opinion QGL should have these same validations - I haven’t looked at that code before. I’ll probably test this to confirm it’s an issue and open a PR to fix it now that you’ve pointed it out.

Phil1988:
What is the appropiate way to delete this request?
It has been 2 years and I am tired.

devond76180:
Wish this feature could have been added.  I still use it and on my system is much better that quad gantry.Lots of new items seem to need reviewers.  Not sure what it takes to be a reviewer.  Sent from my Galaxy
-------- Original message --------From: C0co ***@***.***> Date: 2/4/26  3:19 PM  (GMT-04:00) To: Klipper3d/klipper ***@***.***> Cc: Devon Davis ***@***.***>, Mention ***@***.***> Subject: Re: [Klipper3d/klipper] Added QuadBedLevel feature (PR #6452) Phil1988 left a comment (Klipper3d/klipper#6452)
What is the appropiate way to delete this request?
It has been 2 years and I am tired.

Wulfsta:
> What is the appropiate way to delete this request? It has been 2 years and I am tired.

Closing the PR without deleting the branch is the best approach such that others could still cherry-pick the PR onto their version of klipper. Is the equivalent to using QGL with the motor order reversed? I vaguely recall discovering some way to mutate the inputs to QGL such that was algorithmically the same?

Phil1988:
I am getting spams since opening this PR and dont want to have my information public any more.
@Wulfsta So closing this will not be enough to get this "wiped".
@KevinOConnor Can you please delete this?

Sorry @devond76180 @madnezzm @clearchris and all the silent users who have been using this with no issues.

KevinOConnor:
You can close the PR - just select the close button on this page.

Other than that, I think you'd need to reach out to the github admins if you want the history modified.

-Kevin

PR #6454, opened 2024-01-10
[WIP] Add basic for GPIO extension for RP2040

howard0su:
Plan to add support using 74HC595 to add 8/16/etc output only GPIO to RP2040 to solve RP2040 shortage of IO pins. The PR doesn't include the actual code yet. like to get early feedback while working on hardware prototype.

### DIFF
--- a/src/rp2040/Kconfig
+++ b/src/rp2040/Kconfig
@@ -47,6 +47,21 @@ config STACK_SIZE
     int
     default 512
 
+config GPIO_EXTENSION
+    bool "expend GPIO via 74HC595 chip"
+    default false
+
+config GPIO_EXTENSION_NUM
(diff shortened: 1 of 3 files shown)


TheFuzzyGiggler (commented):
This PR looks dead and is one user specific, doesn't add functionality that likely any other user would use. Recommend rejecting.

bevanweiss:
I'd suggest that the HC595 could be implemented entirely as a Python module.  It could have an optional latch, and an optional reset pin.  And be optionally chained to other such modules.
Normally HC595 are driven by an SPI or similar type of serial interface.  So custom mcu code shouldn't be required for them.  They would just sit on an SPI bus.
@howard0su are you currently using an hc595 in anything that you're doing?
Have you tried to use SPI to drive it instead (whether HW spi or software)

PR #6455, opened 2024-01-10
config: Add Creality Ender 5 S1.

grnbrg:
My first GitHub submission....

From the main commit:

> printer-creality-ender5-s1-2023: New default config file for the Creality Ender 5 S1.
> 
> Creality released the Ender 5 S1 model in November of 2022.  It
> has enough hardware differences from the previous models that
> that the existing Ender 5 configs are not compatible.  This
> configuration is based on one provided by Creality that was then
> tweaked and modified.  I have been using these values (plus some
> additional entries) for about 6 months with no issues.
> 
> I have tidied up the file per the Example_Configs document,
> although I admit I may have missed something.  I'm happy to make
> or accept any required changes.
> 

All IP rights assigned to the Klipper3D project and development team.

Signed-off-by: Brian Greenberg <grnbrg@grnbrg.org>

### DIFF
--- a/config/printer-creality-ender5-s1-2023.cfg
+++ b/config/printer-creality-ender5-s1-2023.cfg
@@ -1,23 +1,16 @@
-# !Ender-5 S1
+# Creality Ender 5 S1
+#
 # printer_size: 220x220x280
 # To use this config, during "make menuconfig" select the STM32F401
 # with a "64KiB bootloader" and serial (on USART1 PA10/PA9)
 # communication.
-
+#
 # Flash this firmware by copying "out/klipper.bin" to a SD card and
 # turning on the printer with the card inserted. The firmware
 # filename must end in ".bin" and must not match the last filename
 # that was flashed.
-
+#
 # See docs/Config_Reference.md for a description of parameters.
-
-[printer]
-kinematics: cartesian
-max_velocity: 300
-max_accel: 5000
-max_z_velocity: 10
-max_z_accel: 1000
-square_corner_velocity: 5.0
 
 [stepper_x]
 step_pin: PC2
@@ -26,7 +19,6 @@ enable_pin: !PC3
 rotation_distance: 40
 microsteps: 16
 endstop_pin: !PA5
-position_min: -5
 position_endstop: 220
 position_max: 222
 homing_speed: 80
@@ -38,9 +30,8 @@ enable_pin: !PC3
 rotation_distance: 40
 microsteps: 16
 endstop_pin: !PA6
-position_min: -2
 position_endstop: 220
-position_max: 225
+position_max: 220
 homing_speed: 80
 
 [stepper_z]
@@ -50,9 +41,8 @@ enable_pin: !PC3
 rotation_distance: 8
 microsteps: 16
 endstop_pin: probe:z_virtual_endstop
-position_min: -2
 position_max: 280
-homing_speed: 4
+homing_speed: 20
 second_homing_speed: 1
 homing_retract_dist: 2.0
 
@@ -65,15 +55,15 @@ microsteps: 16
 nozzle_diameter: 0.400
 filament_diameter: 1.750
 max_extrude_cross_section: 5
-max_extrude_only_distance: 1000.0
+max_extrude_only_distance: 100.0
+pressure_advance: 0.075
 heater_pin: PA1
 sensor_type: EPCOS 100K B57560G104F
 sensor_pin: PC5
-# tuned for stock hardware with 210 degree Celsius target
-control = pid
-pid_kp = 22.698
-pid_ki = 1.363
-pid_kd = 94.481
+control = pid  # tuned for stock hardware with 210 degree Celsius target
+pid_kp = 20.749
+pid_ki = 1.064
+pid_kd = 101.153
 min_temp: 0
 max_temp: 305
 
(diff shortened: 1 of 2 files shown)


config/printer-creality-ender5-s1-2023.cfg
```diff
@@ -0,0 +1,165 @@
+# Creality Ender 5 S1
+#
+# printer_size: 220x220x280
+# To use this config, during "make menuconfig" select the STM32F401
+# with a "64KiB bootloader" and serial (on USART1 PA10/PA9)
+# communication.
+#
+# Flash this firmware by copying "out/klipper.bin" to a SD card and
```
JamesH1978, line 8: Sometimes these boards needs the firmware putting in a specific directory on the SD card, usually STM32F4_UPDATE can you confirm if this is the case?

Thanks
James

config/printer-creality-ender5-s1-2023.cfg
```diff
@@ -0,0 +1,165 @@
...
+fade_start: 1
+fade_end: 10
+fade_target: 0
+
+[mcu]
+serial: /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
+restart_method: command
+
+[safe_z_home]
+home_xy_position: 110,55
```
JamesH1978, line 124: Can you confirm this homes the probe in the middle of the bed? This is useful for the PROBE_CALIBRATE helper as it will use the offsets to move the tool head from the probe being central to the nozzle being central for the paper test in the middle.

Thanks
James

config/printer-creality-ender5-s1-2023.cfg
```diff
@@ -0,0 +1,165 @@
...
+speed: 200
+z_hop: 10
+z_hop_speed: 10
+
+# Probe locations for x-axis twist measurement
+[axis_twist_compensation]
```
JamesH1978, line 130: I am not sure this section should be included in a stock configuration as it has to be calibrated to the person's printer, it is usually not the case that every single frame coming out of the manufacturer is the same. 

Thanks
James

JamesH1978 (changes requested):
Thank you for submitting this PR, please see my comments in the code

Thanks
James.

config/printer-creality-ender5-s1-2023.cfg
```diff
@@ -0,0 +1,165 @@
+# Creality Ender 5 S1
```
Sineos, line 1: As Creality is known to switch around boards, please also provide the exact information of the used board, e.g. "Creality v4.7.11 Rev 38" as indicated on the Silkscreen

config/printer-creality-ender5-s1-2023.cfg
grnbrg: You are correct.  I will update the header comments, and include the fact that in addition to the `STM32F4_UPDATE` directory, the filename must differ from whatever filename the current firmware used.

config/printer-creality-ender5-s1-2023.cfg
grnbrg: `110,55` places the nozzle in the center of the X axis, and 25% from the front edge of the print area, which is a location that is convenient to examine and/or clean the nozzle and hot end.  There is sufficient space for the movements required to run `PROBE_CALIBRATE` from this location.  `110,110` would put the nozzle in the exact center of the print area.  

Given the above reasoning, please confirm the change from `110,55` to `110,110`.

config/printer-creality-ender5-s1-2023.cfg
grnbrg: The section defines the bed locations used by the `AXIS_TWIST_COMPENSATION_CALIBRATE`, but does not actually set any specific calibration.  Based on personal experience and reports in various forums, nearly all Ender 5s suffer from a degree of Y axis tilt.  I could add a comment making this clear, with a suggestion to run the calibration.

Let me know if I should remove the section, add a more detailed comment, or leave as is.

config/printer-creality-ender5-s1-2023.cfg
grnbrg: Will change to 

`Creality Ender 5 S1 (HW Version: CR4NS200141C13)`

Thanks!

config/printer-creality-ender5-s1-2023.cfg
grnbrg: Suggested comment:

```
> 
> # Many Ender 5 S1 printers appear to suffer from a slight twist
> # in the X axis.  This twist can be measured and compensated for
> # using the AXIS_TWIST_COMPENSATION_CALIBRATE G-Code command.  See
> # https://www.klipper3d.org/Axis_Twist_Compensation.html for more
> # information.  This section provides the setup for this optional
> # calibration step.
> [axis_twist_compensation]
> (....)
```

config/printer-creality-ender5-s1-2023.cfg
JamesH1978: Usually we would home with the probe in the centre and then the PROBE_CALIBRATE command would then move the nozzle to the centre of the bed using the defined bltouch offsets, for the paper test. Unless there is a specific reason why you would want to do the paper test at a front location? Most people are not going to do a move before they run the command they're going to home and then they are going to run it. Due to your offsets I would suggest a position of `123,83` to home the probe centrally.

Thanks
James

grnbrg: That makes perfect sense to me.  Final (hopefully) commit has been made.

Thanks for your time.


Brian.

JamesH1978 (approved):
Looks fine to me now,

thanks
James

PR #6458, opened 2024-01-13
Deprecate and remove support for uncommon options in output_pin

KevinOConnor:
Over the last few years the `output_pin` module has gained a handful of options that have made it harder to maintain the the low-level backend code.  This PR simplifies the output_pin module.  The main changes are:
1. The `static_value` parameter is deprecated.  This option has little impact - it only saves a few bytes in the micro-controller.  Going forward it is simple enough to just not specify it.
2. The `maximum_mcu_duration` parameter is deprecated.  The `pwm_tool` module is a better choice for tools that require this capability.  It should be straight forward for users to change their configs from `[output_pin mypin]` to `[pwm_tool mypin]`.
3. The `SET_PIN CYCLE_TIME=x` g-code command for `[output_pin mypin]` config sections has been removed.  Users can convert to a new `[pwm_cycle_time mypin]` config as a replacement.  This change is setup to take effect immediately (no deprecation period).

This is the last PR as part of the `pwm_tool` series of changes (eg, #6369, #6410, #6420).

@Cirromulus - fyi.

-Kevin

### DIFF
--- a/config/generic-mini-rambo.cfg
+++ b/config/generic-mini-rambo.cfg
@@ -84,7 +84,7 @@ pwm: True
 scale: 2.0
 cycle_time: .000030
 hardware_pwm: True
-static_value: 1.3
+value: 1.3
 
 [output_pin stepper_z_current]
 pin: PL4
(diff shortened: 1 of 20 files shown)


Cirromulus:
I think this change is good for code quality.
About the `maximum_mcu_duration` I personally only use it together with `pwm_tool`, but I can imagine that people might use it for something safety related. At the same time, they could just use `pwm_tool` instead, even if it is not actually a tool, so it is probably just fine.

PR #6461, opened 2024-01-17
bed_mesh: Implement adaptive bed mesh

voidtrance:
This PR is for an implementation of adaptive bed meshes. Adaptive bed meshes use the exclude objects data to determine the area of the print bed being used and only create a mesh for that bed area. This has the benefit of greatly decreasing the amount of time needed to start a print without compromising print quality.

These changes have been tested on both Cartesian and Delta printers through the use of SimulAVR (with some custom Klipper modifications to enable it to report probe point z offset through Moonraker) and by many actual users in the Voron Discord.

Below is a ZIP archive containing two HTML files that show the behavior of the adaptive mesh. Plots where generated using test data from the SimulAVR simulation.

[adaptive-mesh-plots.zip](https://github.com/Klipper3d/klipper/files/13969262/adaptive-mesh-plots.zip)

The commands issues to define the print objects that produce these plots are below:

[commands-cartesian.txt](https://github.com/Klipper3d/klipper/files/13969274/commands-cartesian.txt)
[commands-delta.txt](https://github.com/Klipper3d/klipper/files/13969275/commands-delta.txt)


kyleisah:
Amazing! Thank you for all your impressive work on this! 😄

zellneralex (commented):
small typo in md

docs/Bed_Mesh.md
```diff
@@ -370,14 +370,54 @@ are identified in green.
...
+automatically adjust the mesh parameters based on the area occupied by the defined
+print objects.
+
+The adapted mesh area will be computed from the area defined by the boundaries of all
+the defined print object so it covers every object, including any margins defined in
+the configuration. After the area is computed, the number of probe points will be
+scaled down based on the ratio of the default mesh area and the adapted mesh area. To
+illustrate this consider the following example:
+
+For a 150mmx150mm bed with `mesh_min` set to `25,25` and `mesh_max` set to `25,25`,
```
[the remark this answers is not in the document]
voidtrance: Fixed. Thank you.

klippy/extras/bed_mesh.py
```diff
@@ -2,10 +2,14 @@
 #
 # Copyright (C) 2018  Kevin O'Connor <kevin@koconnor.net>
 # Copyright (C) 2018-2019 Eric Callahan <arksine.code@gmail.com>
+# Copyright (C) 2023  Kyle Hansen (Kyleisah) <kyleisah@gmail.com>
+# Copyright (C) 2023  Brandon Nance <bwnance@gmail.com>
+# Copyright (C) 2023  Mitko Haralanov <voidtrance@gmail.com>
```
Arksine, line 7: Generally we only add a copyright on initial submissions.  We'd like to put everyone in there who has contributed, however we want to keep header comments from growing out of control.

Arksine (commented):
Thanks.  Overall the implementation looks good.  I have some comments, however the are minor and mostly centered around code organization.

klippy/extras/bed_mesh.py
```diff
@@ -103,6 +107,7 @@ def __init__(self, config):
         self.log_fade_complete = False
         self.base_fade_target = config.getfloat('fade_target', None)
         self.fade_target = 0.
+        self.margin = config.getfloat('margin', 0.0)
```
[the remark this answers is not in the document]
voidtrance: We wanted to have a new option to the `[bed_mesh]` section. If the attribute was moved to the `BedMeshCalibrate` class, how would it work as a option to the `[bed_mesh]` section?

klippy/extras/bed_mesh.py
```diff
@@ -653,8 +751,17 @@ def cmd_BED_MESH_CALIBRATE(self, gcmd):
         self._profile_name = gcmd.get('PROFILE', "default")
         if not self._profile_name.strip():
             raise gcmd.error("Value for parameter 'PROFILE' must be specified")
+        exclude_object = None
+        adaptive_mesh = bool(gcmd.get_int('ADAPTIVE', 0))
+        margin = gcmd.get_float('ADAPTIVE_MARGIN', self.bedmesh.margin)
+        if adaptive_mesh:
+            try:
+                exclude_object = self.printer.lookup_object("exclude_object")
```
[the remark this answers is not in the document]
voidtrance: > I would recommend calling after the ALGORITHM option is checked, as we don't want any other parameters to override the adaptive mesh.

I added the adaptive logic before the rest of the parameter processing in order to allow parameters specified on the `BED_MESH_CALIBRATE` command line to override the adaptive settings. If the adaptive logic was called after, would the adaptive params override `BED_MESH_CALIBRATE MESH_MIN=x,y`?

klippy/extras/bed_mesh.py
```diff
@@ -584,6 +590,98 @@ def update_config(self, gcmd):
 
         params = gcmd.get_command_parameters()
         need_cfg_update = False
+
+        # If adaptive meshing is enabled and there are defined objects,
+        # generate adaptive mesh
+        if do_adaptive and exclude_object is not None and \
+                len(exclude_object.objects) != 0:
```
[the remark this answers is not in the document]
voidtrance: Why isn't it valid? `ExcludeObject._reset_state()` is being called from the constructor so the member attribute should always be defined. I can switch the code as your suggestion gives me the same result but I just want to understand your comment.

klippy/extras/bed_mesh.py
voidtrance: So how are attributions to the codebase by other people handled? IANAL, but my understanding is that under the GPL, code authors retain their copyright for the code that they've submitted. In such cases, shouldn't the copyright notice be added?

klippy/extras/bed_mesh.py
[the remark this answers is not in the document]
KevinOConnor: It's against our coding conventions as mentioned at: https://www.klipper3d.org/Code_Overview.html#adding-a-host-module

Accessing the internal member variables of other objects leads to a maintenance burden.  Other developers working on ExcludeObject wont know that some other module has accessed its interanl state and those developers may make a local change that could subtly break the bed_mesh module.  So, the convention is to avoid doing that.

-Kevin

klippy/extras/bed_mesh.py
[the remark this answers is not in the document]
Arksine: Yes.  `BedMesh` passes its config object to the `BedMeshCalibrate` class, they share the same config section.  Several of the config items in `[bed_mesh]` are already retrieved there such as the `zero_reference_position`.

klippy/extras/bed_mesh.py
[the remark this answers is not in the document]
Arksine: Yes, if the adaptive logic is added after it will override other params.  I think this makes sense, as that is the point of activating the adaptive mesh.  Mixing parameters with the adaptive logic would seem to produce unexpected results.

klippy/extras/bed_mesh.py
Arksine: I believe the copyright is implied when the DCO is accepted in the commit.  This is covered in point 3 of [what to expect in a review](https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review):

```
New C files and Python files should have an unambiguous copyright statement.
See the existing files for the preferred format. Declaring a copyright on an existing
file when making minor changes to that file is discouraged.
```

This could be considered more than a minor change, so perhaps keeping the copyright is ok.  I'll let Kevin chime in on that.

klippy/extras/bed_mesh.py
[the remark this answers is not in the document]
voidtrance: Done in latest revision.

klippy/extras/bed_mesh.py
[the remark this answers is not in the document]
voidtrance: Done in latest revision.

klippy/extras/bed_mesh.py
voidtrance: OK, the latest revision still contains the copyrights. I will adjust based on Kevin's input.

KevinOConnor: We don't want the copyright header to turn into a "laundry list of everyone that has worked on the code".  Use the commit messages to document who worked on the code - preferably in the form of a signed-off-by line.  In my experience, if the copyright block does become a "laundry list" it tends to grow out of control - requiring scrolling through pages just to get to the code - which becomes a maintenance burden.

When I look at bed_mesh.py it seems to overwhelmingly be Eric's code.  If there is a feeling that a major contribution is being made then it's fine to add additional copyright(s).  I'd say that determination is up to Eric.  For what it is worth, at a quick glance, adding three copyright statements for 100 lines of code seems excessive to me.

FWIW, you can remove my copyright from bed_mesh.py (Eric added my copyright back in the original bed_mesh.py commit - it is not necessary).

Hope that helps.

-Kevin

kyleisah: Just chiming in here regarding the copyright, I am fine with my name being removed as it was VoidTrance who did 99% of the work implementing the logic of KAMP (my project, the repo is in my GitHub profile). That said, some credit somewhere in the klipper repo would be favorable of course. I approached VoidTrance with the idea and helped any way I could, but I cannot take credit for the work he’s done here. I laid some foundation, VoidTrance built the house. 

As for Brandon Nance, he tried to help me for a couple hours when we initially had the idea to bring this into mainline klipper, we wrote a few lines of code together, but easily 90+% of the work was done by VoidTrance. 

I don’t want to complicate things with who did what, just want to give Kevin some backstory so he can make the final decision without missing any nuance here.

voidtrance: Thanks for the input Kevin.

Eric, I guess it's your call. Getting this in is more important to me than having my copyright in the file so I am fine either way.

KevinOConnor: > I am fine with my name being removed 

> Getting this in is more important to me than having my copyright in the file so I am fine either way.

Thanks.  I think we should not add these new copyright statements then.  The more brief we can keep the header the more brief we'll be able to keep it in the future.

If you're in that area of the code, feel free to remove my copyright statement.

Thanks again.

-Kevin

Arksine:
The latest changes look good.  One minor thing I missed earlier is that we should add the `adaptive_margin` option to the [Config Reference](https://github.com/Klipper3d/klipper/blob/master/docs/Config_Reference.md?plain=1#L993) as well.  When that is done I'm ready to approve the PR.

klippy/extras/bed_mesh.py
voidtrance: Done.

voidtrance:
> The latest changes look good. One minor thing I missed earlier is that we should add the `adaptive_margin` option to the [Config Reference](https://github.com/Klipper3d/klipper/blob/master/docs/Config_Reference.md?plain=1#L993) as well. When that is done I'm ready to approve the PR.

Done.

I also re-ran the simulated tests with the latest revision and compared it against the graphs that were generated with the original version submitted with the PR to make sure that thing did not break during the review edits.

Arksine (approved):

KevinOConnor:
Thanks.  Happy to commit.  I think we should give a few days to see if there are any other comments.  Assuming nothing else comes up, I'll look to commit next week.

-Kevin

kyleisah:
> Thanks. Happy to commit. I think we should give a few days to see if there are any other comments. Assuming nothing else comes up, I'll look to commit next week.

Amazing, thank you so much!

docs/Bed_Mesh.md
```diff
@@ -370,14 +370,54 @@ are identified in green.
...
+the configuration. After the area is computed, the number of probe points will be
+scaled down based on the ratio of the default mesh area and the adapted mesh area. To
+illustrate this consider the following example:
+
+For a 150mmx150mm bed with `mesh_min` set to `25,25` and `mesh_max` set to `125,125`,
+the default mesh area is a 100mmx100mm square. An adapted mesh area of `50,50`
+means a ratio of `0.5x0.5` between the adapted area and default mesh area.
+
+If the `bed_mesh` configuration specified `probe_count` as `7x7`, the adapted bed
+mesh will use 4x4 probe points (roundUp(7 * 0.5)).
```
flowerysong, line 391: `roundUp()` is neither an existing term of art in Klipper's documentation nor the actual name of the function called, so this should probably just be written in English.
```suggestion
mesh will use 4x4 probe points (7 * 0.5 rounded up.)
```

docs/Bed_Mesh.md
```diff
@@ -370,14 +370,54 @@ are identified in green.
...
+mesh_min: 35, 6
+mesh_max: 240, 198
+probe_count: 5, 3
+adaptive_margin: 5
+```
+
+- `adaptive_margin` \
+  _Default Value: 0_ \
+  Margin (in mm) to add around the area of the bed used by the defined objects. The diagram
+  below shows the adapted bed mesh area with a `adaptive_margin` set to 5mm. The adapted mesh
```
flowerysong, line 407:
```suggestion
  below shows the adapted bed mesh area with an `adaptive_margin` of 5mm. The adapted mesh
```

docs/Config_Reference.md
```diff
@@ -991,6 +991,9 @@ Visual Examples:
 #   Optional points that define a faulty region.  See docs/Bed_Mesh.md
 #   for details on faulty regions.  Up to 99 faulty regions may be added.
 #   By default no faulty regions are set.
+#adaptive_margin:
+#   An optional margin (in mm) to be added around the bed area used by
+#   the definded print objects when generating an adaptive mesh.
```
flowerysong, line 996:
```suggestion
#   the defined print objects when generating an adaptive mesh.
```

docs/Bed_Mesh.md
voidtrance: This was intended as pseudo-code but I can change it.

docs/Config_Reference.md
voidtrance: Done.

docs/Bed_Mesh.md
voidtrance: Done.

docs/Bed_Mesh.md
```diff
@@ -370,14 +370,54 @@ are identified in green.
...
 
+### Adaptive Meshes
+
+Adaptive bed meshing is a way to speed up the bed mesh generation by only probing
+the area of the bed used by the objects being printed. When used, the method will
+automatically adjust the mesh parameters based on the area occupied by the defined
+print objects.
+
+The adapted mesh area will be computed from the area defined by the boundaries of all
+the defined print object so it covers every object, including any margins defined in
```
thijstriemstra, line 381: object -> objects

klippy/extras/bed_mesh.py
```diff
@@ -1,6 +1,5 @@
 # Mesh Bed Leveling
 #
-# Copyright (C) 2018  Kevin O'Connor <kevin@koconnor.net>
```
thijstriemstra, line 3: why remove this?

Arksine: He removed it [at Kevin's request](https://github.com/Klipper3d/klipper/pull/6461#discussion_r1458008900).

thijstriemstra: sorry, thanks for clarifying

marbocub:
Hello, I'm interested in this PR.
When the purge line in a start gcode is not labeled and is too far from printing objects, the nozzle may scratch the table at the line in the worst case because it is not probed around the line.

I think this has to be careful and think it would be helpful to include a note or a labeling method in the document.

voidtrance:
> Hello, I'm interested in this PR.
> 
> When the purge line in a start gcode is not labeled and is too far from printing objects, the nozzle may scratch the table at the line in the worst case because it is not probed around the line.
> 
> 
> 
> I think this has to be careful and think it would be helpful to include a note or a labeling method in the document.

I am not sure that I understand. Are you asking for the purge line area to be probed or to document that the purge line area is not probed?

Whether the purge line will be probed depends entirely on the slicer-generated gcode. If the slicer generates a purge line that is labeled as an object, it will be probed. Otherwise, it won't.

I think that describing all of that and documenting slicer behavior is beyond the scope of Klipper documentation.

marbocub:
> or to document that the purge line area is not probed?

Yes, this is my opinion.

> Whether the purge line will be probed depends entirely on the slicer-generated gcode. If the slicer generates a purge line that is labeled as an object, it will be probed. Otherwise, it won't.

I think it would be kind to include this content or simply the first half to the document. This may reduce a user accidentally scratching the table.

kyleisah:
> I think it would be kind to include this content or simply the first half to the document. This may reduce a user accidentally scratching the table.

There is the `adaptive_margin` configuration to expand the bounds of the mesh, so purging may not be necessary as you can “purge” using something like a skirt. 

Speaking from experience here, I’ve ran many, many adaptive meshes and have used many different purge configurations (adaptive and static) and have not damaged a buildplate or nozzle. 

I believe it should be left up to the user what they feel is best to handle that, however I do not foresee it being a problem.

marbocub:
I understand. I appreciate your consideration for all.

kyleisah:
> I understand. I appreciate your consideration for all.

I appreciate you understanding. I do, however, agree that although *I've* had no problems with it in the past, if someone's machine is a unique setup or has mechanical issues causing them to rely on a full mesh, that should be laid out clearly. It would require a fairly unique complication with the machine, but I suppose it is possible afterall that issuing print moves outside of the adaptive meshing area can be problematic.

docs/Bed_Mesh.md
kyleisah, line 381:
```suggestion
the defined print objects so it covers every object, including any margins defined in
```

kyleisah (commented):
Added some verbiage for potential pitfalls when attempting print moves outside of an adaptive mesh's bounds on machines that have mechanical issues that a full mesh typically compensates for. I also added a note that adaptive bed meshes are not meant to be reused, and are best utilized by building a mesh prior to every print.

Sineos:
> if someone's machine is a unique setup

@marbocub's concern is essentially valid. In the support activities, I have seen enough bed topologies that range from -0.5 to +0.5 or even worse. It seems that many users mistake bed-meshing as an alternative to leveled bed or generally have completely crooked beds.

Maybe a sound recommendation would be building a mesh of the entire bed and review it. If it deviates more than 1 layer-height, adaptive meshing should be used with caution.

Sineos:
In addition: Not sure if the adaptive meshing changes anything here, but usually the "boundary" points of a mesh are carried over to the unmeshed region.
This might even create another challenge for crooked beds. Let's assume your last meshed point was -0.2 but for whatever reason your bed starts rising to +0.5 outside this point. 
Now, carrying over -0.2 to the "physical +0.5 region" increases your problem to 0.7.

kyleisah:
> Maybe a sound recommendation would be building a mesh of the entire bed and review it. If it deviates more than 1 layer-height, adaptive meshing should be used with caution.

Yeah, I fully agree here. I will adjust the suggestion I've made with some more verbiage about this. I can fully relate with seeing some... "interesting" mesh results, so the concern is totally valid.

>In addition: Not sure if the adaptive meshing changes anything here, but usually the "boundary" points of a mesh are carried over to the unmeshed region.

This is correct, I actually just double-checked this was the behavior today because I was not sure how that is handled. But yes. You can imagine it sort or like a bowl with a very wide flange. The edges of the mesh are scaled out in XY to infinity.

Some of these cases we are imagining are very extreme, almost as if generating a bed mesh on a waffle-iron, but they are absolutely within the realm of possibility.

docs/Bed_Mesh.md
voidtrance: Done.

docs/Bed_Mesh.md
```diff
@@ -370,14 +370,54 @@ are identified in green.
 
 ![bedmesh_interpolated](img/bedmesh_faulty_regions.svg)
 
+### Adaptive Meshes
+
+Adaptive bed meshing is a way to speed up the bed mesh generation by only probing
+the area of the bed used by the objects being printed. When used, the method will
+automatically adjust the mesh parameters based on the area occupied by the defined
+print objects.
+
```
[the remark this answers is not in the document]
voidtrance: @kyleisah, I have added this to the documentation in a reworded/reorg'ed form.

@Arksine, I would appreciate a review of the new documentation section from you, as well?

klippy/extras/bed_mesh.py
```diff
@@ -573,6 +573,112 @@ def _verify_algorithm(self, error):
...
+                self.origin = adapted_origin
+                self.mesh_min = (-self.radius, -self.radius)
+                self.mesh_max = (self.radius, self.radius)
+                self.mesh_config["x_count"] = self.mesh_config["y_count"] = \
+                        max(new_x_probe_count, new_y_probe_count)
+        else:
+            self.mesh_min = adjusted_mesh_min
+            self.mesh_max = adjusted_mesh_max
+            self.mesh_config["x_count"] = new_x_probe_count
+            self.mesh_config["y_count"] = new_y_probe_count
```
Arksine, line 680:
```suggestion
            self.mesh_config["y_count"] = new_y_probe_count
        self._profile_name = None
```

Arksine: To disable the mesh autosave we can set `self._profile name` to `None` here, then we can change line 892:
```python
if self._profile_name is not None:
    self.bedmesh.save_profile(self._profile_name)
```

docs/Bed_Mesh.md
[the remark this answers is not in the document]
Arksine: See the review point [here](https://github.com/Klipper3d/klipper/pull/6461#discussion_r1462263644).  Hopefully its clear enough, Github doesn't seem to allow suggestions outside of the modified code.  Essentially we just set `self._profile_name = None` right before returning `True` from `set_adaptive_mesh()`.  Then we modify the last line `probe_finalize()` to prevent from sending the `None` value `save_profile()`.

klippy/extras/bed_mesh.py
voidtrance: > To disable the mesh autosave we can set `self._profile name` to `None` here, then we can change line 892:
> 
> ```python
> if self._profile_name is not None:
>     self.bedmesh.save_profile(self._profile_name)
> ```

Wouldn't calling `BED_MESH_PROFILE SAVE=...` still end up saving the adapted mesh? I think we would have to prevent this from happening, as well.

voidtrance: > To disable the mesh autosave we can set `self._profile name` to `None` here, then we can change line 892:
> 
> ```python
> if self._profile_name is not None:
>     self.bedmesh.save_profile(self._profile_name)
> ```

What if we set a flag in `set_adaptive_mesh()` (`self.is_adaptive = True`) and then in `ProfileManager.save_config()` we check if it's True. If so, display an error message?

voidtrance: This is the proposed diff that should take care of both the auto- and explicit save:

```diff
diff --git a/klippy/extras/bed_mesh.py b/klippy/extras/bed_mesh.py
index bc053e27..2853ed46 100644
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -90,6 +90,7 @@ class BedMesh:
         self.printer.register_event_handler("klippy:connect",
                                             self.handle_connect)
         self.last_position = [0., 0., 0., 0.]
+        self._is_adaptive = False
         self.bmc = BedMeshCalibrate(config, self)
         self.z_mesh = None
         self.toolhead = None
@@ -216,6 +217,8 @@ class BedMesh:
                     raise self.gcode.error(
                         "Mesh Leveling: Error splitting move ")
         self.last_position[:] = newpos
+    def set_adaptive(self, adaptive):
+        self._is_adaptive = adaptive
     def get_status(self, eventtime=None):
         return self.status
     def update_status(self):
@@ -225,6 +228,7 @@ class BedMesh:
             "mesh_max": (0., 0.),
             "probed_matrix": [[]],
             "mesh_matrix": [[]],
+            "adaptive" : self._is_adaptive,
             "profiles": self.pmgr.get_profiles()
         }
         if self.z_mesh is not None:
@@ -678,6 +682,7 @@ class BedMeshCalibrate:
             self.mesh_max = adjusted_mesh_max
             self.mesh_config["x_count"] = new_x_probe_count
             self.mesh_config["y_count"] = new_y_probe_count
+        self.bedmesh.set_adaptive(True)
         return True
     def update_config(self, gcmd):
         # reset default configuration
@@ -690,6 +695,7 @@ class BedMeshCalibrate:
 
         params = gcmd.get_command_parameters()
         need_cfg_update = False
+        self.bedmesh.set_adaptive(False)
         if self.radius is not None:
             if "MESH_RADIUS" in params:
                 self.radius = gcmd.get_float("MESH_RADIUS")
@@ -889,7 +895,8 @@ class BedMeshCalibrate:
             z_mesh.set_zero_reference(*self.zero_ref_pos)
         self.bedmesh.set_mesh(z_mesh)
         self.gcode.respond_info("Mesh Bed Leveling Complete")
-        self.bedmesh.save_profile(self._profile_name)
+        if not self.bedmesh.get_status().get("adaptive", False):
+            self.bedmesh.save_profile(self._profile_name)
     def _dump_points(self, probed_pts, corrected_pts, offsets):
         # logs generated points with offset applied, points received
         # from the finalize callback, and the list of corrected points
@@ -1332,6 +1339,9 @@ class ProfileManager:
                 "file and restart the printer" %
                 (('\n').join(self.incompatible_profiles)))
     def save_profile(self, prof_name):
+        if self.bedmesh.get_status().get("adaptive", False):
+            self.gcode.respond_info("Adaptive meshes cannot be saved.")
+            return
         z_mesh = self.bedmesh.get_mesh()
         if z_mesh is None:
             self.gcode.respond_info(
```

Arksine: I don't think its necessary to prevent users from explicitly saving the profile.  That is completely opt-in, and perhaps they may have some good reason to do it.  My guess is that for the general user it won't be an issue, because adaptive meshing will only occur during a print.  If they are using a supplied configuration then `BED_MESH_CALIBRATE ADAPTIVE=1` will be in the start macro and `BED_MESH_CLEAR` will be in the cancel and end macros, reducing the chance of unintentionally saving.

docs/Bed_Mesh.md
```diff
@@ -370,14 +370,68 @@ are identified in green.
...
+By nature, adaptive bed meshes use the objects defined by the Gcode file being printed.
+Therefore, it is expected that each Gcode file will generate a mesh that probes a different
+area of the print bed. Therefore, adapted bed meshes should not be re-used. The expectation
+is that a new mesh will be generated for each print if adaptive meshing is used.
+
+It is also important to consider that adaptive bed meshing is best used on machines that can
+normally probe the entire bed and achieve a maximum variance less than or equal to 1 layer
+height. Machines with mechanical issues that a full bed mesh normally compensates for may
+have undesirable results when attempting print moves **outside** of probed area. If a full
```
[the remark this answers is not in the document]
kyleisah: just a small fix here, but otherwise this is all much better than what I wrote, so thank you for that.

klippy/extras/bed_mesh.py
voidtrance: > I don't think its necessary to prevent users from explicitly saving the profile. That is completely opt-in, and perhaps they may have some good reason to do it. My guess is that for the general user it won't be an issue, because adaptive meshing will only occur during a print. If they are using a supplied configuration then `BED_MESH_CALIBRATE ADAPTIVE=1` will be in the start macro and `BED_MESH_CLEAR` will be in the cancel and end macros, reducing the chance of unintentionally saving.

OK. I use your suggested change then.

docs/Bed_Mesh.md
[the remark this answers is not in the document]
voidtrance: Done. Thanks.

klippy/extras/bed_mesh.py
voidtrance: @Arksine, done.

Arksine:
New changes look good to me.  Thanks.

marbocub:
New changes resolved the concern I had. Thank you very much.

kyleisah:
> New changes resolved the concern I had. Thank you very much.

Wonderful! Thank you for bringing those concerns forward. It’s best to correctly set the expectation in the documentation rather than leave it up for interpretation. 😄

KevinOConnor:
Thanks!

I committed this change.  I also added my signed-off-by just to acknowledge the change to the copyright statement .

-Kevin

Grex3D:
Hi guys, after using adaptive bed mesh, i don't understand how to load it (to see it in heighmap) ?
thanks :)

kyleisah:
@KevinOConnor @Arksine @voidtrance we might have a problem here. I just tried this as well and there's no heightmap shpwing up in my web interface (mainsail). However, when I use `BED_MESH_OUTPUT` I do see mesh information there, I just can't see it in the frontend anymore. This might be related to the small bit of code that was changed to help prevent people from saving adaptive meshes and re-using them, so this might actually be a problem that mainsail/fluidd or other frontends might have to patch. I'm not sure, just wanted to report my findings here.

kyleisah:
@meteyou perhaps this is something you should be aware of. Seems adaptive meshes do not show up in the heightmap GUI as they are only meant to be used during prints. If the machine status is "printing" the mesh information is being used, and when the print has finished or is cancelled I believe the mesh information will be removed.

voidtrance:
How do the front ends get the mesh data (coordinates and height)?

kyleisah:
Thankfully it’s not a *pressing* issue or problem, as the mesh information is in fact being used during a print, we just don’t have a visual representation in the frontend currently.

hex-344505:
Hi, not sure if it is right place. How to enable this feature?
my config section looks like 
```
[bed_mesh]
speed: 500
horizontal_move_z: 5
#horizontal_move_z: 10
mesh_min: 30, 30
mesh_max: 320, 320
probe_count: 7,7
adaptive_margin: 5
mesh_pps: 2,2
algorithm: bicubic
bicubic_tension: 0.2
move_check_distance: 3
split_delta_z: 0.01
fade_start: 2.0 
fade_end: 10.0
fade_target: 0
```
But printer probed all the bed. Previously used KAMP to make adaptive bed. Thought it should be enabled by default and did not find documented way to enable it.

voidtrance:
> Hi, not sure if it is right place. How to enable this feature?
> 
> my config section looks like 
> 
> ```
> 
> [bed_mesh]
> 
> speed: 500
> 
> horizontal_move_z: 5
> 
> #horizontal_move_z: 10
> 
> mesh_min: 30, 30
> 
> mesh_max: 320, 320
> 
> probe_count: 7,7
> 
> adaptive_margin: 5
> 
> mesh_pps: 2,2
> 
> algorithm: bicubic
> 
> bicubic_tension: 0.2
> 
> move_check_distance: 3
> 
> split_delta_z: 0.01
> 
> fade_start: 2.0 
> 
> fade_end: 10.0
> 
> fade_target: 0
> 
> ```
> 
> But printer probed all the bed. Previously used KAMP to make adaptive bed. Thought it should be enabled by default and did not find documented way to enable it.

The documentation has not made it to the website yet.

You add the `ADAPTIVE=1` to the `BED_MESH_CALIBRATE` command.

Arksine:
> How do the front ends get the mesh data (coordinates and height)?

Its reported by the `BedMesh.get_status()` method.  My guess is that Mainsail is getting confused by an empty string value in the `profile_name` field, because the mesh data itself does reflect the current adaptive mesh.  It appears that Fluidd doesn't validate the current profile name and therefore renders the heightmap.

Since we introduced a minor change in behavior with profile reporting I think the burden is on us to correct it.  I'm working on a patch now.  I already had some other minor work that needed to be done, so I will add it in the same PR.

voidtrance:
> > How do the front ends get the mesh data (coordinates and height)?
> 
> 
> 
> Its reported by the `BedMesh.get_status()` method.  My guess is that Mainsail is getting confused by an empty string value in the `profile_name` field, because the mesh data itself does reflect the current adaptive mesh.  It appears that Fluidd doesn't validate the current profile name and therefore renders the heightmap.
> 
> 
> 
> Since we introduced a minor change in behavior with profile reporting I think the burden is on us to correct it.  I'm working on a patch now.  I already had some other minor work that needed to be done, so I will add it in the same PR.
> 
> 

Thank you. I was planning on trying to figure it out but Mainsail code is completely new to me plus I wouldn't have had time until Monday.

I appreciate you jumping in with the fix.

zellneralex:
As a short not, yes the empty string is the issue, @meteyou is already working on it. 
‘He deceive to refactor the module so the fix should be part of the next Mainsail release

matmen:
Fluidd correctly shows the adaptive mesh, it just doesn't show up in the profile list (as expected):
image

zellneralex:
@matmen I do not doubt that, we saw the same as long a mesh like default was loaded before the execution with ADAPTIVE=1. Your picture shows also the name default.

So you should test if again if fluidd does not name a mesh with an empty name string default. 

@voidtrance can you check that the name will also be an empty string if a mesh was already loaded, currently it looks like the name stays whatever it was before.

matmen:
@zellneralex The default bed mesh shows up because it's defined in my config - fluidd just shows the reported mesh list, which doesn't include the adaptive mesh (with an empty name). We only use the bed mesh name for showing which mesh profile is active, which is why the "default" profile in the screenshot is showing a load button.

zellneralex:
> @zellneralex The default bed mesh shows up because it's defined in my config - fluidd just shows the reported mesh list, which doesn't include the adaptive mesh (with an empty name). We only use the bed mesh name for showing which mesh profile is active, which is why the "default" profile in the screenshot is showing a load button.
> 
> If there's no bed mesh profiles reported, we just show a message, but the visualization still shows up: image

Perfect, then fluidd should be save.

Arksine:
FWIW, #6473 will add profile names to the adaptive mesh.  It will be something like `adaptive-<hexstring>`, where the hexstring the the mesh's internal object id.  That will make it relatively unique.  It isn't intended that users save these meshes, although they could if they wanted to (perhaps for some experiment).

kyleisah:
Really appreciate the rapid and collaborative work to patch this you guys. 😄 Thanks for doing that.

I've been monitoring some discord servers and forums and it seems like there are some areas the documentation is still lacking. I'll fork klipper and make some changes to the docs to improve any gaps in configuration/setup and make a PR. Really it just boils down to setting up exclude_object correctly and some other small things. Thank you again!

AudreyAP:
@Arksine Is it maybe a good idea to add an option to make save_config ignore adaptive-xxxx meshes?

voidtrance:
> @Arksine Is it maybe a good idea to add an option to make save_config ignore adaptive-xxxx meshes?

Arksine already mentioned in his fix up commit message that he is explicitly allowing the saving of adaptive meshes in case anyone wants to use them for some type of testing/research.

Arksine:
You have to explicitly call `BED_MESH_PROFILE SAVE=xxxx` after probing an adaptive mesh to save it.  You could call it `adaptive-xxx` or anything else.  Without this set `SAVE_CONFIG` will ignore it.

AudreyAP:
🤦‍♀️ whoops, I totally missed that. 

Sounds great, thank you both.

kot0005:
Hi this feature is not working. I am on mainsail. It still probes the whole bed.

I was using a different adaptive bed mesh and it was working.

image


image

image



I dont save my meshes because i have a belted z and the bed needs to be calibrated every time as the bed drops a bit after steppers are disabled.

AudreyAP:
@kot0005 remove the bed_mesh_profile load=default line, and send a screenshot of your start gcode settings in slicer.

kot0005:
> @kot0005 remove the bed_mesh_profile load=default line, and send a screenshot of your start gcode settings in slicer.

hi, this is my start g code in slicer START_PRINT BED_TEMP=[first_layer_bed_temperature] EXTRUDER_TEMP=[first_layer_temperature]

if i remove the line you mentioned, will it still use the mesh it calibrated from the calibrate command ?

ty

AudreyAP:
@kot0005 if you leave the line in there, it most definitely won't work. 

Try adding M117 before you call start_print in slicer start gcode.

kot0005:
> @kot0005 if you leave the line in there, it most definitely won't work.
> 
> Try adding M117 before you call start_print in slicer start gcode.

i removed the load profile code, no change. will try m117
image

kot0005:
> @kot0005 if you leave the line in there, it most definitely won't work.
> 
> Try adding M117 before you call start_print in slicer start gcode.

mate the M117 in the slicer  makes my printer start printing the file without heating anything..

AudreyAP:
@kot0005 if adding M117 before print start makes your print start not work, your config is severely broken. Swing by the discord server and we can get it working, don't want to continue spamming this pr. The feature works exactly as expected, this is a config issue.

kot0005:
> > @kot0005 if you leave the line in there, it most definitely won't work.
> > Try adding M117 before you call start_print in slicer start gcode.
> 
> mate the M117 in the slicer makes my printer start printing the file without heating anything..

Isnt the whole point of PRINT_START is to use the Gcode macros set up in the config file ? adding M117 Ignores all of this, not to mention it could damage the hotend extruder if its starts to print without heating up. Luckily mine was already heated a bit from previous test print. 

I opened a topic on discord with my config file now. I dont think there is anything wrong with it.

AudreyAP:
@kot0005 m117 simply clears the display. It absolutely does not stop print_start from working. I assume that you have misread and have replaced print_start with m117, which is not what I suggested.

kot0005:
> @kot0005 m117 simply clears the display. It absolutely does not stop print_start from working. I assume that you have misread and have replaced print_start with m117, which is not what I suggested.

i did not replace print start.. i added M117 before print start like you suggested.

zellneralex:
What we really need is a sliced gcode file, we have seen that the exclude object definition was placed below the first user defined gcode. As said the feature works as intended, for setup help and debugging of your system use the Klipper discord.

voidtrance:
>> @kot0005 m117 simply clears the display. It absolutely does not stop print_start from working. I assume that you have misread and have replaced print_start with m117, which is not what I suggested.
>
> i did not replace print start.. i added M117 before print start like you suggested.

@kot0005 Please, let's not use the PR thread for debugging issues. There are other forums better suited for this.

> What we really need is a sliced gcode file, we have seen that the exclude object definition was placed below the first user defined gcode. As said the feature works as intended, for setup help and debugging of your system use the Klipper discord.

For what it's worth, a PR was merged a couple of days ago to the preprocessing script that fixes this and the `M117` hack is no longer needed. I don't know how long it will take for it to make it's way into Moonraker's environment, however.

Arksine:
Moonraker's dependency on preprocess-cancellation has been updated.  The next time a user updates Moonraker it should be installed.

voidtrance:
> Moonraker's dependency on preprocess-cancellation has been updated. The next time a user updates Moonraker it should be installed.

Do moonraker updates through the update manager also update the virtual environment? The script is installed in there, if I am not mistaken.

Arksine:
> Do moonraker updates through the update manager also update the virtual environment? The script is installed in there, if I am not mistaken.

Yes, it will update Moonraker's virtualenv.

PR #6462, opened 2024-01-18
stm32g0: Disable internal pull-down resistors on UCPDx CCx pins, because klipper never uses UCPD

bigtreetech:
According to https://www.st.com/resource/en/datasheet/stm32g0b1cb.pdf page 55.

image

### DIFF
--- a/src/stm32/stm32g0.c
+++ b/src/stm32/stm32g0.c
@@ -162,6 +162,8 @@ bootloader_request(void)
 void
 armcm_main(void)
 {
+    // Disable internal pull-down resistors on UCPDx CCx pins
+    SYSCFG->CFGR1 |= (SYSCFG_CFGR1_UCPD1_STROBE | SYSCFG_CFGR1_UCPD2_STROBE);
     SCB->VTOR = (uint32_t)VectorTable;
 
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Seems fine to me.  I'll give a few days to see if there are any other comments.

-Kevin

moggieuk:
Kevin, I just want to comment on the urgency of this.   With the launch of the ERCFv2 project we are advocating the use of the BTT MMB mcu.  This uses the stm32g0.  Without special instruction the default firmware install is causing all kinds of overheating problems with connected steppers.  We have burn out 5 steppers on the test team alone, one of which deformed the ABS mounts.  Many community members have also run into the problem.

Obviously we are trying to get the word out on the care need whilst setting up the mcu but obviously a fix in Klipper soon to minimize the exposure window for what is a safety concern.

Thanks for listening 🙏

KevinOConnor:
Okay, thanks for the info.  I'll target this weekend for the merge.

-Kevin

KevinOConnor:
Thanks.

-Kevin

moggieuk:
Thank you :pray:

PR #6463, opened 2024-01-19
Make MCU Protocol mismatch error message more specific

KevinOConnor:
There is a good chance that an "MCU Protocol Error" may not require all MCUs to be updated, but the current error message seems to indicate that all MCUs that don't exactly match the current version should be updated.  That may be misleading.  Try to word the message to be a little more precise.

There is a risk this message may be more confusing for users, in that it doesn't explicitly state which MCUs to update.

@meteyou , @Drachenkaetzchen - FYI.

-Kevin

### DIFF
--- a/klippy/klippy.py
+++ b/klippy/klippy.py
@@ -166,8 +166,8 @@ class Printer:
         msg = ["MCU Protocol error",
                message_protocol_error1,
                "Your Klipper version is: %s" % (host_version,),
-               "MCU(s) which should be updated:"]
-        msg += msg_update + ["Up-to-date MCU(s):"] + msg_updated
+               "MCU(s) reporting a different version:"]
+        msg += msg_update + ["MCU(s) reporting the same version:"] + msg_updated
         msg += [message_protocol_error2, str(e)]
         return "\n".join(msg)
     def _connect(self, eventtime):


meteyou:
If you take the error message exactly, it is fixed by this PR, but unfortunately, I don't see any improvement here with the "-dirty" tag, which is currently a real problem.

thijstriemstra:
I updated just now to latest klipper and getting:

```
MCU Protocol error

This is frequently caused by running an older version of the
firmware on the MCU(s). Fix by recompiling and flashing the
firmware.

Your Klipper version is: v0.12.0-85-gd785b396-dirty
MCU(s) which should be updated:
mcu: Current version v0.12.0-60-g0665dc89
ext1: Current version v0.12.0-60-g0665dc89
adxl: Current version v0.12.0-60-g0665dc89
Up-to-date MCU(s):


mcu 'ext1': Command format mismatch: query_adxl345 oid=%c rest_ticks=%u vs query_adxl345 oid=%c clock=%u rest_ticks=%u
```

Guess this is new/normal? Will reflash..

meteyou:
> after updating the EBB42 (ext1) the error has not changed though, and I'd expect ext1 to be in the list at bottom. Probably related to the -dirty thing?

yes exactly. that's what i'm talking about in my post above.

quilt-jdockter:
I thought the -dirty string would appear in the mcu version if it was flashed from firmware compiled from a dirty version.  At least in my experience that is what happens.  @thijstriemstra can you confirm that you compiled the firmware from a -dirty version and when you flashed it the -dirty flag disappeared from the mcu version?

dockterj:
I'm not a reviewer, however, I reviewed this change according to the steps in the CONTRIBUTING document and everything looks good to me.

@meteyou - when I compile and flash my mcus (AVR) from a dirty repo the version on the mcu reports with the -dirty suffix.  Is there a discord or discourse discussion of this issue that I can refer to?  @thijstriemstra I'm confused that your ext1 mcu doesn't have the -dirty suffix on the reported version.

meteyou:
sry @dockterj, we talked about that in a non public discord channel before this PR.

KevinOConnor:
Thanks for the feedback.  There doesn't seem to be much interest in this PR (as it is currently implemented).  So, I'm going to close it.

-Kevin

PR #6466, opened 2024-01-22
Fix initial rp2040 SPI transfers

KevinOConnor:
The rp2040 code was not initializing the SCLK line prior to the first SPI transfer.

This issue is likely the cause of common reports of `Invalid adxl345 id (got xx vs e5)` error on the first (and only first) adxl345 query.

@Sineos, @buzztiaan - FYI.

-Kevin


Sineos:
Tested and works 👍

PR #6467, opened 2024-01-25
Update gcode_move.py

CooperGerman:
Updated gcode save and restore commands in order to stack up states until a max is reached and pop those states when restore commands are called.

This allows for macros to be called recursively without overwriting the previously saved state.  (in my case using klipper-macros from https://github.com/jschuh/klipper-macros with modified ercf_software.cfg from https://github.com/EtteGit/EnragedRabbitProject)


CooperGerman:
Hi, i see that the builtin commands.test fails on restoring twice the 'default' gcode state.
My understanding is that the way it normally works is that you can restore a previously saved state mutliple times as the restore_gcode code does not remove it when beeing restored.

In my version this no longer is possible... What expected behavior should the save/restore commands have ?
Is my proposition coherent with what klipper wants to be doing ?

PR #6468, opened 2024-01-25
ci: update & simplify docs contributing - docs: update + cosmetics

docgalaxyblock:
My start focus was to update all sites regarding the CanBoot -> katapult transition.
Stuff I catched on the lines/pages I fixed too.

Changed site can be viewed here: https://docgalaxyblock.github.io/klipper
_Note: It looks like some cometic customization stuff got not applied_

I stopped my work today since the used mkdocs-material version is not uptodate and the official mkdocs-material docs differ a bit.
Will try to update mkdocs-material without breaking anything and then update more pages.

Greetings
Doc

### DIFF
--- a/.github/workflows/klipper3d-deploy.yaml
+++ b/.github/workflows/klipper3d-deploy.yaml
@@ -12,12 +12,12 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: '3.8'
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('docs/_klipper3d/mkdocs-requirements.txt') }}
--- a/.github/workflows/klipper3d-deploy.yaml
+++ b/.github/workflows/klipper3d-deploy.yaml
@@ -1,4 +1,5 @@
 name: klipper3d deploy
+
 on:
   schedule:
     - cron: "0 0 * * *"
@@ -8,11 +9,25 @@ on:
     paths:
       - docs/**
       - .github/workflows/klipper3d-deploy.yaml
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Documentation page deployment source branch (eg: "pr-docs")'
+        required: true
+        default: master
+
+# Required for the deployment to be able to upload to the "gh-pages" branch
+# (since the repo settings for default actions permission defaults to read only)
+permissions:
+  contents: write
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.branch }}
       - name: Setup python
         uses: actions/setup-python@v5
         with:
--- a/docs/Bootloader_Entry.md
+++ b/docs/Bootloader_Entry.md
@@ -21,7 +21,7 @@ Entering bootloader on <DEVICE>
 ```
 
 Where `<DEVICE>` is your serial device, such as
-`/dev/serial.by-id/usb-Klipper[...]` or `/dev/ttyACM0`
+`/dev/serial/by-id/usb-Klipper[...]` or `/dev/ttyACM0`
 
 Note that if this fails, no output will be printed, success is indicated by
 printing `Entering bootloader on <DEVICE>`.
@@ -34,7 +34,7 @@ picocom -b 1200 <DEVICE>
 ```
 
 Where `<DEVICE>` is your serial device, such as
-`/dev/serial.by-id/usb-Klipper[...]` or `/dev/ttyACM0`
+`/dev/serial/by-id/usb-Klipper[...]` or `/dev/ttyACM0`
 
 `<Ctrl-A><Ctrl-P>` means
 holding `Ctrl`, pressing and releasing `a`, pressing and releasing `p`, then
@@ -48,8 +48,7 @@ is being used to connect to it), sending the string
 
 `<SPACE>` is an ASCII literal space, 0x20.
 
-`<FS>` is the ASCII File Separator,
-0x1c.
+`<FS>` is the ASCII File Separator, 0x1c.
 
 Note that this is not a valid message as per the
 [MCU Protocol](Protocol.md#micro-controller-interface), but sync characters(`~`)
@@ -69,8 +68,7 @@ echo $'~ \x1c Request Serial Bootloader!! ~' >> /dev/<DEVICE>
 Where `<DEVICE>` is your serial port, such as `/dev/ttyS0`, or
 `/dev/serial/by-id/gpio-serial2`, and
 
-`<BAUD>` is the baud rate of the serial
-port, such as `115200`.
+`<BAUD>` is the baud rate of the serial port, such as `115200`.
 
 ### CANBUS
 
--- a/docs/Bootloader_Entry.md
+++ b/docs/Bootloader_Entry.md
@@ -83,7 +83,7 @@ This method also applies to devices operating in
 #### Katapult's flashtool.py
 
 ```shell
-python3 ./katapult/scripts/flashtool.py -i <CAN_IFACE> -u <UUID> -r
+python3 ~/katapult/scripts/flashtool.py -i <CAN_IFACE> -u <UUID> -r
 ```
 
 Where `<CAN_IFACE>` is the can interface to use. If using `can0`, both the `-i`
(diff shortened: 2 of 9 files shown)


docgalaxyblock:
I did not touched the `lib/canboot` folder. May that be renamed too and a symlink added for legacy docs forks?

@Arksine The `lib/README` lists a rev I could not find in your katapult repo - how old is the `flash_can.py` in klipper?
Are there be any changes making an update a benefit for the user?

KevinOConnor:
Thanks.  I appreciate you taking the time to review and make fixes to the documentation.

As high-level feedback, there are a lot of changes in this PR and a handful of changes look incorrect to me.  That makes it hard to review and provide feedback.

I appreciate that you've split this PR into commits (as that helps review).  I did see some commits where cosmetic changes (eg, whitespace fixes) were mixed with functional changes (eg, a different path) - that makes it hard to review.

I'll follow up with a few things that I noticed during review.

-Kevin

.github/workflows/klipper3d-deploy.yaml
```diff
@@ -8,16 +9,30 @@ on:
...
+    inputs:
+      branch:
+        description: 'Documentation page deployment source branch (eg: "pr-docs")'
+        required: true
+        default: master
+
+# Required for the deployment to be able to upload to the "gh-pages" branch
+# (since the repo settings for default actions permission defaults to read only)
+permissions:
+  contents: write
```
KevinOConnor, line 22: I'm leery of making a permission change like this.  My fear is that one of the many build scripts could alter the master branch (which would then propagate to thousands of printers to within a few days).  Not sure the trade-off on this commit is worthwhile - might be better to just add documentation on how one can build the documentation (I use a docker instance, for example).

docs/Bootloader_Entry.md
```diff
@@ -101,25 +99,29 @@ for information on finding the CAN UUID of your devices.
...
+
 ## Notes
 
 ### STM32 DFU Warning
 
-Note that on some boards, like the Octopus Pro v1, entering DFU mode can cause
-undesired actions (such as powering the heater while in DFU mode). It is
-recommended to disconnect heaters, and otherwise prevent undesired operations
-when using DFU mode. Consult the documentation for your board for more details.
+!!! danger "STM32 DFU Warning"
```
KevinOConnor, line 121: Heh, a line, a header, another header, and a danger..  I think we may be overdoing this a bit.  :-)

docs/Bootloaders.md
```diff
@@ -452,78 +490,45 @@ The bootloader can be activated by pressing the reset button of the board twice.
...
-
-The `flash_can.py` utility supplied in the `lib/canboot` folder may be used to
-upload Klipper firmware.  The device UUID is necessary to flash.  If you do not
-have a UUID it is possible to query nodes currently running the bootloader:
-```
-python3 flash_can.py -q
-```
-This will return UUIDs for all connected nodes not currently assigned a UUID.
-This should include all nodes currently in the bootloader.
+## STM32F4 micro-controllers with DFU bootloader
```
KevinOConnor, line 493: This section discusses hidflash - so changing this to say "DFU bootloader" is incorrect.

docs/Bootloaders.md
```diff
@@ -553,8 +558,8 @@ Begin by downloading and compiling the software (each step may take
 several minutes and the "make" step may take 30+ minutes):
 
 ```
-sudo apt-get update
-sudo apt-get install autoconf libtool telnet
+sudo apt update
+sudo apt install autoconf libtool telnet
```
KevinOConnor, line 562: I'm not sure about the implications of making this change.  As far as I know, `apt-get` is always installed, but I'm not sure `apt` is always installed.

docs/FAQ.md
```diff
@@ -11,33 +11,39 @@ See the [rotation distance document](Rotation_Distance.md).
 
 ## Where's my serial port?
 
-The general way to find a USB serial port is to run `ls
-/dev/serial/by-id/*` from an ssh terminal on the host machine. It will
-likely produce output similar to the following:
+The general way to find a USB serial port is to run
+``` shell
```
KevinOConnor, line 15: It seems odd to place a space between the backticks and the designator - all the other places in the docs do not place a space there.

docs/FAQ.md
```diff
@@ -11,33 +11,39 @@ See the [rotation distance document](Rotation_Distance.md).
...
+it in the config file and while flashing the micro-controller code.
+For example, a flash command might look similar to:
+``` shell
 sudo service klipper stop
 make flash FLASH_DEVICE=/dev/serial/by-id/usb-1a86_USB2.0-Serial-if00-port0
 sudo service klipper start
 ```
 and the updated config might look like:
-```
+``` yaml
```
KevinOConnor, line 33: It seems odd to mark "ini" file content as "yaml".

docs/FAQ.md
```diff
@@ -210,26 +217,28 @@ information.
...
-
-Long answer: In practice Marlin can typically only step at a rate of
-around 10000 steps per second. If it is requested to move at a speed
-that would require a higher step rate then Marlin will generally just
-step as fast as it can. Klipper is able to achieve much higher step
-rates, but the stepper motor may not have sufficient torque to move at
-a higher speed. So, for a Z axis with a high gearing ratio or high
-microsteps setting the actual obtainable max_z_velocity may be smaller
-than what is configured in Marlin.
+Short answer: \
```
KevinOConnor, line 220: I don't think it is a good idea to end lines with a backslash - it isn't portable markdown and it looks confusing to users reading the text directly.

docs/SDCard_Updates.md
```diff
@@ -41,7 +41,7 @@ board definition as [described below](#board-definitions).
 ## Advanced Usage
 
 The above commands assume that your MCU connects at the default baud rate
-of 250000 and the firmware is located at `~/klipper/out/klipper.bin`.  The
+of 250000 and the firmware is located at `out/klipper.bin`.  The
```
KevinOConnor, line 44: This is explicitly documenting that the tool looks at an absolute path.  The tool does not look at `out/klipper.bin` so this change would be an error.  In general, it's probably best to not change the paths in the docs unless they are obviously incorrect.

docs/FAQ.md
```diff
@@ -210,26 +217,28 @@ information.
...
+configured in Marlin.
 
 ## My TMC motor driver turns off in the middle of a print
 
 If using the TMC2208 (or TMC2224) driver in "standalone mode" then
 make sure to use the
-[latest version of Klipper](#how-do-i-upgrade-to-the-latest-software). A
-workaround for a TMC2208 "stealthchop" driver problem was added to
+[latest version of Klipper](#how-do-i-upgrade-to-the-latest-software).
+An workaround for a TMC2208 "stealthchop" driver problem was added to
```
KevinOConnor, line 241: `A` got incorrectly changed to `An`

.github/workflows/klipper3d-deploy.yaml
docgalaxyblock: Ok but your current setup is less secure.
Without write permissions all current workflows in this repo should fail. Therefore I guess at the repo or organization level the default github workflow permission setting is set to "read & write".
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token



If a threaded harmfull gets merged now it has write permissions to everything.

If your current default token permissions would be set to read only [the pages deployment action](https://github.com/Klipper3d/klipper/blob/72b301a2859c3f7ed26d802dd52fc495eef6c353/.github/workflows/klipper3d-deploy.yaml#L31) would error like this:
```
Error: The deploy step encountered an error: The process '/usr/bin/git' failed with exit code 128 ❌
```

docs/Bootloader_Entry.md
docgalaxyblock: Year but otherwise without the heading it does not get listed at the TOC.
And iirc the current used mkdocs-material version does not support amonitions without title.

docs/FAQ.md
docgalaxyblock: I am following the [mkdos-material guide] here
(https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#usage)

docs/FAQ.md
docgalaxyblock: Year but a backslash or two whitespaces are the only methods I know of making a linebreak without a heel breaking _(if this is the right word)_

Eg: <kbd>Return</kbd> vs <kbd>Shift</kbd> + <kbd>Return</kbd>

docs/FAQ.md
docgalaxyblock: fixed

docs/FAQ.md
docgalaxyblock: Source:
https://spec.commonmark.org/0.31.2/#hard-line-break
https://www.markdownguide.org/basic-syntax/#line-breaks

docs/SDCard_Updates.md
docgalaxyblock: I have reverted this one - I used "replace all" without checking the script. Sry

docs/FAQ.md
docgalaxyblock: I have replaced " \" by "<br>" as the first source suggests it.

docs/FAQ.md
docgalaxyblock: fixed - sry I guessed from the vscode code highlight which one it could be.

docs/Bootloaders.md
docgalaxyblock: I removed the bootloader in the header

docgalaxyblock:
I hope I have got everything now.

Please give your updated opinion on the still open review comments.

Happy 🐇🥚
Doc

Issue #1438, opened 2019-03-22
Klipper "data dictionaries" for regression tests


KevinOConnor:
This is not an issue.  This is a place holder to store Klipper data dictionaries that are used during the regression tests.

To run the regression tests, one would typically download the file below to the home directory of an RPi and then run:
```
cd ~/klipper
tar xfz ~/klipper_dict.tar.gz
~/klippy-env/bin/python scripts/test_klippy.py -d dict/ test/klippy/*.test
```

[klipper_dict.tar.gz](https://github.com/KevinOConnor/klipper/files/2997471/klipper_dict.tar.gz)

PR #6470, opened 2024-01-26
virtual_sdcard: Add an option to sort by date

nhchiu:
Added an option to sort the files with modified date in reverse, 
so the most recently uploaded one is at the top.

It should be useful for many LCD screen users so that they can 
easily reprint the latest file when they are at the printer.
Especially if there are a lot of g-code files or not named properly.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1465,6 +1465,10 @@ path:
 #   be provided.
 #on_error_gcode:
 #   A list of G-Code commands to execute when an error is reported.
+#recent_first: False
+#   Set to True to sort the files with modified date in reverse, so the
+#   most recently uploaded one is at the top. Otherwise they are
+#   sorted alphabetically. The default is False.
 
 ```
 
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -17,6 +17,7 @@ class VirtualSD:
         self.sdcard_dirname = os.path.normpath(os.path.expanduser(sd))
         self.current_file = None
         self.file_position = self.file_size = 0
+        self.recent_first = config.get("recent_first", False)
         # Print Stat Tracking
         self.print_stats = self.printer.load_object(config, 'print_stats')
         # Work timer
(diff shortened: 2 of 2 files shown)


nhchiu:
### Self review

1. Tested on my own printer with and without the option set, and they both worked correctly.
	Result of running [regression test cases](https://www.klipper3d.org/Debugging.html)
 	- `whitespace check` passed
 	- regression test with dictionaries 20230921 from [here](https://github.com/Klipper3d/klipper/issues/1438) failed at `input_shaper.test`.
 	  I assume it's caused by the recent bulk sensor update and the test dictionary is out of date.

```
Protocol error
  File "/home/pi/klipper/./klippy/klippy.py", line 180, in _connect
    cb()
  File "/home/pi/klipper/klippy/mcu.py", line 723, in _connect
    self._send_config(None)
  File "/home/pi/klipper/klippy/mcu.py", line 653, in _send_config
    cb()
  File "/home/pi/klipper/klippy/extras/adxl345.py", line 228, in _build_config
    self.query_adxl345_cmd = self.mcu.lookup_command(
  File "/home/pi/klipper/klippy/mcu.py", line 839, in lookup_command
    return CommandWrapper(self._serial, msgformat, cq)
  File "/home/pi/klipper/klippy/mcu.py", line 86, in __init__
    self._cmd = msgparser.lookup_command(msgformat)
  File "/home/pi/klipper/klippy/msgproto.py", line 317, in lookup_command
    self._error("Command format mismatch: %s vs %s",
  File "/home/pi/klipper/klippy/msgproto.py", line 243, in _error
msgproto.error: mcu 'mcu': Command format mismatch: query_adxl345 oid=%c rest_ticks=%u vs query_adxl345 oid=%c clock=%u rest_ticks=%u
```
	
2. As stated above, I think it's beneficial to LCD screen users that don't have PC or smartphone at the place the printer is. 

3. No new files created and only minor modifications to an existing class. 

4. The new config option is updated in `Config_Reference.md`.
   If not specified it falls back to the default action as before so probably not needed to be documented in `Config_Changes.md`

galagithub:
@nhchiu don't bother, there are several patches made for this before, but the project manager does not want this feature.

scouring the web, you need to be part of some discord channel to get any of your commits looked after.

PR #6459, opened 2024-01-13
Remove previously deprecated features

KevinOConnor:
This PR removes features that we have previously deprecated.  I'll look to make this change in about a month.  Details are in the PR:

Several deprecated features have been removed. Using "NTC 100K beta 3950" as a thermistor name has been removed (deprecated on 20211110). The `SYNC_STEPPER_TO_EXTRUDER` and `SET_EXTRUDER_STEP_DISTANCE` commands have been removed, and the extruder `shared_heater` config option has been removed (deprecated on 20220210). The bed_mesh `relative_reference_index` option has been removed (deprecated on 20230619).

-Kevin

PR #6473, opened 2024-01-27
bed_mesh: behavior fixes

Arksine:
This includes two minor fixes to the behavior of bed_mesh.

1) Avoid probing additional points for the `zero_reference_position` and `faulty_regions` when the user chooses the "manual" method.  Those options aren't useful when probing manually.

2) Fix a reporting regression introduced by the adaptive mesh change.  Since we avoid saving the mesh, the value reported in the `profile_name` field will be whatever was previously stored in `ProfileManger.current_profile`.   This is often an empty string, which is leading some frontends to believe that no mesh is loaded.  This patch moves profile name tracking to the ZMesh class and generates a placeholder name for adaptive meshes.

Kevin, this is likely going to cause a merge conflict with #6459, so if you would like we can hold this PR until its merged and I will resolve the conflicts.

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -233,7 +233,7 @@ class BedMesh:
             mesh_max = (params['max_x'], params['max_y'])
             probed_matrix = self.z_mesh.get_probed_matrix()
             mesh_matrix = self.z_mesh.get_mesh_matrix()
-            self.status['profile_name'] = self.pmgr.get_current_profile()
+            self.status['profile_name'] = self.z_mesh.get_profile_name()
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I don't think we should wait for #6459 - I'll rebase that PR as needed.  Let me know when you'd like me to commit this PR.

-Kevin

kyleisah:
Wow awesome work @Arksine with the quick fix!

Arksine:
Sounds good.  I have tested the changes here and everything works as expected.  You can commit when ready.

KevinOConnor:
Thanks.

-Kevin

PR #6476, opened 2024-01-29
Clarify on some requirements for adaptive meshing

kyleisah:
Just wanted to clarify that adaptive meshing relies on objects being defined by the exclude_object module, as well as some other small changes for better readability. I'm open to any other changes recommended to better fit the formatting of the rest of the documentation. Thank you!

### DIFF
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -439,7 +439,7 @@ that the objects are being defined before the adaptive bed mesh command is execu
 - You must have the [exclude_object](Exclude_Object.md) module enabled by adding `[exclude_object]` to your `printer.cfg` file.
 - You must enable object labeling in your slicer.
   - Certain slicers may do this by default, or do not provide an option because it is an "always-on" feature.
-  - More modern slicers provide a "Firmware-Specific" option for labeling object. If you enable this, you'll need to verify that your slicer supports the `klipper` gcode flavor, and is generating `EXCLUDE_OBJECT_DEFINE` commands at the beginning of the sliced gcode file.
+  - More modern slicers provide a "Firmware-Specific" option for labeling objects. If you enable this, you'll need to verify that your slicer supports the `klipper` gcode flavor, and is generating `EXCLUDE_OBJECT_DEFINE` commands at the beginning of the sliced gcode file.
 - If your slicer does not support firmware-specific object labeling, or does not support the `klipper` gcode flavor, you will need to [configure Moonraker's file manager module](https://moonraker.readthedocs.io/en/latest/configuration/#file_manager) to enable object processing.
 - Finally, to generate an adaptive bed mesh, you must append the `ADAPTIVE=1` flag to `BED_MESH_CALIBRATE` wherever you normally call for a bed mesh to be generated.
   - The command should be `BED_MESH_CALIBRATE ADAPTIVE=1` wherever you normally request it.
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -417,7 +417,7 @@ adaptive_margin: 5
 
 #### Caveats:
 
-By nature, adaptive bed meshes use the objects defined by the [exclude_object](Exclude_Object.md) module 
+By nature, adaptive bed meshes use the objects defined by the [exclude_object](Exclude_Object.md) module
 in the Gcode file being printed, and it is expected that each Gcode file will generate a mesh that probes a different
 area of the print bed. Therefore, adapted bed meshes should **not** be re-used. The expectation
 is that a new mesh will be generated for each print if adaptive meshing is used.
@@ -425,13 +425,13 @@ is that a new mesh will be generated for each print if adaptive meshing is used.
 It is important to consider that adaptive bed meshing is best used on machines that can
 normally probe the entire bed and achieve a maximum variance less than or equal to 1 layer
 height. Machines with mechanical issues that a full bed mesh normally compensates for may
-have undesirable results when attempting print moves **outside** of the probed area. If you 
-normally have a variance greater than 1 layer height when generating a full bed mesh, greater caution 
+have undesirable results when attempting print moves **outside** of the probed area. If you
+normally have a variance greater than 1 layer height when generating a full bed mesh, greater caution
 must be taken when attempting print moves outside of the adapted mesh area.
 
-If there are no objects defined by the [exclude_object](Exclude_Object.md) module before the adaptive 
-bed mesh is requested, the default method will be used instead. If you are requesting an adaptive bed mesh and 
-the default method is still being used, check that you have [exclude_object](Exclude_Object.md) set up correctly, and 
+If there are no objects defined by the [exclude_object](Exclude_Object.md) module before the adaptive
+bed mesh is requested, the default method will be used instead. If you are requesting an adaptive bed mesh and
+the default method is still being used, check that you have [exclude_object](Exclude_Object.md) set up correctly, and
 that the objects are being defined before the adaptive bed mesh command is executed in your Gcode file.
 
 #### How to use Adaptive Meshes:
@@ -441,7 +441,7 @@ that the objects are being defined before the adaptive bed mesh command is execu
   - Certain slicers may do this by default, or do not provide an option because it is an "always-on" feature.
   - More modern slicers provide a "Firmware-Specific" option for labeling object. If you enable this, you'll need to verify that your slicer supports the `klipper` gcode flavor, and is generating `EXCLUDE_OBJECT_DEFINE` commands at the beginning of the sliced gcode file.
 - If your slicer does not support firmware-specific object labeling, or does not support the `klipper` gcode flavor, you will need to [configure Moonraker's file manager module](https://moonraker.readthedocs.io/en/latest/configuration/#file_manager) to enable object processing.
-- Finally, to generate an adaptive bed mesh, you must append the `ADAPTIVE=1` flag to `BED_MESH_CALIBRATE` wherever you normally call for a bed mesh to be generated. 
+- Finally, to generate an adaptive bed mesh, you must append the `ADAPTIVE=1` flag to `BED_MESH_CALIBRATE` wherever you normally call for a bed mesh to be generated.
   - The command should be `BED_MESH_CALIBRATE ADAPTIVE=1` wherever you normally request it.
 
 ## Bed Mesh Gcodes
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -372,11 +372,15 @@ are identified in green.
 
 ### Adaptive Meshes
 
+#### Concept:
+
 Adaptive bed meshing is a way to speed up the bed meshing process as well as waste less mesh information by only probing
 the area of the bed that will be used for printing. When used, this method will
 automatically adjust the mesh parameters based on the area occupied by objects defined by the `EXCLUDE_OBJECT` command. See the [exclude objects guide](Exclude_Object.md) and
 [command reference](G-Codes.md#excludeobject)
 for additional information.
+
+#### How it works:
 
 The adapted mesh area will be computed from the area defined by the boundaries of all
 the defined print objects so it covers every object, including any margins defined in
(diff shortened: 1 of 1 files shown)


Sineos:
According to your KAMPS GH repository, setting
```
[file_manager]
enable_object_processing: True
```
is also a prerequisite. Is this still true for the new implementation?

kyleisah:
> According to your KAMPS GH repository, setting
> 
> ```
> [file_manager]
> enable_object_processing: True
> ```
> 
> is also a prerequisite. Is this still true for the new implementation?

Yes and no, as some slicers have the ability to do it without moonraker needing to process the gcode file. It’s also a setting in moonraker and not klipper, so I wasn’t really sure of the best way to mention all of that. I left it in the realm of “make sure objects are being defined before the command is called” but I guess a new user might get hung up on that.

Sineos:
Moonraker today is the de facto way to use Klipper. Likely, the numbers of Octoprint users are dwindling more and more, if not already insignificant.
If this feature needs an interaction with Moonraker, even if only in some cases, I would vote for mentioning it. Also, in particular, since this Moonraker setting defaults to false. A regular user will not find out, why his setup is not working otherwise.

kyleisah:
@Sineos 
Oh wow, so I just double checked the exclude object modules readme here and there’s no mention of defining [file_manager] in moonraker or anything like that. 

I agree that these things should be mentioned and laid out, but I want them to be in places that make sense. I can make some updates to the .md for exclude_object to include setting it up in moonraker, or have it point to moonraker’s documentation for it. What do you think is best?

kyleisah:
Perhaps I could also expand on the adaptive meshing section a bit further with a small table of prerequisites, similar to the documentation for KAMP

Sineos:
I'm not the expert, but if `enable_object_processing: True` is a general prerequisite for successfully using `[exclude_object]`, then I would mention it in both places, i.e.:
* https://github.com/Klipper3d/klipper/blob/master/docs/Exclude_Object.md?plain=1
* https://github.com/Klipper3d/klipper/blob/9f41f53c5e364694b9b41279b3b3aee34250b93a/docs/Bed_Mesh.md?plain=1#L373

Something like:
> Using Moonraker requires `enable_object_processing: True` in the [[file_manager]](https://moonraker.readthedocs.io/en/latest/configuration/#file_manager) section of the `moonraker.conf` file.

docs/Bed_Mesh.md
```diff
@@ -372,14 +372,15 @@ are identified in green.
 
 ### Adaptive Meshes
 
-Adaptive bed meshing is a way to speed up the bed mesh generation by only probing
-the area of the bed used by the objects being printed. When used, the method will
-automatically adjust the mesh parameters based on the area occupied by the defined
-print objects.
+Adaptive bed meshing is a way to speed up the bed meshing process as well as waste less mesh information by only probing
+the area of the bed that will be used for printing. When used, this method will
```
Sineos, line 376: > When used, this method will [...]

What I find a bit unclear today: What is the criteria, when adaptive meshing is used? Is it just the presence of `[exclude_object]` along with named objects in the gcode? What if I wanted to not use adaptive meshing? Would I need to turn off naming objects in the slicer or take out `[exclude_object]`?

docs/Bed_Mesh.md
```diff
@@ -410,17 +411,16 @@ adaptive_margin: 5
...
 normally probe the entire bed and achieve a maximum variance less than or equal to 1 layer
 height. Machines with mechanical issues that a full bed mesh normally compensates for may
-have undesirable results when attempting print moves **outside** of the probed area. If a
-full bed mesh has a variance greater than 1 layer height, caution must be taken when using
-adaptive bed meshes and attempting print moves outside of the meshed area.
+have undesirable results when attempting print moves **outside** of the probed area. If you normally have a variance greater than 1 layer height when generating a full bed mesh, greater caution must be taken when attempting print moves outside of the adapted mesh area.
+
+If there are no objects defined by the [exclude_object](Exclude_Object.md) module before the adaptive bed mesh is requested, the default method will be used instead. If you are requesting an adaptive bed mesh and the default method is still being used, check that you have [exclude_object](Exclude_Object.md) set up correctly, and that the objects are being defined before the adaptive bed mesh command is executed in your Gcode file.
```
Sineos, line 423: See the comment above

docs/Bed_Mesh.md
kyleisah: if you wanted to not use adaptive meshing, you would just call BED_MESH_CALIBRATE without the ADAPTIVE=1 flag on the command

Sineos: What I wanted to point out:
* Using `ADAPTIVE=1` is not documented in the Adaptive Mesh Chapter (https://github.com/Klipper3d/klipper/blob/master/docs/Bed_Mesh.md#adaptive-meshes)
* It is somewhat "hidden" documented in the next chapter (https://github.com/Klipper3d/klipper/blob/master/docs/Bed_Mesh.md#calibration)
* It is not documented in https://github.com/Klipper3d/klipper/blob/master/docs/G-Codes.md#bed_mesh_calibrate

jernejp21:
I really appreciate you are documenting everything. I just found out today for adaptive mesh functionality and wanted to test it out, but reading documentation I had no success. I tried to put `[exclude_object]` in printer.cfg, but still whole bed is probed. 

I would appreciate more details what is expected in config files or slicer to correctly set up g-code for parsing. I know this is brand new feature and it's early. I just wanted to express that some info is still missing.

docs/Bed_Mesh.md
kyleisah: @Sineos I've just made some updates to the bed_mesh.md for clarification on setup requirements. If this looks okay, I can move forward with making changes to the exclude_object.md about setting up object processing with moonraker.

docs/Bed_Mesh.md
```diff
@@ -410,17 +415,34 @@ adaptive_margin: 5
...
+that the objects are being defined before the adaptive bed mesh command is executed in your Gcode file.
+
+#### How to use Adaptive Meshes:
+
+- You must have the [exclude_object](Exclude_Object.md) module enabled by adding `[exclude_object]` to your `printer.cfg` file.
+- You must enable object labeling in your slicer.
+  - Certain slicers may do this by default, or do not provide an option because it is an "always-on" feature.
+  - More modern slicers provide a "Firmware-Specific" option for labeling objects. If you enable this, you'll need to verify that your slicer supports the `klipper` gcode flavor, and is generating `EXCLUDE_OBJECT_DEFINE` commands at the beginning of the sliced gcode file.
+- If your slicer does not support firmware-specific object labeling, or does not support the `klipper` gcode flavor, you will need to [configure Moonraker's file manager module](https://moonraker.readthedocs.io/en/latest/configuration/#file_manager) to enable object processing.
+- Finally, to generate an adaptive bed mesh, you must append the `ADAPTIVE=1` flag to `BED_MESH_CALIBRATE` wherever you normally call for a bed mesh to be generated.
```
Sineos, line 444: I think a simple wording like

> Call `BED_MESH_CALIBRATE ADAPTIVE=1` either from your `[START_PRINT](Slicers.md#klipper-gcode_macro)` macro or from your slicer's starting Gcode. It is recommended to place this command pretty much at the end of the start-up routine, this means at least after heating up the bed.

would be clearer as these should be the two typical places for a regular user.

docs/Bed_Mesh.md
```diff
@@ -410,17 +415,34 @@ adaptive_margin: 5
...
+the default method is still being used, check that you have [exclude_object](Exclude_Object.md) set up correctly, and
+that the objects are being defined before the adaptive bed mesh command is executed in your Gcode file.
+
+#### How to use Adaptive Meshes:
+
+- You must have the [exclude_object](Exclude_Object.md) module enabled by adding `[exclude_object]` to your `printer.cfg` file.
+- You must enable object labeling in your slicer.
+  - Certain slicers may do this by default, or do not provide an option because it is an "always-on" feature.
+  - More modern slicers provide a "Firmware-Specific" option for labeling objects. If you enable this, you'll need to verify that your slicer supports the `klipper` gcode flavor, and is generating `EXCLUDE_OBJECT_DEFINE` commands at the beginning of the sliced gcode file.
+- If your slicer does not support firmware-specific object labeling, or does not support the `klipper` gcode flavor, you will need to [configure Moonraker's file manager module](https://moonraker.readthedocs.io/en/latest/configuration/#file_manager) to enable object processing.
```
Sineos, line 443: > - If your slicer does not support firmware-specific object labeling, or does not support the `klipper` gcode flavor, you will need to [configure Moonraker's file manager module](https://moonraker.readthedocs.io/en/latest/configuration/#file_manager) to enable object processing.

...to activate `enable_object_processing`

Sineos (approved):
Nice work 👍 

Just two minor nit picks.

docs/Bed_Mesh.md
kyleisah: Sure, that makes sense. I’ll find a spot to work something like this in as well.

docs/Bed_Mesh.md
```diff
@@ -372,14 +372,19 @@ are identified in green.
 
 ### Adaptive Meshes
 
-Adaptive bed meshing is a way to speed up the bed mesh generation by only probing
-the area of the bed used by the objects being printed. When used, the method will
-automatically adjust the mesh parameters based on the area occupied by the defined
-print objects.
+#### Concept:
+
+Adaptive bed meshing is a way to speed up the bed meshing process as well as waste less mesh information by only probing
```
voidtrance, line 377: What does "waste less mesh information" mean? What information is being wasted if adaptive is not being used?

Also, a nit: can you please conform to the document's line length?

docs/Bed_Mesh.md
```diff
@@ -372,14 +372,19 @@ are identified in green.
...
+the area of the bed that will be used for printing. When used, this method will
+automatically adjust the mesh parameters based on the area occupied by objects defined by the `EXCLUDE_OBJECT` command. See the [exclude objects guide](Exclude_Object.md) and
```
voidtrance, line 379: I actually think that the old version ("area occupied by the defined print objects") was a bit better. Talking about how those objects get defined (the `EXCLUDE_OBJECTS` command) seem more appropriate for the "How it works" section.

docs/Bed_Mesh.md
```diff
@@ -410,17 +415,34 @@ adaptive_margin: 5
...
 is that a new mesh will be generated for each print if adaptive meshing is used.
 
-It is also important to consider that adaptive bed meshing is best used on machines that can
+It is important to consider that adaptive bed meshing is best used on machines that can
 normally probe the entire bed and achieve a maximum variance less than or equal to 1 layer
 height. Machines with mechanical issues that a full bed mesh normally compensates for may
-have undesirable results when attempting print moves **outside** of the probed area. If a
-full bed mesh has a variance greater than 1 layer height, caution must be taken when using
-adaptive bed meshes and attempting print moves outside of the meshed area.
+have undesirable results when attempting print moves **outside** of the probed area. If you
```
voidtrance, line 428: I generally prefer documentation be written in a "neutral" way, without using pronouns. Not sure how to explain it so I'll give an example:

Instead of "If you normally have ...", say "If a printer normally has ...".

docs/Bed_Mesh.md
```diff
@@ -410,17 +415,34 @@ adaptive_margin: 5
...
+normally have a variance greater than 1 layer height when generating a full bed mesh, greater caution
+must be taken when attempting print moves outside of the adapted mesh area.
+
+If there are no objects defined by the [exclude_object](Exclude_Object.md) module before the adaptive
+bed mesh is requested, the default method will be used instead. If you are requesting an adaptive bed mesh and
```
voidtrance, line 433: "default method" or "default size"?

docs/Bed_Mesh.md
voidtrance: > I think a simple wording like
> 
> > Call `BED_MESH_CALIBRATE ADAPTIVE=1` either from your `[START_PRINT](Slicers.md#klipper-gcode_macro)` macro or from your slicer's starting Gcode. It is recommended to place this command pretty much at the end of the start-up routine, this means at least after heating up the bed.
> 
> would be clearer as these should be the two typical places for a regular user.

There is no guarantee that I user will be using a `START_PRINT`/`PRINT_START` macro. They could just be putting all of that in the slicer custom gcode.
It's better to leave it as "... add `BED_MESH_CALIBRATE ADAPTIVE=1` to your start GCode".

docs/Bed_Mesh.md
kyleisah: > What does "waste less mesh information" mean? What information is being wasted if adaptive is not being used?
Well, if you're doing a regular full mesh for a small cube on a big bed, you're really only using 1 probed point and some interpolation rather than all of the probed points fitting to the object being printed.

> Also, a nit: can you please conform to the document's line length?

I don't know what I'm doing here, I'm just trying to help out.

docs/Bed_Mesh.md
Sineos: Sorry, but I do not understand this comment:
grafik

voidtrance: > Sorry, but I do not understand this comment: grafik

Oops, sorry! I didn't read the text carefully enough.

KevinOConnor:
Thanks for working on this.  What is the current status of this PR?  Is there a consensus on a documentation change?

-Kevin

mdshw5:
I'd just like to bump this PR since I've gone though the adaptive bed mesh configuration with Klipper and moonraker, as well as helped a few others, and the only reliable place that explains the necessary configuration in detail is on the Klipper Discourse instance (https://klipper.discourse.group/t/native-adaptative-mesh-support/13541/2). I understand that this PR might be difficult since it involves changes to but the Klipper and moonraker documentation, but clarifying the setup process in some way is still necessary.

Sineos:
This should indeed be finalized. In my view, it seems quite comprehensive and at least better than what we have today.

dewi-ny-je:
@kyleisah please review the open point of the review by adapting the text and clicking "resolve" or by leaving a comment.

This change is good but it needs your input. It's almost there!

PR #6479, opened 2024-01-30
controller_fan: add ability to watch temperature_sensors too

slaesh:
I searched a way to have a fan controller that just watches one or more temperature sensors.. like host-temperature. since the enclosure fans of my voron are pretty loud.. and I noticed that most of the stuff is not getting hot at all.. so.. having this fan-controller work without the heater/extruder.. and just watch actual temperatures would be awesome! for my ears.. for the fans.. and hopefully for other guys out there ;)

I found this pretty old [PR](https://github.com/Klipper3d/klipper/pull/4341) from @mnigbur, which did not make it back then.. 

just pinging some of the guys who where active in the past in this thread.. @Ramalama2, @theopensourcerer, @meteyou, @edddeduck, @KevinOConnor

question: any chance to test it without running an actual machine? (in general, not for this one =))

To-Do:
- [x] adjust linting (its not linting, its a hand-written code checker :D)
- [x] test on machine

Signed-off-by: Sascha Ehlers <sascha@ehle.rs>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2791,6 +2791,19 @@ watched component.
 #   is provided here, then the fan will be enabled when any of the given
 #   heaters/steppers are enabled. The default heater is "extruder", the
 #   default stepper is all of them.
+#temperature_sensors:
+#   A list of "temperature_sensor:target_temperature" pairs.
+#   When any of the defined temperature_sensors reaches it threshold, the
+#   fan will be enabled.
+#   For Example, "pi: 60" would enable the fan when the temperature_sensor
+#   "pi" reaches 60 degrees.
+#   The given exmaple would look like this:
+#   temperature_sensors:
+#     pi:60
+#     aanotherone:70
+#temperature_sensors_only:
+#   If set to True, this controller_fan will ONLY watch the given temperature_sensors
+#   and ignore any given heater/extruder. The default is False.
 ```
 
 ### [temperature_fan]
--- a/klippy/extras/controller_fan.py
+++ b/klippy/extras/controller_fan.py
@@ -48,18 +48,23 @@ class ControllerFan:
         self.idle_timeout = config.getint("idle_timeout", default=30, minval=0)
         self.heater_names = config.getlist("heater", ("extruder",))
         self.last_on = self.idle_timeout
-        self.last_speed = 0.0
+        self.last_speed = -1
 
         # get configured temperature_sensors as
         # ((str, str), (str, str)) => (("pi", "55"), ("spider", "77"))
         temperature_sensors_config = config.getlists(
-            "temperature_sensors", seps=(":", "\n"), count=2, default=[]
+            "temperature_sensors",
+            seps=(":", "\n"),
+            count=2,
+            default=[],
         )
+
         # parse given thresholds to floats
         self.temperature_sensors_config = [
             (
                 sensor[0],
                 float(sensor[1]),
+                # 12.4,
             )
             for sensor in temperature_sensors_config
         ]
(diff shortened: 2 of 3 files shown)


Ramalama2:
Hey @slaesh 
Im sorry, but i have to pass, im not active anymore in the 3d printing world xD
Gaved my Voron away, using only Bambu atm if i need to print something rarely at all.
Cheers :-)

JamesH1978:

Thanks
James

slaesh:
> 
> Thanks
> James

Hey James, I am happy to do that. But I am not sure exactly what to do 🙈 Where do I need to upload that "signed" file? :)

BR Sascha ✌️

slaesh:
do I need to add another commit with `--signoff` option?

slaesh:
oh I see, the link was not working on my github app this morning. so.. we add this sign-off thingy at the squashed commit or how do I add this kind of stuff for the already done commits? Oo

```Signed-off-by: Sascha Ehlers <sascha@ehle.rs>```

slaesh:
@JamesH1978 done. Just saw another PR which had it in the description text. Hope this does the job ;)

pedrolamas:
Just a thought, but would the [combined temperature](https://www.klipper3d.org/Config_Reference.html#combined-temperature-sensor) sensor and a regular `temperature_fan` help to achieve a similar result?

TheFuzzyGiggler:
Is this not what's already covered with temperature_fan?

You can set a temperature_fan to check an mcu temp.

Example:

[temperature_fan SB2040_Fan]
pin: sb2040:gpio15
sensor_mcu: sb2040
sensor_type: temperature_mcu
cycle_time: 0.010
hardware_pwm: False
min_temp: 0
max_temp: 100
control = pid
pid_kp = 59.385
pid_ki = 0.153
pid_kd = 5758.870
min_speed: 0.1
target_temp: 45

Ignore my PID settings, they're horrible. I was messing around.

gabevenberg:
> Is this not what's already covered with temperature_fan?
> 
> You can set a temperature_fan to check an mcu temp.


Seems to me the reason is that you cant have a single fan be both a controller_fan and a temperature_fan.

dmit2k:
> I searched a way to have a fan controller that just watches one or more temperature sensors.. like host-temperature. since the enclosure fans of my voron are pretty loud.. and I noticed that most of the stuff is not getting hot at all.. so.. having this fan-controller work without the heater/extruder.. and just watch actual temperatures would be awesome! for my ears.. for the fans.. and hopefully for other guys out there ;)

Hi @slaesh ! Love your Zigbee sticks BTW )
But I've found one problem with your provided solution. I'm trying to resolve the same case: my printer has single controller fan block for cooling the host CPU, mainboard MCU and stepper drivers all at the same time. When the steppers and heaters are off, I need to cool down just the host CPU (which reaches 60-70C even if idle). For cooling the CPU I need only 20-30% of fan speed. But the [controller_fan] either can be Full Off or Full On. 

So your solution does not resolve the primary intention to reduce the fan noice: either it will be all the time Full On because of CPU temp, or Full On because of stepper / heater activity.

I would suggest much simpler solution as a compromise: allow the "min_speed" or "default_speed" option for [controller_fan] so it will run at a predefined minimal speed when not triggered On by steppers / heaters, just to cool down the controllers and CPU.

Otherwise we have to extend the [controller_fan] to a very complex thing with PID controls which can be difficult as we can't get the TMC2209 drivers actual temperature for example.

dmit2k:
> Seems to me the reason is that you cant have a single fan be both a controller_fan and a temperature_fan.

Actually we can, with the [duplicate_pin_override] Klipper option. 

But the problem is - which of them will take precedence over this single fan at specific moment? :) It's like placing an iron into the fridge guessing "who will win" :)

I have tried this workaround but can't imagine how to check the actual fan speed...

nefelim4ag:
https://github.com/Klipper3d/klipper/pull/6695
Template support is merged and should cover that functionality.

Thanks.

PR #6480, opened 2024-02-01
Bed screws: fix horizontal move speed. Bed mesh: add lift speed param

gatools:
During creation and testing of my custom 3d printer (core xy but different implementation: xy/2) I found that toolhead moves very quickly during calibration and was no ability configure this speed, and bed lift was also very quick and with small "run_current" bed was stuck.
Please review my changes to bed screws and and bed mesh. I tested my changes - works as expected.
My configuration file: [printer.cfg](https://gist.github.com/gatools/8eda6220f9c61f30d0a6d033146ac967)


JamesH1978:

Thanks
James

gatools:
Hello @JamesH1978 . Thank you. Please check my updated commits.

PR #6484, opened 2024-02-06
WS2801: Support for WS2801 LEDs using SW SPI and HW SPI

akrto:
Since none of the existing LED drivers worked properly with WS2801 LED modules this driver was added.
Tested on Arduino Uno
Will not pass all regression tests due to dict files being outdated on gh.


KevinOConnor:
Thanks.  I'm not very familiar with this LED type.  How does it differ from the existing "dotstar" module?

-Kevin

akrto:
> Thanks. I'm not very familiar with this LED type. How does it differ from the existing "dotstar" module?
> 
> -Kevin

In short, the protocols are different :-)

The Dotstars (APA102) has framing built into the protocol, meaning that it is not affected by pauses in the data that would be caused by the entire data not fitting into one message from Host to MCU-board running the LEDs.
There is about 40 bytes of data in each message right?
APA102 also has global dimming, which is something that the WS2801 doesn´t.

The WS2801 relies on pauses between messages for framing, meaning that if not all data could fit into one message from host to MCU it would start all over with the data in the next message.
Because of that, I had to implement local LED-registry on the MCU-board, a´ la Neopixel style.

Thank you for all the hard work you guys put into this project, you guys rock!

Framing/protocol for Dotstars/APA102:
image

Framing/protocols for WS2801:
image

bikebastler:
Very exciting topic!
Do I understand correctly that only software SPI is used here?
What happens if I use the HW SPI pins of my MCU?

akrto:
Hi @bikebastler!

I have not touched the subject for a long while and can not really answer your question at this time.

This was kind of one of those times where you spend a lot of time trying to get something that you happen to have laying around to work for the project you are currently working on, when the more practical solution would have been to just buy some _MODERN_ LEDs and use commonly available code to drive them.

At this point I do not really feel the need to pursue merging this code, since the LED drivers are of an outdated kind and not really available anymore except as NOS or similar.

So, feel free to close it, unless anyone else wants to contribute.

akrto:
Haha 😆 

I can´t really recall if I ever got this working well with HW SPI but I have a vague memory of testing out a RPi Pico but that I didn´t nail down the timing which led to flickering and all sorts of issues.

I think that it would be totally doable, if you like to spend some more time on it :)

PR #6487, opened 2024-02-08
scripts: Added shaper tuning parameters to calibrate_shaper script

dmbutyugin:
The added parameters include square_corner_velocity, shaper frequencies to optimize, input shapers to test, input shaper damping ratio and damping ratios to test. All these options can be useful for fine-tuning the input shapers when the default suggestions generated by the tuning script are not optimal.

Also the `SHAPER_CALIBRATE` command was modified to pass some of these parameters to the shaper tuning routine. Specifically, square corner velocity and the maximum tested frequency are used to adjust shaper tuning and maximum acceleration recommendations.

I think the possibility to specify the `square_corner_velocity` is most useful. Some people change it against the recommendations, but may use the same maximum acceleration recommendations from the script without realizing that they should be adjusted. This change allows them to get better estimates for maximum accelerations and potentially make better and more informed decisions about raising `square_corner_velocity`  parameter. Then the possibility to tune only a subset of shapers and specific frequencies may be useful for advanced users who wanted to do fine-tuning of the shapers, but lacked the tools to do so.

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -662,6 +662,19 @@ The same notice applies to the input shaper
 `max_accel` value after the auto-calibration, and the suggested acceleration
 limits will not be applied automatically.
 
+Keep in mind that the maximum acceleration without too much smoothing depends
+on the `square_corner_velocity`. The general recommendation is not to change
+it from its default value 5.0, and this is the value used by default by the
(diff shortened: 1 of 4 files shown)


KevinOConnor:
Thanks.  Let me know when you'd like me to merge this.

Cheers,
-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6489, opened 2024-02-10
adxl345: Add I2C support to the ADXL345 accelerometer implementation

enduity:
Hello,

The purpose of this pull request is to allow others with the same issues as me to use their accelerometers without hardware modding. Namely the issue I had was that my board from Seeed Studio, the [Grove - ADXL345 - 3-Axis Digital Accelerometer(±16g)](https://www.seeedstudio.com/Grove-3-Axis-Digital-Accelerometer-16g.html), was hardwired to use I2C with a trace below the ADXL345 chip itself (between CS and Vin). As I do not own a hot air station, I wanted to avoid desoldering this chip as much as possible. Turns out what I wanted was possible, as the ADXL345 chip supports 400 kHz I2C communication, which is already supported by Klipper in the MPU series.

I changed the ADXL345 Klippy Python code to detect I2C or SPI bus usage from the config, modifying the communication behaviour accordingly. I also implemented a new sensor_adxl345_i2c.c driver, because the code changes were much larger in this case and it is easier to configure the Makefile this way (to include only if I2C is supported on a given MCU). 

I have tested this on my own printer and both I2C and SPI (on another board) work properly. I do not think there is a big risk in breaking any existing functionality, because the added Python code uses basic if statements, where the SPI cases include previously existing code.

I admit the documentation of this addition could be vastly improved, but I think it is currently usable for someone already tinkering with accelerometers using a manual setup (vs a KUSBA or something). I sadly currently lack the time to improve it. Still, I felt it's important to get this change out there so nobody ends up buying another board unnecessarily or maybe breaking one trying to get SPI to work (or maybe even wasting their time trying to do this).

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1631,8 +1631,15 @@ specify an explicit name (eg, [adxl345 my_chip_name]).
 
 ```
 [adxl345]
+#i2c_mcu:
+#i2c_bus:
+#i2c_software_scl_pin:
+#i2c_software_sda_pin:
+#i2c_speed: 400000
+#   See the "common I2C settings" section for a description of the
+#   above parameters. The default "i2c_speed" is 400000.
 cs_pin:
-#   The SPI enable pin for the sensor. This parameter must be provided.
+#   The SPI enable pin for the sensor. This parameter must be provided if using SPI communication.
 #spi_speed: 5000000
 #   The SPI speed (in hz) to use when communicating with the chip.
 #   The default is 5000000.
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -1,20 +1,15 @@
 # Measuring Resonances
 
-Klipper has built-in support for the ADXL345, MPU-9250 and LIS2DW compatible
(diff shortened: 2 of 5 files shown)


KevinOConnor:
Thanks.  I guess the main question is - does the adxl345 actually work well in i2c mode?  When that chip is in i2c mode it is severely limited in the sample rate it can achieve (without losing samples).

So, I'd be curious to understand the testing results you obtained.  I think @dmbutyugin may also be interested.

-Kevin

enduity:
Thanks for replying.

Would running a different ADXL345 on the same bed/extruder be sufficient? I can probably manage to run them simultaneously if necessary. I already own the other one, which has SPI wiring available.

Sineos:
I'd propose two measurements:
* `TEST_RESONANCES AXIS=X OUTPUT=raw_data`
* Same location
* One with SPI connection and one with I2C
* Compare the two raw measurements in terms of how many samples have been taken
* Generate the graphs and compare if there are noteworthy / unexplained differences
* `MEASURE_AXES_NOISE` with both and compare if there are noteworthy differences

PR #6490, opened 2024-02-10
BDsensor: adding support for new probe sensor BDsensor

markniu:
adding support BDsensor for Fast auto bed leveling leveling.
About BDsensor: 
It can measure the bed distance at any point in real time without moving the z axis up and down.
now the fast auto bed leveling is stable for pull, there will be more new features like real time adjust and strain gauge in the future.https://github.com/markniu/Bed_Distance_sensor
Config: https://github.com/markniu/Bed_Distance_sensor/wiki/Installing-for-Klipper

### DIFF
--- a/src/BD_sensor.c
+++ b/src/BD_sensor.c
@@ -2,9 +2,8 @@
 // Mark yue<niujl123@sina.com>
 // This file may be distributed under the terms of the GNU GPLv3 license.
 
-#include<string.h>
+#include <string.h>
 #include <stdlib.h>
-#include "autoconf.h"
 #include "board/gpio.h"
 #include "board/irq.h"
 #include "board/misc.h"
@@ -17,7 +16,6 @@
 #include "board/misc.h" // timer_is_before
 #include "command.h" // DECL_COMMAND
 #include "sched.h" // struct timer
-#include "stepper.h" // stepper_event
 #include "trsync.h" // trsync_add_signal
 
 
@@ -26,23 +24,6 @@
 #define BD_setLow(x)  gpio_out_write(x,0)
 #define BD_setHigh(x) gpio_out_write(x,1)
 
-
-uint32_t delay_m = 20, homing_pose = 0;
-int sda_pin = -1, scl_pin = -1, z_ofset = 0;
-uint16_t BD_Data;
-//extern uint32_t timer_period_time;
-uint16_t BD_read_flag=1018,BD_read_lock=0;
-int switch_mode = 0; //1:in switch mode
-struct gpio_out sda_gpio, scl_gpio;
-struct gpio_in sda_gpio_in;
-uint8_t oid_g,etrsync_oid,endstop_reason=0;
-uint8_t z_oid[4];
-uint32_t endtime_debug=0;
-uint32_t timer_period_endstop=100;
-
-///////////BDsensor as endstop
-struct timer time_bd;
-#include "autoconf.h"
 struct endstop {
     struct timer time;
     uint32_t rest_time, sample_time, nextwake,pin_num;
@@ -52,71 +33,16 @@ struct endstop {
 };
 
 enum { ESF_PIN_HIGH=1<<0, ESF_HOMING=1<<1 };
+
+uint32_t delay_m = 20, homing_pose = 0;
+int sda_pin = -1, scl_pin = -1, z_ofset = 0;
+uint16_t BD_Data;
+uint16_t BD_read_flag=1018,BD_read_lock=0;
+int switch_mode = 0; //1:in switch mode
+struct gpio_out sda_gpio, scl_gpio;
+struct gpio_in sda_gpio_in;
 static uint_fast8_t endstop_oversample_event(struct timer *t);
 static struct endstop e ;
-///////////////
-
-int z_index=0;
-
-enum { POSITION_BIAS=0x40000000 };
-
-struct stepper_move {
-    struct move_node node;
-    uint32_t interval;
-    int16_t add;
-    uint16_t count;
-    uint8_t flags;
-};
-
-struct stepper {
-    struct timer time;
-    uint32_t interval;
-    int16_t add;
-    uint32_t count;
-    uint32_t next_step_time, step_pulse_ticks;
-    struct gpio_out step_pin, dir_pin;
-    uint32_t position;
-    struct move_queue_head mq;
-    struct trsync_signal stop_signal;
-    // gcc (pre v6) does better optimization when uint8_t are bitfields
-    uint8_t flags : 8;
-};
-
-struct step_adjust{
-    uint32_t cur_z;
-    int adj_z_range;
-    int invert_dir;
-    int steps_per_mm;
-    int step_time;
-    int zoid;//oid for all the z stepper
-};
-
-#define NUM_Z_MOTOR  6
-struct step_adjust step_adj[NUM_Z_MOTOR];//x,y,z
-
-struct _step_probe{
-    int min_x;
-    int max_x;
-    int points;
-    int steps_at_zero;
-    int steps_per_mm;
-    int xoid;//oid for x stepper
-    int x_count;
-    int x_dir;
-    int y_dir;
-    int kinematics;//0:cartesian,1:corexy,2:delta
-    ////for y
-    int min_y;
-    int max_y;
-    int y_steps_at_zero;
-    int y_steps_per_mm;
-    int y_oid;//oid for y stepper
-    ////
-    int x_data[64];
-};
-
-struct _step_probe stepx_probe;
-
 
 void BD_i2c_write(unsigned int addr);
 uint16_t BD_i2c_read(void);
@@ -124,7 +50,6 @@ uint16_t BD_i2c_read(void);
 int BD_i2c_init(uint32_t _sda,uint32_t _scl,
     uint32_t delays,uint32_t h_pose,int z_adjust)
 {
-    int i=0;
     sda_pin=_sda;
     scl_pin =_scl;
     homing_pose = h_pose;
@@ -138,13 +63,6 @@ int BD_i2c_init(uint32_t _sda,uint32_t _scl,
 
     gpio_out_write(sda_gpio, 1);
     gpio_out_write(scl_gpio, 1);
-    for (i=0;i<NUM_Z_MOTOR;i++){
-        step_adj[i].cur_z=0;
-        step_adj[i].zoid=0;
-        step_adj[i].adj_z_range=0;
-    }
-    stepx_probe.xoid=0;
-    stepx_probe.y_oid=0;
     BD_i2c_write(1022); //reset BDsensor
     return 1;
 }
@@ -154,7 +72,6 @@ uint32_t nsecs_to_ticks_bd(uint32_t ns)
     return ns * (CONFIG_CLOCK_FREQ / 1000000);
 }
 
-
 void ndelay_bd_c(uint32_t nsecs)
 {
     if (CONFIG_MACH_AVR)
@@ -163,6 +80,7 @@ void ndelay_bd_c(uint32_t nsecs)
     while (timer_is_before(timer_read_time(), end))
         irq_poll();
 }
+
 void ndelay_bd(uint32_t nsecs)
 {
     int i=1;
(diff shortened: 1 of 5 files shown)


Sineos:
Thanks a lot for your contribution. Please also provide the signed off as per point 3 in "What to expect in a review" in [master/docs/CONTRIBUTING.md](https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md)

Also, it would need to come with an update of the documentation in the relevant chapters, e.g. "bed level" and "g-codes"

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+#define BYTE_CHECK_OK     0x01
+#define BYTE_CHECK_ERR    0x00
+#define BD_setLow(x)  gpio_out_write(x,0)
+#define BD_setHigh(x) gpio_out_write(x,1)
+
+
+uint32_t delay_m = 20, homing_pose = 0;
+int sda_pin = -1, scl_pin = -1, z_ofset = 0;
+uint16_t BD_Data;
+//extern uint32_t timer_period_time;
```
slaesh, line 33: can be removed? =)

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+struct gpio_out sda_gpio, scl_gpio;
+struct gpio_in sda_gpio_in;
+uint8_t oid_g,etrsync_oid,endstop_reason=0;
+uint8_t z_oid[4];
+uint32_t endtime_debug=0;
+uint32_t timer_period_endstop=100;
+
+///////////BDsensor as endstop
+struct timer time_bd;
+#include "autoconf.h"
```
slaesh, line 45: maybe move it up to all other imports?

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+
+void ndelay_bd_c(uint32_t nsecs)
+{
+    if (CONFIG_MACH_AVR)
+        return;
+    uint32_t end = timer_read_time() + nsecs_to_ticks_bd(nsecs);
+    while (timer_is_before(timer_read_time(), end))
+        irq_poll();
+}
+void ndelay_bd(uint32_t nsecs)
```
slaesh, line 166:
```suggestion
}

void ndelay_bd(uint32_t nsecs)
```

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+{
+    int i=1;
+    while(i--)
+        ndelay_bd_c(nsecs);
+}
+
+unsigned short BD_Add_OddEven(unsigned short byte)
```
slaesh, line 173: a little bit confusing to name a "short" `byte` =)) was wondering how a byte can have more than 8 (10) bits looking at the loop :)

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+    scl_gpio=gpio_out_setup(scl_pin, 1);
+    BD_setHigh(scl_gpio);
+    BD_setHigh(sda_gpio);
+    ndelay_bd(delay_m);
+    BD_setLow(sda_gpio);
+    ndelay_bd(delay_m);
+    BD_setLow(scl_gpio);
+    ndelay_bd(delay_m);
+}
+void  BD_i2c_stop(void)
```
slaesh, line 233:
```suggestion
}

void  BD_i2c_stop(void)
```

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+    ndelay_bd(delay_m);
+
+    BD_setHigh(sda_gpio);
+    ndelay_bd(delay_m);
+}
+
+uint16_t BD_i2c_read(void)
+{
+    uint16_t b = 1024;
+   // if(BD_read_flag==1014)
```
slaesh, line 250: can be removed? :)

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+            b = b - z_ofset;
+            if(b>1024)
+                b=0;
+        }
+
+    }
+    else
+        b=1024;
+
+#if 0
```
slaesh, line 286: can be removed? or maybe add a speaking comment why this is present :)

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+
+//for gcode command
+void
+command_I2C_BD_receive(uint32_t *args)
+{
+    uint8_t oid = args[0];
+    uint8_t data[8];
+    uint16_t BD_z;
+
+    //if(BD_read_flag==1018)
```
slaesh, line 369: can be removed?

src/BD_sensor.c
```diff
@@ -0,0 +1,541 @@
...
+
+DECL_COMMAND(command_I2C_BD_receive, "I2C_BD_receive oid=%c data=%*s");
+
+
+void
+command_I2C_BD_send(uint32_t *args)
+{
+    int addr=atoi((char *)args[2]);
+    BD_read_flag=addr;
+    if(addr==1015)
```
slaesh, line 420: maybe all these const's should be an enum? 1015, 1023, 1024, .. so many number all over the place :))

slaesh (commented):
cant review/test in a technical perspective.. just some code-quality/hygiene hints :)

I guess this project could need some proper linting rules ;)

KevinOConnor:
Thanks.  Is this sensor based on the ldc1612 chip?  There has been some work done on that chip at #6536 (and #6537).

My high-level feedback is that I think we need to integrate these efforts into a single "probe_eddy_current.py" type of module.  I think we want to avoid having each probe implement their own implementations of PROBE, PROBE_ACCURACY, homing, etc.

-Kevin

markniu:
hello Kevin, thanks! 
 this sensor is not based on ldc1612 chip, it is based on the TCA355G that is a traditional inductive proximity Switch sensor chip,  some other ADC chip and analog electrical component to change the eddy current, not like the ldc1612 which change the eddy frequency to detect the distance. 


> Thanks. Is this sensor based on the ldc1612 chip? There has been some work done on that chip at #6536 (and #6537).
> 
> My high-level feedback is that I think we need to integrate these efforts into a single "probe_eddy_current.py" type of module. I think we want to avoid having each probe implement their own implementations of PROBE, PROBE_ACCURACY, homing, etc.
> 
> -Kevin

markniu:
I agree, and it is no problem to remove the PROBE and PROBE_ACCURACY in the BDsensor.py and use the PROBE, PROBE_ACCURACY in the probe.py.

nefelim4ag:
If there is a will to make it acceptable, the amount of code in MCU should be reduced and look similar to `sensor_ldc1612.c`.
There should be no magic `i2c` command mappings, and no magic i2c reimplementation, no `atoi()` mapping, with no reason (like sending a number as a string, converting it to a string, and then converting an integer to string, to send a response).

There is already I2C infrastructure at Klippy and MCU level, it can be directly used and most of the code can be easily removed.
There are already several examples of I2C device communication in extras, from simple temperature sensors to GPIO multiplexors like `SX1509`, there is even a direct analog of DBSencor, as LDC1612 mentioned above.

Thanks.

markniu:
1)   I think I forget to add the [data protocol](https://github.com/markniu/Bed_Distance_sensor/wiki/Data-Protocol) here and this is not a standard I2C. I also recommend all of us don't to write the I2C code if the system already has that. 
    The reason here is the efficiency of data communication effect the the accuracy, as we know the X/Y axis is moving very fast while doing mesh or real time level, any delay on the communication will effect the accuracy of the X/Y position that we want to probe. in the same way, that delay can also effect the z axis while homing.
   So I change the I2C protocol in this bdsensor, for example, I remove the address transfer, reduce the cycle of 2 bytes(16bit) to 10 bit, that  can save a lot of transfer cycle and time. 
   
 2)  I had removed the `atio `by using `uint32_t `instead of `string `in my [github](https://github.com/markniu/Bed_Distance_sensor/tree/new/klipper) but not sync here, I will sync here later.  

BTW, I am printing with real_time_level all the time and recommend to use the that instead of mesh_bed_level. 

> If there is a will to make it acceptable, the amount of code in MCU should be reduced and look similar to `sensor_ldc1612.c`. There should be no magic `i2c` command mappings, and no magic i2c reimplementation, no `atoi()` mapping, with no reason (like sending a number as a string, converting it to a string, and then converting an integer to string, to send a response).
> 
> There is already I2C infrastructure at Klippy and MCU level, it can be directly used and most of the code can be easily removed. There are already several examples of I2C device communication in extras, from simple temperature sensors to GPIO multiplexors like `SX1509`, there is even a direct analog of DBSencor, as LDC1612 mentioned above.
> 
> Thanks.

nefelim4ag:
> The reason here is the efficiency of data communication effect the the accuracy

According to your protocol, there is a 10 bits of data.
So, you spend 12 bits, instead of 30 bits (3 bytes, address, MSB, LSB).
Which can be translated to 120us, and 300us accordingly.
With 400kHz I2C it is 4 times faster.

How fast Z can be homed?
I have 4 mm rotational distance, and can home with speed of 10 mm/s.
With 8 mm rotational distance, speed must be x2,  and the calculation would be exact.
`10 / 4 * 200 = 500 steps/s = 2 ms per step`
With microstepping 4 we get x4 rate, so 2000 steps/s ~ 500us, 8 = 250us, 16 ~ 125us.
So, the possible overshoot is roughly 500us for normal i2c and 125us for your implementation.
`10 mm/s * 0.000_5 = 0.005 mm`
`10 mm/s * 0.000_125 = 0.00125mm`
For 20 mm/s, there will be an x2 overshoot. 0.01 mm and 0.0025 mm.

According to your documentation, you return data approximately for 0.01 mm difference per bit, with repeatability around +-0.005 mm.
So, from the high-level view, there will be no visible difference, or it can be completely negotiated with either a higher i2c speed or a slower second home move (below).

Again, I recommend taking a look at `ldc1612` and mentioned above `probe_eddy_current`.

Z probe triggered when samples are above threshold.
So, technically at any specific moment, you will be able to check the current height above the bed.
There is a second "slower" home move for higher precision, with half of the speed.
Which can be configured slower as it needs to trigger the sensor between steps.
So, there should be no problem with Z homing.

XY Scan is pretty simple, klipper moves the toolhead over probe points and pauses the toolhead in each point for `100ms`, after the scan is done, data from the sensor with timestamps compared to the toolhead positions and timestamps. 

So, for bed scans, there is plenty of time, to do several samples over a point, and i2c speed matters even less.

> BTW, I am printing with real_time_level all the time and recommend to use the that instead of mesh_bed_level

If there is a custom implementation of a bed scan, it is better to send it as a separate pull request.

Hope it helps.

markniu:
you are right for this calculation, but actually, the speed maybe more slow than that in some case. 
1, need to consider both side of the i2c. 
Master i2c(host mcu): for most mainboard there is no free hardware I2C port, so need to use software i2c with normal gpio that need to consider the other related code consume not only the speed of gpio, if the mcu is AVR 8bit, the speed of gpio and software even slower.
Slave i2c(mcu in sensor):the first version of mcu in sensor is 8bit, and there is even no hareware i2c slave inside. also need to consider the data sample time inside.

2, The other import reason that I choose this custom i2c is the first idea to design this sensor is for real_time_level. there is no bed scan before print, but adjust the z axis with the real distance data from sensor,  that requires more about the real time data and the efficiency of the system, even must do that in mcu side not the python. 

now seems most of the mainboard is 32bit even with very high speed mcu, the bdsensor also had been updated to stm32, but if the software is more efficiency or do more things in sensor side then the system will be more stable.  of course this lose some features like not easy to use for users, code maintenance ..  

 I agree with what you said, very professional and this will help me continue to improve this sensor.


> > The reason here is the efficiency of data communication effect the the accuracy
> 
> According to your protocol, there is a 10 bits of data. So, you spend 12 bits, instead of 30 bits (3 bytes, address, MSB, LSB). Which can be translated to 120us, and 300us accordingly at 100kHz. With 400kHz I2C it is 4 times faster.
> 
> How fast Z can be homed? I have 4 mm rotational distance screw, and can home with a speed of 10 mm/s. With 8 mm rotational distance screw, the speed must be x2, and the per step calculation would be exact. `10 / 4 * 200 = 500 steps/s = 2 ms per step` With microstepping 4 we get x4 rate, so 2000 steps/s ~ 500us, 8 = 250us, 16 ~ 125us. So, the possible overshoot is roughly 500us for normal i2c and 125us for your implementation. `10 mm/s * 0.000_5 = 0.005 mm` `10 mm/s * 0.000_125 = 0.00125mm` For 20 mm/s, there will be an x2 overshoot. 0.01 mm and 0.0025 mm. It can be recalculated to get a microstepping precision, but it sounds unnecessary. So, I say sorry for my rough calculations here.

> If there is a custom implementation of a bed scan, it is better to send it as a separate pull request.
> 
> Hope it helps.

PR #6492, opened 2024-02-15
Draft: custom_component: allow to load python components natively

ayufan:
It become quite common for Klipper users to create custom components and create a dance to symlink them into `extras` folder. This PR provides a first class support for `custom_components` that can be loaded as part of `printer.cfg`:

1. First: Every custom component to be used to be declared as `custom_component`, component name to load, and path where the component is loaded.
2. The component will be lazily loaded on a first use.
3. The custom component will track how many times the component was used, and expose git information if component is repository.
4. It is responsibility of the custom component author to ensure compatibility with Klipper.
5. The `custom_component` load module is only compatible with Python 3.5+, for older versions there's no support for `custom_component`.

#### `printer.cfg`

```yaml
[custom_component hello_world]
path: ~/my_klipper/hello_world.py

[hello_world]
```

#### `hello_world.py`

```python
class HelloWorld:
    def __init__(self, config):

        # Register commands
        self.gcode = self.printer.lookup_object('gcode')
        self.gcode.register_command("HELLO_WORLD", self.cmd_HELLO_WORLD)

    def cmd_HELLO_WORLD(self, gcmd):
        gcmd.respond_info("Hello World")

def load_config(config):
    return HelloWorld(config)
```

#### `GET /printer/objects/query?custom_component%20hello_world`

```json
{
    "result": {
        "eventtime": 2422.219077722,
        "status": {
            "custom_component hello_world": {
                "path": "/home/ayufan/my_klipper/hello_world.py",
                "git_info": {
                    "version": "fc11477a-dirty",
                    "file_status": [],
                    "branch": "master",
                    "remote": "origin",
                    "url": "git@github.com:ayufan/printer-klipper-data.git"
                },
                "used": 1
            }
        }
    }
}
```

## References

1. https://github.com/moggieuk/ERCF-Software-V3/blob/master/install.sh#L101
2. https://github.com/beacon3d/beacon_klipper/blob/master/install.sh#L22
3. https://esphome.io/components/external_components

### DIFF
--- a/klippy/extras/custom_component.py
+++ b/klippy/extras/custom_component.py
@@ -1,0 +1,35 @@
+# Custom Component Module
+#
+# Copyright (C) 2024  Kamil Trzciński <ayufan@ayufan.eu>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging, util, os.path, importlib.util
+
+class CustomComponent:
+    def __init__(self, config):
+        self.name = config.get_name().split()[-1]
+        self.path = config.get('path')
+        self.used = 0
+        self.module = None
+        self.git_info = None
+
+    def load_module(self):
+        self.used += 1
+        if not self.module:
+          full_path = os.path.expanduser(self.path)
+          spec = importlib.util.spec_from_file_location(self.name, full_path)
+          module = importlib.util.module_from_spec(spec)
+          spec.loader.exec_module(module)
+          self.git_info = util.get_git_version(full_path)
+          self.module = module
+        return self.module
+
+    def get_status(self, eventtime):
+        return {
+          'path': self.path,
+          'git_info': self.git_info,
+          'used': self.used
+        }
+
+def load_config_prefix(config):
+    return CustomComponent(config)
--- a/klippy/klippy.py
+++ b/klippy/klippy.py
@@ -110,15 +110,21 @@ class Printer:
             return self.objects[section]
         module_parts = section.split()
         module_name = module_parts[0]
-        py_name = os.path.join(os.path.dirname(__file__),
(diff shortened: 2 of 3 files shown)


nelsongraca:
@ayufan I see this has been untouched for some time now, are you planning to complete this? I think this would be a good addition to klipper

ayufan:
I did not get a feedback. I would love to get this merged, if there is
agreement on this. So, if get preliminary approval will finish it.

nelsongraca:
I would get it ready to be useable and merged into master, without being marked as a draft, ensure the checks pass, self review etc, I don't think anyone will look into something marked as a draft

nelsongraca:
> I started a discussion on something similar at https://klipper.discourse.group/t/possible-klipper-plugins-instead-of-macros/9819 .

Thank you for making me aware of if will have a look and drop my feedback, I'm interested in helping in a solution

ayufan:
> It would be important for me that a "plugin" system have a well understood and stable API. Otherwise, I fear "plugins" could inadvertently destabilize other parts of Klipper, and minor changes in Klipper could inadvertently break 3rd-party code. The Klipper "extras" system was never intended as a "plugin" system, and I don't think it meets those requirements.

Yeah, this is always a problem. However, there's not a good solution. There are ways to maintain out of system information about the compatibility with Klipper, and also making Klipper more aware of what is happening.

The problem today is that the are vendors, and users that simply already have to mess with `extras`, which is likely even worse.

@KevinOConnor As a good example you can see this working pretty well in `Home Assistant` and `ESPHome` ecosystem, where there's a big repository of built-in components, but equally big of external components. Quality of external varies, but I doubt people would drop those things at random.

PR #6498, opened 2024-02-17
shaper_calibrate: Fixed crashes in SHAPER_CALIBRATE and TEST_RESONANCES

dmbutyugin:
Fixed crashes due to wrong parameter passed to the shaper selection function and when the custom FREQ_END is specified.


KevinOConnor:
Thanks.

-Kevin

PR #6499, opened 2024-02-19
feat(klippy, toolhead): add S<seconds> param to G4 command

slaesh:
this adds the ability to pass an `S<seconds>` parameter to the G4 dwell command.
its even possible to combine the existing `P<milliseconds>` with the newly added `S` param.

took me a while to realize that klippers G4 interpreter only used the `P` param.. :))

To-Do:

- [ ] ask/clarify: logging format, do we want to prefix it. or dont log at all??
- [ ] ask/clarify: if not dwelling at all for the None case is okay?
- [x] adjust docs

Signed-off-by: Sascha Ehlers [sascha@ehle.rs](mailto:sascha@ehle.rs)

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -7,7 +7,7 @@ commands that one may enter into the OctoPrint terminal tab.
 
 Klipper supports the following standard G-Code commands:
 - Move (G0 or G1): `G1 [X<pos>] [Y<pos>] [Z<pos>] [E<pos>] [F<speed>]`
-- Dwell: `G4 P<milliseconds>`
+- Dwell: `G4 P<milliseconds>` or `G4 S<seconds>` or `G4 S<seconds> P<milliseconds>`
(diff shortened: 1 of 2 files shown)


pedrolamas:
My understanding is that Klipper is not trying to replicate the full Marlin gcode support, hence why this is not implemented.

For now, you can easily fix this issue by overriding G4 with a custom macro:

```ini
[gcode_macro G4]
rename_existing: G4.1
gcode:
  G4.1 P{ params.P | default((params.S | default(0) | int) * 1000) }
```

The above will allow you to use either `P` or `S` params.

slaesh:
thanks for the hint. before doing that, I will keep adding some zeros here and there ;)

otherwise shared config's wont work anymore on another klipper instance :)

in my feeling this doesnt hurt, but sure it does not add so much value either. since no one requested this earlier, at least cant find any related issue.. I will close this one then =)

PR #6501, opened 2024-02-19
docs: Bed_Mesh.md Correction

nuburt:
mesh_max of (240,198) with 24,5 offset is (216,193) not (206,193)


JamesH1978:

Thanks
James

PR #6504, opened 2024-02-22
remove dead code

kdomanski:
- removes unused param 'eventtime' from heater.check_busy(  )
- removes a few unused assignments that are likely remnants of removed code

### DIFF
--- a/klippy/configfile.py
+++ b/klippy/configfile.py
@@ -189,7 +189,6 @@ class PrinterConfig:
     comment_r = re.compile('[#;].*$')
     value_r = re.compile('[^A-Za-z0-9_].*$')
     def _strip_duplicates(self, data, config):
-        fileconfig = config.fileconfig
         # Comment out fields in 'data' that are defined in 'config'
         lines = data.split('\n')
         section = None
--- a/klippy/extras/dotstar.py
+++ b/klippy/extras/dotstar.py
@@ -10,7 +10,6 @@ BACKGROUND_PRIORITY_CLOCK = 0x7fffffff00000000
 class PrinterDotstar:
     def __init__(self, config):
         self.printer = printer = config.get_printer()
-        name = config.get_name().split()[1]
         # Configure a software spi bus
         ppins = printer.lookup_object('pins')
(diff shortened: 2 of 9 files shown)


dewi-ny-je:
I think you need to sign your contribution. https://www.klipper3d.org/CONTRIBUTING.html

kdomanski:
Yeah, I only signed the first commit and not the second. Thanks, nice catch!

kdomanski:
## Self-review checklist

### 1. Is the submission free of defects and is it ready to be widely deployed?

- Yes, the submission passes regression tests
- it has been tested manually as well
- it doesn't add any logging or debugging code

### 2. Does the submission provide a "high impact" benefit to real-world users performing real-world tasks?

No, however with just a few lines removed it removes a lot of confusion regarding the purpose of the dead code.
Personally I struggled to understand the code in question when debugging certain issues and the presence of dead code served as a distraction. I think getting rid of it will help contributors.

### 3. Is the copyright of the submission clear, non-gratuitous, and compatible?

Yes.

### 4. Does the submission follow guidelines specified in the Klipper documentation?

Yes.

### 5. Is the Klipper documentation updated to reflect new changes?

No documentation is necessary.

### 6. Are commits well formed, address a single topic per commit, and independent?

IMO, yes.

KevinOConnor:
Thanks.  A call to `toolhead.get_last_move_time()` has side effects - it flushes the lookahead queue and ensures that the "print time" advances.  So, I think this change could have a subtle impact on behavior.  Also, although no current `check_busy()` implementation uses `print_time` I don't see a harm in passing it into that function - it can serve as a reminder that the toolhead should be flushed prior to invoking it.

The other parts of the dead code removal look good to me.  Happy to commit those if separated from the changes that can alter behavior.

Cheers,
-Kevin

kdomanski:
Hey Kevin, thanks for the explanation. I plan to tend to this PR and #6503 in a week or two when I'm back from sick leave.

kdomanski:
I restored the calls to `toolhead.get_last_move_time()` and removed the commit which altered the signature of `check_busy()`.

PR #6505, opened 2024-02-22
install-ubuntu-22.04: Mark as executable

Newspaperman57:
This file wasn't marked as executable, and thus didn't autocomplete when pressing tab, causing minor confusion when installing on those systems

mode change 100644 => 100755 scripts/install-ubuntu-22.04.sh`


KevinOConnor:
Thanks.

-Kevin

PR #6506, opened 2024-02-23
lpc176x: don't write to 0x2FC.

jernejp21:
Leave 0x2FC - 0x2FF alone, since it stores values for code read protection. We don't want to lock flash permanently.

According to user manual, 0x2FC is reserved for CRP mechanism. It might happen that generated code is exactly the same as CRP1, CRP2 or CRP3. For that reason we should avoid writing to that area.

slika

### DIFF
--- a/src/lpc176x/Kconfig
+++ b/src/lpc176x/Kconfig
@@ -72,7 +72,7 @@ endchoice
 config FLASH_APPLICATION_ADDRESS
     hex
     default 0x4000 if LPC_FLASH_START_4000
-    default 0x0000
+    default 0x0300
 
 
 ######################################################################
--- a/src/lpc176x/Makefile
+++ b/src/lpc176x/Makefile
@@ -8,8 +8,8 @@ dirs-y += src/lpc176x src/generic lib/lpc176x/device
 CFLAGS += -mthumb -mcpu=cortex-m3 -Ilib/lpc176x/device -Ilib/cmsis-core
 
 CFLAGS_klipper.elf += --specs=nano.specs --specs=nosys.specs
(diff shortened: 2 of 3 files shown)


Arksine:
Interesting.  I think the chances that we would unintentionally write `CRP3` are extremely low, but I see the utility in making it impossible.    If its decided that we need to avoid 0x2FC, I think a simpler solution would be something like this:

```diff
diff --git a/src/generic/armcm_link.lds.S b/src/generic/armcm_link.lds.S
index 2f789f13..46126584 100644
--- a/src/generic/armcm_link.lds.S
+++ b/src/generic/armcm_link.lds.S
@@ -22,6 +22,9 @@ SECTIONS
         _text_vectortable_start = .;
         KEEP(*(.vector_table))
         _text_vectortable_end = .;
+#if CONFIG_LPC_FLASH_START_0000
+        . = 0x300;
+#endif
         *(.text .text.*)
         *(.rodata .rodata*)
     } > rom
```
I'm curious to see Kevin's feedback.  If we make this change to Klipper I will also make it to Katapult.

KevinOConnor:
Interesting, thanks.  I agree it is worthwhile to change the code to avoid this problem.

I agree with Arksine in that I'm unsure about changing FLASH_APPLICATION_ADDRESS and about introducing a new "vector" segment in the linker script.  I do think we should introduce a new `lpc176x_link.lds.S` file though (as I think we should try to avoid adding board specific code to the `src/generic/` directory).  How about copying `src/generic/armcm_link.lds.S` to a new `src/lpc176x/lpc176x_link.lds.S` and adding Arksine's suggestion there (along with a comment about needed to avoid CRP1).

Thanks again,
-Kevin

PR #6507, opened 2024-02-25
scripts: Add octopi installation on Debian Bookworm

farnasirim:
Recent octopi nightly releases are based on Debian Bookworm as shown on the nightly build pages https://unofficialpi.org/Distros/OctoPi/nightly/ with the most recent ones named *-octopi-bookworm-armhf-lite-*.

This copies the original octopi scripts, making python2->python3 tweaks to installation targets.

### DIFF
--- a/scripts/install-octopi-bookworm.sh
+++ b/scripts/install-octopi-bookworm.sh
@@ -1,0 +1,103 @@
+#!/bin/bash
+# This script installs Klipper on a Raspberry Pi machine running the
+# OctoPi distribution.
+
+PYTHONDIR="${HOME}/klippy-env"
+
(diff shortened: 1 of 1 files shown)


Sineos:
Thanks for your contribution.
Not sure how much sense it makes to run after these manual installation scripts nowadays.
Most people likely 
* Do no longer use Octoprint
* Use KIAUH to do the heavy lifting for Moonraker, Fluid, Mainsail etc.
* Potentially also use KIAUH to install Octoprint

farnasirim:
I see, thanks for the pointers it's nice to know about KIAUH! I had just started going through the klipper docs.

PR #6509, opened 2024-02-25

charminULTRA:
Current commands in the documentation, using the *, results in bad chart output when more than one .csv file exists in the tmp folder. This isn't obvious for people who may not know that the * is a wildcard character.

I was pulling out my hair trying to understand my results until I realized my multiple runs of data were all being used in the same graph! I think this simple pitfall should be fixed.


freakydude:
I used this as a feature - so you can measure multiple times (e.g. with a full and empty spool on top) and the results will be merged. 

If that was the intention, I would comment the script line with - "be aware, all matching csv will be merged" or something like that.

Kind regards

JamesH1978:

Thanks
James

dewi-ny-je:
@charminULTRA can you apply the corrections I suggested? then we can hopefully have the changes approved

charminULTRA:
> @charminULTRA can you apply the corrections I suggested? then we can hopefully have the changes approved

I don't see any suggested corrections, where can I view that? @dewi-ny-je

charminULTRA:
> 
> Thanks James

@JamesH1978 How do I add a sign off if I already submitted the commit? 

Not entirely so sure why this is so rigorous considering the Klipper documentation still assumes use of OctoPrint lol

charminULTRA (commented):
Sign-off by Jonathan Williams <jcw122@gmail.com>

dewi-ny-je (changes requested):

docs/Measuring_Resonances.md
```diff
@@ -450,7 +450,7 @@ TEST_RESONANCES AXIS=Y
 ```
 This will generate 2 CSV files (`/tmp/resonances_x_*.csv` and
 `/tmp/resonances_y_*.csv`). These files can be processed with the stand-alone
-script on a Raspberry Pi. To do that, run the following commands:
+script on a Raspberry Pi. To do that, run the following commands (WARNING: Do not run these commands when there is more than one .csv file in your tmp folder. This will result in ALL files being used for the charted results):
```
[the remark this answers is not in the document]
charminULTRA: > Do not change this line: grammatically brackets are NOT meant to include whole sentences. If it's more than few words, it's not for brackets.
> 
> Add a line before line 458, after the code, as suggested:
> 
> ```
> 457bis Keep in mind that these commands (written with the "*") will merge all available CSV files
> from the specified folder to generate a single calibration graph. It may be useful if you tested the
> machine in various conditions and you want a single result which takes all of them into account.
> ```
> 
> Remember to split at column 80, as the file originally was, don't add long lines which make the code difficult to read. Lines are free of charge :)

I don't fully agree with this edit. There's no part of the documentation that already tells or assumes the user should or could use multiple CSV files, so the inherent assumption here is that the user is creating a single CSV. Given that, I don't think we should say things like "keep in mind", but more of a warning like "do not run this with multiple CSVs in your folder, unless you intend to blend the results".

I assume the original intent of using a wildcard was to save the user time from writing out the file name, but again it's not clear from the documentation.

Will submit a modified commit.

KevinOConnor:
Thanks.  Looks fine to me.  I'll wait a few days to see if @dmbutyugin has any comments.

-Kevin

docs/Measuring_Resonances.md
```diff
@@ -450,7 +450,10 @@ TEST_RESONANCES AXIS=Y
 ```
 This will generate 2 CSV files (`/tmp/resonances_x_*.csv` and
 `/tmp/resonances_y_*.csv`). These files can be processed with the stand-alone
-script on a Raspberry Pi. To do that, run the following commands:
+script on a Raspberry Pi. This script is intended to be run with a single CSV
```
dmbutyugin, line 453: Sorry for jumping late into the conversation. But I'd rephrase and expand that sentence to have a note after the commands, something like this:

...
script on a Raspberry Pi. To do that, run the following commands:
```
~/klipper/scripts/calibrate_shaper.py /tmp/resonances_x_*.csv -o /tmp/shaper_calibrate_x.png
~/klipper/scripts/calibrate_shaper.py /tmp/resonances_y_*.csv -o /tmp/shaper_calibrate_y.png
```
Note that '*' in the above commands acts as a wildcard for matching files, thus these commands
should usually be executed after one measurement per axis, with a single CSV file created for each
axis measured. If multiple CSV files were generated, then the script will process them all and
average the results. This can be useful, for example, if a resonance test was done at multiple test
points. On the other hand, if a user was, say, adjusting the belt tension between the test runs,
it is not valid to use the above commands as-is, and the user should either delete all but the latest
measurements, or specify the exact file names of the CSV files for the appropriate test run.
Following the commands above, the script will generate the charts ....

dmbutyugin (commented):
Thanks! I agree that the documentation is not obvious in this regard. I think it may be valuable to provide some additional context, so I wrote some suggestion on how to expand the text without suggesting that multiple files are not supported, but highlighting the typical usecases and that most users would use it after a single test run per axis.

charminULTRA:
Bump

docs/Measuring_Resonances.md
charminULTRA: How about this combination of your comment and my existing PR (placed before commands)

...
script on a Raspberry Pi. This script is intended to be run with a single CSV file for each axis measured, although it can be used with multiple CSV files if you desire to average the results of different resonance tests.  Averaging results can be useful, for example, if a resonance test was done at multiple test points. Delete the extra CSV files if you do not desire to average them.

To run the processing script, execute the following commands:
...

dewi-ny-je (approved):
I approve the final version proposed. I also think the suggestion by @dmbutyugin was too verbose.

KevinOConnor:
Okay, thanks.  If there are no further comments I'll look to commit in a few days.

-Kevin

dmbutyugin (approved):

KevinOConnor:
Thanks.

-Kevin

PR #6510, opened 2024-02-27
avr: enable small code size options for atmega32u4

jaketri:
Enable support for disabling optional features on atmega32u4 because size with all features enable is to big.


KevinOConnor:
Thanks.

-Kevin

PR #6513, opened 2024-02-29
Fixes build on Darwin by using target toolchain gcc 

apullin:
Target toolchain gcc with `-E` is now used for preprocessing.

Also changes `check-gcc.sh` to use target toolchain `readelf`, to make builds work.

This has been tested with `arm-none-eabi-` cross toolchain on OSX 14.3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,17 @@ export KCONFIG_CONFIG     := $(CURDIR)/.config
 
 # Common command definitions
 CC=$(CROSS_PREFIX)gcc
+CXX=$(CROSS_PREFIX)g++
 AS=$(CROSS_PREFIX)as
 LD=$(CROSS_PREFIX)ld
 OBJCOPY=$(CROSS_PREFIX)objcopy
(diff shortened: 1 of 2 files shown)


JamesH1978:

Thanks
James

bevanweiss:
Could you please check whether the more minimal changes in https://github.com/Klipper3d/klipper/pull/6422 do the job?
If they let things build under Darwin, then you could close this PR, and things would be fixed once the other PR is merged.

apullin:
> Could you please check whether the more minimal changes in #6422 do the job? If they let things build under Darwin, then you could close this PR, and things would be fixed once the other PR is merged.

Doesn't work. Still fails at linking, due to using `readelf` in `check-gcc.sh` .

apullin:
> Could you please check whether the more minimal changes in #6422 do the job? If they let things build under Darwin, then you could close this PR, and things would be fixed once the other PR is merged.

Appears to still fail when trying to find `readelf` when running `check-gcc.sh`.

PR #6515, opened 2024-03-03

TheParanoidEngineer:
**[Module]** 
Measuring_Resonances

**[Change Point]**
Changed `libopenblas-base` to `libopenblas-dev`.

**[Change Reason]**
Current instructions cause NumPy installation failure on host machines running Debian Bookworm.

**[Change Background/Evidence]**
Package Support table:
|  | libopenblas-base  | libopenblas-dev | Reference |
| ---| ----------------- | ------------------ | ----------- |
| Debian Buster  | Supported | Supported | [Buster openblas](https://packages.debian.org/source/buster/openblas) |
| Debian Bullseye | Supported  | Supported | [Bullseye openblas](https://packages.debian.org/source/bullseye/openblas) |
| Debian Bookworm | **Not Supported**  | Supported | [Bookworm openblas](https://packages.debian.org/source/bookworm/openblas) |

The table above shows that `libopenblas-base` is not supported by Bookworm, while `libopenblas-dev` is supported by all three.

This change _should_ have improved compatibility with host devices (especially with more devices moving to Bookworm).

**[Test Result]**
The new instructions have been tested on a fresh install of PiOS Lite (Bookworm) on a Pi5. NumPy (and supporting packages) were installed successfully without errors:
image

Signed-off-by: Philip Weber <philiprweb@gmail.com>

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -207,7 +207,7 @@ software dependencies not installed by default. First, run on your Raspberry Pi
 the following commands:
 ```
 sudo apt update
-sudo apt install python3-numpy python3-matplotlib libatlas-base-dev libopenblas-base
+sudo apt install python3-numpy python3-matplotlib libatlas-base-dev libopenblas-dev
 ```
 
 Next, in order to install NumPy in the Klipper environment, run the command:


dmbutyugin:
In principle, I see no problem with this change. It seems this should also work on ubuntu (and ubuntu-based armbian). However, the code contributions need to follow the [guidelines](https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review). Specifically here, the commit(s) need to have a signed-off-by line with the real name (see point 3 there).

TheParanoidEngineer:
> In principle, I see no problem with this change. It seems this should also work on ubuntu (and ubuntu-based armbian). However, the code contributions need to follow the [guidelines](https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review). Specifically here, the commit(s) need to have a signed-off-by line with the real name (see point 3 there).

Maybe I am not understanding, but is the below by-line in my original post not sufficient? 

image

Sineos:
The change makes sense. 
Indeed, I also missed the fact the `libopenblas-base` is handled differently between the major releases. `libopenblas-dev` is a common denominator.

Sineos (approved):

PR #6520, opened 2024-03-06
"static_value" in [output_pin enable_pin] is deprecated

1coderookie:
As the title says, "static_value" is deprecated, so I removed it and replaced it with "value".

### DIFF
--- a/config/printer-anycubic-kobra-go-2022.cfg
+++ b/config/printer-anycubic-kobra-go-2022.cfg
@@ -118,7 +118,7 @@ cycle_time: 0.00005 #20kHz
 
 [output_pin enable_pin]
 pin: PB6
-static_value: 1
+value: 1
     #This pin enables the bed, hotend, extruder fan, part fan.
 
 [mcu]


JamesH1978:

But yes it is deprecated, but we don't remove deprecations until a certain amount of time has passed then Kevin will make the decision to remove them completely.

Please read https://github.com/Klipper3d/klipper/blob/master/docs/Config_Changes.md Which explains this.

Thanks
James

1coderookie:
Sorry for the missing sign-off-line - since I don't really get how/where to add that now and since I don't understand the first line in that doc (with that "module:" example - like would it be "printer: Replaced deprecated `static_value` with `value`?), would it be sufficient if I just add that line here now? If so:   
Signed-off-by: Ulf Dieckmann <1coderookie@quantentunnel.de>  

I've read that doc about the config changes, that's actually why I thought it might be good to change it here in that example printer.cfg already. Because I thought better I remove it right now in that file and make a PR instead of forgetting to do so later and then a user has problems once it's finally being removed - but if you don't want to merge, it's totally fine with me! ;)

JamesH1978:
Apologies I may have jumped the gun on this one I thought you were actually trying to change the base code, and not a printer configuration...

In that case I will approve this.

Sorry about that,

James

JamesH1978 (approved):
This looks fine to me.

Thanks
James

KevinOConnor:
Thanks.

-Kevin

PR #6523, opened 2024-03-09
stm32: Fix USART3 ALT pinout on STM32G0

rakosi2:
I have a BTT E3EZ (STM32G0B1) and the UART on the TFT did not work. I have worked out that the ALT mux was not set correctly. This is the only board that I can verify with so I am not sure if this is a typo or if another chip needs it as AF7. I have not looked at the other USARTs since I can not verify it.

This will resolve this issue: https://github.com/bigtreetech/Manta-E3EZ/issues/8

image

Attila Rakosi <rattila5@hotmail.com>


JamesH1978:

Thanks
James

JamesH1978:
As mentioned over in the discord a lot of other stm32 chips use AF7 for this UART, Especially ones used in the 3D printing world like the F4 and the H7 series. You will need to rewrite your code to be specifically for the G0 series using AF0

Thanks
James

rakosi2:
Set it to only change for STM32G0 series to AF0 and all other STM32 will use AF7
I have also fix the [Signed-off-by](https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#format-of-commit-messages)
Thank you [JamesH1978](https://github.com/JamesH1978)

rakosi2:
I was unaware of it on the first commit, put it on the second commit and edited the first message with the signed-off at the end.

KevinOConnor:
Sorry - I'm still not seeing it.  I need to see a line that says `Signed-off-by: Some Name <some@email.addr>`.  The line can be either in a comment here or in a commit message - I don't see a line like that in either location.

Cheers,
-Kevin

rakosi2:
Signed-off-by: Attila Rakosi <rattila5@hotmail.com>

KevinOConnor:
Thanks.

-Kevin

PR #6524, opened 2024-03-09
Successfully flashed the firmware on my KingRoon KP5L

dick-jones:
I've successfully flashed the firmware using the flash-sdcard.sh script on my KingRoon KP5L with KingRoon KP3 v1.3 board.

### DIFF
--- a/lib/fatfs/ffconf.h
+++ b/lib/fatfs/ffconf.h
@@ -97,7 +97,7 @@
 */
 
 
-#define FF_USE_LFN		0
+#define FF_USE_LFN		1
 #define FF_MAX_LFN		255
 /* The FF_USE_LFN switches the support for LFN (long file name).
 /
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -128,6 +128,15 @@ BOARD_DEFS = {
         'mcu': "stm32g0b1xx",
         'spi_bus': "spi1",
         "cs_pin": "PB8"
+    },
+    'kingroon-kp3-v1.3-board': {
+        'mcu': "stm32f103xe",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        "cs_pin": "PC11",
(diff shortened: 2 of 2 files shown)


JamesH1978:
Thank you for submitting a PR, pleas refer to point 3 in "What to expect in a review" in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and provide a signed off by line. and please fix the failed checks.

Thanks
James

lib/fatfs/ffconf.h
```diff
@@ -97,7 +97,7 @@
 */
 
 
-#define FF_USE_LFN		0
+#define FF_USE_LFN		1
```
bevanweiss, line 100: Was this really required?
It's obviously something that might impact on every other board out there, so ideally if this could not be changed would be preferable.

scripts/spi_flash/board_defs.py
```diff
@@ -128,6 +128,15 @@
         'mcu': "stm32g0b1xx",
         'spi_bus': "spi1",
         "cs_pin": "PB8"
+    },
+    'kingroon-kp3-v1.3-board': {
```
bevanweiss, line 132: I think this would be better called 'mks-robin-nano' since that's what it sounds like the mainboard actually is.

scripts/spi_flash/board_defs.py
```diff
@@ -128,6 +128,15 @@
...
+        'mcu': "stm32f103xe",
+        'spi_bus': "swspi",
+        'spi_pins': "PC8,PD2,PC12",
+        "cs_pin": "PC11",
```
bevanweiss, line 136: It would be good if you could rebase this on master again, including the change to the conversion script.  Which will do away with the 'check_need_convert' change you have made in spi_flash.py also.

If you could try this with the sdio method that I have in PR https://github.com/Klipper3d/klipper/pull/6679 it would be appreciated also.  The SDIO mode means that it can do the verification also, so is a nicer usage flow if possible.

bevanweiss (changes requested):

PR #6525, opened 2024-03-10
scripts: update greenlet to 3.0.3

mkuf:
Installation of requirements fails starting at python 3.12.  
This is due to greenlet requirement of 2.0.2, which is not compatible w/ python 3.12. 

This PR updates greenlet to the latest release 3.0.3

### DIFF
--- a/scripts/klippy-requirements.txt
+++ b/scripts/klippy-requirements.txt
@@ -4,7 +4,8 @@
 #   pip install -r klippy-requirements.txt
 cffi==1.14.6
 pyserial==3.4
-greenlet==2.0.2
+greenlet==2.0.2 ; python_version < '3.12'
+greenlet==3.0.3 ; python_version >= '3.12'
 Jinja2==2.11.3
 python-can==3.3.4
 markupsafe==1.1.1


mkuf:
About the failed checks: 

It does seem that greenlet 3.0.3 is not available for python2. 
Merging this would essentially make the python requirements incompatible with python2.

If you want to go forward with this, I can provide a patch to remove the python2 checks from the review workflow and update the workflow to run with a more recent python3.12.  
This could either be done in the course of this PR or in a separate PR.  

Please provide feedback on how to proceed. 

-Markus

KevinOConnor:
I have noticed this conflict as well.  Alas, I don't have any good suggestions on how to best proceed.

-Kevin

mkuf:
I think I have found a less intrusive solution, which maintains compatibility with previous python versions `< 3.12` as well as versions` >= 3.12`.  

pip supports defining requirement specifiers, which allow to skip packages if certain conditions are not met.  
This way, when installing requirements on python `< 3.12`, greenlet 2.0.2 is used and with python `>= 3.12`, greenlet 3.0.3 is installed.

KevinOConnor:
Ah, thanks.  That seems like a good plan.  I'll give a few days to see if others have comments, but otherwise look to commit.

-Kevin

turk-fx:
> About the failed checks:
> 
> It does seem that greenlet 3.0.3 is not available for python2. Merging this would essentially make the python requirements incompatible with python2.
> 
> If you want to go forward with this, I can try to provide a patch to remove the python2 checks from the review workflow and update the workflow to run with a more recent python3.12. This could either be done in the course of this PR or in a separate PR.
> 
> Please provide feedback on how to proceed.
> 
> -Markus

I can try whatever you want me try. Just let me know what to do.

turk-fx:
Just a quick question. I am new to these type of stuff. So, I should wait till a new release or this plan is committed to a new package?

KevinOConnor:
Thanks.

-Kevin

Piraty:
`greenlet==3.0.3 ; python_version >= '3.12'`
this fails with python3.13 , you need to relax the constraint

RicRavGithub:
greenlet version 3.0.3 is not compatible with actual Pyton 3.13, as reported [here](https://greenlet.readthedocs.io/en/stable/changes.html).

I'm not at all an expert, but to me it seems that the following line of klippy-requirements.txt
`greenlet==3.0.3 ; python_version >= '3.12'`
forces the installation of a greenlet version that clearly doesn't work, as trying to install klipper I get an error:

```
ERROR: Failed building wheel for greenlet
Failed to build cffi greenlet
ERROR: Failed to build installable wheels for some pyproject.toml based projects (cffi, greenlet)
```

PR #6527, opened 2024-03-10
[manual_stepper] Add basic status.

viesturz:
Adding position and enabled in manual_stepper status. Enabled is already available through stepper_enable object. But this makes it more straightforward to access it.


KevinOConnor:
Thanks.  It seems fine to me.  I didn't understand the comment `Note that the stepper may have not reached this position yet if SYNC=1 is used.`  The returned position is always the last requested position, which may differ from the actual position - and as far as I know has no relation to the SYNC command-line option.  Am I missing something?

Cheers,
-Kevin

viesturz:
Oh I meant the opposite, that stepper may not have reached the position yet if sync=0.

Or best to remove that remark altogether?

viesturz:
If SYNC=1 then the command waits for the motor to reach the requested position before returing, effectively ensuring any readers  that the actual position will be the same as commanded position.

Updated the documentation.

viesturz:
I see, thanks, will update the PR tomorrow.

I re-read the code and just to confirm this is my understanding now:
-moves of same manual stepper are sequenced, regardless of sync
- sync ensures next toolhead move is after manual move AND next manual move
is after toolhead moves.
- no explicit way to sequence a manual move starting after a toolhead move.
A zero distance manual move with sync=1 may work as a hack.

- moves between multiple manual steppers cannot be coordinated directly.
Again zero distance move hack could work.

KevinOConnor:
> - moves of same manual stepper are sequenced, regardless of sync

Yes.

> - sync ensures next toolhead move is after manual move AND next manual move
is after toolhead moves.

A manual_stepper move always starts after the completion of the last requested toolhead move, regardless of SYNC.  A SYNC=0 allows a following toolhead move to start prior to the nominal completion of the manual_stepper move.

> - no explicit way to sequence a manual move starting after a toolhead move.

This is always done.  No current way to not do it.

-Kevin

viesturz:
Thanks for the clarification.

This line threw me off, is that redundant?
https://github.com/Klipper3d/klipper/blob/75a40e817db4d5bc9d7d893fad499907d2b910a2/klippy/extras/manual_stepper.py#L105

KevinOConnor:
That line only gets run if a command does not contain a MOVE.  As I understand it, it could be used for explicit syncing of upcoming toolhead moves - possibly useful after some number of moves with `SYNC=1`.  For example, something like `MANUAL_STEPPER MOVE=...some 20 second move...`, `G1 X...some 10 second move...`, `MANUAL_STEPPER SYNC=1`, `G1 X...` - which would allow the first `G1` to overlap, but not the second `G1`.

-Kevin

viesturz:
Thanks, fixed the documentation.

KevinOConnor:
Thanks.

-Kevin

gearslam:
This just causes "Unhandled exception during run" for me.

viesturz:
Seriously? I'm cursed, 2/2 PRs broke stuff.

gearslam:
What's strange is only half of the Fluidd interface loads.

KevinOConnor:
I reverted this commit (in commit 4cfa266e).  The MCU_Stepper class does not have a `is_motor_enabled()` method, so this change just results in an internal exception.

-Kevin

PR #6528, opened 2024-03-10
Docs: sample-macros.cfg: minor change to avoid confusion

JamesH1978:
It has been noted a couple of times that this file states to only use the bare macro names in the slicer. Which is fine if you want to set your own temperatures in klipper but not if you want to pass the actual temperatures through as mentioned in the slicers.md doc linked. This minor change is aimed at causing less confusion.

Thanks
James

### DIFF
--- a/config/sample-macros.cfg
+++ b/config/sample-macros.cfg
@@ -10,7 +10,11 @@
 ######################################################################
 
 # Replace the slicer's custom start and end g-code scripts with
-# START_PRINT and END_PRINT. See docs/Slicers.md for more information on using these macros.
+# START_PRINT and END_PRINT. Please note that using just these g-codes
+# will mean only the default temperatures set in the macro will ever
+# be used. To pass in actual slicer set temperaturees, please see
+# docs/Slicers.md for more information on slicer variables
+# to use.
 
 [gcode_macro START_PRINT]
 gcode:
--- a/config/sample-macros.cfg
+++ b/config/sample-macros.cfg
@@ -12,7 +12,7 @@
 # Replace the slicer's custom start and end g-code scripts with
 # START_PRINT and END_PRINT. Please note that using just these g-codes
 # will mean only the default temperatures set in the macro will ever
-# be used. To pass in actual slicer set temperaturees, please see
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I'm not sure I understand the new comments.  

Maybe an alternative would be to change the existing reference to Slicers.md to something like "Be sure to read docs/Slicers.md before using these macros."

Cheers,
-Kevin

JamesH1978:
Basically some people were not either reading slicers.md and not realizing that to actually get the temperatures passed from a slicer they had to actually pass through the variables/parameters in the slicer gcode(and not reading the linked doc), Or they were viewing it as ignoring the sliceres temperature and just setting the defaults to what they want in the macro as defaults. This change is just to explain this and point them towards the file that was already pointed to for further information.

Does that make any more sense?

Thanks
James

KevinOConnor:
I understand.  However, I fear if I did not already know what the macros did, I would not understand the new comments.  

I guess I'm wondering if it is useful to explain the low-level implications of the macros to those that don't want to read the documentation - as they wont read/understand these new comments either.  Thus maybe we just need to make it more succinct that reading the documentation is necessary.

Thanks for working on this.  We can certainly reword the comments and I'm open to any wording.
-Kevin

JamesH1978:
The other option we have is we just put the correct syntax including the parameter pass in the sample macro file and tell them to read the documentation for more information.

Thanks
James

flowerysong:
Maybe something more like:

```
# START_PRINT and END_PRINT are intended to be called from your slicer's start and end g-code.
# See docs/Slicers.md for more information on using these macros.
```

It avoids the confusing "replace" language, but doesn't attempt to duplicate the documentation or go into excessive detail about the limitations of a specific way of calling the macros that isn't the way most people should be using them.

KevinOConnor:
> The other option we have is we just put the correct syntax including the parameter pass in the sample macro file and tell them to read the documentation for more information.

That's fine.

We could also just set the temperatures explicitly in the sample macros.  That is, if users are confused by the complexity of the example then we can simplify the example.

I'm open to other alternatives as well.

Cheers,
-Kevin

PR #6529, opened 2024-03-13
stm32: Add i2c3_PC0_PC1 for stm32g0

bigtreetech:


KevinOConnor:
Thanks.

-Kevin

PR #6531, opened 2024-03-13
linux: Allow for more i2c buses

CRTified:
Similar to commit df79893, this allows klipper to use up to /dev/i2c-14. Similar to before, this limit is arbitrary.

This is required for some other SoCs, which have even more i2c buses available, e.g. the rk3399:

```
$ ls -1 /dev/i2c-*
/dev/i2c-0
/dev/i2c-3
/dev/i2c-7
```


KevinOConnor:
Thanks.

-Kevin

PR #6536, opened 2024-03-21
Support for ldc1612 eddy current probes

KevinOConnor:
This PR adds support for the ldc1612 inductive sensor chip.  Like regular inductive sensors, this chip can be used to implement a bed probe.  Unlike traditional inductive sensors, sensors using this chip can provide the raw "resonance frequency" of the inductive coil.  These raw values can be used to estimate the distance between toolhead and bed.  This may improve the probing accuracy and may enable faster probing.

The code here provides initial probing support.  There is some information on using the code in the new Eddy_Probe.md file.

-Kevin


kmobs:
Will this support beacon too?

KevinOConnor:
> Will this support beacon too?

I don't know, as I'm not familiar with the low-level details of that hardware.  The support here is for probes that expose the ldc1612 chip via standard i2c.

-Kevin

kmobs:
> > Will this support beacon too?
> 
> I don't know, as I'm not familiar with the low-level details of that hardware. The support here is for probes that expose the ldc1612 chip via standard i2c.
> 
> -Kevin

Oh sorry, I guess I just assumed because it is genre defining in this space that it would be supported with the new code as well.

j90141355:
> 这也支持信标吗？

应该是支持的

ProfDrYoMan:
> > Will this support beacon too?
> 
> I don't know, as I'm not familiar with the low-level details of that hardware. The support here is for probes that expose the ldc1612 chip via standard i2c.
> 
> -Kevin

Can you link one of the existing probes?

sbtoonz:
> > > Will this support beacon too?
> > 
> > 
> > I don't know, as I'm not familiar with the low-level details of that hardware. The support here is for probes that expose the ldc1612 chip via standard i2c.
> > -Kevin
> 
> Can you link one of the existing probes?

https://docs.duet3d.com/en/Duet3D_hardware/Duet_3_family/Duet_3_Scanning_Z_Probe

KevinOConnor:
> Can you link one of the existing probes?

I tested with a prototype board from BTT - I'm not sure if they're shipping that board yet.  My high-level understanding is that the "cartographer probe" also uses the ldc1612 (though I could be wrong on that, and I did not test it).  It seems there is a Duet board as well (as linked above), but I also have not tested that.

Cheers,
-Kevin

PR #6538, opened 2024-03-22
Populate minimum_cruise_ratio to printer.configfile.settings

KevinOConnor:
The default minimum_cruise_ratio setting does not get populated to the printer.configfile.settings information due to the way the max_accel_to_decel backwards compatibility support was implemented. Slightly rework the config reading so that the default for minimum_cruise_ratio is populated there.

Reported by @ReXT3D.

-Kevin


ReXT3D:
Thanks Kevin!

PR #6539, opened 2024-03-22
Fix virtual_sdcard handling of unicode characters on Python2

KevinOConnor:
PR #6472 fixed unicode handling on Python3, but broke Python2 support.  

This could lead to error messages like `UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 12: ordinal not in range(128)` on Python2.

Use an alternate fix that should work for both Python3 and Python2.

@zxy16305, @CODeRUS - fyi.

-Kevin

### DIFF
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -1,9 +1,9 @@
 # Virtual sdcard support (print files directly from a host g-code file)
 #
-# Copyright (C) 2018  Kevin O'Connor <kevin@koconnor.net>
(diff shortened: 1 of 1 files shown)


zxy16305:
This change looks great.
Sorry for only consulting ChatGPT-compatible methods without testing.  ><. hh

thijstriemstra:
Anyone still using Python 2 should have a really really really good reason not to switch to Python 3 (but I really can't think of any reason). A friendly reminder that support for Python 2 ended 4 years ago (01 Jan 2020).

PR #6549, opened 2024-04-04
Add kconfig options for rp2040 uart

amken3d:
Current implementation only supports UART0 on Pin0 and 1. 
This change brings in all the possible UART 0 and 1 pin combinations for the RP2040.
Has been tested on a custom board with UART1 on 24,25

### DIFF
--- a/src/rp2040/serial.c
+++ b/src/rp2040/serial.c
@@ -2,6 +2,7 @@
 //
 // Copyright (C) 2021  Kevin O'Connor <kevin@koconnor.net>
 //
+// This file may be distributed under the terms of the GNU GPLv3 license.
 
 
 #include <stdint.h> // uint32_t
--- a/src/rp2040/serial.c
+++ b/src/rp2040/serial.c
@@ -2,8 +2,7 @@
 //
 // Copyright (C) 2021  Kevin O'Connor <kevin@koconnor.net>
 //
-// This file may be distributed under the terms of the GNU GPLv3 license.
-//Modified 04/04/2024 by Amken3d (info@amken.us) to add all UART pin combos
+
 
 #include <stdint.h> // uint32_t
 #include "autoconf.h" // Include configuration header
--- a/src/rp2040/serial.c
+++ b/src/rp2040/serial.c
@@ -3,14 +3,14 @@
 // Copyright (C) 2021  Kevin O'Connor <kevin@koconnor.net>
 //
 // This file may be distributed under the terms of the GNU GPLv3 license.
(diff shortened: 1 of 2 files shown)


JamesH1978:

Thanks
James

amken3d:
I have included the sign off line for each commit and rebased. 
Thank you

src/rp2040/serial.c
```diff
@@ -67,10 +104,17 @@ serial_enable_tx_irq(void)
...
 serial_init(void)
 {
+
+#if CONFIG_RP2040_SERIAL_UART0
     enable_pclock(RESETS_RESET_UART0_BITS);
+    uint32_t pclk = get_pclock_frequency(RESETS_RESET_UART0_BITS);
+#elif CONFIG_RP2040_SERIAL_UART1
+    enable_pclock(RESETS_RESET_UART1_BITS);
+    uint32_t pclk = get_pclock_frequency(RESETS_RESET_UART1_BITS);
+#endif
```
KevinOConnor, line 114: Please avoid `#if` in the executable code - prefer `if (CONFIG_RP2040_SERIAL_UART0) ...`

KevinOConnor (commented):
Thanks.  We can certainly add additional serial options for the rp2040.  However, please layout the user facing menu (and backend Kconfig code) in the same style as the stm32 serial selection menu.  That is, we want to have a single flat menu with all the meaningful user facing communication options.  Also, like on stm32, it's probably a good idea to tie the less used serial options to CONFIG_LOW_LEVEL_OPTIONS.

I also have a minor comment on the implementation - see below.

-Kevin

amken3d:
@KevinOConnor, thank you for the feedback. I have made the changes you suggested.

image

image

image

src/rp2040/serial.c
```diff
@@ -3,21 +3,65 @@
 // Copyright (C) 2021  Kevin O'Connor <kevin@koconnor.net>
 //
 // This file may be distributed under the terms of the GNU GPLv3 license.
+//Modified 04/04/2024 by Amken3d (info@amken.us) to add all UART pin combos
```
KevinOConnor, line 6: Please don't add changelog information to the code - use the git commit messages for describing and documenting history.  There's a brief note on this at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review .  Sorry - I missed this in the previous review.  Otherwise the changes look fine to me.

Thanks,
-Kevin

amken3d:
Done

KevinOConnor:
Thanks.

-Kevin

PR #6555, opened 2024-04-08
Bulk ADC Sensors

garethky:
Add Bulk ADC Sensors and Load Cell skeleton

* Add support for the [ADS1220](https://www.ti.com/product/ADS1220), HX717 and HX711  ADC sensors.
* Add a `load_cell` skeleton to create the sensors and start taking samples.

## Is this Bug Free / Reliable?
Versions of the HX711/HX717 code have been running for weeks at a time on my test system. Others have tested this code as well and we are pretty confident this version doesn't suffer form any timing defects.

The ADS1220 code is newer but has been running for at least a week non-stop without an error. Since its SPI based it should be more reliable. The only quirk with this chip is that a reset is required before configuring it.

The only issue I'm aware of is with `_finish_measurements()` running when the printer experiences a SHUTDOWN. This causes the communication to the MCU to fail and logs an exception message. This is confusing for users as its at the end of the log but not the root cause of the shutdown. All of the other Bulk Sensors share this behavior. It would be nice if we could detect a shutdown in progress and skip the serial command in `_finish_measurements()`.

## Build Changes
Bulk Sensor assumes that a sensor will be SPI or I2C. Currently it would be automatically excluded on small platforms that don't have these interfaces. But the HX71x chip family uses bit-banging and can run on anything with GPIO. So I altered the build to allow the HX71x chips to be turned off. If they are included they also bring in Bulk Sensor. I'm not sure which platforms we might want to exclude this on or if it was worth bothering about.

## What doesn't this do?
* The implementation of a digital scale is cut from this first PR to keep the size small. The plan is to put that code in the next PR.

* These sensors are all "multiplex" ADCs, meaning they have a single measurement core that is connected to a multiplexer that allows the input channel to be switched at runtime. But they can only produce 1 measurement at a time and some delay is required to switch channels (often a couple of measurement cycles). This code does not make this feature usable, in fact for the ADS1220 I didn't even include an option to select an input channel. This feature has been used by Prusa in their MK4 and XL products. They use the 'A' channel on the HX717 to real the load cell and the 'B' channel to read the the analog filament sensor. They switch inputs at runtime at a regular duty cycle and rout the measurements to separate components. I decided this is out of scope due to the complexity it would create.


garethky:
Looks like the build failed due to the increased binary size on small platforms. I'll have to do an update to exclude the HX71x on those.

sbtoonz (approved):

eazrael:
Sorry for the thermocouple discussion. Have never seen thermocouples before this project. When I made my "finger test", the thermocouple was directly connected to the mainboard, without the extruder PCB and the long cables between them, which also carry heating & extruder power lines. And I doubt they have alumel/chromel PCB traces and cables. So there was only 20cm between them, but it still showed that the ADC is probably not a reliable cold-junction. And will probably never be as the temperature range of the ADC is neglectible when compared to the thermocouples.

I need to do some calculations, check what the error in the usual extruder temperature range is and how to how to fix that for the FFCP2.

garethky:
Not sure I follow how that is relevant to this PR?

eazrael:
Not at all, this was supposed to go into my Pr. Sorry :-(

docs/Config_Reference.md
```diff
@@ -4466,6 +4466,100 @@ adc2:
...
+#   This must be one of the supported Bulk ADC Sensor types
+```
+
+## Bulk ADC Sensors
+
+Klipper includes support for generic Analog to Digital Converter (ADC) chips
+that can sample at a high data rate. These sensors measure a voltage that can be
+converted into various measurements (temperature, force etc.) by other modules.
+These sensors may be used in any config section that accepts a Bulk ADC (such as
+a `[load_cell]`).
```
KevinOConnor, line 4487: I don't think it is a good idea to expose the "bulk adc" abstraction to end-users.  The loadcell documentation and following hx711/hx717/ads1220 seem fine, but I don't see a reason to complicate things with a "bulk adc" intermediary.  Seems simpler to describe this all as part of the loadcell implementation.  If at some future point we need to describe some other (non-loadcell) chip that uses these sensors then we can tackle how to document that at that later point.

klippy/extras/ads1220.py
```diff
@@ -0,0 +1,203 @@
...
+NOOP_CMD = 0x0
+RESET_STATE = bytearray([0x0, 0x0, 0x0, 0x0])
+
+
+# turn bytearrays into pretty hex strings: [0xff, 0x1]
+def hexify(byte_array):
+    return "[%s]" % (", ".join([hex(b) for b in byte_array]))
+
+
+class ADS1220(BulkSensorAdc):
```
KevinOConnor, line 30: I'd generally prefer not to define base classes like this in python - in particular when the base class is in another python file.  In Klipper the coding convention is to not touch the internal member variables of a class if that class is defined in another python file.  Alas, in python, when using base classes like this, it becomes really difficult to identify and maintain these types of dependencies between class instances.

I understand that in this case, the base class has no member variables and does not access any member variables.  However, it also seems like the base class just serves as documentation, and I think it would be easier to maintain the code going forward if comments were used for that purpose instead of a base class.

klippy/extras/ads1220.py
```diff
@@ -0,0 +1,203 @@
...
+        return self.mcu
+
+    def get_samples_per_second(self):
+        return self.sps
+
+    # returns a tuple of the minimum and maximum value of the sensor, used to
+    # detect if a data value is saturated
+    def get_range(self):
+        range_max = (2 ** 24)
+        return -range_max, range_max
```
KevinOConnor, line 105: FWIW, I'm not sure what your plans are for `get_range()`, but as a suggestion, it's probably simpler to normalize the adc values to a float between -1.0 and 1.0.  That eliminates the need for these types of range checks, makes it easier to scale values to voltages or grams, and eases testing when switching to new sensor chips.

klippy/extras/ads1220.py
```diff
@@ -0,0 +1,203 @@
...
+            return {}
+        return {'data': samples}
+
+    def reset_chip(self):
+        # the reset command takes 50us to complete
+        self.send_command(RESET_CMD)
+        # TODO: is a manual 50us delay needed here? Testing says it's not.
+        # read startup register state and validate
+        val = self.read_reg(0x0, 4)
+        if hexify(val) != hexify(RESET_STATE):
```
KevinOConnor, line 164: FWIW, it's certainly seems odd to encode two numeric objects into a string to compare them..

klippy/extras/bulk_adc_sensor.py
```diff
@@ -0,0 +1,54 @@
...
+    # subscribe to data being published from the sensor
+    # callback is called with a list containing (time, counts) tuples:
+    # [(time, counts), (time, counts), ...]
+    # The callback must return True to continue receiving data and False to stop
+    def add_client(self, callback):
+        pass
+
+# Helper for ClockSyncRegression that handles generating timestamps
+# while processing a batch of samples
+class TimestampHelper:
```
KevinOConnor, line 29: I agree it would be good to use shared code for the intricate timestamp manipulation that the bulk sensors implement.  However, I think this should be done for all bulk sensors and implemented in bulk_sensor.py .  I'll put together a separate PR for discussion.

klippy/extras/bulk_sensor.py
```diff
@@ -209,7 +209,7 @@ def __init__(self, clock_sync, bytes_per_sample):
         self.last_sequence = self.max_query_duration = 0
         self.last_overflows = 0
         self.mcu = self.oid = self.query_status_cmd = None
-    def setup_query_command(self, mcu, msgformat, oid, cq):
+    def setup_query_command(self, mcu, msgformat, oid, cq=None):
```
KevinOConnor, line 212: It does not make sense to use the "default cmdqueue" for any of the bulk sensors.  So, this change does not seem correct.

For hx71x.py, just create a cmdqueue with `mcu.alloc_command_queue()`.

klippy/extras/hx71x.py
```diff
@@ -0,0 +1,168 @@
...
+BYTES_PER_SAMPLE = 4  # samples are 4 byte wide unsigned integers
+MAX_SAMPLES_PER_MESSAGE = bulk_sensor.MAX_BULK_MSG_SIZE // BYTES_PER_SAMPLE
+UPDATE_INTERVAL = 0.10
+
+
+# Implementation of HX711 and HX717
+# This supports up to 4 sensor being read in parallel for under-bed load cell
+# applications. It exposes a sum of all sensor outputs via subscribe(). Also,
+# each sensor's data can be read individually via subscribe_sensor(1-4). If any
+# sensor becomes saturated the output of the sum is the saturated value.
```
KevinOConnor, line 23: These comments seems stale.

klippy/extras/load_cell.py
```diff
@@ -0,0 +1,38 @@
...
+        self.printer = printer = config.get_printer()
+        self.sensor = sensor   # must implement BulkAdcSensor
+        # startup, when klippy is ready, start capturing data
+        printer.register_event_handler("klippy:ready", self._handle_ready)
+
+    def _handle_ready(self):
+        self.sensor.add_client(self._on_sample)
+
+    def _on_sample(self, msg):
+        return True
```
KevinOConnor, line 21: Why does this code start sampling at startup?

src/sensor_ads1220.c
```diff
@@ -0,0 +1,206 @@
...
+set_flag(uint8_t mask, struct ads1220_adc *ads1220)
+{
+    ads1220->flags |= mask;
+}
+
+static inline void
+clear_flag(uint8_t mask, struct ads1220_adc *ads1220)
+{
+    ads1220->flags &= ~mask;
+}
```
KevinOConnor, line 55: Seems like dead code.  is_flag_set() and clear_flag() are never used.  Nothing ever clears the flag, so set_flag() is redundant..

src/sensor_ads1220.c
```diff
@@ -0,0 +1,206 @@
...
+    sched_wake_task(&wake_ads1220);
+    return SF_DONE;
+}
+
+// Helper code to reschedule the ads1220_event() timer
+static void
+ads1220_reschedule_timer(struct ads1220_adc *ads1220)
+{
+    irq_disable();
+    set_flag(FLAG_PENDING, ads1220);
```
KevinOConnor, line 73: This code doesn't look right.  It's not clear to me what this flag is intended to do.  If it is a port from sensor_adxl345.c, then it should be clearing the flag here.

src/sensor_ads1220.c
```diff
@@ -0,0 +1,206 @@
...
+
+    // data is ready
+    uint8_t msg[3] = {0, 0, 0};
+    uint32_t start_time = timer_read_time();
+    spidev_transfer(ads1220->spi, 1, sizeof(msg), msg);
+    uint32_t time_diff = timer_read_time() - start_time;
+
+    if (time_diff >= MAX_SPI_READ_TIME) {
+        // some IRQ delayed this read so much that its unusable
+        shutdown("ADS1220 read timing error, read took too long");
```
KevinOConnor, line 121: I don't think we should issue a `shutdown()` in any of this sensor code.  The code should encode an error of some kind in the sensor reading and then let the python code figure out how to handle the event.

Same comment for the sensor_hx71x.c code.

src/sensor_ads1220.c
```diff
@@ -0,0 +1,206 @@
...
+
+    if (counts == -1) {
+        shutdown("ADS1220: Possible bad read");
+    }
+
+    if (counts >= 0x800000) {
+        counts |= 0xFF000000;
+        shutdown("ADS1220: Invalid Counts");
+    }
+    add_sample(ads1220, counts);
```
KevinOConnor, line 137: Using 4 bytes for a 3-byte reading seems wasteful of cpu and bandwidth resources. It would seem better to encode this into the bulk_sensor report using 3-bytes.

Separately, a sample read error could be encoded as 0x800000, and if the sensor actually reports 0x800000 we can turn that into 0x800001 to avoid a conflict.

src/sensor_hx71x.c
```diff
@@ -0,0 +1,267 @@
...
+    gpio_out_write(hx71x->sclk, 1);
+    hx71x_delay_no_irq(hx71x_get_time(), MIN_PULSE_TIME);
+    gpio_out_write(hx71x->sclk, 0);
+    hx71x_delay_no_irq(hx71x_get_time(), MIN_PULSE_TIME);
+    irq_enable();
+}
+
+// hx71x ADC query
+void
+hx71x_read_adc(struct hx71x_adc *hx71x, uint8_t oid)
```
KevinOConnor, line 156: Same comment on using 4-bytes when 3-bytes is sufficient.  Same comment on not invoking shutdown().

src/sensor_hx71x.c
```diff
@@ -0,0 +1,267 @@
...
+    for (uint_fast8_t sample_idx = 0; sample_idx < 24; sample_idx++) {
+        hx71x_pulse_clocks(hx71x);
+        // read 2's compliment int bits
+        counts = (counts << 1) | gpio_in_read(hx71x->dout);
+    }
+
+    // bit bang 1 to 4 more bits to configure gain & channel for the next sample
+    for (uint8_t gain_idx = 0; gain_idx < hx71x->gain_channel; gain_idx++) {
+        hx71x_pulse_clocks(hx71x);
+    }
```
KevinOConnor, line 175: As a minor note, this code could be optimized a bit.  Something to look at after merging though.

src/sensor_hx71x.c
```diff
@@ -0,0 +1,267 @@
...
+    struct hx71x_adc *hx71x = oid_alloc(args[0]
+                , command_config_hx71x, sizeof(*hx71x));
+    hx71x->timer.func = hx71x_event;
+    hx71x->pending_flag = 0;
+    uint8_t gain_channel = args[1];
+    if (gain_channel < 1 || gain_channel > 4) {
+        shutdown("HX71x gain/channel out of range 1-4");
+    }
+    hx71x->gain_channel = gain_channel;
+    hx71x->dout = gpio_in_setup(args[2], -1);
```
KevinOConnor, line 213: A gpio pull down is not supported on all chips.  So, this look strange.  Could a pull up be used instead?

src/sensor_hx71x.c
```diff
@@ -0,0 +1,267 @@
...
+
+void
+command_query_hx71x_status(const uint32_t *args)
+{
+    uint8_t oid = args[0];
+    struct hx71x_adc *hx71x = oid_lookup(oid, command_config_hx71x);
+    const uint32_t start_t = timer_read_time();
+    uint8_t pending_bytes = 0;
+    pending_bytes = hx71x_is_data_ready(hx71x) * BYTES_PER_SAMPLE;
+    const uint32_t end_t = timer_read_time();
```
KevinOConnor, line 247: It seems odd to perform the multiply in the timed critical section above.  A multiply can be several cycles on some chips.  (Realistically, gcc could move the code around, but no reason to hint to gcc that it should add code into the critical section.)

src/sensor_hx71x.c
```diff
@@ -0,0 +1,267 @@
...
+// Background task that performs measurements
+void
+hx71x_capture_task(void)
+{
+    if (!sched_check_wake(&wake_hx71x))
+        return;
+    uint8_t oid;
+    struct hx71x_adc *hx71x;
+    foreach_oid(oid, hx71x, command_config_hx71x) {
+        if (hx71x->pending_flag) {
```
KevinOConnor, line 262: Same comment as before - I'm not sure what is intended with the pending_flag, as it is only ever set during sampling.  Something doesn't look correct.

KevinOConnor (commented):
Thanks!  In general it looks good to me.  I have some comments.

Procedurally it would be preferable if the hx71x and ads1220 code were in separate commits.  Breaking up the commits helps with tools like `git blame` and `git bisect`.  It seems like the two are mostly separated anyway, so hopefully it is not hard to create two atomic commits.  I can also help with this.

I noticed a few, mostly minor, things in the code.  See my comments below.

-Kevin

src/sensor_hx71x.c
garethky: I can make this a ternary

klippy/extras/load_cell.py
garethky: In the next PR the load cell will report the current weight continuously as part of its status, similar to a temperature sensor. This is the smallest part of that I could think if leaving it to get something that actually runs the sensors.

src/sensor_ads1220.c
garethky: Yep, I think I refactored this out but never deleted them, will fix.

src/sensor_ads1220.c
garethky: I agree with you in spirit. I don't like the idea of this ruining a print. I had an implementation of this but it was complex.

If something goes wrong with these chips (particularly the HX71x [reference](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/7c651de64ad3973f8faa10d769d3156d9b2e288a/src/common/hx717.cpp)) the recovery path is to:
* Stop measurements
* Put the chip into a reset state for some minimum time
* Bring the chip out of reset and re-configure it
  * For the HX71x this require running the measurement loop and discarding the first sample
* Restart measurements

While this reset is going on, any process that's consuming measurements now doesn't get them. This might be an issue for the consumer. E.g. calibrating the scale: getting an error should abort the gcode command. So I cant really hide the reset event from downstream systems. Either its communicated as a value, or by disconnecting any clients of the read process.

This also raises the specter of reset loops. When a chip has ESD issues it will fault frequently. 1 or more times per second. How many faults before we shutdown the system?

(Resets on the MCU during probing are handled by the watch dog thread in the load cell endstop. So that part at least is resilient already.)

src/sensor_ads1220.c
garethky: Agree we can encode errors as "-0", 0x800000 or 0x80000000.

To support the endstop on the MCU the sample has to be converted to 32 bits. I didn't want to do this conversion and not see the result in Python. I was/am worried that the conversion is done correctly. Sending back the 32 bits gives me that visibility.

Neither of these sensors can sample so fast that I was worried about the overhead. On the ADS1263, it can sample at 32Khz, but it is 32 bits anyway. The ADXL345 is sending back more data per measurement that this and sampling faster. So I judged the overhead worth the visibility.

klippy/extras/ads1220.py
garethky: That's a good idea, but doesn't work out well later on when I need the integer values to configure the endstop. That works on raw counts, since we cant use floats in the protocol. There would have to be a way to convert back into raw counts which the sensor would need to expose. So one way or another this info will leak from the abstraction. Maybe we can revisit this idea when that code is added.

docs/Config_Reference.md
garethky: As you wish.

I was just trying to highlight that these sensors are not 'load cell sensors'. For anyone new to the codebase its not obvious that they can re-use these if they read the documentation.

klippy/extras/bulk_sensor.py
garethky: Sure I can do that.

This is very non-obvious. The implementations down the stack don't allocate a command queue if one isn't provided, so I figured this was just an oversight. Maybe you should accept `None` here and do the allocation here because `bulk_sensor` has this special requirement?

klippy/extras/ads1220.py
garethky: Well I had to try 🤷‍♂️ 

I'd call this an interface, python calls this an [Abstract Base Class](https://docs.python.org/2.7/glossary.html#term-abstract-base-class) but cant explain what they mean without using the word interface. This is a mechanism. Its discoverable by your IDE (cmd+click takes you right there), it can be checked at build time and run time. With Python 3 type annotations you could enforce that methods accept this type. With 2.7 you can still do type checking at runtime.

I think you are missing out on a simple bit of tooling that could help clearly communicate your intentions in a discoverable way. Responding to everyone that didn't find the right code/doc/comment is exhausting in practice.

src/sensor_hx71x.c
KevinOConnor: Just to be clear, I was referring to performing operations between start_t and end_t.  I wasn't commenting on the operations themselves.

On second look though, you might want to do `irq_disable(); timer_read_time(); hx71x_is_data_ready(); irq_enable();` instead as that provides better timing and has less overhead than two `timer_read_time()` calls.  It's valid on the hx71x because `gpio_in_read()` is safe to run with irqs disabled.  When doing that, it is okay to pass `0` to the time diff field of `sensor_bulk_status()`.

src/sensor_ads1220.c
KevinOConnor: FYI, adxl345 has 3200 samples per second, and 5 bytes per sample are sent to the host.  (So, 10x slower max sampling rate than ADS1283).

src/sensor_ads1220.c
KevinOConnor: Just to be clear, the concern is not with raising an error - the concern is in using `shutdown()` to indicate the error.  A `shutdown()` from the mcu provides little flexibility in handling and reporting the situation - heaters, motors, displays, leds, etc. are all disabled and it is often necessary to analyse the log to determine what happened.

We really want to propagate the error back to the host code so that it can handle it.  If the host code raises an error (or shutdown) that's fine.

You can take a look at how the adxl345 code checks for error conditions and introduces error codes into the sample stream.  The adxl345.py code then checks for those error codes and propagates that information via statistics.

klippy/extras/load_cell.py
KevinOConnor: For testing, you can use the motan interface - something like: `~/klipper/scripts/motan/data_logger.py /tmp/klippy_uds /tmp/mytestdata`

You may need to modify `scripts/motan/data_logger.py` to enable queries by default.  (Or, otherwise use `scripts/whconsole.py` and manually issue the subscribe requests.)

As part of the initial commit, we do want some minimal functionality - I think the web interface is sufficient for that requirement.  I think we should avoid "debugging code" as part of that first commit.

klippy/extras/bulk_sensor.py
KevinOConnor: FYI, there isn't a special requirement for bulk sensors.  We want all devices to have their own queue so that messages to those devices are ordered wrt themselves, but not strictly ordered wrt other devices.  This allows the message priority system to prioritize urgent messages without corrupting ordering of a message stream to a device.

The low-level code can't (easily) assign command queues on its own, as it needs to know all the commands that could be sent to that device.  That is, all of "query_adxl345", "query_adxl345_status", "spi_send", and "spi_transfer" sent to a single adxl345 need to have strict ordering, and no code other than adxl345.py knows that.

src/sensor_hx71x.c
garethky: ok, sounds good

src/sensor_ads1220.c
garethky: the max sampling rate on this chip, ADS1220 is 2000 samples per second. So less than ADXL345.

garethky:
I have a busy few weeks coming up and I'm not sure when I'm going to get to all the comments. If I can grab a few minutes here and there ill knock one out and just tack on a commit. When its all done we can squash git history into 2 commits as requested.

I'm not sure how I'm going to prove the sensor reset stuff works. I cant force an error on command without adding extra code. I might put in a counter that forces a reset from the MCU side every 30 seconds or something and then take that our for the submission. Will probably save this for last.

src/sensor_hx71x.c
garethky: Looks like a pull-up works (on my hardware at least)

src/sensor_hx71x.c
garethky: I think this is fixed now so it works like ADXL345

klippy/extras/ads1220.py
garethky: Removed

src/sensor_ads1220.c
garethky: Removed

src/sensor_ads1220.c
garethky: Resolved to work like ADXL345

KevinOConnor:
Thanks.  I'm not sure what the "sensor reset" stuff is.

There were a bunch of stuff discussed above, but putting aside various code commentary, I'd say only the following are "blockers":
1. This PR should be rebased onto the latest Klipper master (which contains the bulksensor.py FixedFreqReader refactor).
2. The mcu bulk sensor code should not call `shutdown()`.  It should check for errors and propagate those errors to the host code.
3. Don't modify `setup_query_command()` in bulk_sensor.py.
4. The mcu "pending flag" code didn't look correct to me.  Either I'm missing something or the code should be corrected.  (This may already be done.)
5. The code should be split into 2 commits, but I can do that if it's a pain for you to do that.

-Kevin

garethky:
> I'm not sure what the "sensor reset" stuff is.
Well we need to agree on this so I know what to do.

Looking at the accelerometer code, it seems to ignore the errors. I don't see where `AccelCommandHelper` or `ResonanceTester` actually check if the data contains `errors`. This "carry on" approach can't work for load cells.

Here is a primer on the problem:

These chips fail at runtime due to ESD. The big metal load cell acts like a ground path and they are very susceptible to static. This causes the chip to reboot, erasing any settings it had from klippy startup. Continuing to sample after an error wont recover the sensor. Even if it does produce data it is going to be the wrong channel/sample rate/gain etc. So once the chip resets the data stream is dead.

Some chips in this class send a reboot flag with their measurements but these chips are too cheap for that. So all of these checks are heuristics to detect a reset. Even taking too long to collect a sample can change the settings on the HX71x so the next sample is garbage.

When an error is detected we have to stop sampling, re-program the chip and restart sampling. I avoided doing this because its complex. and when installed in a properly grounded printer I don't see these errors. But I'm sure at a large scale they will happen at some non zero rate.

I agree with you that sending the errors to the Host is the right thing to do. But if all the Host does it shutdown on error its a marginal improvement. Ideally I'd like to try and recover so that a stray ESD doesn't ruin a print.

KevinOConnor:
Thanks.  I can only give high-level comments as I'm not familiar with the details of these sensors - so take my comments with "a grain of salt".

> I don't see where AccelCommandHelper or ResonanceTester actually check if the data contains errors

I was referring to how `src/sensor_adxl345.c:adxl_query()` checks for a read error and populates a response of `0xff 0xff 0xff 0xff 0xff`, and how `klippy/extras/adxl345.py:_convert_samples()` looks for that code and drops the sample (and increments the `_process_batch()` error counter).  Also see how `klippy/extras/ldc1612.py:_convert_samples()` checks for warning flags in the top 4 bits of the sample, increments the "errors" counter, but continues to process the data.

> When an error is detected we have to stop sampling, re-program the chip and restart sampling. I avoided doing this because its complex. and when installed in a properly grounded printer I don't see these errors.

For what it is worth, as a user, I'd not want any automatic error recovery like that.  To me, restarting the sensor sounds like it is very likely to lead to overall poor measurement results.  It may be possible to restart the sensor, but there is a very high chance the resulting values will not be of high quality - the nozzle/bed contact may have been missed or distorted, the timing correlation is likely off, etc.  So, it may be possible to restart the sensor, but there is a very good chance I'll get a terrible first level layer height.  As a users, I'd prefer the action to be stopped and to receive an error that says "your hardware is malfunctioning and needs to be fixed", so that I can then fix the hardware and go on to get great first layer quality.

As a developer, I would also prefer not to maintain complex code that, to wit, "makes it easy for users to get bad results".

So, my concern is not with reporting of errors, it is with using `shutdown()` to convey all errors.  Shutdown is useful for internal errors (to alert developers of a coding error), for safety issues (heater outside temperature range), and for situations where the mcu is hopelessly unable to perform the requested actions (eg, very far behind on processing of scheduled timed actions).  For a malfunctioning sensor, we ideally want to stop the action (eg, "sorry we couldn't perform that probe because the load-cell sensor is reporting a ground fault"), but ideally we don't want to completely turn off the motors/heaters with an arcane diagnostic message (eg, "ADS1220: Possible bad read").

Again, I can't give definitive advice on how to handle errors for these sensors.  It seems to me though, that if it's a questionable sensor reading we would want to mark it as a questionable reading and let the high-level code determine what action to take.  If the sensor gets in a state where it can't produce any readings, then we can signal that as well (so that the high-level code knows it wont get further readings and can raise an appropriate error message).

For this initial PR of sensor support, with a goal of getting sensor data to the api server, a `shutdown()` on questionable data seems a drastic response.

Thanks again.  Hope that helps.
-Kevin

garethky:
Thanks, I think I can work with that guidance.

> As a users, I'd prefer the action to be stopped and to receive an error that says "your hardware is malfunctioning and needs to be fixed", so that I can then fix the hardware and go on to get great first layer quality.

This is going to happen via the LoadCellEndstop. When the sensor stops reporting data during a homing event, its a critical failure. The LoadCellEndstop has a watchdog timer and if no data is received for 2 sample windows it will issue a `SHUTDOWN`. (not in this PR)

I think an axis moving without a functioning endstop counts as a `SHUTDOWN` "for safety issues", this is just like a thermistor becoming unplugged.

We don't have to shutdown in the sensor code to achieve that, so they will be removed from this PR.

> As a developer, I would also prefer not to maintain complex code that, to wit, "makes it easy for users to get bad results".

I also don't want to implement this due to the complexity. (Users should ground their printers!!)

The next best thing is to just stop measurements when the Host gets an error from the sensor. So inside the sensor it will not re-schedule the sampling timer on error. On the Host, I can flush the error to clients and then run the finish measurements logic which will disconnect all of the consumers from `bulk_sensor` and run the shutdown logic on the MCU sensor. (I don't think `bulk_sensor` informs consumers that they were disconnected, but in any case they will get the error first)

For something non-critical like a Filament scale it should be possible to restart measurements without halting the print. Or just let the filament scale be broken until the Host restarts. Ideally a non-critical use shouldn't ruin a print.

KevinOConnor:
> I think an axis moving without a functioning endstop counts as a SHUTDOWN "for safety issues", this is just like a thermistor becoming unplugged.

A bit off topic here, but the best way to rapidly stop a homing/probing operation is to issue a `trsync_do_trigger()` with an error code as the "reason".  It's faster than `shutdown()` for multi-mcu homing (as it is propagated between mcus without entering the Python code), and it also allows for better error reporting.

-Kevin

garethky:
I got the C part in, shutdowns are gone. When an error happens the sensor just turns off its pending flag and flushes the error. That will be the last packet it sends before its restarted (if its ever restarted).

I used the `possible_overflows` fields to send back an error code, since that free payload that these sensors don't use.

I have to do the re-base to mainline before I can tackle the Python side

KevinOConnor:
> How are we deciding what to cut from these space limited chips?

There's no documented strategy - generally we disable whatever new feature caused the build to go over in size.  It's really useful to do build all the targets that we have config files for - it catches lots of inadvertent errors.  There isn't a need for these smaller chips to support every possible option though.

-Kevin

cab404:
rn both revisions fail on imports :(

should I start worrying and trying to finish this pr, or am I missing something obvious?)

garethky:
> rn both revisions fail on imports :(
> 
> should I start worrying and trying to finish this pr, or am I missing something obvious?)
Sorry, is this comment related to this PR (Bulk ADC Sensors)?

cab404:
> > rn both revisions fail on imports :(
> > 
> > should I start worrying and trying to finish this pr, or am I missing something obvious?)
> Sorry, is this comment related to this PR (Bulk ADC Sensors)?
> 

Yes, and I mean both commits in the branch

garethky:
> Yes, and I mean both commits in the branch

I just need to get some time to work on it. Looks like I have to exclude the hx71x sensor from all the places that would support it but don't have enough space left in their image to do so. I don't have the local tooling on my Mac setup to test this so I'm doing commits and having github do the testing.

cab404:
rn both revisions fail on imports :(

should I start worrying and trying to finish this pr, or am I missing something obvious?)


> > Yes, and I mean both commits in the branch
> 
> I just need to get some time to work on it. Looks like I have to exclude the hx71x sensor from all the places that would support it but don't have enough space left in their image to do so. I don't have the local tooling on my Mac setup to test this so I'm doing commits and having github do the testing.

I don't want to rush you, sorry if that sounded like it — I just wanted to ask if any help is needed. 

Thank you for working on this!

garethky:
> I don't want to rush you, sorry if that sounded like it — I just wanted to ask if any help is needed.
> 
> Thank you for working on this!

I don't have a way to compile and run the github test suite on my Mac. I looked at some setups but couldn't make it work.

nefelim4ag:
I don't know the actual rom size limits for boards.
But just as a "crazy" suggestion, maybe in the future.
Technically, ROM can store `klipper.dict` in any format.
Json compressed is simplest of course.

For now, the simplest way to squish more is to use a different compression algorithm or use dictionary compression.
Just for reference:
```
$ du -bs klipper.dict*      
7741    klipper.dict # size on master branch
2442    klipper.dict.br # google brotli (uses text dictionary internally)
2839    klipper.dict.gz # zlib compression -9 
2986    klipper.dict.zst # Zstd default
1148    klipper.zstd.dict # Zstd with dictionary trained over current klipper.dict
```

garethky:
stm32f401 - "Too many message ids"

So this PR can't be merged until klipper gets 2 bytes message IDs (or these sensors are excluded from all tests, which seems wrong)

I'll keep working on the other outstanding issues

cab404:
> > I don't want to rush you, sorry if that sounded like it — I just wanted to ask if any help is needed.
> > 
> > Thank you for working on this!
> 
> I don't have a way to compile and run the github test suite on my Mac. I looked at some setups but couldn't make it work.
> 

What Mac do you have?

garethky:
Last word I know about this from Kevin: https://klipper.discourse.group/t/scalability-of-too-many-message-ids/8023/4

"Using a VLQ for the command id was just never implemented in the mcu. Implementing it is likely the best long-term scalable solution."

cab404:
> Intel

Hm, and this tool doesn't work? https://github.com/nektos/act

garethky:
I made a change to the HX71x code to disable interrupts for the entire read cycle. This should last about ~15us. I made the change because of 2 errors that I saw happen while printing:

* "Time too close" while bed meshing. These would usually happen in the first couple of probe points. Multiple people have reported this.
* ERROR_READY_AFTER_READ happening randomly while printing, indicating that the read was delayed long enough for a new sample to arrive.

Combined with seeing `ERROR_READ_TOO_LONG` when the `MAX_READ_TIME` was set to 200us, its pretty clear that the code is being interrupted for a long time. This seems to happen when the serial queue sends instructions to the step queue. The code change seem to resolve the issue but I'm concerned about the increased time with interrupts turned off.

I was trying to follow what neopixel does and only disable interrupts for a minimum of time. So whats reasonable to turn interrupts off for?

garethky:
I think I have a good solution now:
* I'm measuring the start and end times outside of taking the sample and assuming they include IRQ interruptions that extend the execution time. So I've limited the execution time to 1/2 the sample period. This seems to be "fair" as no single module should be taking up 50% of the wall clock time to get executed.
* I moved the check for the data ready pin inside an IRQ fence so its checked at the exact time we expect the chip to turn off. Checking it any later time is fragile as IRQ delay may cause it to be sampled after it comes back on.
* I've gone back to the delay between clock pulses using `irq_poll()`. This tries to limit the time the IRQs are turned off to the absolute minimum. This is what neopixel and prusa_buddy are doing and was the original intent.

I haven't see the "Time too close" error come back but I've only had 2 days to test this.

I've implemented rudimentary sensor resets when an error is reported. They are run from the host side. The sensors stop scheduling themselves when they report an error and flush whatever readings they have to the host. The host also now logs an error when it sees one of the error codes from the sensor. Later the load_cell can cause a shutdown if it sees an error in the middle of a probe. I haven't tested the reset stuff because I cant really cause it at will, but I will get to that soon, probably with a temporary count based trigger in the sensor code.

And I've split the code up into 2 commits. One for the basic load_cell skeleton and hx71x and a second adding the ADS1220.

garethky:
> > Intel
> 
> Hm, and this tool doesn't work? https://github.com/nektos/act

I tries and it didn't work, build ended with an error related to building the doc site. I didn't have time to do any debugging.

garethky:
I merged in the patch from Frederic Lauzon <frederic.lauzon@aon3d.com> (I think [frlauzon](https://github.com/frlauzon) here on github) that makes `msg_id` a VLQ. I need this to build my code now that we have run out of single byte message ids.

garethky:
2-byte message ID support is on the way from @KevinOConnor: https://github.com/Klipper3d/klipper/pull/6613
Will re-base once this is merged.

KevinOConnor:
Hi.

I have merged in the support for 2-byte ids (PR #6613).

I also reviewed this PR again.

I did not see anything that would likely lead to "timer too close" errors.  It's possible there was something in your previous versions that you've now fixed (or, something more subtle that's not obvious to me after reviewing the code).  I did see some opportunities for optimization though.

The PR mostly looks fine, but I did notice a couple of things.

Mainly, I don't think the proposed error handling will work well.  There are two issues: 1) if partial sensor_bulk_data messages are sent it will mess up the sequence number and timestamp tracking in bulk_sensor.py, and 2) messages sent from mcu to host are not guaranteed to be delivered (so, the host code may not detect the error report as the `if not samples:` test may always be taken).  Also, I'm unsure about using `possible_overflows` for error reporting as that could be confusing.

Other minor things I noticed:
* the Config_Reference.md section headers don't look like they are properly nested.
* The `duty_cycle` config option isn't documented.  It looks like it should be a code constant (instead of a config option).
* The code should implement a `_convert_samples()` method - otherwise error reports will pass through as bogus data to clients of the api server.
* The `get_range()` call looks funky to me.  It returns `-(1<<24)..(1<<24)` but shouldn't it return `-(1<23)..((1<<23)-1)`?  Maybe add this method when adding the code that uses this method.
* The ads1220 code uses a config variable `sps` but the documentation says `sample_rate`.
* FYI, the `config.getchoice()` code supports integer keys, and it's preferable to use integer keys when applicable. (I noticed this in the ads1220 code.)

I'll try to put together a code example of some optimizations and an alternative error reporting mechanism.

Thanks,
-Kevin

garethky:
Ok, I can change the error reporting scheme:
• detect an error
• put sensor to sleep, store sleep state in a flag
• continue to run the polling loop
• check the flag at the start of the read and, if set, log an error to the output.

That way eventually an error packet will reach the host.

I should still be able to use the `possible_overflows` field to send the original error type.

Good points on the other items, I will get those addressed.

KevinOConnor:
I put up a branch with sample hx71x code at https://github.com/KevinOConnor/klipper-dev/tree/work-hx71x-20240611

This is untested - I don't have the hx71x hardware.  The goal is to mock up a possible alternative error handling implementation.  The branch also has some proposed code optimizations.

Some notes on the code:
* The timer handler reschedules itself.  The rest_ticks is set to half the query period.  The task is only woken if data is available.  If the task doesn't complete by the time the next timer runs, the `possible_overflows` counter is increased.
* All the low-level bit-banging is moved to a new `hx71x_raw_read()` function.
* The host detects if the sensor stops sending data (and is repeatedly incrementing `possible_overflows`) - it resets the sensor in that case.
* There is a simple `_convert_samples()` function that increments a user facing error count if the sensor reports an error.
* Minor changes to docs/Config_Reference.md for nesting.
* No need to start sensor on load_cell.py startup.

Thoughts?
-Kevin

garethky:
I get what you were going for with a single read loop, removing the conditional blocks etc. But it cant come at the expense of the safety check. `raw_read` needs that check to see if the data ready pin was reset on the 25th clock pulse. If its not, you don't know which clock pulse the chip thinks its on and must reset the chip.

In testing these chips we have had a horrible time with ESD. It can cause the clock to get pulsed or the chip to reboot. When that happens the sync between the code and the chip state is lost and you no longer know which clock pulse the chip is really on. The only thing to do is reset the chip. Testers found the logging really valuable to get their wiring/grounding sorted out.

Let me see if I can incorporate some of this and keep the error checks and reporting.

What is your thinking behind the `barrier()` call?

KevinOConnor:
> But it cant come at the expense of the safety check. raw_read needs that check to see if the data ready pin was reset on the 25th clock pulse.

Agreed.  But, I think the check is still in place - see `hx71x_read_adc()`:
```c
    // Check for errors
    uint_fast8_t extras_mask = (1<<gain_channel) - 1;
    if ((adc & extras_mask) != extras_mask)
        // Transfer did not complete correctly.
        counts = SAMPLE_ERROR;
```
This snippet should check that all the "gain bits" report a logical 1.  (As I read the spec, not only should the 25th be high, but the 26th-28th need to be too as otherwise the read pulses would have counted towards the next sample which would mess up the bit position of the next read.)

Let me know if I missed something.

> What is your thinking behind the barrier() call?

The idea is to tell gcc not to reorder the `hx71x->pending_flag = 0;` statement.  That statement turns off the `hx71x->sb.possible_overflows++` in `hx71x_event()`.  That is, we want to acknowledge that the read was completed immediately after all the "read pulses", and we want to note a possible overflow should half a sample time pass prior to that point.

Cheers,
-Kevin

KevinOConnor:
It occurred to me that we can make the timing more aggressive once the timer handler reschedules itself.  See this incremental patch: https://github.com/KevinOConnor/klipper-dev/commit/3b673b6ed0ecdfbba8b28a195c894aa99f114a3b

The idea here is to check for a new sample 11 times per nominal sample period.  Once a new sample is available, we can give the task 8/11ths of a nominal sample period to completely pull that sample.  (If the task can't get the sample out by then it would be a possible overflow.)

I think this could be more robust than the current proposed code (which samples at twice the nominal rate and gives only half a nominal period to pull that sample).

Cheers,
-Kevin

garethky:
I had something like that in a prior version, before the linear regression time code, to try and narrow the error. But if thats a good idea, why wouldn't we want to do the same timing trick for all other sensors that don't have a buffer?

KevinOConnor:
> But if thats a good idea, why wouldn't we want to do the same timing trick for all other sensors that don't have a buffer?

Most of the existing sensors don't have a fast gpio based check to see a sample is pending.  The only way to check these sensors is to send a spi (or i2c) command which has to be done from task context.  Increasing the number of task wakeups and spi/i2c transactions would increase the mcu load.  If we have a "data pending gpio" though, it's low-overhead to check that gpio in the timer at higher frequencies.

That said, the ldc1612 can be configured with an "irq line" and it might be worthwhile to make it more aggressive.  (FYI, I based part of my code proposal here on my results performing the "gpio check in timer handler" as found in the `sensor_ldc1612.c` code.)

Cheers,
-Kevin

garethky:
I have confirmed that your changes work with the real sensor, at least at idle. I don't have a full stack of changes build up on top of this to prove that is probes. I'd really like to do that just so I can run some prints with it over the weekend and verify that it wont throw any errors over a more extended period when the MCU is under higher load.

KevinOConnor:
Okay, thanks.

I had a couple of additional ideas you may want to consider.  I added a few commits to the branch at https://github.com/KevinOConnor/klipper-dev/tree/work-hx71x-20240611 .  Briefly:
1. If we do want to use dynamic timing, then I think we could probably do 1/10th of a nominal period for standard retries, 8/10ths of a period after sample detection, and 4/10ths of a period if an overflow may have occurred.
2. We could use 0x40000000 as an error indicator (instead of 0x80000000) as that should still be unique and it avoids confusing issues with the sign bit.
3. It would be nice if we exported the adc values over the API Server as floating point numbers (instead of counts).  I understand you need counts in other parts of the code, but it should be trivial to convert back to a count from a float (`int(adc * (1<<23) + .5)`).

Just ideas to consider.

Cheers,
-Kevin

garethky:
#1: I'm scared this is getting complicated

The place this would help is when homing, it would make the trigger delay as short as possible, reducing overshoot. 1/10th the sample period is ~300us at 320SPS and 1250us at 80SPS. But the pin checks for both sample rates take the same amount of time. What may be optimal is to set a fixed interval in microseconds for pin checks for all sample rates, (e.g. 100us). Then, say you want to do 3 pin checks, on average, before seeing a signal.
```
#define PIN_CHECK_INTERVAL timer_from_us(100)
#define PIN_CHECK_WINDOW PIN_CHECK_INTERVAL * 3
rest_ticks_nominal = (hx71x->rest_ticks * 10) - PIN_CHECK_WINDOW;
rest_ticks_retry = PIN_CHECK_INTERVAL;
rest_ticks_overflow = hx71x->rest_ticks * 4;
```
On the ADS1220 it can do 2000 SPS. 1/10th of its interval is 50us, so maybe the PIN_CHECK_INTERVAL needs to be shorter than I put here.

I also though about making the PIN_CHECK_WINDOW adaptive. If the sample is ready on check 1 then the window should have been larger. If its ready on sample 5 then it should get smaller... but that is getting complicated. Lower MCU load and faster sample rates are much better answers to these problems. 

#2 OK with me!

In my head I though that I had implemented only sending back 24 bits per sample. But given that its 32bits (and you are OK with that) we can totally use the high bits for error codes.

#3 Would you be OK with just sending both?

`samples[count] = (round(ptime, 6), round(val * adc_factor, 9), val)`
The conversion is sensor dependent (the next sensor doesn't have to be 24 bits, could be 16, 32 etc.). I didn't want sensor specific code in the load_cell class. And my concerns with the float value are:

The floating point value doesn't properly convey the saturation point of the sensor :
```
>>> adc_factor = 1. / (1<<23)
>>> max = 0x7FFFFF # max positive value
>>> round(max * adc_factor, 9)
0.999999881
```
So checking for `val >=1.0 or val <= -1.0` doesn't work. Deciding how much "wiggle room" to use to detect saturation would be sensor dependent.

Secondly rounding discards data, e.g. 9 digits isn't enough to represent the precision of a 32 bit sensor:
```
>>> adc_factor = 1. / (1<<31)
>>> round(1 * adc_factor, 9)
0.0
```

KevinOConnor:
1 - I was thinking of this:

```c
static uint_fast8_t
hx71x_event(struct timer *timer)
{
    struct hx71x_adc *hx71x = container_of(timer, struct hx71x_adc, timer);
    uint32_t rest_ticks = hx71x->rest_ticks;
    if (hx71x->pending_flag) {
        hx71x->sb.possible_overflows++;
        rest_ticks *= 4;
    } else if (hx71x_is_data_ready(hx71x)) {
        // New sample pending
        hx71x->pending_flag = 1;
        sched_wake_task(&wake_hx71x);
        rest_ticks *= 8;
    }
    hx71x->timer.waketime += rest_ticks;
    return SF_RESCHEDULE;
}
```
Which I don't think is overly complex.  I don't think we need to worry about responsiveness as we'll already be delayed by several milliseconds due to filtering (and worst case multi-mcu homing situations).  So, no point in optimizing for microseconds.  I also think it is fine to hardcode the x4 and x8 in the mcu - the host and mcu code are tightly coupled anyway.

If you think the above is too obscure or too complex then I think the simple unchanging `rest_ticks=.5 * nominal_period` is fine.

2 - okay.

3 - I'm fine with it any way - counts, float, or both.  It's not a big concern.  FWIW, I don't know of any adc sensor that's more accurate than about 15bits in practice, so I don't think precision really matters.  (Sensors claiming more accuracy are just using creative marketing as far as I'm aware.)

Cheers,
-Kevin

garethky:
I got a change stack built up today. I got the machine to home but not probe. I need to work through all the refactoring that have happened in `probe.py`, looking at `probe_eddy_current.py` for guidance. Experimenting with [jj](https://github.com/martinvonz/jj/tree/main) to make the commit stacking easier, so far its better than Git.

garethky:
I've had a chance to try artificially breaking the sensor and forcing the host to restart it and this worked just fine. It goes into the error state on the MCU and just keeps broadcasting errors until it gets reset. This dovetails with the later [endstop implementation](https://github.com/Klipper3d/klipper/commit/c92d84122fdba1aa5e2b03bf9ee3a6e1e83a94f4#diff-843fe921a58492d834879197d17af6a18263ec860909c70a9552dd92296897f5R175) where it wont pass any data to the endstop once it enters the error state. I still need to do this testing on the  ADS1220.

Once the HX71x sensors encounter a detectable error I don't think we should trust its results until its reset and thats what is implemented. If there is a desync the sensor can produce results that are from a different channel than configured and that could get passed to the endstop as valid data to prolong a homing move and damage the machine. On the ADS1220 I've seen it just produce garbage bits when reading the settings registers so I don't trust its output either.

I have a series of branches that build up the upcoming load cell PR sequence ([pr-load-cell-gram-scale](https://github.com/garethky/klipper/tree/pr-load-cell-gram-scale), [pr-load-cell-endstop](https://github.com/garethky/klipper/tree/pr-load-cell-endstop), [pr-load-cell-endstop](https://github.com/garethky/klipper/tree/pr-load-cell-endstop), [pr-probe-load-cell]([pr-probe-load-cell](https://github.com/garethky/klipper/tree/pr-probe-load-cell))) and I have test machine printing again on top of the last one.

garethky:
The latest update has a small change to not return an empty report when an overflow is encountered and instead return whatever data was reported.

I've created a test branch that can probe with the sensor set to fail every n seconds: [test-sensor-errors](https://github.com/Klipper3d/klipper/commit/3a6d037bc11fb1a382e2f102b94853a4be368bf9). I have the load cell endstop set up to trigger the trsync with an error code now instead of a shutdown.

There are several failure modes and the errors they generates are not always clearly of the form "your sensor had an error". Probably the most insidious one happens when the sensor resets and then homing starts. This causes the tearing code to get no data for the first bit of the tear, resulting in not enough samples being collected before it times out. But since the reset already happened the tear process doesn't get any errors to explain the root cause. I need to consider the sensor "offline" if it sends out an error packet and only consider it back "online" when the next non-error packet is sent.

KevinOConnor:
Thanks.  What's the status of this PR?  Is it ready for merging?

I have a few questions/comments about the latest implementation.  These can be adjusted post-merge if changes are needed.
* It's not a good idea to call `logging.error()` in `_convert_samples()` as doing so could flood the log (possibly escalating a minor error into a severe error).
* Is it necessary for the host to "restart" measurements on an error?  Seems like things would be simpler if the host raised an error and stopped measurements on an error (the data is suspect after that point anyway).
* Why does the ads1220 check for `ERROR_READ_TOO_LONG`?  I understand why hx71x does (a bad read could mess up the transfer protocol), but I don't see what would go wrong on the ads1220.  Seems it would just make a minor error (a lost sample) into a notable error (extended sensor outage).
* I don't think sampling should be automatically started on klipper startup.

Thanks again,
-Kevin

KevinOConnor:
Also, just "thinking out loud" - have you considered halting any ongoing homing on a sensor error, while still allowing data reports to continue?  (That is, continue to populate the bulk sensor reports, but immediately notify the homing code that it should abort.)

-Kevin

garethky:
* _have you considered halting any ongoing homing on a sensor error, while still allowing data reports to continue? (That is, continue to populate the bulk sensor reports, but immediately notify the homing code that it should abort.)_
That's the way the first version worked. I decided this was more complex, requiring an extra call to report errors. And if there is some bug that doesn't report errors (as might happen in development) then it would just crash the toolhead. So I ended up with just the watchdog timer on the MCU.

* _I don't think sampling should be automatically started on klipper startup._
Please explain what you have in mind.

My idea is that this works like a temperature sensor. They all start when klipper starts. The `load_cell` object reports its recent gram weight as part of its status. For the folks that want to use this to measure the weight of something during a print (like a spool) this is the behavior they expect.

KevinOConnor:
> > I don't think sampling should be automatically started on klipper startup.

> Please explain what you have in mind.

This PR only exports the loadcell information via the API Server.  If a client connects to Klipper and requests loadcell information then sampling is automatically started at that point.  Thus, there is no reason to start sampling at Klipper startup.

I expected sampling to occur when there are users of the data, and sampling to be disabled when there are no users.  See klippy/extras/ldc1612.py as an example.

Cheers,
-Kevin

garethky:
> I expected sampling to occur when there are users of the data, and sampling to be disabled when there are no users. See klippy/extras/ldc1612.py as an example.
`hx71x` and `ads1220` fulfill that contract as you desire, they are comparable to `ldc1612`. `load_cell` is a different kind of object, its comparable to `temperature_sensor` or `heaters`. If this isn't obvious I suggest that we remove `load_cell` from this PR and just focus on the sensor code. `load_cell` is the whole focus of the next PR in the series anyway.

I'll look at just having the sensor stop sampling and having `load_cell` detect that and trigger a sensor restart.

Does that sound like a good path forward?

KevinOConnor:
Thanks.  My main question is - do you feel this PR is ready for merging?  If so, I think we'd be better off merging as is and then look to make any remaining changes on top of it.

> If this isn't obvious I suggest that we remove load_cell from this PR and just focus on the sensor code.

Well, I'd prefer to merge this PR before going on to the main loadcell work.  I'd also prefer for PRs to provide some user measurable benefit, and providing the ads1220/hx71x measurements via the api server (as the current stubbed out loadcell.py code does) would meet that minimum goal.

> I'll look at just having the sensor stop sampling and having load_cell detect that and trigger a sensor restart.

I'm not sure I understand this, but FWIW it sounds complex.

Cheers,
-Kevin

garethky:
> Well, I'd prefer to merge this PR before going on to the main loadcell work. I'd also prefer for PRs to provide some user measurable benefit, and providing the ads1220/hx71x measurements via the api server (as the current stubbed out loadcell.py code does) would meet that minimum goal.

Great! Lets do that.

> Is it necessary for the host to "restart" measurements on an error? Seems like things would be simpler if the host raised an error and stopped measurements on an error (the data is suspect after that point anyway).

>> I'll look at just having the sensor stop sampling and having load_cell detect that and trigger a sensor restart.

For this PR I mean to just have the sensors stop sampling on error and terminate all clients. Easy enough.

I'll get your other concerns (logging and the `ERROR_READ_TOO_LONG` check) addressed as well.

KevinOConnor:
Thanks.  I did not fully understand your last response.  Would you prefer that I merge what is here now, or would you prefer that I wait for additional changes?

-Kevin

garethky:
> Thanks. I did not fully understand your last response. Would you prefer that I merge what is here now, or would you prefer that I wait for additional changes?
> 
> -Kevin

I am working on this today. Will post changes by end of day.

garethky:
I'm good to merge.

I looked at raising an error in the hx71x code but I left it as-is. When an exception is raised the Batch Bulk Helper drops all the clients but doesn't notify them they were dropped. I think its better to notify clients with errors and make them responsible for deciding if the sensor is stable enough to use.

* Cleaned up the error logging in hx71x.
* `consecutive_fails` is reset to 0 when a payload contains no errors
* Payload with errors from hx71x only has the first error reported, ignoring the duplicates.
* Removed the time check around the SPI call in the ADS1220 and deleted the associated error handling code.
* Cumulative errors and overflows are reported, to match what other sensors are doing.

EllaFoxo:
Just wanna say @garethky - Thank you! I've been following your work for absolutely ages. Hardware to follow 😄

PR #6556, opened 2024-04-08
Add motor resonance reduction to shapers.

andrewmcgr:
This adds an MZV shaper with configurable parameters to each axis.

### DIFF
--- a/klippy/chelper/kin_shaper.c
+++ b/klippy/chelper/kin_shaper.c
@@ -22,7 +22,7 @@ struct shaper_pulses {
     int num_pulses;
     struct {
         double t, a;
-    } pulses[5];
+    } pulses[15];
 };
 
 // Shift pulses around 'mid-point' t=0 so that the input shaper is an identity
--- a/klippy/extras/input_shaper.py
+++ b/klippy/extras/input_shaper.py
@@ -21,6 +21,10 @@ class InputShaperParams:
                                              shaper_defs.DEFAULT_DAMPING_RATIO,
                                              minval=0., maxval=1.)
         self.shaper_freq = config.getfloat('shaper_freq_' + axis, 0., minval=0.)
+        self.motor_damping = config.getfloat('motor_damping_' + axis,
+                                             shaper_defs.DEFAULT_DAMPING_RATIO,
+                                             minval=0., maxval=1.)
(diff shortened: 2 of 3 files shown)


Zeanon:
Could you elaborate a little on how one would calibrate those params?
I was already pretty intrigued when frix posted the initial pr to increase the amount of shaper pulses so I would love to try that out

andrewmcgr:
You'd, for example, use a recent version of [K-ShakeTune](https://github.com/Frix-x/klippain-shaketune) to find your motor resonance and damping factor for each axis (they'll be the same for, e.g., corexy).

dmbutyugin:
This can be an interesting feature indeed. And code-wise, it is rather straightforward. However, it would be great to see some quantitative effect of this feature. Which could be, for instance, measuring the resonance frequency of the steppers on a few printers, showing the value of those frequencies and if/how these resonances affect the printing (e.g. artifacts on the model, audible noise (at accelerations/decelerations?), etc.), and what does this feature do to it. Perhaps Klipper discourse could be a great starting point for these discussions.

Frix-x:
@dmbutyugin just for reference, https://github.com/Klipper3d/klipper/pull/6460 was done in order to implement this. Andrew and I are discussing a lot together and with other Voron contributors about this on Discord. There are already a few people experimenting with it.

It's pretty hard to quantify the effects directly, because we had different results depending on the machines:
- Using accelerometer measurements, it's pretty easy to measure the resonance frequency of the motors using [this Shake&Tune macro](https://github.com/Frix-x/klippain-shaketune/blob/main/docs/macros/vibrations_profile.md) (bottom right graph). On some machines, applying the motor resonance reduction lowered this resonance peak very much (like by ~half). On some other machines, the effect was more anecdotal. But it also helped reduce the amount of global vibration (the other graphs).
- Apparently it's more effective at 36 or 48V than 24V. But we don't have enough feedback and data on that, so it could be a coincidence for now.
- On all machines, the motor noise is changed to something that sounds less aggressive to the ear, smoother and in the higher spectrum. But this is not quantified, it's just a feeling and not really scientific. A few weeks ago I bought some piezoelectric microphones to attach to the motors to record their internal noise: this should give us more information than accelerometers, which have a lower cut-off frequency.

PR #6537, opened 2024-03-21
probe_eddy_current: implement surface scannng

Arksine:
This PR build on #6536, adding support for "surface scanning".  The first commit resolves a minor issue where `PROBE_EDDY_CURRENT_CALIBRATE` removes all previous autosaved configuration, including the `reg_drive_current` option saved by `LDC_CALIBRATE`.  If necessary this commit can be removed from this PR and added to another (or to the original).

The rest of this PR makes the necessary modifications to `probe_eddy_current.py`, `probe.py`, and `bed_mesh.py` to support scanning a surface using an eddy probe.  In addition, a `graph-mesh.py` script has been added to visualize and analyze mesh state, which can be dumped to a json file using the new `BED_MESH_DUMP` command.

PR #6558, opened 2024-04-09
probe_eddy_current: implement surface scannng 

Arksine:
This is a re-creation of PR #6537, rebased against master now that the ldc work has been merged.  The first pull request was against the work-ldc branch, so the PR was closed when the branch was deleted.

### DIFF
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -421,12 +421,95 @@ have undesirable results when attempting print moves **outside** of the probed a
 full bed mesh has a variance greater than 1 layer height, caution must be taken when using
 adaptive bed meshes and attempting print moves outside of the meshed area.
 
+## Surface Scans
+
+Some probes, such as the [Eddy Current Probe](./Eddy_Probe.md), are capable of
+"scanning" the surface of the bed.  That is, these probes can sample a mesh
+without lifting the tool between samples.  To activate scanning mode, the
+`METHOD=scan` probe parameter should be passed in the `BED_MESH_CALIBRATE`
+gcode command.
+
+To accommodate these probes the following additional `probe_parameters` are
+available to `BED_MESH_CALIBRATE`:
+
+- `SCAN_MODE=[detailed | rapid]`:  Choses the scan mode.  The `detailed` mode
+  will pause and collect samples at each probe point.  The `rapid` mode will
+  travel on a continuous path with no pauses, collecting samples near each probe
+  point.
+- `SCAN_SPEED=[speed] `: The maximum X/Y travel velocity of the tool when
+  performing a scan.  The default is the value of the `speed` option in the
+  configuration.
+- `SAMPLE_TIME=[time]`:  The time, in seconds, the tool pauses for sample
+  collection in `detailed` scan mode.  The default is .1 seconds.
+- `SAMPLES_RESULT=[option]`: The type of averaging to perform on collected
+  samples.  Available options are:
+  - `standard`: All collected samples are averaged.
+  - `centered`: Samples are sorted by value.  The first and last quarters
+    are discarded and the remaining samples are averaged.
+  - `weighted`: Samples closer to the desired probe location are assigned
+    more weight in the average than samples farther from the location.
+
+### Scan Height
+
+The scan height is set by the `horizontal_move_z` option in `[bed_mesh]`.  In
+addition it can be supplied with the `BED_MESH_CALIBRATE` gcode command via the
+`HORIZONTAL_MOVE_Z` parameter.
+
+The scan height must be sufficiently low to avoid scanning errors.  Typically
+a height of 2mm (ie: `HORIZONTAL_MOVE_Z=2`) should work well, presuming that the
+probe is mounted correctly.
+
+It should be noted that if the probe is more than 4mm above the surface then the
+results will be invalid.  Thus, scanning is not possible on beds with severe
+surface deviation or beds with extreme tilt that hasn't been corrected.
+
+### Rapid (Continuous) Scanning
+
+When performing a `rapid` scan one should keep in mind that the results will
+have some amount of error.  This error should be low enough to be useful on
+large print areas with reasonably thick layer heights.  Some probes may be
+more prone to error than others.
+
+It is not recommended that rapid mode be used to scan a "dense" mesh.  Some of
+the error introduced during a rapid scan may be gaussian noise from the sensor,
+and a dense mesh will reflect this noise (ie: there will be peaks and valleys).
+
+Bed Mesh will attempt to optimize the travel path to provide the best possible
+result based on the the configuration.  This includes avoiding faulty regions
+when collecting samples and "overshooting" the mesh when changing direction.
+This overshoot improves sampling at the edges of a mesh, however it requires
+that the mesh be configured in a way that allows the tool to travel outside
+of the mesh.
+
+```
+[bed_mesh]
+speed: 120
+horizontal_move_z: 5
+mesh_min: 35, 6
+mesh_max: 240, 198
+probe_count: 5
+scan_overshoot: 8
+```
+
+- `scan_overshoot`
+  _Default Value: 0 (disabled)_\
+  The maximum amount of travel (in mm) available outside of the mesh.
+  For rectangular beds this applies to travel on the X axis, and for round beds
+  it applies to the entire radius.  The tool must be able to travel the amount
+  specified outside of the mesh.  This value is used to optimize the travel
+  path when performing a "rapid scan".  The minimum value that may be specified
+  is 1.  The default is no overshoot.
+
+If no scan overshoot is configured then travel path optimization will not
+be applied to changes in direction.
+
 ## Bed Mesh Gcodes
 
 ### Calibration
 
-`BED_MESH_CALIBRATE PROFILE=<name> METHOD=[manual | automatic] [<probe_parameter>=<value>]
- [<mesh_parameter>=<value>] [ADAPTIVE=[0|1] [ADAPTIVE_MARGIN=<value>]`\
+`BED_MESH_CALIBRATE PROFILE=<name> METHOD=[manual | automatic | scan] \
+[<probe_parameter>=<value>] [<mesh_parameter>=<value>] [ADAPTIVE=[0|1] \
+[ADAPTIVE_MARGIN=<value>]`\
 _Default Profile:  default_\
 _Default Method:  automatic if a probe is detected, otherwise manual_ \
 _Default Adaptive: 0_ \
@@ -436,8 +519,9 @@ Initiates the probing procedure for Bed Mesh Calibration.
 
 The mesh will be saved into a profile specified by the `PROFILE` parameter,
 or `default` if unspecified. If `METHOD=manual` is selected then manual probing
-will occur.  When switching between automatic and manual probing the generated
-mesh points will automatically be adjusted.
+will occur.  If `METHOD=scan` is supplied with a supported probe then surface
+scanning will occur. When switching between probing methods the generated probe
+points will automatically be adjusted.
 
 It is possible to specify mesh parameters to modify the probed area.  The
 following parameters are available:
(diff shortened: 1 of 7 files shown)


KevinOConnor:
Thanks!  I have some high-level comments:
1. This functionality looks useful and I think it is worthwhile to add to Klipper.
2. The interaction between bed_mesh.py, probe.py, and probe_eddy_current.py looks very subtle and I wonder if refactoring that interaction would make sense.  As I understand this new code, bed_mesh.py instantiates probe.py ProbePointsHelper, but instead of passing it a regular Python list it passes in a bed_mesh.py PathGenerator class which mimics a Python list.  Then probe.py's ProbePointsHelper checks for a special HAS_SCANNING flag, and replaces its normal logic with probe_eddy_current.py's ProbeScanHelper class.  That class checks for the existence of a rapid_iter method, and then calls that - which causes execution to jump back to bed_mesh.py's PathGenerator class.  The code flow and data requirements between bed_mesh.py, probe.py, and probe_eddy_current.py seem complex.  I wonder it there is a way to make this interaction more clear, but I'm not sure.
3. Are the bed_mesh.py changes needed for the main "scanning" code changes to probe_eddy_current.py?  Or are those changes an additional enhancement on top of regular "scanning mode"?
4. I noticed this PR adds a third mechanism for gathering eddy current measurements in probe_eddy_current.py ( `EddyCalibration:handle_batch()`, `EddyEndstopWrapper:_add_measurement()`, and `StreamingContext:_bulk_callback()` ).  This looks like something that should be refactored.  That could be done after merge though.

Thanks again,
-Kevin

Arksine:
Hi Kevin, I'll try to address each question comment.

> The code flow and data requirements between bed_mesh.py, probe.py, and probe_eddy_current.py seem complex. I wonder it there is a way to make this interaction more clear, but I'm not sure.

Bed Mesh could instantiate the ProbeScanHelper rather than go through the ProbePointsHelper.  I avoided this since I thought the ProbeScanHelper might be useful for other modules that use the probe.  Other than that, I'm open to ideas, but I'm not sure if there is a better way to handle it.

> Are the bed_mesh.py changes needed for the main "scanning" code changes to probe_eddy_current.py? Or are those changes an additional enhancement on top of regular "scanning mode"?

They are an enhancement, however its necessary to get the best results for a "rapid" scan.  Specifically the rapid mode works best when some overshoot is applied to the points where a change in direction occurs.  It also helps to add some curvature to the direction change.   In addition its necessary to avoid entering faulty regions if possible.

I had considered refactoring Bed Mesh's "points" into a generator previously.  Since point generation is configurable at runtime and `bed_mesh` always generates a new set of points prior to calibration I believe a generator makes sense.

> I noticed this PR adds a third mechanism for gathering eddy current measurements in probe_eddy_current.py ( EddyCalibration:handle_batch(), EddyEndstopWrapper:_add_measurement(), and StreamingContext:_bulk_callback() ). This looks like something that should be refactored. That could be done after merge though.

I agree.  FWIW, when performing a traditional probe (typically during QGL) on occasion I get a "Unable to obtain probe_eddy_current sensor readings" error.  I suspect this is a timing issue, the samples are still being collected by the bulk interface and the bulk callback has yet to be triggered. The fix may be as simple as increasing the delay to .3 seconds.

KevinOConnor:
Hi.

Sorry - been super busy the last couple of weeks.

> >   The code flow and data requirements between bed_mesh.py, probe.py, and probe_eddy_current.py seem complex. I wonder it there is a way to make this interaction more clear, but I'm not sure.

> Bed Mesh could instantiate the ProbeScanHelper rather than go through the ProbePointsHelper. I avoided this since I thought the ProbeScanHelper might be useful for other modules that use the probe. Other than that, I'm open to ideas, but I'm not sure if there is a better way to handle it.

That makes sense.  I'll take another look over the next few days to see if any ideas "pop up" for me.

> They are an enhancement, however its necessary to get the best results for a "rapid" scan.

I agree on the utility of the functionality.  I was more wondering about merge ordering - possibly merging basic scanning and then advanced scanning on top.  Not a big deal though.

> FWIW, when performing a traditional probe (typically during QGL) on occasion I get a "Unable to obtain probe_eddy_current sensor readings" error.

One thing to note is that the current probing code calls `probe.multi_probe_begin()` before starting any probes, and that function calls `sensor_helper.add_client()`.  Thus, we spin up the ldc1612 reporting before any probes, keep it active during all probes, and then only halt measurements after the entire series of probes is done.

I did not see an equivalent mechanism in the new scanning code, but it's possible I missed it.  If `sensor_helper.add_client()` is only called at the start of each probe attempt then it wont be stable.  There is a startup and shutdown delay in the bulk_sensor system - so depending on timing of a stop/restart, samples might continue uninterrupted, or samples might have large gaps in them.  Symptoms would be seeing lots of "LDC1612 starting '%s' measurements" and "LDC1612 finished '%s' measurements" in the logs (as opposed to seeing them just once during a probe series).

If samples are started once for all probes, then there shouldn't be timing gaps in the reports.  So, if you're seeing that then we'll need to track down why it is occurring.  If you've got a test case that shows it I'll try to reproduce it here.

Thanks again,
-Kevin

Arksine:
> Sorry - been super busy the last couple of weeks.

Completely understand, we'll get it done on your timeline.

> I agree on the utility of the functionality. I was more wondering about merge ordering - possibly merging basic scanning and then advanced scanning on top. Not a big deal though.

Ah, I see.  I think it would be possible to just pick commit f359e43cbdefb1608e8e39ca7b22be1bdb7a37af if that is what you would prefer.

I did attempt to perform thorough testing of the Bed Mesh changes, although there is always the chance of a regression with this kind of change.   In the event that something strange pops up a user can dump the entire mesh state using the new `BED_MESH_DUMP` command, which should make analysis and debugging easier.

> If samples are started once for all probes, then there shouldn't be timing gaps in the reports. So, if you're seeing that then we'll need to track down why it is occurring. If you've got a test case that shows it I'll try to reproduce it here.

I haven't run into the issue with scanning.  The `StreamingContext` only adds a client once, then all samples are collected.  Traditional probe attempts (probing that moves the Z axis down) occasionally return with no  valid samples, so the probing procedure raises an exception and returns an error.  Generally speaking, we need to use the traditional functionality with calibration methods like Quad Gantry Level, because the surface could be too far away for scanning to work.  So the issue is not specifically related to this PR, but I thought I would bring it up in case you wanted me to make an attempt to correct it.  For example, something like the following might fix it:

```diff
diff --git a/klippy/extras/probe_eddy_current.py b/klippy/extras/probe_eddy_current.py
index 2b78464d0..a1a1c271c 100644
--- a/klippy/extras/probe_eddy_current.py
+++ b/klippy/extras/probe_eddy_current.py
@@ -266,7 +266,7 @@ class EddyEndstopWrapper:
         while 1:
             systime = reactor.monotonic()
             est_print_time = self._mcu.estimated_print_time(systime)
-            need_delay = self._trigger_time + 0.200 - est_print_time
+            need_delay = self._trigger_time + 0.300 - est_print_time
             if need_delay <= 0.:
                 break
             reactor.pause(systime + need_delay)
```

KevinOConnor:
Thanks again for working on this.  Again, sorry for the delay in responding.

My main feedback is that I think we should refactor the probing interface prior to merging this code.  I can do this refactoring work, and I can have something up for review next week.  I think we should do the refactoring first, as I fear if we add more code with the existing probe interface, it will be increasingly difficult to perform a refactor.

In the future, it seems there are going to be 4 different probing mechanisms with the ldc1612: "descend_until_trigger", "scan", "rapid_scan", and "descend_until_tap".  There are also 4 different "probing classes" in Klipper today: probe.py, bltouch.py, probe_eddy_current.py, and smart_effector.py.  I think we can rework the "probing interface" in all of these classes so that we can better support the different probing mechanisms in the future.

Roughly, here's what I'm thinking we can change the probing interface to:

* All probing commands start by calling `probe.multi_probe_begin()` and the `gcmd` instance is passed to that function.  This will allow the low-level probing implementation to get access to the probe parameters.

* The actual probing will be done with `probe.detect_position()` (instead of `probe.run_probe()`).  This function may move the toolhead (but is not required to), it may perform a full kinematic flush (but is not required to), and may consume some amount of toolhead time (but is not required to).  This function does not return anything.  The calling code may invoke `probe.detect_position()` multiple times and may move the toolhead to new probing positions between each call.

* At the completion of the probing operations, `probe.multi_probe_end()` will be invoked, and this method will return a list with one entry for every `probe.detect_position()` call.  Each entry will contain a 2-tuple: `(nominal_position, detected_position)`.  Where `detected_position` contains the estimated Z height at the given `nominal_position`.  The `nominal_position` will be roughly around the XY coordinate of the toolhead during each `detect_position()` call, but is not required to be any particular exact toolhead position.

I'm hoping that the above interface will scale to future ldc1612 and loadcell requirements.

With this interface, I don't think we would need specific "scan" logic in `ProbePointsHelper` to implement `METHOD=scan` requests.  In this case, the probe_eddy_current.py code can detect `METHOD=scan` in its `probe.multi_probe_begin()` method and then not perform a descending move (but otherwise use the same frequency averaging and height detection that "descend_until_trigger" uses).

What are your thoughts on the interface?

Some other notes:

* The timing in the upstream probe_eddy_current.py `probing_move()` code is indeed busted.  We can increase the delay, but I fear even that may risk "timing race conditions".  I have an alternate fix (wait for the samples to exceed the `end_time`), but we may also just want to rework the interface.  Not sure.

* The upstream probe_eddy_current.py assignment to `self._is_from_home` is indeed a bug.

* The most recent patch series here seems to be missing a `@staticmethod` on `to_str()`.  Not sure if that's a py2/py3 thing.

* I think we should only pass a real list (not a "list generator") to `ProbePointsHelper`.  I fear that will be too subtle for future developers working on `ProbePointsHelper`, as changes to that class may alter the evaluation of the list resulting in subtle behavior changes that are unlikely to get caught in normal testing (as not all developers are likely to have "rapid_scan" capable hardware).

* For performing a "rapid_scan", I wonder if bed_mesh can directly detect a `METHOD=rapid_scan` request in it's `cmd_BED_MESH_CALIBRATE()`.  It can then directly lookup the `PrinterEddyProbe` instance (via `printer.lookup_object()`), and directly invoke it's `add_client()` method.  It can then perform the moves and accumulate the corresponding `time,frequency,z` samples.  I think that may help "flatten" the interactions that this PR has in `PathGenerator`, `MoveSplitter`, `ProbePointsHelper`, `ProbeScanHelper`, and `StreamingContext` classes.

This is all up for discussion, so let me know what your thoughts are.

Thanks again,
-Kevin

Arksine:
Thanks for the feedback Kevin.

> With this interface, I don't think we would need specific "scan" logic in ProbePointsHelper to implement METHOD=scan requests. In this case, the probe_eddy_current.py code can detect METHOD=scan in its probe.multi_probe_begin() method and then not perform a descending move (but otherwise use the same frequency averaging and height detection that "descend_until_trigger" uses).
>
>What are your thoughts on the interface?

This sounds like an improvement.  I agree that we need a better way to handle all of the different ways a probe sample can be taken.  Presumably we would still keep `run_probe()` for situations where we just need to probe a single location?

> The most recent patch series here seems to be missing a @staticmethod on to_str(). Not sure if that's a py2/py3 thing.

Indeed I left it out.  I'm surprised it works without the decorator, that could be a Python 3 thing.

> For performing a "rapid_scan", I wonder if bed_mesh can directly detect a METHOD=rapid_scan request in it's cmd_BED_MESH_CALIBRATE(). It can then directly lookup the PrinterEddyProbe instance (via printer.lookup_object()), and directly invoke it's add_client() method. It can then perform the moves and accumulate the corresponding time,frequency,z samples. I think that may help "flatten" the interactions that this PR has in PathGenerator, MoveSplitter, ProbePointsHelper, ProbeScanHelper, and StreamingContext classes.

Sounds good to me.  In summary, it looks like we can remove `ProbeScanHelper`  from `probe_eddy_current.py`, and I will implement something similar, specific to rapid scanning, in  `bed_mesh`.  It looks like the changes to `bed_mesh` can be done independently from the probe refactor, so I'll begin work on that.

deriyalexey:
Hi both, when we can expect to see this changes?

Arksine:
Closing this PR, as it has been superseded by #6617.

PR #6560, opened 2024-04-11
sensors: add sht31 support

nefelim4ag:
I don't know why no one added this before, because Klipper already has an SHT2x family
So I got one and implemented support.

Hope it will be useful.

Thanks!

---
My current motivation:
I have BME* family for the chamber and SHT because wider temperature support will be used for the filament hot box/dryer, I can and will heat it to 110 C, and support for humidity will help in that case.
/* And SHT3x/SHT4x are easier to get nowadays */

[Datasheet](https://sensirion.com/media/documents/213E6A3B/63A5A569/Datasheet_SHT3x_DIS.pdf)

### DIFF
--- a/klippy/extras/sht3x.py
+++ b/klippy/extras/sht3x.py
@@ -1,0 +1,165 @@
+# SHT3X i2c based temperature sensors support
+#
+# Copyright (C) 2024  Timofey Titovets <nefelim4ag@gmail.com>
+# Based on htu21d.py code
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging
+from . import bus
+
+######################################################################
+# Compatible Sensors:
+#       SHT31  - Tested on octopus pro and Linux MCU
+#
+######################################################################
+SHT3X_I2C_ADDR = 0x44
+
+SHT3X_CMD = {
+    'MEASURE': {
+        'STRETCH_ENABLED': {
+            'HIGH_REP': [0x2c, 0x06], # High   (15ms) repeatability measurement
+            'MED_REP': [0x2c, 0x0D],  # Medium (6ms)  repeatability measurement
+            'LOW_REP': [0x2c, 0x10],  # Low    (4ms)  repeatability measurement
+        },
+        'STRETCH_DISABLED' : {
+            'HIGH_REP': [0x24, 0x00],
+            'MED_REP': [0x24, 0x0B],
+            'LOW_REP': [0x24, 0x16],
+        },
+    },
+    'OTHER': {
+        'STATUS': {
+            'READ': [0xF3, 0x2D],
+            'CLEAN': [0x30, 0x41],
+        },
+        'SOFTRESET': [0x30, 0xA2], # Soft reset
+        'HEATER': {
+            "ENABLE": [0x30, 0x6D],
+            "DISABLE": [0x30, 0x66],
+        },
+        'FETCH': [0xE0, 0x00],
+        'BREAK': [0x30, 0x93],
+    }
+}
+
+class SHT3X:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.name = config.get_name().split()[-1]
+        self.reactor = self.printer.get_reactor()
+        self.i2c = bus.MCU_I2C_from_config(
+            config, default_addr=SHT3X_I2C_ADDR, default_speed=100000)
+        self.report_time = config.getint('sht3x_report_time', 1, minval=1)
+        self.deviceId = config.get('sensor_type')
+        self.temp = self.min_temp = self.max_temp = self.humidity = 0.
+        self.sample_timer = self.reactor.register_timer(self._sample_sht3x)
+        self.printer.add_object("sht3x " + self.name, self)
+        self.printer.register_event_handler("klippy:connect",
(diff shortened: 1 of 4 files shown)


JamesH1978:

Thanks
James

klippy/extras/sht3x.py
```diff
@@ -0,0 +1,165 @@
...
+                    crc = (crc << 1) ^ SHT3X_CRC8_POLYNOMINAL
+                else:
+                    crc <<= 1
+        return crc & 0xFF
+
+    def get_status(self, eventtime):
+        return {
+            'temperature': round(self.temp, 2),
+            'humidity': round(self.humidity, 1),
+        }
```
KevinOConnor, line 160: This exports status information, but it does not look like docs/Status_Reference.md was updated.

klippy/extras/temperature_sensors.cfg
```diff
@@ -1,6 +1,5 @@
 # This file loads the default temperature sensors.
 
-
```
KevinOConnor, line 3: Please don't make unrelated whitespace changes.

klippy/extras/temperature_sensors.cfg
```diff
@@ -36,7 +37,6 @@
 # Load "temperature_combined" sensor
 [temperature_combined]
 
-
```
KevinOConnor, line 39: Please don't make unrelated whitespace changes.

KevinOConnor (commented):
Thanks.  In general it seems fine to me.  I have a few comments - see below.

-Kevin

nefelim4ag:
Just notice,
I have tried to use a macro like  "QUERY_SHT3X" and looks like a console from Fluidd just ignores the last 'X' and shows errors like:
```
Unknown command 'QUERY_SHT3'
```

Maybe if Klipper internally has some sort of special handling of trailing 'X'
The module named with X on End is not perfect?

TheFuzzyGiggler:
I might be blind but I don't see anywhere you implement the GCode command

```
class SHT3X:
    def __init__(self, config):
        self.name = config.get_name().split()[-1]
        self.reactor = self.printer.get_reactor()
        self.i2c = bus.MCU_I2C_from_config(
            config, default_addr=SHT3X_I2C_ADDR, default_speed=100000)
        self.report_time = config.getint('sht3x_report_time', 1, minval=1)
        self.deviceId = config.get('sensor_type')
        self.temp = self.min_temp = self.max_temp = self.humidity = 0.
        self.sample_timer = self.reactor.register_timer(self._sample_sht3x)
        self.printer.add_object("sht3x " + self.name, self)
        self.printer.register_event_handler("klippy:connect",
                                            self.handle_connect)
```

You need to have something like 

```
self.gcode.register_command('QUERY SH3X', self.cmd_QUERY_SHT3X,
           desc=self.cmd_QUERY_SHT3X_help)
```

and then define it

```
cmd_QUERY_SHT3X_help = ("Queries the values from the SHT3X Temp/Humidity Sensor")
def cmd_QUERY_SHT3X(self,gcmd):
     etime = self.reactor.monotonic()
     return self._sample_sht3x(self, etime)
```

nefelim4ag:
@TheFuzzyGiggler 
I mean something like this:
https://github.com/Klipper3d/klipper/blob/2425a74638baa87efef3ca02253804d126101c8a/docs/Command_Templates.md?plain=1#L175

The same copy-paste macro for this sensor has been dropped by request of @KevinOConnor above.

image
image

KevinOConnor:
Thanks.

-Kevin

Murrdo:
Humidity is not displayed
image

nefelim4ag:
@Murrdo which version of Klipper are you running?
What does your macro look like?

From my point of view, this is an impossible error.
```
        return {
            'temperature': round(self.temp, 2),
            'humidity': round(self.humidity, 1),
        }
```

lukas-m5:
I currently have an SHT31 sensor in use and can read the temperature. Unfortunately not the humidity. 
There is always this error message:
 !! Error evaluating 'gcode_macro M521:gcode': jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'humidity'
Here is my Macro:
[gcode_macro M521]
gcode: 
    {% set temp = (printer["temperature_sensor filament_chamber_sensor"].temperature*10)|int %}
    {% set humidity = printer["temperature_sensor filament_chamber_sensor"].humidity %}
	{% set target = printer.save_variables.variables.var_filament_chamber_target_temp*10 %}
    {% set a = 17.27 %}
    {% set b = 237.7 %}
    # Näherungsformel für den Taupunkt
    {% set alpha = ((a * temp) / (b + temp)) + (0.434 * (humidity / 100)) %}
    {% set dewpoint = (b * alpha) / (a - alpha) %}
    M118 Filamentbox T{temp} H{humidity} TP{dewpoint} S{target}

Which command do I have to use to read the humidity?

nefelim4ag:
```
[gcode_macro QUERY_SHT31]
gcode:
    {% set sensor = printer["sht3x my_sensor"] %}
    {action_respond_info(
        "Temperature: %.2f C\n"
        "Humidity: %.2f%%" % (
            sensor.temperature,
            sensor.humidity))}
```

PR #6561, opened 2024-04-11
homing_override: Adds rawparams support

pedrolamas:
The `[homing_override]` has a gcode template that allows usage of the `params` but not `rawparams`.

This PR adds `rawparams` support to it thus making it consistent with the https://github.com/Klipper3d/klipper/blob/master/docs/Command_Templates.md#the-rawparams-variable information.


KevinOConnor:
Thanks.

-Kevin

PR #6563, opened 2024-04-12
Recommend using "ip" instead of "ifconfig" in CANBUS.md

KevinOConnor:
Some Linux systems do not install ifconfig, while ip should always be available.  So, update the canbus documentation to recommend that.

-Kevin

### DIFF
--- a/docs/CANBUS.md
+++ b/docs/CANBUS.md
@@ -31,7 +31,7 @@ adapter. This is typically done by creating a new file named
 allow-hotplug can0
 iface can0 can static
     bitrate 1000000
-    up ifconfig $IFACE txqueuelen 128
+    up ip link set $IFACE txqueuelen 128
 ```
 
 ## Terminating Resistors
@@ -113,7 +113,7 @@ Some important notes when using this mode:
 allow-hotplug can0
 iface can0 can static
     bitrate 1000000
-    up ifconfig $IFACE txqueuelen 128
+    up ip link set $IFACE txqueuelen 128
 ```
 
 * The "bridge mcu" is not actually on the CAN bus. Messages to and


Sineos:
Depending on the Linux distribution and version, it seems that different approaches are needed.

Using `systemd` ( version newer than v248 required):
To use the networking support of `systemd`, one needs to create the file `/etc/systemd/network/80-can.network` with following content:

```
[Match]
Name=can*

[Link]
TransmitQueueLength=128
RequiredForOnline=no

[CAN]
BitRate=1000000
```

The configuration change is applied and activated by running:
```
sudo systemctl daemon-reload
sudo networkctl reload
```

If Linux appears to be rejecting the queue length, the `udev` system can be used to force this parameter. To do so, create a file named `/etc/udev/rules.d/99-persistent-can.rules` with the following content:

```
SUBSYSTEM=="net", ACTION=="add|change", KERNEL=="can*", ATTR{tx_queue_len}="128"
```

This change is activated by issuing the following commands:
```
sudo udevadm control --reload-rules
sudo udevadm trigger
```

KevinOConnor:
So, are you saying this change from "ifconfig" to "ip" could break systems that were previously working?

I understand the `/etc/network/interfaces.d/can0` doesn't necessarily work on all systems, but I'm not sure if you are reporting that this change could be a regression.

Thanks,
-Kevin

Sineos:
As far as I can tell, there are distributions that do not process `/etc/network/interfaces.d/*`, but rely on systemd-networkd or netplan as their network management backend (often desktop flavors, it seems).

Also, there are reports that Linux (for whatever reason) is finicky about accepting TX queue lengths.

These additions are meant to show alternatives in case the user runs into trouble with the regular approach.

PR #6564, opened 2024-04-12
Add information on txqueuelen to CANBUS_Troubleshooting.md

KevinOConnor:
I think there has been some confusion on the CAN bus `txqueuelen` parameter.  This PR is an attempt to clarify the current recommendation, provide information on related error messages, and provide some guidance on things to consider when configuring it.

-Kevin


Sineos (approved):
Interesting. Never thought about the timing effect of the system-sided buffer. 

Seems all very clear to me.

PR #6565, opened 2024-04-13
virtual_sdcard: Fix to process last line on file

pedrolamas:
Due to the way Klipper currently reads the gcode file in 8k buffers, the last line will not be processed as it will remain on the `partial_input` when then last read returns nil.

This PR fixes that by checking that `partial_input` is valid, it will put it on the `lines` array to process and on the next nil read will just terminate the print as it currently does.

This also changes the `next_file_position` calculation to ensure it is always under `file_size` (as we always do `+ 1` on file position to account for the line breaks, this will overflow the file size and cause progress to go above 1.0)

### DIFF
--- a/klippy/extras/virtual_sdcard.py
+++ b/klippy/extras/virtual_sdcard.py
@@ -238,17 +238,21 @@ class VirtualSD:
                 except:
                     logging.exception("virtual_sdcard read")
                     break
-                if not data:
+                if data:
+                    lines = data.split('\n')
+                    lines[0] = partial_input + lines[0]
+                    partial_input = lines.pop()
+                    lines.reverse()
+                elif partial_input:
+                    lines = [partial_input]
+                    partial_input = ""
+                else:
                     # End of file
                     self.current_file.close()
                     self.current_file = None
                     logging.info("Finished SD card print")
                     self.gcode.respond_raw("Done printing file")
                     break
-                lines = data.split('\n')
-                lines[0] = partial_input + lines[0]
-                partial_input = lines.pop()
-                lines.reverse()
                 self.reactor.pause(self.reactor.NOW)
                 continue
             # Pause if any other request is pending in the gcode class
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  As background, g-code has some very arcane rules on how g-code messages are to be processed.  Klipper follows the rule that for every newline character (aka '\n', aka 0x0A) that it receives it will generate exactly one "ok\n" response message.  Getting this correct is important when reading g-code from the pseudo-terminal as print control tools (like OctoPrint) rely on these "ok\n" response messages (and their exact quantity) as part of the g-code flow control system.

This handling of newlines was carried over to the virtual_sdcard tool.  It seems odd to me that a slicer would generate a file without a newline at the end.  Arguably, if the last line doesn't end with a newline then it isn't a command - it's arguably just a partial command.  (It's also really not a good idea to write out text files without a trailing newline as many common tools can have unexpected behavior, and it makes it hard to distinguish between a complete file and an abruptly truncated file.)

What tool produced these files without a trailing newline?

I suppose we could work around this in Klipper.  It'd help to better understand what the impact is.

Or, perhaps I'm misunderstanding this report.

Cheers,
-Kevin

pedrolamas:
Thanks Kevin, I now completely see your point here: in terms of execution, Klipper will only execute commands that terminate with a new line (just like a terminal).

The reality is that the above g-code was post processed with klipper_estimator, so I guess it is missing a final new line in it.

I've performed a quick check in Cura and Orca Slicer and they both add a final empty line, which would then have the correct behavior as you described here.

pedrolamas:
The issue that was identified in klipper_estimator has now been fixed and a new release published.

I will keep this open for now, but considering your comments I think it is safe to assume that the changes here are actually not to be merged.

Given the above, I do believe it makes sense that progress does not go to 100% when a file doesn't terminate in `\n` (as this progress is in terms of file position, not the actual print progress)

PR #6566, opened 2024-04-14
Refactor bulksensor timestamp calculation code

KevinOConnor:
The code to calculate a timestamp for each sample in a bulk_sensor report is complex, and every sensor implementation was performing the calculation.  This PR moves that calculation into the generic ChipClockUpdater class.  This simplifies the sensor specific code.

@garethky - fyi.

-Kevin

### DIFF
--- a/klippy/extras/adxl345.py
+++ b/klippy/extras/adxl345.py
@@ -184,9 +184,6 @@ def read_axes_map(config):
         raise config.error("Invalid axes_map parameter")
     return [am[a.strip()] for a in axes_map]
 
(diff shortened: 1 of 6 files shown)


garethky:
This is exactly what I was hoping for!

KevinOConnor:
I added an additional commit to this PR that further simplifies the bulk sensor implementations.  As part of this additional commit, the ChipClockUpdater class has been renamed to FixedFreqReader (and the new class will handle the low-level queuing of the `sensor_bulk_data` messages).

-Kevin

PR #6567, opened 2024-04-14
Add config files for Recore A6, A7 and A8

eliasbakken:
The ina381 is a current monitor circuit with over current alarm. In order for Recore boards to work, this circuit needs a reset and that is what this plugin does.


docs/Config_Reference.md
```diff
@@ -4652,6 +4652,15 @@ host_mcu:
 #   (True sets CFG5 high, False sets it low). The default is True.
 ```
 
+### [ina381]
+The INA381 is an current monitor chip with a programmable over current
+alarm.
+
+This chip needs a reset pulse in order to operate in latching mode,
+so that is all this plugin does. The rest of the functionality can be
+implemented using already existing functionality in Klipper.
```
KevinOConnor, line 4661: The config reference would need to describe all the available config options (eg, `reset_pin`) and their defaults.

klippy/extras/ina381.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+class ina381:
+    def __init__(self, config):
+        printer = config.get_printer()
+        ppins = printer.lookup_object('pins')
+        ppins.register_chip('ina381', self)
+
+        # Reset over current alarm
+        reset_pin = config.get('reset_pin', 'ar100:PF4')
```
KevinOConnor, line 13: It would not make sense to use a default that contains a user specified mcu name.

klippy/extras/ina381.py
```diff
@@ -0,0 +1,21 @@
...
+        oc_reset = ppins.setup_pin('digital_out', reset_pin)
+        mcu = oc_reset.get_mcu()
+        pin = oc_reset._pin
```
KevinOConnor, line 16: Code should not "peek" into class member variables of classes defined in other python files - this is mentioned briefly at https://www.klipper3d.org/Code_Overview.html#adding-a-host-module .

KevinOConnor (commented):
Thanks.

FYI, I can review code modules, and James reviews config files.  So, it can be a little challenging to mix them in the same PR.

See my code comments below.

-Kevin

docs/Config_Reference.md
eliasbakken: It would be great to be able to fix this without requiring a custom module. I've done some tests with this suggestion and I was not able to get delayed_gcode to work, but after thinking about this a bit I think its just as good to let this be a macro button that the user can press if the over current alarm is tripped. I think that is safer. I will try to implement this in the config files and resubmit.

eliasbakken:
I've implemented this PR as a pure config file change as pr your suggestion.

KevinOConnor:
Thanks.  In general, it seems fine to me.  At a minimum the configs need to be added to the build test cases though ( https://www.klipper3d.org/Example_Configs.html point 4).

@JamesH1978 - do you have any comments?

-Kevin

JamesH1978:
Config looks fine, my only main comment would be why you have such a low min_extrude_temp and a min_temp of 1 kelvin for your extruders?

Once the regression tests are added as per Kevin's comment, i will do a few review.

Thanks
James

eliasbakken:
Thank you both. I've added the config files as tests, but in order to make them pass I had to make some code changes, so I'll try and submit that as a separate PR that can be reviewed first.

PR #6568, opened 2024-04-15
Fix float inaccuracy when checking for probe movement during probe activate/deactivate gcode

Matszwe02:
This PR allows to move the probe between probing, but ensures that the probe returns to the original position. With the previous approach, any movement will cause float inaccuracy, which triggers the condition. With the fixed approach, the movement is allowed, if the probe goes to its original position at the end of the gcode.
That feature is helpful for some exotic probes (like mine) which require movement between probing.

Also fixed typo.


JamesH1978:

Thanks
James

Matszwe02:
Fixed in this PR:
https://github.com/Klipper3d/klipper/pull/6574#issue-2251564564

PR #6570, opened 2024-04-17
bed_mesh: Limits for calculation of z_offset added.

Lebensgefahr:
In case of sending printer head to a high coordinates klipper process loads CPU and in some cases it causes "Timer too close" error thrown.


Lebensgefahr:
Sorry, I need to check it accurately.

PR #6571, opened 2024-04-18
temperature_fan modified to have working pid option

sunbowch:
Module: temperature_fan: ControlPID option modified to get the expected behavior.
Module: Config_Reference: To correspond to the modified ControlPID class.

Whereas the bang-bang control works as expected (fan turns on when measured temp is > reference temp), the ControlPID as designed doesn't have the same behavior and the computed duty cycle is max_speed - PID value, which is not the expected behavior. (in my point of view)

The correction allows the PID to work as expected: the fan will turn on if temperature increase above reference and will speed up if the difference increase. It will eventually shut down if the temperature is below reference. When powered on the fan stays within min_speeed<->max_speed range.

In this case, the minimum fan speed should be defined near the minimum PWM duty cycle value to keep the fan spinning. A tolerance equal to max_delta has been programmed to avoid that the fan oscillate rapidly between 0 and min_speed near the reference value. It only kick-in when the measured temperature is >reference temperature + max_delta and turns off at reference temperature - max_delta, making the fan control effectively back to bang-bang between 0 and min_speed.

The temp_err variable is < 0 when the fan should power-on and if the value is lower, the fan must speed up, so all PID coefficients must be negative. (or the temp_err should be inverted).

Used successfully to control the radiator fan of a water-cooled printer more silently than with the bang-bang control: A stable heat source (head + motors) results in a stable fan rpm near the reference temperature and if the heat stops, the fan stops with some delay.

As modified, it can be used to control the exhaust (or intake) fan of an enclosure, if it is passively heated by the bed, and maintain its reference temperature independently of the bed surface temp (as long as the bed's power is sufficient).

See canceled commit #6547 for a long discussion about the opportunity to have a PID control at all.

The goal of this PR is only to make the option functional without questioning its specific utility.

Signed-off-by: Luc JOB <sunbow@bluewin.ch>

### DIFF
--- a/klippy/extras/fan_generic.py
+++ b/klippy/extras/fan_generic.py
@@ -3,6 +3,7 @@
 # Copyright (C) 2016-2024  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
+import logging
 from . import fan, output_pin
 
 class PrinterFanGeneric:


JamesH1978:

Thanks
James

PR #6574, opened 2024-04-18
probe: Fixed incorrect "Toolhead moved" errors

Matszwe02:
Toolhead can now move between probes, as long as it returns during (de)activate_gcode

[Applied fixes to this PR](https://github.com/Klipper3d/klipper/pull/6568#issue-2244734734)

> This PR allows to move the probe between probing, but ensures that the probe returns to the original position. With the previous approach, any movement will cause float inaccuracy, which triggers the condition. With the fixed approach, the movement is allowed, if the probe goes to its original position at the end of the gcode.
That feature is helpful for some exotic probes (like mine) which require movement between probing.

### DIFF
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -328,16 +328,18 @@ class ProbeEndstopWrapper:
         toolhead = self.printer.lookup_object('toolhead')
         start_pos = toolhead.get_position()
         self.deactivate_gcode.run_gcode_from_command()
-        if toolhead.get_position()[:3] != start_pos[:3]:
+        position_pairs = zip(toolhead.get_position()[:3], start_pos[:3])
+        if any(abs(a - b) > 1e-6 for a, b in position_pairs):
             raise self.printer.command_error(
-                "Toolhead moved during probe activate_gcode script")
+                "Toolhead moved during probe deactivate_gcode script")
     def _lower_probe(self):
         toolhead = self.printer.lookup_object('toolhead')
         start_pos = toolhead.get_position()
         self.activate_gcode.run_gcode_from_command()
-        if toolhead.get_position()[:3] != start_pos[:3]:
+        position_pairs = zip(toolhead.get_position()[:3], start_pos[:3])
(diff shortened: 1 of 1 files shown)


TheFuzzyGiggler:
Out it curiosity (it might be obvious but I'm not thinking of it) what probe requires X/Y movement during triggering? That seems to contradict the definition of being a probe, as the probe is probing that spot.

Matszwe02:
> Out it curiosity (it might be obvious but I'm not thinking of it) what probe requires X/Y movement during triggering? That seems to contradict the definition of being a probe, as the probe is probing that spot.

I have a loadcell mounted on the z axis, which requires moving z up and down a bit to reset very little of backlash that otherwise will trip the probe

KevinOConnor:
Thanks.  This change looks odd to me as the purpose of the error is to warn users that Klipper's internal calculations do not support movement during the macros.  It's not clear to me that these internal calculations support even a little movement - an audit of the code would be needed to ensure the code works correctly.

If you need to take certain movements before or after probing, it's probably best to do that in a macro that invokes the probing process.

Separately, I committed the fix to the error messages (commit 7b490f3e).

-Kevin

Matszwe02:
> Thanks. This change looks odd to me as the purpose of the error is to warn users that Klipper's internal calculations do not support movement during the macros. It's not clear to me that these internal calculations support even a little movement - an audit of the code would be needed to ensure the code works correctly.
> 
> If you need to take certain movements before or after probing, it's probably best to do that in a macro that invokes the probing process.
> 
> Separately, I committed the fix to the error messages (commit [7b490f3](https://github.com/Klipper3d/klipper/commit/7b490f3ec1d3f0477b90c74d96f51ba1fb457a46)).
> 
> -Kevin

Thanks for the response. Do you mean, that klipper's motion queue needs to be checked for that small movements, or my PR's code?

I have been printing for a long time, and it appears to be working well. I will also need to check for edge case scenarios to make sure it will and won't trip when needed.

PR #6576, opened 2024-04-23
bme280: Add BMP388 sensor support to BMxx80

linw1995:
Extends the BMxx80 category with support for the BMP388 sensor, providing temperature and pressure output similar to the existing BMxx80 class of sensors.


KevinOConnor:
Thanks.  Maybe @Arksine could take a look at this.

-Kevin

Arksine (approved):

KevinOConnor:
Thanks.

-Kevin

PR #6579, opened 2024-04-26
Include missing instructions

pfych:
The specified Creality boards requires a baud rate of 115200

This [seems to be a common issue](https://old.reddit.com/r/klippers/comments/skgf5f/mcumcu_unable_to_connect/iotd0rd/) when setting up Ender 3s with Klipper.


JamesH1978:
All boards from creality can run at 250000. I don't know what issue that Reddit person was having but especially the 4.2.2 has no issues running at full speed.

James

PR #6584, opened 2024-04-30
ads1x1x: added support for ADC chip

korsarNek:
Added support for ADC chips ADS1103, ADS1104, ADS1105, ADS1113, ADS1114 and ADS1115.

The PR is rougly based on https://github.com/Klipper3d/klipper/pull/5618 but with a different architecture.

I've tested it with an ADS1115 and a linux mcu host.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4843,6 +4843,9 @@ chip: ADS1115
 #   Default value is 4.096V. The maximum voltage range used for the input. This
 #   scales all values read from the ADC. Options are: 6.144V, 4.096V, 2.048V,
 #   1.024V, 0.512V, 0.256V
+#adc_voltage: 3.3
+#   The suppy voltage for the device. This allows additional software scaling
+#   for all values read from the ADC.
 i2c_mcu: host
 i2c_bus: i2c.1
 #address_pin: GND
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4830,28 +4830,24 @@ ADS1013, ADS1014, ADS1015, ADS1113, ADS1114 and ADS1115 are I2C based Analog to
 Digital Converters that can be used for temperature sensors. They provide 4
 analog input pins either as single line or as differential input.
 
+Note: Use caution if using this sensor to control heaters. The heater min_temp
+and max_temp are only verified in the host and only if the host is running and
+operating normally. (ADC inputs directly connected to the micro-controller
+verify min_temp and max_temp within the micro-controller and do not require a
+working connection to the host.)
+
 ```
 [ads1x1x my_ads1x1x]
 chip: ADS1115
-#pga: 6.144V
+#pga: 4.096V
 #   Default value is 4.096V. The maximum voltage range used for the input. This
 #   scales all values read from the ADC. Options are: 6.144V, 4.096V, 2.048V,
 #   1.024V, 0.512V, 0.256V
-#mode: single
-#    Default value is single. Turn off the chip after a single reading or keep
-#    it running. Turning off saves power but turning it back on will be slower.
-#    Options are single and continuous.
-#samples_per_second: 128
-#    Default value is 128. The amount of samples that the ADC can provide per
-#    second. A lower value makes the samples more accurate, but it takes longer
-#    until a new value is available.
-#    ADS101X's support 128, 250, 490, 920, 1600, 2400, 3300.
-#    ADS111X's support 8, 16, 32, 64, 128, 250, 475, 860.
 i2c_mcu: host
 i2c_bus: i2c.1
 #address_pin: GND
 #   Default value is GND.  There can be up to four addressed devices depending
-#   upon wiring of the device. Check the datasheet for details. The i2_address
+#   upon wiring of the device. Check the datasheet for details. The i2c_address
 #   can be specified directly instead of using the address_pin.
 ```
 
@@ -4861,7 +4857,7 @@ The chip provides pins that can be used on other sensors.
 sensor_type: ...
 #   Can be any thermistor or adc_temperature.
 sensor_pin: my_ads1x1x:AIN0
-#   A combination of the name of the ads1x1x chip that and the pin. Possible
+#   A combination of the name of the ads1x1x chip and the pin. Possible
 #   pin values are AIN0, AIN1, AIN2 and AIN3 for single ended lines and
 #   DIFF01, DIFF03, DIFF13 and DIFF23 for differential between their
 #   correspoding lines. For example
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2737,6 +2737,33 @@ sensor_type: temperature_host
 #   system file on a Raspberry Pi computer.
 ```
 
+### ADS Analog to Digital Converter as temperature sensor
+
+The sensors ADS1013, ADS1014, ADS1015, ADS1113, ADS1114 and ADS1115 are i2c based Analog to Digital Converters to which are thermistor can be connected in a voltage divider setup.
+Needs an [ads1x1x] section to configure the chip. Afterwards chip specific sensor_pins can be used.
+
+```
+sensor_type: ads1x1x
+#   A section for ads1s1x has to specified before the sensor can be used.
+#   Look at the ads1x1x below in this documentation.
+probe_type: Generic 3950
+#   The probe type indicates the calculation profile to use for the raw data
+#   from the ADC device. Check the thermistors list for possible values.
+sensor_pin: my_ads1x1x:AIN0
+#   A combination of the name of the ads1x1x chip that has been defined and
+#   the pin. Possible pin values are AIN0, AIN1, AIN2 and AIN3 for single ended
+#   lines and DIFF01, DIFF03, DIFF13 and DIFF23 for differential between their
+#   correspoding lines. For example
+#   DIFF03 measures the differential between line 0 and 3. Only specific
+#   combinations for the differentials are allowed.
+#pullup_resistor: 4700
+#   A setting that defines the physical wiring method of the sensor.
+#   Default is 4700.
+#voltage_offset: -0.295
+#   The differential voltage from the supplied voltage to the maximum pull-up
+#   voltage that the device under test (thermistor) sees.
+```
+
 ### DS18B20 temperature sensor
 
 DS18B20 is a 1-wire (w1) digital temperature sensor. Note that this sensor is not intended for use with extruders and heater beds, but rather for monitoring ambient temperature (C). These sensors have range up to 125 C, so are usable for e.g. chamber temperature monitoring. They can also function as simple fan/heater controllers. DS18B20 sensors are only supported on the "host mcu", e.g. the Raspberry Pi. The w1-gpio Linux kernel module must be installed.
@@ -4822,6 +4849,41 @@ vssa_pin:
 #   A time value (in seconds) over which the vref and vssa
 #   measurements will be smoothed to reduce the impact of measurement
 #   noise. The default is 2 seconds.
+```
+
+### [ads1x1x]
+
+ADS1013, ADS1014, ADS1015, ADS1113, ADS1114 and ADS1115 are I2C based Analog to
+Digital Converters that can be used for temperature sensors. They provide 4
+analog input pins either as single line or as differential input.
+
+```
+[ads1x1x my_name]
+chip: ADS1115
+#pga: 6.144V
+#   Default value is 4.096V. The maximum voltage range used for the input. This
+#   scales all values read from the ADC. Options are: 6.144V, 4.096V, 2.048V,
+#   1.024V, 0.512V, 0.256V
+#mode: single
+#    Default value is single. Turn off the chip after a single reading or keep
+#    it running. Turning off saves power but turning it back on will be slower.
+#    Options are single and continous.
+#samples_per_second: 128
+#    Default value is 128. The amount of samples that the ADC can provide per
+#    second. A lower value makes the samples more accurate, but it takes longer
+#    until a new value is available.
+#    ADS101X's support 128, 250, 490, 920, 1600, 2400, 3300.
+#    ADS111X's support 8, 16, 32, 64, 128, 250, 475, 860.
+i2c_mcu: host
+i2c_bus: i2c.1
+#adc_voltage: 5
+#   The suppy voltage for the device.
+#   Sensor is reading this amount different from external measuring device.
+#address_pin: GND
+#   Default value is GND.  There can be up to four addressed devices depending
+#   upon wiring of the device. Check the datasheet for details. The i2_address
+#   can be specified directly instead of using the address_pin.
+
 ```
 
 ### [replicape]
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4867,7 +4867,7 @@ chip: ADS1115
 #mode: single
 #    Default value is single. Turn off the chip after a single reading or keep
 #    it running. Turning off saves power but turning it back on will be slower.
-#    Options are single and continous.
+#    Options are single and continuous.
 #samples_per_second: 128
 #    Default value is 128. The amount of samples that the ADC can provide per
 #    second. A lower value makes the samples more accurate, but it takes longer
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2737,33 +2737,6 @@ sensor_type: temperature_host
 #   system file on a Raspberry Pi computer.
 ```
 
-### ADS Analog to Digital Converter as temperature sensor
-
-The sensors ADS1013, ADS1014, ADS1015, ADS1113, ADS1114 and ADS1115 are i2c based Analog to Digital Converters to which are thermistor can be connected in a voltage divider setup.
-Needs an [ads1x1x] section to configure the chip. Afterwards chip specific sensor_pins can be used.
-
-```
-sensor_type: ads1x1x
-#   A section for ads1s1x has to specified before the sensor can be used.
-#   Look at the ads1x1x below in this documentation.
-probe_type: Generic 3950
-#   The probe type indicates the calculation profile to use for the raw data
-#   from the ADC device. Check the thermistors list for possible values.
-sensor_pin: my_ads1x1x:AIN0
-#   A combination of the name of the ads1x1x chip that has been defined and
-#   the pin. Possible pin values are AIN0, AIN1, AIN2 and AIN3 for single ended
-#   lines and DIFF01, DIFF03, DIFF13 and DIFF23 for differential between their
-#   correspoding lines. For example
-#   DIFF03 measures the differential between line 0 and 3. Only specific
-#   combinations for the differentials are allowed.
-#pullup_resistor: 4700
-#   A setting that defines the physical wiring method of the sensor.
-#   Default is 4700.
-#voltage_offset: -0.295
-#   The differential voltage from the supplied voltage to the maximum pull-up
-#   voltage that the device under test (thermistor) sees.
-```
-
 ### DS18B20 temperature sensor
 
 DS18B20 is a 1-wire (w1) digital temperature sensor. Note that this sensor is not intended for use with extruders and heater beds, but rather for monitoring ambient temperature (C). These sensors have range up to 125 C, so are usable for e.g. chamber temperature monitoring. They can also function as simple fan/heater controllers. DS18B20 sensors are only supported on the "host mcu", e.g. the Raspberry Pi. The w1-gpio Linux kernel module must be installed.
@@ -4858,7 +4831,7 @@ Digital Converters that can be used for temperature sensors. They provide 4
 analog input pins either as single line or as differential input.
 
 ```
-[ads1x1x my_name]
+[ads1x1x my_ads1x1x]
 chip: ADS1115
 #pga: 6.144V
 #   Default value is 4.096V. The maximum voltage range used for the input. This
@@ -4876,14 +4849,24 @@ chip: ADS1115
 #    ADS111X's support 8, 16, 32, 64, 128, 250, 475, 860.
 i2c_mcu: host
 i2c_bus: i2c.1
-#adc_voltage: 5
-#   The suppy voltage for the device.
-#   Sensor is reading this amount different from external measuring device.
 #address_pin: GND
 #   Default value is GND.  There can be up to four addressed devices depending
 #   upon wiring of the device. Check the datasheet for details. The i2_address
 #   can be specified directly instead of using the address_pin.
+```
 
+The chip provides pins that can be used on other sensors.
+
+```
+sensor_type: ...
+#   Can be any thermistor or adc_temperature.
+sensor_pin: my_ads1x1x:AIN0
+#   A combination of the name of the ads1x1x chip that and the pin. Possible
+#   pin values are AIN0, AIN1, AIN2 and AIN3 for single ended lines and
+#   DIFF01, DIFF03, DIFF13 and DIFF23 for differential between their
+#   correspoding lines. For example
+#   DIFF03 measures the differential between line 0 and 3. Only specific
+#   combinations for the differentials are allowed.
 ```
 
 ### [replicape]
--- a/klippy/extras/ads1x1x.py
+++ b/klippy/extras/ads1x1x.py
@@ -6,7 +6,6 @@
 import logging
 import pins
 from . import bus
-from . import adc_temperature
 
 # Supported chip types
 ADS1X1X_CHIP_TYPE = {
@@ -173,22 +172,10 @@ class ADS1X1X_chip:
         if config.get('address_pin', None) is not None:
             address = config.getchoice('address_pin', ADS1X1X_CHIP_ADDR)
 
-        if isADS101X(self.chip):
-            self.samples_per_second = config.getchoice('samples_per_second',
-                ADS101X_SAMPLES_PER_SECOND, '128')
-            self.samples_per_second_numeric = config.getint(
-                'samples_per_second', 128)
-        else:
-            self.samples_per_second = config.getchoice('samples_per_second',
-                ADS111X_SAMPLES_PER_SECOND, '128')
-            self.samples_per_second_numeric = config.getint(
-                'samples_per_second', 128)
-
         self._ppins = self._printer.lookup_object("pins")
         self._ppins.register_chip(self.name, self)
 
         self.pga = config.getchoice('pga', ADS1X1X_PGA, '4.096V')
-        self.mode = config.getchoice('mode', ADS1X1X_MODE, 'single')
         # Comparators are not implemented, they would only be useful if the
         # alert pin is used, which we haven't made configurable.
         # But that wouldn't be useful for a normal temperature sensor anyway.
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Thanks.  As high-level feedback, this sounds like useful functionality.  I have some comments and questions:

1. I'm confused why this new module ties into the temperature system (and is described in the temperature section of the documentation).  I would have thought that one could add a new adc pin via the `setup_pin()` mechanism, and then users could define temperatures sensors (or adc buttons) using the existing `my_ads_chip:pin1` mechanism.  That is, wouldn't the generic adc pin system be able to handle existing thermistors, analog buttons, and other features?
2. How does this module relate to (if at all) the ads1220 code proposed in #6555 and the ads1118 code proposed in #6553?

-Kevin

KevinOConnor:
I should also note that most temperature sensors in Klipper today are implemented with mcu code that can check the min/max temperature setting and directly invoke a shutdown if the range is exceeded.  This code would be a different in that only the non-realtime host python code would check the min/max temperature range.  I don't think that is a "blocker" but something to be aware of.

-Kevin

korsarNek:
1. Users can define temperature sensors with the my_ads_chip:pin1 mechanism. I have one definition for the chip in general that controls stuff like the i2c definition and then there is a separate definition for temperature sensors where the sensor_pin and sensor_type get used with the option to add a couple extra config values. The chip object also provide a setup_pin function. I thought that is the same system as the other temperature sensors.
The separation of chip, sensor and pin and not including it all in the sensor is one of the things I've changed compared to the original PR where everything was defined on the sensor directly.
Do you maybe have an example where I can see how it is supposed to look like?
2. The ADS1118 is very similar to the ADS chips I've implemented here and support the same state machine, the big difference would just be that SPI is used instead of I²C. It could make sense to try to unify them.
Concerning the ADS1220, from skimming through the data sheet, it looks like there are too many differences between that and the other ADS chips to reasonably try to unify them.

korsarNek:
I'll check the min max temperature part.

korsarNek:
Actually, the ADS1118 has a similar format, but supports a couple different commands. Maybe they are also different enough that it wouldn't make too much sense sharing the code.

KevinOConnor:
> I have one definition for the chip in general that controls stuff like the i2c definition and then there is a separate definition for temperature sensors where the sensor_pin and sensor_type get used with the option to add a couple extra config values.

Can you elaborate on when/why a user would use `sensor_type: ads1x1x` instead of just using `sensor_pin: my_ads1x1x:AIN0` with an existing sensor_type?

> Do you maybe have an example where I can see how it is supposed to look like?

I don't know of a "one-to-one" example.  You can look at `klippy/extras/adc_scaled.py` and `config/generic-duet2-maestro.cfg` for an example of a pseduo-adc chip.  You can look at `klippy/extras/sx1509.py` and `config/generic-duet2-duex.cfg` for an example of an i2c digital gpio expander chip.

-Kevin

KevinOConnor:
Also, as recently discussed in #6553, src/thermocouple.c and klippy/extras/spi_temperature.py could also be looked at as an example - though as mentioned there it may not be the best example.

-Kevin

korsarNek:
```
[ads1x1x adc]
chip: ADS1115
pga: 6.144V
samples_per_second: 16
i2c_mcu: host
i2c_bus: i2c.1
address_pin: GND
adc_voltage: 5

[temperature_fan chamber_fan]
pin: host:gpio17
min_temp: 0
max_temp: 80
max_power: 0.8
off_below: 0.1
shutdown_speed: 0
control: watermark
sensor_type: ads1x1x
sensor_pin: adc:AIN0
probe_type: Generic 3950
gcode_id: C
target_temp: 0
voltage_offset: -0.2
pullup_resistor: 4700
```

That's the config I'm using for my setup. There are sensors like HTU21D, where this would not be compatible with, they define their own communication protocol and would be incompatible with the ADS just reading out the voltage on the pin. I thought that was what you originally meant with sensor_type, but there are other sensor_types which are equivalent to the probe_type here. I guess I could have a look in how to make the thermistor and adc_temperature objects talk with the ads chip object to do the voltage to temperature conversion, then the probe_type property would not be necessary anymore and would become the sensor_type instead.

KevinOConnor:
I'm not sure I understand.

To take a step back, on one of my printers, I have the following:
```
[heater_bed]
heater_pin: PC9
sensor_type: NTC 100K MGB18-104F39050L32
sensor_pin: PC4
control: pid
min_temp: 0
max_temp: 130
pid_Kp: 57.006
pid_Ki: 3.334
pid_Kd: 243.700
```

I guess my question is, if I had an ads1115, could I just do:
```
[ads1x1x my_adc]
chip: ADS1115
pga: 6.144V
samples_per_second: 16
i2c_mcu: host
i2c_bus: i2c.1
address_pin: GND
adc_voltage: 5

[heater_bed]
heater_pin: PC9
sensor_type: NTC 100K MGB18-104F39050L32
sensor_pin: my_adc:AIN0
control: pid
min_temp: 0
max_temp: 130
pid_Kp: 57.006
pid_Ki: 3.334
pid_Kd: 243.700
```

That is, if `setup_pin()` is implemented and returns a value between 0.0 and 1.0, would the existing temperature code work as is?

-Kevin

korsarNek:
As the code currently is, that would not work, the pin is missing the setup_minmax function. Apparently that's needed for an adc pin. Additionally to that, I have the code for requesting the voltage currently on the sensor instead of the pin, I'll have to move some stuff around from the sensor to the pin.

A config that currently would work is this:

```
[ads1x1x my_adc]
chip: ADS1115
pga: 6.144V
samples_per_second: 16
i2c_mcu: host
i2c_bus: i2c.1
address_pin: GND
adc_voltage: 5

[heater_bed]
heater_pin: PC9
>sensor_type: ads1x1x<
>probe_type: NTC 100K MGB18-104F39050L32<
sensor_pin: my_adc:AIN0
control: pid
min_temp: 0
max_temp: 130
pid_Kp: 57.006
pid_Ki: 3.334
pid_Kd: 243.700
```

When I move the code around instead of using probe_type, it should be possible to use the NTC for the sensor_type instead.

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,431 @@
...
+        else:
+            self.samples_per_second = config.getchoice('samples_per_second',
+                ADS111X_SAMPLES_PER_SECOND, '128')
+            self.samples_per_second_numeric = config.getint(
+                'samples_per_second', 128)
+
+        self._ppins = self._printer.lookup_object("pins")
+        self._ppins.register_chip(self.name, self)
+
+        self.pga = config.getchoice('pga', ADS1X1X_PGA, 'V4.096')
```
jackwakefield, line 193: `V4.096` should be `4.096V`

docs/Config_Reference.md
```diff
@@ -4606,6 +4633,41 @@ vssa_pin:
...
+[ads1x1x my_name]
+chip: ADS1115
+#pga: 6.144V
+#   Default value is 4.096V. The maximum voltage range used for the input. This
+#   scales all values read from the ADC. Options are: 6.144V, 4.096V, 2.048V,
+#   1.024V, 0.512V, 0.256V
+#mode: single
+#    Default value is single. Turn off the chip after a single reading or keep
+#    it running. Turning off saves power but turning it back on will be slower.
+#    Options are single and continous.
```
jackwakefield, line 4652: `continous` should be `continuous`

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,431 @@
...
+ADS101X_PGA_SCALAR = {
+    0x0000: 6.144 / 2047.0,  # +/-6.144V range = Gain 2/3
+    0x0200: 4.096 / 2047.0,  # +/-4.096V range = Gain 1
+    0x0400: 2.048 / 2047.0,  # +/-2.048V range = Gain 2
+    0x0600: 1.024 / 2047.0,  # +/-1.024V range = Gain 4
+    0x0800: 0.512 / 2047.0,  # +/-0.512V range = Gain 8
+    0x0A00: 0.256 / 2047.0  # +/-0.256V range = Gain 16
+}
+ADS1X1X_MODE = {
+    'continous': 0x0000,  # Continuous conversion mode
```
jackwakefield, line 109: `continous` should be `continuous`

jackwakefield:
Thanks for your work on this! I'm running into a couple of issues.

The first is, when using all 4 ADC pins, the readings are jumping from one pin to another. For example, AIN0 might be 35c, and AIN2 50c, but then it might suddenly show both AIN0 and AIN2 as 50c, and then next AIN2 as 35c and AIN0 as 50c.

The second issue is (and it might be a wiring issue on my end), I'm seeing `ADS1X1X: command queue is overflowing, dropping requests` in klippy.log. I do have 2 ADS1115's, and can see one is noticebly slower than the other. One will begin reading temperatures immediately, while the other is showing 0c for several seconds, and then refreshes about 10x slower than the other ADS1115. Again this might be an issue on my end, and I'll let you know if I find anything.

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,431 @@
...
+        if self._command_queue:
+            command = self._command_queue.pop()
+            pin_object = self._pins[command['sensor'].pin]
+            sample = 0
+            try:
+                self._write_register(ADS1X1X_REG_POINTER['CONFIG'],
+                                     pin_object.config)
+
+                self._reactor.pause(self._reactor.monotonic() + 1.0 /
+                                    (self.samples_per_second_numeric + 1))
```
jackwakefield, line 255: Am I right in thinking this wouldn't pause for a long enough time for the ADC conversion to complete, based on `1.0 / (self.samples_per_second_numeric + 1)`?

For example, if we were taking 128 samples a second, then the ADC delay would be 7.8125ms per sample, but `1.0 / (128 + 1)` would be roughly 7.75ms.

Would it make sense to change this to `1.0 / self.samples_per_second_numeric + 0.001`?

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,431 @@
...
+
+                sample = self._read_register(ADS1X1X_REG_POINTER['CONVERSION'])
+            except Exception:
+                logging.exception("ADS1X1X: error while sampling")
+                return self._reactor.NEVER
+
+            command['callback'](sample)
+
+        return self._reactor.monotonic() + 1.0 / self.samples_per_second_numeric
```
jackwakefield, line 264: I believe if we're taking samples from multiple ADC pins per sensor then this waketime would cause issues.

Is this specific delay needed if we're pausing above before reading the ADC value?

klippy/extras/ads1x1x.py
jackwakefield: According to the spec, bit 15 is a operational status pin which we can check to see whether the value's available.

> For a read status:
> 0 : Device is currently performing a conversion
> 1 : Device is not currently performing a conversion

Perhaps it's worth treating the delay as a timeout instead, and polling the config to see whether bit 15 is 1?

jackwakefield: I've tested this and it works very well, solving the first issue I describe in https://github.com/Klipper3d/klipper/pull/6584#issuecomment-2240574284.

https://github.com/jackwakefield/klipper/commit/72e923761c65a357adadb823f5f40df697abd8b8#diff-a00870c95aa335ed2e7afd38885cc50c69987c1d819811014a80c380f85bdd95R246

korsarNek: The intent was to have a little extra delay, that's why added the +1 on the samples, but it should've been a -1 actually. I saw you are using a fixed delay now, the values seem like they are related to the default sample count for that device, not the configured one.
ADS1115 default is 128, 1 / 128 ~= 0.008
ADS1015 default is 1600, 1 / 1600 ~= 0.001
Theoretically, the sample count should relate to how long the operation takes, I'm gonna test it out, but I think instead of the fixed delay we'll go back to 1 / sampleCount but with a better margin.

Checking the conversion bit seems like a fine adjustment, an extra layer of security here is nice.

I'll have to check some stuff, but I would then cherry-pick your commit over.

klippy/extras/ads1x1x.py
korsarNek: I'll try to check it, but I think you might be right. That looks like an unnecessary double wait.

klippy/extras/ads1x1x.py
korsarNek: I've now updated the code, seems to be working fine. On my raspberry pi 3b, the sample rate seems to cap out somwhere at around 80 fps, but lower values seem to be working fine, too.

korsarNek:
@KevinOConnor Concerning the sensor setup, I have a question on how this could be solved.

image

The problem is, the pullup resistor is technically not part of the chip, but when someone would use the thermistor, there has to be a little circuit that could theoretically be different every time. It would be unusual for the pull up resistor to be different at different parts of the machine, but the voltage offset might be different, if you want an accurate reading, a different length of the wire could result in a different voltage offset that someone might want to account for.

On the pin, I don't have access to that information. The pins provided by the chip object, can also only be really used in connection with the ads1x1x sensor object, at the moment, they are not general pins that can be reused in other contexts.

Is there a way to have the thermistor as the sensor_type under these circumstances? I don't see how I could make that work with the current architecture.

> I'm not sure I understand.
> 
> To take a step back, on one of my printers, I have the following:
> 
> ```
> [heater_bed]
> heater_pin: PC9
> sensor_type: NTC 100K MGB18-104F39050L32
> sensor_pin: PC4
> control: pid
> min_temp: 0
> max_temp: 130
> pid_Kp: 57.006
> pid_Ki: 3.334
> pid_Kd: 243.700
> ```
> 
> I guess my question is, if I had an ads1115, could I just do:
> 
> ```
> [ads1x1x my_adc]
> chip: ADS1115
> pga: 6.144V
> samples_per_second: 16
> i2c_mcu: host
> i2c_bus: i2c.1
> address_pin: GND
> adc_voltage: 5
> 
> [heater_bed]
> heater_pin: PC9
> sensor_type: NTC 100K MGB18-104F39050L32
> sensor_pin: my_adc:AIN0
> control: pid
> min_temp: 0
> max_temp: 130
> pid_Kp: 57.006
> pid_Ki: 3.334
> pid_Kd: 243.700
> ```
> 
> That is, if `setup_pin()` is implemented and returns a value between 0.0 and 1.0, would the existing temperature code work as is?
> 
> -Kevin

KevinOConnor:
Ah, the simple way to handle that is to not implement `voltage_offset` nor `adc_voltage` in ads1x1x.py .  The existing thermistor (and adc_temperature) code already handles all that logic.

So, something like:
```
[ads1x1x my_adc]
chip: ADS1115
i2c_mcu: host
i2c_bus: i2c.1

[heater_bed]
heater_pin: PC9
sensor_type: NTC 100K MGB18-104F39050L32
sensor_pin: my_adc:AIN0
control: pid
min_temp: 0
max_temp: 130
pid_Kp: 57.006
pid_Ki: 3.334
pid_Kd: 243.700
```

Said another way, just invoke the adc callback with the raw adc value (eg, see `mcu.py:_handle_analog_in_state()`) and let the existing Klipper code convert that to a voltage/temperature.

-Kevin

legend069:
will this PR include support for a ADC chips connected to the printer MCU?

korsarNek:
> will this PR include support for a ADC chips connected to the printer MCU?

@legend069 I did not test a setup with an ADC chip connected to the printer MCU. I have the I2C pins needed for this connected to my linux mcu. From my understanding, Klipper should abstract a proper printer MCU and a linux MCU enough that they should behave the same.

My setup:
Hintergrund

The setup that I think, you mean:
Hintergrund2

Both should be working fine.

What's a bit special about this implementation is that most sensors have their loop to check if the temperature is out of bounds on the mcu, this implementation has it only in the python code. It might not be quite as real time as other sensors.

I'm currently using the chip only to measure the chamber temperature, so that setup is sufficient for me.

---

For the final version the config will change a bit. I've already tested a new version that doesn't use sensor_type: ads1x1x anymore but allows using the normal thermistors. The pushes from yesterday were just a rebase, the code wasn't quite compatible with the most recent version of klipper. I'll probably finish it up next weekend and then push the newest version.

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,358 @@
...
+                # possible inputs. So sample_count 16 on 1 input will result
+                # in 4 samples per second.
+                delay = 1 / self.samples_per_second_numeric
+                self._reactor.pause(self._reactor.monotonic() + delay)
+                while not self.is_ready():
+                    self._reactor.pause(self._reactor.monotonic() + 0.001)
+                sample = self._read_register(ADS1X1X_REG_POINTER['CONVERSION'])
+            except Exception:
+                logging.exception("ADS1X1X: error while sampling")
+                return None
```
KevinOConnor, line 249: Can you provide a little more explanation for this error handling?  If I understand the code correctly, if there is some error in this interaction the code will just stop sending temperature updates to the higher-level code.  That sounds like that could cause a subtle failure (the higher level code will just assume the temperature hasn't changed for potentially extended periods).  I'm not sure what the best solution is, but perhaps invoking a shutdown should be considered.

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,358 @@
...
+        self.invalid_count = 0
+
+        self.chip._printer.register_event_handler("klippy:connect", \
+                                            self._handle_connect)
+
+    def _handle_connect(self):
+        self._reactor = self.chip._printer.get_reactor()
+        self._sample_timer = \
+                self._reactor.register_timer(self._timer)
+        self._reactor.update_timer(self._sample_timer, self._reactor.NOW)
```
KevinOConnor, line 312: FYI, you can do these two things directly with `reactor.register_timer(self._timer, reactor.NOW)`

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def _timer(self, eventtime):
+        sample = self.chip.sample(self)
+        if sample is not None:
+            self._process_sample(sample)
+
+        return self._reactor.monotonic() + self.report_time
```
KevinOConnor, line 319: It seems odd that this does not use `eventtime + self.report_time`.  Rescheduling based on the current time will result in an ongoing drift in the period between readings.

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,358 @@
...
+    def _process_sample(self, sample):
+        # The sample is encoded in the top 12 or full 16 bits
+        # Value's meaning is defined by ADS1X1X_REG_CONFIG['PGA_MASK']
+        if isADS101X(self.chip.chip):
+            sample >>= 4
+            target_value = sample / ADS101X_RESOLUTION
+        else:
+            target_value = sample / ADS111X_RESOLUTION
+
+        if self.maxval > self.minval:
```
KevinOConnor, line 330: The calling code should verify `maxval > minval` and always provide values - so this looks like an odd check.

klippy/extras/ads1x1x.py
korsarNek: I could adjust it so that a failed sampling adds to the invalid_count and invokes a shutdown if it happens often enough in a row.

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,358 @@
...
+                raise pins.error(
+                    '%s pin %s for chip %s is used multiple times' \
+                        % (self.chip, pin.pin, self.name))
+            self._pins[pin.pin] = pin
+            return pin
+        raise pins.error('Wrong pin or incompatible type: %s with type %s! ' % (
+            pin_params['pin'], pin_type))
+
+    def _handle_ready(self):
+        self.reset_all_devices()
```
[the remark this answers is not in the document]
korsarNek: Sounds good, I wasn't quite sure which one was the correct one for this.

klippy/extras/ads1x1x.py
```diff
@@ -0,0 +1,358 @@
...
+        self.comp_queue = ADS1X1X_COMPARATOR_QUEUE['QUEUE_NONE']
+        self._i2c = bus.MCU_I2C_from_config(config, address)
+
+        self.mcu = self._i2c.get_mcu()
+
+        self._printer.add_object("ads1x1x " + self.name, self)
+        self._printer.register_event_handler("klippy:ready", \
+                                            self._handle_ready)
+        self._printer.register_event_handler("klippy:shutdown", \
+                                            self.reset_all_devices)
```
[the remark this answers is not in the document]
korsarNek: I'll remove it, reset on initialize should suffice anyway.

klippy/extras/ads1x1x.py
korsarNek: Ah, I wasn't sure if it does for certain, so I just added it to be safe. I'll remove that part.

docs/Config_Reference.md
```diff
@@ -4824,6 +4824,51 @@ vssa_pin:
...
+#mode: single
+#    Default value is single. Turn off the chip after a single reading or keep
+#    it running. Turning off saves power but turning it back on will be slower.
+#    Options are single and continuous.
+#samples_per_second: 128
+#    Default value is 128. The amount of samples that the ADC can provide per
+#    second. A lower value makes the samples more accurate, but it takes longer
+#    until a new value is available.
+#    ADS101X's support 128, 250, 490, 920, 1600, 2400, 3300.
+#    ADS111X's support 8, 16, 32, 64, 128, 250, 475, 860.
```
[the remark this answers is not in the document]
korsarNek: Someone might want to change it for higher accuracy, how many samples_per_second you can get away with also depends on how many inputs of the chip are used as one sampling would have to be done after the other.

It would make more sense to try to calculate that automatically, someone would need to know too much about how klipper and the ADS chip works to choose a good value.

A report_time in setup_adc_callback of for example 0.010s, would be 100 samples per second. On a single input I would choose 128. If 4 inputs are used, I would need to choose 920 (128 X 4 = 512, next best is 920) for ADS101X and 860 for ADS111X, reducing the accuracy but fitting in the time frame. I would then output a warning if the sample_rate needs to be higher than the chip can support.

It's a bit of extra logic finding the next best value and counting the used inputs, but then I could kick out the samples_per_second.

Concerning the single or continous mode, there is a theoretical power saving on single mode. I just wanted to include it for completeness.

If I compute the sample_rate anyway, this could be hard coded to continous.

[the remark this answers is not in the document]
KevinOConnor: Okay.  Just to be clear, my comments here are not "blockers" to merging.

As for "power savings" - I can't see anyone caring, so it sounds like continuous is a better default and one can use a code constant to make it possible to enable sleep mode.

As for auto-calculating wakeup times - I'll defer to you if that's something worthwhile to implement.

klippy/extras/ads1x1x.py
KevinOConnor: Sounds good to me.

docs/Config_Reference.md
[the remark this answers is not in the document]
korsarNek: I have now added a calcuation for the sample rate. I made single mode the default though, it turns out that the idle state that we check on to see if the sampling is done, only gets set on single mode, not on continuous mode.

KevinOConnor:
Thanks.  Seems fine to me.  Let me know when you think it is ready to commit.

I did notice a couple of minor things in the documentation.  See below.

-Kevin

docs/Config_Reference.md
```diff
@@ -4824,6 +4824,50 @@ vssa_pin:
...
+analog input pins either as single line or as differential input.
+
+Note: Use caution if using this sensor to control heaters. The heater min_temp
+and max_temp are only verified in the host and only if the host is running and
+operating normally. (ADC inputs directly connected to the micro-controller
+verify min_temp and max_temp within the micro-controller and do not require a
+working connection to the host.)
+
+The sample rate of the chip gets automatically calculated based on the number of
+pins used and the refresh rate requested by klipper.
```
KevinOConnor, line 4840: FWIW, seems a little odd to describe parameters the user doesn't need to configure - it might be confusing.

docs/Config_Reference.md
```diff
@@ -4824,6 +4824,50 @@ vssa_pin:
...
+chip: ADS1115
+#pga: 4.096V
+#   Default value is 4.096V. The maximum voltage range used for the input. This
+#   scales all values read from the ADC. Options are: 6.144V, 4.096V, 2.048V,
+#   1.024V, 0.512V, 0.256V
+i2c_mcu: host
+i2c_bus: i2c.1
+#address_pin: GND
+#   Default value is GND.  There can be up to four addressed devices depending
+#   upon wiring of the device. Check the datasheet for details. The i2_address
```
KevinOConnor, line 4853: "i2c_address"?

docs/Config_Reference.md
```diff
@@ -4824,6 +4824,50 @@ vssa_pin:
...
+#   can be specified directly instead of using the address_pin.
+```
+
+The chip provides pins that can be used on other sensors.
+
+```
+sensor_type: ...
+#   Can be any thermistor or adc_temperature.
+sensor_pin: my_ads1x1x:AIN0
+#   A combination of the name of the ads1x1x chip that and the pin. Possible
```
KevinOConnor, line 4863: Extra "that"?

comradef191:
In anticipation of this PR being merged, a few weeks ago I designed a set of ADS1115 prototype boards and have been testing them using various versions of this PR over the past ~2 weeks.

_While I have a GH link there, I haven't yet published that repo as I'm still testing the boards at the moment_

Electrically, my boards checkout with the correct voltage & resistance measured across my thermistor.
image
_Edited from actual schematic for clarity, hence missing designators & only 1 thermistor_

In Klipper, however, what I'm seeing is entirely different from what i SHOULD be seeing.
With the PGA set to 4.096, and a room temperature of about ~22°C, my ADS1115s read 67°C without a Thermistor, and 71.4°C with a TL 104NT. This is with the 4.7kΩ voltage divider & 3.3V voltage source as pictured above - This same layout wired directly to a MCU is working perfectly fine for all my other thermistors.
image

_____
After checking a few dozen times with various versions of this PR, trying to run my ADS1115s off the rPis GPIO instead of an STM32F401, and trying with several different resistors, I went and spent a few hours talking with my friend @dead.lock trying to get to the bottom of what I'm seeing, with them concluding the following:

> Since the ADS1115 has an internal voltage reference for the ADCs instead of using external reference, the suspected source of the behaviour is lack of definition of thermistor pull-up voltage. The reasoning is following:
> At 22°C the thermistor used (104NT-4-R025H42G) has a nominal resistance of 115.3kΩ, with a 4.7kΩ pull-up resistor and 3.3V pull-up voltage, the output of the resistive divider is 3.17V, which was confirmed with a multimeter.
> Since the ADS PGA was set for ±4.096V the reading is 0.7741 of full scale of ADC.
With 4.7kΩ pull-up resistor, resistive divider giving 0.7741 of its full scale would have 16.11kΩ pull-down resistor.
> Correlation with thermistor chart shows that it has 16.11kΩ at 69°C.
> Hence it's deemed plausible that the absolute voltage reading given by the ADC is treated by the firmware as relative, as the mathematical model shown above is reflected by both readings shown by Klipper and code analysis.

korsarNek:
@comradef191 Hi, could you add the configuration for one of the sensors that worked and one that didn't? Together with the related ADS configuration. That would remove some guess work.

korsarNek:
@KevinOConnor I've now updated the documentation.

Let's see if @comradef191 adds some more context, then I'll have a look what's going on there, otherwise it should be fine to merge.

comradef191:
> @KevinOConnor I've now updated the documentation.
> 
> Let's see if @comradef191 adds some more context, then I'll have a look what's going on there, otherwise it should be fine to merge.
(sorry for delayed reply- SMRRF & college held me up!)

What details would you like?

The thermistors I'm using for testing are known-good, verified as reading an accurate room temperature both on a normal mainboard (SKR Pico), and when connected directly to the MCU im using with an identical PCB layout & schematic for the voltage dividers (quite-literally copy-pasted between the two boards.)
I see no improvement by using a variety of different Thermistors including:
- The 104NT on a TL CHCPro
- An E3D 104NT
- Multiple separate TL 3950s
- A TL 104NT

The MCU i am using for testing is an STM32F401CDU6 BlackPill, with four ADS1115s connected via hardware I2C bus 1.
Each ADS is wired up so that all have unique I2C addresses, per the datasheet. I am able to get all four to display a reading at once just fine.

The ADS' run off an extremely stable 3V3 source, which is entirely separate from the on-board supply from the BlackPill. This 3V3 source is also used for the  4.7kΩ pull-up resistors on the thermistors ports.
Each ADS' has four thermistor ports wired up to it identically per the earlier schematic & picture.

I see no improvement in what the thermistors read no matter what I set the PGA to.
I see no change in what the thermistors read when I use a set of jumper wires to connect it directly to my rPis GPIO I2C & 3V3 pins and use Host MCU to control it. 
While doing the above, I saw no improvement by running the board off of 5V power instead of 3V3 even when I fiddled with PGA.

When I probe the board with my Multimeter:
- 3V3 source is accurate and stable
- only 0.1Ω (the lowest my meter goes) between the between the pin of the thermistor port & corresponding pin of the ADS
- All 4.7kΩ resistors within tolerance
- Capacitors within tolerance

KevinOConnor:
Thanks.  The PR looks fine to me.  I can't comment on the testing results, so @korsarNek let me know when you think I should commit.

-Kevin

recrudesce:
How can you realistically accept a PR to a mainline 3d printer firmware without testing it based off comments from others saying there's an issue with it ?

korsarNek:
> What details would you like?

@comradef191 I would've liked to see the klipper config.

> @comradef191 Hi, could you add the configuration for one of the sensors that worked and one that didn't? Together with the related ADS configuration. That would remove some guess work.

Anyway, I've pushed another version that supports a new config entry. Additionally to the pga, you can also set the adc_voltage.

The thermistors expect the adc readings they receive to be between 0 and 1. Some of them offer options to provide an offset, like voltage_offset or their own adc_voltage, but that is just for voltage based sensors (adc_temperature). Resistance based sensors don't really offer a mechansim for that. With the added adc_voltage on the ads1x1x you can provide an offset that helps rescale the values between 0 and 1.

A pga of 4.096 and an adc_voltage of 3.3 remaps a voltage of for example 3 from ~0.73 (3 / 4.096) to ~0.91 (3 / 3.3). That should account for the unexpected temperature you were seeing.

comradef191:
I am able to confirm that by using `adc_voltage: 3.3` temperatures are now in-line with what should be expected. I suggest defaulting to 3.3 unless otherwise specified, since its already what most-everything uses with Thermistors.

image
Temperature readings are stable and noise-free, and quickly track any changes cause by squeezing on the thermistor or such.
When taped to the bottom of my heated bed, two seperate thermistors are accurately tracking its 60c temperature within a few degrees of the silicone mats integrated thermistor.

Working config used for this testing, with four ADS' on one bus, each with one Thermistor:
[thermex16.cfg.txt](https://github.com/user-attachments/files/18386136/thermex16.cfg.txt)

legend069:
I'll be happy to test this, will plan on connecting the adc sensors to the btt octopus

Has anyone got a reference config I could use ?

korsarNek:
I've adjusted the default value, makes sense to have it 3.3V.

korsarNek:
@legend069 feel free to test this, but 3 people already tested this so I think it works fine now. comrade and I have added example configs for the ADS in the ticket if you mean that.

korsarNek:
@KevinOConnor From my side it is now fine to merge. You should have another look over the code, if it is fine for you, too, it can be merged.

KevinOConnor:
Thanks.

-Kevin

PR #6585, opened 2024-05-02
STM32 module fixes to support Stm32g474 with FDCAN.

Dropeffect:
This pull request adds support for `stm32g474` of the evaluation board [NUCLEO-G474RE](https://www.st.com/en/evaluation-tools/nucleo-g474re.html), The PR mainly Fixes the following.

  1. ADC3 channel common interface registers should use the ADC345 interface, The original code does not compile for `stm32g474` which has the ADC3 channel, and gives the following error:
  ```
src/stm32/stm32h7_adc.c: In function 'gpio_adc_setup':
src/stm32/stm32h7_adc.c:192:22: error: 'ADC3_COMMON' undeclared (first use in this function); did you mean 'ADC345_COMMON'?
         adc_common = ADC3_COMMON;
                      ^~~~~~~~~~~
                      ADC345_COMMON
src/stm32/stm32h7_adc.c:192:22: note: each undeclared identifier is reported only once for each function it appears in
make: *** [Makefile:65: out/src/stm32/stm32h7_adc.o] Error 1
   ```

  2. Allow the use of FDCAN2 needed for "PB12, PB13" and "PB5, PB6" alternate pins defined in Klipper Kconfig. without the applied changes the original code will mistakenly route the IRQ to `FDCAN1_IT0_IRQn`, and the FDCAN peripheral clock will not be enabled making writing to FDCAN registers ineffective.

  3. Add support for `stm32g474cc` in `stm32` Kconfig.

The code is tested with both CAN and USB communication channels on `STM32G474CC` as a secondary MCU responsible for extrusion, regression tests and whitespace checks passed successfully.
We are potentially developing a product around this MCU(`STM32G474CC`). We think this processor holds a lot of potential and it would be great if you could add support for it.

Thanks,
Amr.

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -88,6 +88,9 @@ choice
     config MACH_STM32G431
         bool "STM32G431"
         select MACH_STM32G4
+    config MACH_STM32G474
+        bool "STM32G474"
+        select MACH_STM32G4
     config MACH_STM32H723
         bool "STM32H723"
         select MACH_STM32H7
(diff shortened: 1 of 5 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #6587, opened 2024-05-03
Add Artillery Genius Pro config

jayofelony:


JamesH1978:

Also please refer to point 4 in the guidelines in the following link and add the printers.test with your new file for regression testing https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Thanks
James

JamesH1978:
Please also refer to the example config document for what to include and what not to include in an example configuration. You have no header explaining flash instructions, and you have included superfluous macros and you have commented out PID sections in your heaters. I will do a full review once you have added the signed off by line and the regressions test.

Thanks
James

Issue #3468, opened 2020-10-25
Feature Request: Sort vsdlist by creation date


dushyantahuja:
Hi, I've seen that this was rejected - https://github.com/KevinOConnor/klipper/pull/1387/commits/c7cf8c5095a8bb744a0223eb299400efeaa40d1f
However, is it possible to relook at this - I have a lot of gcode files, and need to scroll down on the LCD menu to find the latest one that has been uploaded. Sorting by creation date would be really helpful, and make the menu more usable.

klipper-gitissuebot:
Hi @dushyantahuja,

It did not look like there was a Klipper log file attached to this ticket. The log file has been engineered to answer common questions the Klipper developers have about the software and its environment (software version, hardware type, configuration, event timing, and hundreds of other questions).

Unfortunately, too many people have opened tickets without providing the log. That consumes developer time; time that would be better spent enhancing the software. If this ticket references an event that has occurred while running the software then the Klipper log must be attached to this ticket. Otherwise, this ticket will be automatically closed in a few days.

For information on obtaining the Klipper log file see: https://github.com/KevinOConnor/klipper/blob/master/docs/Contact.md

The log can still be attached to this ticket - just add a comment and attach the log to that comment.

Best regards,
~ Your friendly GitIssueBot

PS: I'm just an automated script, not a human being.

dushyantahuja:
@KevinOConnor Please let me know if this can be included - adds a slight complexity to the config file (only for those that want it), but would really help out on small LCDs.

KevinOConnor:
I have no preference here.  If someone wants to implement it, it's fine with me.  I'd say it should be done in the display code as part of specifying the menu items.

-Kevin

PR #6589, opened 2024-05-06
Implement sorting options for vsdlist menu items

nmaggioni:
Solves #3468 by reimplementing the logic removed in c7cf8c5 in the display code instead of the virtual sdcard itself, as requested in the issue.

Adds the new `sort` and `reverse` options to menu entries of the `vsdlist` type.


howdy-vulcan:
I am an complete amateur to this, but was able to to edit my local RPI Klipper code & printer configuration via Win SCP + Virtual Studio today. I copied and pasted these edits in and **confirmed it works great** on my Voron V0 screen. **Thanks for making this! It really improves how I use my printer!**  Also for anyone else who sees this i'd recommend adding the month&date at the start of the exported gcode filename in your slicer configuration.  I also made a backup of the RPI using the Unbutu virtual machine and  "clone and compress raspberry pi image" video by  techsapphire, didnt need to use it, but great to have.

nmaggioni (commented):
This PR adheres to [the specs](https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review).

MarSoft:
Please update `p2cmd_O50` in `klippy/extras/palette2.py`: it expects `virtual_sdcard.get_file_list` to return tuples of 2 items.

nmaggioni:
Done, let me know if there are any hiccups.

MarSoft:
Unfortunately I don't use that module, I just found that it uses `get_file_list` method by using a simple `grep`.
Probably we should instead modify `get_file_list` to accept an optional `with_mtime` parameter? With it set to True it would return list of 3-tuples, otherwise (by default) 2-tuples? This would avoid breaking any external code.

nmaggioni:
I think that would introduce unneeded complexity as people would need to look into `get_file_list` itself when using it instead of safely relying on existing examples. After all pulling the mtime in only requires a `stat`-like call for each file (especially cheap on flash based systems) and if the third parameter isn't needed one can simply ignore it after unpacking the tuple.

klippy/extras/virtual_sdcard.py
```diff
@@ -76,16 +76,21 @@ def get_file_list(self, check_subdirs=False):
                     full_path = os.path.join(root, name)
                     r_path = full_path[len(self.sdcard_dirname) + 1:]
                     size = os.path.getsize(full_path)
-                    flist.append((r_path, size))
+                    mtime = os.path.getmtime(full_path)
+                    flist.append((r_path, size, mtime))
```
kageurufu, line 80: Instead of calling both `os.path.getsize` and `os.path.getmtime` and returning both, could we not use `os.stat` and return `(filename, stat_result)`? 

Then in M23, just do

```
for fname, fstat in files:
    gcmd.respond_raw("%s %d" % (fname, fstat.st_size))
```

Nothing else uses the size, and thus wouldn't be affected, and then VSDList would have easy access to both size and mtime, plus any additional fields that could be useful in the future

nmaggioni: We could very well do that, but before moving more (currently unneeded) data around the stack I'd like to get the input of a maintainer on this PR in general. There's no point in reducing syscall invocations if there's no interest in merging this anyway.

Also, since this code runs on the host and typically accesses flash-based media with read only queries, I'd favour code readability over premature optimisation.

PR #6592, opened 2024-05-07
klippy: add stallguard monitoring tool

nefelim4ag:
Hi, as https://github.com/Klipper3d/klipper/pull/6139 is merged, I can now show the stall guard measurement stuff.
I'm not sure that it is the perfect solution to poll without batch processing, and not sure about timings/frequency.
I try to mimic ADXL345 code, to avoid adding any new complexity.
My [discourse thread](https://klipper.discourse.group/t/tmc-coolstep-and-how-to-make-it-works) with more graphs, and some comments.

So, there is the problem that I'm trying to solve:
Coolstep is simple and hard to configure because there is no near real-time visibility from the user side.
The original TMC solution is to directly reproduce and measure your load.
We can't see Stallguard pin output outside of homing actions, to track stalls.
But we can poll it.

*And there are many different printers and half of them now are CoreXY, which makes things tricky here.

High-level solution:
- Monitor driver for status
- Store History with stallguard/speed/current data.
- Make it human-readable.

For now, I added the command: `MEASURE_STALLGUARD STEPPER=<stepper>`
It will start to poll/filter/store data, and on the second call - store them in csv.
```
cat /tmp/stepper_y-20240515_024227.csv 
#time,velocity,sg_result,csactual
43668.927338,191.3,67,30
43668.952901,203.8,67,28
43668.977954,203.8,67,28
43669.003832,203.8,67,28
43669.363363,203.8,74,27
```

There is a dump script for data processing:
```
scripts/calibrate_stallguard.py /tmp/stepper_y-20240512_010858.csv -o ~/klipper/stepper_y-20240512_010858.png
```
Example output for one hour print with CoolStep enabled, unfiltered, cumulative graph.
```
~/klipper/scripts/calibrate_stallguard.py stepper_y-20240515_004650.csv -o stepper_y-20240515_004650_cumulative_unfiltered.png -e 100000
```
image
As you can see there is crazy noise on low velocities, also there are crazy speeds calculated from step timing - possibly because of step compression.

Let's filter that, and analyze it together. (min 1 mm/s, max 200 mm/s)
```
~/klipper/scripts/calibrate_stallguard.py stepper_y-20240515_004650.csv -o stepper_y-20240515_004650_cumulative_low.png -s 1 -e 200
```
image
There is StallGuard noise below ~35 mm/s - not enough back EMF.
The suggested SG range for StallGuard2 is in range 0..100, assuming there is more or less correct SGT (-10..0..10) value (mine is 1 or 2, depending on the motor).
We can track min & mean SG value, till they stabilize.
Min becomes zero on speeds above ~125 mm/s (> 3 RPS)
Mean becomes acceptable slightly early at ~80 mm/s. (2 RPS)
*CS is 31 on TMC5160 with reasonable sense resistor and equal CS max here.
The minimum threshold for Coolstep is 32, because SEMIN must be > 0, SEMIN = 1, (SG = SEMIN * 32)

So, I picked up the value 150 mm/s as a threshold with a safe margin, and I know my printer is limited to 1000 mm/s.
(let's just ignore high-speed diagonal moves here)

So, the last one with adequate range.
```
~/klipper/scripts/calibrate_stallguard.py stepper_y-20240515_004650.csv -o stepper_y-20240515_004650_cumulative.png -s 130 -e 1000
```
image
Mean SG is around ~60, if SG is above 64 current can be reduced SEMIN=1, SEMAX=0 ~ (SEMIN + SEMAX+1)*32
If below 32, the current should be bumped up.

There we are, with enabled CoolStep there is a mean CS of around 25 and min down to 16, because of SEIMIN=0 (50%).
25/31 * 100 = 80 %, 20% current reduction on medium velocity moves.

There is missing part, I didn't test the "high" velocities where backEMF is too big, because for me they are [too large](https://www.reprapfirmware.org/emf.html).
image
So, I can suggest assuming half of the supply voltage as back EMF is a safe margin, according to the graph.
Because up to 1000 mm/s there is still adequate response from CoolStep in my test.

---
My 2 cents:
I have only TMC5160 48V & TMC2209, so only this combination was tested.
The motors are LDO 2504AC.
I initially tuned SGT values by sensorless homing.
I so far have no issues with Coolstep, with almost default values.
CoolStep threshold will control sensorless homing, and IIRC, sensorless homing speed must be above that threshold.
On my printer, sensorless homing works fine with 50 mm/s, but with CoolStep at such a low threshold and with allowed reduction SEIMIN=1 (25% of CS) leads to skipping steps.
With an extruder, CoolStep can work, but its usefulness really depends on setup and extrusion flow.
On orbiter 2, flow shall be above 25 mm3^s. to get 2 RPS.

Will be happy to hear any feedback, even if these things are useless.
Thanks!

Raw data:
[stepper_y-20240515_004650.csv](https://github.com/Klipper3d/klipper/files/15315327/stepper_y-20240515_004650.csv)

### DIFF
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -5,6 +5,7 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import logging, collections
 import stepper
+from . import bulk_sensor
 
 
 ######################################################################
@@ -220,6 +221,96 @@ class TMCErrorCheck:
             self.last_drv_fields = {n: v for n, v in fields.items() if v}
         return {'drv_status': self.last_drv_fields, 'temperature': temp}
 
+######################################################################
+# Record driver status
+######################################################################
+
+class TMCStallguardDump:
+    def __init__(self, config, mcu_tmc):
+        self.printer = config.get_printer()
+        self.stepper_name = ' '.join(config.get_name().split()[1:])
+        self.mcu_tmc = mcu_tmc
+        self.mcu = self.mcu_tmc.get_mcu()
+        self.fields = self.mcu_tmc.get_fields()
+        self.sg2_supp = False
+        self.sg4_reg_name = None
+        # It is possible to support TMC2660, just disable it for now
+        if not self.fields.all_fields.get("DRV_STATUS", None):
+            return
+        # Collect driver capabilities
+        if self.fields.all_fields["DRV_STATUS"].get("sg_result", None):
+            self.sg2_supp = True
+        # New drivers have separate register for SG4 result
+        if self.mcu_tmc.name_to_reg.get("SG_RESULT", 0):
+            self.sg4_reg_name = "SG_RESULT"
+        # 2240 supports both SG2 & SG4
+        if self.sg4_reg_name is None:
+            if self.mcu_tmc.name_to_reg.get("SG4_RESULT", 0):
+                self.sg4_reg_name = "SG4_RESULT"
+        # TMC2208
+        if self.sg2_supp is None and self.sg4_reg_name is None:
+            return
+        self.optimized_spi = False
+        # Bulk API
+        self.samples = []
+        self.query_timer = None
+        self.error = None
+        self.batch_bulk = bulk_sensor.BatchBulkHelper(
+            self.printer, self._dump, self._start, self._stop)
+        api_resp = {'header': ('time', 'sg_result', 'cs_actual')}
+        self.batch_bulk.add_mux_endpoint("tmc/stallguard_dump", "name",
+                                         self.stepper_name, api_resp)
+    def _start(self):
+        self.error = None
+        status = self.mcu_tmc.get_register_raw("DRV_STATUS")
+        if status.get("spi_status"):
+            self.optimized_spi = True
+        reactor = self.printer.get_reactor()
+        self.query_timer = reactor.register_timer(self._query_tmc,
+                                                  reactor.NOW)
+    def _stop(self):
+        self.printer.get_reactor().unregister_timer(self.query_timer)
+        self.query_timer = None
+        self.samples = []
+    def _query_tmc(self, eventtime):
+        sg_result = -1
+        cs_actual = -1
+        recv_time = eventtime
+        try:
+            if self.optimized_spi or self.sg4_reg_name == "SG4_RESULT":
+                #TMC2130/TMC5160/TMC2240
+                status = self.mcu_tmc.get_register_raw("DRV_STATUS")
+                reg_val = status["data"]
+                cs_actual = self.fields.get_field("cs_actual", reg_val)
+                sg_result = self.fields.get_field("sg_result", reg_val)
+                is_stealth = self.fields.get_field("stealth", reg_val)
+                recv_time = status["#receive_time"]
+                if is_stealth and self.sg4_reg_name == "SG4_RESULT":
+                    sg4_ret = self.mcu_tmc.get_register_raw("SG4_RESULT")
+                    sg_result = sg4_ret["data"]
+                    recv_time = sg4_ret["#receive_time"]
+            else:
+                # TMC2209
+                if self.sg4_reg_name == "SG_RESULT":
+                    sg4_ret = self.mcu_tmc.get_register_raw("SG_RESULT")
+                    sg_result = sg4_ret["data"]
+                    recv_time = sg4_ret["#receive_time"]
+        except self.printer.command_error as e:
+            self.error = e
+            return self.printer.get_reactor().NEVER
+        print_time = self.mcu.estimated_print_time(recv_time)
+        self.samples.append((print_time, sg_result, cs_actual))
+        if self.optimized_spi:
+            return eventtime + 0.001
+        # UART queried as fast as possible
+        return eventtime + 0.005
+    def _dump(self, eventtime):
+            if self.error:
+                raise self.error
+            samples = self.samples
+            self.samples = []
+            return {"data": samples}
+
 
 ######################################################################
 # G-Code command helpers
@@ -233,6 +324,7 @@ class TMCCommandHelper:
         self.mcu_tmc = mcu_tmc
         self.current_helper = current_helper
         self.echeck_helper = TMCErrorCheck(config, mcu_tmc)
+        self.record_helper = TMCStallguardDump(config, mcu_tmc)
         self.fields = mcu_tmc.get_fields()
         self.read_registers = self.read_translate = None
         self.toff = None
--- a/klippy/extras/tmc2130.py
+++ b/klippy/extras/tmc2130.py
@@ -200,12 +200,21 @@ class MCU_TMC_SPI_chain:
         cmd = self._build_cmd([reg, 0x00, 0x00, 0x00, 0x00], chain_pos)
         self.spi.spi_send(cmd)
         if self.printer.get_start_args().get('debugoutput') is not None:
-            return 0
+            return {
+                "spi_status": 0,
+                "data": 0,
+                "#receive_time": .0,
+            }
         params = self.spi.spi_transfer(cmd)
         pr = bytearray(params['response'])
         pr = pr[(self.chain_len - chain_pos) * 5 :
                 (self.chain_len - chain_pos + 1) * 5]
-        return (pr[1] << 24) | (pr[2] << 16) | (pr[3] << 8) | pr[4]
+        return {
+            "spi_status": pr[0],
+            "data": (pr[1] << 24) | (pr[2] << 16) | (pr[3] << 8) | pr[4],
+            "#receive_time": params["#receive_time"],
+        }
+
     def reg_write(self, reg, val, chain_pos, print_time=None):
         minclock = 0
         if print_time is not None:
@@ -258,11 +267,13 @@ class MCU_TMC_SPI:
         self.tmc_frequency = tmc_frequency
     def get_fields(self):
         return self.fields
-    def get_register(self, reg_name):
+    def get_register_raw(self, reg_name):
         reg = self.name_to_reg[reg_name]
         with self.mutex:
-            read = self.tmc_spi.reg_read(reg, self.chain_pos)
-        return read
+            resp = self.tmc_spi.reg_read(reg, self.chain_pos)
+        return resp
+    def get_register(self, reg_name):
+        return self.get_register_raw(reg_name)["data"]
     def set_register(self, reg_name, val, print_time=None):
         reg = self.name_to_reg[reg_name]
         with self.mutex:
--- a/klippy/extras/tmc2130.py
+++ b/klippy/extras/tmc2130.py
@@ -272,6 +272,13 @@ class MCU_TMC_SPI:
         with self.mutex:
             resp = self.tmc_spi.reg_read(reg, self.chain_pos)
         return resp
+    def decode_spi_status(spi_status):
+        return {
+            "standstill": spi_status >> 3 & 0x1,
+            "sg2": spi_status >> 2 & 0x1,
+            "driver_error": spi_status >> 1 & 0x1,
+            "reset_flag": spi_status & 0x1
+        }
     def get_register(self, reg_name):
         return self.get_register_raw(reg_name)["data"]
     def set_register(self, reg_name, val, print_time=None):
--- a/klippy/extras/tmc2130.py
+++ b/klippy/extras/tmc2130.py
@@ -232,6 +232,8 @@ class MCU_TMC_SPI_chain:
         pr = pr[(self.chain_len - chain_pos) * 5 :
                 (self.chain_len - chain_pos + 1) * 5]
         return (pr[1] << 24) | (pr[2] << 16) | (pr[3] << 8) | pr[4]
+    def get_mcu(self):
+        return self.spi.get_mcu()
 
 # Helper to setup an spi daisy chain bus from settings in a config section
 def lookup_tmc_spi_chain(config):
@@ -292,6 +294,8 @@ class MCU_TMC_SPI:
             "Unable to write tmc spi '%s' register %s" % (self.name, reg_name))
     def get_tmc_frequency(self):
         return self.tmc_frequency
+    def get_mcu(self):
+        return self.tmc_spi.get_mcu()
 
 
 ######################################################################
--- a/klippy/extras/tmc2660.py
+++ b/klippy/extras/tmc2660.py
@@ -198,11 +198,14 @@ class MCU_TMC2660_SPI:
         self.fields = fields
     def get_fields(self):
         return self.fields
-    def get_register(self, reg_name):
+    def get_register_raw(self, reg_name):
         new_rdsel = ReadRegisters.index(reg_name)
         reg = self.name_to_reg["DRVCONF"]
         if self.printer.get_start_args().get('debugoutput') is not None:
-            return 0
+            return {
+                'data': 0,
+                '#receive_time': .0,
+            }
         with self.mutex:
             old_rdsel = self.fields.get_field("rdsel")
             val = self.fields.set_field("rdsel", new_rdsel)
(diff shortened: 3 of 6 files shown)


JamesH1978:

Thanks
James

nefelim4ag:
I will update that branch with measurement stuff, as current changes superseded by: https://github.com/Klipper3d/klipper/pull/6139

KevinOConnor:
Thanks.  I have a few high-level comments:
1. I can see where it would be useful to gather this information, and it seems like a useful feature.
2. I think it would be preferable to add this to the existing "motan" system instead of creating a new set of custom text files and graphing tools.  That is, it would be preferable to use `bulk_sensor.BatchBulkHelper()` to gather the samples and write them out over the API server.  You can take a look at how `motion_report.py` uses this helper to gather and report results - `adxl345.py` is another example.  You can also look at `scripts/motion_report/` and https://www.klipper3d.org/Debugging.html#motion-analysis-and-data-logging for information on analyzing the data.  Using the "motan" system allows for cross-analysis with motor movements and other sensors.
3. The `cs_actual` is a good catch and makes sense to change.
4. Minor, but I think it would be preferable to add this new logic to a new class in tmc.py (eg, `TMCStallguardDump()`) instead of adding it to `TMCCommandHelper()`.  You can take a look at how `TMCCommandHelper()` instantiates `TMCErrorCheck()` as an example.
5. Also minor, but I don't think it is necessary to change all the tmcxxxx.py code to activate this feature - the `TMCStallguardDump()` code can self activate whenever there is a `cs_actual` field present (or some other appropriate field).

Cheers,
-Kevin

nefelim4ag:
2. bulk_sensor.BatchBulkHelper() reused, api endpoint names/format up to you (just set something). _I initially assumed it is hardly coupled to sensor bulk - so tried to do magic around spi unsuccessfully._ :)
3. Thanks, rebased
4. I also think so - moved to a separate class.
5. yep, done.

Currently, I'm in motan, trying to do any useful graphs with this.
I feel like a simple return of value makes most of graph empty/zero

```
    def pull_data(self, req_time):
        jmsg = self.jdispatch.pull_msg(req_time, self.name)
        ....
        if jmsg is None:
            return
        time, velocity, sg_result, cs_actual = data[0]
        if self.filter == "sg_result":
            return sg_result
```

P.S. We can actually return sg_result/cs_actual in a standstill, but sg_resuult would be meaningless. cs actual can show work of IHold and maybe lose of microstep precision in pair with encoder?

nefelim4ag:
So, more or less done, I'm not sure that I'm correctly getting/calculating print time, maybe estimated_print_time() is better.

Feels like the perfect solution is to get SPI answer clock timing, then convert it to print_time somehow, but this looks like too much work I guess.

Thanks.

nefelim4ag:
image
The estimated print time is overall better here, now shift is around 15 ms.
In my setup query, TMC uses around 1 ms.

But I have no idea how to get better accuracy here.

KevinOConnor:
Thanks.

I fear I may not have understood the goal of this PR when I first reviewed it.

Is your goal, 1) to merge a tool for configuring coolstep, 2) to merge support for analyzing tmc drivers for motion analysis, 3) to not merge but discuss possible changes, or 4) something else?

If the goal is to merge a tool to configure coolstep, then I think the tool would need to be tested by a number of interested users, and have feedback showing the resulting calibration improves real-world results for a notable audience.

If the goal is to enhance motion analysis, then I think the PR should be streamlined for that goal.  In particular, I'd avoid adding a MEASURE_STALLGUARD command, avoid the custom data file format, and avoid adding the calibrate_stallguard.py script.  It may also be worthwhile to consider dumping the raw tmc fields (eg, DRV_STATUS, SG_RESULT) to the output instead of processing it in the main klipper process (that is, let the user, or motan scripts, post process the fields of interest).

> But I have no idea how to get better accuracy here.

I'm not sure there is a way to get good timing accuracy with the tmc drivers.  The current code seems to query three fields (TSTEP, DRV_STATUS, SG_RESULT) which means three round-trip times between host -> mcu -> driver.  Inherently that's going to be difficult to get good timing with.  I'm also not sure why you want TSTEP as we already know how frequently Klipper sends steps to the driver (but maybe I'm missing something).

That said, the most accurate timestamp is likely to pull out the `params['#receive_time']` from the low-level `self.spi.spi_transfer()` (or `self.tmcuart_send_cmd.send()`) request.  Converting the systime to eventtime (as you are doing) is likely a lot easier though, and likely only marginally less accurate.

Cheers,
-Kevin

nefelim4ag:
TLDR: Something in between. Let's focus on motan.

My initial goal is to get CoolStep working thresholds (Superseded by: https://github.com/Klipper3d/klipper/pull/6139).
My second goal is to add any tooling to work with it.
The TMC datasheet states to monitor "live" SG value & etc, and it is just impossible to do right now - that is what I initially tried to do.
_That said, I'm flexible here. Motan & batch logger is more powerful/flexible solution - it is implemented._

Of course, I will be glad if someone has an interest and also gives it a try, but for now in my 3d printing circles, no interest so far.
I agree with not adding a new command and tooling, even if it feels simpler.
Motan interface is more complex of course, but superior.
_IIRC they are in separate commits, so can be easily dropped/skipped. I should recheck and squash some commits, of course, they are here for the history of war against timing precision_

---
About raw fields, maybe they are just slightly different between drivers.
2130 & 5160 have slightly different numbers of fields in status.
SG_RESULT exists in 2209, but not in 2130/5160
2240 use both.

Feels like very complex to postprocess, no?
As there is "low" frequency dump, maybe just dump them parsed as a dictionary?

---
TSTEP was initially used because it is a direct drive sense of step frequency, because I has no way to dump raw speed of motor, and TSTEP allow matching it against thresholds.

With motan:
- I don't know if there is a way to show stepper speed for CoreXY. But it can be calculated of corse.
- is stealth threshold for 2240?

But I also feel it wasteful in some way to collect TSTEP.

---
Yep, it looks like pass low level data to TMC is a way, but I didn't have enough courage to pass it through.
So, If it is fine for now - we can leave it as is.

klippy/extras/tmc.py
```diff
@@ -3,8 +3,12 @@
 # Copyright (C) 2018-2020  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
+import os
+import time
+import multiprocessing
```
yunline, line 8: These libraries are imported but it seems they are left unused.

nefelim4ag:
Some time has passed, and my understanding of code and features has changed.
So, I reworked the patchset.

I think the goals right now are simpler:
- Provide some API to work with the Stallguard value. Example: debug/calibrate sensorless homing. (I think, then it makes sense to update sensorless documentation, to actually document it.)
- Track the CS value; it is "free" to track. It should allow tracking the IHOLD behavior and Coolstep if required.

I can additionally dump the standstill state if it makes sense.
AFAIK, there are no more registers I could suggest for dynamic tracking.

There are 2 parts:
- Preparation work in the TMC SPI/UART to allow gathering low-level information, like SPI_STATUS or send/recv time.
- For high frequency UART requests, there are knob to increase the UART speed.
- Code now mostly unconditionally dumps the DRV_STATUS/StallGuard registers.

Some minor tweaks, I do not dump the TSTEP or use it, because assuming we have precise enough stepq/trapq data, there is no reason to do so. Where it does make sense, the stealth flag is used to guess the current mode (TMC2240).

Tested.

---
Hmmm... it seems that current UART implementation works unreliable at high speeds >125k, even on the STM32H7 with plenty of power. 
TMC is supposed to work up to 12Mhz/16 = ~750k
Hmmm...

:D It does work. Well, nice to know.
Seems okay at 350k, when going higher, it seems to also fail more often.
(it is a bad solution, at least in this specific implementation, because it will often fail due to timers activity).
```
diff --git a/src/tmcuart.c b/src/tmcuart.c
index a2f5f8fcd..808a57a1d 100644
--- a/src/tmcuart.c
+++ b/src/tmcuart.c
@@ -186,6 +186,17 @@ DECL_COMMAND(command_config_tmcuart,
              "config_tmcuart oid=%c rx_pin=%u pull_up=%c"
              " tx_pin=%u bit_time=%u");
 
+static void
+execute_inline(struct tmcuart_s* t)
+{
+    struct timer* timer = &t->timer;
+    timer->waketime = timer_read_time();
+    while (timer->func(timer) == SF_RESCHEDULE) {
+        while (timer_is_before(timer_read_time(), timer->waketime))
+            ;
+    }
+}
+
 // Parse and schedule a TMC UART transmission request
 void
 command_tmcuart_send(uint32_t *args)
@@ -210,6 +221,10 @@ command_tmcuart_send(uint32_t *args)
         t->bit_time = t->cfg_bit_time;
         t->timer.func = tmcuart_send_event;
     }
+    if (t->cfg_bit_time < timer_from_us(8)) {
+        execute_inline(t);
+        return;
+    }
     irq_disable();
     t->timer.waketime = timer_read_time() + timer_from_us(200);
     sched_add_timer(&t->timer);
```

nefelim4ag:
TMC5160 and IHOLD
image
TMC2240 and stealth switching at 3 mm/s
image

KevinOConnor:
Thanks.  In general, it seems useful to me.  A couple of comments:
* I think the `#sent_time` does something a little different from what you are expecting.  It reports the query time associated with a response message *if* that response is generated from a synchronous query.  So, on tmc_uart messages, the `#sent_time` doesn't contain any useful information (as the response isn't synchronous with the request command).  Also, even on spi messages, `#sent_time` could be zero if the serialqueue.c code retransmits a message.  I suspect you want to use `#receive_time`.
* I'd be leery about exposing a new tmc_uart `uart_speed` option.  It seems a bit of a "foot gun" option as it's really easy to set it wrong and not easy to set it to a valid value.

Separately, just out of curiosity, what's the reason for querying "cs_actual"?  Can the host calculate it from the microsteps, run_current, and stepper phase information - or is this some kind of "coolstep" thing?

Cheers,
-Kevin

nefelim4ag:
> Separately, just out of curiosity, what's the reason for querying "cs_actual"? Can the host calculate it from the microsteps, run_current, and stepper phase information - or is this some kind of "coolstep" thing?

It is an automatic current scaling/energy-saving thing.
It does show the configured/requested motor current, not the coil current (so microstep/phase are unrelated).
So, in the normal case, while printing IRUN == CS.
In the standstill with IHOLD != IRUN, CS would show how the current is decreased over time (shown above with tmc5160).
In CoolStep, it would show how the current decreased based on the SG measurements.

Aside from that, it is mostly useless.
So, I'm not sure if it is really needed. _If someone wants to use it, it is pretty easy to add it back._
I could also easily remove it from the dump output, if we think it is not needed or could cause harm.

Hmm..
If we drop it, technically, I could slim down the reading code.

KevinOConnor:
I don't have a preference on `cs_actual`.

The change looks fine to me, so let me know when you'd like me to commit it.

Separately, invoking a shutdown on a failed query is pretty severe.  Might be better to unregister the timer and have the next `_dump()` invocation raise a command error.  That should propagate the error back up to the webapi clients.

Also, seems a little odd that `data_logger.py` tries to subscribe to every stepper, even those that may not be using tmc drivers.

Cheers,
-Kevin

nefelim4ag:
> Separately, invoking a shutdown on a failed query is pretty severe. Might be better to unregister the timer and have the next _dump() invocation raise a command error. That should propagate the error back up to the webapi clients.

~~I'm hoping that I can can more or less safely ignore errors here.~~
~~There is a TMC Error check, if it detects an issue it would shutdown everything.~~

Ah, I got it.
Done.
_I agree, it is better to not shutdown and to somewhat notify clients that there is something broken underneath, and my approach with borked data is not really useful._

> Also, seems a little odd that data_logger.py tries to subscribe to every stepper, even those that may not be using tmc drivers.

Fixed.

JFYI:
I slightly reworked and simplified the dump class and reused the field helper.
Assuming the main target is sensorless homing.
I prefer to only read the SG value on UART drivers.
TMC2240 UART also covered, under current normal conditions it does sensorless homing by the SpreadCycle/SG2, instead of SG4.

I hope it is more or less ready.
I could try to update the sensorless homing documentation with direct examples for how to peek inside.

---
I have a dream where a field helper could cache the reads, so we could reuse the recent result from the DRV_STATUS.
But right now it would be an overoptimization, so just sharing strange ideas.

KevinOConnor:
Okay, thanks.  Let me know when you are ready for me to commit it.

Cheers,
-Kevin

nefelim4ag:
It seems to work fine.
Let's merge,

Thanks.

---
I did some tests for the sake of completeness with sensorless homing:
image
```
./scripts/motan/motan_graph.py -g '[["stallguard(stepper_y,sg_result)"],["stallguard(stepper_x,sg_result)"],["adxl345(adxl345,y)"],["derivative(stepq(stepper_y))"],["step_phase(tmc5160 stepper_y)"]]' ./sensorless_y -s 1 -d 7
```
It seems difficult to get the data where the MCU actually stops the toolhead, which is somewhat expected.
It is easy to compare data with an accelerometer:
image
It seems close to the stepper phase and a little late in comparison to the accelerometer.

---
This is a stall test, where I just hit the axis limit after the sensorless home:
image
```
./scripts/motan/motan_graph.py -g '[["stallguard(stepper_y,sg_result)"],["stallguard(stepper_x,sg_result)"],["adxl345(adxl345,y)"],["deviation(angle(mt6826s),stepq(stepper_x))"],["trapq(toolhead,y_velocity)"],["step_phase(tmc5160 stepper_y)"]]' ./stall_y -s 1.4 -d 0.8
```
Seems like generally the stallguard data is late for ~10ms

---
Normal move (without stall, acceleration decreased 10k -> 2k to easier find where the speed limit):
image
```
./scripts/motan/motan_graph.py -g '[["stallguard(stepper_y,sg_result)"],["stallguard(stepper_x,sg_result)"],["trapq(toolhead,y_velocity)"],["step_phase(tmc5160 stepper_y)"]]' ./normal_move_y -s 1 -d 1
```
And the minimal speed where my drivers show something at my sensitivity level is around 30mm/s
image
The test at 40mm/s, where the Y driver barely shows something.
image

Sineos:
Wow, quite insightful. 👍 

Since sensorless homing is quite popular, wouldn't it make sense to support users in the setup process with some `SH_CALIBRATE` command that determines the optimal sensitivity, speed, and potentially current automatically?

KevinOConnor:
Thanks.

-Kevin

nefelim4ag:
@Sineos, honestly, idk
_There are my thoughts, maybe they can help. If I try to do something like that, it would probably initially happen on the Discouse._

An additional investigation is needed first to at least have some model of the actual Stallguard2 values behaviour.
Then it would be nice to make a decision of how to properly calibrate those values without an endstop or user input.
(Basically, macro, to alter the toolhead position and repeatedly crush into the frame, at the known point, until the code would see the expected values?)
```
G91
G1 Y-110
// hit at -100
G1 Y100
// repeat
G1 Y-110
...
```

The only thing that I can say is that SG4 (TMC2209/TMC2240) is more sensitive and predictable, at least it seems to be from my small tests.
_(Maybe it would be nice to have it enabled on the TMC2240, instead of the SG2)._

I did small tests right now.
Like 2A -> 1A -> 1.5A, just removes the signal from the data.
Decreasing the SGT value does not help with the restoration of the signal 2 -> 1 -> 0.
(SGT=0 default sensitivity, SGT 1..63 decrease sensitivity, SGT -1..-64 increase).

With 1.5A/SGT=0, the first speed point where I actually start to see something again is around 80 mm/s.
Why so a small decrease in the current and an increase in the sensitivity leading to this? idk.

If I try to somewhat guess - why, I probably want to estimate the backEMF:
image

But if I try to calculate the BackEMF from before, with my default settings, it would look like less BackEMF:
image

So, I can say something like: 2-3V of total backEMF seems like a bare minimum for me.
Closer to the motor rated current seems better.
Probably the increase of the current increases SG2 output, increase of the speed also does so. So, the sensitivity should also be higher (and vice versa).

But it is only my small data points, I'm not sure it would be true for different power voltages, motors, and SG2 drivers.
So, it is hard for me, to say, that I can easily come up with some autotuning tool here.
Ah, yes, also there is a mysterious thing, that same motors (LDO2504) with the same driver (TMC5160)
X and Y drivers, does output different values sometimes.
(You probably already noticed this on my graphs, above).

_In comparison, my TMC2209 (SG4) on the Z axis, are pretty agreed between each other, I even think that it is possible to sync AWD motors with them._

Thanks.

---
_Maybe, it would be helpful to write a WebUI/Graph stallguard tool as the Moonraker client.
It seems like Motan is already too hard to use for the general public._

Basic WebUI: https://klipper.discourse.group/t/sensorless-stallguard-webui-graph-tool/24231

PR #6593, opened 2024-05-08
Fixes exceptions regarding bytes vs str

limitz:
Clean install on raspberry pi CM4, Bookworm, Python 3.11.2:

scripts/whconsole.py is throwing exception:
```
TypeError: a bytes-like object is required, not 'str'
```
I simply added encode / decode where it seemed necessary to get rid of the exception.


JamesH1978:

Thanks
James

kabroxiko (changes requested):
It works, but you must change shebang at line 1 to python3.

PR #6597, opened 2024-05-11
Update printer-creality-ender6-2020.cfg

Neelrhi:
Added in dimensions for the the custom 3D-printed mount.

### DIFF
--- a/config/printer-creality-ender6-2020.cfg
+++ b/config/printer-creality-ender6-2020.cfg
@@ -50,32 +50,32 @@ dir_pin: PB5
 enable_pin: !PC3
 microsteps: 16
 rotation_distance: 8
-position_endstop: 0.0                     # disable to use BLTouch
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thankyou for submitting a PR, our stock configurations are for bone stock printers as if they have just come out of the box from factory, as users may not have custom additions. This printer did not ship with a bltouch as standard, so therefore the code for it is commented out and left for the user to uncomment if they add this option.

Thanks
James

bevanweiss:
@Neelrhi I'm unsure that this PR adds significant value.
As noted by James the Creality Ender 6 did not ship with a BLTouch, so such a configuration should never be the default for this printer.  The commented out section in the current source represents the typical 'from Vendor' addition.
Adding as default, configurations for additional 3rd party items would almost certainly result in confusion for end users.

I believe you should close this PR.

zimmertr:
If I had to guess, this user mistakenly opened this PR.

PR #6598, opened 2024-05-11
MANUAL_STEPPER STOP_ON_ENDSTOP and SYNC description update

dfrog3:
Added a line in the MANUAL_STEPPER gcode refrence that explains how the STOP_ON_ENDSTOP setting ignores the SYNC setting. I think this will be useful to users so they can better understand the limmitations of the MANUAL_STEPPER command. This behavior can be found in manual_stepper.cmd_MANUAL_STEPPER line 96.


JamesH1978:

Thanks
James

PR #6599, opened 2024-05-14
Update install-octopi.sh

samascience:
fixing this issue 

###### Installing packages...
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'libusb-1.0-0-dev' for regex 'libusb-1.0' Note, selecting 'libusb-1.0-doc' for regex 'libusb-1.0' Note, selecting 'libusb-1.0-0' for regex 'libusb-1.0' Package python-dev is not available, but is referred to by another package. This may mean that the package is missing, has been obsoleted, or is only available from another source
However the following packages replace it:
  python-dev-is-python3

### DIFF
--- a/scripts/install-octopi.sh
+++ b/scripts/install-octopi.sh
@@ -8,7 +8,7 @@ PYTHONDIR="${HOME}/klippy-env"
 install_packages()
 {
     # Packages for python cffi
-    PKGLIST="virtualenv python-dev libffi-dev build-essential"
+    PKGLIST="virtualenv python-dev-is-python3 libffi-dev build-essential"
     # kconfig requirements
     PKGLIST="${PKGLIST} libncurses-dev"
     # hub-ctrl


samascience:
fixing dependency issue

JamesH1978:

Also can you confirm this issue happens on a stock octopi image downloaded from official sources?

Thanks
James

Sineos:
I'm not sure that this fix is correct:
* `python-dev-is-python3` ist just a symlink from `python3 `to `python`
* Due to its dependencies it will force the installation of python3
* Newer Debian or Ubuntu distributions no longer default to python2 on such packages but require you to explicitly choose either python2 or python3
* The script still tries to create a python2 virtual env here:
https://github.com/Klipper3d/klipper/blob/79930ed99a1fc284f41af5755908aa1fab948ce1/scripts/install-octopi.sh#L36-L37

PR #6600, opened 2024-05-14
Fix adaptive probe count on delta printers

voidtrance:
Round beds require an odd number of probe points in order to prevent erroneously truncating the mesh.

The adaptive mesh algorithm did not consider that and as a result, it was possible to generate adaptive
meshes with even number of probe points.

This change fixes this by increasing the probe point count by 1 in cases where the adaptive probe points are even.


voidtrance:
@Arksine, would you mind reviewing this?

yzheka:
I have tested it on kossel linear plus and now adaptive bed mesh works without problems for me.

Arksine (approved):

Arksine:
The change looks good to me.  Thanks.

KevinOConnor:
Thanks.

-Kevin

PR #6601, opened 2024-05-15
Minor tmc driver improvements

KevinOConnor:
A couple of minor enhancements that I noticed after the merge of #6139 .
1. Simplify the `TMCtstepHelper()` code.
2. There is no need to warn the user about setting `high_velocity_threshold` when using sensorless homing as the sensorless homing can save/restore `thigh`.

@leptun - it would be great if you could take a look at these changes.

-Kevin

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3327,11 +3327,8 @@ run_current:
 #high_velocity_threshold:
 #   The velocity (in mm/s) to set the TMC driver internal "high
 #   velocity" threshold (THIGH) to. This is typically used to disable
-#   the "CoolStep" feature at high speeds. Important - if
-#   high_velocity_threshold is set and "sensorless homing" is used,
-#   then one must ensure that the homing speed is below the high
-#   velocity threshold! The default is to not set a TMC "high
-#   velocity" threshold.
+#   the "CoolStep" feature at high speeds. The default is to not set a
+#   TMC "high velocity" threshold.
 #driver_MSLUT0: 2863314260
 #driver_MSLUT1: 1251300522
 #driver_MSLUT2: 608774441
@@ -3646,11 +3643,8 @@ run_current:
 #high_velocity_threshold:
 #   The velocity (in mm/s) to set the TMC driver internal "high
 #   velocity" threshold (THIGH) to. This is typically used to disable
-#   the "CoolStep" feature at high speeds. Important - if
-#   high_velocity_threshold is set and "sensorless homing" is used,
-#   then one must ensure that the homing speed is below the high
-#   velocity threshold! The default is to not set a TMC "high
-#   velocity" threshold.
+#   the "CoolStep" feature at high speeds. The default is to not set a
+#   TMC "high velocity" threshold.
 #driver_MSLUT0: 2863314260
 #driver_MSLUT1: 1251300522
 #driver_MSLUT2: 608774441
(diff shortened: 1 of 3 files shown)


nefelim4ag (approved):

leptun (approved):
I agree with these changes. Though I did not have time yet to test them on a printer

nefelim4ag:
Defining CoolStep above homing velocities effectively broke sensorless homing.

leptun:
I think that is expected? defining a custom coolstep threshold shall be used at all times. So if the coolstep velocity is already set, it doesn't get overridden to 0xFFFFF during the homing move. I actually _need_ to have the coolstep threshold set to 15mm/s during homing, or else the stallguard triggers during the acceleration move. With a non-zero coolstep threshold during homing, I can increase the stallguard sensitivity without causing false positives, making the homing tap softer and more precise.

If you need such a high coolstep threshold, then you should probably use a script that reduces the coolstep threshold at the beginning of home then restores it at the end. This is the exact approach I use for lower homing currents:
```
[gcode_macro G28]
rename_existing: G28.1
description: Advanced TMC homing move
gcode:
  {% set x = 'X' in params %}
  {% set y = 'Y' in params %}
  {% set z = 'Z' in params %}
  {% set all = not (x or y or z) %}
  {% if all or x %}
    SET_TMC_CURRENT STEPPER=stepper_x CURRENT=0.15
  {% endif %}
  {% if all or y %}
    SET_TMC_CURRENT STEPPER=stepper_y CURRENT=0.18
  {% endif %}
  {% if all or z %}
    SET_TMC_CURRENT STEPPER=stepper_z CURRENT=0.3
  {% endif %}

  G28.1 {rawparams}

  {% if all or x %}
    SET_TMC_CURRENT STEPPER=stepper_x CURRENT={printer['tmc2130 stepper_x'].run_current} HOLDCURRENT={printer['tmc2130 stepper_x'].hold_current}
  {% endif %}
  {% if all or y %}
    SET_TMC_CURRENT STEPPER=stepper_y CURRENT={printer['tmc2130 stepper_y'].run_current} HOLDCURRENT={printer['tmc2130 stepper_y'].hold_current}
  {% endif %}
  {% if all or z %}
    SET_TMC_CURRENT STEPPER=stepper_z CURRENT={printer['tmc2130 stepper_z'].run_current} HOLDCURRENT={printer['tmc2130 stepper_z'].hold_current}
  {% endif %}
```

nefelim4ag:
@leptun cares about sensorless homing reliability
@nefelim4ag cares about CoolStep current reduction while printing

So, we did a small talk about coolstep threshold and approaches.
CoolStep threshold can be used to get reliable sensorless homing and avoid false positivities on the acceleration phase.
That is the reason why we honor config value in this specific code if defined.

(we have crazy different setups like Prusa MK3 and 0.5A steppers (possibly high inductance), and RatRig V-core 3.1 with LDO 2504 and 1.768A RMS (low inductance))

My doubt there was that such a low threshold used for homing would be unreliable for printing speeds on CoreXY because if one motor decreases current other one can just trigger motor stall and steps skipping.

I decreased the cool step threshold from 150 mm/s to 50 mm/s, to test reliability on the dummy test print (where I got step skips before with SEIMIN=1 (25% of rated current)), but so far, so good I didn't get step skips
(30 min with 300% speed up to 1000 mm/s & 20k acceleration).

So let's assume for now, the threshold for homing speeds shall work for printing in the correct setup. Someone with doubt (like me) who does not use sensorless homing, can set the threshold to value appropriate to personal taste.

And leave the code as is.

Thanks

KevinOConnor:
Thanks.  I'm guessing the comments above are about the existing behavior of not always clearing `tcoolthrs` during sensorless homing (that is, only set it to 0xfffff if it was zero).  This PR should not have changed the behavior of `tcoolthrs`.   If there was a regression introduced in this PR, let me know.

Separately, I don't have a particular preference on sensorless homing wrt `tcoolthrs`.  Unless there is a problem, I'd just assume keep the behavior as it was.

Cheers,
-Kevin

PR #6602, opened 2024-05-16
temperature_probe: probe thermal drift compensation

Arksine:
This adds a `temperature_probe` module that supports configuring a temperature sensor for use with inductive probes.  This module implements the smoothing algorithm present in heaters and provides a temperature drift calibration procedure.  The module is intended to be generic, it can be used with any kind of inductive probe, where the calibration logic is farmed out to the probe specific implementation.  Also included in this PR is an implementation for the `probe_eddy_current` module.

The calibration procedure works by using the bed to heat the probe and requesting a manual probe before each sample.  This is intended to filter out thermal expansion.  Samples are taken at a specified temperature interval, by default this is set to every 2C.  The best calibration is achieved by taking samples across the widest temperature range possible.   After a manual probe, the `probe_eddy_current` module will take samples every .5mm, up to 4mm.  These samples are used to create a series of 2nd degree polynomials at each height.

Compensation uses the current frequency and coil temperature the find the upper and lower polynomials, then performs linear interpolation to calculate the correct frequency at the target temperature (ie: the temperature of the coil during `PROBE_EDDY_CURENT_CALIBRATE`).

Attached is a plot of the "stock" height calculate vs the corrected height for a probe that has a good temperature calibration.  This plot was measured by using the bed to heat the probe and performing a probe every 30 seconds, after which the height reported by the probe is collected.  `PROBE_EDDY_CURRENT_CALIBRATE` was performed with the probe around 35C.

eddy-mounted-drift-correction

For comparison, I also collected data at a 1kHz sample rate.  In addition, this data was collected with `PROBE_EDDY_CURRENT_CALIBRATE` close to 60C:

eddy-mounted-drift-correction-1khzrate

What stands out most in the above data is that the linear regression of the z-height is closer to the `z_offset` of `.5`.

This PR is built on #6558, so  we likely need to address any outstanding issues with that PR first.  Specific commits for this PR start at 108985cdc8af6c7cd315120b19b9fa06a9af2351.  The last two commits are minor fixes to `probe_eddy_current`, one fixing a stale attribute name and one slightly bumping the delay time after a probe to avoid "Unable to obtain probe_eddy_current sensor readings" errors.  If necessary I can drop these commits from this PR.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2401,6 +2401,65 @@ temperature sensors that are reported via the M105 command.
 #   parameter.
 ```
 
+### [temperature_probe]
+
+Reports probe coil temperature.  Includes optional thermal drift
+calibration for eddy current based probes.   A `[temperature_probe]`
+section may be linked to a `[probe_eddy_current]` by using the same
+postfix for both sections.
+
+```
+[temperature_probe my_probe]
+#sensor_type:
+#sensor_pin:
+#min_temp:
+#max_temp:
+#   Temperature sensor configuration.
+#   See the "extruder" section for the definition of the above
+#   parameters.
+#smooth_time:
+#   A time value (in seconds) over which temperature measurements will
+#   be smoothed to reduce the impact of measurement noise. The default
+#   is 2.0 seconds.
+#gcode_id:
+#   See the "heater_generic" section for the definition of this
+#   parameter.
+#speed:
+#   The travel speed [mm/s] for xy moves during calibration.  Default
+#   is the speed defined by the probe.
+#horizontal_move_z:
+#   The z distance [mm] from the bed at which xy moves will occur
+#   during calibration. Default is 2mm.
+#resting_z:
+#   The z distance [mm] from the bed at which the tool will rest
+#   to heat the probe coil during calibration.  Default is .4mm
+#calibration_position:
+#   The X, Y, Z position where the tool should be moved when
+#   probe drift calibration initializes.  This is the location
+#   where the first manual probe will occur.  If omitted, the
+#   default behavior is not to move the tool prior to the first
+#   manual probe.
+#calibration_bed_temp:
+#   The maximum safe bed temperature (in C) used to heat the probe
+#   during probe drift calibration.  When set, the calibration
+#   procedure will turn on the bed after the first sample is
+#   taken.  When the calibration procedure is complete the bed
+#   temperature will be set to zero.  When omitted the default
+#   behavior is not to set the bed temperature.
+#calibration_extruder_temp:
+#   The extruder temperature (in C) set probe during drift calibration.
+#   When this option is supplied the procedure will wait for until the
+#   specified temperature is reached before requesting the first manual
+#   probe.  When the calibration procedure is complete the extruder
+#   temperature will be set to 0.  When omitted the default behavior is
+#   not to set the extruder temperature.
+#extruder_heating_z: 50.
+#   The Z location where extruder heating will occur if the
+#   "calibration_extruder_temp" option is set.  Its recommended to heat
+#   the extruder some distance from the bed to minimize its impact on
+#   the probe coil temperature.  The default is 50.
+```
+
 ## Temperature sensors
 
 Klipper includes definitions for many types of temperature sensors.
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1415,3 +1415,39 @@ command will probe the points specified in the config and then make independent
 adjustments to each Z stepper to compensate for tilt. See the PROBE command for
 details on the optional probe parameters. The optional `HORIZONTAL_MOVE_Z`
 value overrides the `horizontal_move_z` option specified in the config file.
+
+### [temperature_probe]
+
+The following commands are available when a
+[temperature_probe config section](Config_Reference.md#temperature_probe)
+is enabled.
+
+#### TEMPERATURE_PROBE_CALIBRATE
+`TEMPERATURE_PROBE_CALIBRATE [PROBE=<probe name>] [TARGET=<value>] [STEP=<value>]`:
+Initiates probe drift calibration for eddy current based probes.  The `TARGET`
+is a target temperature for the last sample.  When the temperature recorded
+during a sample exceeds the `TARGET` calibration will complete.  The `STEP`
+parameter sets temperature delta (in C) between samples. After a sample has
+been taken, this delta is used to schedule a call to `TEMPERATURE_PROBE_NEXT`.
+The default `STEP` is 2.
+
+#### TEMPERATURE_PROBE_NEXT
+`TEMPERATURE_PROBE_NEXT`: After calibration has started this command is run to
+take the next sample.  It is automatically scheduled to run when the delta
+specified by `STEP` has been reached, however its also possible to manually run
+this command to force a new sample.  This command is only available during
+calibration.
+
+#### TEMPERATURE_PROBE_COMPLETE:
+`TEMPERATURE_PROBE_COMPLETE`:  Can be used to end calibration and save the
+current result before the `TARGET` temperature is reached.  This command
+is only available during calibration.
+
+#### ABORT
+`ABORT`:  Aborts the calibration process, discarding the current results.
+This command is only available during drift calibration.
+
+### TEMPERATURE_PROBE_ENABLE
+`TEMPERATURE_PROBE_ENABLE ENABLE=[0|1]`: Sets temperature drift
+compensation on or off. If ENABLE is set to 0, drift compensation
+will be disabled, if set to 1 it is enabled.
--- a/klippy/extras/probe_eddy_current.py
+++ b/klippy/extras/probe_eddy_current.py
@@ -14,6 +14,7 @@ class EddyCalibration:
     def __init__(self, config):
         self.printer = config.get_printer()
         self.name = config.get_name()
+        self.drift_comp = DummyDriftCompensation()
         # Current calibration data
         self.cal_freqs = []
         self.cal_zpos = []
@@ -37,8 +38,10 @@ class EddyCalibration:
         self.cal_freqs = [c[0] for c in cal]
         self.cal_zpos = [c[1] for c in cal]
     def apply_calibration(self, samples):
+        cur_temp = self.drift_comp.get_temperature()
         for i, (samp_time, freq, dummy_z) in enumerate(samples):
-            pos = bisect.bisect(self.cal_freqs, freq)
+            adj_freq = self.drift_comp.adjust_freq(freq, cur_temp)
+            pos = bisect.bisect(self.cal_freqs, adj_freq)
             if pos >= len(self.cal_zpos):
                 zpos = -OUT_OF_RANGE
             elif pos == 0:
@@ -51,7 +54,7 @@ class EddyCalibration:
                 prev_zpos = self.cal_zpos[pos - 1]
                 gain = (this_zpos - prev_zpos) / (this_freq - prev_freq)
                 offset = prev_zpos - prev_freq * gain
-                zpos = freq * gain + offset
+                zpos = adj_freq * gain + offset
             samples[i] = (samp_time, freq, round(zpos, 6))
     def freq_to_height(self, freq):
         dummy_sample = [(0., freq, 0.)]
@@ -71,7 +74,8 @@ class EddyCalibration:
         prev_zpos = rev_zpos[pos - 1]
         gain = (this_freq - prev_freq) / (this_zpos - prev_zpos)
         offset = prev_freq - prev_zpos * gain
-        return height * gain + offset
+        freq = height * gain + offset
+        return self.drift_comp.unadjust_freq(freq)
     def do_calibration_moves(self, move_speed):
         toolhead = self.printer.lookup_object('toolhead')
         kin = toolhead.get_kinematics()
@@ -86,6 +90,7 @@ class EddyCalibration:
             return True
         self.printer.lookup_object(self.name).add_client(handle_batch)
         toolhead.dwell(1.)
+        self.drift_comp.note_z_calibration_start()
         # Move to each 40um position
         max_z = 4.0
         samp_dist = 0.040
@@ -112,6 +117,7 @@ class EddyCalibration:
             times.append((start_query_time, end_query_time, kin_pos[2]))
         toolhead.dwell(1.0)
         toolhead.wait_moves()
+        self.drift_comp.note_z_calibration_finish()
         # Finish data collection
         is_finished = True
         # Correlate query responses
@@ -188,6 +194,8 @@ class EddyCalibration:
         # Start manual probe
         manual_probe.ManualProbeHelper(self.printer, gcmd,
                                        self.post_manual_probe)
+    def register_drift_compensation(self, comp):
+        self.drift_comp = comp
 
 # Tool to gather samples and convert them to probe positions
 class EddyGatherSamples:
@@ -265,16 +273,18 @@ class EddyGatherSamples:
             freq = self._pull_freq(start_time, end_time)
             if pos_time is not None:
                 toolhead_pos = self._lookup_toolhead_pos(pos_time)
-            self._probe_results.append((freq, toolhead_pos))
+            sensor_z = None
+            if freq:
+                sensor_z = self._calibration.freq_to_height(freq)
+            self._probe_results.append((sensor_z, toolhead_pos))
             self._probe_times.pop(0)
     def pull_probed(self):
         self._await_samples()
         results = []
-        for freq, toolhead_pos in self._probe_results:
-            if not freq:
+        for sensor_z, toolhead_pos in self._probe_results:
+            if sensor_z is None:
                 raise self._printer.command_error(
                     "Unable to obtain probe_eddy_current sensor readings")
-            sensor_z = self._calibration.freq_to_height(freq)
             if sensor_z <= -OUT_OF_RANGE or sensor_z >= OUT_OF_RANGE:
                 raise self._printer.command_error(
                     "probe_eddy_current sensor not in valid range")
@@ -435,6 +445,20 @@ class PrinterEddyProbe:
             return EddyScanningProbe(self.printer, self.sensor_helper,
                                      self.calibration, z_offset, gcmd)
         return self.probe_session.start_probe_session(gcmd)
+    def register_drift_compensation(self, comp):
+        self.calibration.register_drift_compensation(comp)
+
+class DummyDriftCompensation:
+    def get_temperature(self):
+        return 0.
+    def note_z_calibration_start(self):
+        pass
+    def note_z_calibration_finish(self):
+        pass
+    def adjust_freq(self, freq, temp=None):
+        return freq
+    def unadjust_freq(self, freq, temp=None):
+        return freq
 
 def load_config_prefix(config):
     return PrinterEddyProbe(config)
--- a/klippy/extras/temperature_probe.py
+++ b/klippy/extras/temperature_probe.py
@@ -1,0 +1,716 @@
+# Probe temperature sensor and drift calibration
+#
+# Copyright (C) 2024 Eric Callahan <arksine.code@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import logging
+from . import manual_probe
+
+KELVIN_TO_CELSIUS = -273.15
+
+######################################################################
+# Polynomial Helper Classes and Functions
+######################################################################
+
+def calc_determinant(matrix):
+    m = matrix
+    aei = m[0][0] * m[1][1] * m[2][2]
+    bfg = m[1][0] * m[2][1] * m[0][2]
+    cdh = m[2][0] * m[0][1] * m[1][2]
+    ceg = m[2][0] * m[1][1] * m[0][2]
+    bdi = m[1][0] * m[0][1] * m[2][2]
+    afh = m[0][0] * m[2][1] * m[1][2]
+    return aei + bfg + cdh - ceg - bdi - afh
+
+class Polynomial2d:
+    def __init__(self, a, b, c):
+        self.a = a
+        self.b = b
+        self.c = c
+
+    def __call__(self, xval):
+        return self.c * xval * xval + self.b * xval + self.a
+
+    def get_coefs(self):
+        return (self.a, self.b, self.c)
+
+    def __str__(self):
+        return "%f, %f, %f" % (self.a, self.b, self.c)
+
+    def __repr__(self):
+        parts = ["y(x) ="]
+        deg = 2
+        for i, coef in enumerate((self.c, self.b, self.a)):
+            if round(coef, 8) == int(coef):
+                coef = int(coef)
+            if abs(coef) < 1e-10:
+                continue
+            cur_deg = deg - i
+            x_str = "x^%d" % (cur_deg,) if cur_deg > 1 else "x" * cur_deg
+            if len(parts) == 1:
+                parts.append("%f%s" % (coef, x_str))
+            else:
+                sym = "-" if coef < 0 else "+"
+                parts.append("%s %f%s" % (sym, abs(coef), x_str))
+        return " ".join(parts)
+
+    @classmethod
+    def fit(cls, coords):
+        xlist = [c[0] for c in coords]
+        ylist = [c[1] for c in coords]
+        count = len(coords)
+        sum_x = sum(xlist)
+        sum_y = sum(ylist)
+        sum_x2 = sum([x**2 for x in xlist])
+        sum_x3 = sum([x**3 for x in xlist])
+        sum_x4 = sum([x**4 for x in xlist])
+        sum_xy = sum([x * y for x, y in coords])
+        sum_x2y = sum([y*x**2 for x, y in coords])
+        vector_b = [sum_y, sum_xy, sum_x2y]
+        m = [
+            [count, sum_x, sum_x2],
+            [sum_x, sum_x2, sum_x3],
+            [sum_x2, sum_x3, sum_x4]
+        ]
+        m0 = [vector_b, m[1], m[2]]
+        m1 = [m[0], vector_b, m[2]]
+        m2 = [m[0], m[1], vector_b]
+        det_m = calc_determinant(m)
+        a0 = calc_determinant(m0) / det_m
+        a1 = calc_determinant(m1) / det_m
+        a2 = calc_determinant(m2) / det_m
+        return cls(a0, a1, a2)
+
+class TemperatureProbe:
+    def __init__(self, config):
+        self.name = config.get_name()
+        self.printer = config.get_printer()
+        self.gcode = self.printer.lookup_object("gcode")
+        self.speed = config.getfloat("speed", None, above=0.)
+        self.horizontal_move_z = config.getfloat(
+            "horizontal_move_z", 2., above=0.
+        )
+        self.resting_z = config.getfloat("resting_z", .4, above=0.)
+        self.cal_pos = config.getfloatlist(
+            "calibration_position", None, count=3
+        )
+        self.cal_bed_temp = config.getfloat(
+            "calibration_bed_temp", None, above=50.
+        )
+        self.cal_extruder_temp = config.getfloat(
+            "calibration_extruder_temp", None, above=50.
+        )
+        self.cal_extruder_z = config.getfloat(
+            "extruder_heating_z", 50., above=0.
+        )
+        # Setup temperature sensor
+        smooth_time = config.getfloat("smooth_time", 2., above=0.)
+        self.inv_smooth_time = 1. / smooth_time
+        self.min_temp = config.getfloat(
+            "min_temp", KELVIN_TO_CELSIUS, minval=KELVIN_TO_CELSIUS
+        )
+        self.max_temp = config.getfloat(
+            "max_temp", 99999999.9, above=self.min_temp
+        )
+        pheaters = self.printer.load_object(config, "heaters")
+        self.sensor = pheaters.setup_sensor(config)
+        self.sensor.setup_minmax(self.min_temp, self.max_temp)
+        self.sensor.setup_callback(self._temp_callback)
+        pheaters.register_sensor(config, self)
+        self.last_temp_read_time = 0.
+        self.last_measurement = (0., 99999999., 0.,)
+        # Calibration State
+        self.cal_helper = None
+        self.next_auto_temp = 99999999.
+        self.target_temp = 0
+        self.expected_count = 0
+        self.sample_count = 0
+        self.in_calibration = False
+        self.step = 2.
+        self.last_zero_pos = None
+        self.total_expansion = 0
+        self.start_pos = []
+
+        # Register GCode Commands
+        pname = self.name.split(maxsplit=1)[-1]
+        self.gcode.register_mux_command(
+            "TEMPERATURE_PROBE_CALIBRATE", "PROBE", pname,
+            self.cmd_TEMPERATURE_PROBE_CALIBRATE,
+            desc=self.cmd_TEMPERATURE_PROBE_CALIBRATE_help
+        )
+
+        self.gcode.register_mux_command(
+            "TEMPERATURE_PROBE_ENABLE", "PROBE", pname,
+            self.cmd_TEMPERATURE_PROBE_ENABLE,
+            desc=self.cmd_TEMPERATURE_PROBE_ENABLE_help
+        )
+
+        # Register Drift Compensation Helper with probe
+        full_probe_name = "probe_eddy_current %s" % (pname,)
+        if config.has_section(full_probe_name):
+            pprobe = self.printer.load_object(config, full_probe_name)
+            self.cal_helper = EddyDriftCompensation(config, self)
+            pprobe.register_drift_compensation(self.cal_helper)
+            logging.info(
+                "%s: registered drift compensation with probe [%s]"
+                % (self.name, full_probe_name)
+            )
+        else:
+            logging.info(
+                "%s: No probe named %s configured, thermal drift compensation "
+                "disabled." % (self.name, pname)
+            )
+
+    def _temp_callback(self, read_time, temp):
+        smoothed_temp, measured_min, measured_max = self.last_measurement
+        time_diff = read_time - self.last_temp_read_time
+        self.last_temp_read_time = read_time
+        temp_diff = temp - smoothed_temp
+        adj_time = min(time_diff * self.inv_smooth_time, 1.)
+        smoothed_temp += temp_diff * adj_time
+        measured_min = min(measured_min, smoothed_temp)
+        measured_max = max(measured_max, smoothed_temp)
+        self.last_measurement = (smoothed_temp, measured_min, measured_max)
+        if self.in_calibration and smoothed_temp >= self.next_auto_temp:
+            self.printer.get_reactor().register_async_callback(
+                self._check_kick_next
+            )
+
+    def _check_kick_next(self, eventtime):
+        smoothed_temp = self.last_measurement[0]
+        if self.in_calibration and smoothed_temp >= self.next_auto_temp:
+            self.next_auto_temp = 99999999.
+            self.gcode.run_script("TEMPERATURE_PROBE_NEXT")
+
+    def get_temp(self, eventtime=None):
+        return self.last_measurement[0], self.target_temp
+
+    def _collect_sample(self, kin_pos, tool_zero_z):
+        probe = self._get_probe()
+        x_offset, y_offset, _ = probe.get_offsets()
+        speeds = self._get_speeds()
+        lift_speed, _, move_speed = speeds
+        toolhead = self.printer.lookup_object("toolhead")
+        cur_pos = toolhead.get_position()
+        # Move to probe to sample collection position
+        cur_pos[2] += self.horizontal_move_z
+        toolhead.manual_move(cur_pos, lift_speed)
+        cur_pos[0] -= x_offset
+        cur_pos[1] -= y_offset
+        toolhead.manual_move(cur_pos, move_speed)
+        return self.cal_helper.collect_sample(kin_pos, tool_zero_z, speeds)
+
+    def _prepare_next_sample(self, last_temp, tool_zero_z):
+        # Register our own abort command now that the manual
+        # probe has finished and unregistered
+        self.gcode.register_command(
+            "ABORT", self.cmd_TEMPERATURE_PROBE_ABORT,
+            desc=self.cmd_TEMPERATURE_PROBE_ABORT_help
+        )
+        probe_speed = self._get_speeds()[1]
+        # Move tool down to the resting position
+        toolhead = self.printer.lookup_object("toolhead")
+        cur_pos = toolhead.get_position()
+        cur_pos[2] = tool_zero_z + self.resting_z
+        toolhead.manual_move(cur_pos, probe_speed)
+        cnt, exp_cnt = self.sample_count, self.expected_count
+        self.next_auto_temp = last_temp + self.step
+        self.gcode.respond_info(
+            "%s: collected sample %d/%d at temp %.2fC, next sample scheduled "
+            "at temp %.2fC"
+            % (self.name, cnt, exp_cnt, last_temp, self.next_auto_temp)
+        )
+
+    def _manual_probe_finalize(self, kin_pos):
+        if kin_pos is None:
+            # Calibration aborted
+            self._finalize_drift_cal(False)
+            return
+        if self.last_zero_pos is not None:
+            z_diff = self.last_zero_pos[2] - kin_pos[2]
+            self.total_expansion += z_diff
+            logging.info(
+                "Estimated Total Thermal Expansion: %.6f"
+                % (self.total_expansion,)
+            )
+        self.last_zero_pos = kin_pos
+        toolhead = self.printer.lookup_object("toolhead")
+        tool_zero_z = toolhead.get_position()[2]
+        try:
+            last_temp = self._collect_sample(kin_pos, tool_zero_z)
+        except Exception:
+            self._finalize_drift_cal(False)
+            raise
+        self.sample_count += 1
+        if last_temp >= self.target_temp:
+            # Calibration Done
+            self._finalize_drift_cal(True)
+        else:
+            try:
+                self._prepare_next_sample(last_temp, tool_zero_z)
+                if self.sample_count == 1:
+                    self._set_bed_temp(self.cal_bed_temp)
+            except Exception:
+                self._finalize_drift_cal(False)
+                raise
+
+    def _finalize_drift_cal(self, success, msg=None):
+        self.next_auto_temp = 99999999.
+        self.target_temp = 0
+        self.expected_count = 0
+        self.sample_count = 0
+        self.step = 2.
+        self.in_calibration = False
+        self.last_zero_pos = None
+        self.total_expansion = 0
+        self.start_pos = []
+        # Unregister Temporary Commands
+        self.gcode.register_command("ABORT", None)
+        self.gcode.register_command("TEMPERATURE_PROBE_NEXT", None)
+        self.gcode.register_command("TEMPERATURE_PROBE_COMPLETE", None)
+        # Turn off heaters
+        self._set_extruder_temp(0)
+        self._set_bed_temp(0)
+        try:
+            self.cal_helper.finish_calibration(success)
+        except self.gcode.error as e:
+            success = False
+            msg = str(e)
+        if not success:
+            msg = msg or "%s: calibration aborted" % (self.name,)
+            self.gcode.respond_info(msg)
+
+    def _get_probe(self):
+        probe = self.printer.lookup_object("probe")
+        if probe is None:
+            raise self.gcode.error("No probe configured")
+        return probe
+
+    def _set_extruder_temp(self, temp, wait=False):
+        if self.cal_extruder_temp is None:
+            # Extruder temperature not configured
+            return
+        toolhead = self.printer.lookup_object("toolhead")
+        extr_name = toolhead.get_extruder().get_name()
+        self.gcode.run_script_from_command(
+            "SET_HEATER_TEMPERATURE HEATER=%s TARGET=%f"
+            % (extr_name, temp)
+        )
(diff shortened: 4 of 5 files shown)


Arksine:
For those wondering about the status of this branch, I'm currently working on rebasing and resolving conflicts against the master branch.  My goal is to push an updated version this weekend (by June 22).

smwoodward:
> For those wondering about the status of this branch, I'm currently working on rebasing and resolving conflicts against the master branch. My goal is to push an updated version this weekend (by June 22).

Are we able to move off the BTT branch of klipper?

Arksine:
I have updated this branch to bring it in sync with the current master branch.  It should not be necessary to perform a new `PROBE_DRIFT_CALIBRATE` after updating to the newest implementation, however it will be necessary to perform a new `PROBE_EDDY_CURRENT_CALIBRATE`.

Another change to be aware of is that there is no longer a `SCAN_MODE` parameter in `BED_MESH_CALIBRATE`.  Instead there are two new options for the `METHOD` parameter, `scan` and `rapid_scan`.

docs/Config_Reference.md
```diff
@@ -2401,6 +2405,65 @@ temperature sensors that are reported via the M105 command.
...
+#   default behavior is not to move the tool prior to the first
+#   manual probe.
+#calibration_bed_temp:
+#   The maximum safe bed temperature (in C) used to heat the probe
+#   during probe drift calibration.  When set, the calibration
+#   procedure will turn on the bed after the first sample is
+#   taken.  When the calibration procedure is complete the bed
+#   temperature will be set to zero.  When omitted the default
+#   behavior is not to set the bed temperature.
+#calibratation_extruder_temp:
```
bigtreetech, line 2453: should be `calibration_extruder_temp`

KevinOConnor:
Thanks for working on this.  Sorry for the delay in responding.

In general, this seems like useful functionality.  I have some high-level questions that would help me finish reviewing.

It looks like the temperature compensation algorithm works by creating 9 z mappings at z=0.05 to z=4.55 (0.500mm between each mapping).  Each of these Z mappings can translate a temperature to the expected ldc1612 sensor frequency at the given Z height.  The mappings are implemented using a quadratic formula (`expected_freq = c*temp*temp + b*temp + a`).  Each mapping is obtained by finding the least squares solution for a, b, c given all the (temperature, freq) pairs measured at the given Z height.  During runtime, the code compensates for temperature by adjusting a measured frequency - it finds the appropriate Z bucket given the current temperature and measured frequency, finds the expected frequency range for that Z bucket if the temperature was the nominal calibration temperature, and then interpolates the measured frequency back to the expected frequency at nominal temperature given those ranges.  Did I understand this system correctly?

What is the expected user workflow for this system?  Are users expected to run PROBE_DRIFT_CALIBRATE on each printer, or is it expected that most users will populate the temperature settings from some upstream source?  If the latter, do you have examples of tuned calibration settings?

In some of my previous tests, I got the impression that the ldc1612 was sensitive to both coil temperature and bed temperature.  Is it expected that: a) bed temperature isn't a big contributor to drift, b) bed temperature does contribute to drift but it is easy for users to stabilize the bed temperature prior to probing, c) bed temperature does contribute to drift but accounting for coil temperature is still better than nothing, or d) something else?

How would a typical user utilize SET_PROBE_DRIFT_ADJ_FACTOR?

Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement?  My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions.  I feel I may be missing something subtle.

Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today?  Has there been user feedback on calibration issues, accuracy tests, or similar?

Thanks again!
-Kevin

smwoodward:
> Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.
> 
> Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?

I just set this up recently off of their forked version, and what it did at each temperature measurement was to do a paper test with the nozzle, and when I clicked accept it moved and did measurements, then went back to the position and waited till the next temperature which I used the paper again to set the manual probe and it repeated.

It wasn't difficult to do, but it would be nice if there was a different way of doing it than sitting by the printer for 30+ minutes to do the paper test over and over.

Sineos:
It seems the Beacon and Cartographer are using different approaches (or I have understood them wrong)

**Basic Calibration:**
* [Beacon](https://docs.beacon3d.com/quickstart/#6-calibrate-beacon)
* [Cartographer ](https://docs.cartographer3d.com/cartographer-probe/installation-and-setup/v2-and-v3-probes-w-input-shaping/cartographer-with-input-shaper-v2-and-v3-hybrid)

Both are using the paper test and only at one temperature.


**Extended Calibration:**
* Beacon: n/a
* [Cartographer ](https://docs.cartographer3d.com/cartographer-probe/fine-tuning/temperature-differential-calibration-beta)

In Cartographer, an automated process takes various frequency samples at different temperatures and computes profiles from them. No manual paper method is required.

For Beacon, I did not find such a procedure. It seems they rely on a factory-built calibration profile.

I cannot assess the accuracy of either product, but Cartographer appears to have the advantage of not requiring the paper method. 
It is likely that applying the manual paper method multiple times in a row will result in a significant measurement error due to the inherent limitations of the human user, who is unable to perform the paper method with the precision and accuracy required.

Arksine:
Hi Kevin, thanks for the review.

> It looks like the temperature compensation algorithm works by creating 9 z mappings at z=0.05 to z=4.55 (0.500mm between each mapping). Each of these Z mappings can translate a temperature to the expected ldc1612 sensor frequency at the given Z height. The mappings are implemented using a quadratic formula (expected_freq = c*temp*temp + b*temp + a). Each mapping is obtained by finding the least squares solution for a, b, c given all the (temperature, freq) pairs measured at the given Z height. During runtime, the code compensates for temperature by adjusting a measured frequency - it finds the appropriate Z bucket given the current temperature and measured frequency, finds the expected frequency range for that Z bucket if the temperature was the nominal calibration temperature, and then interpolates the measured frequency back to the expected frequency at nominal temperature given those ranges. Did I understand this system correctly?

This is correct.

> What is the expected user workflow for this system? Are users expected to run PROBE_DRIFT_CALIBRATE on each printer, or is it expected that most users will populate the temperature settings from some upstream source? If the latter, do you have examples of tuned calibration settings?

Its expected that the user run `PROBE_DRIFT_CALIBRATE` for each printer.  I compared data from multiple BTT Eddy probes and drift was not consistent between them, so I don't believe it is going to be possible to provide a factory preset.  That said, it may be possible that future iterations of the Eddy or other probe designs based on the ldc1612 are able to deliver a preset.

> In some of my previous tests, I got the impression that the ldc1612 was sensitive to both coil temperature and bed temperature. Is it expected that: a) bed temperature isn't a big contributor to drift, b) bed temperature does contribute to drift but it is easy for users to stabilize the bed temperature prior to probing, c) bed temperature does contribute to drift but accounting for coil temperature is still better than nothing, or d) something else?

In my testing bed temperature was not a significant factor with regard drift.  Coil temperature and distance from the bed were the primary factors.  For example, below are two plots captured with the bed at 85C AND 110C respectively.   This data was captured with the bed and chamber preheated, using a printed mount that decoupled the probe from the gantry.  The mount put the coil approximately ~3mm above the bed.

85C

eddy-drift-85c-base-height


110C

eddy-drift-110c-base-height

> How would a typical user utilize SET_PROBE_DRIFT_ADJ_FACTOR?

I used it to disable thermal compensation by setting a factor of 0 during testing.  As I hinted mentioned, it may be possible for some probe designs to deliver a stock calibration and tweak the result by adjusting the factor.   I need to do more testing, but it might also be useful for tweaking calibration when switching build plates.

That said, I 100% understand if you want to remove this functionality until we receive confirmation with regard to its usage.
 
> Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.

A manual probe is necessary to remove the influence of thermal expansion from the frame and bed during the probe warming procedure.  A good calibration requires that we collect samples across the widest coil temperature range possible, so unfortunately we can't remove this influence by preheating the printer.  In the future, if its possible to use the ldc1612 to perform contact based probing, this operation could be automated. 

> Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?

BTT has users running this code based on the version of `probe_eddy_current` prior to the recent refactor.  General feedback I have received is that it works well when calibrated correctly, however the calibration procedure is difficult.

smwoodward:
@KevinOConnor Something that I wanted to as or run by you with these new probes, I did install on of the eddy's on my ratrig and I kept the Pinda probe on also, and I moved it to be a z-endstop which works for homing, but is there a way to add the capability to do things like homing and QGL/z-tilt with one probe and do the mesh with the different probe?

KevinOConnor:
> In my testing bed temperature was not a significant factor with regard drift.

Thanks for the responses!

Just to further clarify the impact of bed temperature - the two graphs you provided don't look that similar to me.  For example, at 85c bed temperature the probe at 60c seems to report 3.270Mhz while at 110c bed temperature the probe at 60c seems to report 3.269Mhz.  Granted that's small relative to the coil swing (1khz for 25c bed temperature change relative to about 5Khz for 25c probe temperature change).  I'm not sure it's small relative to our probe Z height accuracy goals though.  I think I may be missing something.

-Kevin

KevinOConnor:
@smwoodward - for questions on using Klipper please follow the contact directions at: https://www.klipper3d.org/Contact.html

-Kevin

KevinOConnor:
Thanks for working on this.  In general it seems worthwhile to add to Klipper.

My main comment is about the interface between module components.  What do you think about an alternate interface between temperature_probe.py and probe_eddy_current.py?  This PR changes probe_eddy_current to detect and register itself with the temperature_probe code (`EddyDriftCompensation:self.temp_sensor.register_calibration_helper(self)`).  Could we flip that around and have the temperature_probe code lookup probe_eddy_current and call a new `probe_eddy_current.register_temperature_compensation(self)`?

The reason I think that may be useful is that it seems the probe temperature calibration interface (TemperatureProbe to EddyDriftCompensation) is more complex than the frequency adjustment interface (EddyDriftCompensation to EddyCalibration).  Thus if we can move EddyDriftCompensation into temperature_probe.py then I think the code interfaces might be more apparent.

I think we could also move `calibration_temp` to EddyDriftCompensation.  The EddyCalibration code could call a `self.drift_comp.note_start_z_calibration()` at the start of calibration and call a `self.drift_comp.note_finish_z_calibration()` at the end of calibration (which could signal to EddyDriftCompensation to store the `calibration_temp` in the temperature_probe config section during auto save).  Thus EddyCalibration would have only a few temperature specific methods (`def register_temperature_compensation()`, `self.drift_comp.note_start_z_calibration()`, `self.drift_comp.note_finish_z_calibration()`, `self.drift_comp.adjust_freqs()`, `self.drift_comp.unadjust_freq()`).

I have a few other code comments which I'll follow up with below.

Cheers,
-Kevin

klippy/extras/probe_eddy_current.py
```diff
@@ -38,7 +41,8 @@ def load_calibration(self, cal):
         self.cal_zpos = [c[1] for c in cal]
     def apply_calibration(self, samples):
         for i, (samp_time, freq, dummy_z) in enumerate(samples):
-            pos = bisect.bisect(self.cal_freqs, freq)
+            adj_freq = self.drift_comp.adjust_freq(freq, self.cal_temp)
+            pos = bisect.bisect(self.cal_freqs, adj_freq)
```
KevinOConnor, line 45: This looks like it could cause a subtle race condition.  The `apply_calibration()` call works on a block of frequency measurements.  However, `adjust_freq()` could pull a new temperature on each individual call (as `TemperatureProbe:_temp_callback()` runs in a background thread).  That could mean that in some subtle situations a different temperature could be applied midway through a set of samples.  Probably better to consistently use a single temperature for each block of samples.

One way to do this would be to extract the temperature outside of the loop and then pass that temperature to each `adjust_freq()` call.  Another option (that might be simpler and more efficient) could be to pass all the frequencies as a list to a new `adjust_freqs()` call.

klippy/mathutil.py
```diff
@@ -135,3 +135,76 @@ def matrix_sub(m1, m2):
...
+        self.c = c
+
+    def __call__(self, xval):
+        return self.c * xval * xval + self.b * xval + self.a
+
+    def get_coefs(self):
+        return (self.a, self.b, self.c)
+
+    def __str__(self):
+        return "%f, %f, %f" % (self.a, self.b, self.c)
```
KevinOConnor, line 166: I don't think `Polynomial2d` is a good fit for mathutil.py.  We can certainly add the linear squares code and/or the matrix determinant code to mathutil.py.  (Interestingly, I think @dmbutyugin recently made a similar implementation for his generic_cartesian code.)

The `EddyDriftCompensation` code has some subtle dependencies on `Polynomianl2d` that I'm unsure would be apparent to developers looking at just mathutil.py (eg, the `str()` format becomes a user-facing api as it is written directly to the config file).

One simple solution would be to relocate `Polynomial2d` to temperature_probe.py .

klippy/extras/probe_eddy_current.py
```diff
@@ -436,5 +446,252 @@ def start_probe_session(self, gcmd):
...
+                "polynomial_coefs": [c.get_coefs() for c in polynomials],
+                "legend": ["temperature", "frequency"],
+                "data": cal_samples,
+                "start_z": .05,
+                "sample_z_dist": .5
+            }
+            with open(tmpfname, "w") as f:
+                f.write(json.dumps(out))
+        except Exception:
+            logging.exception("Failed to write %s" % (tmpfname))
```
KevinOConnor, line 633: I think we should avoid writing ad-hoc files from Klipper.  If this is just debug code, maybe we should remove it before merging in the main Klipper repo.  If it's generally useful information then I think we should look to export it via the Klipper API interface.

The ad-hoc files are hard for users to use, hard for us to document, and can become a burden to maintain as future devs wont necessarily know if they can change the contents.

klippy/extras/temperature_probe.py
```diff
@@ -0,0 +1,375 @@
...
+            adj_time = min(time_diff * self.inv_smooth_time, 1.)
+            self.smoothed_temp += temp_diff * adj_time
+            self.measured_min = min(self.measured_min, self.smoothed_temp)
+            self.measured_max = max(self.measured_max, self.smoothed_temp)
+            st = self.smoothed_temp
+            if self.in_calibration and st >= self.next_auto_temp:
+                self.next_auto_temp = 99999999.
+                self.printer.get_reactor().register_async_callback(
+                    lambda e, gc=self.gcode: gc.run_script("PROBE_DRIFT_NEXT")
+                )
```
KevinOConnor, line 93: This looks like it could have subtle race conditions (as `_temp_callback()` runs in a background thread).

The `self.temp_lock` is only ever taken here, which doesn't seem correct.  (The lock only works if other places in the code also acquire the lock.)

Many of the variables (measured_min, measured_max, in_calibration, next_auto_temp) could change mid-execution (both here and in `get_status()` etc) due to the background thread.

Also, technically, it's possible for PROBE_DRIFT_NEXT to be queued from `_temp_callback()` at a similar moment that the user manually calls `PROBE_DRIFT_NEXT` - which could cause two back-to-back executions.

docs/G-Codes.md
```diff
@@ -1415,3 +1422,34 @@ command will probe the points specified in the config and then make independent
...
 details on the optional probe parameters. The optional `HORIZONTAL_MOVE_Z`
 value overrides the `horizontal_move_z` option specified in the config file.
+
+### [temperature_probe]
+
+The following commands are available when a
+[temperature_probe config section](Config_Reference.md#temperature_probe)
+is enabled.
+
+#### PROBE_DRIFT_CALIBRATE
```
KevinOConnor, line 1432: What about `TEMPERATURE_PROBE_CALIBRATION` instead of `PROBE_DRIFT_CALIBRATE`?  Putting the module name in the command helps reduce the chance of command conflicts in the future.

docs/G-Codes.md
```diff
@@ -1006,6 +1006,13 @@ appropriate DRIVE_CURRENT for the sensor. After running this command
 use the SAVE_CONFIG command to store that new setting in the
 printer.cfg config file.
 
+#### SET_PROBE_DRIFT_ADJ_FACTOR
+`SET_PROBE_DRIFT_ADJ_FACTOR PROBE=<config_name> FACTOR=<new_factor>`:
+When thermal drift compensation is calibrated this command may be
+used to set the `drift_adj_factor` during runtime.  This value
+is a multiplier applied to the calculated drift compensation and may be
+used to fine tune the result.
```
KevinOConnor, line 1014: For what it is worth, this command seems very subtle to me.  Might be better to hold off on it until we know it is needed.  Similar with the config option.

Or another alternative might be to have something like a `PROBE_TEMPERATURE_ENABLE` command to let users manually enable/disable temperature_probe without the subtle fine tuning.

klippy/extras/temperature_probe.py
```diff
@@ -0,0 +1,375 @@
...
+        self.in_calibration = True
+        self.cal_helper.start_calibration()
+        self.target_temp = target_temp
+        self.step = step
+        self.sample_count = 0
+        self.expected_count = expected_count
+        # If configured move to heating position and turn on extruder
+        try:
+            self._move_to_start()
+        except Exception:
```
KevinOConnor, line 317: Very minor, but this should probably be `except self.printer.config_error` (as above).  An internal coding error probably shouldn't be caught here.

KevinOConnor (commented):
Thanks again - see my comments below.

-Kevin

bigtreetech:
> @KevinOConnor Something that I wanted to as or run by you with these new probes, I did install on of the eddy's on my ratrig and I kept the Pinda probe on also, and I moved it to be a z-endstop which works for homing, but is there a way to add the capability to do things like homing and QGL/z-tilt with one probe and do the mesh with the different probe?

This is already a standard part of how klipper works but as Kevin said, this is not the forum to discuss questions unrelated to the specifics of this PR which are dealing with thermal comp.

Arksine:
> Just to further clarify the impact of bed temperature - the two graphs you provided don't look that similar to me. For example, at 85c bed temperature the probe at 60c seems to report 3.270Mhz while at 110c bed temperature the probe at 60c seems to report 3.269Mhz. Granted that's small relative to the coil swing (1khz for 25c bed temperature change relative to about 5Khz for 25c probe temperature change). I'm not sure it's small relative to our probe Z height accuracy goals though. I think I may be missing something.

I can explain the discrepancy.  When collecting data it became immediately apparent that I needed to decouple the probe from the gantry to remove the influence of thermal expansion.  BTT designed a mount that encases the probe and places the coil ~3mm above the surface.  My collection procedure was to preheat the chamber, connect the probe and place it roughly in the center of the bed, collect data, then remove when done to allow the probe to cool down. 

Since the probe isn't fixed, small deviations can change the nominal frequency.  Conditions such as the PEI being slightly thicker in one area, imperfections in the printed mount, imperfections in the build plate, etc can all influence then nominal frequency.  Put differently, I could collect two data sets at 110C and they would show the same minor deviation.

> My main comment is about the interface between module components. What do you think about an alternate interface between temperature_probe.py and probe_eddy_current.py? This PR changes probe_eddy_current to detect and register itself with the temperature_probe code (EddyDriftCompensation:self.temp_sensor.register_calibration_helper(self)). Could we flip that around and have the temperature_probe code lookup probe_eddy_current and call a new probe_eddy_current.register_temperature_compensation(self)?

Sure, that wont be a problem.


> I think we could also move calibration_temp to EddyDriftCompensation. The EddyCalibration code could call a self.drift_comp.note_start_z_calibration() at the start of calibration and call a self.drift_comp.note_finish_z_calibration() at the end of calibration (which could signal to EddyDriftCompensation to store the calibration_temp in the temperature_probe config section during auto save). Thus EddyCalibration would have only a few temperature specific methods (def register_temperature_compensation(), self.drift_comp.note_start_z_calibration(), self.drift_comp.note_finish_z_calibration(), self.drift_comp.adjust_freqs(), self.drift_comp.unadjust_freq()).

Sound good.  I'll work on that change.  In addition, I'll address each review comment individually as I go along.

klippy/extras/probe_eddy_current.py
Arksine: One thing I considered was moving the call to `adjust_freq()` from `apply_calibration` to `EddyCalibration.freq_to_height()`.  This would reduce overhead from performing the adjustment on all samples, with the downside that the adjustment would't be applied to the z values in the dump command.

That said, I'm fine with capturing the probe temperature outside of the loop and applying it to all samples.

klippy/extras/probe_eddy_current.py
Arksine: I'll remove it.  Its purpose was primarily for collecting data from other users testing the module.

klippy/extras/temperature_probe.py
Arksine: Ah, I must have had async programming on my mind.  FWIW, its probably not a great plan to acquire a thread lock on the main thread since acquisition can be a blocking operation.  I wonder if it would be a good idea to create a wrapper around the lock using the reactor, testing if the lock is held and pausing until it isn't.

docs/G-Codes.md
Arksine: I'm good with that.

klippy/mathutil.py
Arksine: Sounds good.

docs/G-Codes.md
Arksine: Sounds good.

klippy/extras/probe_eddy_current.py
KevinOConnor: I think it would be helpful if the API server continued to export a Z value that is temperature adjusted.

klippy/extras/temperature_probe.py
KevinOConnor: I'm not sure the reactor can help much as it only has limited thread safety itself (only `register_async_callback()` can be called from the background thread) and the background thread isn't running a reactor (only the main thread has it).

One possible alternative is to store groups of values in tuples, as Python then assures each assignment is atomic.  The clocksync.py code does this.

Something like (totally untested):
```python
    def _temp_callback(self, read_time, temp):
        time_diff = read_time - self.last_temp_read_time
        self.last_temp_read_time = read_time
        smoothed_temp, measured_min, measured_max = self.last_measurement
        temp_diff = temp - smoothed_temp
        adj_time = min(time_diff * self.inv_smooth_time, 1.)
        smoothed_temp += temp_diff * adj_time
        measured_min = min(measured_min, smoothed_temp)
        measured_max = max(measured_max, smoothed_temp)
        self.last_measurement = (smoothed_temp, measured_min, measured_max)
        if self.in_calibration and smoothed_temp >= self.next_auto_temp:
            self.printer.get_reactor().register_async_callback(
                self._check_kick_next)

    def _check_kick_next(self, eventtime):
        smoothed_temp, measured_min, measured_max = self.last_measurement
        if self.in_calibration and smoothed_temp >= self.next_auto_temp:
            self.next_auto_temp = 99999999.
            self.gcode.run_script("TEMPERATURE_PROVE_NEXT")
```

The idea here is that smoothed_temp, measured_min, measured_max are always atomically written as a group, and thus any readers will get those values as a group (never a mix from two possible `_temp_callback()` invocations).  Also, a "dirty read" of self.in_calibration and self.next_auto_temp isn't an issue as long as the values are rechecked in the main thread.

klippy/mathutil.py
dmbutyugin: Yes, the code can be found here
https://github.com/dmbutyugin/klipper/blob/5f4df09c99d9e92a3f9896c9e8107659d0efc0d0/klippy/mathutil.py#L143-L152
and the calculation of position using pseudoinverse:
https://github.com/dmbutyugin/klipper/blob/generic-cartesian/klippy/kinematics/generic_cartesian.py#L124-L128
which is also an least squares problem solution. I think this one is a bit less verbose, so feel free to use, but up to you.

Arksine:
I have pushed some updates, the changes of which are outlined below:

- moved the `Polynomial2D` class from `mathutil.py` to `temperature_probe.py`
- moved the `EddyDriftCompensation` class from `probe_eddy_current.py` to `temperature_probe.py`.  The `probe_eddy_current` module now contains a "dummy" class serves as the default when no temperature probe is registered.
- Reversed registration logic.  The `temperature_probe` now registers the compensation helper with the underlying probe implementation by calling its `register_drift_compensation()` method.
- Moved calibration temperature tracking from the `probe_eddy_current.EddyCalibration` class to `temperature_probe.EddyDriftCompensation`.  Eddy Probe Calibration calls the suggested `note_z_calibration_start()` and `note_z_calibration_finish()` methods of the compensation helper.
- Use one temperature reading for each batch of ldc samples received when applying drift compensation.
- Remove the debug file output
- Remove the thread lock in `TempertureProbe`.  Use the suggested method of storing the smoothed, min, and max tempertures in a tuple for atomic writes.  Implement the suggested `_check_kick_next()` callback to avoid a potential race when calling "TEMPERATURE_PROBE_NEXT"
- Rename all gcode commands prefixed withe `PROBE_DRIFT_` to `TEMPERATURE_PROBE_`
- Remove the `SET_PROBE_DRIFT_ADJ_FACTOR` gcode command
- Implement `TEMPERATURE_PROBE_ENABLE` gcode command.  Report the current enabled state in `TemperatureProbe.get_status()`
- In the `TEMPERATURE_PROBE_CALIBRATE` gcode handler, catch `self.printer.command_error` when calling `_move_to_start()`.

I believe that covers everything.  This change moves all drift compensation related items in the autosave from the `probe_eddy_current` section to the `temperature probe` section.  Testers that don't wish to redo their calibration can manually update the section (generally it isn't recommended to modify the autosave, but it seems appropriate for this purpose).  For example, the autosave section should now contain a snippet that looks like the following.

```ini
#*# [temperature_probe btt_eddy]
#*# calibration_temp = 42.919721
#*# drift_calibration =
#*# 	3265957.923222, -161.557392, 0.448391
#*# 	3256524.613344, -106.776906, 0.145617
#*# 	3249162.938671, -68.684932, -0.075612
#*# 	3243413.959281, -44.068844, -0.206610
#*# 	3238864.481265, -26.708818, -0.301046
#*# 	3235263.558343, -15.621579, -0.355331
#*# 	3232252.151603, -4.210931, -0.421378
#*# 	3229739.335973, 7.083410, -0.497587
#*# 	3227831.886549, 9.409186, -0.497820
#*# drift_calibration_min_temp = 32.18233127179636
```

KevinOConnor:
Thanks.  It looks fine to me.  If there are no further comments I'll look to commit in a few days.  I guess I'll look to do a "squash and merge".

-Kevin

klippy/extras/temperature_probe.py
```diff
@@ -0,0 +1,718 @@
...
+            )
+        else:
+            logging.info(
+                "%s: No probe named %s configured, thermal drift compensation "
+                "disabled." % (self.name, pname)
+            )
+
+    def _temp_callback(self, read_time, temp):
+        if not temp:
+            return
```
KevinOConnor, line 166: FWIW, this test looks a bit odd to me as no sensor should be passing on a bogus zero value (if a zero value is relayed it would indicate zero Celsius).  If there are sensors passing on bad data we should definitely look to fix those sensors.

-Kevin

Arksine: Ok, I'll remove it.  IIRC I included a similar check to the temperature callback in `temperature_sensor.py` when adding support for min/max temps. It was reported that some sensors would report an initial value of 0 which would impact the minimum stored value.  I agree that its better to debug the sensor reporting that value.

Arksine:
> Thanks. It looks fine to me. If there are no further comments I'll look to commit in a few days. I guess I'll look to do a "squash and merge".

Sounds good.  I just removed the `if temp` check, and did a fixup on the commits so you can rebase and merge if that is your preference.

KevinOConnor:
Thanks.  I committed this change.  I was just giving a few days to see if others had comments.

Cheers,
-Kevin

smwoodward:
> > > Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.
> > > Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?
> > 
> > 
> > I just set this up recently off of their forked version, and what it did at each temperature measurement was to do a paper test with the nozzle, and when I clicked accept it moved and did measurements, then went back to the position and waited till the next temperature which I used the paper again to set the manual probe and it repeated.
> > It wasn't difficult to do, but it would be nice if there was a different way of doing it than sitting by the printer for 30+ minutes to do the paper test over and over.
> 
> I totally agree. the paper testing requires opening the chamber door, which causes rapid loss of chamber temperature and introduces rapid changes in coil temperature.

I don't think the thermal loss is that great to the coil temperature from what I saw, I can say that the thermal loss of the air in the chamber would extend the required time in heating up the coil as the chamber environment would need to heat back up to keep increasing the coil temp.  There would still be a good amount of radiation heat transfer because of the distance between the bed and coil, but the convection transfer would take time to regain the temperature once the chamber is closed again.

qqqwqk:
> > > > Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.
> > > > Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?
> > > 
> > > 
> > > I just set this up recently off of their forked version, and what it did at each temperature measurement was to do a paper test with the nozzle, and when I clicked accept it moved and did measurements, then went back to the position and waited till the next temperature which I used the paper again to set the manual probe and it repeated.
> > > It wasn't difficult to do, but it would be nice if there was a different way of doing it than sitting by the printer for 30+ minutes to do the paper test over and over.
> > 
> > 
> > I totally agree. the paper testing requires opening the chamber door, which causes rapid loss of chamber temperature and introduces rapid changes in coil temperature.
> 
> I don't think the thermal loss is that great to the coil temperature from what I saw, I can say that the thermal loss of the air in the chamber would extend the required time in heating up the coil as the chamber environment would need to heat back up to keep increasing the coil temp. There would still be a good amount of radiation heat transfer because of the distance between the bed and coil, but the convection transfer would take time to regain the temperature once the chamber is closed again.

I'm just raising an idea: maybe thermal expansion of the frame and bed should not be taken into account when calibrating the probe, which would simplify the manual work of calibration(As far as I know: Not even Hannibal would try to cook his fingers). 
Thermal expansion should be considered as a separate item (perhaps we should consider using z-endstop to try to detect/calculate thermal expansion)

smwoodward:
> > > > > Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.
> > > > > Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?
> > > > 
> > > > 
> > > > I just set this up recently off of their forked version, and what it did at each temperature measurement was to do a paper test with the nozzle, and when I clicked accept it moved and did measurements, then went back to the position and waited till the next temperature which I used the paper again to set the manual probe and it repeated.
> > > > It wasn't difficult to do, but it would be nice if there was a different way of doing it than sitting by the printer for 30+ minutes to do the paper test over and over.
> > > 
> > > 
> > > I totally agree. the paper testing requires opening the chamber door, which causes rapid loss of chamber temperature and introduces rapid changes in coil temperature.
> > 
> > 
> > I don't think the thermal loss is that great to the coil temperature from what I saw, I can say that the thermal loss of the air in the chamber would extend the required time in heating up the coil as the chamber environment would need to heat back up to keep increasing the coil temp. There would still be a good amount of radiation heat transfer because of the distance between the bed and coil, but the convection transfer would take time to regain the temperature once the chamber is closed again.
> 
> I'm just raising an idea: maybe thermal expansion of the frame and bed should not be taken into account when calibrating the probe, which would simplify the manual work of calibration(As far as I know: Not even Hannibal would try to cook his fingers). Thermal expansion should be considered as a separate item (perhaps we should consider using z-endstop to try to detect/calculate thermal expansion)

Perhaps someone else can speak more about exactly what is going on, but from my understanding the temperature calibration isn't more about thermal expansion as the readings the eddy current sensor change based on temperature.  It's a bit different than thermal expansion which would cause the physical parts to change shape.  It's more that everything else being equal (sensor being exactly the same distance away from the metal surface no matter the temperature) that if the sensor is reading 2.7 gigawatts (or what ever it is measuring) at 24*, that if the sensor alone goes to 35* it will read 3.7 gigawatts, but nothing else has changed except for the temp of the sensor.  Thermal expansion would cause the gap to physically change between the two items.

qqqwqk:
> > > > > > Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.
> > > > > > Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?
> > > > > 
> > > > > 
> > > > > I just set this up recently off of their forked version, and what it did at each temperature measurement was to do a paper test with the nozzle, and when I clicked accept it moved and did measurements, then went back to the position and waited till the next temperature which I used the paper again to set the manual probe and it repeated.
> > > > > It wasn't difficult to do, but it would be nice if there was a different way of doing it than sitting by the printer for 30+ minutes to do the paper test over and over.
> > > > 
> > > > 
> > > > I totally agree. the paper testing requires opening the chamber door, which causes rapid loss of chamber temperature and introduces rapid changes in coil temperature.
> > > 
> > > 
> > > I don't think the thermal loss is that great to the coil temperature from what I saw, I can say that the thermal loss of the air in the chamber would extend the required time in heating up the coil as the chamber environment would need to heat back up to keep increasing the coil temp. There would still be a good amount of radiation heat transfer because of the distance between the bed and coil, but the convection transfer would take time to regain the temperature once the chamber is closed again.
> > 
> > 
> > I'm just raising an idea: maybe thermal expansion of the frame and bed should not be taken into account when calibrating the probe, which would simplify the manual work of calibration(As far as I know: Not even Hannibal would try to cook his fingers). Thermal expansion should be considered as a separate item (perhaps we should consider using z-endstop to try to detect/calculate thermal expansion)
> 
> Perhaps someone else can speak more about exactly what is going on, but from my understanding the temperature calibration isn't more about thermal expansion as the readings the eddy current sensor change based on temperature. It's a bit different than thermal expansion which would cause the physical parts to change shape. It's more that everything else being equal (sensor being exactly the same distance away from the metal surface no matter the temperature) that if the sensor is reading 2.7 gigawatts (or what ever it is measuring) at 24*, that if the sensor alone goes to 35* it will read 3.7 gigawatts, but nothing else has changed except for the temp of the sensor. Thermal expansion would cause the gap to physically change between the two items.

Heating the coil sensor will also heat the bed and frame, and during this process, the distance between the sensor and the bed cannot remain constant.

smwoodward:
> > > > > > > Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.
> 
> > > > > > > Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?
> 
> > > > > > 
> 
> > > > > > 
> 
> > > > > > I just set this up recently off of their forked version, and what it did at each temperature measurement was to do a paper test with the nozzle, and when I clicked accept it moved and did measurements, then went back to the position and waited till the next temperature which I used the paper again to set the manual probe and it repeated.
> 
> > > > > > It wasn't difficult to do, but it would be nice if there was a different way of doing it than sitting by the printer for 30+ minutes to do the paper test over and over.
> 
> > > > > 
> 
> > > > > 
> 
> > > > > I totally agree. the paper testing requires opening the chamber door, which causes rapid loss of chamber temperature and introduces rapid changes in coil temperature.
> 
> > > > 
> 
> > > > 
> 
> > > > I don't think the thermal loss is that great to the coil temperature from what I saw, I can say that the thermal loss of the air in the chamber would extend the required time in heating up the coil as the chamber environment would need to heat back up to keep increasing the coil temp. There would still be a good amount of radiation heat transfer because of the distance between the bed and coil, but the convection transfer would take time to regain the temperature once the chamber is closed again.
> 
> > > 
> 
> > > 
> 
> > > I'm just raising an idea: maybe thermal expansion of the frame and bed should not be taken into account when calibrating the probe, which would simplify the manual work of calibration(As far as I know: Not even Hannibal would try to cook his fingers). Thermal expansion should be considered as a separate item (perhaps we should consider using z-endstop to try to detect/calculate thermal expansion)
> 
> > 
> 
> > Perhaps someone else can speak more about exactly what is going on, but from my understanding the temperature calibration isn't more about thermal expansion as the readings the eddy current sensor change based on temperature. It's a bit different than thermal expansion which would cause the physical parts to change shape. It's more that everything else being equal (sensor being exactly the same distance away from the metal surface no matter the temperature) that if the sensor is reading 2.7 gigawatts (or what ever it is measuring) at 24*, that if the sensor alone goes to 35* it will read 3.7 gigawatts, but nothing else has changed except for the temp of the sensor. Thermal expansion would cause the gap to physically change between the two items.
> 
> 
> 
> heating the coil sensors also heat the bed and frame simultaneously, and during this process the distance between the sensor and the bed does not remain constant.
> 
> 

That's why you do the paper test to make sure that the nozzle is still at the distance it should be, then it takes readings of the eddy currents at the exact same spot to know how they change. From my understanding the sensor readings change much more drastically due to temperature than thermal expansion affects the measurements.

qqqwqk:
> > > > > > > > Why does PROBE_DRIFT_CALIBRATE perform a manual probe at the start of each temperature measurement? My initial thought would be to perform the Z/freq/temp sampling at identical Z carriage positions. I feel I may be missing something subtle.
> > 
> > 
> > > > > > > > Do you know if this PROBE_DRIFT_CALIBRATE code is in wide use today? Has there been user feedback on calibration issues, accuracy tests, or similar?
> > 
> > 
> > > > > > >
> > 
> > 
> > > > > > >
> > 
> > 
> > > > > > > I just set this up recently off of their forked version, and what it did at each temperature measurement was to do a paper test with the nozzle, and when I clicked accept it moved and did measurements, then went back to the position and waited till the next temperature which I used the paper again to set the manual probe and it repeated.
> > 
> > 
> > > > > > > It wasn't difficult to do, but it would be nice if there was a different way of doing it than sitting by the printer for 30+ minutes to do the paper test over and over.
> > 
> > 
> > > > > >
> > 
> > 
> > > > > >
> > 
> > 
> > > > > > I totally agree. the paper testing requires opening the chamber door, which causes rapid loss of chamber temperature and introduces rapid changes in coil temperature.
> > 
> > 
> > > > >
> > 
> > 
> > > > >
> > 
> > 
> > > > > I don't think the thermal loss is that great to the coil temperature from what I saw, I can say that the thermal loss of the air in the chamber would extend the required time in heating up the coil as the chamber environment would need to heat back up to keep increasing the coil temp. There would still be a good amount of radiation heat transfer because of the distance between the bed and coil, but the convection transfer would take time to regain the temperature once the chamber is closed again.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > I'm just raising an idea: maybe thermal expansion of the frame and bed should not be taken into account when calibrating the probe, which would simplify the manual work of calibration(As far as I know: Not even Hannibal would try to cook his fingers). Thermal expansion should be considered as a separate item (perhaps we should consider using z-endstop to try to detect/calculate thermal expansion)
> > 
> > 
> > > 
> > 
> > 
> > > Perhaps someone else can speak more about exactly what is going on, but from my understanding the temperature calibration isn't more about thermal expansion as the readings the eddy current sensor change based on temperature. It's a bit different than thermal expansion which would cause the physical parts to change shape. It's more that everything else being equal (sensor being exactly the same distance away from the metal surface no matter the temperature) that if the sensor is reading 2.7 gigawatts (or what ever it is measuring) at 24*, that if the sensor alone goes to 35* it will read 3.7 gigawatts, but nothing else has changed except for the temp of the sensor. Thermal expansion would cause the gap to physically change between the two items.
> > 
> > 
> > heating the coil sensors also heat the bed and frame simultaneously, and during this process the distance between the sensor and the bed does not remain constant.
> 
> That's why you do the paper test to make sure that the nozzle is still at the distance it should be, then it takes readings of the eddy currents at the exact same spot to know how they change. From my understanding the sensor readings change much more drastically due to temperature than thermal expansion affects the measurements.

I'm just expressing my opinion and suggesting a way to make it easier to use. After all, doing paper calibration multiple times with the heated bed on is a painful process.

Arksine:
> I'm just expressing my opinion and suggesting a way to make it easier to use. After all, doing paper calibration multiple times with the heated bed on is a painful process.

I understand.  Unfortunately the manual probe is required to take valid samples.  Without the manual probe the distance between the probe's coil and the bed will change due to thermal expansion.   This would result in a calibration that is invalid.

As mentioned previously in this thread, if its possible to implement contact probing using the ldc1612 then the manual probe can be replaced with an automatic probe.  Until then, the current solution is the only workable solution for a probe mounted to the gantry.

Another alternative would be to calibrate the probe before mounting/installation.  This would require designing a printed holder that sits on the bed, with a single motor capable of moving the probe 4 to 5 mm vertically.  This would allow users to preheat the bed/chamber, then insert the probe/holder and perform the calibration.

Sineos:
> Another alternative would be to calibrate the probe before mounting/installation. This would require designing a printed holder that sits on the bed, with a single motor capable of moving the probe 4 to 5 mm vertically. This would allow users to preheat the bed/chamber, then insert the probe/holder and perform the calibration.

Is this so much different from what Cartographer does?
It seems to take the profile data at various heights and various temperatures via a macro

Arksine:
> Is this so much different from what Cartographer does?
It seems to take the profile data at various heights and various temperatures via a macro

Its possible, I haven't looked at the details of the Cartographer implementation.  Since their extension seemed to be heavily derived from the beacon extension  I assumed they performed thermal compensation in a similar fashion, where the probe is shipped with some presets in flash memory.

Sineos:
> I assumed they performed thermal compensation in a similar fashion

I have not seen something similar for the beacon. Details here: https://docs.cartographer3d.com/cartographer-probe/fine-tuning/temperature-differential-calibration-beta

qqqwqk:
> 
> As mentioned previously in this thread, if its possible to implement contact probing using the ldc1612 then the manual probe can be replaced with an automatic probe. Until then, the current solution is the only workable solution for a probe mounted to the gantry.

Maybe we can try using adxl345-probe?
https://github.com/jniebuhr/adxl345-probe
Does klipper support using multiple probes now?

Arksine:
> I have not seen something similar for the beacon. Details here: https://docs.cartographer3d.com/cartographer-probe/fine-tuning/temperature-differential-calibration-beta

Interesting.  That method appears to me that their method will introduce error from thermal expansion.  Personally I think performing ~20 manual probes in 30 minutes is preferable to waiting hours for a calibration procedure.  They also note that its required to re-calibrate the "model", which seems incorrect to me.  Thermal drift calibration should have no impact on the base calibration.

> Maybe we can try using adxl345-probe?
https://github.com/jniebuhr/adxl345-probe
Does klipper support using multiple probes now?

No, I don't believe that klipper supports multiple probes.  I was referring to the possibility of using the ldc1612 to detect contact.  The beacon apparently supports this method of probing.  Presumably such an implementation would support both a contact and a non-contact probing mode.

One other possibility would be to temporarily fix a switch to the tool that triggers just before nozzle contact.  It could be configured as a `z endstop`, and the module could replace the manual probe with "z home".  We would need to know its offsets, so it would effectively be a temporary secondary probe.

qqqwqk:
> 
> Interesting. It appears to me that method will introduce error from thermal expansion. Personally I think performing ~20 manual probes in 30 minutes is preferable to waiting hours for a calibration procedure. They also note that its required to re-calibrate the "model", which seems incorrect to me. Thermal drift calibration should have no impact on the base calibration.

In fact, if you read the installation process rigorous
(https://docs.cartographer3d.com/cartographer-probe/installation-and-setup/v2-and-v3-probes-w-input-shaping/cartographer-with-input-shaping-v2-and-v3-hybrid),
 you will find that this detection is actually achieved by using the accelerometer's tap detection function( 
they named it "tap-probe" 
https://github.com/Cartographer3D/cartographer-klipper/blob/master/idm.py#L176). 
Unfortunately, this function has not been merged into the master branch of klipper.

Sineos:
> Personally I think performing ~20 manual probes in 30 minutes is preferable to waiting hours for a calibration procedure.

My point is less the comfort argument. Doing 20 manual probes once does not seem too bad.
A while back, I did a small measuring campaign on manual probing and I really tried hard, but the accuracy and repeatability of the manual probes have been really bad. 
Actually, my goal was to compare the "paper method" with using a feeler gauge and IIRC, the paper method was even better than the feeler gauge. Not sure that I kept the data somewhere, though.

Of course, I cannot judge the impact of this measuring error in the probe's calibration.

> One other possibility would be to temporarily fix a switch to the tool that triggers just before nozzle contact. It could be configured as a `z endstop`, and the module could replace the manual probe with "z home". We would need to know its offsets, so it would effectively be a temporary secondary probe.

The relative position between switch and nozzle would be affected by thermal expansion as well. Wouldn't this not just introduce another kind of error?

qqqwqk:
I just installed voron-tap and btt-eddy on the toolhead 
so maybe we can make some changes

aaronevers:
If the issue is coil drift with coil temperature changes, why heat the chamber at all?  Seems like you could eliminate a lot of error if you could heat the coil directly and measure the drift without moving the toolhead at all.  Perhaps a future version could include a small heater inside the Eddy housing.

PR #6605, opened 2024-05-24
Internal rework of PrinterProbe class

KevinOConnor:
This is an internal refactoring of the probe.py PrinterProbe class.  Basically the goal is to split up the functionality in that class into a series of more manageable classes (ProbeCommandHelper, ProbeSessionHelper, HomingViaProbeHelper, and ProbeOffsetsHelper).  The goal is to separate functionality so that it is easier to add features in the future.  There should be no user visible changes as part of this PR.

@Arksine - fyi.

-Kevin

### DIFF
--- a/klippy/extras/axis_twist_compensation.py
+++ b/klippy/extras/axis_twist_compensation.py
@@ -38,10 +38,13 @@ class AxisTwistCompensation:
 
         # setup calibrater
         self.calibrater = Calibrater(self, config)
+        # register events
+        self.printer.register_event_handler("probe:update_results",
+                                            self._update_z_compensation_value)
 
-    def get_z_compensation_value(self, pos):
+    def _update_z_compensation_value(self, pos):
         if not self.z_compensations:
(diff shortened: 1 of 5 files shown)


KevinOConnor:
I have made some notable changes to this branch.  The goal is still to split up the PrinterProbe() implementation into helper classes.  This branch now also reworks the probe implementations (bltouch, probe_eddy_current, smart_effector, probe) to be the main interface point for probing.  That is, the users of the probe (eg, bed_calibrate) will now see a probe implementation class (eg, `BLTouchProbe`) as the class registered as the `printer.probe` object.  This allows the probe implementations more control over probing actions, and allows for an improved interface between probe users and probe implementations.

I don't have a good test setup for bltouch, smart_effector, nor axis_twist_compensation.  Tests from others would be helpful.

The main goal of this PR is in preparation for improved probe_eddy_current support - which I'll place in a separate PR.

-Kevin

PR #6607, opened 2024-05-25
Add support for STM32F031x6 which is the 32 KB version of the STM32F032 MCU

eliasbakken:
Recore A7 and A8 has the 32 KB flash version of the STM32F031 chip, while Recore A6 has the 16 KB version. This PR adds support for choosing how much flash is available in the config files. This is necessary in order to add support for those Recore boards which will come as a separate PR once this has been merged. 


KevinOConnor:
Thanks.  Some feedback:
1. The i2c changes don't look correct - the pins should be ordered as SCL,SDA (in both the enumeration description and in the `i2c_bus[]` array).  I'm guessing you got the order backwards?
2. We generally don't track flash sizes for each variant of a chip, as there are too many variants out there.  We generally just define the flash size to the largest reasonable size available for that class of chip.  So, probably best to just increase `FLASH_SIZE` to 0x8000 for all stm32f031 chips?

Cheers,
-Kevin

eliasbakken:
OK, I've changed the order for the I2C pins and removed the check for the flash size. That also removes the need for the extra MCU config file, so I've taken that out of the PR.

KevinOConnor:
Thanks.

-Kevin

PR #6609, opened 2024-05-31
RFC: tmc/motan: calculate mscnt with stepper phase

nefelim4ag:
What I'm trying to do:
While working on phase stepping and messing around with accelerometer data against stepper driver phase, I assumed phase is equal to driver `mscnt` or `1023 - mscnt`, to try to map accelerometer data specific micro steps or analyze the log instead of gluing together adxl csv files.

TLDR, I went in, and checked, and they do not match.
Debug some things, and got that this is caused by mcu_phase_offset, which is subtracted in tmc.py.
```
phase = int(float(driver_phase) / 1024 * phases + .5) % phases
moff = (phase - stepper.get_mcu_position()) % phases
self.mcu_phase_offset = moff
```
Found that it is subtracted again in readlog.py, and should do the trick, but looks like it is not.
```
return (step_pos - self.mcu_phase_offset) % self.phases
```
I don't know the initial intentions for the stepper phase code or how it works with the end-stop phase, angle sensor,
so I can not claim inequality to driver mscnt as a bug (and not sure I'm able to fix it properly :D).

*I used code for stallguard tracking, to sample mscnt from driver

Some graphs:
Machine enabled, phase is +1 from mscnt, okay:
image
Machine homed, do not match at all:
broken_stepper_phase

So, I tried to export mscnt from the same code and compute it.
So, the last mscnt line is dumped from the driver, and the second from the bottom is computed by the stepper position.
fixed_phase
fixed_phase_256
```
./scripts/motan/motan_graph.py -s 0 -d 20 --segment-time=0.01 ./mscnt-test -g '[["trapq(toolhead,x)"], ["step_phase(tmc5160 stepper_x,microstep)"], ["step_phase(tmc5160 stepper_x)"], ["step_phase(tmc5160 stepper_x,mscnt)"], ["stallguard(stepper_x,mscnt)"]]'
```

So, for now, this is mostly RFC, I can successfully extract needed data now, but it looks strange that the stepper phase, which is based on TMC mscnt, does not match the actual mscnt.

Thanks

### DIFF
--- a/klippy/extras/tmc.py
+++ b/klippy/extras/tmc.py
@@ -237,6 +237,8 @@ class TMCCommandHelper:
         self.read_registers = self.read_translate = None
         self.toff = None
         self.mcu_phase_offset = None
+        self.tmc_mscnt = None
+        self.mcu_phas_offset_pos = None
         self.stepper = None
         self.stepper_enable = self.printer.load_object(config, "stepper_enable")
         self.printer.register_event_handler("stepper:sync_mcu_position",
@@ -328,6 +330,7 @@ class TMCCommandHelper:
             return
         try:
             driver_phase = self._query_phase()
+            self.tmc_mscnt = driver_phase
         except self.printer.command_error as e:
             logging.info("Unable to obtain tmc %s phase", self.stepper_name)
             self.mcu_phase_offset = None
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Hi.  For what it is worth, I don't understand what this Pull Request does.  That is, I'm not sure what functionality it provides nor the tasks that users will accomplish with it.

If you have questions on how the existing `mcu_phase_offset` and `phase_offset_position` work (or you think there is a defect) then I'd recommend opening up a new thread on Klipper Discourse.  I'll try to answer the questions (and/or look into the defect) there.

Cheers,
-Kevin

nefelim4ag:
That would be very kind of you,
thanks
JFYI: https://klipper.discourse.group/t/motion-analysis-stepper-phase-what-is-it/16789

PR #6610, opened 2024-06-01
Add ldc1612 support for "METHOD=scan" probing

KevinOConnor:
This is an alternate implementation to the "scan" mode introduced in PR #6558 .  The main difference is that this PR refactors the internal probe classes with a goal to "flatten" some of those interactions between classes.

This PR is on top of PR #6605 .

@Arksine - the main goal of this PR is to demonstrate a possible internal interface between the classes for the "detailed scan" mode.  That is, the toolhead movements still go through ProbePointsHelper, but now `PrinterEddyProbe` can detect a "scan" request and substitute a different probe implementation.  Also, the same sample gathering code is used for both "descend until trigger" and "scan probing" modes.  I understand your PR had improved user options, improved documentation, and other improvements.  Happy to change this PR further, or go with an alternate implementation.

Let me know your thoughts.
-Kevin

### DIFF
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -128,14 +128,16 @@ class ProbeCommandHelper:
         fo_gcmd = gcode.create_gcode_command("", "", fo_params)
         # Probe bed sample_count times
         probe_session = self.probe.start_probe_session(fo_gcmd)
-        positions = []
-        while len(positions) < sample_count:
+        probe_num = 0
+        while probe_num < sample_count:
             # Probe position
-            pos = probe_session.run_probe(fo_gcmd)
-            positions.append(pos)
+            probe_session.run_probe(fo_gcmd)
+            probe_num += 1
             # Retract
+            pos = toolhead.get_position()
             liftpos = [None, None, pos[2] + params['sample_retract_dist']]
             self._move(liftpos, params['lift_speed'])
+        positions = probe_session.pull_probed_results()
         probe_session.end_probe_session()
         # Calculate maximum, minimum and average values
         max_value = max([p[2] for p in positions])
@@ -259,6 +261,7 @@ class ProbeSessionHelper:
                                              minval=0)
         # Session state
         self.multi_probe_pending = False
+        self.results = []
         # Register event handlers
         self.printer.register_event_handler("gcode:command_error",
                                             self._handle_command_error)
@@ -276,10 +279,12 @@ class ProbeSessionHelper:
             self._probe_state_error()
         self.mcu_probe.multi_probe_begin()
         self.multi_probe_pending = True
+        self.results = []
         return self
     def end_probe_session(self):
         if not self.multi_probe_pending:
             self._probe_state_error()
+        self.results = []
         self.multi_probe_pending = False
         self.mcu_probe.multi_probe_end()
     def get_probe_params(self, gcmd=None):
@@ -349,8 +354,13 @@ class ProbeSessionHelper:
                 toolhead.manual_move(
                     probexy + [pos[2] + params['sample_retract_dist']],
                     params['lift_speed'])
-        # Calculate and return result
-        return calc_probe_z_average(positions, params['samples_result'])
+        # Calculate result
+        epos = calc_probe_z_average(positions, params['samples_result'])
+        self.results.append(epos)
+    def pull_probed_results(self):
+        res = self.results
+        self.results = []
+        return res
 
 # Helper to read the xyz probe offsets from the config
 class ProbeOffsetsHelper:
@@ -441,19 +451,20 @@ class ProbePointsHelper:
         if self.horizontal_move_z < self.probe_offsets[2]:
             raise gcmd.error("horizontal_move_z can't be less than"
                              " probe's z_offset")
-        results = []
         probe_session = probe.start_probe_session(gcmd)
+        probe_num = 0
         while 1:
-            self._raise_tool(not results)
-            if len(results) >= len(self.probe_points):
+            self._raise_tool(not probe_num)
+            if probe_num >= len(self.probe_points):
+                results = probe_session.pull_probed_results()
                 done = self._invoke_callback(results)
                 if done:
                     break
-                # Caller wants a "retry" - clear results and restart probing
-                results = []
-            self._move_next(len(results))
-            pos = probe_session.run_probe(gcmd)
-            results.append(pos)
+                # Caller wants a "retry" - restart probing
+                probe_num = 0
+            self._move_next(probe_num)
+            probe_session.run_probe(gcmd)
+            probe_num += 1
         probe_session.end_probe_session()
     def _manual_probe_start(self):
         self._raise_tool(not self.manual_results)
@@ -476,7 +487,8 @@ class ProbePointsHelper:
 # Helper to obtain a single probe measurement
 def run_single_probe(probe, gcmd):
     probe_session = probe.start_probe_session(gcmd)
-    pos = probe_session.run_probe(gcmd)
+    probe_session.run_probe(gcmd)
+    pos = probe_session.pull_probed_results()[0]
     probe_session.end_probe_session()
     return pos
 
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -438,7 +438,7 @@ class ProbePointsHelper:
         def_move_z = self.default_horizontal_move_z
         self.horizontal_move_z = gcmd.get_float('HORIZONTAL_MOVE_Z',
                                                 def_move_z)
-        if probe is None or method != 'automatic':
+        if probe is None or method == 'manual':
             # Manual probe
             self.lift_speed = self.speed
             self.probe_offsets = (0., 0., 0.)
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -386,7 +386,7 @@ class ProbePointsHelper:
         # Internal probing state
         self.lift_speed = self.speed
         self.probe_offsets = (0., 0., 0.)
-        self.results = []
+        self.manual_results = []
     def minimum_points(self,n):
         if len(self.probe_points) < n:
             raise self.printer.config_error(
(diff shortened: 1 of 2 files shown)


Arksine:
Hi Kevin,

Having done a basic review of the code it looks good to me.  I'll do some testing and an in depth review as soon as I am able.  With regard to additional options and documentation, I can add that in a follow-up PR that adds "rapid" probing to `bed_mesh`.  Speaking of which, I think I can reuse most of your code with a small modification to the `EddyScanningProbe` class:

```diff
--- a/klippy/extras/probe_eddy_current.py
+++ b/klippy/extras/probe_eddy_current.py
@@ -346,12 +346,16 @@ class EddyScanningProbe:
         self._gather = EddyGatherSamples(printer, sensor_helper,
                                          calibration, z_offset)
         self._sample_time_delay = 0.50
-        self._sample_time = 0.100
+        self._sample_time = gcmd.get_float("SAMPLE_TIME", 0.100, above=0.0)
+        self._mode = gcmd.get("SCAN_MODE", "detailed")
     def run_probe(self, gcmd):
         toolhead = self._printer.lookup_object("toolhead")
         printtime = toolhead.get_last_move_time()
-        toolhead.dwell(self._sample_time_delay + self._sample_time)
-        start_time = printtime + self._sample_time_delay
+        if self._mode == "detailed":
+            toolhead.dwell(self._sample_time_delay + self._sample_time)
+            start_time = printtime + self._sample_time_delay
+        else:
+            start_time = printtime - self._sample_time / 2
         self._gather.note_probe(start_time, start_time + self._sample_time)
     def pull_probed_results(self):
         results = self._gather.pull_probed()
```

I believe with this change we can keep the scanning code in `probe_eddy_current.py` as `bed_mesh` can reuse the probe session interface when generating the rapid probe path.  If you think this is appropriate I can make this change in the follow-up PR, otherwise I can use the `add_client()` method and grab the samples directly as originally proposed.

KevinOConnor:
> I believe with this change we can keep the scanning code in probe_eddy_current.py as bed_mesh can reuse the probe session interface when generating the rapid probe path.

Yeah, if we can reuse the code, I think that is a good idea.  (As an aside, the code will also need to be updated to use `toolhead.register_lookahead_callback()`.)

What do you think about using `METHOD=rapid` (or similar) instead of adding a new user-facing `SCAN_MODE` setting?  It's not a big deal either way though.

Thanks.
-Kevin

KevinOConnor:
Yeah, that should work.

I just realized though, that my code has a systemic inaccuracy that should be fixed.  The "nominal position" should really be calculated from the stepper motor positions - not the toolhead/trapq positions.  For example, if one is using a Z with rotation_distance=8, full_steps_per_rotation=200, and microsteps=16 then there is a step_distance of 0.0025.  But, if one were to move to a Z position that isn't an exact multiple of that amount it could result in a systemic bias to the probe results.  This wasn't an issue for "descend until trigger" mode because the toolhead position is recalculated at the end of the trigger, but in scan (and rapid scan) modes, the actual stepper Z may deviate from the requested Z.  Probably not a huge precision issue on most printers, but could become an issue in some setups.

I'll update this branch to calculate the "nominal Z" from the stepper motor positions.  The one upside of this change is that we wont have to track the nominal toolhead position upfront - it can be calculated in EddyGatherSamples from the stepper movement history and start/end_times.

> I'm not sure if toolhead.wait_moves() is the best approach before pulling the results

Just "thinking out lout", simplest may be to call `toolhead.get_last_move_time()` in `pull_probed_results()` as that will flush the lookahead call and ensure all `register_lookahead_callback()` calls are completed.

Cheers,
-Kevin

KevinOConnor:
I reworked and rebased this branch.

Notes:
* This branch is still relative to the underlying probe refactoring (PR #6605).
* This branch now includes the sensor_bulk size change (for 2-byte response ids) as described at #6613.
* The scan mode code now calculates the low-level kinematic position from the steppers - which should provide a more precise position.
* The code (both scan and descend-until-trigger) will now average the frequencies and then calculate a Z height from that average (as opposed to averaging the calculated Z heights for each sample).  I think this should be slightly more accurate.
* I also tested a variant of @Arksine's "rapid_scan" snippet above.  I included it on this branch, but I'll drop it from this series if it is in the way.

-Kevin

c0psrul3:
testing this branch and found that i'm able to successfully do BED_MESH_CALIBRATE
* in detailed mode: successful
* in scan mode: error "probe_eddy_current sensor not in valid range"
* in rapid mode: error "probe_eddy_current sensor not in valid range"

any ideas here?

KevinOConnor:
Thanks for testing.

If you're using one of the scan modes, you'll need to make sure the horizontal_move_z field is relatively low - try a command like: `bed_mesh_calibrate method=scan horizontal_move_z=1.5`  (And, you'll need to make sure the bed is relatively flat so that the nozzle does not strike the bed while moving at that low Z height.)

If that fails, run `M112` immediately after the unexpected result, and then attach the full Klipper log file here.

-Kevin

c0psrul3:
testing this branch and found that i'm able to successfully do BED_MESH_CALIBRATE
* in detailed mode: successful
* in scan mode: error "probe_eddy_current sensor not in valid range"
* in rapid mode: error "probe_eddy_current sensor not in valid range"

any ideas here?

c0psrul3:
@KevinOConnor thanks, that seems to have solved my issue.
for usability, do you think there's anything that can be done to meet n00b user expectations?  default values?
allowing `horizontal_move_z` to be set in `probe_eddy_current` config section?  just ideas.

thank you again for the guidance.

c0psrul3:
testing the different methods (scan works great) and got this error:
`Probe triggered prior to movement`
from command:  `BED_MESH_CALIBRATE METHOD=rapid PROFILE=rapid-test-1 horizontal_move_z=1.5`

Arksine:
I have had a chance to test the latest version if this branch and refactor `bed_mesh` to accommodate these changes.  Overall it looks good to me, however I do have some observations:

1) For the `rapid_scan` method, I noticed that for samples where there is a direction change, the kinematic XY positions returned by `EddyGatherSamples` may slightly deviate from the requested position.  This can cause an error in the `probe_finalize()` in when `bed_mesh` attempts to validate faulty regions and/or when it transposes the results into a Z matrix.  That said, I can increase the tolerance in bed mesh to accommodate the deviation, as long a we know it won't be too large.   In my local work I have changed bed mesh to allow up to .5mm deviation on X and/or Y.

2) For the `scan` method, registering lookahead callbacks would likely speed up the probing procedure a bit, although it works fine as things stand.  This change would likely also introduce the deviations noted above.


>for usability, do you think there's anything that can be done to meet n00b user expectations? default values?
allowing horizontal_move_z to be set in probe_eddy_current config section? just ideas.

@c0psrul3 The `[bed_mesh]` section has a `horizontal_move_z` option that can be set to apply a default.

> testing the different methods (scan works great) and got this error:
Probe triggered prior to movement
from command: BED_MESH_CALIBRATE METHOD=rapid PROFILE=rapid-test-1 horizontal_move_z=1.5

The correct method name is `rapid_scan`.  The error you received was the result of the scanning procedure falling back to the `automatic` method, ie: standard probing.  FWIW, when you test `rapid_scan` you may encounter an error due to the issue I noted in `#1` above.

c0psrul3:
@Arksine thank you for correcting me, as i obv did not read the diff close enough.
i was referring to the fact that horizontal_move_z default is `5` and does not indicate much on how to resolve it.

@KevinOConnor rapid_scan immediately gives me error `Probe triggered prior to movement`, but does appear to *work* with horizontal_move_z=2 however the scan process then does hop between scans.
so, with `BED_MESH_CALIBRATE METHOD=rapid_scan PROFILE=rapid-test-1 horizontal_move_z=2`  the mesh appears to look nearly identical to the mesh from `method=scan`

[klippy.log.20240609-1109.txt](https://github.com/user-attachments/files/15753061/klippy.log.20240609-1109.txt)

Arksine:
> @KevinOConnor rapid_scan immediately gives me error Probe triggered prior to movement, but does appear to work with horizontal_move_z=2 however the scan process then does hop between scans.
so, with BED_MESH_CALIBRATE METHOD=rapid_scan PROFILE=rapid-test-1 horizontal_move_z=2 the mesh appears to look nearly identical to the mesh from method=scan

According to your log you are on commit aa507c897.  The lastest commit in this branch after Kevin's last push is 2ac8242c18520cce38aa7813a8ee21306a09a6ce.  You likely need to fetch and reset (or checkout detached) to update to the latest code.

The `rapid_scan` should be one continuous movement.

c0psrul3:
@Arksine thanks for pointing this out :+1:   rapid_scan appears to work, went through the motions, but got error message:
`bed_mesh: Invalid y-axis table length`
however, `METHOD=scan` appears to still work without issue.

[klippy.log.20240609-1651.txt](https://github.com/user-attachments/files/15754212/klippy.log.20240609-1651.txt)

KevinOConnor:
Thanks for testing and reviewing.

> I noticed that for samples where there is a direction change, the kinematic XY positions returned by EddyGatherSamples may slightly deviate from the "requested position".

Yeah - the latest code is calculating the x, y, and z position from the stepper motors.  There are a few reasons the resulting "stepper position" may deviate from the requested position:
1. Due to minor rounding differences in floating point numbers.
2. Due to microstep coarseness.  (For example, a requested position X=100.007 may not correspond to a whole microstep on the stepper.)
3. Due to input shaper.  We'll actually be reporting the position of the steppers, and input shaper alters those positions relative to the requested positions - in particular during cornering.

It is possible to return requested XY positions, but it may also be worthwhile to return the actual XY positions.  Not sure.  Note that probe XY positions could always deviate slightly from requested positions (due to stepper coarseness - in particular on delta type kinematics).

> For the scan method, registering lookahead callbacks would likely speed up the probing procedure a bit

Unless I'm missing something, there shouldn't be a difference.  Calling `dwell()` should immediately flush all the lookahead callbacks.  So, as far as I'm aware, calling `register_lookahead_callback()` followed by `dwell()` should produce the same result as `print_time = toolhead.get_last_move_time()` followed by `dwell()`.

Cheers,
-Kevin

KevinOConnor:
> but after a closer look its likely due to self._sample_time_delay being set to .5 rather than .05 which is what was previously used.

Eeks!  I must have goofed on one of my rebase jobs.  It was intended to be `0.050`.

I restored the value to 0.050.  I also merged in the backend probe work (PR #6605), merged the 2-byte command id work (PR #6613), and rebased this branch on top of the latest master.

-Kevin

c0psrul3:
@KevinOConnor thanks for putting some updates.
i'm still getting the error `bed_mesh: Invalid y-axis table length`
with the following additional output to console:
```
bed_mesh: Invalid y-axis table length
Probed table length: 30 Probed Table:
...
```

what i do notice is the output `Probed table length: 30`   while i have configured in bed_mesh: `probe_count: 15,15`

thanks!

Sineos:
@c0psrul3, thanks for testing. 
Without this case in particular, it is generally recommended to call a `M112` immediately after an issue and just post the log. This will give the devs the best possible information to continue working with.

KevinOConnor:
> i'm still getting the error bed_mesh: Invalid y-axis table length

As indicated at https://github.com/Klipper3d/klipper/pull/6610#issuecomment-2156634469 , the current bed_mesh code would need changes in order to be used with "rapid_scan".

I'll drop the "rapid_scan" part of this PR prior to committing.

-Kevin

Arksine:
I have created PR #6617, adding support for rapid scanning in `bed_mesh`.  It should resolve the errors discussed above.

KevinOConnor:
Thanks for testing.

> the map for METHOD=scan looks smoothed when compared to the map from METHOD=rapid_scan. this is also revealed by comparing each profile's deviation (0.137 for "scan" :: 0.170 for "rapid_scan").

Well, I guess that's what I would have expected - the "scan" mode would produce a more accurate map, and the "rapid_scan" would produce a map with more "noise".  That is, rapid_scan runs a little faster but isn't as accurate.

Or are you saying you've done some kind of analysis showing that your bed is actually very "crinkly" like the rapid scan shows?

Cheers,
-Kevin

c0psrul3:
lol, my n00b is showing.
I've not done any analysis and noise didn't occur to me.  it is textured,
but that wouldn't show at 30,30.  noise would make better sense than
accuracy.

in any case, it all looks good and ready imo

KevinOConnor:
I merged the "scan" mode support in this PR into the main repo.  I did not merge the initial "rapid_scan" work previously discussed here - it looks like that can be done in the upcoming PR #6617.

-Kevin

PR #6613, opened 2024-06-06
Support 2-byte message ids in host/mcu protocol

KevinOConnor:
Currently the message protocol only supports 128 possible message ids.  However, some microcontroller builds need up to ~125 messages today and some proposed code additions are exceeding the limit.

This PR increases the limit from 128 message ids to 16384 ids.  It does this by adding support for either 1 or 2 byte message ids in the low-level message parsing code.

This does mean that the bulk sensor code should use no more than 51 bytes of data when reporting results back to the host.  (A message can be up to 64 bytes long: 5 bytes for the headers, 2 bytes for the response id, 2 bytes for oid, 3 bytes for the sequence number, 1 byte for data length, 51 bytes of data.)  This is a breaking change for the ldc1612 sensor support - users with this sensor will need to recompile and reflash their mcu code after applying this update.

@garethky - fyi.

-Kevin

### DIFF
--- a/klippy/extras/bulk_sensor.py
+++ b/klippy/extras/bulk_sensor.py
@@ -198,7 +198,7 @@ class ClockSyncRegression:
         inv_freq = clock_to_print_time(base_mcu + inv_cfreq) - base_time
         return base_time, base_chip, inv_freq
 
-MAX_BULK_MSG_SIZE = 52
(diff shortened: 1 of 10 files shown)


garethky:
Thanks for picking this up so quickly! I didn't want to be responsible for knowing all the ways this could break something.

PR #6617, opened 2024-06-11
bed_mesh: add support for "rapid" scanning

Arksine:
This pull request builds on #6610, adding support for optimized travel paths when a `rapid_scan` is requested.   In addition, the `probe_finalize()` method no longer depends on on the XY coordinates received from the probing procedure to populate the Z matrix.

The final commit in this series updates `gnupru` to the latest release in `ci-install.sh`.  It appears that version `2023.1` fails due to a missing dependency.

### DIFF
--- a/docs/Bed_Mesh.md
+++ b/docs/Bed_Mesh.md
@@ -421,12 +421,75 @@ have undesirable results when attempting print moves **outside** of the probed a
 full bed mesh has a variance greater than 1 layer height, caution must be taken when using
 adaptive bed meshes and attempting print moves outside of the meshed area.
 
+## Surface Scans
+
+Some probes, such as the [Eddy Current Probe](./Eddy_Probe.md), are capable of
+"scanning" the surface of the bed.  That is, these probes can sample a mesh
+without lifting the tool between samples.  To activate scanning mode, the
+`METHOD=scan` or `METHOD=rapid_scan` probe parameter should be passed in the
+`BED_MESH_CALIBRATE` gcode command.
+
+### Scan Height
+
+The scan height is set by the `horizontal_move_z` option in `[bed_mesh]`.  In
+addition it can be supplied with the `BED_MESH_CALIBRATE` gcode command via the
+`HORIZONTAL_MOVE_Z` parameter.
+
+The scan height must be sufficiently low to avoid scanning errors.  Typically
+a height of 2mm (ie: `HORIZONTAL_MOVE_Z=2`) should work well, presuming that the
+probe is mounted correctly.
+
+It should be noted that if the probe is more than 4mm above the surface then the
+results will be invalid.  Thus, scanning is not possible on beds with severe
+surface deviation or beds with extreme tilt that hasn't been corrected.
+
+### Rapid (Continuous) Scanning
+
+When performing a `rapid_scan` one should keep in mind that the results will
+have some amount of error.  This error should be low enough to be useful on
+large print areas with reasonably thick layer heights.  Some probes may be
+more prone to error than others.
+
+It is not recommended that rapid mode be used to scan a "dense" mesh.  Some of
+the error introduced during a rapid scan may be gaussian noise from the sensor,
+and a dense mesh will reflect this noise (ie: there will be peaks and valleys).
+
+Bed Mesh will attempt to optimize the travel path to provide the best possible
+result based on the configuration.  This includes avoiding faulty regions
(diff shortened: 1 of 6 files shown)


KevinOConnor:
Thanks.  I cherry-picked your gcc-pru build patch.

This PR seems good to me.  I think we should give a few days for others to comment, but otherwise I'm ready to commit when you are.

Not a "blocker", but have you considered using a webhooks endpoint instead of the BED_MESH_DUMP command (and have `graph_mesh.py` directly pull the json from the unix domain socket)?  The gain would be that it would also enable the web frontends to access all this information in the future.  It would also avoid having to describe the low-level details of the command, the file name, file location, file contents to users.  Again, not a "blocker".

Thanks again,
-Kevin

docs/Bed_Mesh.md
```diff
@@ -421,12 +421,75 @@ have undesirable results when attempting print moves **outside** of the probed a
...
+have some amount of error.  This error should be low enough to be useful on
+large print areas with reasonably thick layer heights.  Some probes may be
+more prone to error than others.
+
+It is not recommended that rapid mode be used to scan a "dense" mesh.  Some of
+the error introduced during a rapid scan may be gaussian noise from the sensor,
+and a dense mesh will reflect this noise (ie: there will be peaks and valleys).
+
+Bed Mesh will attempt to optimize the travel path to provide the best possible
+result based on the the configuration.  This includes avoiding faulty regions
```
KevinOConnor, line 458: `the the` typo

Arksine: Whoops, this should be fixed.

Arksine:
> Not a "blocker", but have you considered using a webhooks endpoint instead of the BED_MESH_DUMP command (and have graph_mesh.py directly pull the json from the unix domain socket)? The gain would be that it would also enable the web frontends to access all this information in the future. It would also avoid having to describe the low-level details of the command, the file name, file location, file contents to users. Again, not a "blocker".

I think this is a good idea.  I'll rework `BED_MESH_DUMP` into a `bed_mesh/dump_mesh` API request.

c0psrul3:
:+1:  this works great

Arksine:
I have removed the `BED_MESH_DUMP` gcode command in favor of a `bed_mesh/dump_mesh` API.  The `graph_mesh` script can fetch the data in one of three ways:

- A direct connection to Klipper's unit socket
- The bridge connection via Moonraker
- A json file

The json file can be created with the `graph_mesh dump <input>` command.  This allows for users to share the dump when troubleshooting if necessary.  The Moonraker connection allows users to run the script on a Desktop class PC enabling animations and live mesh previews.

KevinOConnor:
Thanks.  It looks good to me.  If there are no further comments, I'll look to commit in a couple of days.

-Kevin

klippy/extras/bed_mesh.py
```diff
@@ -282,6 +287,32 @@ def cmd_BED_MESH_OFFSET(self, gcmd):
             gcode_move.reset_last_position()
         else:
             gcmd.respond_info("No mesh loaded to offset")
+    cmd_BED_MESH_DUMP_help = "Dump mesh data to file for analysis"
```
KevinOConnor, line 290: Minor - dead line of code?

-Kevin

Arksine: Oops.  Should be fixed now.

KevinOConnor:
Thanks!

-Kevin

Cereal-Killa:
Question, are there more patches to get the BTT Eddy to work on the official Klipper branch or is this it?

KevinOConnor:
There are a number of features that have been discussed or are in progress.  Here's my understanding of the status:
* Basic probing support was added to mainline in PR #6536.
* "Detailed scan" support was added in PR #6610.
* "Rapid scan" support was added in this PR (#6617).

I expect the above is what most users will expect.

There are a few things in discussion that are not in mainline:
* Automated thermal adjustment (PR #6602)
* Support for homing with the ldc1612 (it is possible with the mainline code, but requires several steps).
* Probing in "tap" mode - still in discussion - for example: https://klipper.discourse.group/t/tap-detection-with-ldc1612-inductive-sensors/16234

Hope that helps,
-Kevin

Cereal-Killa:
> There are a number of features that have been discussed or are in progress. Here's my understanding of the status:
> 
> 
> Hope that helps, -Kevin

Absolutely, in this case I'll stick to the BTT branch for the time being, specially due to the homing feature.
Thanks Kevin

PR #6622, opened 2024-06-18
servo42d: Initial implementation

Maccraft123:
This PR adds support for Makerbase's Servo42D-CAN closed loop stepper drivers.
I wasn't really sure how to handle the CAN connectivity for something that isn't an MCU so ATM this is more of an RFC than a "real" PR.

The configuration options have been tested with a Fysetc UCAN board used for CAN communication.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3267,6 +3267,44 @@ pins:
 #   parameter must be provided.
 ```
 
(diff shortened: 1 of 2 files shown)


Maccraft123:
Closing after getting too frustrated with getting those drivers to be accurate under FOC mode and feeling like I'm polishing a turd.

PR #6628, opened 2024-06-25
Update Dockerfile to use the official python2 image

NeoTheFox:
Switched the provided Dockerfile from unsupported for over a year ubuntu 18.04 image to official python2 image, based on Debian. This ensures that the python2 package is provided and the Debian base is supported.

### DIFF
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -14,13 +14,13 @@
 #       /home/klippy/.config/
 # For more Dockerfile examples with Klipper (and Octoprint) see:
 # https://github.com/sillyfrog/OctoPrint-Klipper-mjpg-Dockerfile
-FROM ubuntu:18.04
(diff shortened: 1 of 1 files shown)


thijstriemstra (approved):
Looks good to me.

scripts/Dockerfile
```diff
@@ -14,13 +14,13 @@
 #       /home/klippy/.config/
 # For more Dockerfile examples with Klipper (and Octoprint) see:
 # https://github.com/sillyfrog/OctoPrint-Klipper-mjpg-Dockerfile
-FROM ubuntu:18.04
+FROM python:2
```
bevanweiss, line 17: Would 'python:3' now work here?

NeoTheFox: It would, is python3 the preferred version now?

PR #6632, opened 2024-07-01
spi_flash: New flash-sdcard board definitions for Creality v4.5.x motherboards

Palakis:
This pull request adds two board definitions for Creality v4.5.2 and v4.5.3 motherboards found in Creality CR-6 SE printers. These definitions are actually aliases pointing to the existing `creality-v4.2.2` definition, which has the same MCU and SDIO pinout as the v4.5.x boards.


JamesH1978:

Thanks
James

Palakis:
Oops, my bad. Fixed!

Palakis:
There is an issue with firmware filename case-sensitivity with these boards. The stock bootloader on Creality v4.5.x motherboards looks for a lowercase `firmware.bin` while FatFS writes it as uppercase `FIRMWARE.BIN`, which means the stock bootloader won't see the new firmware file upon reboot and won't update.

It seems like FatFS only does uppercase DOS-style filenames, and it would require some level of refactor to support these specific boards. At this point, users of these boards would be better off with Katapult or the standard firmware update procedure.

PR #6633, opened 2024-07-03
sw spi bus: Added support for configured width and encoding

bazarovdev:
Adds option to work with any device/sensor unrelated to HW SPI bus width.
Implemented to work with `YHCB2004` display on `Geeetech A10/M/T` printers that require 9-bit SPI commands.

### transmitting 20 bits `0x80f05`
ff0
### testing all 4 options of encoding of `0x80f05`
**big/little-endian**
**lsb/msb bit first**
4 opt4

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4826,6 +4826,13 @@ SPI bus.
 #spi_software_sclk_pin:
 #spi_software_mosi_pin:
 #spi_software_miso_pin:
+#spi_software_bus_width: number of bits from `5` to `64` to send/receive per
+#                        transaction on the software SPI bus (default 8 bits).
+#spi_software_is_little_endian: `True`/`False` - if `True` and width above `8`
+#    bits, the least significant byte will be sent first, otherwise last
+#    (default `False`).
+#spi_software_is_lsb_first: `True`/`False` - if `True` then the least
+#    significant bit will be sent first (default `False`).
 #   Specify the above parameters to use "software based SPI". This
 #   mode does not require micro-controller hardware support (typically
 #   any general purpose pins may be used). The default is to not use
--- a/klippy/extras/bus.py
+++ b/klippy/extras/bus.py
@@ -40,11 +40,22 @@ def resolve_bus_name(mcu, param, bus):
 # Helper code for working with devices connected to an MCU via an SPI bus
 class MCU_SPI:
     def __init__(self, mcu, bus, pin, mode, speed, sw_pins=None,
-                 cs_active_high=False):
+                 cs_active_high=False, sw_bus_width=8, is_little_endian=False,
+                 is_lsb_first=False):
         self.mcu = mcu
         self.bus = bus
         # Config SPI object (set all CS pins high before spi_set_bus commands)
         self.oid = mcu.create_oid()
+        self.width = sw_bus_width
+        self.is_little_endian = is_little_endian
+        self.is_lsb_first = is_lsb_first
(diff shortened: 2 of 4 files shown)


nwash57:
The code is black magic to me, but I can confirm this makes the display work on the A10T I converted to Klipper. At the very least it makes getting these junky and bug-ridden printers into a stable configuration that much easier.

bazarovdev:
Self review:
1. Defects: the changes were minimized and instead of introducing new commands, added modification to `spi_software_transfer()` that depending on initialization treats the data instead of array of bytes as concatenated bits of specified width.
It supports encodings `big/small endian` and transmitting`lsb/msb` bit first. So in all calls in Python the data is encoded/decoded and in C the bits are just transmitted without adding too much complexity.
Attached the scope pictures of main tests.
2. Impact: it makes software SPI flexible to meet any need of users. In addition, it gives option to integrate support to Geeetech A10/M/T, ender clones with mixing extruders (M=2in1out, T=3in1out) increasing user base of Klipper and potentially helping test features related to mixing.
(Currently, most users stuck with Marlin or mainly replace stock board+display to work with Klipper. The facebook group of owners contains 6.3k users).
3. Licensing: No new files introduced, commits signed-off.
4. Guidelines: Tried to follow to my best understanding.
5. Docs: PR contains update to docs related to new fields in `spi_software_*`section
6. Commits: 1 commit with 1 feature.

KevinOConnor:
Thanks.  As high-level feedback, I am leery of making such an intricate change to a common component in support of very rare hardware.  The concern is that a change to this common code could introduce a regression (either directly or via a performance degradation).

Which devices did you have in mind for this support?  Is there any other way to drive these devices without requiring a full 9-bit (or other) byte order?  For example, arranging to always send commands in multiples of nine 8-bit bytes.

I have not done a full code review, but I did notice that the C code on this PR introduces an arbitrary integer division (along with a modulo).  Most micro-controllers do not directly support division (nor modulo) and using these operators can result in slow and bloated code being introduced by the compiler.

Cheers,
-Kevin

bazarovdev:
Thanks for the feedback.
1. I got stuck on a problem with `YHCB2004` display and, while solving it, dug into the software SPI code. The idea was that the change to make it completely generic wasn't that hard after I got what is going on. It solves my problem and might help others down the road (it took me a lot of time to understand and solve the problem, while initially it seemed just as minor change in the display driver).
2. sending 8 commands that add up to 9 bytes might work, thanks for the idea, I will test it.
3. I used `division`/`mod` for input checking and calculating number of chunks to avoid adding additional `chunks_num` parameter. But it seems, I can drop divisions and instead just track number of bits left as it sends them. And it seems, there is a bug of ambiguity for chunks less that 8 bits, so I will set lower limit to 8 bits.

Hopefully, the 8 commands solution will work, so I will be able to update https://github.com/Klipper3d/klipper/pull/6639 without this PR (and I hope adding display driver doesn't fall in "no go zone":).
And will try to address division issues here, to make this PR more likely to be merged someday.

bazarovdev:
The proposed sending 8 commands as 9 bytes actually works.
Updated https://github.com/Klipper3d/klipper/pull/6639

I'm not fixing this, as it is not intendent to be merged now.
For future, if somebody decides to use it, TODO: 
* limit minimal width to 8 bits (otherwise it might make mistake on number of commands encoded in byte list)
* replace `division`/`mod` by counting actual number of commands econded as it goes

nefelim4ag:
As the display for which it was initially written is merged, there is no plan to work on this specific topic.

This seems obsolete and stale.

-Timofey

PR #6634, opened 2024-07-03
sht3x: use periodic report mode

nefelim4ag:
SHT3X supports periodic mode, which allows us to completely skip the "write" phase and wait, and just read data in one command.

Also, I corrected reactor pauses to the right numeric order (they are also wrong in htu21.py).

Thanks.

### DIFF
--- a/klippy/extras/bus.py
+++ b/klippy/extras/bus.py
@@ -192,6 +192,9 @@ class MCU_I2C:
             return
         self.i2c_write_cmd.send([self.oid, data],
                                 minclock=minclock, reqclock=reqclock)
+    def i2c_write_wait_ack(self, data, minclock=0, reqclock=0):
+        self.i2c_write_cmd.send_wait_ack([self.oid, data],
+                                minclock=minclock, reqclock=reqclock)
     def i2c_read(self, write, read_len):
         return self.i2c_read_cmd.send([self.oid, write, read_len])
     def i2c_modify_bits(self, reg, clear_bits, set_bits,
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.  Happy to commit this when you are ready.

Note, though, that host based `pause()` commands do not enforce a chip delay.  The `i2c_write()` commands are queued and sent to the mcu for execution - pausing between queuing of commands does not ensure the mcu will have a pause between execution of commands.

There are a variety of ways to fix this.  I'd guess the simplest would be to add an `i2c_write_wait_ack()` to bus.py  (that uses `self.i2c_write_cmd.send_wait_ack()`).  I'm not entirely sure what the sht3x chip requirements are though.

-Kevin

nefelim4ag:
I'm not an experienced embedded developer, so I write pauses according to the datasheet with some additional sane margin "just in case".

In HTU code they all look like x10 of max datasheet values, _maybe because reactor pause does not guarantee timings?_
So, Initially, without complete understanding, I used the same order of values.
(like 20ms = 0.20, instead of 0.020)

With that said:
* If you would like, I can add and test wait_ack for i2c here, and after for some bme sensors, because I also have them (in separate PR if their timing differs as here). It looks like wait_ack is not blocking, and should not cause issues.
* Or, I can return pauses to x10 of max values and leave everything as is.

Thanks

KevinOConnor:

The `pause()` timing is accurate - I'd guess that other code exaggerates delays to try to account for host delays not easily correlating to mcu delays.  If you can make and test fixes for the bme code then that would be great.

-Kevin

PR #6642, opened 2024-07-11
Update tmc5160.py

HonestBrothers:
TMC5160.py incorrectly implements CS values to where they are always 31. 

The calculation of globalscaler in current Klipper code leaves out the CS values, and then calculates CS after globalscaler.

According to page 74 of the datasheet, Calculation of RMS Current, the end result of not specifying a CS value, automatically sets the CS to 31. 

(CS+1)/32 = 1. 

In the current code, globalscaler is calculated and then CS is back calculated using globalscaler, but the only value it can be set to is 31, which the globalscaler calculation has already determined by setting it to 1. 

Analog Devices is less than forthcoming on how to calculate the CS value, as the equation used on page 74 has two unknowns and thus can not be calculated. However, the tuning spreadsheet TMC220x_TMC222x_Calculations.xlsx allows us to back calculate the CS value:

cs = int(100 * (current * math.sqrt(2)) * self.sense_resistor) - int(1 - 2 * (current * math.sqrt(2)))

Implementing this bug fix allows us to effectively set the CS value before globalscaler, and then set globalscaler based on the chosen CS value. Which is how I believe Analog Devices intended this driver to be tuned. This has the end result of CS being used to effectively scale Rsense, ie. when the driver is rated for more currrent than is being used, Rsense does not need to be changed and instead CS can scale the Rsense input. 

The practical issue is that the way Klipper currrently calculates CS makes the motors hard to tune on high amp 5160 drivers, and high voltages practically impossible on some motors because the chopper parameters can not be correctly set. Current word around town is that 2804's don't do well above 24V and will result in VFAs if used at 48V.

I have implemented this fix on my own device and am using 2804's at 60V, which run at a cool 65C. Before implementing the motors were very noisy and melted my motor mounts. Now they are regular level of 1980's printer noisy. 

The CS value is calculated automatically so it requires no user input. This fix should allow better motor tuning, the calc sheet to be used effectively, and certain motor, voltage, driver combos to be used where they could not before. 

Signed off by: Brandon Smith honestbrotherstv@gmail.com

### DIFF
--- a/klippy/extras/tmc5160.py
+++ b/klippy/extras/tmc5160.py
@@ -254,7 +254,7 @@ VREF = 0.325
 MAX_CURRENT = 10.000  # Maximum dependent on board, but 10 is safe sanity check
 
 
-class TMC5160CurrentHelper(tmc.BaseTMCCurrentHelper):
+class TMC5160CurrentHelper:
     def __init__(self, config, mcu_tmc):
         super().__init__(config, mcu_tmc, MAX_CURRENT)
 
--- a/klippy/extras/tmc5160.py
+++ b/klippy/extras/tmc5160.py
@@ -256,20 +256,21 @@ MAX_CURRENT = 10.000  # Maximum dependent on board, but 10 is safe sanity check
 
 class TMC5160CurrentHelper:
     def __init__(self, config, mcu_tmc):
-        super().__init__(config, mcu_tmc, MAX_CURRENT)
-
-        self.sense_resistor = config.getfloat(
-            "sense_resistor", 0.075, above=0.0
-        )
-        self.cs = 31
-        gscaler, irun, ihold = self._calc_current(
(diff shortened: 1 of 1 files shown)


Sineos:
Thanks for your contribution. While I cannot judge the correctness of it, please mind  https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md. In particular:
- Signed-off line
- Avoid unrelated white space / formatting changes

HonestBrothers:
> Thanks for your contribution. While I cannot judge the correctness of it, please mind https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md. In particular:
> 
> * Signed-off line
> * Avoid unrelated white space / formatting changes

I will work on the formatting. Not sure why it's such a mess.

PR #6645, opened 2024-07-13
BME280/BME680 reduce bandwidth usage

nefelim4ag:
BME280/BMP280 has a normal (periodic) mode where we only read data - let's use it and simplify data read and timings.
Technically we know that the maximum sampling rate will be above or equal to 0.8s and that the maximum possible sample time is `(1.25 + (2.3 * 16) + ((2.3 * 16) + .575) + ((2.3 * 16) + .575)) / 1000 = 0.1128 ms`.
That allows for specific standby times, such as 500 ms or 250 ms.

BME680 officially does not have the same mode.
But all registers that have been written are stored between cycles.

_BTW: BME680 is heating (+2.5C in my case) because of frequent gas heating cycles, and according to the datasheet, it is supposed to be tracked once at 3s or less often._

Also, I touched everything else as little as possible because refactoring should be done in separate PR and I did not have the sensors to test changes.

Thanks

### DIFF
--- a/klippy/extras/bme280.py
+++ b/klippy/extras/bme280.py
@@ -157,6 +157,7 @@ class BME280:
             return
         self.printer.register_event_handler("klippy:connect",
                                             self.handle_connect)
+        self.last_gas_time = 0
 
     def handle_connect(self):
         self._init_bmxx80()
@@ -297,8 +298,7 @@ class BME280:
         if self.chip_type == 'BME680':
             self.max_sample_time = \
                     (1.25 + (2.3 * self.os_temp) + ((2.3 * self.os_pres) + .575)
-                     + ((2.3 * self.os_hum) + .575)
-                     + self.gas_heat_duration) / 1000
+                     + ((2.3 * self.os_hum) + .575)) / 1000
             self.sample_timer = self.reactor.register_timer(self._sample_bme680)
             self.chip_registers = BME680_REGS
         elif self.chip_type == 'BMP180':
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Maybe @Arksine can take a look at this PR.

-Kevin

klippy/extras/bme280.py
```diff
@@ -319,15 +322,17 @@ def read_calibration_data_bmp180(calib_data_1):
 
             self.sample_timer = self.reactor.register_timer(self._sample_bmp388)
         else:
-            self.max_sample_time = \
-                (1.25 + (2.3 * self.os_temp) + ((2.3 * self.os_pres) + .575)
-                 + ((2.3 * self.os_hum) + .575)) / 1000
+            if self.chip_type == 'BME280':
```
Arksine, line 325: Minor nitpick, but I believe this conditional block can be an extension of its parent block.

Arksine:
Aside from the minor refactoring above the changes look good to me.  I'll try to test the bme280 implementation on one of my machines today for additional validation.  I don't have a bmp280 or bme680, but I presume they were tested as well.

nefelim4ag:
I also don't have BMP280, so I can only say that according to datasheet and previous code BME/BMP differs only by Humidity sensor.

We can of course split it to separate code, and don't touch it.

nefelim4ag:
I did some further digging into BME680 self-heating thermal drift, I have the following data points ('cold' chamber ~27.5C):

```
gas freq    Temp Hum   Pressure     VOC
disabled   27.9C 54.8% 1009.8 hPa 0 Om
0.8 s(cur)  29.1C 50.9% 1009.8 hPa 479.4 kOm
>2s            28.3C 53.2% 1009.9 hPa 265.1 kOm
>3s            28.2C 53.4% 1009.9 hPa 256.5 kOm
>5s            28.1C 53.8% 1009.9 hPa 228.6 kOm
>7s            28.0C 53.8% 1009.8 hPa 218.9 kOm (decreased to 147 kOm after > 15 min)
```

by datasheet VOC levels explained like on the picture, so there is a massive difference not only by absolute number, but also by sense.
image

So, I would suggest code to query the sensor once in a while

KevinOConnor:
What was the end result of this PR?  Should I commit it?

-Kevin

Arksine:
Hi Kevin, the changes look good to me and run well on my BME280.  You can proceed to merge when you are ready.

KevinOConnor:
Thanks.

-Kevin

PR #6646, opened 2024-07-14
Add stm32g0b0xx to board_defs.py

bryan065:
Added board definition for stm32g0b0xx variant of the SKR Mini v3.0.

I couldn't flash this variant of the board with the flash_sdcard.sh script and had to add an additional definition.

### DIFF
--- a/scripts/spi_flash/board_defs.py
+++ b/scripts/spi_flash/board_defs.py
@@ -31,6 +31,11 @@ BOARD_DEFS = {
         'spi_bus': "spi1",
         "cs_pin": "PA4"
     },
+    'btt-skr-mini-v3-b0': {
+        'mcu': "stm32g0b0xx",
+        'spi_bus': "spi1",
+        "cs_pin": "PA4"
+    },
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thankyou for submitting a PR

There is no need to define a new definition, you only need to create a new alias `'btt-skr-mini-e3-v3': BOARD_DEFS['btt-skr-mini-v3'],`

You can see this method being used for other boards in the same family in the file.

Thanks
James

bryan065:
That was the first thing I tried as it's the same method used for other boards of the family. If I just add the alias, this is the error I'm getting below as I have the stm32g0b0 variant of the SKR mini 3.0. The board definition for `'btt-skr-mini-v3'` targets the stm32g0b1.

image

JamesH1978:
Yes, you are completely right. same connections, different chip, so an alias cant be used.

JamesH1978 (approved):
Looks fine to me

Thanks
James

KevinOConnor:
Thanks.

-Kevin

PR #6647, opened 2024-07-14
install-debian.sh: Highlight "NOT" in root check

Kreyren:
Got confused by that, the console is sometimes hard to read over serial.


JamesH1978:

Thanks
James

Kreyren:
As stated in https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2248660638 this is considered obsolete.

PR #6648, opened 2024-07-14
install-debian-12.sh: Initial commit

Kreyren:
Made the mistake of looking at the source code of random script that deploys services to Debian on a single board computer to be again annoyed about the quality and disfunctionality to refactor it. Why is it so difficult for people in the open-source community to learn how to write shell scripts. (=-=)

Please someone punch me each time i do this so that i don't end up making these merge requests and hopefully resist refactoring the rest of them.

Peer-Review requested in libera.chat's #bash channel.

Squash commits into one to maintain merge request history with the commit message from the first commit and add `Reviewed-by: @sorpigal`

_<sup>[If you find value in my work, then you might just find a way to brighten up monerochan's day](https://github.com/Kreyren#donate)</sup>_

### DIFF
--- a/scripts/install-debian-12.sh
+++ b/scripts/install-debian-12.sh
@@ -3,15 +3,12 @@
 
 # This script installs Klipper3D on an Debian 12
 
-set -e # Exit on false return
-
-
 srcDir="$(git rev-parse --show-toplevel)"
 
 # Exit trap
 die() { printf "FATAL: %s\n" "$2"; exit "$1" ;}
 
-# Check for bare repo
+# Check for bare repo, note that the used command outputs boolean in a form of a string
 [ "$(git rev-parse --is-bare-repository)" != "true" ] || die 1 "This script is NOT designed to run outside of the Klipper3D repository" 
 
 pythonDir="$HOME/klippy-env"
@@ -54,10 +51,10 @@ create_virtualenv() {
     report_status "Updating python virtual environment..."
 
     # Create virtualenv if it doesn't already exist
-    [ ! -d "$pythonDir" ] && virtualenv -p python3 "$pythonDir"
+    [ -d "$pythonDir" ] || { virtualenv -p python3 "$pythonDir" || die 1 "Failed to set up python environment" ;}
 
     # Install/update dependencies
-    "$pythonDir/bin/pip" install -r "$srcDir/scripts/klippy-requirements.txt"
+    "$pythonDir/bin/pip" install -r "$srcDir/scripts/klippy-requirements.txt" || die 1 "Failed to install/Update dependencies"
 }
 
 # Step 3: Install startup script
--- a/scripts/install-debian-12.sh
+++ b/scripts/install-debian-12.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env sh
+# shellcheck shell=sh # POSIX
+
+# This script installs Klipper3D on an Debian 12
+
+set -e # Exit on false return
+
+srcDir="$(git rev-parse --show-toplevel)"
+
+# Exit trap
+die() { printf "FATAL: %s\n" "$2"; exit 1 ;}
+
+# Check for bare repo
+[ "$(git rev-parse --is-bare-repository)" != "true" ] || die 1 "This script is NOT designed to run outside of the Klipper3D repository" 
+
+PYTHONDIR="$HOME/klippy-env"
+SYSTEMDDIR="/etc/systemd/system"
+klipperUser="$USER"
+# shellcheck disable=SC2034 # Keep this available in case it's ever needed
+klipperGroup="$klipperUser"
+
+# Step 1: Install system packages
+install_packages() {
+	# Packages for python cffi
+	pkgsList="$pkgsList virtualenv python-dev-is-python3 libffi-dev build-essential"
+	# kconfig requirements
+	pkgsList="$pkgsList libncurses-dev"
+	# hub-ctrl
+	pkgsList="$pkgsList libusb-dev"
+	# AVR chip installation and building
+	pkgsList="$pkgsList avrdude gcc-avr binutils-avr avr-libc"
+	# ARM chip installation and building
+	pkgsList="$pkgsList stm32flash libnewlib-arm-none-eabi"
+	pkgsList="$pkgsList gcc-arm-none-eabi binutils-arm-none-eabi pkg-config"
+	# Working with the repository
+	pkgsList="$pkgsList git"
+
+	# Update system package info
+	report_status "Running apt-get update..."
+	sudo apt-get update
+
+	# Install desired packages
+	report_status "Installing packages..."
+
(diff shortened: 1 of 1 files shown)


Kreyren:
image

Terrorism!

sorpigal:
I would not use `set -e` at all in any script. Check the exit code of each program where a failure would be a problem instead.

 Your `die` function ignores its first arg, should probably say `exit $1`.

You still use UPPER CASE variable names without good reason, e.g. PYTHONDIR and SUSTEMDDIR. 

I would not use double quote when my string does not require interpolation.

Kreyren:
> I would not use set -e at all in any script. Check the exit code of each program where a failure would be a problem instead. -- @sorpigal (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227483439)

Python doesn't seem to return a sane exit and it would introduce an unwanted complexity (i want to keep the script easy to read and by the average user)

Sane?

> Your die function ignores its first arg, should probably say exit $1. -- @sorpigal (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227483439)

Thanks for noticing fixed in next commit (It's a quick function that i usually make to avoid doing echo; exit following the semi-standard `die` implementation from other languages)

> You still use UPPER CASE variable names without good reason, e.g. PYTHONDIR and SUSTEMDDIR. -- @sorpigal (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227483439)

Adjusted

> I would not use double quote when my string does not require interpolation. -- @sorpigal (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227483439)

Shellcheck complains about that and disabling it script-wide would obstruct a legitimate SC2086 so i prefer to double-quote them. Sane?

---

Noticed also the change to `python-dev` which doesn't have an install candidate and apt suggests `python-dev-is-python3`

sorpigal:
> Python doesn't seem to return a sane exit and it would introduce an unwanted complexity (i want to keep the script easy to read and use by the average user)
> 
> Sane?

`set -e` is also checking the exit code the same way `prog || die` would, so you have the same problem either way.

> > Your die function ignores its first arg, should probably say exit $1. -- @sorpigal ([#6648 (comment)](https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227483439))
> 
> Thanks for noticing fixed in next commit (It's a quick function that i usually make to avoid doing echo; exit following the semi-standard `die` implementation from other languages)

I would suggest: `die () { printf '%s\n' "$1"; exit ${2-1}; }` so the exit code can be omitted and default to 1, but having it always exit with 1 is reasonable.

> > I would not use double quote when my string does not require interpolation. -- @sorpigal ([#6648 (comment)](https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227483439))
> 
> Shellcheck complains about that and disabling it script-wide would obstruct a legitimate SC2086 so i prefer to double-quote them. Sane?

I said you should not double quote when you don't need interpolation meaning you *should* single quote, not that you should not quote at all.

Kreyren:
> set -e is also checking the exit code the same way prog || die would, so you have the same problem either way. -- @sorpigal (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227490782)

Fair, added `die` traps, but keeping the `exit -e` in case e.g. cat fails to parse the docstring or something unexpected alike to stop the script from continuing, sane?

I would suggest: die () { printf '%s\n' "$1"; exit ${2-1}; } so the exit code can be omitted and default to 1, but having it always exit with 1 is reasonable. -- @sorpigal (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227490782)

`die()` in other languages e.g. python and rust follows the syntax of: `die [exit-code] (message)` as it (usually) provides a default message for the selected exit code. So trying to follow this somewhat-standard integration.

> I said you should not double quote when you don't need interpolation meaning you should single quote, not that you should not quote at all. -- @sorpigal (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2227490782)

Oh i see, unsure how i feel about that styling.. Seems like it makes the code look like someone who cared wrote it so lets use it.

--

Ready to approve?

thijstriemstra:
I'm not sure if it's a great idea to add a debian version-specific installation script. Like, will there be a install-debian-13.sh in the future?

Kreyren:
> I'm not sure if it's a great idea to add a debian version-specific installation script. Like, will there be a install-debian-13.sh in the future? 14, 15? -- @thijstriemstra (https://github.com/Klipper3d/klipper/pull/6648)

Like ideally: Management to generate deb files and submit them to Debian (I don't have time for that atm, please do not send me donations for that)

This works enough for me to not have to run around the hackerspace to control the 3D printer which is the sole reason for me to submit this instead of keeping it private as in my mind someone else might find it useful, beyond that it could be quite easily adjusted to implement a conditional that checks for debian version and adjusts the workflow, but at that point we are basically developing a very weird and fucked up LFS on top of debian and it will take much more developer resources in the future to keep it running.. At that point the repo should just be adjusted to generate a deb files and port them to debian upstream imho. 

For me personally i will be making an OSHW PCB for the printer off of (likely) smoothieboard v2 or using some of OLIMEX's products to do wires going across multiple boards until i figure out how i want it to work and then implement https://github.com/NiXium-org/NiXium where klipper implementation is just a matter of adding `services.klipper.enable = true;` to not need this anymore.

Also the version-specific scripts [are already done for ubuntu](https://github.com/Klipper3d/klipper/blob/master/scripts/install-ubuntu-22.04.sh), which is terrible idea, but it's done away..

FWIW if klipper works for my printer well enough i will likely submit more merge requests that implement automation and refactor some parts of the code that i wouldn't consider sufficient for mission-critical environment.

Sineos:
Unfortunately, the installation instructions are not up to date and at least to my feelings, KIAUH today is the standard way to install Klipper and its ecosystem.
The shell scripts feel a bit like something of the past.

JamesH1978:
Thank you for submitting a PR, please refer to point 3 in "What to expect in a review" in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md and provide a signed off by line. Like a lot of open source we require implicit acceptance of the Developer Certificate of Origin

Thanks
James

JamesH1978:
and on a side note, i hope to have a new `installation.md` ready to go and PR soon

Kreyren:
> Unfortunately, the installation instructions are not up to date and at least to my feelings, KIAUH today is the standard way to install Klipper and its ecosystem. -- @Sineos (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2230651504)

tbh it feels like making a bad solution even worse.. 

> and on a side note, i hope to have a new installation.md ready to go and PR soon -- @JamesH1978 (https://github.com/Klipper3d/klipper/pull/6648#issuecomment-2232682352)

Lets not merge this and do it that way instead then, plz ping me when you have PR ready.

PR #6652, opened 2024-07-24
create config file for creality ender3 max neo

vlad-bil:
Created the configuration file for Creality Ender3 Max Neo based on Creality Ender3 S1 and Creality Ender3 Max for the stock hardware.


JamesH1978:
Thank you for submitting a PR, please refer to point 4 in the guidelines in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md to add the regression tests, and provide a signed off by line as outlined in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

Thanks
James

JamesH1978:
The code you have submitted includes sections that are not available in mainline klipper such as `[bl24c16f]` and we would prefer a [fan] section and not an output pin to run the main parts fan

I will do a full review once you have added printer.test and signed off your PR

Thank
James

config/printer-creality-ender3max-neo-2024.cfg
```diff
@@ -0,0 +1,131 @@
...
+serial: /dev/serial/by-id/usb-1a86_USB_Serial-if00-port0
+restart_method: command
+
+[printer]
+kinematics: cartesian
+max_velocity: 300
+max_accel: 5000
+max_z_velocity: 10
+max_z_accel: 1000
+square_corner_velocity: 5.0
```
jhartleycaw, line 95: There is no need to set square_corner_velocity, 5 is default

Thanks
James

config/printer-creality-ender3max-neo-2024.cfg
```diff
@@ -0,0 +1,131 @@
...
+
+[bltouch]
+sensor_pin: ^PC14
+control_pin: PC13
+x_offset: -31.8
+y_offset: -40.5
+z_offset: 0
+probe_with_touch_mode: true
+stow_on_each_sample: false
```
jhartleycaw, line 104: Please do not set stow_on_each_sample as false, this is a feature that can have safety issues, as the pin is not stowed, therefore obstacles can impede progress or at worse cause bad things to happen. Setting this option is up to the end end user.

Thanks
James

jhartleycaw (changes requested):
Apart from my comments, all looks good.

Thanks
James

PR #6653, opened 2024-07-24
idex_modes: Improved restoring position in RESTORE_DUAL_CARRIAGE_STATE

dmbutyugin:
Previous implementation could crash the idex carriages into each other. The new code moves the idex carriages together, eliminating this risk and decreasing the time needed to restore the carriages positions.

### DIFF
--- a/klippy/kinematics/idex_modes.py
+++ b/klippy/kinematics/idex_modes.py
@@ -4,7 +4,7 @@
 # Copyright (C) 2023  Dmitry Butyugin <dmbutyugin@google.com>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
-import math
+import logging, math
 import chelper
 
 INACTIVE = 'INACTIVE'
@@ -202,14 +202,31 @@ class DualCarriages:
         move_speed = gcmd.get_float('MOVE_SPEED', 0., above=0.)
         toolhead = self.printer.lookup_object('toolhead')
         toolhead.flush_step_generation()
-        pos = toolhead.get_position()
         if gcmd.get_int('MOVE', 1):
+            homing_speed = 99999999.
+            cur_pos = []
             for i, dc in enumerate(self.dc):
                 self.toggle_active_dc_rail(i)
-                saved_pos = saved_state['axes_positions'][i]
-                toolhead.manual_move(
-                        pos[:self.axis] + [saved_pos] + pos[self.axis+1:],
-                        move_speed or dc.get_rail().homing_speed)
+                homing_speed = min(homing_speed, dc.get_rail().homing_speed)
+                cur_pos.append(toolhead.get_position())
+            move_pos = list(cur_pos[0])
+            dl = [saved_state['axes_positions'][i] - cur_pos[i][self.axis]
+                  for i in range(2)]
+            primary_ind = 0 if abs(dl[0]) >= abs(dl[1]) else 1
+            self.toggle_active_dc_rail(primary_ind)
+            move_pos[self.axis] = saved_state['axes_positions'][primary_ind]
+            dc_mode = INACTIVE if min(abs(dl[0]), abs(dl[1])) < 0.000000001 \
+                    else COPY if dl[0] * dl[1] > 0 else MIRROR
+            if dc_mode != INACTIVE:
+                self.dc[1-primary_ind].activate(dc_mode, cur_pos[primary_ind])
+                self.dc[1-primary_ind].override_axis_scaling(
+                        abs(dl[1-primary_ind] / dl[primary_ind]),
+                        cur_pos[primary_ind])
+            toolhead.manual_move(move_pos, move_speed or homing_speed)
+            toolhead.flush_step_generation()
+            # Make sure the scaling coefficients are restored with the mode
+            self.dc[0].inactivate(move_pos)
+            self.dc[1].inactivate(move_pos)
         for i, dc in enumerate(self.dc):
             saved_mode = saved_state['carriage_modes'][i]
             self.activate_dc_mode(i, saved_mode)
@@ -257,3 +274,8 @@ class DualCarriagesRail:
         self.scale = 0.
         self.apply_transform()
         self.mode = INACTIVE
+    def override_axis_scaling(self, new_scale, position):
+        old_axis_position = self.get_axis_position(position)
+        self.scale = math.copysign(new_scale, self.scale)
+        self.offset = old_axis_position - position[self.axis] * self.scale
+        self.apply_transform()
--- a/test/klippy/dual_carriage.cfg
+++ b/test/klippy/dual_carriage.cfg
@@ -61,7 +61,7 @@ pid_Kd: 114
 min_temp: 0
 max_temp: 250
 
-[gcode_macro PARK_extruder0]
+[gcode_macro PARK_extruder]
 gcode:
     G90
     G1 X0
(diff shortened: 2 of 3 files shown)


dmbutyugin:
When working on `generic_cartesian` kinematics, I realized that the current implementation of restoring the positions of IDEX carriages has a bug: depending on the relative positions of the carriages and their intended final destinations, `RESTORE_DUAL_CARRIAGE_STATE` may crash one of the carriages into another one. Instead of fiddling with the restoration order (basically, reimplementing homing heuristics) I decided to implement simultaneous restoration of the carriages positions. Since they move together in a linear motion (and IDEX is supported only on cartesian-style linear kinematics), it can be guaranteed that the minimum distance between the carriages will be `min(abs(start_pos_dc0-start_pos_dc1), abs(end_pos_dc0-end_pos_dc1))`, and assuming both the current and stored positions were valid, it means that the carriages cannot crash into each other during the motion. And besides the safety, which is the primary concern, this implementation also speeds up restoring carriages positions and can probably be used in the carriage switch macros (`T0`/`T1`) to speed up the tool change process.

I decided to create this PR first such that I don't bundle too many changes into `generic_cartesian` PR later and so that the latter has no functional changes. Besides, preparing that PR will take some time, so it is better to commit this fix ahead of that.

@miklschmidt @HelgeKeck FYI, though I doubt this would break RatOS. Also, I think you have some similar implementation for the tool change process, however I haven't looked at yours and it probably does it differently.

HelgeKeck:
hey, i monitored all the discussions you had with keving abotu the new branch. sorry that i couldnt test anything yet, i jstu cant find the time atm. will try to give it a shot soon. seems you are makign progress, thats good. thank you for that

HelgeKeck:
> I realized that the current implementation of restoring the positions of IDEX carriages has a bug
we are completely avoiding this command in ratos bc we had multiple issues with it

about the toolchanges, we are doing htem completely in jinja, they are just basically 2 copy moves, so toolheads move always at the same time. not sure if this is waht you are talking about

dmbutyugin:
> we are completely avoiding this command in ratos bc we had multiple issues with it

Can you elaborate what kind of issues you have had? In case something can be fixed or improved.

FWIW, I think that now it can be used roughly like this:

In the start code, one can do
```
G28
; Initial initialization of the states
SET_DUAL_CARRIAGE CARRIAGE=1
SAVE_DUAL_CARRIAGE_STATE NAME=tool1
SET_DUAL_CARRIAGE CARRIAGE=0 
SAVE_DUAL_CARRIAGE_STATE NAME=tool0
...
```
then for the actual tool changing macros
```
[gcode_macro T0]
gcode:
  {% if printer.dual_carriage.carriage_0 == 'INACTIVE' %}
  SAVE_DUAL_CARRIAGE_STATE NAME=tool1
  ACTIVATE_EXTRUDER EXTRUDER=extruder
  {% if cleanup_nozzle %}
  SET_DUAL_CARRIAGE CARRIAGE=0
  G0 X... F...
  G0 X... F...
  ...
  {% endif %}
  RESTORE_DUAL_CARRIAGE_STATE NAME=tool0 MOVE=1 MOVE_SPEED=300
```
and similarly for `T1`, just swapping carriage indices and extruder. And of course, you may need to insert temperature adjusting commands, extruder purging, etc.

> about the toolchanges, we are doing htem completely in jinja, they are just basically 2 copy moves, so toolheads move always at the same time. not sure if this is what you are talking about

Yes, with that you have to do two moves. In case of this PR, it is just one COPY or MIRROR move - depending on the destination positions relative to the current positions of the carriages. But this is only possible with the special code modification to adjust the scaling of the motion of one of the carriages, since they may need to travel different distances. I just did not like the stutter of the two moves that one toolhead experiences, so I went with this implementation.

HelgeKeck:
sorry if i havent been clear enough, in the toolchange macro we have absolutely no use for this command at all, the reason is that we already know at the time we want to make the toolchange where the next toolhead has to be, this happens all on the fly. In our case there is just no situation during the toolchanges where this command makes any sense.

but in other situations where this command would be useful i haad in the past some very strange behaviour, i dont rememebr now everything, have to dig the code and tr to remember what happend there. can do that tomorrow for you.

HelgeKeck:
jsut to clarify, our T0 or T1 macro receives parameters from the gcode where the next move happens, means we are moving the new toolhead directly to this position and the old toolhead at the same time to its parking position, hop this makes more sense

HelgeKeck:
this looks in our case like this, x, y and z are the coordinates where the next gcode move for the demanded toolhead starts

```
[gcode_macro T1]
gcode:
	{% set x = params.X|default(-1.0)|float %}
	{% set y = params.Y|default(-1.0)|float %}
	{% set z = params.Z|default(0.0)|float %}

```

HelgeKeck:
> Yes, with that you have to do two moves. In case of this PR, it is just one COPY or MIRROR move - depending on the destination positions relative to the current positions of the carriages. But this is only possible with the special code modification to adjust the scaling of the motion of one of the carriages, since they may need to travel different distances. I just did not like the stutter of the two moves that one toolhead experiences, so I went with this implementation.

i always wanted to have this feature, independen toolhead moves at the same time, for exaclty this reason

but even with 2 moves we do not have any stutter at all, it is even absolutely flawless and super fast, way below 0.5 seconds for the complete process, barely to see with the naked eye that these are two moves

HelgeKeck:
will it be possible to make this independent moves manually form gcode?

dmbutyugin:
> sorry if i havent been clear enough, in the toolchange macro we have absolutely no use for this command at all,

No problem, this was just a demo for other readers and reviewers of this PR, how this functionality can be useful. Your approach is also fine, to be clear.

> will it be possible to make this independent moves manually form gcode?

The code I wrote takes 2 pairs of coordinates - starting positions of the 2 carriages and their final positions, and executes the coordinated move to get them to the final requested positions. So in principle, it would be possible to expose it as a separate GCode command as well. However, since Klipper natively supports only 4 coordinates in G0/G1, this may have implications on the API of a supposedly newly introduced GCode command with my `generic_cartesian` kinematics which may very well support `dual_carriage` on both X and Y axes. Basically, the new command would have to support two coordinates per carriage, something like `MOVE_DUAL_CARRIAGE CARRIAGE0=<X0>,<Y0> CARRIAGE1=<X1>,<Y1>` as opposed to a more simple case of single-axis `dual_carriage` (`MOVE_DUAL_CARRIAGE CARRIAGE0=<C0> CARRIAGE1=<C1>`). Also notably the code I currently wrote does not perform any checks on the validity of the move it attempts to execute, assuming that the saved position being restored was valid and the user somehow managed to get from that position to the current position, and this assumption is sufficient to prove the validity of the move to restore the positions. However, in a more generic case such validity checks must be performed as usual. Thus I don't want to deal with that right now, so maybe I'll get to that after (and if) `generic_cartesian` gets merged.

HelgeKeck:
> Also notably the code I currently wrote does not perform any checks on the validity of the move it attempts to executeyeah, 
yeah, thats the reason our toolchange macro are extremly long, there is a lot to teke into account

however, i think it would be super beneficial if thsi would be exposed as a gcode command, this would make seamless toolchanges for idex possible, imo this is how idex is supposed to work

KevinOConnor:
Thanks.  Dmitry, let me know when you feel this is ready to commit, and I will do so.

Cheers,
-Kevin

dmbutyugin:
Thanks Kevin. I think the PR is ready to merge, I think there's no more feedback, and I do not plan to make any further adjustments to the code.

KevinOConnor:
Thanks.

-Kevin

PR #6656, opened 2024-07-31
servo: Asynchronous adjustments of servo position

dmbutyugin:
This change follows the same approach as implemented for fan control. The change removes the move queue flushing when changing servo position, which does not appear to be necessary. This can be beneficial, for example, for WS7040-based cooling on IDEX setups where the servo can be used to control the air flow between the toolheads, with this change eliminating micro-stutters of the toolhead on servo position adjustment.

### DIFF
--- a/klippy/extras/servo.py
+++ b/klippy/extras/servo.py
@@ -58,13 +58,15 @@ class PrinterServo:
         return width * self.width_to_value
     cmd_SET_SERVO_help = "Set servo angle"
     def cmd_SET_SERVO(self, gcmd):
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Looks fine to me.  I'll give a couple of days to see if there are any comments, and then look to commit.

Cheers,
-Kevin

dmbutyugin:
OK, thanks, sounds good from my side.

KevinOConnor:
Thanks.

-Kevin

PR #6657, opened 2024-08-01
gcode_arc: use arc generator - reduce blocking time

nefelim4ag:
TLDR:
On doing arcs klipper can spend a 'lot' of time locked in planArc creating segments.
5..30ms depends on host performance and arc length, segments count/length.
Performance-wise, this change is marginal +1% in normal circumstances - this is more like suggestion.
In the limit it will be better, the more segments we have.

Thanks.

_For profiling, I used perf/py-spy, which is not perfectly accurate. So, long sampling durations, fixed CPU freq, and high sampling count._

---
For context:
I sometimes heard and experienced strange CPU spikes when using arcs.
So, I was curious.
Real usage is beneath in C code, at itersolve_gen_steps_range, pa_range_integrate & etc.
Because a lot of segments generate a lot of small moves & etc.
This is a different topic.

### DIFF
--- a/klippy/extras/gcode_arcs.py
+++ b/klippy/extras/gcode_arcs.py
@@ -39,8 +39,6 @@ class ArcSupport:
         self.gcode.register_command("G18", self.cmd_G18)
         self.gcode.register_command("G19", self.cmd_G19)
 
-        self.Coord = self.gcode.Coord
-
         # backwards compatibility, prior implementation only supported XY
         self.plane = ARC_PLANE_X_Y
 
@@ -64,24 +62,28 @@ class ArcSupport:
         if not gcodestatus['absolute_coordinates']:
             raise gcmd.error("G2/G3 does not support relative move mode")
         currentPos = gcodestatus['gcode_position']
+        absolut_extrude = gcodestatus['absolute_extrude']
 
         # Parse parameters
-        asTarget = self.Coord(x=gcmd.get_float("X", currentPos[0]),
-                              y=gcmd.get_float("Y", currentPos[1]),
-                              z=gcmd.get_float("Z", currentPos[2]),
-                              e=None)
+        asTarget = [gcmd.get_float("X", currentPos[0]),
+                    gcmd.get_float("Y", currentPos[1]),
+                    gcmd.get_float("Z", currentPos[2])]
 
         if gcmd.get_float("R", None) is not None:
             raise gcmd.error("G2/G3 does not support R moves")
 
         # determine the plane coordinates and the helical axis
-        asPlanar = [ gcmd.get_float(a, 0.) for i,a in enumerate('IJ') ]
+        I = gcmd.get_float('I', 0.)
+        J = gcmd.get_float('J', 0.)
+        asPlanar = (I, J)
         axes = (X_AXIS, Y_AXIS, Z_AXIS)
         if self.plane == ARC_PLANE_X_Z:
-            asPlanar = [ gcmd.get_float(a, 0.) for i,a in enumerate('IK') ]
+            K = gcmd.get_float('K', 0.)
+            asPlanar = (I, K)
             axes = (X_AXIS, Z_AXIS, Y_AXIS)
         elif self.plane == ARC_PLANE_Y_Z:
-            asPlanar = [ gcmd.get_float(a, 0.) for i,a in enumerate('JK') ]
+            K = gcmd.get_float('K', 0.)
+            asPlanar = (J, K)
             axes = (Y_AXIS, Z_AXIS, X_AXIS)
 
         if not (asPlanar[0] or asPlanar[1]):
@@ -91,20 +93,20 @@ class ArcSupport:
         asF = gcmd.get_float("F", None)
 
         # Build list of linear coordinates to move
-        coords = self.planArc(currentPos, asTarget, asPlanar,
-                              clockwise, *axes)
+        segments, coords = self.planArc(currentPos, asTarget, asPlanar,
+                                        clockwise, *axes)
         e_per_move = e_base = 0.
         if asE is not None:
-            if gcodestatus['absolute_extrude']:
+            if absolut_extrude:
                 e_base = currentPos[3]
-            e_per_move = (asE - e_base) / len(coords)
+            e_per_move = (asE - e_base) / segments
 
         # Convert coords into G1 commands
         for coord in coords:
             g1_params = {'X': coord[0], 'Y': coord[1], 'Z': coord[2]}
             if e_per_move:
                 g1_params['E'] = e_base + e_per_move
-                if gcodestatus['absolute_extrude']:
+                if absolut_extrude:
                     e_base += e_per_move
             if asF is not None:
                 g1_params['F'] = asF
@@ -162,20 +164,20 @@ class ArcSupport:
         coords = []
         for i in range(1, int(segments)):
             dist_Helical = i * linear_per_segment
-            cos_Ti = math.cos(i * theta_per_segment)
-            sin_Ti = math.sin(i * theta_per_segment)
+            c_theta = i * theta_per_segment
+            cos_Ti = math.cos(c_theta)
+            sin_Ti = math.sin(c_theta)
             r_P = -offset[0] * cos_Ti + offset[1] * sin_Ti
             r_Q = -offset[0] * sin_Ti - offset[1] * cos_Ti
 
-            # Coord doesn't support index assignment, create list
-            c = [None, None, None, None]
+            c = [None, None, None]
             c[alpha_axis] = center_P + r_P
             c[beta_axis] = center_Q + r_Q
             c[helical_axis] = currentPos[helical_axis] + dist_Helical
-            coords.append(self.Coord(*c))
+            coords.append(c)
 
         coords.append(targetPos)
-        return coords
+        return segments, coords
 
 def load_config(config):
     return ArcSupport(config)
--- a/klippy/extras/gcode_arcs.py
+++ b/klippy/extras/gcode_arcs.py
@@ -89,29 +89,9 @@ class ArcSupport:
         if not (asPlanar[0] or asPlanar[1]):
             raise gcmd.error("G2/G3 requires IJ, IK or JK parameters")
 
-        asE = gcmd.get_float("E", None)
-        asF = gcmd.get_float("F", None)
-
-        # Build list of linear coordinates to move
-        segments, coords = self.planArc(currentPos, asTarget, asPlanar,
-                                        clockwise, *axes)
-        e_per_move = e_base = 0.
-        if asE is not None:
-            if absolut_extrude:
-                e_base = currentPos[3]
-            e_per_move = (asE - e_base) / segments
-
-        # Convert coords into G1 commands
-        for coord in coords:
-            g1_params = {'X': coord[0], 'Y': coord[1], 'Z': coord[2]}
-            if e_per_move:
-                g1_params['E'] = e_base + e_per_move
-                if absolut_extrude:
-                    e_base += e_per_move
-            if asF is not None:
-                g1_params['F'] = asF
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Hi,

I'm not sure what you are proposing.  The code changes here would increase cpu load and would not improve latency.  It would move the cpu intensive part of `_cmd_inner()` from the call to `self.palnArc()` just a few lines down to `for coord in coords`.  So no change in latency.

If one wanted to ensure other events can run in the main Python thread then adding calls to `reactor.pause(0.0)` in the inner loop would accomplish that.  However, care should be taken as too many `pause()` calls can increase overhead.

For what it is worth, a simpler solution might be to change to: `self.mm_per_arc_segment = config.getfloat('resolution', 1., minval=0.100)`.  That is, fix by prohibiting absurdly small arc segments.  There should not be a problem with the existing code if the arc size is reasonable.

I may be missing something though.

Cheers,
-Kevin

KevinOConnor:
Oh - I see now.  You are proposing to intersperse `gcode_move.cmd_G1()` in the middle of the calculations so that it will call `reactor.pause()` in the inner loop if the resulting move commands are scheduled far into the future.

I don't see an issue with doing that.  FWIW, it wasn't obvious to me that was the intent.  Might be better to refactor the loops to make that more clear.

-Kevin

nefelim4ag:
_I think about klippy as an async application where it should not block if possible, to have the opportunity to do other stuff.
Maybe I'm wrong here in that assumption or just overestimated the importance of that at some point._

So current code is doing the:
G3 -> _cmd_innter -> planArc() (block for some time)
_cmd_innder -> cmd_G1 -> ... -> toolhead.move()

So, below toolhead.move() it will actually do progress, like pause()/add_move()/flush()(->itersolve).
That is what I mean by reducing latency/blocking here, my bad for not making it clear.
By unblocking the loop at planArc() there actually should be a more stable/predictable saturation of trapq.

If we limit minimal resolution, it will be harder to generate load on Klipper
How I am supposed to overload it then? (joking)

Yes, refactoring it out will be better, as this is acceptable, I will do it.
Initially, I chose the generator here because it looks cleaner, but you are right about hiding the real intention.

Thanks

KevinOConnor:
> So, below toolhead.move() it will actually do progress, like pause()/add_move()/flush()(->itersolve).

Good point.  Note that only occurs if the moves take enough time to be scheduled sufficiently into the future.  Lots of tiny, near instantaneous, moves will likely not result in a delay or lookahead flush.

> By unblocking the loop at planArc() there actually should be a more stable/predictable saturation of trapq.

That's fine if you want to do that.

FWIW, one could move planArc() to its own class.  Something like the following (totally untested):
```diff
--- a/klippy/extras/gcode_arcs.py
+++ b/klippy/extras/gcode_arcs.py
@@ -39,8 +39,6 @@ class ArcSupport:
         self.gcode.register_command("G18", self.cmd_G18)
         self.gcode.register_command("G19", self.cmd_G19)
 
-        self.Coord = self.gcode.Coord
-
         # backwards compatibility, prior implementation only supported XY
         self.plane = ARC_PLANE_X_Y
 
@@ -66,10 +64,9 @@ class ArcSupport:
         currentPos = gcodestatus['gcode_position']
 
         # Parse parameters
-        asTarget = self.Coord(x=gcmd.get_float("X", currentPos[0]),
-                              y=gcmd.get_float("Y", currentPos[1]),
-                              z=gcmd.get_float("Z", currentPos[2]),
-                              e=None)
+        asTarget = (gcmd.get_float("X", currentPos[0]),
+                    gcmd.get_float("Y", currentPos[1]),
+                    gcmd.get_float("Z", currentPos[2]))
 
         if gcmd.get_float("R", None) is not None:
             raise gcmd.error("G2/G3 does not support R moves")
@@ -91,16 +88,18 @@ class ArcSupport:
         asF = gcmd.get_float("F", None)
 
         # Build list of linear coordinates to move
-        coords = self.planArc(currentPos, asTarget, asPlanar,
-                              clockwise, *axes)
+        coords = planArc(currentPos, asTarget, asPlanar,
+                         clockwise, self.mm_per_arc_segment, *axes)
+        num_segments = coords.get_num_segments()
         e_per_move = e_base = 0.
         if asE is not None:
             if gcodestatus['absolute_extrude']:
                 e_base = currentPos[3]
-            e_per_move = (asE - e_base) / len(coords)
+            e_per_move = (asE - e_base) / num_segments
 
         # Convert coords into G1 commands
-        for coord in coords:
+        for i in range(num_segments):
+            coord = coords.get_next_segment()
             g1_params = {'X': coord[0], 'Y': coord[1], 'Z': coord[2]}
             if e_per_move:
                 g1_params['E'] = e_base + e_per_move
@@ -111,16 +110,24 @@ class ArcSupport:
             g1_gcmd = self.gcode.create_gcode_command("G1", "G1", g1_params)
             self.gcode_move.cmd_G1(g1_gcmd)
 
-    # function planArc() originates from marlin plan_arc()
-    # https://github.com/MarlinFirmware/Marlin
-    #
-    # The arc is approximated by generating many small linear segments.
-    # The length of each segment is configured in MM_PER_ARC_SEGMENT
-    # Arcs smaller then this value, will be a Line only
-    #
-    # alpha and beta axes are the current plane, helical axis is linear travel
-    def planArc(self, currentPos, targetPos, offset, clockwise,
-                alpha_axis, beta_axis, helical_axis):
+# planArc code originates from marlin plan_arc()
+# https://github.com/MarlinFirmware/Marlin
+#
+# The arc is approximated by generating many small linear segments.
+# The length of each segment is configured in MM_PER_ARC_SEGMENT
+# Arcs smaller then this value, will be a Line only
+#
+# alpha and beta axes are the current plane, helical axis is linear travel
+class planArc:
+    def __init__(self, currentPos, targetPos, offset, clockwise,
+                 mm_per_arc_segment, alpha_axis, beta_axis, helical_axis):
+        self.currentPos = currentPos
+        self.targetPos = targetPos
+        self.offset = offset
+        self.alpha_axis = alpha_axis
+        self.beta_axis = beta_axis
+        self.helical_axis = helical_axis
+
         # todo: sometimes produces full circles
 
         # Radius vector from center to current location
@@ -128,8 +135,8 @@ class ArcSupport:
         r_Q = -offset[1]
 
         # Determine angular travel
-        center_P = currentPos[alpha_axis] - r_P
-        center_Q = currentPos[beta_axis] - r_Q
+        self.center_P = center_P = currentPos[alpha_axis] - r_P
+        self.center_Q = center_Q = currentPos[beta_axis] - r_Q
         rt_Alpha = targetPos[alpha_axis] - center_P
         rt_Beta = targetPos[beta_axis] - center_Q
         angular_travel = math.atan2(r_P * rt_Beta - r_Q * rt_Alpha,
@@ -154,28 +161,31 @@ class ArcSupport:
             mm_of_travel = math.hypot(flat_mm, linear_travel)
         else:
             mm_of_travel = math.fabs(flat_mm)
-        segments = max(1., math.floor(mm_of_travel / self.mm_per_arc_segment))
-
-        # Generate coordinates
-        theta_per_segment = angular_travel / segments
-        linear_per_segment = linear_travel / segments
-        coords = []
-        for i in range(1, int(segments)):
-            dist_Helical = i * linear_per_segment
-            cos_Ti = math.cos(i * theta_per_segment)
-            sin_Ti = math.sin(i * theta_per_segment)
-            r_P = -offset[0] * cos_Ti + offset[1] * sin_Ti
-            r_Q = -offset[0] * sin_Ti - offset[1] * cos_Ti
-
-            # Coord doesn't support index assignment, create list
-            c = [None, None, None, None]
-            c[alpha_axis] = center_P + r_P
-            c[beta_axis] = center_Q + r_Q
-            c[helical_axis] = currentPos[helical_axis] + dist_Helical
-            coords.append(self.Coord(*c))
-
-        coords.append(targetPos)
-        return coords
+        self.segments = max(1., math.floor(mm_of_travel / mm_per_arc_segment))
+
+        self.theta_per_segment = angular_travel / segments
+        self.linear_per_segment = linear_travel / segments
+        self.current_segment = 0
+
+    def get_num_segments(self):
+        return self.segments
+
+    def get_next_segment(self):
+        self.current_segment += 1
+        if self.current_segment >= self.segments:
+            return self.targetPos
+
+        dist_Helical = self.current_segment * self.linear_per_segment
+        cos_Ti = math.cos(self.current_segment * self.theta_per_segment)
+        sin_Ti = math.sin(self.current_segment * self.theta_per_segment)
+        r_P = -self.offset[0] * cos_Ti + self.offset[1] * sin_Ti
+        r_Q = -self.offset[0] * sin_Ti - self.offset[1] * cos_Ti
+
+        c = [None, None, None, None]
+        c[self.alpha_axis] = self.center_P + r_P
+        c[self.beta_axis] = self.center_Q + r_Q
+        c[self.helical_axis] = self.currentPos[self.helical_axis] + dist_Helical
+        return c
 
 def load_config(config):
     return ArcSupport(config)
```

-Kevin

nefelim4ag:
I separated the file and did some mocking with the actual code arc, to validate code and get some perf numbers.
And used loop to run code N times under profiler:
```
for i in range(0, 200):
    arcs.cmd_G3(gcmd)
```
[gcode_arcs.py.gz](https://github.com/user-attachments/files/16484345/gcode_arcs.py.gz)
`python3 -m cProfile ./gcode_arcs.py`

Python 3.12, I chose the best result
Current:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.827    2.827 {built-in method builtins.exec}
        1    0.000    0.000    2.827    2.827 gcode_arcs.py:1(<module>)
      200    0.031    0.000    2.826    0.014 gcode_arcs.py:75(cmd_G3)
      200    0.368    0.002    2.795    0.014 gcode_arcs.py:87(_cmd_inner)
   696400    1.481    0.000    1.481    0.000 gcode_arcs.py:360(create_gcode_command)
      200    0.515    0.003    0.752    0.004 gcode_arcs.py:176(planArc)
   696400    0.121    0.000    0.193    0.000 gcode_arcs.py:352(cmd_G1)
```
Generator
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.736    2.736 {built-in method builtins.exec}
        1    0.000    0.000    2.736    2.736 gcode_arcs.py:1(<module>)
      200    0.000    0.000    2.735    0.014 gcode_arcs.py:75(cmd_G3)
      200    0.469    0.002    2.735    0.014 gcode_arcs.py:87(_cmd_inner)
   696400    1.487    0.000    1.487    0.000 gcode_arcs.py:360(create_gcode_command)
   696600    0.422    0.000    0.567    0.000 gcode_arcs.py:235(_gen_segments)
   696400    0.133    0.000    0.210    0.000 gcode_arcs.py:352(cmd_G1)
``` 
Manual generator with class
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.836    2.836 {built-in method builtins.exec}
        1    0.000    0.000    2.836    2.836 gcode_arcs.py:1(<module>)
      200    0.000    0.000    2.835    0.014 gcode_arcs.py:75(cmd_G3)
      200    0.537    0.003    2.835    0.014 gcode_arcs.py:87(_cmd_inner)
   696400    1.474    0.000    1.474    0.000 gcode_arcs.py:360(create_gcode_command)
   696400    0.475    0.000    0.612    0.000 gcode_arcs.py:306(get_next_segment)
   696400    0.133    0.000    0.210    0.000 gcode_arcs.py:352(cmd_G1)
```

I don't know why generators are faster, but I think the custom class is slower because accessing instance attributes is slower than local vars.

nefelim4ag:
I came up with the solution I think is less complicated - to move G1 emit below to planArc.

Thanks

KevinOConnor:
Thanks!  Looks good to me.  I'll give a couple of days to see if there are any further comments, and otherwise look to commit.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6658, opened 2024-08-02
Modify Ender3 V2 config comments to warn users to disconnect the display

wildjerry:
I spent about three hours chasing firmware issues caused by this issue, and wanted to save others from wasting their time. 

See also https://youtu.be/gfZ9Lbyh8qU?t=422

Should this note be applied to other Creality Printers? I'm pretty sure the E3V2 isn't the only printer that ships with this screen, and it's available as an upgrade module too.

### DIFF
--- a/config/printer-creality-ender3-v2-2020.cfg
+++ b/config/printer-creality-ender3-v2-2020.cfg
@@ -12,7 +12,8 @@
 # Flash this firmware by copying "out/klipper.bin" to a SD card and
 # turning on the printer with the card inserted. The firmware
 # filename must end in ".bin" and must not match the last filename
-# that was flashed.
+# that was flashed. You may need to remove the display in order to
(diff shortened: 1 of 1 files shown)


jhartleycaw:
You shouldn't need to remove the screen to flash these, in fact they can help identify a flashed board, as these boards have no other flash signs, having a screen that only shows the creality logo or a blank screen after flash is a good identifier that the stock firmware has been overwritten. It should still flash. 

I do agree though that a comment should be added to mention the screen on these printers and most "modern" creality screens are not compatible(the old style blue screen box style ender pro screen is supported though).

Also you will need to sign off your PR for it to valid to merge. Please see point 3 in `What to expect in a review` in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

Thanks
James

PR #6659, opened 2024-08-02
Assorted minor hx71x changes

KevinOConnor:
Hi @garethky,

I noticed a few minor things in the recent bulk_adc merge.

1. The bulk adc report includes the adc value as a float, but the header and docs weren't updated to reflect that.
2. I think it would be a little better to check for hx71x timing overflows via the timer handler.  This check should be a little more accurate as it includes the time needed to wake up the query task.  It should also be a little more efficient to check.
3. I wonder if we could automatically restart hx71x samples after an error if the code receives 4 valid looking samples.  The idea here is: In the event of a read overrun, it's likely the sample is bad and it should be discarded; the following sample is also likely bad as some leading bits may have been erroneously read in the previous sample; the next sample is also likely bad, because the gain/channel settings may have been altered; the next sample is likely okay, but prudent to discard as well; the following samples should be valid.  If the sensor fails to recover, the host will still automatically restart the sensor.

I don't actually have this hardware to test, so I'd appreciate feedback.

The 3rd commit is a bit speculative on my part, so I understand if it's not a good idea.

-Kevin

### DIFF
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -375,9 +375,10 @@ A request may look like:
 `{"id": 123, "method":"hx71x/dump_hx71x",
 "params": {"sensor": "load_cell", "response_template": {}}}`
 and might return:
-`{"id": 123,"result":{"header":["time","counts"]}}`
+`{"id": 123,"result":{"header":["time","counts","value"]}}`
 and might later produce asynchronous messages such as:
-`{"params":{"data":[[3292.432935, 562534], [3292.4394937, 5625322]]}}`
+`{"params":{"data":[[3292.432935, 562534, 0.067059278],
+[3292.4394937, 5625322, 0.670590639]]}}`
 
 ### ads1220/dump_ads1220
 
@@ -390,9 +391,10 @@ A request may look like:
 `{"id": 123, "method":"ads1220/dump_ads1220",
 "params": {"sensor": "load_cell", "response_template": {}}}`
 and might return:
-`{"id": 123,"result":{"header":["time","counts"]}}`
+`{"id": 123,"result":{"header":["time","counts","value"]}}`
 and might later produce asynchronous messages such as:
-`{"params":{"data":[[3292.432935, 562534], [3292.4394937, 5625322]]}}`
+`{"params":{"data":[[3292.432935, 562534, 0.067059278],
+[3292.4394937, 5625322, 0.670590639]]}}`
 
 ### pause_resume/cancel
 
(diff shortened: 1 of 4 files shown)


garethky:
1. Doc updates look correct, my bad, I missed that

2. I think this makes sense. 

3. I spent an hour looking at this and at the data sheet and I'm not sure if this works or not. It must be that the output register refills and the counter resets when a conversion is finished. But the edge cases when the read happens late are hard to reason about. e.g. you cant really tell if the `extras_mask` check is valid because it could just be that those bits of the sample were 1's. If you read across a conversion boundary that could happen. Its much safer to read near the data ready signal and then be strict about not allowing the read to take longer than 1 conversion cycle (- time between checks). If you do have an overflow

Doing the chip reset procedure (`PD_SCK` high for 60us) guarantees a recovery. This seems like an edge case and I wanted to ship whats guaranteed to work. I don't care about faster recovery times in these edge cases.

Also, I though we had to be concerned that serial packet delivery from the MCU is not guaranteed. I let the errors latch so that if a packet gets dropped one will eventually make it to the Host with errors in it.

KevinOConnor:
Okay, thanks for reviewing.  I'll drop commit 3 from this PR.

> This seems like an edge case and I wanted to ship whats guaranteed to work. I don't care about faster recovery times in these edge cases.

That's a good point.

> Also, I though we had to be concerned that serial packet delivery from the MCU is not guaranteed. I let the errors latch so that if a packet gets dropped one will eventually make it to the Host with errors in it.

Yes - that is correct.  Resuming measurements would only be valid if it was felt those measurements were accurate.  It wouldn't make a difference during homing operations as the loadcell code would need to issue a `trsync_do_trigger()` on the first error anyway (to get a rapid halt of the steppers as no other error mechanism responds as fast).  Faster recovery would only be uesful for debugging, weight measurements, extruder pressure measurements or other niche contexts.  You make a good point about this being an edge case though.

> But the edge cases when the read happens late are hard to reason about. e.g. you cant really tell if the extras_mask check is valid because it could just be that those bits of the sample were 1's. 

Agreed.  My thinking is, if the code definitely does issue the reads in under 0.8 sample time, the correct trailing 1s are present, the sensor definitely does change the line to 0 on its own, and this pattern repeats a minimum of 3 times - then the code must be back in sync with the sensor.  That is, repeatedly issuing timely reads and repeatedly observing the sensor transition from 1 to 0 on its own seems like a good indicator that synchronization has been obtained.

Cheers,
-Kevin

PR #6660, opened 2024-08-03
Fix getting wrong ADC value on PA0 of STM32G431

Huskia:
I built my own 3D printer board with STM32G431 and flashed klipper firmware. I connected PA0 to heater bed sensor but always get random value. Finally I found that in array "adc_pins", the first element is 0 but GPIO('A', 0) equals 0, too. This causes the "gpio_adc_setup" functoin will get the wrong ADC channel of PA0. Here I changed it to 0xFF and it works fine.

### DIFF
--- a/src/stm32/stm32h7_adc.c
+++ b/src/stm32/stm32h7_adc.c
@@ -92,7 +92,7 @@ static const uint8_t adc_pins[] = {
   #endif
     0,             //             VREFINT
 #elif CONFIG_MACH_STM32G4
-    0,                      // [0] vssa
+    0xFF,                   // [0] vssa
     GPIO('A', 0),           // [1]
     GPIO('A', 1),           // [2]
     GPIO('A', 2),           // [3]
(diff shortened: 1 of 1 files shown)


jhartleycaw:
Thank you for submitting a PR, please provide a signed off by line as outlined in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

Thanks
James

Huskia:
complement information is as follows:
module: src/stm32/stm32h7_adc.c
Signed-off-by: Nicholas Huskie <huskie@idealfuture.org.cn>
Thanks

KevinOConnor:
Thanks.  I agree we should make this change.  Any chance you could change this patch to introduce a define (eg, `INVALID`) for 0xff, and extend the fix to CONFIG_MACH_STM32H7 and stm32l4 parts of the existing ifdef?

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6665, opened 2024-08-10
Warning Fix: Current code produces warnings for possible value overflows.

bevanweiss:
As the input values are uint8_t types, any shift may result in value loss.
Explicit promotion to the output type (uint32_t) keeps things safe.

Have also changed the int32_t in ads1220_read_adc to uint32_t, type promotion and bit manipulation are a bit 'weird' on signed integers, so keep it as an unsigned to align with following function call parameter type.
Have retained the prior explicit sign extension logic however.

### DIFF
--- a/src/sensor_hx71x.c
+++ b/src/sensor_hx71x.c
@@ -27,8 +27,8 @@ struct hx71x_adc {
 };
 
 #define BYTES_PER_SAMPLE 4
-#define SAMPLE_ERROR_DESYNC 1 << 31
-#define SAMPLE_ERROR_READ_TOO_LONG 1 << 30
+#define SAMPLE_ERROR_DESYNC 1L << 31
+#define SAMPLE_ERROR_READ_TOO_LONG 1L << 30
 
 static struct task_wake wake_hx71x;
 
--- a/src/sensor_ads1220.c
+++ b/src/sensor_ads1220.c
@@ -86,7 +86,9 @@ ads1220_read_adc(struct ads1220_adc *ads1220, uint8_t oid)
     barrier();
 
     // create 24 bit int from bytes
-    int32_t counts = (msg[0] << 16) | (msg[1] << 8) | msg[2];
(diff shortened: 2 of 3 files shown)


KevinOConnor:
Thanks.  The change looks fine, but commits need to have a signed-off-by line - see https://www.klipper3d.org/CONTRIBUTING.html .

Also, fwiw, the issue isn't with uint8_t as C will implicitly convert most operations to an `int` type.  The problem is that `int` is 16-bit on AVR, which can cause overflows (and compile warnings) on that architecture.

Cheers,
-Kevin

bevanweiss:
> Thanks. The change looks fine, but commits need to have a signed-off-by line - see https://www.klipper3d.org/CONTRIBUTING.html .

Was there a different issue?
the commit definitely had a signed-off line (I've done a great job spelling my last name wrong of course...)
image

KevinOConnor:
Sorry, I missed that the signed-off-by line was in fact present in the commit message.

Just for completeness, can you provide the signed-off-by line with the correct spelling?

-Kevin

bevanweiss:
> Just for completeness, can you provide the signed-off-by line with the correct spelling?
> 
> -Kevin

Done, thanks

KevinOConnor:
Thanks.

-Kevin

src/sensor_hx71x.c
```diff
@@ -27,8 +27,8 @@ struct hx71x_adc {
 };
 
 #define BYTES_PER_SAMPLE 4
-#define SAMPLE_ERROR_DESYNC 1 << 31
-#define SAMPLE_ERROR_READ_TOO_LONG 1 << 30
+#define SAMPLE_ERROR_DESYNC 1L << 31
+#define SAMPLE_ERROR_READ_TOO_LONG 1L << 30
```
thetic, line 31: `UINT32_C(1)`

https://en.cppreference.com/w/c/types/integer#Function_macros_for_minimum-width_integer_constants

PR #6666, opened 2024-08-12
Update generic-fysetc-s6-v2.cfg

Citruzzzor:
Fixed making MCU receipt


Sineos:
IMO this change is not correct and also contradicts the information in the Fysetc Wiki here: https://wiki.fysetc.com/FYSETC_S6/#3-upload-the-firmware

JamesH1978:
as far as i can see, for the v2 variants of this board are now using a 0x8008000 bootloader, so 32kib, and not "no bootloader". This should be added either as an addition to the main S6 file in the comments to explain, or as a seperate S6 V2 config if the pins are different, which it doesnt look as if they are.

Using "no bootloader" will work, as klipper will still run at 0x8000000 but the bootloader will be wiped away.

Thanks

James

PR #6668, opened 2024-08-12
temperature_probe: add 'max_valid_temp' option

Arksine:
This pull request addresses an issue with users unable to calibrate across a wide coil temperature range.  Specifically, unenclosed printers commonly achieve a maximum temperature between 50-55C.  The result is a calibration with polynomials that intersect at higher temperatures.  The `temperature_probe` module checks for this condition post-calibration, and said users are experiencing failures.

This patch introduces the `max_valid_temp` config option, which replaces the hard coded maximum validation temperature of 120C.  This option will not impact current calibrations.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2458,6 +2458,10 @@ postfix for both sections.
 #   "calibration_extruder_temp" option is set.  Its recommended to heat
 #   the extruder some distance from the bed to minimize its impact on
 #   the probe coil temperature.  The default is 50.
+#max_validation_temp: 60.
+#   The maximum temperature used to validate the calibration.  It is
+#   recommended to set this to a value between 100 and 120 for enclosed
+#   printers.  The default is 60.
 ```
 
 ## Temperature sensors
--- a/docs/Eddy_Probe.md
+++ b/docs/Eddy_Probe.md
@@ -78,7 +78,9 @@ for further details on how to configure a `temperature_probe`.  It is
 advised to configure the `calibration_position`,
 `calibration_extruder_temp`, `extruder_heating_z`, and
 `calibration_bed_temp` options, as doing so will automate some of the
-steps outlined below.
+steps outlined below.  If the printer to be calibrated is enclosed, it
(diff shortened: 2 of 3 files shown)


docs/Config_Reference.md
```diff
@@ -2458,6 +2458,10 @@ postfix for both sections.
 #   "calibration_extruder_temp" option is set.  Its recommended to heat
 #   the extruder some distance from the bed to minimize its impact on
 #   the probe coil temperature.  The default is 50.
+#max_valid_temp: 60.
+#   The maximum temperature used to validate the calibration.  It is
+#   recommended to set this to a value between 100 and 120 for enclosed
+#   printers.  The default is 60.
```
bevanweiss, line 2464: This seems a bit confusing...

> This patch introduces the max_valid_temp config option, which replaces the hard coded maximum validation temperature of 120C.

Should the default not then be 120degC, such that existing installations are NOT impacted by this change?

And why have a recommendation of between 100 and 120deg, but then a default of 60deg?
shouldn't the default fall within the recommended range?  (I get that enclosed vs open will be different... but you haven't made this clarification in your description, even though you've obviously based the default 60deg value on it)

I think the name should also be changed to read 'max_validation_temp'.
'max_valid_temp' sounds like if the readback temperature exceeds this value then it will be considered as 'invalid'

Arksine: Existing calibrations won't be affected by this change, they have already been validated up to 120C.  The recommendation of 100-120C is for **enclosed** printers.    Defaulting to 60 makes it less likely users will fail validation after calibration.

I'm ok with changing the name for clarity.  In addition, I can tweak the post-calibration validation procedure a bit to use the maximum value between the `max_validation_temp` and the maximum measured temp.  This ensures that the entire range of measured temperatures is validated.

KevinOConnor:
Thanks.  Let me know when you are ready to commit and I will do so.

-Kevin

Arksine:
Oops, this one slipped my mind.  It is good to merge whenever you are ready Kevin.  Thanks.

KevinOConnor:
Thanks.

-Kevin

PR #6672, opened 2024-08-30
Klipper lib CMSIS: bump version to 6.1.0

bevanweiss:
Update CMSIS to v6.1.0.

Tested with STM32F103 build.

### DIFF
--- a/lib/README
+++ b/lib/README
@@ -1,8 +1,8 @@
 This directory contains external library code.
 
 The cmsis directory contains code from:
-  https://github.com/ARM-software/CMSIS_5
-version 5.9.0 (2b7495b8535bdcb306dac29b9ded4cfb679d7e5c). Contents
+  https://github.com/ARM-software/CMSIS_6/tree/v6.1.0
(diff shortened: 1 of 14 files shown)


KevinOConnor:
We can certainly update these files, but I think we should have a use-case for it first.  That is, if we're adding some new chip that needs the later version then I think it's fine to update as part of that work.  I'm leery of making changes that could introduce a regression without some benefit to taking that risk.

Thanks,
-Kevin

nefelim4ag:
Would add a small excuse for 5.9.
Here is [device](https://github.com/markniu/bd_pressure/blob/main/hardware/bdpressureE/Schematic_Pressure_sensor_E3D_2026-01-03.pdf)  which utilizes STM32C011.
And it seems to me that nearest CMSIS https://github.com/STMicroelectronics/cmsis-device-c0
5.9 https://htmlpreview.github.io/?https://github.com/STMicroelectronics/STM32CubeC0/blob/main/Release_Notes.html


But I'm don't really know what I'm talking about.

Cheers,
-Timofey.

PR #6673, opened 2024-08-30
Klipper lib/stm32*: Update STM32 header references.

bevanweiss:
The README referenced paths in a github that were themselves git submodules.
So have altered README to point to the referenced git repo itself rather than the containing git repo.


KevinOConnor:
Alas, the STM provided code is of low-quality.  The code often has bugs, they don't publish meaningful changelogs, they frequently change the license statements (often making them more ambiguous), and they have a history of changing the repo locations.

As near as I can tell the current Klipper lib/README file is accurate (one can obtain the code from the given repos by checking out the specified commit).  Until that changes or some new functionality requires an update, I would avoid "chasing" STM's changes.

Thanks,
-Kevin

PR #6675, opened 2024-09-02
temperature_combined: Handle None in combined temp sensor values

Swamp-Ig:
Improves `temperature_combined` so it handles undefined values (by ignoring them).

This allows `temperature_combined` to use for example the temps reported by tmc2240 stepper drivers, currently if these are included there is a crash on boot up.

### DIFF
--- a/klippy/extras/temperature_combined.py
+++ b/klippy/extras/temperature_combined.py
@@ -70,7 +70,8 @@ class PrinterSensorCombined:
         for sensor in self.sensors:
             sensor_status = sensor.get_status(eventtime)
             sensor_temperature = sensor_status['temperature']
-            values.append(sensor_temperature)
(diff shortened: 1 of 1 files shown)


Swamp-Ig:
[klippy.log](https://github.com/user-attachments/files/16846044/klippy.log)

Here is the klippy.log that demonstrates the bug that is getting fixed.

nefelim4ag:
Honestly,
I think it is better to handle `None` here and return `raise self.printer.config_error()` with sensor name
Because otherwise, there is possibility to define combined sensor with all TMC2240 as example, which will not work until they are enabled, which is tolerable. But this allows to try to bind it to something else, like temperature fan - then there should be a special handling & etc.

I hope I was able to explain the idea and basic problem here with that.

PR #6676, opened 2024-09-02
Amend octoprint install instructions with `klippy.serial`

mlewis-everley:
The OctoPrint install instructions say 'add ~/printer_data/comms/klippy.sock' but it actually appears to be `~/printer_data/comms/klippy.serial`, so I have updated this.

I also seemed to have issues connecting without `/tmp/printer` in additional ports, so added more information on this.

### DIFF
--- a/docs/OctoPrint.md
+++ b/docs/OctoPrint.md
@@ -52,12 +52,12 @@ Navigate to the Settings tab (the wrench icon at the top of the page).
 Under "Serial Connection" in "Additional serial ports" add:
 
 ```
-~/printer_data/comms/klippy.sock
+~/printer_data/comms/klippy.serial
 ```
+
 Then click "Save".
 
-_In some older setups this address may be `/tmp/printer`_
-
+_In some older setups this address may be `/tmp/printer`, depending on your setup, you may need to keep this line as well_
 
 Enter the Settings tab again and under "Serial Connection" change the "Serial Port"
 setting to the one added above.


JamesH1978:
Ah yes, this would be my cock up :) i did indeed put the sock and not the serial.

Please can you sign off your PR, and i will review and accept it as per point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks
James

rongith:
I have an issue with ~/printer_data/comms/klippy.serial
~/printer_data/comms/klippy.serial is not an existing link. It should be linked to what user try to open that link. I didn't find a way to discover what user it is.
Then in my case I have to specify the full path: /home/pi/printer_data/comms/klippy.serial

On my system in /etc/systemd/system/octoprint.service, I have:
```
User=pi
ExecStart=/home/pi/OctoPrint/venv/bin/octoprint serve
```

JamesH1978:
fixed by https://github.com/Klipper3d/klipper/pull/6716

JamesH1978:
@rongith are you saying that octoprint itself does not recognize ~ as a placeholder for /home/<user>/

If so, i will make some amendments.

Thanks, and sorry for missing your reply

James

rongith:
@JamesH1978. No worries for the delay because I found a solution. It just took time (I wasn't efficient) to find it because first I was on a quite old version of klipper with the /tmp/printer and I goggled my issue instead of having a look at the documentation and dig a bit.
I thought it was a user issue but you are probably right that the tilde expansion is not recognize in Octoprint. But I can't find any reference about that. Here the consequence of the declaration of the path to the serial:
Screenshot 2024-10-22 174743
Screenshot 2024-10-22 174919
Screenshot 2024-10-22 174942
Screenshot 2024-10-22 175015

Here defaults fields for other paths in Octoprint:
Screenshot 2024-10-22 175538

rongith:
If the tilde is not supported, one cause may be that Octoprint is OS agnostic.

JamesH1978:
i will do some testing and confirm if the tilde is supported or not, and then make a change accordingly, maybe just state `/home/<your user>/printer_data` , it shouldnt matter for older /tmp/printer 

Thanks
James

rongith:
The tilde expansion works in Octoprint for Folders and for plugins like Octoklipper. But not for me for the serial ports. :thinking:
image
image
Moreover the Octoprint Folders section replace automatically ~ by /home/username when you save. There should be some tricks under the hood here.

PR #6677, opened 2024-09-03
tmc: Add `overvoltage_threshold` and `overtemperature_warning_threshold` for TMC2240

yunline:
This PR adds `overvoltage_threshold` and `overtemperature_warning_threshold` to the config of tmc2240 driver.

## overvoltage_threshold
Accroding to tmc2240 datasheet:
> A stepper motor application can generate significant overvoltage, especially when the motor becomes quickly decelerated from a high velocity, or when the motor stalls 
To protect the driver as well as connected circuitry, the TMC2240 has an overvoltage detection and protection
mechanism.
The OV output allows attaching an NPN or MOSFET with a power resistor (brake resistor) to dump the excess energy
into the resistor.
The transistor chops with approximately 3kHz to 4kHz (depending on the clock frequency) to keep the supply within the limits.

By setting the `overvoltage_threshold` in config, we can determine the the threshold of the overvoltage detection of tmc2240 to adapt different supply voltage.

## overtemperature_warning_threshold
This config sets the threshold of overtemperature warning. When the temperature of tmc2240 reaches this threshold, a overtemperature warning will be reported.

## Example config
```
[tmc2240 stepper_x]
...
overvoltage_threshold: 13.0 
# 13V overvoltage threshold for a 12V system.

overtemperature_warning_threshold: 100
# When temperature reaches 100℃, a warning will be reported.
...
```

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3718,6 +3718,16 @@ run_current:
 #   velocity" threshold (THIGH) to. This is typically used to disable
 #   the "CoolStep" feature at high speeds. The default is to not set a
 #   TMC "high velocity" threshold.
+#overvoltage_threshold: 37.735
+#   The threshold voltage (in V) of overvoltage protection. The decelerating
+#   stepper motor can cause a raise of V_supply of the driver. When the V_supply
+#   is higher than overvoltage_threshold, the mosfet is turned on and dump the
+#   excess energy into the power resistor, keep the supply voltage below
+#   the threshold.
+#overtemperature_warning_threshold: 120.0
+#   The threshold temperature (in ℃) of the overtemperature warning.
+#   When the temperature of the driver is higher than this threshold,
+#   a warning will be reported.
 #driver_MSLUT0: 2863314260
 #driver_MSLUT1: 1251300522
 #driver_MSLUT2: 608774441
--- a/klippy/extras/tmc2240.py
+++ b/klippy/extras/tmc2240.py
@@ -336,6 +336,34 @@ class TMC2240CurrentHelper:
         val = self.fields.set_field("irun", irun)
         self.mcu_tmc.set_register("IHOLD_IRUN", val, print_time)
 
+######################################################################
+# TMC stepper overvoltage config helper
+######################################################################
+
+def TMC2240OvervoltageHelper(config, mcu_tmc):
+    fields = mcu_tmc.get_fields()
+    ov_thres = config.getfloat(
+        'overvoltage_threshold',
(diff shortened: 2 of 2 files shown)


JamesH1978:
Thank you for submitting a PR, please provide a signed off by line as outlined in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

Thanks

James

yunline:
> Thank you for submitting a PR, please provide a signed off by line as outlined in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md

Updated

yunline:
I didn't find a significant quality improvement with `overvoltage_threshold` set to 13v (my power supply is 12v) or 30v.
That could because the printer is running at a low speed (200mm/s 10000mm/s^2), or because the design of circuit of the driver module is not good. Or maybe, changing `overvoltage_threshold` doesn't really matter to the print quality?

Anyway, before this PR, there is no way to access field `overtempprewarning_vth` and `overvoltage_vth`. This is the importances of this PR.

nefelim4ag:
The main purpose of OVERVOLTAGE_VTH is to enable the break resistor to avoid damage to components during deceleration.
On the current available BTT TMC2240, it is a tiny 1W 2k Ohm resistor.

So, the only situation where it can be needed or helpful is if we have 36V power, and it is easy to get 36V+ output on the motor leads.
So, I'm not sure it makes sense to adjust it in any other situation.

It is possible to query the current supply voltage, add an offset of 0.5V, and probably set the threshold to this value.
Not sure if it makes sense.

It is possible to set this field with SET_TMC_FIELD if necessary.
It seems to me that the only thing we can/should fix is to reset it to default values upon initialization.
So, I guess, this should be enough:
```
diff --git a/klippy/extras/tmc2240.py b/klippy/extras/tmc2240.py
index 2d154684a..7320e184a 100644
--- a/klippy/extras/tmc2240.py
+++ b/klippy/extras/tmc2240.py
@@ -417,6 +417,9 @@ class TMC2240:
         set_config_field(config, "sg4_angle_offset", 1)
         #   DRV_CONF
         set_config_field(config, "slope_control", 0)
+        #   OTW_OV_VTH
+        set_config_field(config, "overvoltage_vth", 0x0f25)
+        set_config_field(config, "overtempprewarning_vth", 0x0b92)
 
 def load_config_prefix(config):
     return TMC2240(config)
```

So, a knowledgeable enough person can configure it (it is not so hard), and they are at default state in any other circumstances.

Regards,
-Timofey

PR #6678, opened 2024-09-04
Fix hx71x on python2

KevinOConnor:
This minor code change should fix the hx71x support on python2 installations.  It also adds a load_cell regression test case.

@garethky , @alchemyEngine - fyi.

-Kevin


garethky (approved):
Hmm, I'll have to see what else I might have used this pattern on. Wish Python2 would go bye-bye.

Neko-vecter:
Python 2 support?? When will we fully migrate to python3?

-Neko.vecter

PR #6684, opened 2024-09-10
SX1509: drop i2c modify bits

nefelim4ag:
i2c_modify_bits is tricky and used only in one place where it actually seems overcomplicated.
We do not modify registers in other places, they are reset before setting new bits.
So they can be simplified to i2c_write.

Datasheet: https://cdn.sparkfun.com/datasheets/BreakoutBoards/sx1509.pdf

Thanks.

---
#6674

### DIFF
--- a/klippy/extras/bus.py
+++ b/klippy/extras/bus.py
@@ -160,7 +160,7 @@ class MCU_I2C:
                 % (self.oid, speed, addr))
         self.cmd_queue = self.mcu.alloc_command_queue()
         self.mcu.register_config_callback(self.build_config)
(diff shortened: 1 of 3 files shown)


KevinOConnor:
Thanks.  Seems useful to me.

I believe the sx1509 was only used on the duet2 with the duex add-on board.  It would be great if someone could confirm everything still works after this change.  I'm not sure how many users there are now for that board.  Maybe @Rext3d would know?

-Kevin

ReXT3D:
Unfortunately I am unable to comment.  I never used the sx1509 and both my Duet2 cards are currently not in active use.

KevinOConnor:
Thanks.

-Kevin

PR #6685, opened 2024-09-10
Update generic-creality-v4.2.7.cfg

louiskatz:
Corrected pinoutss for the 4 stepper motor on the 4.2.7 board. The correct ones can also be found on the klipper/config/printer-creality-ender3-v2-neo-2022.cfg

I am using this on my v4.2.7 board . 
The step_pin and dir_pin are are inverted in the original. 
Thanks to whoever looks these over. 
Louis

### DIFF
--- a/config/generic-creality-v4.2.7.cfg
+++ b/config/generic-creality-v4.2.7.cfg
@@ -16,8 +16,8 @@
 # See docs/Config_Reference.md for a description of parameters.
 
 [stepper_x]
-step_pin: PB9
-dir_pin: PC2
+step_pin: PC2
+dir_pin: PB9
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thankyou for submitting a PR, but the pins on a 4.2.7 are indeed different/swapped compared to a 4.2.2. This has always been the way, unless you have the "freeruns" version which does follow the 4.2.2 pins, but also has controlled TMC drivers.

Thanks
James

JamesH1978:
Also the mentioned config for the v2 neo also follows the 4.2.2 pin map and not the 4.2.7 one.

Thanks
James

PR #6686, opened 2024-09-11
Update Bed_Level.html

DragonSlayerSE:
changed the word "level" to layer on row 1584
(my first github contribution)


JamesH1978:
Thankyou for submitting a PR, just to let you know the html is created from the actual docs, so you would actually edit https://github.com/Klipper3d/klipper/blob/master/docs/Bed_Level.md and that change will be auto built into the web based documentation. Please also be aware that you need to sign off a PR, please see point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks
James

PR #6688, opened 2024-09-13
Queue and collate fan requests (fixed PR)

KevinOConnor:
PR #6670 had an error causing timing problems with heater_fan, temperature_fan, and controller_fan modules.  The fan queuing changes of that PR were reverted in commit 900bf2be.  

This PR adds fan queuing for g-code commands while preserving the previous behaviour for asynchronous updates from the heater_fan, temperature_fan, and controller_fan modules.  This should, hopefully, fix the issues with the previous PR.

-Kevin


KevinOConnor:
I've reworked the code a little to better improve timing when multiple redundant requests are issued.

-Kevin

PR #6690, opened 2024-09-15
flash_usb: Wait for busnum file to exist when flashing with picoboot

jackwakefield:
This solves an issue where the USB directory could exist, but the busnum file itself may not exist immediately. This was encountered when flashing a Pico connected to a Raspberry Pi 5.


KevinOConnor:
Thanks.

-Kevin

PR #6691, opened 2024-09-16
Add new fan_multi module that can have tachometer pins 

xPatrikTwitch:
Adds a new "fan_multi" module that can accept multiple tachometer pins and then reports their rpm in an array (Used with a board that drives multiple fans from one mosfet but each fan has its own tachometer pin connected to gpio)

image

Signed-off-by: xPatrikTwitch <xpatriktwitch@gmail.com>

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3009,6 +3009,29 @@ with the SET_FAN_SPEED [gcode command](G-Codes.md#fan_generic).
 #   See the "fan" section for a description of the above parameters.
 ```
 
+### [fan_multi]
+
+Manually controlled fan that can have multiple tachometer pins (one may define any number of sections with a
+"fan_multi" prefix). The speed of a manually controlled fan is set
+with the SET_FAN_SPEED [gcode command](G-Codes.md#fan_generic).
+
+```
+[fan_multi other_fans]
+#pin:
+#max_power:
+#shutdown_speed:
+#cycle_time:
+#hardware_pwm:
+#kick_start_time:
+#off_below:
+#tachometer_pins:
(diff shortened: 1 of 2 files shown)


bevanweiss:
Is there a way this could be done with the core 'fan control' provided by inheriting from the existing Fan, or perhaps even with just modifications to Fan?
Otherwise this adds quite a lot of copy-paste, which when making changes to Fan will require others to keep them in sync all the time.

You'll also need to fix the signoff to have your correct name.

xPatrikTwitch:
> Is there a way this could be done with the core 'fan control' provided by inheriting from the existing Fan, or perhaps even with just modifications to Fan? Otherwise this adds quite a lot of copy-paste, which when making changes to Fan will require others to keep them in sync all the time.
> 
> You'll also need to fix the signoff to have your correct name.

yeah i agree that this is 90% from the normal fan.py but if all wans returned their rpm in a array it would break how mainsail/fluidd currently display the rpm and both would need to be updated for this change

bevanweiss:
> > Is there a way this could be done with the core 'fan control' provided by inheriting from the existing Fan, or perhaps even with just modifications to Fan? Otherwise this adds quite a lot of copy-paste, which when making changes to Fan will require others to keep them in sync all the time.
> > You'll also need to fix the signoff to have your correct name.
> 
> yeah i agree that this is 90% from the normal fan.py but if all wans returned their rpm in a array it would break how mainsail/fluidd currently display the rpm and both would need to be updated for this change

Couldn't it return the standard value (as it currently does) where there's a single fan tacho?
Only when there are multiple fan tachos would it then return an array.

And couldn't you still implement this as an inherited class from the `fan`, so that you're not increasing the maintenance burden so much?

PR #6692, opened 2024-09-17
rp2040/i2c.c: Check for NACK/Start NACK

nefelim4ag:
For now, RP2040 silently ignores any kind of errors, and we only try to catch them by timeout condition.
I did some tests, write can't be timeouted actually, at least with a disconnected device.

So, these are specific checks for NACK.
I did reproduce Start NACK/Start Read NACK by disconnecting SDA line from the device.
Write NACK has not been tested, I'm a little puzzled about how to reproduce write NACK.

Thanks.


KevinOConnor:
Thanks.

-Kevin

PR #6695, opened 2024-09-21
Add support for output_pin and fan templates

KevinOConnor:
This PR adds support for specifying a math formula for output_pin and fan_generic objects.  This works similarly to the "led template" support - one can specify a `[display_template my_template]` config section that contains a Jinja2 macro providing a formula, and that formula will then be evaluated periodically to determine the pin/fan setting to apply.

The settings from the formula do not pass through the g-code command queue, so the updates are applied even during homing and other long running commands.

This should make it possible to define more flexible pin and fan control schemes.

-Kevin

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -476,6 +476,20 @@ enabled.
 `SET_FAN_SPEED FAN=config_name SPEED=<speed>` This command sets the
 speed of a fan. "speed" must be between 0.0 and 1.0.
 
+`SET_FAN_SPEED PIN=config_name TEMPLATE=<template_name>
+[<param_x>=<literal>]`: If `TEMPLATE` is specified then it assigns a
+[display_template](Config_Reference.md#display_template) to the given
+fan. For example, if one defined a `[display_template
+my_fan_template]` config section then one could assign
+`TEMPLATE=my_fan_template` here. The display_template should produce a
+string containing a floating point number with the desired value. The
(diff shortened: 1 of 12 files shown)


docs/G-Codes.md
```diff
@@ -476,6 +476,20 @@ enabled.
 `SET_FAN_SPEED FAN=config_name SPEED=<speed>` This command sets the
 speed of a fan. "speed" must be between 0.0 and 1.0.
 
+`SET_FAN_SPEED PIN=config_name TEMPLATE=<template_name>
+[<param_x>=<literal>]`: If `TEMPLATE` is specified then it assigns a
+[display_template](Config_Reference.md#display_template) to the given
```
bevanweiss, line 481: Perhaps `display_template` could get a secondary name of just `template`?
It feels very confusing to be connecting an output pin / fan to a 'display_template' when the intent is to not display anything, but to have it all work in the background...

Maybe `dynamic_variable` or `dynamic_value` or something else might be a good secondary name.  Just something that doesn't tie it specifically to a 'display' usage.

KevinOConnor: Thanks for reviewing.

I agree that display_template is a confusing name.  Roughly I'd like to introduce something like `macro_template`, make `render()` always available from Jinja2, and deprecate `display_template`.  That change seems orthogonal though, so didn't want to complicate this PR with a config rename.

-Kevin

Issue #6641, opened 2024-07-11
G4 add seconds


mikemerryguy:
https://github.com/Klipper3d/klipper/blob/248d3dbf8bcda69a51e3a303862bbcc64a32119d/klippy/toolhead.py#L611C3-L614C26

def cmd_G4(self, params):
        # Dwell
        if 'S' in params:
            delay = self.get_float('S', params, minval=0.)
        else:
            delay = self.get_float('P', params, 0., minval=0.) / 1000.
        self.toolhead.dwell(delay)

jakep82:
We don't use GitHub to report or track issues.

https://www.klipper3d.org/Contact.html#i-found-a-bug-in-the-klipper-software

PR #6696, opened 2024-09-25
Add support of S parameter to Dwell command (G4)

EmJay276:
Add support of G4 S< seconds >

Inspired by https://github.com/Klipper3d/klipper/issues/6641

Signed-off-by: Michael Jäger michael@mjaeger.eu

### DIFF
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -7,7 +7,7 @@ commands that one may enter into the OctoPrint terminal tab.
 
 Klipper supports the following standard G-Code commands:
 - Move (G0 or G1): `G1 [X<pos>] [Y<pos>] [Z<pos>] [E<pos>] [F<speed>]`
-- Dwell: `G4 P<milliseconds>`
+- Dwell: `G4 P<milliseconds>` or  `G4 S<seconds>`
 - Move to origin: `G28 [X] [Y] [Z]`
 - Turn off motors: `M18` or `M84`
 - Wait for current moves to finish: `M400`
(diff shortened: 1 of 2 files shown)


Zeanon:
idea:
Why not add S and P up, that way you can do
G4 S10 P5 for 10.5 seconds for example

```
        delay = gcmd.get_float("P", 0.0, minval=0.0) / 1000.0
        delay += gcmd.get_float('S', 0., minval=0.)
```

EmJay276:
Sounds good, I did not think about this idea. 

But 
G4 S10 P5 would be 10.005 sec ;)

Sineos:
I'd think this would be better solved with either:
* A standard Klipper macro
* The upcoming "plugin" system --> https://klipper.discourse.group/t/new-proposal-for-klipper-extension-support/19150

EmJay276:
> I'd think this would be better solved with either:

This is standard G-Code, why not support it?

Sineos:
> This is standard G-Code, why not support it?

Because there is no real Gcode standard and about every firmware has a slightly different interpretation. Just because Marlin is doing X, this does not mean X is a standard.

Klipper's overall philosophy so far has been to keep the supported Gcodes as short as possible. This means not extending Gcodes unless there is a compelling need or no other solution.

As there is one existing option to extend G4 via a standard macro, another option is coming up and the whole G4 topic is far from changing the world, I see no real reason to do it. On the other hand, the code change seems pretty trivial without much risk of a regression. 
Finally it will be Kevin's call.

Arksine:
FWIW, Klipper previously had support for the S parameter and it was removed.  The commit message indicates that it is non-standard.

https://github.com/Klipper3d/klipper/commit/8e87ab0939eb2ac12b5d41e590da1494596577e3

PR #6699, opened 2024-09-29
Update ender 3 v2 re: GD32F303 variants.

berg-michael:
Ender started shipping GD32F303 based mainboards sometime in 2022. These require "Disable SWD at startup" to be toggled. This PR adds a note regarding this setting. The note is lifted from the ender3pro-2020 config file, but I changed GD32F**1**03 to GD32F**3**03 as I believe this is the chip used in these models.

### DIFF
--- a/config/printer-creality-ender3-v2-2020.cfg
+++ b/config/printer-creality-ender3-v2-2020.cfg
@@ -3,6 +3,12 @@
 # STM32F103 with a "28KiB bootloader" and serial (on USART1 PA10/PA9)
 # communication.
 
+# It should be noted that newer variations of this printer shipping in
+# 2022 may have GD32F303 chips installed and not STM32F103. You may
+# have to inspect the mainboard to ascertain which one you have. If it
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thanks for submitting a PR, generally the GD32F303 variants do not need that setting, only the GD32F103 ones. 

It seems to be a weird quirk that the GD32F303 is a better clone of the STM32F103 than the similar numbered GD32F103.

This is why these configs have not been changed to say otherwise.

Thanks
James

jimbovia:
So it took finding @JamesH1978 's pinned post about the GD32F303 before I unchecked the "Disable SWD at startup" and was eventually successful in flashing Klipper to a GD32F303 board that came in an Ender 3 Pro (not a v2).  What are the chances that someone can add a line or two clarifying the differences between the 'make menuconfig' process settings for the GD32F103 and the GD32F303 chipset 4.2.2 boards in both the Ender3v2 and Ender3Pro starting/example printer.cfg's in the Klipper repository?

PR #6701, opened 2024-10-02
docs: Fix typo - XH711 instead of HX711

kabroxiko:
Fix documentation typo


KevinOConnor:
Thanks.  Commits need to have a signed-off-by line for me to commit directly (see https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md ).  If that is an issue, I'll make the corresponding fix using my sign off.

-Kevin

KevinOConnor:
I fixed this in commit 55339998.

Thanks,
-Kevin

PR #6702, opened 2024-10-02
Extend documentation of klippy:connect and klippy:read event handlers.

rschaeuble:
Describe more clearly what is and what isn't allowed in the klippy:connect and klippy:ready event handlers.


rschaeuble:
I'm not sure how I should self-review a documentation change.

@KevinOConnor: could you have a quick look, as we were having the discussion about these event handlers in another PR recently?

KevinOConnor:
Sorry - this "slipped" on my side.

We can certainly change the documentation, but for what it is worth I find the paragraph flow of this change a little odd.  The "reactor" is first described a few bullets down in that list - so referring to the reactor before describing it seems a little odd.

Maybe just change the "ready" event description to add something like: "This event is only intended as notification of a state change - do not raise an error in this callback, send commands to a micro-controller, nor perform any long running operations."  Not sure.

Thanks for working on this,
-Kevin

PR #6703, opened 2024-10-04
Add support for HSE bypass on STM32.

liampwll:


src/stm32/stm32g4.c
```diff
@@ -76,8 +76,12 @@ gpio_clock_enable(GPIO_TypeDef *regs)
...
 }
 
-#if !CONFIG_STM32_CLOCK_REF_INTERNAL
-DECL_CONSTANT_STR("RESERVE_PINS_crystal", "PF0,PF1");
+#if CONFIG_STM32_CLOCK_REF_INTERNAL
+// No pins required.
+#elif CONFIG_STM32_CLOCK_HSE_BYPASS
+DECL_CONSTANT_STR("RESERVE_PINS_hse_clock", "PF0");
+#else // CONFIG_STM32_CLOCK_HSE_CRYSTAL
+DECL_CONSTANT_STR("RESERVE_PINS_hse_clock", "PF0,PF1");
```
liampwll, line 84: This is a bit ugly, but it works for now.

KevinOConnor:
I'm not sure what this does, or why one would want it.  Can you elaborate?

Thanks,
-Kevin

liampwll:
HSE bypass mode only uses a single pin as an input for a clock signal
rather than connecting the two pins to a crystal driver. It's required when
using a clock generation IC rather than a crystal.

KevinOConnor:
Okay, thanks.  How about we change the menu so that it lists: 8Mhz crystal, 12Mhz crystal, ..., 12Mhz external clock, ..., Internal clock.  That is, add options for external clocks to the menu instead of adding a new "HSE bypass" config option (which I suspect may confuse many people).

Cheers,
-Kevin

liampwll:
Doing so would likely involve breaking existing configs as we would likely want to change `STM32_CLOCK_REF_8M` to `STM32_CLOCK_REF_CRYSTAL_8M` etc..

I'm not sure if it makes sense to do that given that the HSE bypass options is hidden away in the low level options.

PR #6704, opened 2024-10-04
Add support for USART3 on PC11/PC10 on STM32G474.

liampwll:
This can likely be extended to other devices, I just don't want to go check all the datasheets.


KevinOConnor:
Thanks.

-Kevin

PR #6705, opened 2024-10-05
config: Lulzbot TAZ5 config, and TAZ6 and RAMBo config improvements

eraserhd:
This adds a new config for TAZ5 with a DualExtruder v3.

The TAZ6 has mostly the same guts (same mainboard for sure), so I started from that.  In the process I made some fixes to the TAZ6 configs:

* Added missing chassis fans for the TAZ6 configs.
* Added a commented out chassis fan for the generic RAMBo config.
* Added T0 and T1 macros for the TAZ6 dual extruder.

The chassis fan was verified from the [RAMBo schematic](https://github.com/ultimachine/RAMBo-1.4/blob/1.4/Project%20Outputs/Schematic%20Prints_RAMBo_1.4a.PDF) and the [wiring instructions for the TAZ6](https://download.lulzbot.com/TAZ/6.0/production_docs/OHAI/03_ControlBoxWiring-OHAI-T6.pdf).

The TAZ5 config is tested and works on a real machine.  The machine is stock in all the ways that matter for this config.

Differences TAZ5 -> TAZ6 differences were largely about homing.  These I things changed:

* Two endstop pins were different.
* `[bed_screws]` section
* Homing location and some axis extents.
* Reworked START_PRINT and END_PRINT from Lulzbot's Cura profiles.

The stepper digipot is reconfigured, and I think it's more accurate, but since I don't have enough knowledge to fully confirm it, I didn't change TAZ6 or RAMBo.  If I'm right, the TAZ6 and RAMBo configs are underpowering their steppers.

### DIFF
--- a/config/generic-rambo.cfg
+++ b/config/generic-rambo.cfg
@@ -78,6 +78,10 @@ pin: PH5
 #[heater_fan heatbreak_cooling_fan]
 #pin: PH3
 
+#[controller_fan controller_fan]
+#RAMBo labels this "FAN-2", and it is often used as the chassis fan.
+#pin: PE4
+
 [mcu]
 serial: /dev/ttyACM0
 
--- a/config/printer-lulzbot-taz5-dual-v3-2015.cfg
+++ b/config/printer-lulzbot-taz5-dual-v3-2015.cfg
@@ -1,0 +1,259 @@
+#This file contains pin mappings for the Lulzbot TAZ 5 circa 2015 using RAMBo and Dual v3 toolhead.
+#To use this config, the firmware should be compiled for the AVR atmega2560.
+
+# See docs/Config_Reference.md for a description of parameters.
+
+#-------------------------------------------------------------------------------------------------
+#-------------------------------------------------------------------------------------------------
+#-------------------------------------------------------------------------------------------------
+# LULZBOT TAZ5 (RAMBo) with Dual v3 Extruder Master Config
+#-------------------------------------------------------------------------------------------------
+#-------------------------------------------------------------------------------------------------
+#-------------------------------------------------------------------------------------------------
+#Notes:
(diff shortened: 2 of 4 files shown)


JamesH1978:
Thank you for submitting a PR, as you have added a whole new config file, you will need to also submit for the regression tests as per point 4 in the guidelines in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md

Thanks
James

eraserhd:
Updated, thanks!

JamesH1978:
Thanks, generally it looks fine to me, although i am not sure why you have set a min_extrude_temp to 120 in both extruder configs, unless there is a good reason to set 120. We generally do not override default safety behaviour.

Also can you confirm the taz5 was released with the dual extruder option out of the box, ie you could purchase at taz5 with this already fitted and not just an after market option? Our configs are generally only for printers that someone would buy off the shelf in default configuration.

Thanks
James

eraserhd:
Re: `min_extrude_temp`, that was just copied from the TAZ6 config and not otherwise touched.  I'll remove it (from the TAZ6, too?)

Re: Dual extruder, it wasn't sold that way... but I'm just thinking about it and I didn't realize it was sold on the TAZ6.  I figured it was a common upgrade.

I'll make up a stock TAZ5.  My original tool head works enough to verify.

JamesH1978:
Apologies It has taken a while for me to get back to this, looks fine except for the macros in your new file for start print and end print. Usually we don't include macros that aren't explicitly needed. @KevinOConnor what are your thoughts on the T0 and T1 macros, added though?

Thanks
James

KevinOConnor:
I don't have a strong preference.  If it's a "printer" config file and the macros are relevant to successfully printing on that particular printer than I can see value in adding them to the example configs.

Cheers,
-Kevin

eraserhd:
bump?

PR #6708, opened 2024-10-10
Increase speeds of same70 by using Tightly Coupled Memory and DCache

Wulfsta:
Currently, this MCU is _slow_. Much slower than a 300MHz chip should be. This is due to the bus speeds being half the core clock, and the high number of wait states required to retrieve instructions from flash. This MCU can avoid these wait states by locating part of the code in Instruction Tightly Coupled Memory (ITCM). Klipper is currently less than one quarter of the size of the ITCM when enabled (USB build, slightly larger for USBCANBUS). These commits do the following:
1. If Tightly Coupled Memory (TCM) is not configured, sets it.
2. Relocates all of Klipper into ITCM after boot via Direct Memory Access (DMA), then enables ITCM.
    a. This avoids large changes, especially to the linker script.
    b. This is possible because ARM Cortex M7 MCUs remap access to the ITCM region to flash memory if ITCM is not enabled, but DMA hardware can access TCM regions if configured.
4. Enables the Data Cache.
    a. This is easier than trying to move everything to Data Tightly Coupled Memory (DTCM), since Bossa cannot set the GPNVM bits required to enable TCM.
    b. This is about as fast as DTCM anyways.
5. Locates CAN messages in DTCM.
    a. This is a simple way to avoid the messages getting cached and causing issues with a nice side effect of being slightly faster.
6. Adds necessary barriers in the USB code.
7. Adds benchmarks to the documentation.

To quantify how slow this MCU currently is, I ran a three stepper benchmark. Current mainline is 728 ticks, 1236K steps. This commit brings this to 190 ticks, 4737K steps - possibly making it the second fastest MCU in Klipper?

### DIFF
--- a/src/atsam/Makefile
+++ b/src/atsam/Makefile
@@ -20,9 +20,10 @@ CFLAGS-$(CONFIG_MACH_SAM4E) += -Ilib/sam4e/include
 CFLAGS-$(CONFIG_MACH_SAME70) += -Ilib/same70b/include
 CFLAGS += $(CFLAGS-y) -D__$(MCU)__ -mthumb -Ilib/cmsis-core -Ilib/fast-hash
 
-CFLAGS_klipper.elf += -nostdlib -lgcc -lc_nano
-CFLAGS_klipper.elf += -T $(OUT)src/generic/armcm_link.ld
-$(OUT)klipper.elf: $(OUT)src/generic/armcm_link.ld
+samlink-y := $(OUT)src/generic/armcm_link.ld
+samlink-$(CONFIG_MACH_SAME70) := $(OUT)src/atsam/same70_link.ld
+CFLAGS_klipper.elf += -nostdlib -lgcc -lc_nano -T $(samlink-y)
+$(OUT)klipper.elf: $(samlink-y)
 
 # Add source files
 src-y += atsam/main.c atsam/gpio.c atsam/i2c.c atsam/spi.c
--- a/docs/Benchmarks.md
+++ b/docs/Benchmarks.md
@@ -354,6 +354,26 @@ micro-controller.
 | 1 stepper (200Mhz)   | 39    |
 | 3 stepper (200Mhz)   | 181   |
 
+### SAME70 step rate benchmark
+
+The following configuration sequence is used on the SAME70:
+```
+allocate_oids count=3
+config_stepper oid=0 step_pin=PC18 dir_pin=PB5 invert_step=-1 step_pulse_ticks=0
+config_stepper oid=1 step_pin=PC16 dir_pin=PD10 invert_step=-1 step_pulse_ticks=0
+config_stepper oid=2 step_pin=PC28 dir_pin=PA4 invert_step=-1 step_pulse_ticks=0
+finalize_config crc=0
+```
+
+The test was last run on commit `e3f1140d` with gcc version
+`arm-none-eabi-gcc (NixOS 10.3-2021.10) 10.3.1` on a SAME70Q20B
+micro-controller.
+
+| same70               | ticks |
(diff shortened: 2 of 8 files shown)


KevinOConnor:
Interesting.  Thanks.

@eamaclean - fyi.

Some comments/questions:

* I don't think src/generic/armcm_link.lds.S should be changed.  Instead, copy that file to a new src/atsam/same70_link.lds.S, modify it as needed, and change the makefile to use that for the same70.  See src/rp2040/ for an example.
* I don't understand the changes to sam3_usb.c - the `usb_write_packet()` code writes to device memory which shouldn't be cached.  If it is cached, then wouldn't `usb_read_packet()` need changes as well?
* Out of curiosity, did you test what the results are if one enables the ICache (`SCB_EnableDCache()`), Dcache, but doesn't use ITCM?  (That is what the stm32h7 chip does.)

Cheers,
-Kevin

Wulfsta:
@eamaclean knows, I have been annoying him with progress on this for a few weeks. 😆

In order:
1. I opted to add this to the generic linker because many M7 chips like the H7 have TCM that behaves this way; I thought they might be able to benefit from this pattern as well. Notice the ratio of the steps between this commit and the H7 is not the same as the ratio of their clock speeds, possibly an indicator that this is better than ICache?
2. I have no idea either; I arrived at this by trial and error. It worked when built with `O0`, so I added barriers everywhere in the USB driver and bisected until it worked. Sort of brute forced this one because I couldn’t spot an issue.
3. I did not do a formal benchmark, but have anecdotal evidence to believe this is better - I tried tinkering with the microsteps in my config with both this change and one only using ICache, and I was able to get a higher microstepping value with ITCM. I think this is due to the size of the ICache being half the size of Klipper’s `.text` section, leading to cache misses and flash reads. I do wonder if benchmarks would be similar, but I don’t really want to setup and run another test. 😅

Wulfsta:
Updated based on conversation offline - the USB code now uses `__DMB()`, and the change has been split into its own commit.

Wulfsta:
Updated again, moved the linker script to the `atsam` directory and stripped out a Kconfig directive.

src/atsam/Makefile
```diff
@@ -21,8 +21,14 @@ CFLAGS-$(CONFIG_MACH_SAME70) += -Ilib/same70b/include
...
-CFLAGS_klipper.elf += -T $(OUT)src/generic/armcm_link.ld
-$(OUT)klipper.elf: $(OUT)src/generic/armcm_link.ld
+
+ifeq ($(CONFIG_MACH_SAME70),y)
+    CFLAGS_klipper.elf += -T $(OUT)src/atsam/same70_link.ld
+    $(OUT)klipper.elf: $(OUT)src/atsam/same70_link.ld
+else
+    CFLAGS_klipper.elf += -T $(OUT)src/generic/armcm_link.ld
+    $(OUT)klipper.elf: $(OUT)src/generic/armcm_link.ld
+endif
```
KevinOConnor, line 31: I'm not sure that does the right thing.  (Makefile syntax is cryptic.)

How about:
```
samlink-y := $(OUT)src/generic/armcm_link.ld
samlink-$(CONFIG_MACH_SAME70) := $(OUT)src/atsam/same70_link.ld
CFLAGS_klipper.elf += -nostdlib -lgcc -lc_nano -T $(samlink-y)
$(OUT)klipper.elf: $(samlink-y)
```

-Kevin

KevinOConnor:
Thanks.  In general, it looks fine to me.  If there are no further comments, I'll look to commit in a few days.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6713, opened 2024-10-19
Add input_mux and vref options to ADS1220 sensor

garethky:
* fix type comparison bug that stopped the sensor from initializing (no one could actually use it)
* correct mismatch between docs and code for `sample_rate` (fixed to work same as hx71x)
* add `input_mux`, `pga_bypass` and `vref` options
* update configuration reference & fix typos


KevinOConnor:
Thanks.

-Kevin

PR #6714, opened 2024-10-21
Add support for named/multiple probes and update bed_mesh to support …

cekim-git:
Add support for named/multiple probes via [probe NAME] directive.
- backwards compatible with default [probe]
- update bed_mesh.py to support named probes via probe: configuration directive
- quad_gantry_level already supports passing this configuration item
- Add PROBE= argument to macros (PROBE_ACCURACY PROBE=other_probe)
- Internally lookup_object(self.probe_name) can be used to get the probe


JamesH1978:

Thanks
James

cekim-git:
My apologies... I have just amended the commit.  Hopefully, that satisfies this requirement?  If not, let me know.

cekim-git:
I craf5,poo

LittleReckliss:
Does this allow an eddy sensor and a bltouch to run at the same time?

cekim-git:
> Does this allow an eddy sensor and a bltouch to run at the same time?

Long answer:

In my case I am running both a TAP sensor (aka: 'probe') and a superPINDA (aka: 'pinda0') attached to Toolhead0 (the first of 6 heads attached via CAN/EBB36 pins PB9 and PB6 respectively).  quad_gantry_level worked without edits by just adding a "probe: pinda0" to its configuration in printer.cfg.  I can freely switch back and forth between them.  BedMesh required the attached edits to function the same (switch freely between the 2 probes).  The edits to bed_mesh.py were to replace the lookup for the "probe" object from using the string 'probe' and instead use a configuration value (aka: self.probe_name) retrieved from the printer.cfg values.  I have also modified my macros to allow me to freely switch between TAP and PINDA as the primary Z-home values in [homing_override].  I simultaneously rely on the TAP sensor for detecting that the head has moved in Z relative to the gantry (an indication of multi-headed printer malfunction or tool-change) while relying on the PINDA for homing, QGL and meshing... 

Without having tested bltouch/eddy I can't say for sure what they would require.  Looking at the code for eddy - there is a reliance on the default "probe" object:
        self.cmd_helper = probe.ProbeCommandHelper(
            config, self, self.mcu_probe.query_endstop)
        self.probe_offsets = probe.ProbeOffsetsHelper(config)
        self.probe_session = probe.ProbeSessionHelper(config, self.mcu_probe)

I would expect there may need to be re-factoring here to request the new named probe rather than rely on the default probe object.  Doing so would potentially allow decoupling as I have done with TAP/PINDA case and thus concurrent operation.

But, I would view this extension/PR as a first step in adding extensibility to the probe definition which then requires other objects wishing to de-couple from the hard-coded default "probe" object to make relatively simple edits to do so.

Short Answer: Potentially...  Facilitating this is the precise intention of this modification, but as of yet changes and testing have not taken place with that combination.  May require some edits to eddy/bltouch to remove reliance on default "probe" object often referenced directly rather than fetched with "lookup_object()"

LittleReckliss:
I appreciate the effort you've put into this PR. I've never edited Klipper directly myself but I'm sure you have a decent understanding given you made this. 

I was wondering about the structure for declaring multiple probes. Would it be feasible to add another line to the probe declaration, like this?

`[probe <probe_type (e.g., "eddy_current", "microswitch", "bltouch")> <probe_name>]`

This approach could allow users to:

1. Declare the probe type: This way, you could include specific configuration options based on the type (e.g., Eddy needing a 
sensor type, BLTouch needing a control pin).
2. Assign a unique name to each probe: This would help users manage multiple probes more clearly, making it easier to call a specific probe for different operations.

I faced an issue with declaring multiple probes and trying to use z_virtual_endstop for Z-homing when there were multiple “probe” boards involved. I realized that it might just require changing the board assignment to match the specific probe name. Still, I think the added declaration line could make configuration more intuitive for users, if possible.

cekim-git:
> Would it be feasible to add another line to the probe declaration, like this?
> `[probe <probe_type (e.g., "eddy_current", "microswitch", "bltouch")> <probe_name>]`
Given the way klipper works, this might not be necessary to accomplish your goal with named probes added to the base probe class.  As for feasibility, I just don't know if provision for more than a "prefix" (second argument) exists...  (you now have 3 arguments).  However, a configuration might have a "type:" value for conditional functionality like this for example, but also.... there is the inherent function of class definition here.

In klipper there's a base presumption that [<class_name>] will, if not already defined, try to open a file in extras/ named <class_name>.py and find it dynamically. Such a file presumably contained a python object (loaded via the "load_config()" declaration in that file).  That file may then either create a entirely new "class" or it may inherit an existing class (as is the case with eddy/bltouch inheriting "probe").

Further [<class_name> <prefix>] can be supported by the new class.  So, Something like Eddy or Bltouch could themselves support this prefix and multiple named instances if they so choose having inherited the enhanced base probe class.
 
> This approach could allow users to:
> 1. Declare the probe type: This way, you could include specific configuration options based on the type (e.g., Eddy needing a
>    sensor type, BLTouch needing a control pin).
See above, the nature of [<class name>} loading <class_name>.py provides for this already, I think?  Either via the derived class searching for differing configuration arguments period or relying on a "type" configuration value to determine which arguments are valid.

> 2. Assign a unique name to each probe: This would help users manage multiple probes more clearly, making it easier to call a specific probe for different operations.
This is the exact intention of the base class (probe.py) supporting named probes.  From this point forward, new classes and configurations can now reach these multiple probe objects by name.  Either in the .cfg files or in the python.  Whether a named probe is a child/derived class of probe.py (bltouch/eddy for example) or a simple pin/stop it would be reachable via name from this point forward.

In short, it is unclear to me if the klipper infrastructure supports arguments beyond 2... I just don't know, but it may not be required to resolve the issue you are describing IF named probes are supported.  Both the differing configuration items (supported already by derived class objects corresponding to the first value in [<class> <prefix>]) and the ability to reference by name to "call a specific probe for different options" are addressed here, I think.

klippy/extras/probe.py
```diff
@@ -40,26 +40,50 @@ def __init__(self, config, probe, query_endstop=None):
...
+        probe_cmd                = 'PROBE'
+        probe_calibrate_cmd      = 'PROBE_CALIBRATE'
+        probe_accuracy_cmd       = 'PROBE_ACCURACY'
+        z_offset_apply_probe_cmd = 'Z_OFFSET_APPLY_PROBE'
+        if self.probe_name != 'probe':
+            query_probe_cmd          += '_' + self.probe_name.upper()
+            probe_cmd                += '_' + self.probe_name.upper()
+            probe_calibrate_cmd      += '_' + self.probe_name.upper()
+            probe_accuracy_cmd       += '_' + self.probe_name.upper()
+            z_offset_apply_probe_cmd += '_' + self.probe_name.upper()
```
Wulfsta, line 58: This is a bit of an odd pattern to use - it would make much more sense to specify which probe to use as an input into the command. For instance, `QUERY_PROBE` would become `QUERY_PROBE PROBE=<name>`, with a default name for the backwards compatibility case.

cekim-git: The functionality you describe is already there...  

Any of the existing commands can be run as you describe in the existing edited version of probe.py:
PROBE PROBE=pinda0
QUERY_PROBE PROBE=pinda0
PROBE_ACCURACY PROBE=pinda0
and so on...

If no "PROBE=" is provided it defaults to "probe" by default.

The added functionality you are highlighting is that of re-naming commands which primarily serves to avoid causing an error by registering duplicate macros.

These renamed macros end up being duplicate functions.  The following will perform the same action:
PROBE PROBE=pinda0
OR
PROBE_PINDA0

So, this could be removed by blocking the registration of these macros for any but the default probe.

garethky:
This should also come with some documentation updates:
* Add the new, optional, `PROBE=` parameter to the various GCode commands in GCode.md
* Update to Configuration Reference to note that multiple `[probe xxx]` sections can be defined
* Update to Status Reference to explain how data from various probes can be retrieved
There are patterns in the existing docs for how to make these updates. Things like fans and extruders already do this.

cekim-git:
> FWIW, `display.py` demonstrates a pattern for registering "muxed" gcode commands that include a default when its possible to configure an extra with a standard section (`[display]`) and prefix sections(`[display my_lcd]`. Below are the relevant lines:
> 
> https://github.com/Klipper3d/klipper/blob/f2e69a3703f97725fb0db626ff592a223e22b3a2/klippy/extras/display/display.py#L208-L213
> 
> What happens here is that if the user runs `SET_DISPLAY_GROUP GROUP=mygroup` it will execute the command for the default `[display]`, however if the user runs `SET_DISPLAY_GROUP DISPLAY=my_lcd GROUP=mygroup`, it will execute the command for the `[display my_lcd]` object.

Thank you, this was what I was hoping to find... some more common way to handle this issue... I figured it must exist.... this makes sense, I will employ this method.

Wulfsta:
@cekim-git are you still planning to work on this?

PR #6716, opened 2024-10-22
Update OctoPrint.md - wrong serial address

JamesH1978:
This PR corrects a simple mistake where I gave the Unix socket not the serial pts.

Thanks
James


KevinOConnor:
Thanks.

-Kevin

PR #6717, opened 2024-10-23
Improve timer vs task priority check

KevinOConnor:
Rename `sched_tasks_busy()` to `sched_check_set_tasks_busy()` and change it to only return true if tasks are active (running or requested) for two consecutive calls.  This makes it less likely that timers will yield to tasks except when tasks really are notably backlogged.

This also makes it less likely that multiple steppers controlling the same rail will be interrupted by tasks mid-step.  This should slightly improve the timing, and make it less likely that a halt during homing/probing will occur with these steppers taking a different number of total steps.

This was discussed at: https://klipper.discourse.group/t/z-stepper-microstep-shift-on-multi-mcu-homing-loadcell-ldc1612/19524/

-Kevin


garethky:
Tested on the Voron 2.4 with the Load Cell code that is similarly impacted. Verified to work.

PR #6718, opened 2024-10-24
Probe position skew warning and ldc1612 irq fix

KevinOConnor:
During a probe attempt, the internal tracking of positions should not change.  Verify this after each probe and log a warning if the post-probe position tracking does not match the pre-probe position tracking.

Also, fix an error in the ldc1612 probe trigger notification - it should have called `trsync_do_trigger()` with irqs disabled.  Fix by changing `trsync_do_trigger()` to internally disable irqs to avoid issues like this in the future.

-Kevin

EDIT: This was discussed at: https://klipper.discourse.group/t/z-stepper-microstep-shift-on-multi-mcu-homing-loadcell-ldc1612/19524/

### DIFF
--- a/klippy/extras/homing.py
+++ b/klippy/extras/homing.py
@@ -1,6 +1,6 @@
 # Helper code for implementing homing operations
 #
-# Copyright (C) 2016-2021  Kevin O'Connor <kevin@koconnor.net>
+# Copyright (C) 2016-2024  Kevin O'Connor <kevin@koconnor.net>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import logging, math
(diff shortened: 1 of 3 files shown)


garethky:
I'll test this out and get back to you with a verification. The code currently in `endstop.c` does not explicitly disable the irq's when it calls `trsync_do_trigger()`. When I tried a calling `trsync_do_trigger()` from the load cell endstop's timer function, with IRQ's disabled, I got step drift. But I used `irq_enable` instead of `irq_restore` which, if I understand correctly, wont enable the IRQs inside the timer functions as they were already disabled.

garethky:
OK, checks out good for me.

PR #6719, opened 2024-10-24
Update Measuring_Resonances.md for some NumPY version issues

JamesH1978:
It has been noted over the last six to eight months that some versions of Numpy have issues with the klipper python environment on some machines. This PR introduces a fixed version that is known to work and a small test for people to do to make sure there are no output issues from the get go. These have been pulled from the pinned posts in the discord, from a time when 1.26 was causing issue, and now it seems v2 is also having some issues, hence the change. 

Thanks
James

### DIFF
--- a/docs/Measuring_Resonances.md
+++ b/docs/Measuring_Resonances.md
@@ -212,12 +212,20 @@ sudo apt install python3-numpy python3-matplotlib libatlas-base-dev libopenblas-
 
 Next, in order to install NumPy in the Klipper environment, run the command:
 ```
-~/klippy-env/bin/pip install -v numpy
+~/klippy-env/bin/pip install -v numpy==1.25.2
 ```
 Note that, depending on the performance of the CPU, it may take *a lot*
 of time, up to 10-20 minutes. Be patient and wait for the completion of
 the installation. On some occasions, if the board has too little RAM
-the installation may fail and you will need to enable swap.
+the installation may fail and you will need to enable swap. Also note
+the forced version, due to newer versions of NumPY having requirements
+that may not be satisfied in some klipper python environments.
+
+Once installed please check that no errors show from the command:
+```
+~/klippy-env/bin/python -c 'import numpy;'
+```
+The correct output should simply be a new line.
 
 #### Configure ADXL345 With RPi
 


dmbutyugin:
In general, I have no concerns about this change, as we are not using any special features from more recent releases. However, should we really force a specific version (1.25.2). Would it make sense to specify either ~=1.25.0 or <1.26 ? TBH, I do not really mind it either way.

JamesH1978:
I believe 1.26 was the first version to show issues, and 1.25.2 was the last known not to basically

dmbutyugin:
Sorry, I was probably unclear with my comment. What I meant was, instead of fixing a very specific version, wouldn't it be better to let pip choose the best available version (for a given version of installation) prior to 1.26?

JamesH1978:
Ah sorry, i reread and got your meaning now! yes i could do less than 1.26

dmbutyugin (approved):

KevinOConnor:
Okay, thanks.  I'll give a few days to see if there are other comments, and otherwise look to commit.

-Kevin

Sineos:
Thanks @JamesH1978. While you are already working on this document, please check https://klipper.discourse.group/t/documentation-update-for-input-shaper-page/19149

JamesH1978:
@Sineos yes we have seen a few people wonder why the docs don't say the right thing, when they don't follow them correctly ;) to be fair that doc needs a complete rewrite and maybe splitting out into a separate doc for adxl, mpu and lis, as it has become a tome.

Sineos:
I wholeheartedly agree @JamesH1978 
Guess it is the same as with software: Features are added, bugs are fixed and at a certain point you need to start refactoring as it has become hard to maintain.
Still, I think, Klipper has an outstanding documentation overall, given the complexity of the product.

In this case, it would be a small bug fix by exchanging `/tmp/resonances_x_*.csv` with `/tmp/calibration_data_x_*.csv`

Anyway, thank you for taking a little care of the documentation 👍

JamesH1978:
Its not a bug though :) the instructions that give that line are linked to the TEST_RESONANCES command, not SHAPER_CALIBRATE, where the former does output `resonances_x_*.csv`, unlike the latter which outputs `calibration_data_x_*.csv`. There is no part of this doc that references a png conversion for SHAPER_CALIBRATE

Sineos:
You are right. Everywhere the preceding command is indeed `TEST_RESONANCES` when it is about using the data files to generate a graph afterwards.

KevinOConnor:
Thanks.

-Kevin

PR #6721, opened 2024-10-24
Support atsame51n19 chip (for Duet3HC)

KevinOConnor:
Discussed at: https://klipper.discourse.group/t/klipper-on-duet3hc-without-usb/19273/13

-Kevin


Wulfsta:
What code is this using for 12MHz clock? I don’t see an implementation in the samd51 clock?

KevinOConnor:
Thanks for reviewing.  That does look like an issue.  Seems setting that is equivalent to setting "internal clock".  Probably best to raise on Discourse or Discord so that Anna Trubitcyna can respond.

-Kevin

KevinOConnor:
I removed the change to enable 12Mhz in Kconfig and rebased this branch.  (A follow up  PR can add support for 12Mhz if needed.)

-Kevin

PR #6724, opened 2024-10-29
Slightly expand BL Touch explanation of Pin configuration

do3cc:
Stating why the pullup is needed might help people who found wrong examples in the internet for probe configuration.


JamesH1978:

Thanks
James

Sineos:
TBH, I'm not sure if this not rather increases the confusion.
Most users do not know what the pull-up actually does and even less so what an open-drain is. For those who know both terms and what they do, the explanation is not needed.

do3cc:
There is already an ignored Pull Request and at least 7 open issues regarding that PIN and at least mentioning Klipper. 19 Bugs when also looking for Marlin. Meh, I didn't know how BTT treats the support of their hardware.
I will create a new pull request for the example config of the SKR Mini Board with a section for BLTouch which is outcommented. My problem seems to be SKR Mini specific.

PR #6725, opened 2024-10-29
Add support for rp2350 micro-controllers

KevinOConnor:
This adds support for the new Raspberry Pi RP2350 micro-controllers.  This has also been discussed at: https://klipper.discourse.group/t/support-for-rp2350-micro-controllers/19656

Note that the rp2350 has a serious hardware defect (RP2350-E9 errata) and as a result I do not recommend using this chip currently.  See the link above for more details.

The code on this PR has a few known limitations (to be addressed in future PRs):
* Only the first 30 GPIOs are supported. Support for the additional GPIOs on the rp2350b has not been added (nor support for ADC in general on the rp2350b).

-Kevin

EDIT: The chipid and reboot into bootloader functionality has also been implemented in this PR.


thijstriemstra:
In the thread you mention: " Implemented. Regular make flash should now work." Does this also apply to rp2040 or only rp2350?

KevinOConnor:
The rp2040 already supports `make flash`.  The changes here extend that logic to include rp2350 chips.  The rp2040 support is otherwise unchanged.

-Kevin

PR #6726, opened 2024-10-30
resonance_tester: Fix chips selection, add accel_per_hz selection

MRX8024:
-The current code does not have the option to "live" select a non-adxl345 `chips` when running `TEST_RESONANCES`, since it adds the adxl prefix by default.
-Added the ability to "live" select `accel_per_hz` when running `TEST_RESONANCES`.
-Minor fixes in the documentation.

Signed-off-by: Maksim Bolgov maksim8024@gmail.com

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20241112: Option `CHIPS=<chip_name>` in `TEST_RESONANCES` and
+`SHAPER_CALIBRATE` requires specifying the full name(s) of the accel
+chip(s). For example, `adxl345 rpi` instead of short name - `rpi`.
+
 20240912: `SET_PIN`, `SET_SERVO`, `SET_FAN_SPEED`, `M106`, and `M107`
 commands are now collated. Previously, if many updates to the same
 object were issued faster than the minimum scheduling time (typically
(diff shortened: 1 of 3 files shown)


nefelim4ag:
Every commit is expected to have a sign-off line.
PR on the other hand can just have a basic description, arguing, and description of a problem:
why it is a problem and how you are trying to solve it.

Also, take a look at: https://www.klipper3d.org/CONTRIBUTING.html
Specifically the commit format.

Thanks.

MRX8024:
Thanks for the tip, I think I fixed the formatting of the commits

dmbutyugin:
Thanks. I think this makes sense. However, given that this is a not a backwards-compatible change, I think it should be mentioned in `docs/Config_Changes.md` that the `CHIPS=` parameter for `TEST_RESONANCES` and `SHAPER_CALIBRATE` commands must specify the full accelerometer name(s).

MRX8024:
> the full accelerometer name(s).

Do you want me to add a straight `CHIPS=<full_chip_name>`, or just description?
The doc already mentions: "for example `CHIPS="adxl345, adxl345 rpi"`".
Thanks

dmbutyugin:
I think the G-Codes documentation is OK already. But since this is a change that will break any existing usages of `CHIPS` parameter, it is better to document this in `docs/Config_Changes.md` (it says 'config', however breaking gcode changes are also documented there).

KevinOConnor:
Thanks.

-Kevin

PR #6727, opened 2024-11-01
fan_generic: fixes missing logging import

pedrolamas:
At the moment, using `SET_FAN_SPEED` with a display template that does not resolve to a string that can get converted to a valid float will cause a crash as the `import logging` is missing.


KevinOConnor:
Thanks.

-Kevin

PR #6729, opened 2024-11-05
Load cell gram scale

garethky:
This PR adds the features needed to make a load cell actually work as a gram scale. This PR will let you weight filament or measure a force in grams. (no homing or probing yet!)

* Convert raw sensor `counts` to grams and make this force data available via unix socket and printer object status
* Track sensor calibration state and allow sensors to go from un-calibrated to calibrated at runtime
* Add basic GCodes for tearing and reading the load cell
* Add a guided calibration tool so users can calibrate a load cell with a known mass or measured force.
* Add diagnostic GCode to check the health and performance of the load cell and sensor

This code has been tested by several people in the community and some last minute bug fixes and changes were landed:
* The `reverse` option allows the polarity of the force readings to be inverted. You can have your probing collision graphs in either orientation now.
* `reference_tare_counts` is written to config and read as an integer. [link](https://github.com/garethky/klipper/pull/3)
* Fixed crash when guided calibration tool being entered twice
* Fixed crash when `CALIBRATE` was used before `TARE`

A lot of work (months) and bug fixing went into parts of this, particularly `LOAD_CELL_DIAGNOSTIC` and the `LoadCellSampleCollector`. These bits have to work when the sensor is buggy and still produce usable output. `LoadCellSampleCollector` also underpins later work for probing.

`numpy` is required for this PR. We could change that. But PR's for probing will absolutely need it. If you want to merge `[load_cell]` and `[load_cell_probe]` then it seems there is no point in trying to keep it out of this PR.

### DIFF
--- a/docs/API_Server.md
+++ b/docs/API_Server.md
@@ -364,37 +364,21 @@ and might later produce asynchronous messages such as:
 The "header" field in the initial query response is used to describe
 the fields found in later "data" responses.
 
-### hx71x/dump_hx71x
+### load_cell/dump_force
 
-This endpoint is used to subscribe to raw HX711 and HX717 ADC data.
-Obtaining these low-level ADC updates may be useful for diagnostic
-and debugging purposes. Using this endpoint may increase Klipper's
-system load.
+This endpoint is used to subscribe to force data produced by a load_cell.
+Using this endpoint may increase Klipper's system load.
 
 A request may look like:
-`{"id": 123, "method":"hx71x/dump_hx71x",
+`{"id": 123, "method":"load_cell/dump_force",
 "params": {"sensor": "load_cell", "response_template": {}}}`
 and might return:
-`{"id": 123,"result":{"header":["time","counts","value"]}}`
+`{"id": 123,"result":{"header":["time", "force (g)", "counts", "tare_counts"]}}`
 and might later produce asynchronous messages such as:
-`{"params":{"data":[[3292.432935, 562534, 0.067059278],
-[3292.4394937, 5625322, 0.670590639]]}}`
+`{"params":{"data":[[3292.432935, 40.65, 562534, -234467]]}}`
 
-### ads1220/dump_ads1220
-
-This endpoint is used to subscribe to raw ADS1220 ADC data.
-Obtaining these low-level ADC updates may be useful for diagnostic
-and debugging purposes. Using this endpoint may increase Klipper's
-system load.
-
-A request may look like:
-`{"id": 123, "method":"ads1220/dump_ads1220",
-"params": {"sensor": "load_cell", "response_template": {}}}`
-and might return:
-`{"id": 123,"result":{"header":["time","counts","value"]}}`
-and might later produce asynchronous messages such as:
-`{"params":{"data":[[3292.432935, 562534, 0.067059278],
-[3292.4394937, 5625322, 0.670590639]]}}`
+The "header" field in the initial query response is used to describe
+the fields found in later "data" responses.
 
 ### pause_resume/cancel
 
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -4758,6 +4758,16 @@ scale.
 [load_cell]
 sensor_type:
 #   This must be one of the supported sensor types, see below.
+#counts_per_gram:
+#   The floating point number of sensor counts that indicates 1 gram of force.
+#   This value is calculated by the LOAD_CELL_CALIBRATE command.
+#reference_tare_counts:
+#   The integer tare value, in raw sensor counts, taken when LOAD_CELL_CALIBRATE
+#   is run. This is the default tare value when klipper starts up.
+#sensor_orientation:
+#   Change the sensor's orientation. Can be either 'normal' or 'inverted'.
+#   The default is 'normal'. Use 'inverted' if the sensor reports a
+#   decreasing force value when placed under load.
 ```
 
 #### HX711
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -721,6 +721,40 @@ and RAW sensor value for calibration points.
 #### DISABLE_FILAMENT_WIDTH_LOG
 `DISABLE_FILAMENT_WIDTH_LOG`: Turn off diameter logging.
 
+### [load_cell]
+
+The following commands are enabled if a
+[load_cell config section](Config_Reference.md#load_cell) has been enabled.
+
+### LOAD_CELL_DIAGNOSTIC
+`LOAD_CELL_DIAGNOSTIC [LOAD_CELL=<config_name>]`: This command collects 10
+seconds of load cell data and reports statistics that can help you verify proper
+operation of the load cell. This command can be run on both calibrated and
+uncalibrated load cells.
+
+### LOAD_CELL_CALIBRATE
+`LOAD_CELL_CALIBRATE [LOAD_CELL=<config_name>]`: Start the guided calibration
+utility. Calibration is a 3 step process:
+1. First you remove all load from the load cell and run the `TARE` command
+1. Next you apply a known load to the load cell and run the
+`CALIBRATE GRAMS=nnn` command
+1. Finally use the `ACCEPT` command to save the results
+
+You can cancel the calibration process at any time with `ABORT`.
+
+### LOAD_CELL_TARE
+`LOAD_CELL_TARE [LOAD_CELL=<config_name>]`: This works just like the tare button
+on digital scale. It sets the current raw reading of the load cell to be the
+zero point reference value. The response is the percentage of the sensors range
+that was read and the raw value in counts.
+
+### LOAD_CELL_READ load_cell="name"
+`LOAD_CELL_READ [LOAD_CELL=<config_name>]`:
+This command takes a reading from the load cell. The response is the percentage
+of the sensors range that was read and the raw value in counts. If the load cell
+is calibrated a force in grams is also reported.
+
+
 ### [heaters]
 
 The heaters module is automatically loaded if a heater is defined in
--- a/docs/Load_Cell.md
+++ b/docs/Load_Cell.md
@@ -1,0 +1,122 @@
+# Load Cells
+
+This document describes Klipper's support for load cells. Basic load cell
+functionality can be used to read force data and to weigh things like filament.
+A calibrated force sensor is an important part of a load cell based probe.
+
+## Related Documentation
+
+* [load_cell Config Reference](Config_Reference.md#load_cell)
+* [load_cell G-Code Commands](G-Codes.md#load_cell)
+* [load_cell Status Reference](Status_Reference.md#load_cell)
+
+## Using `LOAD_CELL_DIAGNOSTIC`
+
+When you first connect a load cell its good practice to check for issues by
+running `LOAD_CELL_DIAGNOSTIC`. This tool collects 10 seconds of data from the
+load cell and resport statistics:
+
+```
+$ LOAD_CELL_DIAGNOSTIC
+// Collecting load cell data for 10 seconds...
+// Samples Collected: 3211
+// Measured samples per second: 332.0
+// Good samples: 3211, Saturated samples: 0, Unique values: 900
+// Sample range: [4.01% to 4.02%]
+// Sample range / sensor capacity: 0.00524%
+```
+
+Things you can check with this data:
+* The configured sample rate of the sensor should be close to the 'Measured
+samples per second' value. If it is not you may have a configuration or wiring
+issue.
+* 'Saturated samples' should be 0. If you have saturated samples it means the
+load sell is seeing more force than it can measure.
+* 'Unique values' should be a large percentage of the 'Samples
+Collected' value. If 'Unique values' is 1 it is very likely a wiring issue.
+* Tap or push on the sensor while `LOAD_CELL_DIAGNOSTIC` runs. If
+things are working correctly ths should increase the 'Sample range'.
+
+## Calibrating a Load Cell
+
+Load cells are calibrated using the `LOAD_CELL_CALIBRATE` command. This is an
+interactive calibration utility that walks you though a 3 step process:
+1. First use the `TARE` command to establish the zero force value. This is the
+`reference_tare_counts` config value.
+2. Next you apply a known load or force to the load cell and run the
+`CALIBRATE GRAMS=nnn` command. From this the `counts_per_gram` value is
+calculated. See [the next section](#applying-a-known-force-or-load) for some
+suggestions on how to do this.
+3. Finally, use the `ACCEPT` command to save the results.
+
+You can cancel the calibration process at any time with `ABORT`.
+
+### Applying a Known Force or Load
+
+The `CALIBRATE GRAMS=nnn` step can be accomplished in a number of ways. If your
+load cell is under a platform like a bed or filament holder it might be easiest
+to put a known mass on the platform. E.g. you could use a couple of 1KG filament
+spools.
+
+If your load cell is in the printer's toolhead a different approach is easier.
+Put a digital scale on the printers bed and gently lower the toolhead onto the
+scale (or raise the bed into the toolhead if your bed moves). You may be able to
+do this using the `FORCE_MOVE` command. But more likely you will have to
+manually moving the z axis with the motors off until the toolhead presses on the
+scale.
+
+A good calibration force would ideally be a large percentage of the load cell's
+rated capacity. E.g. if you have a 5Kg load cell you would ideally calibrate it
+with a 5kg mass. This might work well with under-bed sensors that have to
+support a lot of weight. For toolhead probes this may not be a load that your
+printer bed or toolhead can tolerate without damage. Do try to use at least 1Kg
+of force, most printers should tolerate this without issue.
+
+When calibrating make careful note of the values reported:
+```
+$ CALIBRATE GRAMS=555
+// Calibration value: -2.78% (-59803108), Counts/gram: 73039.78739,
+Total capacity: +/- 29.14Kg
+```
+The `Total capacity` should be close to the theoretical rating of the load cell
+based on the sensor's capacity. If it is much larger you could have used a
+higher gain setting in the sensor or a more sensitive load cell. This isn't as
+critical for 32bit and 24bit sensors but is much more critical for low bit width
+sensors.
+
+## Reading Force Data
+Force data can be read with a GCode command:
+
+```
+LOAD_CELL_READ
+// 10.6g (1.94%)
+```
+
+Data is also continuously read and can be consumed from the load_cell printer
+object in a macro:
+
+```
+{% set grams = printer.load_cell.force_g %}
+```
+
+This provides an average force over the last 1 second, similar to how
+temperature sensors work.
+
+## Taring a Load Cell
+Taring, sometimes called zeroing, sets the current weight reported by the
+load_cell to 0. This is useful for measuring relative to a known weight. e.g.
+when measuring a filament spool, using `LOAD_CELL_TARE` sets the weight to 0.
+Then as filament is printed the load_cell will report the weight of the
+filament used.
+
+```
+LOAD_CELL_TARE
+// Load cell tare value: 5.32% (445903)
+```
+
+The current tare value is reported in the printers status and can be read in
+a macro:
+
+```
+{% set tare_counts = printer.load_cell.tare_counts %}
+```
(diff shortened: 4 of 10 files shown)


garethky:
The tests failed because of the numpy dependency. I guess I could move the check out of `__init__` to get them to pass.

JamesH1978:
such a pity numpy is so bloaty and time consuming to just add it to requirements

Sineos:
> such a pity numpy is so bloaty and time consuming to just add it to requirements

I'm not sure that this has a real world impact. I guess most, or at least a significant portion also use input shaping and have it installed anyway.
Might be even better to add it to the requirements and call for a version <1.26 to avoid clashing with input shaping.

JamesH1978:
i think its the sheer time to compile numpy is the issue if the host is sub par, it can take hours on some

garethky:
I would love to get on my "just require numpy" soap box (Binaries are installed automatically for 32bit, you don't need to compiling it. Its shockingly fast vs plain python. ["Your job as a Python programmer is to write C"]https://www.youtube.com/watch?v=vVUnCXKuNOg)) but... I want to ship this and be 'done' with the project. So what do I have to do?

I can push the numpy require check back to the point where its clear that you are configuring a probe. Most of what I'm using it for in this PR (`average`, `min`, `max`, `unique`) is relatively easy to replace or remove.

For probing, I need `linalg.lstsq`. There are some optimizations I discussed with @KevinOConnor to replace the bulk of those calls, but I still have to make several calls to plain least squares. numpy/c is so much faster than Python that any sort of optimizations that we do might be washed away because they aren't in c (plus the code is very complicated). There is also, very very likely, a numpy/vectorized version of what we want to do that will crush the pure Python version (it has loops, its probably 50x slower than numpy). If this code is slow, it causes noticeable pauses.

docs/Load_Cell.md
```diff
@@ -0,0 +1,152 @@
+# Load Cells
+
+This document describes Klipper's support for load cells and load cell based
+probes.
+
```
thijstriemstra, line 5: not knowing what a load cell does/means for a 3d printer, could you add a sentence or two describing it's usage/benefits for 3d printers/klipper?

garethky: I made some updates to the documentation including a little bit more exposition at the top and a section about how to read the load cell gram force

garethky:
> Is there any way to provide / use a prebuilt v1.26 wheel?

For users: if you use Python 3.0 on a 32 bit RPI OS then [PiWheels](https://www.piwheels.org/) will just install a binary. That's the default behavior. You have to kinda work hard to not get that to happen. i.e. willfully use Python 2 or pick a 64 bit OS.

For the git build system: I'm not sure how that works.

Sineos:
> For users: if you use Python 3.0 on a 32 bit RPI OS then [PiWheels](https://www.piwheels.org/) will just install a binary.

[PyPi](https://pypi.org/project/numpy/1.26.4/#files) has some arm64 wheels as well 

> For the git build system: I'm not sure how that works.

There are examples of project who built the necessary wheels as CI actions. E.g. https://github.com/matrix-org/synapse/pull/14212/files

Maybe it would be an option to built the needed wheels (Python 2 and Python3?) as Klipper CI and the install scripts (or KIAUH) could automatically draw upon them.

garethky:
This was updated to remove the numpy dependency

docs/Load_Cell.md
```diff
@@ -0,0 +1,102 @@
+# Load Cells
+
+This document describes Klipper's support for load cells. Basic load cell
+functionality can be used to read force data and to weigh things like filament.
```
henrivdr, line 4: Consider adding a note that weighing things is only possible on printers with the load cells in the bed, and not those with the cell in the extruder.

garethky: I don't want to get into applications because there are many. You don't even have to put it in the printer! Several people want to use this functionality to weigh filament spools mounted outside the printer. E.g. https://klipper.discourse.group/t/filament-spool-scale-hx711/19800/
- You could check before a print if there is enough filament on the spool based on the print weight
- Track filament used by weight.
- Track drying if the filament by weight lost
In the future this may be in an application that is just a dry box and not even a printer. 
One community member is even working on runout detection in a resin printer.

docs/API_Server.md
```diff
@@ -396,6 +396,22 @@ and might later produce asynchronous messages such as:
 `{"params":{"data":[[3292.432935, 562534, 0.067059278],
 [3292.4394937, 5625322, 0.670590639]]}}`
 
+### load_cell/dump_force
+
+This endpoint is used to subscribe to force data produced by a load_cell.
+Using this endpoint may increase Klipper's system load.
```
KevinOConnor, line 402: It seems odd that there is an additional endpoint added for this information.  I would have thought the information could be added to the existing `hx71x/dump_hx71x` endpoint (or that endpoint replaced with this one).

docs/API_Server.md
```diff
@@ -396,6 +396,22 @@ and might later produce asynchronous messages such as:
...
+
+A request may look like:
+`{"id": 123, "method":"load_cell/dump_force",
+"params": {"sensor": "load_cell", "response_template": {}}}`
+and might return:
+`{"id": 123,"result":{"header":["time", "force (g)", "counts", "tare_counts"]}}`
+and might later produce asynchronous messages such as:
+`{"params":{"data":[[3292.432935, 40.65, 562534, -234467]]}}`
```
KevinOConnor, line 410: Just out of curiosity, does "tare_counts" change with each sample?

docs/Config_Reference.md
```diff
@@ -4659,6 +4659,15 @@ scale.
...
 #   This must be one of the supported sensor types, see below.
+#counts_per_gram:
+#   The floating point number of sensor counts that indicates 1 gram of force.
+#   This value is calculated by the CALIBRATE_LOAD_CELL command.
+#reference_tare_counts:
+#   The integer tare value, in raw sensor counts, taken when CALIBRATE_LOAD_CELL
+#   is run. This is the default tare value when klipper starts up.
+#reverse:
+#   Reverses the polarity of the load cell. This is a boolean value, the
+#   default is False.
```
KevinOConnor, line 4670: Out of curiosity, is "reverse" effectively the same as a negative "counts_per_gram"?

docs/G-Codes.md
```diff
@@ -700,6 +700,40 @@ and RAW sensor value for calibration points.
...
+`TARE_LOAD_CELL [LOAD_CELL=<config_name>]`: This works just like the tare button
+on digital scale. It sets the current raw reading of the load cell to be the
+zero point reference value. The response is the percentage of the sensors range
+that was read and the raw value in counts.
+
+### READ_LOAD_CELL load_cell="name"
+`READ_LOAD_CELL [LOAD_CELL=<config_name>]`:
+This command takes a reading from the load cell. The response is the percentage
+of the sensors range that was read and the raw value in counts. If the load cell
+is calibrated a force in grams is also reported.
```
KevinOConnor, line 734: As a general comment, it's generally preferable to implement less commands.  For example, could this be reduced to just two commands: LOAD_CELL_READ (for reading and diagnostics) and LOAD_CELL_CALIBRATE (for calibrating and changing the tare).

docs/Load_Cell.md
```diff
@@ -0,0 +1,102 @@
+# Load Cells
```
KevinOConnor, line 1: If a new document is added then it's also necessary to update docs/Overview.md and docs/_klipper3d/mkdocs.yml

klippy/extras/load_cell.py
```diff
@@ -3,21 +3,548 @@
...
+
+    def add_mux_endpoint(self, path, key, value, webhooks_start_resp):
+        self.webhooks_start_resp = webhooks_start_resp
+        wh = self.printer.lookup_object('webhooks')
+        wh.register_mux_endpoint(path, key, value, self._add_api_client)
+
+# Adapter for WebhooksHelper that transforms the response using a function
+# Anything that implements the add_client contract can be a message source
+# Outputs to its own clients
+class WebhooksTransformer(WebhooksHelper):
```
KevinOConnor, line 51: Can you elaborate on what this does (as well as WebhooksHelper)?

At first glance, it seems a little odd to make a slightly different interface for just loadcell.  If the BatchWebhooksClient code isn't flexible enough, maybe we should expand the webhooks interface and refactor the other users (like ldc1612 and accelerometers).

klippy/extras/load_cell.py
```diff
@@ -3,21 +3,548 @@
...
+        # startup, when klippy is ready, start capturing data
+        printer.register_event_handler("klippy:ready", self._handle_ready)
+
+    def _handle_ready(self):
+        self.add_client(self._on_sample)
+        # announce calibration status on ready
+        if self.is_calibrated():
+            self.printer.send_event("load_cell:calibrate", self)
+        if self.is_tared():
+            self.printer.send_event("load_cell:tare", self)
```
KevinOConnor, line 429: Can you elaborate on what these events will be used for?

KevinOConnor (commented):

In general it seems fine to me.  I have a few comments and questions - see below.

Cheers,
-Kevin

docs/G-Codes.md
garethky: Ok, I can make those changes.

docs/Load_Cell.md
garethky: Will do

docs/API_Server.md
garethky: No, it only changes when the tare command is executed.

docs/API_Server.md
garethky: Well there are 2 ADC endpoints `hx71x/dump_hx71x` and `ads1220/dump_ads1220`. 

My goal with this was to try to keep everything "load cell" related out of the ADC code. The `load_cell` is acting as a decorator to the raw ADC data stream. The raw stream is still useful as a diagnostic tool (or at least it was during development).

This goes with your other question about extending/updating the socket code. What we decide here is going to determine what to do about that comment.

At a high level, what are you thinking about here?
- Is the performance penalty of the duplicate streams a concern?
- Is the duplication of the raw counts data in 2 streams confusing?
- Everything is one stream is desired but then the ADC stream is redundant/confusing?

klippy/extras/load_cell.py
garethky: The probing code uses these events to get notified about the state of the gram scale. When the user interactively calibrates the `load_cell` for the first time it goes from uncalibrated -> calibrated. That state transition allows probing to work without a restart. Before that it will throw an error that you are not allowed to probe or home with a load cell that isn't calibrated.

I still have the probing code in a separate file because I believe this chunk of functionality is a nice cleanly separate unit. My intention is to pitch you on keeping them split up. These events are one part of the decoupling that makes that work.

klippy/extras/load_cell.py
garethky: This goes back to your first comment, lets resolve what we want for a data stream structure. If this is still required I can do a refactor.

docs/API_Server.md
KevinOConnor: A bit of all three.  Mostly concerned with maintaining two different interfaces that provide redundant capabilities, and there resulting developer confusion on why there are two redundant interfaces.

Perhaps take a look at how ldc1612.py takes a calibration object and calls `self.calibration.apply_calibration(samples)`, which then allows probe_eddy_current.py to populate a "Z" field in the samples.

docs/API_Server.md
KevinOConnor: Okay - seems a little odd to repeat this value for every sample.  Maybe just export the information via `get_status()`?

docs/API_Server.md
garethky: I look at the `ldc1612` code and I think the other option is a better fit for this use case. The "that endpoint replaced with this one" idea.

First, I think its more developer friendly to have the endpoint be named according to the load cell in the config and not the sensor implementation. That's a leaky abstraction. That way a front end doesn't need to look up the sensor type in the config and try to build the endpoint URI. The URI can be built from the load cell name alone and its always consistent. Force data comes from a thing that is about force. That feels consistent and logical. (You could argue that LDC1212 is only about Z so that's not necessarily inconsistent)

Second but less important: the 'B' channel on the HX717 is used to measure an RTD and a hall effect filament sensor in Prusa's new machines. I think it would make supporting that kind of thing less awkward later on if we deleted the endpoint from the sensors. Objects that consume the data (LoadCell, Temperature etc.) can transform and publish it (or not) as required.

I'll work on this first.

rcloran: > less important: the 'B' channel on the HX717 is used to measure an RTD and a hall effect filament sensor in Prusa's new machines

Hey, that is important! ;)

I've currently got a very rough proof of concept implementation for the filament sensor, based on your load-cell-probe-community-testing branch -- ie, using the WebhooksHelper/WebhooksTransformer in this PR. Probably the main feature that would be useful for my code is being able to subscribe to a particular channel only. At the moment my code has 3 or 4 places all receiving the whole data stream and doing their own filtering on it.

In day-to-day usage the raw counts data is not that useful, especially to the user -- dropping the raw sensor API endpoint seems fine. However, if a load cell ADC were to be reused by existing modules that currently just rely on regular ADC pins (gcode_button, hall_filament_width_sensor), the user will need access to the raw counts or the normalized 0..1 value somehow during the setup process in order to calibrate their new gadgets, since they won't necessarily have specific commands like in this PR. For the filament sensors it doesn't have to be the existing endpoint, though. The values change dramatically enough that getting even just one data point each with filament inserted and not inserted would allow for a reliable configuration.

docs/API_Server.md
garethky: Because a Tare can happen in the middle of a batch of results you wouldn't know which samples used what tare value. A UI would not correctly reflect whats really going on. And as pointed out, tare can happen frequently while probing.

docs/Config_Reference.md
garethky: Basically. This lets you do what you wanted, where if the force graph is naturally negative you can reverse its polarity.

docs/G-Codes.md
garethky: This does not work, it results in a `!! Malformed command` [error](https://github.com/Klipper3d/klipper/blob/master/klippy/gcode.py#L265). I don't see anywhere in the existing GCode command that tries to make 1 command perform different actions. The only way this would work would be:
```
LOAD_CELL READ=TRUE
LOAD_CELL TARE=TRUE
LOAD_CELL DIAGNOSTIC=TRUE
```
That looks horrible and is hard to use/discover in the front ends. I'd like to stick with separate commands. I have no problem prefixing them all with `LOAD_CELL_`, that actually improves discoverability. Looking at Bed Mesh, Angle, Accelerometer etc, that's the pattern they all follow.

docs/API_Server.md
garethky: Correction: it cant happen in the middle of a batch. The tare value would always apply to a whole batch.

docs/Load_Cell.md
garethky: Fixed

docs/G-Codes.md
garethky: Updated commands to use a consistent prefix `LOAD_CELL_`

docs/API_Server.md
garethky: Updated to not have the sensors register websocket endpoints.

There is still a class for handling both internal and websocket clients: ApiClientHelper. I need it in subsequent PR to emit events for the probe. If that looks like a class that should be in webhooks.py let me know.

For now I've kept things in separate commits so the changes are easier to see. But my intention is to squash them before its merged.

KevinOConnor:
Okay, thanks.   In general it seems fine to me.  The only thing I noticed is that docs/API_Server.md should be updated now that the hx71x/ads1220 api is no longer available, and Config_Changes.md should have a note that the previous API endpoints were removed.

As for the commands, I was thinking something like `LOAD_CELL_READ`, `LOAD_CELL_READ FULL_DIAGNOSTIC=1`, `LOAD_CELL_CALIBRATE`, and `LOAD_CELL_CALIBRATE UPDATE_TARE_ONLY=1`.  It's not a big deal though if you want to keep 4 commands.

FWIW, the "reverse" config option seems a little add to me, though I'm not sure if a negative counts_per_gram would be better.  It's not a big deal either way.  Note that the added docs have described "reverse" twice.

Cheers,
-Kevin

garethky:
Thanks for taking a look @KevinOConnor 

- I will address the comments about the docs, thanks for the feedback and the eyeballs
- If its all the same to you I will keep the 4 commands. I think that's move consistent with existing gcodes
- `reverse` has one advantage over negative `counts_per_gram`: it wont be overwritten in a calibration. The orientation stays the same between the calibrations with no user input. I could call it `invert`, maybe that's a better name.

🍻

cab404:
Just checked it. Appears to be working with hx711 — alas the calibration doesn't fix the sign for some reason.

garethky:
> alas the calibration doesn't fix the sign for some reason.

Can you explain what you mean? Seems like are reporting a bug?

garethky:
* hx71x & ads1220 API server doc entries deleted.
* made the `sensor_orientation` change but went with `normal` and `inverted`. "If the scale goes negative set the orientation to `inverted`" just sounded right to me?
* Squashed into 2 commits

I think this is ready to merge.

KevinOConnor:
Okay, thanks.  I'll give a few days to see if there are comments and otherwise look to commit.

```
#sensor_orientation:
#   Change the sensor's orientation. Can be either 'normal' or 'inverted'.
#   The default is 'inverted'.
```
I must admit, I got quite a chuckle out of that default.

Separately, if Load_Cell.md is a new document, it should also be added to `docs/_klipper3d/mkdocs.yml` (briefly mentioned at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review ).

Cheers,
-Kevin

garethky:
> I must admit, I got quite a chuckle out of that default.

😂 I'm doing too many things today! Its 'normal'.

Load_Cell.md is in the Overview:

cab404:
> > alas the calibration doesn't fix the sign for some reason.
> 
> Can you explain what you mean? Seems like are reporting a bug?

```
02:11:47  $ CALIBRATE GRAMS=520
02:11:48  // Calibration value: -3.63% (-304549), Counts/gram: 324.90769,             Total capacity: +/- 25.40Kg
02:12:24  $ CALIBRATE GRAMS=926
02:12:25  // Calibration value: -5.00% (-419427), Counts/gram: 306.51296,             Total capacity: +/- 26.93Kg
02:13:50  $ CALIBRATE GRAMS=1458
02:13:51  // Calibration value: -6.13% (-514252), Counts/gram: 259.70850,             Total capacity: +/- 31.78Kg
02:13:56  $ ACCEPT
02:13:56  // Load cell calibration settings:
// 
// counts_per_gram: 259.708505
// reference_tare_counts: -135596
```

All of our measurements displayed going into negative values, yet inversion wasn't reflected in the final results, nor was asked to be attributed in config during calibration itself

garethky:
> All of our measurements displayed going into negative values, yet inversion wasn't reflected in the final results, nor was asked to be attributed in config during calibration itself

In the calibration tool this is the expected output. Everything there that is negative is the raw counts from the sensor. The only measurement that gets inverted is the grams when reading the load cell.

cab404:
> In the calibration tool this is the expected output. Everything there that is negative is the raw counts from the sensor. The only measurement that gets inverted is the grams when reading the load cell.

The problem is that after the calibration it shows negative values, cause the fact that counts per gram is negative is not saved)

garethky:
> > In the calibration tool this is the expected output. Everything there that is negative is the raw counts from the sensor. The only measurement that gets inverted is the grams when reading the load cell.
> 
> 
> 
> The problem is that after the calibration it shows negative values, cause the fact that counts per gram is negative is not saved)

Yes. If the wiring is 'backwards' it will always show negative grams, unless you go into the config and set `sensor_orientation: inverted`. This is very similar to how steppers work. If your stepper moves in the wrong direction, you can swap 2 wires or add a `!` to the `dir_pin`.

cab404:
> Yes. If the wiring is 'backwards' it will always show negative grams, unless you go into the config and set sensor_orientation: inverted. This is very similar to how steppers work

yeah, but if in case of steppers there is no automatic calibration, here code can actually deduce your orientation during calibration process

maybe this is actually good, and is an additional protection from assembly mistakes or something like that — but I still thought it is worth asking about

KevinOConnor:
Thanks.

-Kevin

PR #6730, opened 2024-11-06
config: add warning about new-stock ender 3s

BluCobalt:
My Ender 3 (bought in 2021?) came with a 4.2.2 board out of the box. I just want to add this note to hopefully help someone else, because I lost a good couple of hours trying to figure this out ;)


JamesH1978:

Thanks
James

BluCobalt:
Should be good now.

JamesH1978:
Thanks, looks fine to me as an informational, as this is indeed the case. The other option would be to create a whole new file referencing the date these started shipping without the melzi boards, but I think that could get confusing. @KevinOConnor what do you think?

Thanks
James

BluCobalt:
Actually, that might make more sense. A little bit of research says that this hardware change happened in July of 2020. Let me make that new config.

PR #6733, opened 2024-11-09
stm32: Enable hw pwm for pin PA3

Jyrka98:
This adds support for using hardware pwm on pin PA3 on STM32H7 based boards. In my case, for FAN3 on a BTT Octopus Max EZ


JamesH1978:

Thanks
James

PR #6734, opened 2024-11-11
ldc_1612: Supports configurable external crystal frequency

kluoyun:
ldc_1612: Supports configurable external crystal frequency

You can use the 40Mhz crystal oscillator recommended by TI official manual to get the best performance.
refer to: [ldc1612.pdf](https://www.ti.com/cn/lit/ds/symlink/ldc1612.pdf) 7.3.4

Signed-off-by: XiaokuiZhao<xiaok@zxkxzk.cn>


KevinOConnor:
Thanks.  We can certainly add this option.  It looks like this PR wont merge as is though.  Also, why the change in the signed-off-by line (relative to your previous PRs)?

-Kevin

kluoyun:
Same as before, it's me, just written differently.

nefelim4ag (approved):

KevinOConnor:
Thanks.  The change itself looks fine, but the signed-off-by line has me confused - at first glance it looks like an alias instead of a full name.  Is there a reason not to use the previous name you've submitted in past PRs?

Also, any chance we could make the config option "frequency" instead of "freq"?

-Kevin

kluoyun:
Sure, thanks a lot for the reminder.

KevinOConnor:
Thanks.

-Kevin

vvuk:
Note that when operating in single-channel mode, the [datasheet specifies that the valid Fref range is <= 35MHz](https://www.ti.com/document-viewer/LDC1612/datasheet#clocking-architecture-snoscz04376/SNOSCZ04376); 40MHz is technically not valid for single channel, and a clock divider of 2 must be used to get it into a valid range. Mellow Fly docs currently currently tell users to patch klipper to change the ldc1612 frequency to 40MHz (I'm guessing this is where this PR comes from, to avoid that patch), and it "probably" works fine, but it's not to spec.

nefelim4ag:
@vvuk
JFYI, I've seen the data from SHT36v3.
It seems that the sensor is 5MHz, and that it works fine without the divider with CLKIN 40MHz.

_I've played with the internal clock source, which is 46MHz in my case. The addition of the divider does not change the noise. So, I guess it is out of spec, but it is fine._

PR #6735, opened 2024-11-12
Angle: add support for MT6816, MT6826S sensors

nefelim4ag:
For now, this is just a proposal - till there are other real users.

MT6816 - simple 14-bit angle sensor, can be read and that's it.
MT6826S - Has calibration capabilities - implemented, hysteresis, and 15-bit output
MT6835 - (21 bit) _should_ work with code from MT6826S, we anyway will ignore additional bits.

For now, I mostly play around with MT6826S.

Thanks.

---
BTW, it would be nice to have a generic query command for the angle sensor, like `ACCELEROMETER_QUERY`.


KevinOConnor:
Thanks.  As very high-level feedback, we can certainly add support for these additional sensors.  The code would need to pass the regression tests though.

Cheers,
-Kevin

nefelim4ag:
But for now it is failing because of image limitations:
```
  /home/runner/work/klipper/klipper/ci_build/or1k-linux-musl-cross/bin/../lib/gcc/or1k-linux-musl/10.2.1/../../../../or1k-linux-musl/bin/ld: Klipper image is too large
```
The only idea that I have is to disable angle sensors for ar100.

KevinOConnor:
Thanks.  I'm fine with adding this.

I noticed a couple of minor things:
* if a new ANGLE_CHIP_CALIBRATION command is added then it should be added to docs/G-Codes.md
* I think ar100 wants sensor_adxl345.c, but doesn't care about sensor_angle.c.  So, I think it would be best if we removed CONFIG_WANT_SENSORS and replaced it with CONFIG_WANT_THERMOCOUPLE, CONFIG_WANT_ADXL345, and CONFIG_WANT_ANGLE_SENSOR.  I know this is a bit of a pain to change, but it should only involve modifying src/Makefile and src/Kconfig.

Cheers,
-Kevin

PR #6736, opened 2024-11-12
Corrected path for printer.cfg

c15yi:
Looks like the path for `printer.cfg` changed a while ago according to this [forum post](https://klipper.discourse.group/t/printer-cfg-in-blank/5411/2).


JamesH1978:

You also have a whitespace issue you need to clear up, so the build checks can pass.

Thanks
James

docs/Installation.md
```diff
@@ -179,8 +179,8 @@ Another option is to use a desktop editor that supports editing files
 over the "scp" and/or "sftp" protocols. There are freely available tools
 that support this (eg, Notepad++, WinSCP, and Cyberduck).
 Load the printer config file in the editor and then save it as a file
-named "printer.cfg" in the home directory of the pi user
-(ie, /home/pi/printer.cfg).
+named "printer.cfg" under `~/printer_data/config`
+(ie, /home/pi/printer_data/config/printer.cfg).
```
JamesH1978, line 183: Probably best to word it so that the user knows that pi may not be the username

Thanks
James

JamesH1978 (changes requested):
Thanks for signing off, please see my one comment.

James

docs/Installation.md
c15yi: Makes sense. Pushed an update.

PR #6738, opened 2024-11-13
sgp40: Add sgp40 support 

meteyou:
Add support for the SGP40 Sensor for VOC mesurement. This sensor needs a temp + huminity sensor to work correct. If you don't add one, it will use 25°C and 50% huminity as reference.

Here is an example config:
```yaml
[temperature_sensor BME_OUT]
sensor_type: BME280
i2c_address: 118
i2c_mcu: pico
i2c_bus: i2c1e
i2c_speed: 400000

[temperature_sensor SGP_OUT]
sensor_type: SGP40
i2c_mcu: pico
i2c_bus: i2c1e
ref_temp_sensor: bme280 BME_OUT
ref_humidity_sensor: bme280 BME_OUT
```

Link to discourse: https://klipper.discourse.group/t/support-for-sgp40-sensors/20033

### DIFF
--- a/klippy/extras/sgp40.py
+++ b/klippy/extras/sgp40.py
@@ -1,0 +1,914 @@
+# SGP40 I2C-based VOC sensor support
+#
+# Copyright (C) 2024 Stefan Dej <meteyou@gmail.com>
+# Copyright (C) 2022 Adrien Le Masle
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+#
+# Based on the Nevermore sgp40.py code:
+# https://github.com/nevermore3d/Nevermore_Max/blob/master/Software/Klipper/
+
+import logging
+import math
+from . import bus
+from struct import unpack_from
+
+SGP40_I2C_ADDR = 0x59
+SGP40_WORD_LEN = 2
+
+SGP40_CMD = {
+    "GET_SERIAL": [0x36, 0x82],
+    "HEATER_OFF": [0x36, 0x15],
+    "SOFT_RESET": [0x00, 0x06],
+    "SELF_TEST": [0x28, 0x0E],
+    "MEASURE_RAW_NO_COMP": [0x26, 0x0F, 0x80, 0x00, 0xA2, 0x66, 0x66,
+                             0x93],
+    "MEASURE_RAW": [0x26, 0x0F]
+}
+
+# Constants for the VOC algorithm
+_VOC_SAMPLING_INTERVAL = 1
+_VOC_INITIAL_BLACKOUT = 45
+_VOC_INDEX_GAIN = 230
+_VOC_SRAW_STD_INITIAL = 50
+_VOC_SRAW_STD_BONUS = 220
+_VOC_TAU_MEAN_VARIANCE_HOURS = 12
+_VOC_TAU_INITIAL_MEAN = 20
+_VOC_INITI_DURATION_MEAN = 2700
+_VOC_INITI_TRANSITION_MEAN = 0.01
+_VOC_TAU_INITIAL_VARIANCE = 2500
+_VOC_INITI_DURATION_VARIANCE = 5220
+_VOC_INITI_TRANSITION_VARIANCE = 0.01
+_VOC_GATING_THRESHOLD = 340
+_VOC_GATING_THRESHOLD_INITIAL = 510
+_VOC_GATING_THRESHOLD_TRANSITION = 0.09
+_VOC_GATING_MAX_DURATION_MINUTES = 180
+_VOC_GATING_MAX_RATIO = 0.3
+_VOC_SIGMOID_L = 500
+_VOC_SIGMOID_K = -0.0065
+_VOC_SIGMOID_X0 = 213
+_VOC_INDEX_OFFSET_DEFAULT = 100
+_VOC_LP_TAU_FAST = 20
+_VOC_LP_TAU_SLOW = 500
+_VOC_LP_ALPHA = -0.2
+_VOC_PERSISTENCE_UPTIME_GAMMA = 10800
+_VOC_MEAN_VARIANCE_ESTIMATOR__GAMMA_SCALING = 64
+_VOC_MEAN_VARIANCE_ESTIMATOR__FIX16_MAX = 32767
+_FIX16_MAXIMUM = 0x7FFFFFFF
+_FIX16_MINIMUM = 0x80000000
+_FIX16_OVERFLOW = 0x80000000
+_FIX16_ONE = 0x00010000
+
+class SGP40:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.name = config.get_name().split()[-1]
+        self.reactor = self.printer.get_reactor()
+        self.i2c = bus.MCU_I2C_from_config(
+            config, default_addr=SGP40_I2C_ADDR, default_speed=100000)
+        self._error = self.printer.command_error
+        self._callback = None
+        self.report_time = config.getint('sgp40_report_time', 1, minval=1)
+        self.temp_sensor = config.get('ref_temp_sensor', None)
+        self.humidity_sensor = config.get('ref_humidity_sensor', None)
+        self.soft_reset_on_startup = config.getboolean(
+            'soft_reset_on_startup', False)
+
+        self.temp = 25
+        self.humidity = 50
+        self.min_temp = self.max_temp = 0.
+        self.sample_timer = self.reactor.register_timer(self._sample_sgp40)
+
+        self.voc_algorithm = VOCAlgorithm()
+        self.raw = self.voc = 0.
+        self._measuring = False
+        self.printer.add_object("sgp40 " + self.name, self)
+        self.printer.register_event_handler("klippy:connect",
+                                            self.handle_connect)
+
+    def handle_connect(self):
+        self._init_sgp40()
+        self.reactor.update_timer(self.sample_timer, self.reactor.NOW)
+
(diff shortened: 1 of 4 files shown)


scripts/klippy-requirements.txt
```diff
@@ -9,3 +9,4 @@ greenlet==3.0.3 ; python_version >= '3.12'
 Jinja2==2.11.3
 python-can==3.3.4
 markupsafe==1.1.1
+sensirion-gas-index-algorithm==3.2.1
```
thetic, line 12: this library requires at least python 3.6[^1], but klipper is still testing against python 2. I'm sure this is intentional, but I'm not sure why. It went end of life years ago[^2].

[^1]: https://pypi.org/project/sensirion-gas-index-algorithm/
[^2]: https://www.python.org/doc/sunset-python-2/

meteyou: Thx for your feedback!

That was also a reason why I opened the post in the Klipper discourse, "if I should use a python dependency" or upload the algorithm itself in a file in the Klipper repo.

I didn't continue to develop on this PR until I get an answer for this question...

thetic: I think this test failure answers the question for you, lol. To make this python2 compatible, another implementation of the algorithm is needed. Putting it in klipper itself seems like the most stable option to me.

meteyou: There are also python 2 compatible libs available. That's why I ask... I hope I have time this weekend to continue here. Thx for your feedback.

klippy/extras/sgp40.py
```diff
@@ -0,0 +1,190 @@
...
+            crc ^= byte
+            for _ in range(8):
+                if crc & 0x80:
+                    crc = (crc << 1) ^ 0x31
+                else:
+                    crc = crc << 1
+        return crc & 0xFF
+
+    def _log_message(self, message):
+        return f"SGP40 {self.name}: {message}"
```
thetic, line 177: Python 2 does not have f-strings.

```suggestion
        return f"SGP40 %s: %s" % (self.name, message)
```

klippy/extras/sgp40.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # Get the serial number to verify communication
+        serial = self._write_command(SGP40_CMD["GET_SERIAL"], read_len=6,
+                                     delay=.001)
+        if serial is None:
+            logging.error(self._log_message("Failed to read serial number, "
+                                            "communication error"))
+        else:
+            serial_str = ''.join([f'{x:02x}' for x in serial])
+            logging.info(self._log_message(f"Serial number: {serial_str}"))
```
thetic, line 74: f-string

klippy/extras/sgp40.py
```diff
@@ -0,0 +1,190 @@
...
+        self._write_command(SGP40_CMD["SOFT_RESET"], delay=.001)
+
+        # Get the serial number to verify communication
+        serial = self._write_command(SGP40_CMD["GET_SERIAL"], read_len=6,
+                                     delay=.001)
+        if serial is None:
+            logging.error(self._log_message("Failed to read serial number, "
+                                            "communication error"))
+        else:
+            serial_str = ''.join([f'{x:02x}' for x in serial])
```
thetic, line 73: f-string

meteyou:
@thetic thx! I refactored all f-strings

klippy/extras/sgp40.py
```diff
@@ -0,0 +1,871 @@
...
+
+    def get_status(self, eventtime):
+        return {
+            'temperature': round(self.temp, 2),
+            'humidity': round(self.humidity, 1),
+            'gas_row': self.raw,
+            'gas': self.voc,
+        }
+
+class VOCAlgorithm:
```
thetic, line 228: the license for this code (MIT) and attribution must be preserved.

klippy/extras/sgp40.py
```diff
@@ -0,0 +1,872 @@
...
+_FIX16_ONE = 0x00010000
+
+class SGP40:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.name = config.get_name().split()[-1]
+        self.reactor = self.printer.get_reactor()
+        self.i2c = bus.MCU_I2C_from_config(
+            config, default_addr=SGP40_I2C_ADDR, default_speed=100000)
+        self.report_time = config.getint('sgp40_report_time', 1, minval=1)
```
thetic, line 74: sampling must be at 1Hz for this algorithm according to the datasheet. The algorithm parameters all assume such. It should not be configurable.

CRTified:
Hey, I wanted to test your PR as I would really love a more native support for the SGP40 after the [Nevermore Max repo](https://github.com/nevermore3d/Nevermore_Max/tree/master/Software/Klipper) got archived. My setup worked with that code before.

However, I fail to get your changes running with my printer, running into an i2c NACK at boot. The SGP40 is on the same i2c bus of the host (Rock 4C+) as the BME280 I wanted to use as reference. Using just the BME280 works fine.

[Here's the klipper.log](https://github.com/user-attachments/files/19045263/klipper.log). Maybe it helps with debugging.

meteyou:
hey @CRTified thx for your response! I know this issue, but haven't time until now to debug this issue. Maybe @thetic found a solution here: https://github.com/thetic/klipper-sgp40/issues/4#issuecomment-2692320560

I haven't test his fix until now.

CRTified:
Thanks for the pointer. I don't have a lot of time for debugging as well, but I'll check the fix out and report back there - in the worst case, it's just one more data point.

meteyou:
I fixed all reported issues, upgraded to the latest i2c stuff and added the docs. I tested it on my Voron Trident with a CAN Stealthmore PCB:

Config:
```yaml
[temperature_sensor _BME_OUT]
sensor_type: BME280
i2c_address: 119
i2c_mcu: nevermore
i2c_bus: i2c2_PB10_PB11

[temperature_sensor Nevermore_OUT]
sensor_type: SGP40
i2c_mcu: nevermore
i2c_bus: i2c2_PB10_PB11
ref_temp_sensor: bme280 _BME_OUT
ref_humidity_sensor: bme280 _BME_OUT
#soft_reset_on_startup: False

[temperature_sensor _BME_IN]
sensor_type: BME280
i2c_address: 119
i2c_mcu: nevermore
i2c_bus: i2c1_PB8_PB9

[temperature_sensor Nevermore_IN]
sensor_type: SGP40
i2c_mcu: nevermore
i2c_bus: i2c1_PB8_PB9
ref_temp_sensor: bme280 _BME_IN
ref_humidity_sensor: bme280 _BME_IN
#soft_reset_on_startup: False
```

Screenshot from the Mainsail Temperature panel:
Bildschirmfoto 2026-06-30 um 15 12 09

KevinOConnor:
Thanks.  In general it seems fine to me.  I have a few minor comments.

I don't understand the comments in the top copyright block.  If the nevermore folks have a copyright, then please state their copyright explicitly.  Alternatively, if they don't have a copyright and you are adding commentary to the code, then please place that commentary separate from the main copyright block.

Similarly, if the VOCAlgorithm class was directly ported from some other source, then please place that statement (and associated copyright information) directly with that class.

FWIW, it's a little "goofy" that VOCAlgorithm is implementing 16bit divide, multiply, and sqrt code instead of using the more precise builtin Python math functions.  If you want to do that just to reuse the existing code, then it is okay with me, but some commentary at the top of the class would help future devs.

Cheers,
-Kevin

meteyou:
Thanks for the review!

I've updated the copyright/comment blocks. I moved the VOCAlgorithm copyright block to the VOCAlgorithm class incl. note for the 16bit math stuff. I also fixed the file copyright block and added the nevermore copyright author, but there was no email address in the original nevermore file to add it in the header here.

thetic:
I've been maintaining a fork of the old nevermore code at https://github.com/thetic/klipper-sgp40. I dropped the old fixed-point math and made some (rather invasive) tweaks to avoid the pervasive shutdowns caused by I2C NACKs from the sensors. It's GPL, so you're welcome to copy or adapt whatever you'd like. I will warn you that my I2C handling is more of a hack than anything else; I think these sensors would require some more invasive changes to the I2C stack for robust support.

meteyou:
@thetic thx for your feedback!

@KevinOConnor I checked the repo from @thetic and I found these features in his repo:
- CRC validation on reads (easy to adapted)
- persistant calibration (gcode CALIBRATE_SGP40 and save with SAVE_CONFIG)
- Heater-aware calibration gating: pause baseline learning while a heater is active/hot
- multi-sensor sync

I would implement the CRC validation here before you merge it. The rest of the list is optional in my opinion. What do you think about these features?

KevinOConnor:
Thanks.  The changes here look okay to me.  I have an additional question - why does spg40 export `gas_raw` while other sensors report `gas`?  How is a user expected to distinguish between these two fields?

> I would implement the CRC validation here before you merge it. The rest of the list is optional in my opinion. What do you think about these features?

At a quick glance, I'd suggest adding desired features as follow up commits in a new PR after merging this one.  I don't know enough about the sensor to give feedback on the features being needed or worthwhile.

Let me know what you want to do.

-Kevin

EDIT: If your preference is to add crc checking to this PR and possibly tackle other features later then that is okay with me.

thetic:
this is the patch i'm talking about: https://github.com/thetic/klipper-sgp40/blob/main/src/klipper_sgp40/__init__.py#L234-L255. Without this, the printer **will** shutdown eventually (usually every 2-3 days).

PR #6739, opened 2024-11-15
axis_twist_compensation: remove auto parameter from axis_twist_compensation

yochiwarez:


KevinOConnor:
Okay, thanks.  If there are no further comments I'll commit this PR in a couple of days.

@blastrock - fyi.

-Kevin

blastrock (approved):

KevinOConnor:
What's the status of this PR?  It looks like it was changed to be in the "draft" state.

-Kevin

yochiwarez:
I've been working on enabling the calibration of any type of inaccuracy across the entire bed. This time, I've used a mesh and interpolation to calculate the difference between the probe mesh and the manual mesh, allowing compensation at any point based on that difference.

I've tested it, and it seems to be working correctly. However, I look forward to your feedback and suggestions.

image

image

tiagomaricate:
Hi guys!

Just checking, what is the last stable version?
I'm trying following the official documentation and I'm get several errors.

yochiwarez:
> Hi guys!
> 
> Just checking, what is the last stable version? I'm trying following the official documentation and I'm get several errors.

In the current official version, you can apply compensation on either the X or Y axis, but not both at the same time. This will cause issues in the central area where they intersect. The AUTO parameter does not work.

KevinOConnor:
Is this ready to be merged?

Thanks,
-Kevin

JamesH1978:
Can we merge this please? It has commented in other support channels that this auto system does not work. @yochiwarez we will do this tomorrow, unless a good reason otherwise.

Thanks
James

blastrock (approved):
I just noticed that a review was requested from me, don't hesitate to ping me if I don't notice such things for such a long time (assuming I was the blocking factor) ^^

Anyway, lgtm

KevinOConnor:
Thanks.  I committed this change along with fbd5b492 on top.

-Kevin

Sineos:
For future reference: https://github.com/Klipper3d/klipper/pull/6624#issuecomment-2476961864

moccor:
Copy and pasting my discord message, since this change just made things worse for anybody that used/uses axis_twist_compensation. -
"So apparently the reason it wouldn't fix is because it simply only wants X or Y in the "save_config". Previously both X + Y worked. I wish changes like this would come with warnings. Especially something like twist_compensation, since people that are using it, are using it as a last resort. Nobody uses it cuz they want to, they use it cuz they have to. So when it changes and forces you to pick one or the other, now you gotta redo mesh and other stuff and hope it works fine again because twists are the biggest pain"

blastrock:
Well, I don't like it either when such regression happen. I admit I didn't think this worked at all, since the math seems wrong to me, which is why I approved it.

I think it's not too hard to actually implement both axis compensation. I think the only thing that makes the current math wrong is the fact that we normalize the measurements by their average: https://github.com/Klipper3d/klipper/blob/47aa28e5300d75764add52e34b23b1804147626c/klippy/extras/axis_twist_compensation.py#L326-L329 . Here's my math, tell me if it's completely off:

> The twist on the X axis and Y axis add a relative error `ex(x)` and `ey(y)` respectively. So we want to apply the compensation `c(x, y) = ex(x) + ey(y)` to any point on the bed. When we do the calibration of one of the axis, let's say X, we fix the Y axis and we measure manually `c(x, fixed_y)`. Thus, we don't actually measure `ex(x)` alone, but `ex(x) + ey(fixed_y)`.
>
> My proposal is that in addition to the above definition `c(x, y) = ex(x) + ey(y)`, we add two more definitions `ex(125) = 0` and `ey(125) = 0` (assuming a bed size of 250x250). This is allowed because we are only interested in relative values of `c(x, y)` (right?).
>
> With these additional definitions, when we calibrate X, we measure `c(x, 125) = ex(x) + ey(125) = ex(x) + 0`, so `ex(x)` alone. Same thing for the Y axis.
>
> I took the center of the bed as reference, but I think any point would work. I think in practice we should take the homing position `safe_z_home.home_xy_position` ( https://www.klipper3d.org/Config_Reference.html#safe_z_home ).

TL;DR: Here's my proposal:

1. We revert part of this PR to allow compensating both x and y (i.e. removing https://github.com/Klipper3d/klipper/pull/6739/files#diff-bbcb105a3a471b724cf76b5977fbb0b4aba70efc389f922ed729abd0a536e9ddR48-R61 ). This will allow people who use the feature to continue using it as they did before.
2. We (might be me) make a PR with the above changes to make this properly work. I think I can make the code quickly enough, maybe this weekend, but I won't have much time to test it, any help there will be appreciated.

Thoughts?

PS: as an immediate workaround for people who want the feature back right now, you can checkout the code from just before this merge:
```
git checkout b50d740
```

KevinOConnor:
Thanks.  I don't use axis_twist_compensation and I'm not super familiar with it - so I'm looking for others to tell me what needs to be done.

If you want to create a PR that allows one to use a previously configured x/y calibration then I'll go ahead and merge that.

If you want to create a subsequent PR that adds back in a new x/y calibration system then that's fine too.

Thanks again,
-Kevin

torte71:
> 2. We (might be me) make a PR with the above changes to make this properly work. I think I can make the code quickly enough, maybe this weekend, but I won't have much time to test it, any help there will be appreciated.

Many thanks for addressing this.
If I get notified about that change (that should happen automatically if you mention this PR in the upcoming PR), then I should be able to test it in a timely manner.

blastrock:
I have thought this through, and tried to implement my solution. Halfway through, I concluded that what I tried to implement would have no effect at all, and I am now convinced that the current code actually works as expected.

The offsets we calculate for Z compensation over X or Y are relative and not absolute. When we calibrate or compensate X axis during Y calibration, we only add a constant to the measurements, and this has no effect to the relative values. This constant would just move the bed mesh up or down, and the bed mesh itself is relative too anyway, it only becomes absolute when the printer homes the Z axis

**tl;dr**: I changed my mind, the math now seems correct to me. I'm not sure why @yochiwarez wanted to prevent this use. I hope users like this feature :)

moccor:
Just want to make sure, but from my understanding it sounds like the X and Y compensation is confirmed to be working as intended and was re-added back to the latest builds of Klipper? Just curious for if I update

blastrock:
Yes

borislavstefanov:
Are you using QUAD_GANTRY_LEVEL before and is it suitable for crow, because as far as I can see it is only for 3 points along Y, it does not calibrate the entire plate along X and Y?

PR #6742, opened 2024-11-18
buildcommands: Add support for a versionstamp file

Laikulo:
Adds support for a versionstamp file at src/.version, similar to the one used by klippy at klippy/.version.

If git cannot determine the version being built (either because git is not present on the system, or klipper was downloaded as a tar/zip from github) buildcommands will read src/.version as a fallback. Versions determined this way are always considered "clean".


KevinOConnor:
I think this should be addressed with PR #7134 .

-Kevin

Laikulo:
Closing in favor of #7134 as per above

PR #6743, opened 2024-11-18
TMC_Driver docs:  More info for OverTemp error

Hello1024:
Add info useful to advanced users on other ways to deal with overtemp errors.

A few extra suggestions for basic solutions too.

Remove the suggestion to cool motors more (this error is specifically about the controllers, and will not be triggered by a motor overheat)

### DIFF
--- a/docs/TMC_Drivers.md
+++ b/docs/TMC_Drivers.md
@@ -540,9 +540,13 @@ Some common errors and tips for diagnosing them:
 #### TMC reports error: `... ot=1(OvertempError!)`
 
 This indicates the motor driver disabled itself because it became too
-hot. Typical solutions are to decrease the stepper motor current,
-increase cooling on the stepper motor driver, and/or increase cooling
-on the stepper motor.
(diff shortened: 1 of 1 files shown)


Sineos:
TBH, I would be leery of such advice. 
1. Typical OT errors are pretty rare in my experience.
2. On a well-built system, there should be no need to adjust these TMC registers.
3. Sentences like "Advanced users might want to..." are read as: I definitely have to mess with them to fix something with a completely different root cause, and I have no idea what they mean.
4. Playing with these registers often does more harm than good, as the documentation and dependencies on other settings are not fully transparent.

Please be also aware that you need to sign off to accept the developer certificate of origin, please see point 3 in [master/docs/CONTRIBUTING.md#what-to-expect-in-a-review](https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md?rgh-link-date=2024-10-29T08%3A55%3A40Z#what-to-expect-in-a-review)

dewi-ny-je:
Wouldn't advanced users already know that?

PR #6745, opened 2024-11-19
Update printer-anycubic-kobra-plus-2022.cfg

billzball:


JamesH1978:
Please can you describe what these changes achieve? As far as i am aware these boards are hardware locked to 16 microsteps and with no tmc control, this cant be changed in software. Also why the 2 identical PR's?

and finally please be aware that you need to sign off to accept the developer certificate of origin, please see point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks
James

PR #6749, opened 2024-11-26
gcode: Improve handling of extended g-code commands with '*;#' characters

KevinOConnor:
The g-code command parser did not allow three characters to be passed as parameters to commands (asterisk, semicolon, pound sign).  Rework the parsing code to better leverage the python shlex package so that these characters can be supported.

In particular, this should allow better support for printing g-code files that have unusual characters in the filename.

-Kevin

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,12 @@ All dates in this document are approximate.
 
 ## Changes
 
+20241201: In some cases Klipper may have ignored leading characters or
+spaces in a traditional G-Code command. For example, "99M123" may have
+been interpreted as "M123" and "M 321" may have been interpreted as
+"M321". Klipper will now report these cases with an "Unknown command"
+warning.
+
 20241112: Option `CHIPS=<chip_name>` in `TEST_RESONANCES` and
 `SHAPER_CALIBRATE` requires specifying the full name(s) of the accel
 chip(s). For example, `adxl345 rpi` instead of short name - `rpi`.
--- a/klippy/gcode.py
+++ b/klippy/gcode.py
@@ -198,16 +198,14 @@ class GCodeDispatch:
                 line = line[:cpos]
             # Break line into parts and determine command
             parts = self.args_r.split(line.upper())
-            numparts = len(parts)
-            cmd = ""
-            if numparts >= 3 and parts[1] != 'N':
-                cmd = parts[1] + parts[2].strip()
-            elif numparts >= 5 and parts[1] == 'N':
+            if ''.join(parts[:2]) == 'N':
                 # Skip line number at start of command
-                cmd = parts[3] + parts[4].strip()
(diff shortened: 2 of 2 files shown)


meteyou:
Thx @KevinOConnor! I tested this PR with the "start print" command via webinterface (the origin issue) and it works fine with your fix:

image

pedrolamas:
Just ran a bunch of tests on this, worked perfectly with whatever strings I threw at it, even with escaped characters!

Thanks Kevin!

ealin1994:
I plan to use params to transfer the hex color code in gcode file to klipper, and then convert it into RGB signal to drive ws2812 to the color of consumables, but the HEX color code starts with #, which will be recognized by klipper as a comment and report an error. Is there any way to solve this problem?以上翻译结果来自有道神经网络翻译（YNMT）· 通用场景

ealin1994:
I intend to utilize params to transfer the hex color code in the gcode file to klipper and subsequently convert it into RGB signal to drive ws2812 to the color of filament. However, the HEX color code begins with '#', which will be regarded by klipper as a comment and result in an error. Is there any solution to this issue?

pedrolamas:
> Is there any solution to this issue?

Until the above code gets merged, your best option is to send the hex code without the `#` or send each of the RGB values separately in decimal form.

ealin1994:
> > Is there any solution to this issue?
> 
> Until the above code gets merged, your best option is to send the hex code without the `#` or send each of the RGB values separately in decimal form.

After employing this issue, there is still no means to solve my problem. Of course, I would prefer to use HEX code without #, but the slicer lacks such an argument. Here is my gcode_mcaro. In order for klipper to be capable of reading the filament color of the slicer, set the RGB indicator color for the corresponding color channel.

ealin1994:
> > Is there any solution to this issue?
> 
> Until the above code gets merged, your best option is to send the hex code without the `#` or send each of the RGB values separately in decimal form.

After employing this issue, there is still no means to solve my problem. Of course, I would prefer to use HEX code without #, but the slicer lacks such an argument. Here is my gcode_mcaro. In order for klipper to be capable of reading the filament color of the slicer, set the RGB indicator color for the corresponding color channel.

[gcode_macro testRGB]
gcode:
    {% set hex1 = params.HEX %}
    {% set hex = hex1[1:] %}
    {% set r_str = hex[0] + hex[1] %}
    {% set g_str = hex[2] + hex[3] %}
    {% set b_str = hex[4] + hex[5] %}
    {% set r = r_str|int(base=16) %}
    {% set g = g_str|int(base=16) %}
    {% set b = b_str|int(base=16) %}
    {% set rr_str = "%.2f"|format(r/255) %}
    {% set gg_str = "%.2f"|format(g/255) %}
    {% set bb_str = "%.2f"|format(b/255) %}
    
    SET_LED LED=sb_leds GREEN={gg_str} RED={rr_str} BLUE={bb_str}

PR #6756, opened 2024-12-02
Add support for ICM20948 accelerometer

phansel:
By popular request, adding support for the non-obsolete ICM20948 accelerometer by porting the MPU9250 codebase. There are several register changes which obstruct integration with the MPU9250 driver, e.g. automatically detecting that a chip is ICM20948 before initializing it.

Seems to work with an RP2040 Pico:
```
[mcu pico]
serial: /dev/serial/by-id/usb-Klipper_rp2040_abcdef0123-if00

[icm20948]
i2c_mcu: pico
i2c_bus: i2c0a

[resonance_tester]
accel_chip: icm20948
probe_points:
        100, 100, 20

[static_digital_output pico_3v3pwm]
pins: pico:gpio23
```
```
accelerometer values (x, y, z): 29103.915381, 15275.006592, -21317.971582
--------------
ACCELEROMETER_QUERY
```

I'm not sure if these values are scaled correctly. Are they?

Signed-off-by: Paul Hansel github@paulhansel.com


JamesH1978:

Thanks
James

KevinOConnor:
We can certainly add support for a new accelerometer, but it should be done by refactoring the existing mpu9250 code, not duplicating it.  (Or, if no code changes are needed, by updating the documentation to indicate that the icm20948 is supported by the mpu9250 config.)

Thanks,
-Kevin

phansel:
Hi @KevinOConnor,

Thanks for taking a look at the PR.

> We can certainly add support for a new accelerometer, but it should be done by refactoring the existing mpu9250 code, not duplicating it. (Or, if no code changes are needed, by updating the documentation to indicate that the icm20948 is supported by the mpu9250 config.)
> 
> Thanks, -Kevin

Around 90% of the register addresses and values are different, so ICM20948 is not supported by the MPU9250 config. The sample frequency is also different: 4.5 kHz instead of 4 kHz.

- Functions achieved by one register on MPU9250 are set by multiple registers on ICM20948
- Functions configured by two registers on MPU9250 are set by one on ICM20948
- FIFO format and I2C address is the same

There are two ways this could be addressed on the front-end:
- A: mpu9250.py and sensor_mpu9250.c silently detect that a particular sensor is an ICM and not an MPU, set registers correctly
- B: the user simply enters icm20948 in printer.cfg instead of mpu9250 and the duplicated code is retained.

The branch as of [020f250](https://github.com/Klipper3d/klipper/pull/6756/commits/020f2503cbb2c2b671243767a1024e26755fa967) does work, at least for the ICM20948. I don't have an MPU9250 to test.

I took a poke at making sensor_mpu9250.c and mpu9250.py work separately to automatically detect ICM20948 based on the WHO_AM_I register:[a78b650](https://github.com/Klipper3d/klipper/pull/6756/commits/a78b650ef09ff009badf9d5b4c7c2d54aff829ac)

However, something is lacking in my understanding of how I2C is accessed through klipper. I'll concede that I'm stuck.

```
  File "/home/prusa/klipper/klippy/klippy.py", line 130, in _connect
    self._read_config()
  File "/home/prusa/klipper/klippy/klippy.py", line 123, in _read_config
    self.load_object(config, section_config.get_name(), None)
  File "/home/prusa/klipper/klippy/klippy.py", line 112, in load_object
    self.objects[section] = init_func(config.getsection(section))
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/prusa/klipper/klippy/extras/mpu9250.py", line 258, in load_config
    return MPU9250(config)
           ^^^^^^^^^^^^^^^
  File "/home/prusa/klipper/klippy/extras/mpu9250.py", line 93, in __init__
    dev_id_icm = self.read_reg(REG_DEVID_ICM20948)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/prusa/klipper/klippy/extras/mpu9250.py", line 122, in read_reg
    params = self.i2c.i2c_read([reg], 1)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/prusa/klipper/klippy/extras/bus.py", line 196, in i2c_read
    return self.i2c_read_cmd.send([self.oid, write, read_len])
           ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'send'
```

KevinOConnor:
Ah, I thought your earlier comment was indicating that this chip was mostly compatible with the mpu9250.  If they are totally different chips with different protocols then it would not make sense to use the same code for both.

Cheers,
-Kevin

phansel:
I see - I should have been more precise! It was a copy paste but one followed by quite a few breaking changes. I'd agree that the incompatibility justifies a new file. Paul On Dec 2, 2024, at 4:11 PM, KevinOConnor ***@***.***> wrote:﻿
Ah, I thought your earlier comment was indicating that this chip was mostly compatible with the mpu9250.  If they are totally different chips with different protocols then it would not make sense to use the same code for both.
Cheers,
-Kevin

Wulfsta:
@phansel you might want to take a look at my recent changes in #6715 - the situation is similar there. There is a lot of difference between the registers of these chips, but many of the functions are similar or the same... I am not sure if it is truly desirable to try to consolidate the code in this case, but is probably doable?

phansel:
@Wulfsta Thanks for the comments! 

In the specific case of the ICM20948 vs. MPUxxxx, the max sample rate isn't even the same, so a few sequence-dependent components of the init are broken by having to probe the I2C to identify the device. To make matters worse, the DEVICE_ID register doesn't even have the same address.

Someone more familiar with the klipper codebase could probably figure that first issue out, but the net savings are only a few hundred lines of Python.

Since these drivers generally will only be used once or twice, and there are only so many accelerometers, I generally prefer to keep them in separate files to (a) reduce risk of breaking behavior of Chip A by modifying Chip B driver and (b) improve readability, since a bunch of switches in the middle of the driver can be quite confusing.

It's ultimately up to the owners of the repo. I'm happy to help but don't have access to an MPU9250 for regression testing and don't have unlimited time to figure out how klippy/extras/bus.py works.

Wulfsta:
The only reason I see to consolidate these is if we are concerned about binary size - which should not be an issue anyways, since there is a mechanism in place to optionally include sensors for chips that are sensitive to this. As I said, I am not sure it is desirable to consolidate in this case, I just wanted to provide another resource to consider. :slightly_smiling_face:

phansel:
@KevinOConnor I've updated this PR to match the new CONFIG_WANT_[sensor] format.

klippy/extras/icm20948.py
```diff
@@ -0,0 +1,175 @@
...
+                "(e.g. faulty wiring) or a faulty chip."
+                % (dev_id))
+        else:
+            logging.info("Found %s with id %x"% (ICM_DEV_IDS[dev_id], dev_id))
+
+        # Setup chip in requested query rate
+        self.set_reg(REG_PWR_MGMT_1, SET_PWR_MGMT_1_WAKE)
+        self.set_reg(REG_PWR_MGMT_2, SET_PWR_MGMT_2_ACCEL_ON)
+        # Add 20ms pause for accelerometer chip wake up
+        self.read_reg(REG_DEVID) # Dummy read to ensure queues flushed
```
nefelim4ag, line 132: there are `i2c_write_wait_ack` which can be used above
Which ensures that that request is acknowledged by the device

klippy/extras/icm20948.py
```diff
@@ -0,0 +1,175 @@
...
+        systime = self.printer.get_reactor().monotonic()
+        next_time = self.mcu.estimated_print_time(systime) + 0.020
+        self.set_reg(REG_ACCEL_SMPLRT_DIV1, SAMPLE_RATE_DIVS[self.data_rate])
+        self.set_reg(REG_ACCEL_SMPLRT_DIV2, SAMPLE_RATE_DIVS[self.data_rate],
```
nefelim4ag, line 136: Those will reorder writes, cause your next set_reg will execute with minclock=0
I may be wrong about exact numbers but in general, such print time conversion will send it in the future by 200ms.

klippy/extras/icm20948.py
phansel: I'll swap that in. Thanks for the suggestion @nefelim4ag

klippy/extras/icm20948.py
```diff
@@ -0,0 +1,175 @@
...
+        self.axes_map = adxl345.read_axes_map(config, SCALE, SCALE, SCALE)
+        self.data_rate = config.getint('rate', 4500)
+        if self.data_rate not in SAMPLE_RATE_DIVS:
+            raise config.error("Invalid rate parameter: %d" % (self.data_rate,))
+        # Setup mcu sensor_icm20948 bulk query code
+        self.i2c = bus.MCU_I2C_from_config(config,
+                                           default_addr=ICM20948_ADDR,
+                                           default_speed=400000)
+        self.mcu = mcu = self.i2c.get_mcu()
+        self.oid = oid = mcu.create_oid()
```
nefelim4ag, line 69: `oid` is not used anywhere

klippy/extras/icm20948.py
nefelim4ag: Yes, I already found that, but this is your code now :D 
I just suggest you not mindlessly copy things
minclock=0/reqclock=0 is the default behavior anyway

With wait_ack() you can just add a pause there, before write, or order other writes by minclock
If you merge coupled registers write, that will also make things cleaner, like:
```
SAMPLE_RATE_DIVS = { 4500: [0x00, 0x0] }
i2c_write_wait_ack([REG_ACCEL_SMPLRT_DIV1] + SAMPLE_RATE_DIVS[self.data_rate])
# or
set_reg(REG_ACCEL_SMPLRT_DIV1, SAMPLE_RATE_DIVS[self.data_rate])
```

nefelim4ag: According to the datasheet, there is a 20ms delay between the chip wake and accelerometer ready (how ready, what does it mean?) but looks like all registers can be just written

KevinOConnor: > Those will reorder writes, cause your next set_reg will execute with minclock=0

Commands are never reordered within a cmdqueue (and all the i2c commands will be on the same cmdqueue).  The `minclock` and `reqclock` control the timing of when messages are released from the command queue, but they never reorder that queue.

There are a few ways to enforce a delay between i2c writes.  One way is to do what mpu9250 does:
```
        # Add 20ms pause for accelerometer chip wake up
        self.read_reg(REG_DEVID) # Dummy read to ensure queues flushed
        systime = self.printer.get_reactor().monotonic()
        next_time = self.mcu.estimated_print_time(systime) + 0.020
        self.set_reg(REG_SMPLRT_DIV, SAMPLE_RATE_DIVS[self.data_rate],
                     minclock=self.mcu.print_time_to_clock(next_time))
```
This forces an i2c read so that we know previous commands have been completed at the micro-controller (that is, nothing in the command queue).  Then we schedule the next i2c message to have a minimum clock that is 20ms in the future (from the current time as measured on the host).  Note that the `minclock` is on the very next write.

Another mechanism would be something like:
```
        # Add 20ms pause for accelerometer chip wake up
        self.read_reg(REG_DEVID) # Dummy read to ensure queues flushed
        reactor = self.printer.get_reactor()
        reactor.pause(reactor.monotonic() + 0.020)
        self.set_reg(REG_SMPLRT_DIV, SAMPLE_RATE_DIVS[self.data_rate])
        ...
```
This differs from the above in that we pause the host instead of just scheduling the next message at a particular time.  (Both will enforce at least 20ms of delay between i2c messages.)

As mentioned, it's possible to eliminate the dummy read by using `i2c_write_ack()` but that seems like more trouble than it is worth.  This is an infrequently run "slow path" for sensor initialization.

Of course, if no delay is actually needed, that is preferable.

Hope that helps,
-Kevin

KevinOConnor:
Thanks.  In general it seems fine to me.  I only noticed the unusual pause, as discussed above.

Cheers,
-Kevin

klippy/extras/icm20948.py
phansel: I'm not too familiar with this codebase, but I don't recall it working without line 69.

nefelim4ag: `self.oid = mcu.create_oid()`

phansel:
@KevinOConnor @nefelim4ag I've tested the requested changes ([dc190b6](https://github.com/Klipper3d/klipper/pull/6756/commits/dc190b62fdfc33992c42f7b179b433761cff4f2e)) as working with ICM20948/RP2040. The new resonance test features were a nice surprise.

Let me know your preference for resolving conflicts. Happy to squash this down to a single commit off the current master on a new PR.

shaper_calibrate_x
shaper_calibrate_y

nefelim4ag (approved):
Squash and LGTM

I've noticed your:
> accelerometer values (x, y, z): 29103.915381, 15275.006592, -21317.971582

If it is still in that range, it would be nice to look at scaling, but it does not affect shaper frequency, only scale.
Normally, there is around 9800 for Z and noise level for the x/y axes.

According to the datasheet, the current values look correct.
0x04 (0b100) written to 0x14 is +- 8g scale.
So
```
0xffff/16 = 4095.9375
1 / 4096 = 0.000244140625
```

Thanks.

KevinOConnor:
Thanks.

-Kevin

PR #6757, opened 2024-12-02
Correct maximum TMC2240 UART address.

liampwll:
There are 3 address pins on this device, allowing a maximum address of 7.


KevinOConnor:
Thanks.

-Kevin

PR #6758, opened 2024-12-02
Support white neopixel strings

oresk:
Support white neopixel strings

This change allows white single color neopixel cob strips to work like this one https://vi.aliexpress.com/item/1005005946691268.html?spm=a2g0o.order_list.order_list_main.10.57261802SVIh97&gatewayAdapt=glo2vnm 

Tested it on my setup, and it works as expected. 

I have updated the relevant documentation. 

The new check allows you also to enter color_order of one or two characters. This allows to enter just W and then have all the white subparts adresable on their own. 

Signed-off-by: Lovro Oreskovic <lovro@oreskovic.me>

### DIFF
--- a/klippy/extras/neopixel.py
+++ b/klippy/extras/neopixel.py
@@ -34,12 +34,7 @@ class PrinterNeoPixel:
             raise config.error("color_order does not match chain_count")
         color_indexes = []
         for lidx, co in enumerate(color_order):
-            if sorted(co) not in (
-                sorted("RGB"),
-                sorted("RGBW"),
-                sorted("WWWW"),
-                sorted("WWW"),
-            ):
+            if any(c not in "RGBW" for c in co):
                 raise config.error("Invalid color_order '%s'" % (co,))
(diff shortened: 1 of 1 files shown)


klippy/extras/neopixel.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, config):
             raise config.error("color_order does not match chain_count")
         color_indexes = []
         for lidx, co in enumerate(color_order):
-            if sorted(co) not in (sorted("RGB"), sorted("RGBW")):
+            if any(c not in "RGBW" for c in co):
```
flowerysong, line 37: This validation is insufficient, it allows strings like `RRRRR`.

flowerysong: I'm not sure why you're bringing up single colors. The best approach would be to perform sufficient validation so that strings like `RRRRRRRRRRRRRRRRRRGBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB` aren't accepted.

PR #6759, opened 2024-12-04
Add gcode_shell_command.py

master3218:


JamesH1978:
This module will likely not be included in mainline for the reasons given in the KIAUH documentation

**Warning**
This extension may have a high potential for abuse if not used carefully! Also, depending on the command you execute, high system loads may occur and can cause system instabilities. Use this extension at your own risk and only if you know what you are doing!

Thanks
James

PR #6760, opened 2024-12-06
lib: fix rp2040_flash

bigtreetech:
`model_t` is a enum value [here](https://github.com/Klipper3d/klipper/blob/master/lib/rp2040_flash/picoboot_connection.h#L44)

It should be `rp2040` (actual value is 0, `rp2350` is 1, and `unknown` is 2)


KevinOConnor:
Thanks.

-Kevin

PR #6762, opened 2024-12-07
stm32: Add SPI6 support for PB3-PB5 on STM32H723

aam335:
The STM32H723 hardware SPI on pins PB3-PB5 supports alternate functions for SPI1, SPI3, and SPI6. Currently, these pins work only with SPI1 (spi1a config) and cannot be used with the hardware driver since SPI1 is already utilized by the onboard MAX PT100 amplifier (on PA5-7 pins).

This commit introduces SPI6 hardware support for PB3-PB5 using the spi6a configuration.

Signed-off-by: Andrey Melnikov Aam335@gmail.com


nefelim4ag:
image

I tried to test it with Octopus PRO v1.1 with STM32H723.
And it does not work, also I tested the same pins in the same order but in software mode (it works):
```
spi_software_mosi_pin: PB4
spi_software_miso_pin: PB3
spi_software_sclk_pin: PB5
```
Maybe you set them in the wrong order?

PR #6763, opened 2024-12-08
Allow more control of compiled sensors and displays and features

Links2004:
This Helps to get the size down e.g. for usage with atmega328p.


Links2004:
the builds fail e.g. with:
```
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: address 0x800945 of out/klipper.elf section `.bss' is not within region `data'
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: address 0x800945 of out/klipper.elf section `.bss' is not within region `data'
collect2: error: ld returned 1 exit status
make: *** [Makefile:73: out/klipper.elf] Error 1
```

Links2004:
added Signed-off-by to commit

KevinOConnor:
Thanks.  In general it seems fine to me.  There were some recent changes to the Kconfig file (doing something similar) so this PR will need to be rebased to the latest code.

Also, at a quick glance it seems this patch misses the conversion of `HAVE_GPIO_I2C` to `WANT_GPIO_I2C` in a few places.  I'm not sure why the conversion was made for WANT_GPIO_SPI but not WANT_GPIO_I2C.

Cheers,
-Kevin

Links2004:
rebased and reworked the selection, new menu:

KevinOConnor:
Thanks.  In general it seems fine.  However, I don't think we should organize the menu by "support for i2c" and "support for spi".  I think we should keep the accelerometers together, the high speed adcs together, the displays together, etc. .  That is, I think most users will find an orientation by chip type (not chip interface) to be more convenient.

Also, it looks like you introduce a duplicate `WANT_HX71X`.

Also, it doesn't make sense to add `CONFIG_WANT_xxx=y` definitions to atmega328p.config as these already default to 'y' - it only makes sense to add explicit 'n' definitions.

Cheers,
-Kevin

Links2004:
the problme with this is that the chips require I2C / SPI which basicly forces you to decide on that first.
but I dont think that there will by and problem finding the chips since the default for I2C / SPI is y (if the platform / MCU supports it), which result in the full visable list of chips.

if we sort it diffently its hard to under stand why options are simingly randomly there and not.

KevinOConnor:
Thanks.  I agree that sorting by sensor type could be confusing when spi/i2c are disabled.  However, I think sorting by spi/i2c could also cause confusion.  I think I'd prefer to keep the current sorting by sensor type.

-Kevin

KevinOConnor:
FYI, see the similar #6884 

-Kevin

PR #6766, opened 2024-12-18
z_tilt: return done when max retries is 0

krautech:
- Set `printer.z_tilt.applied = True` when max_retries is 0
- and then raises error when `[z_tilt] retry_tolerance = 0` 

[Discourse Link](https://klipper.discourse.group/t/possible-bug-printer-z-tilt-applied-not-true-if-no-retries-in-config/20584/2)

### DIFF
--- a/klippy/extras/z_tilt.py
+++ b/klippy/extras/z_tilt.py
@@ -108,7 +108,7 @@ class RetryHelper:
         return self.increasing > 1
     def check_retry(self, z_positions):
         if self.max_retries == 0:
-            return
+            return "done"
         error = round(max(z_positions) - min(z_positions),6)
         self.gcode.respond_info(
             "Retries: %d/%d %s: %0.6f tolerance: %0.6f" % (


Esoterical:
Tested on my trident, on klipper v0.12.0-396 but with this z_tilt.py file added.

no retries, no tolerance: applied=true
3 retries, no tolerance: error about no tolerance
no retries, 0.1 tolerance: applied=true
3 retries, 0.1 tolerance: retried and applied as normal

So it looks like it fixes the original problem (applied=false after a z_tilt if there was no retries set in config) and also won't let you try multiple retries if the tolerance is accidently not set/set to 0.

KevinOConnor:
Thanks.  Happy to make the `return "done"` change.  However, I'm not sure about the retry_tolerance change - it is technically valid to use a retry_tolerance=0 - it means keep retrying until two results produce identical results.  So, it'd be a bit of a silly setting, but it's not technically an error.  I guess we could change it if we thought it was problematic, but then it would need to be documented as a behavior change in Config_Changes.md .

-Kevin

krautech:
Ahhh I see, I wasn't aware that its behaviour was tied to looking for an equal result.

What would your preference be? Simplely, changing the return to done would be suffice IMO.

KevinOConnor:
Yeah - lets fix the `return "done"` bug as one commit.  Other changes/fixes can be done as separate commits if needed.

Thanks.
-Kevin

krautech:
No worries, have updated the commit 

Cheers,

Chris

KevinOConnor:
Thanks.

-Kevin

PR #6769, opened 2024-12-23
Temperature combined: handle temp monitors and I2C sensors

nefelim4ag:
There are 2 patches:
1. Duct tape fix for race conditions with temperature combined sensor and other sensors during klippy initialization, where sensors can return 0. This may be by design, but it currently limits the minimum possible deviation between sensors.  Like there, it is adc + i2c: image
 
2. Temperature combined allows us to set "any" object in the sensor list, but with tmc2240, there is a corner case, it has a method `get_status()` and `temperature` field, but it has no value. So, it will crash Klippy cause max/min can't handle `None`. 

---
Should close: https://github.com/Klipper3d/klipper/pull/6675

### DIFF
--- a/klippy/extras/temperature_combined.py
+++ b/klippy/extras/temperature_combined.py
@@ -41,14 +41,22 @@ class PrinterSensorCombined:
             sensor = self.printer.lookup_object(sensor_name)
             # check if sensor has get_status function and
             # get_status has a 'temperature' value
-            if (hasattr(sensor, 'get_status') and
-                    'temperature' in sensor.get_status(
-                            self.reactor.monotonic())):
-                self.sensors.append(sensor)
-            else:
+            if not hasattr(sensor, 'get_status'):
+                raise self.printer.config_error(
+                        "'%s' does not have a status."
+                        % (sensor_name,))
+            status = sensor.get_status(self.reactor.monotonic())
+            if 'temperature' not in status:
                 raise self.printer.config_error(
                         "'%s' does not report a temperature."
                         % (sensor_name,))
+            # Handle temperature monitors
+            if status["temperature"] is None:
+                raise self.printer.config_error(
+                        "Temperature monitor '%s' is not supported"
+                        % (sensor_name,))
+
+            self.sensors.append(sensor)
 
     def _handle_ready(self):
         # Start temperature update timer
--- a/klippy/extras/temperature_combined.py
+++ b/klippy/extras/temperature_combined.py
@@ -52,8 +52,10 @@ class PrinterSensorCombined:
 
     def _handle_ready(self):
         # Start temperature update timer
+        # There is a race condition with sensors where they can be not ready,
+        # and return 0 or None - initialize a little bit later.
         self.reactor.update_timer(self.temperature_update_timer,
-                                  self.reactor.NOW)
+                                  self.reactor.monotonic() + 1.)
 
     def setup_minmax(self, min_temp, max_temp):
         self.min_temp = min_temp


KevinOConnor:
Thanks.  In general the first two patches look fine.  The new comments seem a little odd though - I'm not aware of any sensor that will return 0 - we should definitely fix any sensor doing that.

I'm not sure about the third patch.  Why would one use temperture_combined on a tmc2240 temperature?  What would they combine it with?  If the goal is to use a tmc2240 temperature sensor as a regular sensor, maybe we should consider adding an internal mechanism to indicate to the tmc2240 that it always needs to provide temperature (that is, have the temperature code signal the tmc2240 instead of a user config option signalling tmc2240).

-Kevin

nefelim4ag:
> I'm not aware of any sensor that will return 0 - we should definitely fix any sensor doing that.

I think that about the first patch and the delay in initialization.
It will be nice, but all temperature sensors do that, the only exception is ADC ones.
They all initialized with the default value equal to `.0`: 
https://github.com/Klipper3d/klipper/blob/master/klippy/extras/ds18b20.py#L20
(same for BME, HTU, SHT, LM75).

So, get_status which always just returns the last value of `self.temp`, can be called and will return `.0`
Of course, after the first query cycle, they will return the actual value or crash the system.

Maybe there is a more clever solution, like ... do `pause()` in get_status() to wait for initialization in every sensor?
I'm not so familiar with the whole call chain here, to suggest something sane here.
Looks like ADC is always making calls to MCU instead of providing cached property.

---
> I'm not sure about the third patch. Why would one use temperture_combined on a tmc2240 temperature? What would they combine it with?

Maybe I misunderstood the goal of temperature combined and misused it, but I have only 2 use cases, like:
- use avg/max over drivers' temperatures to control temperature FAN (I use only one pin to control fan output to cool drivers)
- Duplicate sensor, because heater object can't have humidity or gas, and I use bme680 as the control for chamber temperature, and ADC + temperature fan to control PTC heater.

_I do not see a reason to use any individual data from the drivers to control anything, AFAIK there is no external TMC2240 like TMC5160, which can be cooled individually._

If there is a way to avoid adding strange config options and do some magic behind the scenes - I agree it is a better solution (like query always if there are consumers of data - it will be a better solution).
_But I can't imagine how to do that without modifying every possible temperature consumer to add some callback for that._ I even had thought about allowing to define something like `temperature_sensor` where type is `tmc2240 stepper_*` which will do something like that - change internal behavior. But it also feels weird a little.

My original intention here is to just allow us to use it somehow at least because right now it is only for graphs.
_Graphs btw look strange to anyone who doesn't know [why](https://github.com/Klipper3d/klipper/pull/6145) there is no data from the driver if I can easily query it from the console._

nefelim4ag:
I took another look at the code, ADC sensors were provided through the heater factory, and in the end have queried through the callback which is protected with a lock.
I'm not sure about global ordering here, there is still a possibility to query get_status before the first callback runs.
But it does not happen.

As this is a callback any structure above it will receive data when they are ready, but combined sensor - query object directly, without a callback, so it can receive data before they are ready and between readings.

I think there are 2.5 possibilities to work with:
1. Migrate combined sensor to callback infrastructure, but it will break its flexibility and possible existing users, like I can define it over a heater, but I don't know if it is valid.
1.5 Above + allows to setup of several callbacks for the sensor to handle that case (I think objects are unique and if there are two callback users one will steal a callback from the other).
2. Introduce a similar mechanism to get_status() for sensors, which will block till the first data is ready (I think that can break initialization of the combined sensor, there is a call to get_status on connect())
2.5 Above + use event time for order, but that will make any users of get_status() lag till fresh data with `query eventtime > get_status eventtime` is available.

For me, it feels like callback is the correct way to go here, cause all internal users should use it, and it will automatically disallow the use of tmc2240, cause there is no setup_callback, and will update its own data on each update of sensors beneath. But it is a lot of changes.

I'm not sure I'm qualified here to choose the correct way to fix this, I will be pleased to hear your opinion on this, maybe I just have a strong misunderstanding there.

nefelim4ag:
I tried to migrate temperature combined to callback infrastructure, but it happens most config objects just do not provide `setup_callback`, maybe for some reason.
It is still possible, but then I should add setup_callbacks to every object that can be used in a combined sensor.

Not sure if it is the right way to fix the combined sensor race condition, cause it still be present till all callbacks happen at least once.

nefelim4ag:
Hi, as there is an upcoming V0.13
I may suggest merging the first 2 commits, as they are pure fixes.

If that makes sense, I can split them to other PR.

Thanks.

KevinOConnor:
Okay, thanks.  If you want to separate out the 2 changes I'll go ahead and commit.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6771, opened 2024-12-28
Expose slope_control field on TMC2240.

liampwll:
The chip has a default value of 0, corresponding to 100V/µs. The default is overridden with 2 here, corresponding to 400V/µs to approximately match the TMC2209. This should also solve some overheating issues as it lowers the power dissipation at a 50kHz chopper frequency by around 1W.

This is currently untested.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3846,6 +3846,7 @@ run_current:
 #driver_SEIMIN: 0
 #driver_SFILT: 0
 #driver_SG4_ANGLE_OFFSET: 1
+#driver_SLOPE_CONTROL: 2
 #   Set the given register during the configuration of the TMC2240
 #   chip. This may be used to set custom motor parameters. The
 #   defaults for each parameter are next to the parameter name in the
--- a/klippy/extras/tmc2240.py
+++ b/klippy/extras/tmc2240.py
@@ -408,6 +408,8 @@ class TMC2240:
         set_config_field(config, "tpowerdown", 10)
         #   SG4_THRS
         set_config_field(config, "sg4_angle_offset", 1)
+        #   DRV_CONF
+        set_config_field(config, "slope_control", 2)
 
 def load_config_prefix(config):
     return TMC2240(config)


Sineos:
Would it make sense to expose `DRV_CONF` register bits 1..0 as well. This seems to be a nice way to tune the driver closer to the actual motor current (see data sheet, page 47). 
I'm currently using 01 instead of the default 11 on my z-axis and at least have the feeling that it somewhat reduced z-banding (could be wishful thinking, but at least no negative effect observed).

liampwll:
It probably makes more sense for Klipper to do that internally based off the current specified by the user, but it wouldn't hurt to have an override that the user can specify.

nefelim4ag:
Slope control defines how fast the MOSFETs will be switched.
*Didn't know that, so it is for someone who will also struggle with that.

For now, I can confirm the information about power dissipation:
image
The green line is fan PWM. 0->2 looks like a lot of difference.

nefelim4ag:
> Would it make sense to expose `DRV_CONF` register bits 1..0 as well. This seems to be a nice way to tune the driver closer to the actual motor current (see data sheet, page 47). I'm currently using 01 instead of the default 11 on my z-axis and at least have the feeling that it somewhat reduced z-banding (could be wishful thinking, but at least no negative effect observed).

image
2A. next one 1A
image

It is calculated and set automatically, based on your desired current.

wight554:
Have to check regressions, but did speed and resonance tests so far on mellow drivers and results are amazing.
I've got temp drop from from 100-110C to 70-75C on 1.414A.
Was able to raise current to 2.5A motors RMS (~1.77A) with temp at 85-90C level vs overtemp error before the change.

Have to mention that I use LH Stinger Ebox with 2 8025 fans blowing on all electronics inside and TMC Autotune, it has different change to drop temps on 2240, by dropping freqs iirc, so idk if these play well together

liampwll:
The lower frequency in the TMC Autotune module shouldn't conflict with
this, but I suspect it should also no longer be required with this change
for most users.

KevinOConnor:
Thanks for working on this.  We can certainly add a new `driver_SLOPE_CONTROL` config option.

However, I'd be leery about setting a default that differs from the Trinamic default, unless we can find some strongly worded statement from Trinamic that indicates the setting should be changed.  That is, I'm unsure we'll be able to select a better default that Trinamic can.

If we can find a strong indication from Trinamic that the default should be different, then we'll need to document the change in docs/Config_Changes.md .

-Kevin

liampwll:
The only things that are impacted by this setting outside of the switching losses should be the EMI and power supply ripple. As noted earlier, setting this to 400V/µs matches the typical value of the TMC2209 that is widely used on stepsticks on DIY boards (see attached image), so I doubt that will cause any issues for users of DIY boards. It might be an issue for any non-DIY vendors that have gone through proper EMI testing or are pushing what the caps on their boards can handle, but if they're just blindly applying updates then that's their own problem.

I think the fact that no one has made a PR to expose this setting before now makes a good case for making this the default as it shows that most users will not read the datasheet to investigate why their drivers are overheating and will instead reduce the current or reduce the chopper frequency, which can both decrease performance.

image

KevinOConnor:
Thanks.  I made a change to the Klipper code to make it possible to alter this setting (commit cf3b0475).  However, I have not changed the default (the default remains zero as Trinamic documents the default and as Klipper previously had as a default).

-Kevin

liampwll:
I think there should at least be a message presented to the user as there's little reason to prefer a lower value, but I'm not sure what the best way to go about that is.

PR #6772, opened 2024-12-31
docs: page about python versions

Laikulo:
Based on the information in this discourse topic: https://klipper.discourse.group/t/discuss-python-versions/20942


KevinOConnor:
Thanks.  In general it seems fine to me.  New documentation files should also be added to Overview.md (mentioned briefly in point 5 at https://www.klipper3d.org/CONTRIBUTING.html#what-to-expect-in-a-review ).

I don't think the official documentation should link to discussion on discourse, as the discussions there tend to be ephemeral.

Cheers,
-Kevin

PR #6774, opened 2025-01-01
probe_as_z_home: Initial implementation and documentation

nickweedon:
This pull request adds the ability to perform multiple probes when issuing a G28 command. More specifically it calls into the 'probe' logic to execute a probe in a way similar to what would be performed during bed mesh creation. Since it uses the same logic this means that the resulting probe also applies the same ```calc_probe_z_average``` method to the result.

The benefit of this feature is it allows users with less than perfect probing hardware to still have an unsupervised and reliable homing experience which is crucial to ensure print quality and preventing failed prints due to an incorrect Z homing. Bad homing can even lead to extruder jams and waste significant amounts of the 3D printer user's time. This is a feature that has been requested many times with one such example being [Discourse: multiple-samples-on-home-z](https://klipper.discourse.group/t/multiple-samples-on-home-z/11864).

To configure this tool a new configuration section called ```[probe_as_z_home]``` should be placed anywhere in the config before the either ```[safe_z_homing]``` or ```[homing_override]```.

The implementation is completely decoupled from any particular homing or probing method and does not modify any existing code.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2095,7 +2095,7 @@ sensor_type: ldc1612
 
 This tool allows the homing procedure (```G28```) to use probing (i.e. the same as the ```PROBE``` command) so that in some cases the Z axis can be more accurately probed.
 
-You should only use this option if you are occasionally experiencing innacurate homing.
+You should use this option if you are occasionally experiencing innacurate homing.
 
 Configuring this tool will also cause the ```G28``` G-Code to emit a console message stating the correction that was performed. You can use this to measure the level of benefit that the tool is providing.
 
@@ -2107,6 +2107,11 @@ See [command reference](G-Codes.md#G-Code_commands) for further information.
 
 ```
 [probe_as_z_home]
+#g28_probe_cmd_args:
+#   Optional command arguments passed to the PROBE command while homing.
+#   See [g-code PROBE command](G-Codes.md#probe).
+#   Example:
+#      g28_probe_cmd_args: SAMPLES=3 SAMPLES_TOLERANCE_RETRIES=5
 ```
 
 _Note: The standing homing procedure is still perofrmed before probing since this is required to ensure safety and compatibility with the probing method since probing requires homing to first be performed_
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -2091,6 +2091,26 @@ sensor_type: ldc1612
 #   See the "probe" section for information on these parameters.
 ```
 
+### [probe_as_z_home]
+
+This tool allows the homing procedure (```G28```) to use probing (i.e. the same as the ```PROBE``` command) so that in some cases the Z axis can be more accurately probed.
+
+You should only use this option if you are occasionally experiencing innacurate homing.
+
+Configuring this tool will also cause the ```G28``` G-Code to emit a console message stating the correction that was performed. You can use this to measure the level of benefit that the tool is providing.
+
+To enable this tool, add the following configuration section anywhere before either ```[safe_z_homing]``` section or the ```[homing_override]``` section.
+
+There is no specific G-Code for this command since it simply alters the behavior of the
+existing standard 'G28' G-Code command.
+See [command reference](G-Codes.md#G-Code_commands) for further information.
+
+```
+[probe_as_z_home]
+```
+
+_Note: The standing homing procedure is still perofrmed before probing since this is required to ensure safety and compatibility with the probing method since probing requires homing to first be performed_
+
 ### [axis_twist_compensation]
 
 A tool to compensate for inaccurate probe readings due to twist in X or Y
--- a/klippy/extras/probe_as_z_home.py
+++ b/klippy/extras/probe_as_z_home.py
@@ -17,6 +17,7 @@ class ProbeAsZHome:
         self.printer.load_object(config, 'homing')
         self.prev_G28 = self.gcode.register_command("G28", None)
         self.gcode.register_command("G28", self.cmd_G28)
+        self.g28_probe_cmd_args = config.get("g28_probe_cmd_args", None)
 
     def cmd_G28(self, gcmd):
         logging.debug("Probe As Home: Performing G28")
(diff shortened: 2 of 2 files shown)


vgdh:
great!

AchilleBailly:
This seems really nice ! I tested it and it works great indeed using the probe config.
However, passing the usual PROBE params to G28 does not work, it would be nice to do so, mainly to be able to do let's say 10 samples for homing but keep 1 or 2 samples when doing the bed mesh.
Do you think that would be possible ?

nickweedon:
> This seems really nice ! I tested it and it works great indeed using the probe config. However, passing the usual PROBE params to G28 does not work, it would be nice to do so, mainly to be able to do let's say 10 samples for homing but keep 1 or 2 samples when doing the bed mesh. Do you think that would be possible ?

@AchilleBailly Thanks for checking this out! It is definately possible i believe and in fact that had been the intended behavior, to "pass through" the parameters. Given this context I see this as a bug. 

Thanks for pointing this out, i will look at fixing this this evening when i get home from work. I will give you a heads up once I have done this.

nickweedon:
OK so now that I have taken a look at the code again, I actually take that back about this being a bug lol. When you said "passing the usual PROBE params to G28" I was a bit too busy to think clearly and misinterpreted this as meaning that you can usually pass through the PROBE params to G28 which of course is actually not standard behavior.

Do I think this would be a useful change? Absolutely, I think it would be very useful as it would allow you to do a 'quick and dirty home' with just one probe if you just wanted to do a filament change for example and didn't care about precision.

However...
While its totally possbile and even easy to make this change it would mean that I would actually need to change some existing code both in the 'Probe' and 'BLTouch' modules in order pass these parameters through. Since currently this change is very clean and does not modify any existing code, I would like to keep it this way for now. I also don't want to add too much scope to this initial change.

If this PR gets approved however then I will create another PR with the change that you suggest :)

Sorry I can't implement this right away but thanks for the great idea!

AchilleBailly:
Okay I understand you don't wanna have scope creep ahah

Though I quickly tested locally and it seems I can just pass arguments to the probe command this way (l.66-68 of `probe_as_z_home.py`):
```
 gcmd = self.gcode.create_gcode_command("PROBE", "PROBE", {"SAMPLES": self.sample_count})
 probe_session = probe.start_probe_session(gcmd)
 probe_session.run_probe(gcmd)
 ```
 with `self.sample_count` retrieved from the config.

nickweedon:
@AchilleBailly Nice!
This option makes sense to me so I took your idea and added it to the PR.
I instead implemented this as a optional field in the [probe_as_z_home] config section called 'g28_probe_cmd_args' which allows us to pass additional arguments to the PROBE command.
FYI, I chose to go this route as opposed to adding the same fields in the [probe] section to the [probe_as_z_home] for these reasons:
1. Hard coding these fields would mean that they would need to be updated whenever the [probe] or [bltouch] code changes and adds a new field.
2. Passing these through dynamically does not appear possible the code base will detect unused fields and it would not be possible to validate these fields without refactoring [probe], [bltouch] and any other or future code.
3. Using the 'g28_probe_cmd_args' field is slightly more intuitive IMO since it makes it clear to the use that this takes affect at run time and will not be validated as startup.

Let me know your thoughts though and thanks again for the idea!

nickweedon:
@AchilleBailly As an aside, I am curious about why you choose to use a different sample count when homing? For me it makes sense to keep this the same as when doing other probing operations.

vgdh:
> @AchilleBailly As an aside, I am curious about why you choose to use a different sample count when homing? For me it makes sense to keep this the same as when doing other probing operations.

For example I do first home only probing once and when bed heated enough i do final probing 5 times.

vgdh:
> > > @AchilleBailly As an aside, I am curious about why you choose to use a different sample count when homing? For me it makes sense to keep this the same as when doing other probing operations.
> > 
> > 
> > For example I do first home probing only once and when bed heated enough i do final probing 5 times.
> 
> That makes sense. I figured there would be use cases like this. I am also curious to know about @AchilleBailly 's use case where he wants to statically configure the probe to act differently during G28 (i..e. via the config, such as what is now possible via the more recent change i made to this PR)
> 
> @vgdh Hopefully if this change gets merged in then i will add a new PR to add that feature to be able to specify PROBE parameters via the G28 command. I wonder in the meantime if its possible to do similar via a some kind of jenky macro now with the new `g28_probe_cmd_args` configuration field I added?

I write some crazy macro do to this functionality, that was not a simple thing 😁.
So it would be nice to have this from the box.

nickweedon:
@vgdh & @AchilleBailly, yall might also like my other changes: #6778 and #6780.
Combining these with this change i managed to turn my potato 3D printer into a reliable "fire & forget" unit that prints perfect first layers every time.

KevinOConnor:
Thanks, and sorry for the delay.

If I understand correctly, this new module overrides the typical `G28` homing sequence with a new one that also calls `PROBE`.  It some ways it is similar to `[safe_z_home]` and in some ways it is similar to an elaborate `G28` macro that uses `PROBE` and `SET_KINEMATIC_POSITION`.  It's an interesting approach and I agree it looks useful.

There are a few other development efforts going on in this area of the code right now - "generic_cartesian" support, some homing/probing refactoring that I've been looking at, and a generic "klipper extension" system.  So, I agree this looks useful, but I'm inclined to wait a little longer on this PR to see how those efforts progress first.

Thanks again,
-Kevin

nickweedon:
> Thanks, and sorry for the delay.
> 
> If I understand correctly, this new module overrides the typical `G28` homing sequence with a new one that also calls `PROBE`. It some ways it is similar to `[safe_z_home]` and in some ways it is similar to an elaborate `G28` macro that uses `PROBE` and `SET_KINEMATIC_POSITION`. It's an interesting approach and I agree it looks useful.
> 
> There are a few other development efforts going on in this area of the code right now - "generic_cartesian" support, some homing/probing refactoring that I've been looking at, and a generic "klipper extension" system. So, I agree this looks useful, but I'm inclined to wait a little longer on this PR to see how those efforts progress first.
> 
> Thanks again, -Kevin

Thanks Kevin,
That makes total sense, I agree it is better to wait for the ongoing efforts in this area to complete. I'm always happy to refactor this afterwards if need be. Alternatively, if this is no longer needed at that point then even better :)

Felix14-v2:
Just tested, works perfectly!

The only feature I miss is a separate command to automatically probe and apply the coordinates without issuing G28 in macros. I use this in my macro that uses PROBE to detect inductive probe temperature drift and continue printing when it's stabilized, and the only way (that I know) of applying the new coordinate without modifying offsets is G28, so I have to perform unnecessary procedure instead of calling your solution directly.

nefelim4ag:
Aha, so it is not directly related, but related: https://github.com/Klipper3d/klipper/pull/7268

So, technically, homing now happens with `PROBE`, but the sample count is internally overwritten (for different reasons).
The main problem which are attempt to be solved here is to use the average/median of several samples, so it is still not covered.

-Timofey

Felix14-v2:
Found a parity problem that may be critical to someone – after performing G28, Z coordinate is displayed as negative. Let's say your z_offset is 2.5 mm, and you started a print: Fliudd and printer monitor will display Z=-2.5 at the first layer (Gcode z=0). This can confuse end users and lead to attempts to "fix" z_offset (which is already correct, just displayed incorrectly).

Removing `[probe_as_z_home]` returns the default homing behavior and fixes the issue.

PR #6779, opened 2025-01-05
probe: Fixed premature firing of 'homing:homing_move_begin'

nickweedon:
When probing multiple points such as when calibrating a bed mesh, the ProbePointsHelper start_probe method initiates a horizontal repositioning move and then immediately begins to probe without waiting for the repositioning move to complete. This causes the 'homing:homing_move_begin' to trigger before the Z direction move has even began. This is problematic when running BED_MESH_CALIBRATE with [homing_heaters] configured since it means that the bed will never attempt to re-heat.

This fix addresses this issue simply by waiting for the moves to complete before begining the actual probe.

### DIFF
--- a/klippy/extras/probe.py
+++ b/klippy/extras/probe.py
@@ -463,6 +463,10 @@ class ProbePointsHelper:
                 # Caller wants a "retry" - restart probing
                 probe_num = 0
             self._move_next(probe_num)
+            # Make sure that the toolhead is positoned before probing
+            # otherwise this will cause 'homing:homing_move_begin' to trigger
+            # while moving horizontally.
+            self.printer.lookup_object('toolhead').wait_moves()
             probe_session.run_probe(gcmd)
             probe_num += 1
         probe_session.end_probe_session()


KevinOConnor:
Thanks.  This change seems like it would cause significant delays during probe sessions (as wait_moves() takes a long time to complete).  I'm not sure I understand the problem being reported - why does homing_heaters care where the toolhead is when homing nominally starts?

-Kevin

nickweedon:
Thanks for checking this out! 
So homing_heaters cares about this because it needs to know when the actual BLTouch probe is about to begin as opposed to when the toolhead is simply being horizontally re-positioned. Otherwise it will keep the heaters turned off during the entire movement from the previous probe to the next probe. The impact is quite apparent as it means that the entire bed mesh probing operation becomes one giant chained sequence of probes where the bed heater is never re-enabled for more than a few milliseconds and thus the bed will usually be cold by the end of bed mesh calibration.

Here is an example of the old behavior:

Turn off heater -> probe -> turn on heater -> turn off heater -> probe -> turn on heater -> turn off heater -> move to next probe position and probe -> turn on heater -> turn off heater -> ...

New behavior (bed will be reheated between toolhead horizontal moves):
Turn off heater -> probe -> turn on heater -> turn off heater -> probe -> turn on heater -> move to next probe position -> turn off heater -> probe -> turn on heater -> turn off heater -> ...

Note that I have not observed any slowdowns from 'wait_moves()' since it the printer is waiting on the horizontal movement but its quite possible i overlook/misunderstand something here?

As an aside, this fix is also necessary for my related bed reheating feature PR to work #6780

nefelim4ag:
I did something like this to test:
```
[homing_heaters]
steppers: stepper_z
heaters: heater_bed
```
And tested the bed mesh sequence without and with patch:
image

I don't think this is the right way to work around that behavior. 

As an example, that is my `bed_mesh` config:
```
[bed_mesh]
horizontal_move_z: 3
mesh_min:  20, 5
mesh_max: 380, 355
speed: 1000
probe_count: 7, 7
```
So `380/7 = 54 mm` maximum distance between probes.
By using the [Prusa calculator](https://blog.prusa3d.com/calculator_3416/#speed)
With cruise ratio = 50%, move profile will look like this:
image

Let's roughly estimate travel time to move with that speed, assuming half of the time it will accelerate or deaccelerate.
`54 / 2 / 400 + 54 / 2 / 200 = 0.2s` this is barely enough to enable and disable heaters - heaters always scheduled 300ms ahead.

I tried to limit the speed to 50 mm/s, so I have around 1s between probes, and there is enough time to enable and disable them.

Looks like there are just no events that will trigger heaters enabled/disabled, and `self.printer.lookup_object('toolhead').wait_moves()` works as a `hack`.

nickweedon:
Thanks for checking out this change @nefelim4ag.

So just to point something out here, irrespective of whatever algorthm you use and given current bed heater technology, if you are spending the majority of the time probing as opposed to anything else and you are also wishing to turn off the bed to avoid EMI interferance, then you will always have the bed go cold.

There is no free lunch here, it is impossible to keep the bed hot while running at the kind of speed and acceleration you are using.
I'm not sure also why you refer to 'wait_moves' as a hack as this is acutally correct behavior. Repositining the toolhood should not be considered part of probing. This is part of the mesh calibration and it is illogical to conflate these things and it is in essence a bug.

This bug fix is actually not supposed to address bed heating in particular since technically this bug fix has nothing to do with bed heating, just the probing event which is logically decoupled from bed heating via an observer pattern. This event could actually be consumed by all kinds of different code for various reasons.

If you take a look at #6780 then you will see the actual solution that I have for bed heating which relys on this fix. It will wait for the bed to re-heat between probes but within a threshold. This is the only possible way to solve this problem.

Sineos:
While I cannot comment on how the proposed code is working, I do agree with the intention behind it.

When using the `[homing_heaters]` functionality, the temperature drops for both the bed and the extruder are quite significant, as they stay off for longer periods, such as during Quad Gantry Levelling or adaptive meshing.

I'd also appreciate an improvement here.

nefelim4ag:
I think I know how to "fix" this behaviour.
```diff
diff --git a/klippy/extras/homing.py b/klippy/extras/homing.py
index 5b484529b..fa93e7bff 100644
--- a/klippy/extras/homing.py
+++ b/klippy/extras/homing.py
@@ -76,8 +76,6 @@ class HomingMove:
                 for cp, tp in zip(cpos, thpos[:3])] + thpos[3:]
     def homing_move(self, movepos, speed, probe_pos=False,
                     triggered=True, check_triggered=True):
-        # Notify start of homing/probing move
-        self.printer.send_event("homing:homing_move_begin", self)
         # Note start location
         self.toolhead.flush_step_generation()
         kin = self.toolhead.get_kinematics()
@@ -86,6 +84,8 @@ class HomingMove:
         self.stepper_positions = [ StepperPosition(s, name)
                                    for es, name in self.endstops
                                    for s in es.get_steppers() ]
+        # Notify start of homing/probing move
+        self.printer.send_event("homing:homing_move_begin", self)
         # Start endstop checking
         print_time = self.toolhead.get_last_move_time()
         endstop_triggers = []
```

`flush_step_generation()` basically ensures that any previous movement has been finished before probing/homing starts.
As a bonus, it adds some time ~0.25s, for the heater to heat.

This is the current master:
image
This is the suggested "fix":
image

The only other user of the event is sensorless homing, where register writing is blocking, so it seems safe to do it that way, without many additional pauses for probing.

Thanks,
-Timofey

KevinOConnor:
If that fix is sufficient, why not just call `toolhead.flush_step_generation()` from `homing_heaters:handle_homing_move_begin()`?

Cheers,
-Kevin

nefelim4ag:
Sure, it provides the same effect:
```DIFF
diff --git a/klippy/extras/homing_heaters.py b/klippy/extras/homing_heaters.py
index 4bb9ba040..a06d03bb1 100644
--- a/klippy/extras/homing_heaters.py
+++ b/klippy/extras/homing_heaters.py
@@ -49,6 +49,7 @@ class HomingHeaters:
     def handle_homing_move_begin(self, hmove):
         if not self.check_eligible(hmove.get_mcu_endstops()):
             return
+        self.printer.lookup_object("toolhead").flush_step_generation()
         for heater_name in self.disable_heaters:
             heater = self.pheaters.lookup_heater(heater_name)
             self.target_save[heater_name] = heater.get_temp(0)[1]
```

Thanks,
-Timofey

---
I just thought about it in the way "how to send an event right before the descent phase", otherwise, as a fix for `homing_heaters.py` your suggestion is more appropriate.

KevinOConnor:
Okay thanks.  I don't have any printers that use homing_heaters, but if someone that does finds that change useful I'm fine with merging it.

Cheers,
-Kevin

plampix:
This change is very useful. By lowering my bed_mesh speed, my printer is able to keep bed temp at ±0.2 degrees of the target temp during mesh creation.

PR #6785, opened 2025-01-15
Draft: Add new eddy current probe support with tap functionality

vvuk:
Latest standalone code for this is here: https://github.com/vvuk/eddy-ng

I've been working on updated eddy current sensor support for a week or so now, with a goal of supporting both homing and accurate z-offset setting via "tap" ("touch" etc.) like functionality. This works, though I'm still gathering testers (my sample size is 2 currently of people it works for, including myself). I'm opening this draft PR to get early feedback.

This is implemented as separate "ng" versions of ldc1612 code and eddy current sensor code. I ended up rewriting large parts of these, and I wanted to make sure that the current eddy current support can continue to be used as-is to make it easy for people to switch back and forth (especially during development).

Homing is similar to current `probe_eddy_current`. However, as there's no temperature calibration or compensation, homing is "coarse" at print temps. Tap is used to set an accurate physical offset at print temperatures.

"Tap" is based on an accumulated sum of a simple average of the derivative of a windowed moving average of the sensor frequencies (say that 10 times fast). I spent a while with a Jupyter notebook and a bunch of data, and this seems to give a very reliable signal:

image

Another plot with tap detection starting and ending bars (all the way to the right; orange start, green end at which point it's triggered):

image

The code generates these interactive plots in `/tmp/tap.html` if `plotly` is installed in klippy's venv at tap time.

Post-tap, the sensor's reading is taken at toolhead Z=2 (or whatever the home trigger height is set, because that's also used as the probe height), and the difference between that value and 2 is saved as a "tap offset" which is then added to sensor readings for post-tap bed mesh (as I type this out I realize there's a bug here, actually; will fix). The idea being that post-tap, the steppers + z offset is the most accurate physical distance, so I can figure out what the offset is from the sensor's reading to the true value. Then, for bed mesh, I assume that in bed-mesh-level variance ranges, the sensor will still accurately enough detect +/- 0.2-0.3mm from that point, even if the absolute value isn't correct.

There's also functionality for testing sensor ranges at different drive currents; BTT Eddy's coil design in particular I think leads to a much less useful sensor range, and I end up needing to use a higher drive current for "tap" than I do for home to get useful physical ranges.

I've also written up [way too many words for documentation](https://hackmd.io/yEF4CEntSHiFTj230CdD0Q?view).

Also, if someone who knows more about the physical behaviour of stepper motors when there's contact involved between nozzle/build plate wants to verify that I'm doing the right thing for calculating the true Z offset, it would be appreciated. I've talked myself into this, which could be entirely off base:

- The toolhead's position at the tap start time is true Z=0 (plus a configurable constant value, typically to "back it off" a bit)
- When the drip move is stopped, the steppers will have been commanded to a tiny bit beyond that point, even though there's physical contact with the build plate
- I don't believe the Z steppers will have skipped at this point, instead the force of nozzle against build plate is spread out amongst the entire mechanical assembly 
- So an accurate z offset is purely the toolhead's position (well, the negative of) at the tap start time.
- The toolhead lifting will just "undo" the motion of the steppers, again without skipping, and will just release that force that was spread out. Meaning that touching the build plate will not cause any Z stepper skips or otherwise incorrect toolhead Z values.

I would appreciate any feedback, as well some idea of whether this would be considered for merging or not. This PR also includes a simple MIT-licensed `printf` -- I added a way to send back some runtime debug logging messages back to klippy to preserve my sanity :) This can be stripped out, or I'm happy to move this to a separate PR.


gueee:
Hi Vladimir!
I've got 6 Cartographers and two Mellow Fly Pros to test and would be happy to help you testing and pushing your project along.

Sineos:
Thanks a lot for this PR. Just some "non-code" comments as food for thought:
* Probably you are aware of https://github.com/KevinOConnor/klipper-dev/tree/work-ldctap-20240430
* From a user perspective, I'd consider the fragmentation between BTT Eddy, Cartographer, Beacon, etc., a major issue.
  * It forces the user to install third-party code, causing extra effort and marking Klipper as "dirty."
  * It potentially creates issues on main-line Klipper updates, as suddenly the third-party code no longer works.

Personally, I think that some effort to consolidate the interfaces and functionality between these products would greatly benefit the users (and potentially the developers as well in terms of maintenance and effort).

Of course, this depends on the willingness and ability of the different actors, e.g., @dalegaard, @mattthebaker, @ksavinash9, @wohe, @bigtreetech, @flymaker, @jaysuk, to align on a common approach.

dalegaard:
Hi,

Kevin has made his stance on this stuff abundantly clear and I'm not interested in being involved in the Klipper project. Please refrain from highlighting me.

Best regards,
Lasse

vvuk:
> Probably you are aware of https://github.com/KevinOConnor/klipper-dev/tree/work-ldctap-20240430

Yep! I chatted briefly with Kevin before starting down this path. I started building on top of that branch but things evolved pretty rapidly. (I also 100% realize that "rewrite the whole thing" isn't great -- a lot of that was due to me needing to more deeply understand the code, and a lot was code cleanup along the way so that I could have an easier time debugging/diagnosing issues.)

> From a user perspective, I'd consider the fragmentation between BTT Eddy, Cartographer, Beacon, etc., a major issue.

Yeah, absolutely. It's one of the reasons I started doing this -- I looked at the code for eddy, carto, etc. and they were all doing similar things in slightly dissimilar ways... and other things in wildly different ways, but unnecessarily so. I have BTT Eddys, Mellow Fly, Cartographer here that I'm planning on testing with myself. The code here should support any LDCxxx sensor (and technically really any sensor that provides some kind of value that can be smoothly mapped to a height -- I have one of these "Bed Distance Sensor" probes that I haven't looked into how they work).

There's a lot that can and should be shared so that improvements are uniform. There's still value in the different products, as coil design significantly impacts precision and range.

vvuk:
A few notes about the approach I took...

Calibration:
  * I originally did calibration via polynomial fitting, increasing in degree until a RMSE threshold was crossed. This worked, but was never really perfect and even a 0.005mm divergence matters for e.g. bed mesh. I reverted back to smoothing out the data and interpolating values. It's still worth exploring fitting, perhaps with some better data conditioning.
  * TODO: test doing multiple calibration moves and average out the data.
  * TODO: expand out the eddy-ng PROBE_ACCURACY (vs. the standard PROBE_ACCURACY) to do a ManualProbe, and then check sensor height reading vs true-Z at a few different heights. This will give a better idea of true accuracy, especially in the 1mm-3mm range which (IMO) is the most important: bed mesh, QGL, etc. would be done at 2.0mm (or centered around whatever the "home height" is).

For tap detection, I looked at the following:
  * Simple sum-of-frequency-change for last N sensor readings, and then waiting for that value to start dropping rapidly, usually expressed as a % drop. This can detect a tap, but for it to be reliable the % has to be very high meaning that tap gets detected much later (50ms or so in my test cases at 3mm/s)
  * 1D Canny edge detection. Very late detection, needs a lot of smoothing to not give false triggers.
  * Higher-order [Daubechies wavelet transform](https://en.wikipedia.org/wiki/Daubechies_wavelet) (suggestion from someone who knows way more about stats/signals than I do) -- might be promising
  * TODO: Use higher accuracy [finite difference coefficients](https://en.wikipedia.org/wiki/Finite_difference_coefficient#Central_finite_difference) for derivative approximation -- this looks promising, and I'm going to explore this for the derivative calculations
  * TODO: make some of the window sizes configurable, and possibly even implement a few different runtime-selectable detection approaches (for testing, mainly).
  * TODO: right now the tap speed defaults to 3mm/s. This is quite slow, and while it works for me, it does make the touch very gentle and doesn't work well for some testers. I need to gather some data at faster speeds. I've got a Sacrificial Ender (tm) that I won't mind toolhead crashes and whatnot to test with, probably over the weekend.

I'll clean up and share the Jupyter notebook I've been using for data analysis; right now every tap dumps data to `/tmp/tap-samples.csv` that can be imported and looked at.

General implementation:
  * I see the whitespace test failing :) I've got lots of lines over 80 characters, because I'm not working on a VT100. We can reformat down the line :)
  * Right now it's a full klipper fork, despite just a few added files. I'm going to create a separate repo that's a separately-installable set of files and a simple shell script to patch some of the compilation things. (Klipper could actually benefit from an `EXTRA_SOURCES` in the makefiles to add things to `src-y` without needing to edit makefiles.)

vvuk:
(Note: this branch lags behind my `vlad/eddy-ng` branch in my repo -- this branch has just the eddy-related things, so I manually update it every few days)

pellcorp:
Seems like this would have a much greater chance of being merged if the changes were the minimum required to make it work and of course were changes to the existing files rather than new, also what is with the printf files is that just for convenience?   I would love to consider this for Simple AF project I have a ton of btt eddy users and probably some carto users who would be up for it

pellcorp:
In our testing we also noticed eddy does not like to be used colder than when it was calibrated, is this something your changes are proposing to fix?

vvuk:
> Seems like this would have a much greater chance of being merged if the changes were the minimum required to make it work and of course were changes to the existing files rather than new

The majority of the work is new, and changes to existing files would be extensive. There isn't much point in looking at a diff (I mean you can diff e.g. ldc1612.py and ldc1612_ng.py, and there will be some common/shared code, but not much). If this gets merged it should be a super set of the previous code.

> also what is with the printf files is that just for convenience

Yep, like I mentioned at the start, it's there for some debugging prints, but is disabled/turned off in this PR. Because of the way this PR was made it's still in here, but I'll update things to remove it shortly.

> In our testing we also noticed eddy does not like to be used colder than when it was calibrated, is this something your changes are proposing to fix?

Temperature is pretty irrelevant with the approach I'm taking, except if the extremes are such that the coil simply does not function as needed in those temperatures (i.e. requires drive current or other adjustments).

pellcorp:
> > Seems like this would have a much greater chance of being merged if the changes were the minimum required to make it work and of course were changes to the existing files rather than new
> 
> The majority of the work is new, and changes to existing files would be extensive. There isn't much point in looking at a diff (I mean you can diff e.g. ldc1612.py and ldc1612_ng.py, and there will be some common/shared code, but not much). If this gets merged it should be a super set of the previous code.
> 
> > also what is with the printf files is that just for convenience
> 
> Yep, like I mentioned at the start, it's there for some debugging prints, but is disabled/turned off in this PR. Because of the way this PR was made it's still in here, but I'll update things to remove it shortly.
> 
> > In our testing we also noticed eddy does not like to be used colder than when it was calibrated, is this something your changes are proposing to fix?
> 
> Temperature is pretty irrelevant with the approach I'm taking, except if the extremes are such that the coil simply does not function as needed in those temperatures (i.e. requires drive current or other adjustments).

I think you would have way more luck getting this merged it if changed the existing code rather than added new, but of course I am just an interested bystander :-)

We get the `Error during homing probe: Eddy current sensor error` if we use the probe colder than we calibrated it, so for example if someone calibrates the eddy at 35c and tries to use it for bed mesh at 30c they will often (almost always in my experience) get the eddy current sensor error.   In my project Simple AF (its a way to support Creality K1s / K1Ms with eddy, carto etc) we actually have a lot of hacks to work around this.

Anyway I will probably be creating a branch of my own klipper fork with the changes from this PR and asking some Simple AF users to have a play with it, we will let you know our experience :-)

It would be really awesome if this could be merged at some point.

zingo:
I just got a extra PEI sheet from a another band and my eddy calibration was trown off :(  I think that the "Tap" at each print start you do in this PR might actually remove the need to recallibrate each time I select different sheets. Good work :)

zingo:
Tried and it seem to help the PEI switching, without any calibration I don't get the same first layers on both, but kind of close enough, e.g. second sheet seem to work but not 100 pretty. I vade just done some fast test but it's a great improvement in my workflow not needing to care about it. Thanks 🙏

vvuk:
Latest standalone/integratable code for this is here: https://github.com/vvuk/eddy-ng -- I'm going to close this PR!

KevinOConnor:
Thanks for working on this.  Is there a desire to merge this into the mainline Klipper branch?  If so, what are the next steps for that.

-Kevin

vvuk:
> Thanks for working on this. Is there a desire to merge this into the mainline Klipper branch? If so, what are the next steps for that.

I'd like to get it into Klipper -- without tap, it works pretty much the same as the current eddy support. From my prespective it's ready to go; it's been tested by at least a few dozen people by now. Only future improvements would be making some improvements to the auto-setup experience, and maybe thinking about how to handle an edge case where tap is detected too early if a significant amount of noise happens _just_ before the actual tap. (Though that issue is mostly resolved by doing a few tap samples and throwing out outliers.)

Sineos:
Does this fix the "jumping" Z-Offset issue that some users are experiencing? (Including myself, which is why I reverted the BTT Eddy mod on one of my printers.)
If so, it would definitely be something I'd much appreciate being addressed, and thus the PR be reopened and followed up.

KevinOConnor:
> I'd like to get it into Klipper

Okay, let me know how you'd like to proceed with that.  It also seems to me that it would be useful to add.

> Does this fix the "jumping" Z-Offset

I'm not familiar with that issue.  Do you have a link describing the issue?

Thanks,
-Kevin

vvuk:
> Okay, let me know how you'd like to proceed with that. It also seems to me that it would be useful to add.

I think this is more a question for you, as far as how you'd like me to proceed :) I'm happy to clean up/revive this PR if that's the right path?

KevinOConnor:
Thanks.  The biggest issue on my side is time constraints.  If you can point me to some core areas of code that you'd like me to take a look at, then I can try to give some initial feedback.  If your documentation link above is still pertinent I'll try to familiarize myself with that as well.

-Kevin

vvuk:
Yep, understood. Up to date documentation is [in my standalone repo](https://github.com/vvuk/eddy-ng/wiki).

Until I update this PR, the latest code is [over there as well](https://github.com/vvuk/eddy-ng). At this point it's pretty extensively used/tested by many people, so I don't really have any specific areas that need feedback. The code is pretty lengthy, though much of it is "support" -- e.g. code for writing plots of taps/calibration to help diagnose issues, code for automatic setup to try to help people with BTT Eddys figure out parameters, code for various tests, etc. So maybe the best thing would be feedback on the high level approach:

- Add new `_ng` variants, to not break/change anybody's existing probe_eddy_current configuration.
- The "hard pwm" additions for stm32f042 in order to support Cartographer (this is largely a port of the code from the Cartographer klipper repo)
- Also, this is python 3 only

KevinOConnor:
Thanks.  As high-level comments, I think we'd want one driver in the mainline code.  It'd also be generally preferable to simultaneously support python2/3 (as it's typically relatively simple to extend support).  As for "hard pwm", maybe that could be added to mainline Klipper as a separate (and prerequisite) PR?

All that said, I'm not sure it'd be a good use of your time to redo this PR multiple times.  So, I can try to do an initial high-level review of your links above first.  I think it'll be a few days out before I can get that feedback to you though.

Cheers,
-Kevin

vvuk:
No problem, that works -- and the hard-pwm bits are easy to split out, so I'll do that.

For python2.. I think it would be difficult. There's e.g. a lot of type hints on the code, as well as I don't know what else that's python 3 specific. Python 2 is pretty dead... so if that's a requirement this might be a non-starter.

Sineos:
> I'm not familiar with that issue. Do you have a link describing the issue?

See the discussion here: https://klipper.discourse.group/t/btt-eddy-variance-with-home-command/21827 and some language from BTT here: https://github.com/bigtreetech/Eddy#my-z-offset-doesnt-seem-to-save-and-resets-is-there-a-work-around-or-fix

For me both approaches had various success and never "felt stable"

vvuk:
> > I'm not familiar with that issue. Do you have a link describing the issue?
> 
> See the discussion here: https://klipper.discourse.group/t/btt-eddy-variance-with-home-command/21827 and some language from BTT here: https://github.com/bigtreetech/Eddy#my-z-offset-doesnt-seem-to-save-and-resets-is-there-a-work-around-or-fix
> 
> For me both approaches had various success and never "felt stable"

Not really related to eddy-ng but -- "homing" is not really intended to set a very precise Z offset with these types of probes. You're supposed to home for a coarse Z home (i.e. get Z roughly where it should be), and then do a regular PROBE, and then just set Z to the probed value. That's what (stock) PROBE/SET_Z_FROM_PROBE does.

The BTT "beta z offset" macros are a hack and I wouldn't use them.

vvuk:
I should also add -- eddy-ng doesn't really change anything related to homing. It will behave the same. eddy-ng does add a "tap" which is kind of like a home, but there's more processing on the probe so that the Z=0 position can be determined more accurately.

KevinOConnor:
> For python2.. I think it would be difficult. There's e.g. a lot of type hints on the code, as well as I don't know what else that's python 3 specific. Python 2 is pretty dead... so if that's a requirement this might be a non-starter.

Well, python2 wouldn't be needed for an initial review.  However, I'm not looking to change Klipper's minimum requirements at this time, so changing ldc1612 support to require python3 (along with the corresponding build checks) would certainly add notable complications to merging.

Cheers,
-Kevin

nefelim4ag:
Well, I did read parts of the code.
Random thoughts:
- This PR https://github.com/Klipper3d/klipper/pull/6734 will cover different crystal frequency, which is hardcoded in this PR per sensor.
- IIRC, it is really unwanted to do runtime divisions and floating operations on MCU, so this logic should be handled in high-level code.
- If configurable Deglitch makes sense, there is a limited amount of boards, so I think it also makes sense to define boards where current defaults should be replaced.

For cartographer:
stm32 pwm is used to replace crystal for LDC1612.
I found the https://github.com/vvuk/klipper/commit/b17c55ed7c0ac5caaffd09b0c6c3596290082367:
```
static const struct gpio_pwm_info pwm_regs[] = {
#if CONFIG_MACH_STM32F0
+  #if CONFIG_MACH_STM32F0x2
+    {TIM3, GPIO('B', 4), 1, GPIO_FUNCTION(1)}
+  #endif
  #if CONFIG_MACH_STM32F070
```
Well, It looks like it can be easily ported.

The second part allows for the change of MAX_PWM in gpio_pwm_setup() to drive up to 24Mhz.
Idk. Technically, it should be possible to just pull down that pin and use the internal LDC1612 oscillator.

vvuk:
> it is really unwanted to do runtime divisions and floating operations on MCU, so this logic should be handled in high-level code.

This really should be moved to fixed point math; I didn't bother (yet?) for simplicity during development, and because the primary targets have more than enough CPU power to handle the fp ops. Doing it in high level code isn't really desirable because the latency of the toolhead continuing to move down while the "processing" happens on the python side would be too high.

> The second part allows for the change of MAX_PWM in gpio_pwm_setup() to drive up to 24Mhz.
Idk. Technically, it should be possible to just pull down that pin and use the internal LDC1612 oscillator.

LDC1612 accuracy is directly tied to the accuracy of the reference oscillator, and the internal one is not amazing. TI recommends if at all possible to use an external one.

KevinOConnor:
Thanks for working on this.  I got a chance to review the code in the repo mentioned above.  I have some initial comments.

My main feedback is that the new functionality looks useful, but in order for it to be merged it would need to be broken up into a series of incremental commits that updates the existing system.  The new implementation appears like "a complete rewrite" of the existing eddy code (and parts of the probe code).  I don't think it is feasible for us to have two separate implementations for the eddy/probe code (as many have probably noticed, we have trouble maintaining all the functionality currently in Klipper - having to maintain multiple redundant systems would only make that much much worse).  Replacing the existing eddy system with a "big bang rewrite" would also not be feasible, as the risk of introducing a regression would be too high and we would lose the use of tools like `git bisect` and `git revert` to narrow down the cause of any new issues.

The preferred way to introduce large changes like this is to break them into a series of atomic commits - one commit per logical change.  That makes it easier to review each change and allows us to use `git bisect` to find regressions.  If at the end of that series of changes the resulting code looks nothing like the start, then that is fine.  If there is interest in doing this, we should discuss the best way to breakdown the changes into separate commits (eg, sensor_ldc1612 tap support, documentation updates, etc.).  I think we should both have a good feel for the final product before starting the laborious work of reorganizing changes.

Putting aside merging, I have some comments on the new implementation if you are interested in feedback:
* I like how you've been able to implement the wma tap detection based entirely on an increasing derivative.  This seems like a useful way to detect taps without fragile calibration code.  I suspect the "windowed moving average" part of the code may be a bit susceptible to noise - as the last `wma_d_avg` is always heavily weighted to the last sample read and thus seems susceptible to that last sample being noisy.
* I didn't fully understand how the SOS filter code is used.  It looks like the code looks for the last filtered sample to decrease in frequency from the peak filtered frequency observed.  Normally we'd expect the frequency to increase as the sensor gets closer to the bed, and then stop increasing once bed contact is made - at first glance looking for a decreasing frequency seems odd.  I'm not sure if the code is actually detecting the "bed bouncing" or if it utilizes some mathematical property of the butterworth filter to exhibit a decreasing value.
* I may have missed something, but it looks to me like this code has the mcu determine the final tap position.  I suspect it will be very hard to get good accuracy that way, due to the amount of noise in each individual ldc1612 samples.  For what it is worth, I think it would be better for the mcu to be used solely to determine when to stop the motors, and then use the host to figure out where the tap point actually is.  That is, the host can analyze the ldc1612 samples for 100+ms prior to contact, along with the samples taken while the toolhead lifts away from the bed.  With this info it can use a large number of samples to determine the "tap point" - it'll be the "elbow" in the resulting "graph" of samples.  @garethky has done a lot of work in this area with loadcell sensors.
* I noticed a couple areas of the sensor_ldc1612.c code that replaced explicit flags with "sentinel values" that do not look correct to me.  Nothing prevents a "pin" from being -1, nothing prevents an oid from being zero, and nothing prevents a 32bit mcu clock from being zero.  So, comparing with these values to determine if they are "unset" could result in subtle defects.
* The improved ldc1612 error resilience during homing looks useful (eg, in `check_error()`).  In particular, the observation that a high amplitude warning mainly occurs during low frequencies (probe far away from the bed) is very interesting.  Seems useful to get that support merged.
* I noticed that the mcu code implements the sos filter using "float".  Note that @garethky does have a fixed-point mcu based sosfilter system implemented for loadcells.  (Also, fwiw, even if we did want to use "float" I don't think we can rely on the host and mcu having identical bit representations, so copying the binary host values looks problematic.)
* FYI, I noticed some debugging code, areas for optimization, and areas where the documentation should be updated.  (Note that the mcu code does have an `output()` function that looks similar to the added `dprint()`.  Also, note that all timers are automatically unregistered on a shutdown - so `ldc1612_ng_shutdown()` looks odd.)  We can discuss those things in more detail if there is interest.

Thanks again for working on this,
-Kevin

vvuk:
> My main feedback is that the new functionality looks useful, but in order for it to be merged it would need to be broken up into a series of incremental commits that updates the existing system. The new implementation appears like "a complete rewrite" of the existing eddy code (and parts of the probe code).

Thanks for the feedback, I'll respond below (super useful and good catches on some things). Unfortunately at this point it's not really feasible to break it up into a series of commits; or rather, I don't really want to invest the time to do so. Even if I did, I'm not sure it would be useful for bisecting as there would probably be 2-3 large commits to move chunks of the code to the new setup that would work individually. I would suggest as an alternative to just include both. It also may actually be best to just keep it as a separate extension, as there is further work I want to do here and that would be easier outside of the klipper code (and/or would then need to keep the code in klipper up to date). 

> Putting aside merging, I have some comments on the new implementation if you are interested in feedback:

Absolutely :)
 
> * I like how you've been able to implement the wma tap detection based entirely on an increasing derivative.  This seems like a useful way to detect taps without fragile calibration code.  I suspect the "windowed moving average" part of the code may be a bit susceptible to noise - as the last `wma_d_avg` is always heavily weighted to the last sample read and thus seems susceptible to that last sample being noisy.

I'm actually planning on entirely removing the `wma` approach in the near future. It doesn't have any benefits over the filtering approach; it's more susceptible to noise, spurious detection, etc. It was what I implemented first, and then moved to the filtering approach which has been very reliable across many users' machines.

> * I didn't fully understand how the SOS filter code is used.  It looks like the code looks for the last filtered sample to decrease in frequency from the peak filtered frequency observed.  Normally we'd expect the frequency to increase as the sensor gets closer to the bed, and then stop increasing once bed contact is made - at first glance looking for a decreasing frequency seems odd.  I'm not sure if the code is actually detecting the "bed bouncing" or if it utilizes some mathematical property of the butterworth filter to exhibit a decreasing value.

The way I understand it is this -- the frequency increases until the bed is hit, at which point it levels off rapidly and becomes flat. The filter is doing a 2nd order butterworth bandpass filter between 5-25hz (empirically derived). The filtered data is flat during the period when the frequency is steadily increasing. When the signal levels off, that's essentially an additional high frequency component that's introduced at that point. The filter tries to smooth this out (vs the previous steady state) by introducing a large negative impulse. It's that negative impulse that I look for.

> * I may have missed something, but it looks to me like this code has the mcu determine the final tap position.  I suspect it will be very hard to get good accuracy that way, due to the amount of noise in each individual ldc1612 samples.  For what it is worth, I think it would be better for the mcu to be used solely to determine when to stop the motors, and then use the host to figure out where the tap point actually is.  That is, the host can analyze the ldc1612 samples for 100+ms prior to contact, along with the samples taken while the toolhead lifts away from the bed.  With this info it can use a large number of samples to determine the "tap point" - it'll be the "elbow" in the resulting "graph" of samples.  @garethky has done a lot of work in this area with loadcell sensors.

The code for this has changed a bit (and is about to change even more, actually). I think in what you looked at, I record the time at which the signal started dropping, and if the distance crosses the threshold, then the "tap" is recorded at the position where the signal started dropping. There's noise in each individual sample, but it's not really significant as all I care is the time at which a consistently negative slope began in the signal, and not the actual frequency value. If that time is one or two samples earlier or later than the "true" value, it does mean some variance (3mm/s tap, default 250 samples/second, so 0.012mm travel per sample; I typically run at at least 500 s/s here). But that's why I do multiple taps and then pull out a cluster that has a low stddev.

But in that, the threshold is only used to detect the tap; the tap time is always the start of that downwards slope. In the new code, the tap time is a point in the region between the start of the tap and when the threshold is crossed. If this is at 0, then it's the same as the previous code. But setting it to 0.3 (30% of the way between) results in much more consistent and accurate results, because it's much more robust to the occasional "soft tap" (where for a single tap, there just happens to be a consistent downward slope in the filtered values... even if at the beginning it's a very slight downward slope, meaning it takes longer in time to actually hit the threshold). 

> * I noticed a couple areas of the sensor_ldc1612.c code that replaced explicit flags with "sentinel values" that do not look correct to me.  Nothing prevents a "pin" from being -1, nothing prevents an oid from being zero, and nothing prevents a 32bit mcu clock from being zero.  So, comparing with these values to determine if they are "unset" could result in subtle defects.

Good catch, I should've caught the mcu clock as zero but didn't think of the others! (Though, really? pins can be negative values, and oid=0 is valid? .. why? :) ) I'll sort this out. 

> * The improved ldc1612 error resilience during homing looks useful (eg, in `check_error()`).  In particular, the observation that a high amplitude warning mainly occurs during low frequencies (probe far away from the bed) is very interesting.  Seems useful to get that support merged.

Yeah, that might be a good improvement for the current code even if this doesn't land. One thing that's odd is that in my testing, on pretty much all of my printers, I get non-error readings even very high off the build plate (>200mm). They're wrong (frequencies way too high from what I'd expect; basically everything above 20mm seems to give very similar values), but I don't get amplitude errors. Others absolutely do.

> * I noticed that the mcu code implements the sos filter using "float".  Note that @garethky does have a fixed-point mcu based sosfilter system implemented for loadcells.  (Also, fwiw, even if we did want to use "float" I don't think we can rely on the host and mcu having identical bit representations, so copying the binary host values looks problematic.)

Yep, this was just for efficiency in implementation as I wanted to get things working. All the current eddy probes that this runs on (btt eddy, cartographer, mellow fly) have good enough FP support (even if soft fp) that it hasn't been an issue, but fixed point is on the list at some point. I'll take a look at the loadcell code, would be great to just share that.

As for float representation... I mean, yes, it's possible that something doesn't use IEEE 754, but at that point you'd also have to support e.g. big endian and non-2's-complement integers, and while I guess I could run Klipper on my motorola 68k or MIPS machines, it's not likely :) Is there any hardware that klipper runs on that doesn't do IEEE 754?

> * FYI, I noticed some debugging code, areas for optimization, and areas where the documentation should be updated.  (Note that the mcu code does have an `output()` function that looks similar to the added `dprint()`.  Also, note that all timers are automatically unregistered on a shutdown - so `ldc1612_ng_shutdown()` looks odd.)  We can discuss those things in more detail if there is interest.

I had problems getting output() to go somewhere where I could actually see it, and instead of trying to hack that I just added my own along with richer formatting. Certainly not required (and not enabled by default). The timer issue... yeah, I need to remove the shutdown. I had a few instances early on in testing where it looked like I had a timer fire after shutdown, and things got confused. I'll take the shutdown out and make sure that isn't happening.

> Thanks again for working on this, -Kevin

Thanks for the feedback! Even if this doesn't make it into klipper, it's appreciated and gives me some things to improve.

garethky:
I'll get you the SOS fixed point filter broken out into its own MCU object in my next load cell PR. It will be limited to a Q12 fixed point representation.

The code would be a lot simpler if we could use 32bit floating point. To get floating point to work correctly on all targets, to the best of my understanding, I would need the following:
* Update the serial protocol to explicitly support floating point. For every architecture you need to implement converting wire protocol bits to system float 32. I.e. solve endianness and representation standars (maybe we are lucky and they are all the same!)
* Where applicable, enable the FPUs. My understanding is that they are not turned on by default, so even on a chip with an FPU, a floating point multiplication may be done in a software library, because the compiler cant use the FPU instructions.
* Allow targeting of the SOS filter only to chips that have a 32 bit FPU in the build system.

I decided the above was out of scope for my project. Having to convert everything to a fixed point representation comes with limitations that impact other parts of the code that I don't like. But fixed point was the path that let me cut scope.

vvuk:
> I decided the above was out of scope for my project. Having to convert everything to a fixed point representation comes with limitations that impact other parts of the code that I don't like. But fixed point was the path that let me cut scope.

Makes sense. For me, because the targets needed specific hardware (ldc1612), a general solution wasn't required. I'll double check the fpu enabling though.

rcsez:
If I'm reading this correctly, the tap function should be agnostic to build plates when it comes to setting Z height? 

A problem with the standard Eddy setup for me has been that if I setup and calibrate Eddy with one build plate, it works fine, but if I switch to a different brand plate, the Z height can be off by ~0.5mm, so the Z-offset needs to be dialed in all over again.

PR #6786, opened 2025-01-18
Updating homing speed for sensorless homing

thomasfjen:
Hello there, 

Update homing speed for sensorless homing
The TMC 2209 documention recommends a higher speed.
Currently the recommended starting homing speed is half as rotation speed, which would equal half a rotation per second and that is in my thinking well below one revolution per second as the documentation is stating. 

```
StallGuard4 operates best at medium motor velocities. Very low motor velocities (for many motors
well below one revolution per second) [...]
```
https://www.analog.com/media/en/technical-documentation/data-sheets/TMC2209_datasheet_rev1.08.pdf S.59 Section 11.5

If there are any objections please let me know.

### DIFF
--- a/docs/TMC_Drivers.md
+++ b/docs/TMC_Drivers.md
@@ -187,12 +187,12 @@ the end of the rail. However, the TMC drivers can't reliably detect a
 stall at very slow speeds.
 
 A good starting point for the homing speed is for the stepper motor to
-make a full rotation every two seconds. For many axes this will be the
-`rotation_distance` divided by two. For example:
+make a full rotation every second. For many axes this will be equal to the
+`rotation_distance`. For example:
 ```
 [stepper_x]
 rotation_distance: 40
-homing_speed: 20
+homing_speed: 40
 ...
 ```
 


flowerysong:
You appear to have misread the documentation. "for many motors well below one revolution per second" is not a recommendation of one revolution per second.

thomasfjen:
> You appear to have misread the documentation. "for many motors well below one revolution per second" is not a recommendation of one revolution per second.

You are right the documentation does not recommend that speed, i have worded that not great.

My objection still is that the half a rotation per second from the klipper docs would fall in to the range of "for many motors well below one revolution per second".

nefelim4ag:
I will add my 2 cents, speed will only depend on a specific motor, to be specific on motor inductance.
You can use https://www.reprapfirmware.org/emf.html to check.
My current understanding will suggest something around  60 mm/s for 1,5mH, 40 mm/s for 3mH, and 20 mm/s for 6mH.
I think with proper testing someone can reverse engineer the TMC detection logic and get the minimal backEMF value, which is required for stallguard to work reliably.

So, there is no right default value, this one is good for old motors used on Prusa printers (they have 6mH afaik).

BTW, I [suggested](https://klipper.discourse.group/t/tmc-coolstep-and-how-to-make-it-works/16259) some tool (https://github.com/Klipper3d/klipper/pull/6592) to at least peek at the actual motor values, but there is a lack of interest.

PR #6789, opened 2025-01-21
Crazy3DPrinter CZ-300

DragonSlayerSE:
Motherboard is named "3FD1X ENTINE T1 2018/10/22"

### DIFF
--- a/config/printer-crazy3dprint-cz-300-2018.cfg
+++ b/config/printer-crazy3dprint-cz-300-2018.cfg
@@ -1,0 +1,189 @@
+# To install Klipper on your Raspberry pi (octoprint) or PC (any debian flavor)
+# try KIAUH https://github.com/th33xitus/kiauh#-download-and-use-kiauh
+
+
+# This file contains common pin mappings for XYZPrint Crazy3DPrint CZ-300.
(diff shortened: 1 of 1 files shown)


JamesH1978:
Thankyou for submitting a PR, couple of points before i review, please be aware you need to sign off your PR for the certificate of origin as per point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review and all new config example added need to also pass regression tests as per point 4 in https://github.com/Klipper3d/klipper/blob/master/docs/Example_Configs.md#guidelines

Thanks
James

DragonSlayerSE:
Commit message:

This is a config file for the Crazy3DPrint CZ-300 printer. Sold as a 300x300x400 printer, but with the right glass clamps you can use all the 310x330. Since Klipper3D does not support multi microstepping modes, the A5984 stepper is forced to 32 microsteps via [static_digital_output]
There are several unpopulated pinheaders (jst 2.0) on the motherboard, Like (burning, debug, calibrate, e-sensor, led, fan2)
English is not my native language, there could be som spelling problems.

Signed-off-by: Kire Dyfvelsten <kire@dyfvelsten.se>

PR #6790, opened 2025-01-22
Correct Pin number for Filament Runout Sensor

theminor:
Corrected Filament Runout Sensor Pin (as per schematic - see https://github.com/LeeOtts/Ender3v2-Klipper-Configs/blob/main/Creality.4.2.7.-.Schematic.28-5-22-1.pdf)

The alias `FIL_RUNOUT` was set to the wrong pin. Pin `EXP1_1` was set to `PC6` and `FIL_RUNOUT` was also set to `PC6`. As per the schematic, the filament runout sensor is pin `PA4`


JamesH1978 (approved):
I can confirm that PA4 is the correct pin for those boards

Thanks
James

KevinOConnor:
Thanks.  Happy to make a change, but I did not see a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html

-Kevin

theminor:
> Thanks. Happy to make a change, but I did not see a signed-off-by line - see: https://www.klipper3d.org/CONTRIBUTING.html
> 
> -Kevin

Sorry about that. Amended the commit to add sign-off.

KevinOConnor:
Thanks.

-Kevin

PR #6791, opened 2025-01-24
screws_tilt_adjust: initialize status result as a dictionary

pedrolamas:
On [my previous PR](https://github.com/Klipper3d/klipper/pull/5992) to make this an object I forgot to correctly initialize `results` as an object; this PR fixed that.


KevinOConnor:
Thanks.

-Kevin

PR #6792, opened 2025-01-24
Improve developer documentation 

shitwolfymakes:
I noticed there's 0 explanation of how to compile klipper without installing it to the host system, so I added a section to the developer docs to explain how.

I don't like that it involves single-use modification of the build scripts, but that seems like the only real way to do this aside from adding scripts just to install dependencies for every OS with an installer


JamesH1978:
All PR's need signing off as mentioned in point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks
James

JamesH1978:
I would think a simple list of the software needed to compile firmware would be enough, then those developers wanting to only build firmware and not fully install klipper and its env, would be able to extrapolate the actual packages needed for their particular flavor of linux. For example, developer's libraries for ncurses, Arm GNU Toolchain, GCC, etc or even just the list of debian packages as klipper is usually installed on those flavors.

Thanks
James

PR #6793, opened 2025-01-24
Add constant to klippy.py

shitwolfymakes:
This replaces all instances of `'__init__.py'` with the constant `init_filename`

### DIFF
--- a/klippy/klippy.py
+++ b/klippy/klippy.py
@@ -8,6 +8,8 @@ import sys, os, gc, optparse, logging, time, collections, importlib
 import util, reactor, queuelogger, msgproto
 import gcode, configfile, pins, mcu, toolhead, webhooks
 
+init_filename = "__init__.py"
+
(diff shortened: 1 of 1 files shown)


JamesH1978:
Klipper is python 2 and 3 compatible to support older installs, i believe __init__.py is required in the former. 

Also please remember all PR's need signing off as mentioned in point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks
James

shitwolfymakes:
As for the first part, that *should* be irrelevant in this instance, as I'm replacing a hardcoded string in a string comparison with a constant storing the same exact value.

As for part two, I've never had to sign off commits before, i'll go figure out how to do that

flowerysong:
I don't see any reason to make this change. It just obfuscates the code and makes it harder to understand what it's doing.

PR #6794, opened 2025-01-25
fix(gc): freeze gc objects after startup 

ammmze:
This is my attempt to address Timer Too Close errors on some slower machines. Through some debugging that started with something @KevinOConnor pointed out in #6784, I observed a correlation between generation 2 collection happening at the same time as homing moves (in particular during happy hare drip/homing moves...with happy hare the user will experience many more "homing" moves than a standard non-mmu enabled printer because it "homes" filament to various sensors each time filament is swapped. I found that on my Orange Pi CM4, the generation 2 garbage collection takes between ~150-300ms depending on whether I have the shake and tune module enabled. Based on the discussion in #6784, those homing moves require low latency (<100ms) communication. So having gc that takes longer than this 100ms (which would block all other threads) is problematic. More details about my setup can be found [here](https://gist.github.com/ammmze/5770bb4833de3aafc63ab589e46cd371).

~What I am doing here in this PR is tracking how long each generation garbage collection takes (with a cap of 90% of the largest free time we've seen). With this information, we can prevent doing garbage collection if there's a possibility that gc will take us past the next scheduled timer.~

What I am doing here in this PR is performing a full garbage collection after klipper has finished loading and then performing a garbage collection freeze. This will move all the remaining mostly static objects to the "permanent" generation, leaving a lot less data for future gen 2 collections to process, making gen 2 collections much faster.

I am open to other ideas if you've got any. I don't think my solution is in any way bullet proof...there's still a possibility that a long running garbage collection could take too long and take us past the next timer, but I think it will be less likely.

I will acknowledge that standard klipper does do garbage collection much less frequently, but the reality is many people have extra modules like cartographer, led effects, happy hare, etc. While it would be nice if these all did their best to avoid leaving garbage for the garbage collector, the reality is that it would be a large effort and different communities have varying levels of knowledge to make this happen. So I think we should try to accommodate the best we can.

### DIFF
--- a/klippy/extras/garbage_collection.py
+++ b/klippy/extras/garbage_collection.py
@@ -1,0 +1,31 @@
+# Garbage collection optimizations
+#
+# Copyright (C) 2025  Branden Cash <ammmze@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+import gc
+import logging
+
+class GarbageCollection:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        # feature check ... freeze/unfreeze is only available in python 3.7+
+        can_freeze = hasattr(gc, 'freeze') and hasattr(gc, 'unfreeze')
+        if can_freeze:
+            self.printer.register_event_handler("klippy:ready",
+                                                self._handle_ready)
+            self.printer.register_event_handler("klippy:disconnect",
+                                                self._handle_disconnect)
+
+    def _handle_ready(self):
+        logging.debug("Running full garbage collection and freezing")
+        for n in range(3):
+            gc.collect(n)
+        gc.freeze()
+
+    def _handle_disconnect(self):
+        logging.debug("Unfreezing garbage collection")
+        gc.unfreeze()
+
+def load_config(config):
+    return GarbageCollection(config)
--- a/klippy/toolhead.py
+++ b/klippy/toolhead.py
@@ -290,7 +290,7 @@ class ToolHead:
                                             self._handle_shutdown)
         # Load some default modules
         modules = ["gcode_move", "homing", "idle_timeout", "statistics",
-                   "manual_probe", "tuning_tower"]
+                   "manual_probe", "tuning_tower", "garbage_collection"]
         for module_name in modules:
             self.printer.load_object(config, module_name)
     # Print time and flush tracking


JamesH1978:
All PR's need signing off as mentioned in point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks
James

burkfers:
I have been trying this branch for about 16 hours of print time and not suffered any TTC (the standard causes such as canbus errors, high neopixel frame rate, throttling host etc have been ruled out).

Will update as time passes.

ammmze:
Another idea that was given via Discord chat was possibly doing a `gc.freeze()` at the start of a print and `gc.unfreeze()` at the end of the print. My understanding from that conversation is that it would move everything over to the permanent generation, which would then mean gen 2 collection doesn't have to go through as much stuff, which should make it faster. One issue with that is that feature is only available in python 3.7+, so we'd need to do some feature detection to ensure it doesn't completely break for older python versions and people would need to be on 3.7+ if they want this. 

Also I was thinking since there really shouldn't be much that gets loaded once klippy has fully loaded, perhaps we could have klippy do a full gc once everything has finished loading, then just do the `gc.freeze()` and never actually un-freeze them. So then collection would only run on stuff created and dropped since klippy started.

KevinOConnor:
Interesting, thanks.

I have some high-level thoughts.  These are just some initial thoughts, so don't take them too seriously.  Listed in no particular order.
* We can certainly optimize the Python garbage collection in Klipper.
* The current reporting is in place because we've ran into issues like this in the past.  The code does do a full garbage collection sweep on each restart (see klippy.py).  This could be extended to also run after connect I guess.
* If we limit running of gc based on previous scan times, we risk never running a gc.  It's common for the API clients to setup a "get_status()" watchlist which runs every 250ms.  So, if a gc sweep ever takes more than 250ms then we'll never gc.  Not gc'ing could introduce subtle failures due to swapping and/or OOM issues.
* A good way to avoid gc'ing in Python is to avoid having circular links between objects when those objects are temporary.  Or, if a circular link is needed, then manually break the link prior to dismissing the last reference to it.  For example, see: `git log 3efb424d..1af4a4ae9f`
* I agree it is difficult to get many other developers to think about these types of gc issues.  However, the code will *never* be stable with 3rd party "extras" modifications.  The only way to allow "plugins" in a way that will scale and be stable is to define an API and run those "plugins" in a separate python process.  https://klipper.discourse.group/t/new-proposal-for-klipper-extension-support/19150 
* FYI, even if `self._next_timer` is several hundred milliseconds in the future, it's still not safe for the Python process to become busy for an extended period.  At any point, the mcu could send a message to the host, and the host may need to respond within about 100ms.  (The heater code is 300ms, but other areas of the code could be less.)  New mcu message wakeup the reactor outside of the timer dispatch mechanism.

Again, just some high-level initial thoughts.
-Kevin

ammmze:
Thanks for sharing your thoughts!

> If we limit running of gc based on previous scan times, we risk never running a gc. It's common for the API clients to setup a "get_status()" watchlist which runs every 250ms. So, if a gc sweep ever takes more than 250ms then we'll never gc. Not gc'ing could introduce subtle failures due to swapping and/or OOM issues.

This is one of my concerns as well. I did try to accommodate it by essentially limiting the tracked gc time to 90% of the highest time between events. Though that 90% selection was arbitrary. But like you pointed out, I can confirm that we never see 300ms between events. I think the max I ever saw was like ~150ms, which means my current code would essentially track the 300ms gen 2 collection as 135ms, meaning it would allow gen 2 collection to run if it sees a 135ms window. And if it ends running that 300ms collection, there is a chance it could still trigger a TTC. So like I mentioned, this solution is not bullet proof...just reduces the chances of a TTC due to gc further.

> The only way to allow "plugins" in a way that will scale and be stable is to define an API and run those "plugins" in a separate python process. https://klipper.discourse.group/t/new-proposal-for-klipper-extension-support/19150

I am intrigued by this and will look forward to more on this. I hadn't see it before. Thanks for sharing!

> FYI, even if self._next_timer is several hundred milliseconds in the future, it's still not safe for the Python process to become busy for an extended period. At any point, the mcu could send a message to the host, and the host may need to respond within about 100ms. (The heater code is 300ms, but other areas of the code could be less.) New mcu message wakeup the reactor outside of the timer dispatch mechanism.

This is a very good point! And i think is a big reason why my current code would only be considered a bandaid. Ultimately, because garbage collection is a blocking process, the only true resolution is to be able to reduce the amount of time it takes to do garbage collection. And I think the separate plugin process will be extremely valuable for that.

In the meantime, I'd still like to try some `gc.collect()` and `gc.freeze()` stuff after klippy `connect` (or freeze on print start and unfreeze on print end/cancel) to see if that will reduce gen 2 collection time. If that proves to be successful, would you be open to accepting code that does some feature detection (i.e. detect if `freeze` is a method on `gc`) and perform that functionality? This would mean this gc optimization would only be available for users using python 3.7+. Users on older python would just skip that functionality.

ammmze:
Okay...this is seems promising...

I just did a test where i created a new module (`garbage_collection`) that just does the following upon receiving `klippy:connect`:

```python
def handle_connect(self):
    logging.info("Running full garbage collection and freezing")
    gc.collect(0)
    gc.collect(1)
    gc.collect(2)
    gc.freeze()
```

In the reactor, I just have my logging active so i can see the timings before and after. I've only included one set of logs for each gc generation. These are both after restarting klipper and the printer just sitting idle for a few minutes. We can see gen 2 collection is significantly faster if we freeze all the objects present after the initial load of klipper. Now in theory memory from this point _should_ be pretty stable as collection will still run on new stuff getting created. Thanks to @nefelim4ag for pointing me to the `gc.freeze()` functionality.

### BEFORE (i.e. **without** `gc.freeze()`)
```
Running gc gen 0 with 1.05ms freetime. Counts (701, 9, 9)
Finished gc of 716 items in 3.74ms
Running gc gen 1 with 40.69ms freetime. Counts (702, 10, 9)
Finished gc of 814 items in 2.13ms
Running gc gen 2 with 24.96ms freetime. Counts (701, 10, 10)
Finished gc of 1487 items in 326.70ms
```

### AFTER (i.e. **with** `gc.freeze()`)
```
Running gc gen 0 with 40.61ms freetime. Counts (705, 9, 9)
Finished gc of 676 items in 1.69ms
Running gc gen 1 with 40.64ms freetime. Counts (708, 10, 9)
Finished gc of 822 items in 2.45ms
Running gc gen 2 with 39.44ms freetime. Counts (704, 10, 10)
Finished gc of 996 items in 7.65ms
```

Thoughts?

Perhaps this could be an opt-in module initially (i.e. users would need to add `[garbage_collection]` to their config and then eventually if we find it works well for people who are having TTC errors, perhaps we make it enabled by default?

KevinOConnor:
Thanks.  That sounds useful.

> Perhaps this could be an opt-in module initially

It seems like it could be useful to enable it by default.  You can look at how toolhead.py auto-loads a bunch of modules.

You might also want to look at freezing after the "klippy:ready" callback as there are likely many objects created during the connect itself.

Cheers,
-Kevin

KevinOConnor:
Oh, if you "freeze" be sure to also "unfreeze" via the "klippy:disconnect" handler.  Otherwise, you'll leak memory on every RESTART command.

-Kevin

klippy/extras/garbage_collection.py
```diff
@@ -0,0 +1,26 @@
+import gc
+import logging
+
+class GarbageCollection:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        can_freeze = hasattr(gc, 'freeze') and hasattr(gc, 'unfreeze')
+        if can_freeze and config.getboolean("freeze", True):
```
ammmze, line 9: I put a config option in there, so the freeze functionality _can_ be disabled if for whatever reason a user needs to disable it. I'm not sure if we want to document it in the config reference. If we do, any thoughts where to put it? I didn't see a section for configs that are more related to klippy functionality vs all the parts of the printer.

klippy/toolhead.py
```diff
@@ -290,7 +290,7 @@ def __init__(self, config):
                                             self._handle_shutdown)
         # Load some default modules
         modules = ["gcode_move", "homing", "idle_timeout", "statistics",
-                   "manual_probe", "tuning_tower"]
+                   "manual_probe", "tuning_tower", "garbage_collection"]
```
ammmze, line 293: It was mentioned to look at how toolhead auto loads modules. I'm not certain that pointer was intended to say we should do the autoload here in toolhead, or if there would be a more preferred place to do the `load_object`?

ammmze (commented):
@KevinOConnor, I've pushed up my changes. It reverts all the reactor changes in favor of the `garbage_collection` module, which will autoload currently via the `toolhead.py`. The `garbage_collection` module will only do the freeze/unfreeze, if those are present in  the `gc` module. With the freeze moved from `klippy:connect` to `klippy:ready`, I am now seeing gen 2 collection taking 4-5ms vs the ~7ms when doing it at connect, so as expected that works even better, so thank you for that suggestion. Let me know what you think!

ammmze:
> 2\. I can't see much value in adding a config option for this, as it seems to me that the option either works in which case everyone wants it, or it doesn't work in which case no one wants it.  So, not sure how a user can realistically configure it.

I'm good with that...copyright has been added and config option removed.

KevinOConnor:
Thanks.  Seems fine to me.  I'll give a few days to see if there are any further comments and otherwise look to commit.

Cheers,
-Kevin

ammmze:
FYI to fully test this variation of gc fix I did my test print that initially in this endeavor always triggered TTC (because it had many filament swaps within a short time) and I can confirm that the print finished and that the gen 2 collection times looked good the entire time. The highest I saw was just under 12ms. And memory usage was stable at ~140MB.

burkfers:
I have updated to `37403ca` and can report zero issues so far, having subjected the printer to usually ttc-intensive conditions but experiencing none.

cgarwood82:
Also throwing in confirmation that this has resolved time sync issues for me that I was experiencing consistently while calibrating my box turtle. Merging this with a local branch has resolve my issues.

weemantella:
Running led_effects on a CB1 has been inconsistent but TTC have been common. After adding a BoxTurtle running AFC the only way to run well was disabling led_effects and klipper screen. I only have a couple hours running with this PR but reenabling led_effects I've had zero TTC errors. Good work!

KevinOConnor:
Thanks!  I committed this change.  I made some minor change to the commit message to better fit the normal Klipper commit format and to omit references to "plugins" as Klipper does not have a plugin system and I don't want to give an impression that it does.

Cheers,
-Kevin

53Aries:
Just wanted to chime in. Unfortunately I still get TTC errors every now and then. Most common is upon toolhead load or toolhead unload of my boxturtle using Happy Hare. It's always my toolhead board that gives the TTC. Just a FYI really and maybe my logs help. Thanks. 
[klippy - 2025-03-15T011718.799.log](https://github.com/user-attachments/files/19259365/klippy.-.2025-03-15T011718.799.log)
[mmu (8).log](https://github.com/user-attachments/files/19259367/mmu.8.log)

PR #6796, opened 2025-01-26
stepper: optimize event full

nefelim4ag:
Problem: a buddy of mine runs servos with really high microstepping because of some servo controller limitation (like 80000 per rotation). Servo controllers use pulses instead of edges for stepping, so it is easily got Stepper too far in the past.

Currently `stepper_event_full` is much slower than a `stepper_event_edge`.
Like for rp2040 (for 1 stepper and 3 steppers):
- edge code produces 5 ticks (2400K), 22 ticks (1636K)
- event full produces  16 ticks (750K), 61 ticks (590K) (step_pulse_ticks=2)
- event full optimized 15 ticks (800K), 58 ticks (621K)

I also tested this on stm32h7 (at 550Mhz, cause I testing it):
- event full produces 169 (3254K) & 527 (3130K) (step_pulse_ticks=44)
- event full optimized 161 (3416K) & 501 (3293K)

So, on average this is faster by 5%.

Thanks.

---
Btw, I think the main limit right now is the speed of the timer/rescheduler code, so there is another solution.
But I think it can be useful only with a high enough resolution timer (rp2040 feels a little too coarse to guarantee minimal pulse time)

```
@@ -267,6 +305,8 @@ command_config_stepper(uint32_t *args)
     if (HAVE_EDGE_OPTIMIZATION) {
         if (!s->step_pulse_ticks && invert_step < 0)
             s->flags |= SF_SINGLE_SCHED;
           // guess the minimal reschedule time ~ 250ns
+        else if (HAVE_FULL_SHORT && s->step_pulse_ticks <= nsecs_to_ticks(250))
+            s->time.func = stepper_event_full_short;
         else if (HAVE_FULL_OPTIMIZATION)
             s->time.func = stepper_event_full_pstart;
         else
.....
// Uber fast "double scheduled" step function
uint_fast8_t
stepper_event_full_short(struct timer *t)
{
    struct stepper *s = container_of(t, struct stepper, time);
    gpio_out_toggle_noirq(s->step_pin);
    uint32_t curtime = timer_read_time();
    uint32_t min_next_time = curtime + s->step_pulse_ticks;
    s->count--; // same as from current PR
    while (unlikely(timer_is_before(timer_read_time(), min_next_time)));
    gpio_out_toggle_noirq(s->step_pin);
    min_next_time += s->step_pulse_ticks;
    if (likely(s->count)) {
        s->next_step_time += s->interval;
        s->interval += s->add;
        s->time.waketime = s->next_step_time;
        if (unlikely(timer_is_before(s->next_step_time, min_next_time)))
            // The next step event is too close - push it back
            s->time.waketime = min_next_time;
        return SF_RESCHEDULE;
    }
    uint_fast8_t ret = stepper_load_next(s);
    if (ret == SF_DONE || !timer_is_before(s->time.waketime, min_next_time))
        return ret;
    // Next step event is too close to the last unstep
    int32_t diff = s->time.waketime - min_next_time;
    if (diff < (int32_t)-timer_from_us(1000))
        shutdown("Stepper too far in past");
    s->time.waketime = min_next_time;
    return SF_RESCHEDULE;
}
```

This feels superior by a high margin and closer to the expected x2 difference (step_pulse_ticks=2), numbers:
rp2040
- 11 ticks (1090K) 
- 39 ticks (923K)
So, around ~33%

There already was something similar: f8b0c884 in 2017

### DIFF
--- a/src/stepper.c
+++ b/src/stepper.c
@@ -18,15 +18,18 @@
  #define HAVE_SINGLE_SCHEDULE 1
  #define HAVE_EDGE_OPTIMIZATION 1
  #define HAVE_AVR_OPTIMIZATION 0
+ #define HAVE_FULL_SHORT 1
  DECL_CONSTANT("STEPPER_BOTH_EDGE", 1);
 #elif CONFIG_INLINE_STEPPER_HACK && CONFIG_MACH_AVR
  #define HAVE_SINGLE_SCHEDULE 1
  #define HAVE_EDGE_OPTIMIZATION 0
  #define HAVE_AVR_OPTIMIZATION 1
+ #define HAVE_FULL_SHORT 0
 #else
  #define HAVE_SINGLE_SCHEDULE 0
  #define HAVE_EDGE_OPTIMIZATION 0
  #define HAVE_AVR_OPTIMIZATION 0
+ #define HAVE_FULL_SHORT 0
 #endif
 
 struct stepper_move {
@@ -59,6 +62,12 @@ enum {
     SF_LAST_DIR=1<<0, SF_NEXT_DIR=1<<1, SF_INVERT_STEP=1<<2, SF_NEED_RESET=1<<3,
     SF_SINGLE_SCHED=1<<4, SF_HAVE_ADD=1<<5
 };
+
+static unsigned int
+nsecs_to_ticks(uint32_t ns)
+{
+    return timer_from_us(ns * 1000) / 1000000;
+}
 
 // Setup a stepper for the next move in its queue
 static uint_fast8_t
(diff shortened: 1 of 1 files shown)


KevinOConnor:
FYI, if you want to enable the "inline stepper" optimization while not having the "step on both edges" optimization, then just modify the board Kconfig file to not enable `HAVE_STEPPER_BOTH_EDGE`.  See src/pru/Kconfig and src/linux/Kconfig as examples.

Otherwise, could you elaborate on what the code on this PR does?  In general, we'll get better performance if we don't specify a `s->time.func` callback as that allows the code to get inlined into the core scheduler (see `CONFIG_INLINE_STEPPER_HACK`.

Cheers,
-Kevin

nefelim4ag:
> if you want to enable the "inline stepper" 

I ignored it because I thought it would not provide huge benefits.
But you are right it gives a nice boost:
11 and 48 ticks, so ~25% over defaults.

This is more than I expected.

> Otherwise, could you elaborate on what the code on this PR does? 

Several instructions can be cut off, if we just run code unconditionally.
So, here it is, we know `step` can be just executed unconditionally and unconditionally rescheduled with constant time to the unstep function, where we still have to do something with step time calculation and do additional checks.

It _can_ replace stepper_event_full, but it is not compatible with `inline hack`, so I just incorporated a safety net in `stepper_event`.

_And yes, juggling functions look like spaghetti a little._

Now, I know the inlined stepper event is fast, so I'm not sure that cut of several instructions makes sense.

---
On the other hand. for high step rates, I think it is better to avoid reschedule in the middle of the pulse, like if the pulse width is shorter than the time to reschedule.
it is cheaper to unstep in the same event, and the threshold to do so looks equal to a single stepper edge event.

But there is a limitation of timer resolution, it can violate a short pulse width because of integer rounding.
Like 125ns can't be set on rp2040: `125 * (12_000_000 / 1000_000) / 1000 = 1.5`.

Maybe if people already try to modifying their pulse width and trying to get a high step rate, it can be left as is, cause they will not see steps on 125ns, but will start to see them on 167ns.
Right now even on stm32h7, there is a limit around ~2100k steps on default configuration, so any values below 500ns just do not make any sense.

Thanks.

nefelim4ag:
I replaced the code with just an optional short version, it will only work if there a short pulse required and it will be comparable to the performance of inlined step_event_full, for several steppers slightly better. But works alongside with edge and only if needed.

`400ns` is just "short" enough, it is disputable, maybe we want to lower it to only support really short pulses, like <250ns.
I chose it, because of [stm32h7 limits](https://klipper.discourse.group/t/klipper-400mhz-limitations/15807/17?u=nefelim4ag)
Where one stepper can only provide: 171 ticks or 2339K (400Mhz)
`171 / 400_000_000 = 0.000_000_427`

Same performance numbers are actual (step_pulse_ticks=2) rp2040
* Stock 16 (750K), 61 ticks (590K) 
* Inlined 11 and 48 ticks ~25% over defaults.
* Patch 11 ticks (1090K), 39 ticks (923K) ~33% over defaults.

Thanks.

KevinOConnor:
Okay thanks.  For what it is worth I'm inclined to say that the minor speed improvement is not worth the additional code complexity.

Cheers,
-Kevin

PR #6801, opened 2025-01-30
stm32: h7 spi support reload mode & frequency

nefelim4ag:
Currently, there is an issue with spi bus frequency and possibly mode.
If several different devices are present on the bus because of the logical `OR` gate on CFG registers, mode and div will be merged.
So, if they share mode speed will be lower, in my specific case 4Mhz TMC (MBR 6) + 10Mhz encoder (MBR 5), causes div to be 6|5 = 7.

So, this is a simple fix to allow a proper reload of CFG registers.
STM32H7 (at 480Mhz) so expected frequencies:
- 480_000_000 >> (5 + 1) = 7500000
- 480_000_000 >> (6 + 1) = 3750000

There are oscilloscope probes, while both are running:
image
image

Thanks.

### DIFF
--- a/src/stm32/gpio.h
+++ b/src/stm32/gpio.h
@@ -38,7 +38,13 @@ void gpio_adc_cancel_sample(struct gpio_adc g);
 
 struct spi_config {
     void *spi;
-    uint32_t spi_cr1;
+    union {
+        uint32_t spi_cr1;
+        struct {
+            uint8_t div;
+            uint8_t mode;
+        };
+    };
 };
 struct spi_config spi_setup(uint32_t bus, uint8_t mode, uint32_t rate);
 void spi_prepare(struct spi_config config);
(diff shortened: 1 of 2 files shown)


KevinOConnor:
@adelyser - FYI.

-Kevin

D4SK:
Are the stm32h7 chips now running at 480mhz? Used to be set at 400, but maybe I missed this

nefelim4ag:
@D4SK, don't worry, mainline Klipper still runs on 400Mhz.
480 Mhz is just my local modifications, where I tried to push the platform to its limits.
I just pointed it out to show that frequencies are different from mainline ones (3.125Mhz and 6.25Mhz accordingly) to avoid confusion (and I just try to avoid reflashing it too frequently).

If that matters, a small [topic](https://klipper.discourse.group/t/klipper-400mhz-limitations/15807/20) on the discourse about going higher.

KevinOConnor:
Thanks.  If there are no further comments I'll look to commit this in a few days.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6802, opened 2025-01-31
save_variables: Check lowercase variable names

pedrolamas:
This is a follow up on [this Discorse topic](https://klipper.discourse.group/t/save-variable-only-works-with-lower-case-variable-names/21032), ensuring the variable name is converted to lowercase before saving and documenting this change to fix the incorrect behavior.

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20250131: Option `VARIABLE=<name>` in `SAVE_VARIABLE` requires lowercase
+value. For example, `extruder` instead of mixedcase `Extruder` or
+uppercase `EXTRUDER`. Using any uppercase letter will raise an error.
+
 20241203: The resonance test has been changed to include slow sweeping
 moves. This change requires that testing point(s) have some clearance
 in X/Y plane (+/- 30 mm from the test point should suffice when using
--- a/docs/G-Codes.md
+++ b/docs/G-Codes.md
@@ -1201,8 +1201,9 @@ has been enabled.
 
 #### SAVE_VARIABLE
 `SAVE_VARIABLE VARIABLE=<name> VALUE=<value>`: Saves the variable to
-disk so that it can be used across restarts. All stored variables are
-loaded into the `printer.save_variables.variables` dict at startup and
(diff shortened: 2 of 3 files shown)


Sineos:
grafik

From the Discord.

KevinOConnor:
Thanks.

As a question, can we raise an error if there's an issue with the name instead of silently renaming the request?  For example, `if varname.lower() != varname: raise gcmd.error("Variables must not contain upper case")` .  I'm a little concerned that silently renaming the variables could cause confusion.

Cheers,
-Kevin

pedrolamas:
> As a question, can we raise an error if there's an issue with the name instead of silently renaming the request?

Thanks Kevin, the only problem I see with that is for users with existing macros using upper/mixed case variable names: their macros will start to fail, and they won't immediately understand why that is now happening!

But having said that, the suggested approach of using `tolower()` on the variable name will also cause failures on the variable access in existing macros when using upper/mixed case names (that won't exist anymore with this)

Given this, I guess I agree with you that erroring might indeed be the better solution!

KevinOConnor:
Okay, thanks.  If there is a concern this could break existing macros then we should probably add a note to Config_Changes.md as well.

In general it looks good to me.

-Kevin

pedrolamas:
As requested, I've now added a small note to Config_Changes.md

KevinOConnor:
Thanks.  Looks good to me.  I'll give a couple of days to see if there are further comments, and otherwise look to commit.

-Kevin

whyme12:
I dont think there is a problem with transforming the names to lower case. It happens already when a variable is the first time stored using SAVE_VARIABLE.

It cant break macros because the variable key is always read back lower case, which is a problem on its own.
I did a small test:
variables file with following content (manually added)
```
[Variables]
LaLa = 12345
```

a small macro:
```
[gcode_macro TEST]
gcode:
  M118 lala: {printer.save_variables.variables.lala}
  M118 LaLa: {printer.save_variables.variables.LaLa}
  M118 LALA: {printer.save_variables.variables.LALA}
```

only the first print returns a result, the other two are empty.

KevinOConnor:

I agree that silently converting the variable to lowercase is an option, but I'd feel better (at least for now) with reporting an error so that users that may have inadvertently depended on the old behavior are made aware of the change.

Cheers,
-Kevin

PR #6803, opened 2025-02-02
heaters: faster PID controller, with friendly logging

samv:
After demystifying the PID control loop to myself, I thought I would go ahead and make the logs explain the algorithm.  At each adjustment (the rate of which is now configurable), the numbers are printed with their units, along with the amount of the total power is attributed to the P, I and D components.


nefelim4ag:
Can you please elaborate and describe what problem you are trying to solve?

Why do we need to propagate update filtering to a user?
Why do we need to enable debug logging and replace some of them with info?
How will this additional information help?

Thanks

PR #6806, opened 2025-02-03
RFC: Added support for fast MCU

nefelim4ag:
This is the patch set from the discussion: https://klipper.discourse.group/t/klipper-400mhz-limitations/15807

This PR is the group of patches, which fixes or move all known to me limitations, which can cause issues with STM32H7 running with timer frequency, which violates current requirements: `2**32/10` hz
Shortly, allow Klipper to work on MCU where timer overflows in less than 10s.

My personal opinion:
It feels more right to recompute constants instead of just changing them for everyone.

List of required changes, because implementation may change before merge:
- Stats are reported every 5s, which causes computation overflow. 
- Heaters expect a max duration of 5s.
- PWM Tools expect a max duration of 5s.
~~- Step on edge will violate 100ns limit between edges.~~ https://github.com/Klipper3d/klipper/pull/6852
- ~~Software SPI will run faster than 4Mhz - breaks TMC SPI communication.~~ https://github.com/Klipper3d/klipper/pull/6856

My short self-review here:
- Software spi can be limited in different ways, with a wait loop it can run slower than before or maybe enabled only for "fast" MCUs.
- ~~Stats limited to 3s for simplicity (initially, I just recomputed it based on timer frequency).~~ clever fix https://github.com/KevinOConnor/klipper-dev/commit/0553c191d47023a18edc4242dd111ad379d31293
- ~~ClockSync changes added "just to be sure", it should work without them.~~ not required
- PWM max duration reduced to 3.5 because `2000_000_000 / 550_000_000  = 3.63`
- PWM recomputed based on MCU frequency as PoC, to not alter behavior for normal ones. 
- Step on edge limited in the most non-intrusive way that I can find.

Step on edge numbers:
550Mhz: 1 stepper is 49 ticks, so formally there are 11224K. 3 steppers as on the benchmark page - 195 ticks.
On the oscilloscope, there are 210ns per pulse - 105 ns per step ~ 57 ticks.
480Mhz: 1 stepper is 46 ticks, so formally there are 10435K.


KevinOConnor:
Just to be clear, I'm not against making these changes.  I think we should have a good feel for the end result though.  For example, have a goal to move from a maximum of 5 seconds scheduled in the future to a maximum of 3 seconds.  Or perhaps, make the goal more aggressive, and make most areas of the code auto-check the duration.

FWIW, it is not immediately obvious to me the need for the clocksysc change.

FYI, busy waiting in the stepper code seems a bit odd as `INLINE_STEPPER_HACK` basically just moves the busy wait to the scheduling code which should be in the same function.  The gain of busy waiting in the scheduling code is that if something else wants to run it can (and that is pretty common in multi-stepper axis setups).

Cheers,
-Kevin

nefelim4ag:
> (mcu.py:_handle_mcu_stats) or change the downstream users of these stats.

My error, I totally overlooked that `graphstats.py` and other derivatives use 5s, I just assumed they show raw log values.

> So, I guess the big question is: Is the risk of a regression worth the benefits? What are the main benefits?

I agree that it is not worth it, and it should not be broken.

Just my wild suggestion, I did not yet try to implement that, just thoughts.
`stats_update()` is already accumulating the sum of diffs, so we still can calculate 5s and then just report when `sum > timer_from_us(5000000)`.
`uptime high=N` _I'm a little puzzled with the overflow tracking here_. At second glance, even in case of tracking `sum` against `timer_from_us` it should still works.

If it simply works that way, I hope it is okay and does not overcomplicate everything.

> Separately, the software spi change looks like it could be useful regardless. By convention though, we try to avoid board specific logic tests (CONFIG_MACH_STM32) in the common code and try to avoid divide operations. Ideally, the host would tell the mcu how much to delay by on each operation (eg, passing delay ticks instead of rate).

I feel uncomfortable to slow down other MCUs without a reason or force anyone to reflash without a serious reason.
Generally agree.
If it is fine for you, I can make a separate PR with SPI-specific changes where I can suggest implementation, and we can discuss the preferred one. 

> FWIW, it is not immediately obvious to me the need for the clocksysc change.

I tested without them, now it sounds like there is no hidden stuff behind it - it can be dropped.

> FYI, busy waiting in the stepper code seems a bit odd as INLINE_STEPPER_HACK basically just moves the busy wait to the scheduling code which should be in the same function.

You already gave me a good explanation of `INLINE_STEPPER_HACK` in the stepper full PR: https://github.com/Klipper3d/klipper/pull/6796

> The gain of busy waiting in the scheduling code is that if something else wants to run it can (and that is pretty common in multi-stepper axis setups).

Not sure that I get the thought.
There is a documented busy wait in the `timer_dispatch_many()`, I'm aware of it and understand, I hope.

TLDR:
if we track and detect in the `stepper_event_edge` that 100ns is violated.
To exit from the timer code, we need to reschedule.
To run another timer here, we also need to reschedule.
Our target metric is the step rate benchmark.
From my testing, there is just not enough time to do anything except a busy loop.
Busy loop here only runs if we try to violate 100ns.
Most of the time, the busy loop is "free".

Long story:
[There is a patch on discurse](https://klipper.discourse.group/t/klipper-400mhz-limitations/15807/16) with logic from `stepper_full` to reschedule.
[There is a test info](https://klipper.discourse.group/t/klipper-400mhz-limitations/15807/18), it takes 85 ticks at 550Mhz, and 75 at 450Mhz.

My specific busy loop here exists only because rescheduling and/or handling rescheduling in the `stepper_event_edge` adds too much overhead and kills all performance even with `INLINE_STEPPER_HACK`.
There is just not enough time.

In a perfect scenario, [pin toggle takes](https://metebalci.com/blog/stm32h7-gpio-toggling/): `1 / 24_000_000 = 0.000_000_042` + (ODR toggle) Read Modify Write
_Pin toggle speed scales more or less linearly with MCU Clock according to software SPI behaviour.
So, around 17.4Mhz at 400Mhz clock speed._
 
The current code executes in ~44 ticks, so with the highest possible step frequency for stm32h7 right now:
`44 / 550_000_000 = 0.000_000_080`.
1 tick is 1.8 ns.
So, there are ~11 ticks to spend until 100ns.

From above, we know ~half of that is the pin toggle because of MCU architecture.
Assume RWM (ODR) takes 1 additional tick in comparison to BSRR.
`44 - (42/1.8) - 1 = 19.6` ticks or ~36ns is overhead of other code, mostly rescheduling.
In the hypothetical scenario, where that code can reschedule itself for "free", the overhead of rescheduling and load other code to run, will be more than avaliable 20ns.

Another stepper will take at least the same amount of ticks (44).
From tests with different clock speeds, we know that 3 steppers always take 195 ticks.
`195  - 44*3 = 63` or 115ns rough estimation of rescheduling overhead with 3 steppers.

Busy loop, only do something if we violate 100ns restriction.
So, there are only 2 cases where we can try to violate 100ns:
- Clean one-stepper benchmark
- When timer is lagging behind and tries to catch

For the first one, it seems suboptimal as it burns 20% of CPU time, but there is no time to do anything useful instead.
For the second one, from the count of processed timers per second, it still will be faster to execute them with a busy loop until any other timer in the list.

I hope I was able to explain myself.

Thanks.

nefelim4ag:
Stats are now reported by accumulated passed time sum; it looks like working solution.
SW SPI search for effective power of 2 div, similar to STM32 HW SPI.

We still can change the way how SW SPI is configured, so this is for now, the short/simple solution.

KevinOConnor:
> To exit from the timer code, we need to reschedule.

I'm not sure why one would reschedule.  If I'm reading the new code correctly, the basic idea is that `stepper_event_edge()` checks if it has been called to early and busy loops if so.  However, a more natural solution seems to be to not request the scheduler to call `stepper_event_edge()` earlier than it can run.

For example, something like (totally untested):
```
uint_fast8_t
stepper_event_edge_super_fast(struct timer *t)
{
    struct stepper *s = container_of(t, struct stepper, time);
    gpio_out_toggle_noirq(s->step_pin);
    uint32_t curtime = timer_read_time();
    uint32_t min_next_time = curtime + nsecs_to_ticks(100);
    uint32_t count = s->count - 1;
    if (likely(count)) {
        s->count = count;
        s->next_step_time += s->interval;
        s->interval += s->add;
        if (unlikely(timer_is_before(s->next_step_time, min_next_time)))
            goto reschedule_min;
        s->time.waketime = s->next_step_time;
        return SF_RESCHEDULE;
    }
    uint_fast8_t ret = stepper_load_next(s);
    if (ret == SF_DONE || !timer_is_before(s->time.waketime, min_next_time))
        return ret;
    // Next step event is too close
    int32_t diff = s->time.waketime - min_next_time;
    if (diff < (int32_t)-timer_from_us(1000))
        shutdown("Stepper too far in past");
reschedule_min:
    s->time.waketime = min_next_time;
    return SF_RESCHEDULE;
}
```
(Note the above would require associated changes to `stepper_load_next()`.)

The advantage of not busy looping in the stepper code is that the scheduler could make better choices on what event to run next.  Maybe I'm missing something though.

Thanks for working on this.
-Kevin

nefelim4ag:
> For example, something like (totally untested)

This patch is what I referenced above [from the discourse](https://klipper.discourse.group/t/klipper-400mhz-limitations/15807/16). _The first one, that I have suggested initially in thread._
 
 If I compress mine a little, it looks very similar:
 ```
uint_fast8_t
stepper_event_edge(struct timer *t)
{
    struct stepper *s = container_of(t, struct stepper, time);
    gpio_out_toggle_noirq(s->step_pin);
    uint32_t curtime = timer_read_time();
    uint32_t min_next_time = curtime + nsecs_to_ticks(100);
    uint32_t count = s->count - 1;
    if (likely(count)) {
        s->count = count;
        s->next_step_time += s->interval;
        s->interval += s->add;
        // The next step event is too close - push it back
        if (unlikely(timer_is_before(s->next_step_time, min_next_time)))
            s->time.waketime = min_next_time;
        else
            s->time.waketime = s->next_step_time;
        return SF_RESCHEDULE;
    }
    uint_fast8_t ret = stepper_load_next(s);
    if (ret == SF_DONE || !timer_is_before(s->time.waketime, min_next_time))
        return ret;
    // Next step event is too close to the last unstep
    int32_t diff = s->time.waketime - min_next_time;
    if (diff < (int32_t)-timer_from_us(1000))
        shutdown("Stepper too far in past");
    return SF_RESCHEDULE;
}
 ```
 
 Unfortunately, it is just too slow; as I mentioned, numbers will be like 85 ticks at 550Mhz, and 75 at 450Mhz.
 
 > The advantage of not busy looping in the stepper code is that the scheduler could make better choices on what event to run next. Maybe I'm missing something though.
 
I totally get it, but this specific "fast" mcu (stm32h7) is still not fast enough to do something else or execute full code with timing control.
I was unable to optimize that code in any way to not break check for lagging behind.

Right now, I think the only way to squeeze more is to somehow re-implement the scheduling code in the stepper.c
So, all steppers will be run in one stepper timer from the main scheduler code, and internally, pin access can be aliased and/or checked for "too far in past" and can be shared, and overhead on each call can be reduced.

But this is a complete redesign, and I don't think it is worth it or even give any gain in the end.

KevinOConnor:
> Unfortunately, it is just too slow; as I mentioned, [numbers will be like](https://klipper.discourse.group/t/klipper-400mhz-limitations/15807/18) 85 ticks at 550Mhz, and 75 at 450Mhz.

Ah, okay.  I missed that you've already attempted this.

However, I fear your new code may be unrealistically optimized to the benchmark and not to real-world usages.  Sure, it'll be faster in the single stepper case, but the single stepper case is kinda "bonkers" - sending 1+ million steps per second to a single stepper seems absurd.  I suspect the new code is faster because it effectively takes the next step exactly when it can, but that came at the expense of wasting cpu cycles.  That is, scheduling to the actual wakeup time will result in a wakeup that is not exactly the requested time, but it is more capable of utilizing the available cpu resources when there is contention.

Said another way, I think the 3+ stepper benchmark is more interesting, and I suspect the older version of your code (similar to mine above) is no slower in that benchmark.

Remind me - what is your target printer environment that you're optimizing for?

---

Separately, thinking about this more, I think we could enable the "step on both edge optimization" for all mcus on the slow path.  Something like:
```diff
--- a/src/stepper.c
+++ b/src/stepper.c
@@ -146,7 +146,7 @@ stepper_event_full(struct timer *t)
     uint32_t curtime = timer_read_time();
     uint32_t min_next_time = curtime + s->step_pulse_ticks;
     s->count--;
-    if (likely(s->count & 1))
+    if (s->count & 1 && s->flags & SF_NEED_UNSTEP)
         // Schedule unstep event
         goto reschedule_min;
     if (likely(s->count)) {

```
The idea is that the "slow path" (ie, `stepper_event_full()`) is already so slow that checking for "step on both edge" is a minor cost, even if it isn't enabled.  However, if we have to use the slow path and we can use "step on both edge" then we can basically double performance.

This wouldn't heavily optimize the >400Mhz case, but at least disabling HAVE_STEPPER_BOTH_EDGE wouldn't result in half the performance..

-Kevin

nefelim4ag:
> I fear your new code may be unrealistically optimized to the benchmark and not to real-world usages. Sure, it'll be faster in the single stepper case, but

Yep

> I suspect the older version of your code (similar to mine above) is no slower in that benchmark.

You are right, it is pretty similar.

> Remind me - what is your target printer environment that you're optimizing for?

My initial motivation was "Stepper too far" on the buddy printer on servos. So, there are stepper_full PR: https://github.com/Klipper3d/klipper/pull/6796
Higher frequency will be only useful for magento x or printers with servos, where they can run fast, but microsteps are small, and the slow path is too slow sometimes.
(Like Magneto X just disabled that check if firmware, which sounds silly.)

> Separately, thinking about this more, I think we could enable the "step on both edge optimization" for all mcus on the slow path.

If I got it correctly, it is a brilliant idea.
That way, the edge code can be disabled, and full step can be used instead, inlined, used for both.
Without not pretty additions to edge code.

I will take a look.

nefelim4ag:
I pushed mine, I will test and update it, and post some numbers; it should be backward compatible.
I see your refactor - so dropped `HAVE_SINGLE_SCHEDULE`.

new is a combined stepper_full().

RP2040 (200Mhz) (I debug on it :D)
 ""                | edge | full | new edge | new full 
--                 | --        | --     | --                | --
1 stepper   | 3       | 10    | 3                | 8
3 steppers | 14     | 38    | 14              | 31

STM32H7 (480Mhz) (the left is semi old data. I did not retest, I forgot)
 ""                | edge     | full    | new edge | new full 
--                 | --            | --        | --                | --
1 stepper   | 46          | 184   | 78              | 166
3 steppers | 196        | 535    | 225            | 435
1 step rate | 10436K | 2608K | 6153K     | 2891K
3 step rate | 7346K   | 2691K | 6400K     | 3310K

Smoke tested, my printer is able to home itself.

KevinOConnor:
Interesting, but I'm not sure I understand your numbers.  What is "new edge"?  Is it the inlined `stepper_event_full()` with "combined" support?  If so, those are impressive numbers (3-stepper edge of 6400k vs stock 6061K on stm32h7), but the numbers look strange to me.  I would not expect rp2040 to have the same performance as stock, and I would not expect the "new edge" to be more than twice as fast as "new full".

-Kevin

nefelim4ag:
> What is "new edge"? Is it the inlined stepper_event_full() with "combined" support?

You are correct.

> If so, those are impressive numbers (3-stepper edge of 6400k vs stock 6061K on stm32h7)

They only make sense with a high MCU clock frequency.
With the default 400Mhz, I suspect they will be closer to 5k.

> I would not expect rp2040 to have the same performance as stock, and I would not expect the "new edge" to be more than twice as fast as "new full".

I suspect RP2040, just does not have enough timer resolution in comparison to a 200Mhz clock to measure difference there.
`step_pulse_ticks` will also somewhat limit the maximal number here.
Because it will force to reorder steps on CPU contention or wait in scheduler busy loop.
Probably with 12Mhz timer and 100ns rescheduling, `step_pulse_ticks == 1`, so contention, reordering and busy loop waiting is limited. 

FWIW, as I already mentioned, rescheduling is not free. This is why I tried to hack around stepper_full in https://github.com/Klipper3d/klipper/pull/6796#issuecomment-2624955800.
With the full event, twice as many are rescheduled ops. So, the overhead is just bigger.
I suspect the rescheduling cost around 30ns from the data above and from the data of the stepper_full PR.

Thanks.

KevinOConnor:
I ran some tests on mcus I had handy.  I'm seeing different results when using edge on the "combined" `stepper_event_full()`.

Tests were done on both the master branch ( fbd5b492 ) and on the code at https://github.com/KevinOConnor/klipper-dev/tree/work-stepperedge-20250309 ( commit b918fe896d5c3c71988e3012133fb144b9f718eb ).  It's slightly different code than you have, but I'd expect similar benchmark results.

The "edge" test was with `invert_step=-1`, and "combined" test was with `invert_step=0`.  All tests were with `step_pulse_ticks=0`, but I didn't notice any difference in spot tests if I set it to 100ns on these mcus.

All tests ran were with `arm-none-eabi-gcc (Fedora 14.1.0-1.fc40) 14.1.0`

mcu | 1-stepper edge | 3-stepper edge | 1-stepper full | 3-stepper full
-- | -- | -- | -- | --
rp2040 master | 3 | 14 | 10 | 38
rp2040 combined | 4 | 17 | 7 | 31
rp2350 master | 36 | 169 | 132 | 514
rp2350 combined | 48 | 207 | 87 | 379
stm32f446 master | 46 | 211 | 170 | 646
stm32f446 combined | 65 | 269 | 119 | 500

So, I always see a slight performance degradation when using "edge" on combined (compared to edge on master).  Like you, I see a notable performance benefit to "full" when on combined (compared to stock master).

Note, after testing various permutations, I found using the hot-path test of `if (likely(s->count & 1 && !(s->flags & SF_SINGLE_SCHED)))` provided the best performance (for both edge and full).  It's how I was able to slightly improve on your rp2040 combined "full" result (7 vs 8).

In general, the results look very promising.
-Kevin

nefelim4ag:
I switched to your branch from klipper-dev.
All tests ran were with `arm-none-eabi-gcc (15:12.2.rel1-1) 12.2.1 20221205`
combined (b918fe89)

edge (invert_step=-1)
full (invert_step=0)

MCU                                              | 1 stepper full pulse_ticks=0    | 3 stepper full pulse_ticks=0
--                                                    | --                                                   | --
STM32H7 (400Mhz) combined | 106                                              | 430

1 stepper full pulse_ticks=0 106 ticks, 3 stepper full pulse_ticks=0 430
1 stepper full pulse_ticks=40 140 ticks, 3 stepper full pulse_ticks=40 430
1 stepper edge pulse ticks=0 53 ticks, 3 stepper edge pulse_ticks=0 221
1 stepper edge pulse_ticks=40 73 ticks, 3 stepper edge pulse_ticks=40 221

KevinOConnor:
Interesting.  I pulled out my stm32h723 board and ran some tests:

Same gcc as above, same revisions for "master" and "combined".  The "480mhz" version is with commit https://github.com/KevinOConnor/klipper-dev/commit/0553c191d47023a18edc4242dd111ad379d31293 - note I've used a slightly different stats patch.  The "spt=40" is with `step_pulse_ticks=40` and "spt=48" is with `step_pulse_ticks=48`.

mcu | 1-stepper edge | 3-stepper edge | 1-stepper full | 3-stepper full
-- | -- | -- | -- | --
stm32h723 master | 44 | 186 | 142 | 526
stm32h723 master spt=40 | | | 160 | 526
stm32h723 combined | 54 | 225 | 108 | 438
stm32h723 combined spt=40 | 74 | 225 | 151 | 438
stm32h723 combined 480mhz | 52 | 224 |106 | 437
stm32h723 combined 480mhz spt=48 | 80 | 224 | 166 | 437

-Kevin

nefelim4ag:
JFYI:
> The "480mhz" version is with commit https://github.com/KevinOConnor/klipper-dev/commit/0553c191d47023a18edc4242dd111ad379d31293 ...

stats, tested, much better than mine - lgtm.

> I suspect that because I tested with step_pulse_ticks == 48 (100ns), the code spent some of the time in the return path, hit the busy loop: ...

 I've tried to measure launch overhead between the busy loop and the actual timer call:
 ```
 timer_dispatch_many():
         while (unlikely(diff > 0))
            diff = next - timer_read_time();
            ...
sched_timer_dispatch():
            res = t->func(t);
 ```
 To try to subtract overhead here, like:
 `diff = next - timer_read_time() - overhead;`
 
 So, I've tried to measure the difference between `timer->waketime` and `timer_read_time()`.
 I've expected something around stm32h7 ~10 ticks, rp2040 ~0 ticks.
 MCU | Ticks | Nanosec | Reproducible
  -- | -- | -- | --
  rp2040 | 0 | < 1000 | Y
  Linux MCU | 50..500 | > 1000 | N
  stm32h7 (480Mhz) | 9 | 18.7 | Y  
   
 It may give another ~10% with non zero `step_pulse_ticks`.
So, just to consider, it is possible to measure it after a board init, maybe.
And just use it during runtime.
It will spare some performance and reduce static lag, but will add complexity.

 <details>
<summary>Measure code</summary>

```
diff --git a/src/basecmd.c b/src/basecmd.c
index 9d5e58f9..efbd16ec 100644
--- a/src/basecmd.c
+++ b/src/basecmd.c
@@ -234,6 +234,40 @@ command_allocate_oids(uint32_t *args)
 }
 DECL_COMMAND(command_allocate_oids, "allocate_oids count=%c");
 
+/****************************************************************
+ * Scheduler tune
+ ****************************************************************/
+
+struct {
+    struct timer timer;
+    uint32_t overhead;
+    uint8_t counter;
+} sched_overhead;
+
+static uint_fast8_t
+measure_event(struct timer *timer)
+{
+    uint32_t cur = timer_read_time();
+    sched_overhead.counter++;
+    if (sched_overhead.counter < 4)
+        sched_overhead.overhead += cur - timer->waketime;
+    if (sched_overhead.counter < 20) {
+        timer->waketime += timer_from_us(1) / 2 ? timer_from_us(1) / 2 : timer_from_us(1);
+        return SF_RESCHEDULE;
+    }
+    sched_overhead.overhead = sched_overhead.overhead >> 4;
+    sendf("sched overhead=%u", sched_overhead.overhead);
+    return SF_DONE;
+}
+
+void
+measure_sched(void){
+    sched_overhead.timer.func = measure_event;
+    sched_overhead.timer.waketime = timer_read_time() + timer_from_us(100);
+    sched_overhead.counter = 0;
+    sched_overhead.overhead = 0;
+    sched_add_timer(&sched_overhead.timer);
+}
 
 /****************************************************************
  * Config CRC
@@ -252,6 +286,7 @@ DECL_COMMAND_FLAGS(command_get_config, HF_IN_SHUTDOWN, "get_config");
 void
 command_finalize_config(uint32_t *args)
 {
+    measure_sched();
     move_finalize();
     config_crc = args[0];
 }
```

</details>

KevinOConnor:
FYI, I was looking at something similar.  It turns out that `gpio_out_toggle_noirq()` is absurdly slow on the stm32h7.  It looks like it is taking around 15 cpu cycles - while most mcus can do a toggle in 1-4 ticks.

It seems the gpio registers are on the AHB4 bus, which is on the AHB3 bus, which is on the AXI bus, which is what the cpu accesses.  Because the toggle on stm32 requires a read/modify/write operation, the read of the gpio output register state blocks the cpu.

That's contributing to the weird scheduling delays.  If one were to do something like:
```diff
--- a/src/stepper.c
+++ b/src/stepper.c
@@ -144,8 +144,8 @@ static uint_fast8_t
 stepper_event_full(struct timer *t)
 {
     struct stepper *s = container_of(t, struct stepper, time);
-    gpio_out_toggle_noirq(s->step_pin);
     uint32_t curtime = timer_read_time();
+    gpio_out_toggle_noirq(s->step_pin);
     uint32_t min_next_time = curtime + s->step_pulse_ticks;
     s->count--;
     if (likely(s->count & 1 && !(s->flags & SF_SINGLE_SCHED)))
```
Then you'd probably not notice the delay.

Or more accurately:
```diff
--- a/src/stepper.c
+++ b/src/stepper.c
@@ -144,8 +144,12 @@ static uint_fast8_t
 stepper_event_full(struct timer *t)
 {
     struct stepper *s = container_of(t, struct stepper, time);
+    uint32_t curtime;
+    if (CONFIG_HAVE_STRICT_TIMING)
+        curtime = timer_read_time();
     gpio_out_toggle_noirq(s->step_pin);
-    uint32_t curtime = timer_read_time();
+    if (!CONFIG_HAVE_STRICT_TIMING)
+        curtime = timer_read_time();
     uint32_t min_next_time = curtime + s->step_pulse_ticks;
     s->count--;
     if (likely(s->count & 1 && !(s->flags & SF_SINGLE_SCHED)))
```

But as a whole I don't think it's worth the complexity to work around it.  It only seems to impact the 1-stepper case, which is not a very realistic benchmark anyway.

One could also, in theory, replace all the `gpio_out_toggle_noirq()` calls with explicit `gpio_out_write()` calls on stm32h7, but that seems way overly complex to workaround this one goofy mcu.

Cheers,
-Kevin

KevinOConnor:
Heh - world's biggest hack:
```diff
--- a/src/stm32/gpio.c
+++ b/src/stm32/gpio.c
@@ -86,16 +86,20 @@ gpio_out_setup(uint32_t pin, uint32_t val)
     return g;
 }
 
+static uint32_t ODR_HACK[16];
+
 void
 gpio_out_reset(struct gpio_out g, uint32_t val)
 {
     GPIO_TypeDef *regs = g.regs;
     int pin = regs_to_pin(regs, g.bit);
     irqstatus_t flag = irq_save();
+    uint32_t *odr_hack = &ODR_HACK[((uint32_t)g.regs >> 10) & 0x0f];
     if (val)
-        regs->BSRR = g.bit;
+        *odr_hack |= g.bit;
     else
-        regs->BSRR = g.bit << 16;
+        *odr_hack &= ~g.bit;
+    regs->ODR = *odr_hack;
     gpio_peripheral(pin, GPIO_OUTPUT, 0);
     irq_restore(flag);
 }
@@ -104,7 +108,9 @@ void
 gpio_out_toggle_noirq(struct gpio_out g)
 {
     GPIO_TypeDef *regs = g.regs;
-    regs->ODR ^= g.bit;
+    uint32_t *odr_hack = &ODR_HACK[((uint32_t)g.regs >> 10) & 0x0f];
+    *odr_hack ^= g.bit;
+    regs->ODR = *odr_hack;
 }
 
 void
@@ -119,10 +125,14 @@ void
 gpio_out_write(struct gpio_out g, uint32_t val)
 {
     GPIO_TypeDef *regs = g.regs;
+    uint32_t *odr_hack = &ODR_HACK[((uint32_t)g.regs >> 10) & 0x0f];
+    irqstatus_t flag = irq_save();
     if (val)
-        regs->BSRR = g.bit;
+        *odr_hack |= g.bit;
     else
-        regs->BSRR = g.bit << 16;
+        *odr_hack &= ~g.bit;
+    regs->ODR = *odr_hack;
+    irq_restore(flag);
 }
 
 
```

It actually improves the performance.  Test of 480Mhz combined spt=0 "edge" stepper-1/stepper-3 went from 52/224 to 41/187.

-Kevin

nefelim4ag:
> It actually improves the performance. Test of 480Mhz combined spt=0 "edge" stepper-1/stepper-3 went from 52/224 to 41/187.

Wow, that's impressive and unexpected.
I was aware of slow GPIO performance on h7,
But I eliminated the idea of using BSRR, assuming RMW is just ~3 ticks, and any magic workaround BSRR will be the same.

KevinOConnor:
FYI, I cleaned up the ODR caching code, implemented some further optimizations, and added it to my work-stepperedge-20250309 branch ( https://github.com/KevinOConnor/klipper-dev/commit/85924af7da0e9588dc5ac987a0d22f30762feb25 ).

Test of 480Mhz combined spt=0 "edge" stepper-1/stepper-3 at 39/183.

-Kevin

nefelim4ag:
FYI, I've applied https://github.com/KevinOConnor/klipper-dev/commit/85924af7da0e9588dc5ac987a0d22f30762feb25
Tested, lgtm

SW SPI 4.6Mhz, 3.2Mhz, 1.7Mhz, to my great surprise sligtly slower then [before (4.71Mhz, 3.25Mhz, and 1.72Mhz)](https://github.com/Klipper3d/klipper/pull/6806#issuecomment-2708967579)
Not a big deal, but I'm surprised.

480Mhz spt=1 (because over my branch, effectively the same)
w+ and w- oscilloscope probe data, 387ns/391ns swapped randomly.
Edge

test                        | ticks | w+        | w- 
--                             | --       | --          | -- 
1 stepper  spt=1  | 38     | 87.5ns | 100ns
3 stepper  spt=1  | 180   | 391ns  | 387ns
1 stepper  spt=48| 66     | 150ns | 158ns
3 stepper  spt=48| 180   | 391ns | 387ns

So, there is ~50ns overhead of time to run after the busy loop (28 ticks).

Those numbers are just insane.
Like those are already more than can be utilized.

---
Well, I played with measure and compensating time to run after the busy loop.
With -28 offset pulses are around 97-103ns.

spt=48 sched overhead=14 (it is automatic, half to be on the safe side)
test                           | ticks | w+        | w- 
--                               | --       | --          | -- 
1 stepper  spt=48  | 51     | 117ns | 129ns
3 stepper  spt=48  | 180   | 387ns  | 391ns

I don't think it makes much sense here, with those numbers.
I hope it can be useful for weaker MCUs.
There is a scheduler measurement patch.

<details>

```
diff --git a/src/basecmd.c b/src/basecmd.c
index 9d5e58f9e..d68efd42b 100644
--- a/src/basecmd.c
+++ b/src/basecmd.c
@@ -252,6 +252,7 @@ DECL_COMMAND_FLAGS(command_get_config, HF_IN_SHUTDOWN, "get_config");
 void
 command_finalize_config(uint32_t *args)
 {
+    sendf("sched overhead=%u", ttr_overhead());
     move_finalize();
     config_crc = args[0];
 }
diff --git a/src/generic/armcm_timer.c b/src/generic/armcm_timer.c
index e77081804..4f3b68674 100644
--- a/src/generic/armcm_timer.c
+++ b/src/generic/armcm_timer.c
@@ -147,7 +147,7 @@ timer_dispatch_many(void)
         // Next timer in the past or near future - wait for it to be ready
         irq_enable();
         while (unlikely(diff > 0))
-            diff = next - timer_read_time();
+            diff = next - timer_read_time() - ttr_overhead();
         irq_disable();
     }
 }
diff --git a/src/sched.c b/src/sched.c
index 51d159bc6..14020141f 100644
--- a/src/sched.c
+++ b/src/sched.c
@@ -336,6 +336,50 @@ sched_shutdown(uint_fast8_t reason)
     longjmp(shutdown_jmp, reason);
 }
 
+/****************************************************************
+ * Measure timer to run overhead
+ ****************************************************************/
+static uint32_t overhead;
+
+uint32_t
+ttr_overhead(void)
+{
+    return overhead;
+}
+
+static uint_fast8_t
+measure_event(struct timer *t)
+{
+    uint32_t cur = timer_read_time();
+    uint32_t diff = cur - t->waketime;
+    uint32_t repeat = timer_from_us(1);
+    if (repeat > 1)
+        repeat = repeat / 2;
+
+    static uint8_t counter;
+    static uint32_t min = UINT32_MAX;
+    counter++;
+    if (min > diff)
+        min = diff;
+    if (counter < 8) {
+        t->waketime += repeat;
+        return SF_RESCHEDULE;
+    }
+    overhead = min >> 1;
+    sendf("sched overhead=%u", overhead);
+    return SF_DONE;
+}
+
+static struct timer measure_timer = {
+    .func = measure_event,
+};
+
+void
+sched_measure(void)
+{
+    measure_timer.waketime = timer_read_time() + timer_from_us(100000);
+    sched_add_timer(&measure_timer);
+}
 
 /****************************************************************
  * Startup
@@ -356,5 +400,6 @@ sched_main(void)
         run_shutdown(ret);
     irq_enable();
 
+    sched_measure();
     run_tasks();
 }
diff --git a/src/sched.h b/src/sched.h
index 9ec96631e..8c792cfd1 100644
--- a/src/sched.h
+++ b/src/sched.h
@@ -40,6 +40,7 @@ void sched_try_shutdown(uint_fast8_t reason);
 void sched_shutdown(uint_fast8_t reason) __noreturn;
 void sched_report_shutdown(void);
 void sched_main(void);
+uint32_t ttr_overhead(void);
 
 // Compiler glue for DECL_X macros above.
 #define _DECL_CALLLIST(NAME, FUNC)     
```

</details>

KevinOConnor:
Interesting, thanks.

> Well, I played with measure and compensating time to run after the busy loop.

FWIW, with PR #6852 the `step_pulse_ticks` comes directly from the user `step_pulse_duration` configuration option.  So, instead of an automatic compensation, a user could just decrease the `step_pulse_duration` to get the actual pulse width that they desire.  That is, if a `step_pulse_ticks=48` actually results in a delay of 76 ticks, then the user could arrange to configure the driver with `step_pulse_ticks=20` (to effectively get the desired 48 ticks).

FWIW, though, I can't see anyone doing the above as the risk vs reward is not appealing.  Get it just right and you can send more than 6 million steps to a single driver (useless) and get it wrong and you'll be dropping steps (very bad).

I suppose if someone was really interested in this, one could build an automated calibration tool that detects the minimum pulse duration needed.  The tool could put the tmc driver in 256 microstep mode, disable the motor power in the driver, send 1000 steps at a low step_pulse_ticks, and verify that the drivers MSCNT field moves by exactly 1000.  Repeat until one finds the minimum pulse duration.  Probably useless as an optimization, but might be useful for verifying that the driver can actually receive pulses at 100ns (as board traces may have sufficient capacitance/inductance that the driver actually needs more than 100ns).  All in all, probably not worth the effort.

Cheers,
-Kevin

nefelim4ag:
FWIW,
You did all the hard work already with https://github.com/Klipper3d/klipper/pull/6852 and https://github.com/Klipper3d/klipper/pull/6853
It would be useful to also add your stats patch in some PR: https://github.com/KevinOConnor/klipper-dev/commit/0553c191d47023a18edc4242dd111ad379d31293
Or I can cherry-pick them to this PR, if that is okay.

That will leave us with only SW SPI & PWM 5s -> 3s.
They seem small, and in their current state, I think they can be handled in this PR.

Then I can add change to 480Mhz, if that still makes sense.

What do you think?

KevinOConnor:
I'd move the software spi timing to a new PR.  I think we want to discuss changing the host/mcu interface to pass on delay ticks (the host should be able to auto-detect the change so we can keep backwards compatibility if desired).  I'm also not sure about the timing in that patch - should an irq occur between timer read and gpio update, it could cause the externally observed duration between pulses to not meet the minimum.  In any case, it seems orthogonal to the changes here.

Feel free to cherry-pick my changes.

Cheers,
-Kevin

KevinOConnor:
Also, it seems the stm32h723 can go up to 550Mhz.  Out of curiosity, what is the significance of 480Mhz?

-Kevin

nefelim4ag:
> Also, it seems the stm32h723 can go up to 550Mhz. Out of curiosity, what is the significance of 480Mhz?

There are known issues to me:
- Split MCU Kconfig H723 (550Mhz) and H743 (480Mhz).
- HW I2C should use a different nominal clock because of presc overflow.
- IIRC, I got random errors on SPI.

To summarize, 480Mhz is just much easier to get working right now.

EmperorArthur:
I have been running this PR along with #6853, and adjusting the frequency to 480Mhz.  It fixed the issues I was having with my Magneto X.

No problems so far.

nefelim4ag:
> What's remaining on this? Is it just changing the mcu stats, and changing the host to stop using 5 seconds?

That is it.

> Also, I was looking at the stm32h7 specs and it seems the pll is limited to 836mhz. Does this mean the pll configuration needs to change to go above 418Mhz cpu speed?

I also stumbled upon this. https://klipper.discourse.group/t/klipper-400mhz-limitations/15807/32

For some reason, it does work at 480Mhz (960Mhz PLL) without any changes.
It would be correct to reconfigure the PLL according to the datasheet for the STM32H723.
```
    uint32_t pll_freq = CONFIG_CLOCK_FREQ * 2;
    // H723 has a different VOC limit 836Mhz
    if (CONFIG_MACH_STM32H723 && pll_freq > 836000000) {
        pll_freq = CONFIG_CLOCK_FREQ;
    }
 ```
 Should do the trick.

PR #6813, opened 2025-02-08
ldc1612: handle i2c errors

nefelim4ag:
I feel somewhat obligated to finish i2c error handling, or at least push it a little forward after https://github.com/Klipper3d/klipper/pull/6689

I got some idea of how error handling happens around accelerometers, and can incorporate it into other i2c bulk data sensors.

So, here is a suggestion for LDC1612, the host code already handles cases where there are errors bit set in MSB. Then it should be easy to unconditionally set any of them, if i2c got an error, because `ret` is more than 0 on error, and less than 8.
```
    def _convert_samples(self, samples):
        ...
        for ptime, val in samples:
            mv = val & 0x0fffffff
            if mv != val:
                self.last_error_count += 1
```
If it would be better, we can manually hardcode enum values for consistency, to avoid assumptions.

`query_status` handling is done in the same way as in the `sensor_adxl345`.
Technically `> 64` would be enough (`1 << 6 == 64`).
image

Not tested, because I just don't have ldc1612.

Thanks.


ZeSlammy:
So ELI5, would this work on Temp, pressure, Gaz sensors plugged onto a PCB for a Stealth Max filter ?
I somehow have NACK i2C errors and they are seen as critical even in Kalico/DangerKlipper

nefelim4ag:
> So ELI5

This PR will touch only inductive bad sensors, like BTT Eddy.
So, not yet.
I hope, there will be a series of patch sets with error handling in different places, where it is easy to do with the current code infrastructure. But that will not touch Temp, Pressure, or Gaz sensors.

To generally handle I2C errors, as discussed in the previous topic,
there should be some command changes, and forwarding error information back to the host where they will be handled (possibly retried once). This can break something, this is a lot of work and will take several iterations.

> Temp, pressure, Gaz sensors plugged onto a PCB for a Stealth Max filter ?
> I somehow have NACK i2C errors and they are seen as critical even in Kalico/DangerKlipper

I don't know anything about your setup and think you can try to ask for help on the discourse about what you can do.
If you have separate MCUs for sensors and they do not control heaters.
I can only suggest something like:
```
 void i2c_shutdown_on_err(int ret)
 {
+    return;
     switch (ret) {
     case I2C_BUS_NACK:
         shutdown("I2C NACK");
```
To disable I2C errors completely. This is bad, this is a hack and if you don't know how to do it, you should not.

Hope it helps.

thijstriemstra:
> I feel somewhat obligated to finish i2c error handling

Thanks!

KevinOConnor:
Thanks for working on this.  I'll try to test this on my ldc1612.  Alas, I'm quite behind on projects so it may take me some time.

Cheers,
-Kevin

Supergun777:
Patch applied, everything works as before, don't know how to reproduce I2C errors

nefelim4ag:
Just rebased, I did recheck the code if there is a appropriate error handling on in the LDC1612 code on the MCU and the Host side.

Seems it also should correctly report "Eddy current sensor error" on I2C communication error during the homing procedure.
```
    // High 4 bits are for errors
    d[0] |= ret << 4;

    // Check for endstop trigger
    uint32_t data =   ((uint32_t)d[0] << 24)
                    | ((uint32_t)d[1] << 16)
                    | ((uint32_t)d[2] << 8)
                    | ((uint32_t)d[3]);
    check_home(ld, data);
 ....
check_home(struct ldc1612 *ld, uint32_t data)
    if (data > 0x0fffffff) {
        // Sensor reports an issue - cancel homing
        ld->homing_flags = 0;
```

Thanks.

nefelim4ag:
Taking into account that I now have the LDC, it would probably make sense to utilize error bits (we actually do intentionally enable them and report them to the host).

I think that for now, the handling of the status can be okay. For data, I'm not sure about a good solution.
Probably it can be some special value like `x0` or `0xdeadbeef`.
For now, I feel like setting sensor state bits inside data report with I2C errors may not be a great idea.

nefelim4ag:
I spent some more time with sensors and all of the application notes.
I guess we can utilize the under/over range bits: https://www.ti.com/lit/an/snoa959/snoa959.pdf

Under-range can't happen because we do not use offset, and if we do want it can be done in software to compact the bulk data output, and/or pack more status bits.

_My (carto) board would output something about raw values in the range: 30_000_000...40_000_000. BTT Eddy, AFAIK, uses 12MHz, and has a similar frequency range. So the raw values would be twice as much. 60_000_000..80_000_000._

Overrange, on the other hand, is always `0xfffffff` = 268435455.

Watchdog/Zero count is something that should not happen, and honestly, both of them sound fatal to me.

nefelim4ag:
image

Probably I've found an appropriate solution.
Over/Under range are mutually exclusive, and even if they happen - they can be decoded by the value itself.
So they can be used for error reporting.
The `Zero count` is currently not reported anywhere, and I would not bother to report it specifically.
It should be handled within watchdog error.

The only inconvenience is that there is low/high amplitude error aliasing.
It is possible, though, to encode status and errors in 15 distinct values (4 bits). But I do not see that it is neccessary curently.
If the driver current is calibrated (it should be), then there are predictable results of where amplitude errors happen.

Thanks.

nefelim4ag:
Patch works.
```
LDC1612 starting 'carto' measurements
LDC1612: I2C IO error
LDC1612: I2C IO error
```

_BTW, watchdog errors can be produced by shortening the coil._

Thanks.

nefelim4ag:
After the https://github.com/Klipper3d/klipper/pull/7125#discussion_r2628992308
I prototyped the min side safe coil frequency guard with use of frequency offset and Under-range error.

Well, the only downside, if frequency goes in this range (< offset). Sensor not only response with Under-range errors.
But also seems to return I2C NACKs randomly. One time it even hangs, so upon FIRMWARE_RESET the read of DEVICE_ID failed with NACK.

Thats unfortunate.

-Timofey

nefelim4ag:
@chengxg,
After your [comment](https://github.com/Klipper3d/klipper/pull/7125#discussion_r2629422992), I implemented some error checking.

I will be glad if you can check how it will behave now.
There is still no direct watchdog implementation, though.
But I hope Zero Count should cover the case where you disconnected the sensor and changed the settle time.

Thanks!

nefelim4ag:
Overall, I refactored it and tested that I can.
As the overall consensus looks to me, we still want to be able to see Amplitude errors.
So, they are logged along with any other possible errors.
But now they are ignored during homing.

I2C errors and Zero Count errors should force data readout. (IDK if Zero count triggers DRDY flag).
So, upon those errors, error checking will be triggered.
That relaxes a needs of watchdog for now.

The upper limit for the frequency is decreased to half of the reference frequency.
Probably, in the future, it would be feasible to decrease it further to 1/4 of the reference frequency.
The limitation factor here is BTT Eddy, which is out of spec, with coil frequency > F_ref / 4.

Low-frequency guard https://github.com/Klipper3d/klipper/commit/d6f4e9745e20f930f85b104a7e52c249ca5e5c95 is dropped for now, simply because it looks ugly to my eye, adds even more complexity, and so I wanted to limit the overall scope of changes in this PR to a digestible amount.

Thanks,
-Timofey

nefelim4ag:
Okay, I've refactored it a little.
Added some docs which should describe errors for people who are not me.
It is honestly hard to check and reproduce everything.

Verified to happen:
- Firmware watchdog - with no clock signal or with low ODR.
- Over range - reducing clock input closer to sensor frequency (24MHz -> 4MHz in my case).
- Conversion watchdog - by closing the coil pins
- ~Zero count - by closing the coil pins~ 
- Over range - by closing the coil pins
- I2C IO error - by closing of the I2C pins.

Messages in the log looks like so:
```
ldc1612: Amplitude Low/High (24)
ldc1612: Amplitude Low/High (36)
ldc1612: Amplitude Low/High (24)
ldc1612: Frequency over valid range (3)
ldc1612: Conversion Watchdog timeout (1)
ldc1612: Frequency over valid range (3)
```

To reduce amplite warnings rate, we can technically disable the amplitude high bit, I think.
But for now messages are still here.

That is it on my side.

Thanks,
-Timofey

---
I removed the zero count, because I can't reproduce there is no good place to encode it.

KevinOConnor:
Thanks.  In general I think it is a good idea to check for these errors and to properly report them from the host software.  However, I would ask for some changes to this PR.

My main concern is that I fear some of the changes here would make the mcu ldc1612.c code closely tied to the host ldc1612.py code in a way that would be hard to maintain.  That is, I fear future developers making minor changes to the mcu ldc1612.c could inadvertently break subtle expectation of the host code.  I think it would be preferable if we could improve the host/mcu interface in this PR to facilitate future changes.

I have some more specific comments, which I will follow up with.

Cheers,
-Kevin

klippy/extras/ldc1612.py
```diff
@@ -161,12 +161,25 @@ def clear_home(self):
...
+            if not errors.get(msg, 0):
+                errors[msg] = 0
+            errors[msg] += 1
         for ptime, val in samples:
             mv = val & 0x0fffffff
             if mv != val:
+                flags = val >> 28
+                if (flags & (0x8 | 0x4)) == (0x8 | 0x4):
+                    log_once("I2C IO error")
+                    continue
```
KevinOConnor: I wonder if we could improve this error reporting interface so that it is more "regular".  Perhaps we could specify that if the top 15 bits are all ones (`if mv >> 16 == 0xffff`) it indicates a specific error condition code is in the low bits (`error_code = mv & 0xffff`).  And then use the enumeration system to lookup a string for the error (`error_enums = mcu.get_enumerations("ldc1612_sensor_errors:")` ... `error_msg = error_enums.get(error_code); log_once(error_msg)`).

It is possible that the chip could send a message with all high bits set.  In that case, the mcu could detect that condition and report an error for that (`if (data >> 16 == 0xffff) { error_code = MY_DATA_ERROR; }` ... `DECL_ENUMERATION("ldc1612_sensor_errors:", "sensor data invalid" MY_DATA_ERROR);` ).

Normally mcu enumerations describe parameters in host/mcu messages, but we could use a special character (eg, ":") to denote an enumeration not intended to directly describe a message parameter.

klippy/extras/probe_eddy_current.py
```diff
@@ -383,6 +383,7 @@ def note_probe_and_position(self, start_time, end_time, pos_time):
 # Helper for implementing PROBE style commands (descend until trigger)
 class EddyDescend:
     REASON_SENSOR_ERROR = mcu.MCU_trsync.REASON_COMMS_TIMEOUT + 1
+    REASON_BUS_IO_ERROR = REASON_SENSOR_ERROR + 12
```
KevinOConnor: I have a similar comment here.  I don't think it would be a good idea to invent "magic numbers" like 12 here.

Can we also use the enumeration system here to provide the error messages?  Maybe something like `DECL_ENUMERATION("ldc1612_home_error:", "Eddy I2C IO error", HOME_ERROR_BUS_IO)` and the host could do something like: `error_code = res - REASON_SENSOR_ERROR; raise command_error(home_enums.get(error_code))`

src/sensor_ldc1612.c
```diff
@@ -143,19 +157,24 @@ check_home(struct ldc1612 *ld, uint32_t data)
...
 // Read the status register on the ldc1612
 static uint16_t
 read_reg_status(struct ldc1612 *ld)
 {
     uint8_t data_status[2];
-    read_reg(ld, REG_STATUS, data_status);
+    int ret = read_reg(ld, REG_STATUS, data_status);
+    // Bits 5:4 are unused, report I2C
+    if (ret != I2C_BUS_SUCCESS)
+        data_status[0] |= STATUS_I2C_ERROR;
```
KevinOConnor: I don't think it is a good idea to mix local data with remote data like this.  The issue is, it is possible that the chip is misbehaving (or there are severe noise issues on the i2c line) which cause these bits to be set in the data read from the device.  That could cause very confusing debug messages.  (It also makes gcc's job harder as it wont know that these bits are only local).  I think it would be preferable to return two separate fields from `read_reg_status()` - for example (`static int read_reg_status(struct ldc1612 *ld, uint16_t *status)`).

src/sensor_ldc1612.c
```diff
@@ -44,6 +45,39 @@ check_intb_asserted(struct ldc1612 *ld)
     return !gpio_in_read(ld->intb_pin);
 }
 
+// Default TRSYNC_TIMEOUT, limits ODR > 40
+#define WATCHDOG_TIMEOUT timer_from_us(25000)
+#define WATCHDOG_TIMEOUT_ERROR (1<<4)
```
KevinOConnor: I agree that we want to implement a "sensor timeout" feature for ldc1612.

I think we also need to look hard to see if now is a good time to refactor the code though.  That is, I think we may want to consider if ldc1612 can share the infrastructure in load_cell_probe.c , or alternatively perhaps the "sensor timeout" could be moved into trsync.c ..

I also think that if there is a long running timer we probably want its duration to be set by the host and it probably should not be hardcoded into the mcu.

So, I think it might be worthwhile to move the "sensor watchdog" change to a separate PR as I think it is a larger topic.

KevinOConnor:
Thanks again.  That completes my comments.

Cheers,
-Kevin

src/sensor_ldc1612.c
nefelim4ag: Sure.
If you don't mind.

I would suggest to have at least some host side temporary duct tape, where `EddyGatherSamples` would wait for samples to be recieved at least once (even if, it will be done not in this PR).
It is surprisingly easy to make sensor hung (at least for me with static_pwm_clock or when messing up with registers).

src/sensor_ldc1612.c
nefelim4ag: Ah, I forgot that bits always can be noisy.
Fixed.

klippy/extras/ldc1612.py
nefelim4ag: This is was mostly done to have backward compatibility with existing interface.
The only complication with this approach, that bits can be set sumultaniously.
For example "conversion watchdog + amplitude error".

So, I'm a little bit sceptical about reencoding all error bits with enumeratons.
Probably the middle ground is to still send the bits as before, but in case of I2C errors do 0xffff encoding like above?

~The possible workaround is to assume some priority for errors within MCU~
So:
~I2C -> Conversion watchdog -> Over/under range ->~ ~amplitude~ (Ah, I want to ignore them and they is not fatal for measurments).

---
Ah, and we can't trigger under range without enabling of OFFSET, but probably it is still better to define it.

---
I guess middle ground for I2C errors should work.
Because, as far as I understand, we still want to report amplitude errors.

klippy/extras/probe_eddy_current.py
nefelim4ag: I like the enumeration for trsync error decoding.
I only don't like a little to blend the sensor specific errors with generic eddy handler.
Probably, I can hide sensor specific decoding within some call to the ldc1612 class.

klippy/extras/ldc1612.py
KevinOConnor: > Probably the middle ground is to still send the bits as before, but in case of I2C errors do 0xffff encoding like above?

Right.  I don't think we want to reencode the error bits already present in the `REG_DATA0` register - as there's no confusion on the meaning of those bits.  I think it would be preferable to have a more regularized encoding for errors not already in `REG_DATA0` - for example: invalid data (all high bits set), i2c error, etc.

Using enumerations is just a "random idea" - I don't have strong feelings about it one way or another.

nefelim4ag:
Thanks, I hope I've addressed everything.

-Timofey

klippy/extras/ldc1612.py
nefelim4ag: I think enumeration should be fine for now.

Probably, in the future, when there would be a refactor of TRSYNC, there would be general TRSYNC errors with watchdog, IO errors, whatever.

As bonus, it allows for now to return "specific" errors for homing.

KevinOConnor:
Thanks for working on this.

I reviewed your latest changes.  I'm still concerned about the mixing of the local error codes with remote data from the sensor.  In particular, I'm concerned that if the sensor sent something like 0xffffffff that the user would get very confusing error messages (sensor_ldc1612.c could treat that as a kind of i2c error).  I'm also concerned that if there is an i2c error that sensor_ldc1612.c encodes as 0xffffxxxx that ldc1612.py would pass that along as a valid measurement with a very high frequency.  I think it would be preferable if the code did more to keep the two error paths distinct.

As an alternative, how about the changes at https://github.com/KevinOConnor/klipper-dev/tree/work-ldc1612error-20260107 .  This is my attempt to keep the error paths more separate.

Thanks again,
-Kevin

nefelim4ag:
Looks good to me overall.

Ah, yeah, I just was not really successful with filtering I2C error data from the stream, because the only solution that I obviously see is to copy data into the new array, and so, I decided it is "rare situation, whatever".
Yes, I think this sort of host dropping the I2C error data is the right approach.

I agree with the general intention to separate code paths, because otherwise it is quite a mess here.
I just wanted to avoid code duplication. Also, I like the separete handling of I2C status/data messages, and detection of 0xffff.

I will try to incorporate your work, thanks!

-Timofey.

nefelim4ag:
I incorporated your ideas with minor changes.
Sensor error is 0, so the same error path as before - Generic sensor homing error.
Added goto to decrease overall code duplication.

I've retested: over range (sensor error), conversion watchdog, I2C STATUS error.

Docs describing errors and bits decoding can be questionable, we can either change it/split/drop.

-Timofey.

---
I think goals for this PR in decreasing priorities:
- Fix amplitude warning thing
- Add some safe guards
- Have log output upon errors, instead of silent "Sensor reports error"
- Handle I2C errors because of specific sensor
- +Document something

So, documentation can be scapped / added later I guess.

---
Minor formatting changes.

KevinOConnor:
Thanks.

The first two commits (handle i2c errors, ignore amplitude errors during homing) look good to me.  I'll commit when you are ready.

The "trigger error on high frequency" commit seems a little odd to me.  I think we'd be better off telling the mcu code what the valid range is when starting a homing attempt.  Or, use the infrastructure already in load_cell_probe.c which already supports doing that (along with watchdog support).  I'll try to mock up a refactor of load_cell_probe.c to see if that approach makes sense.

The "decode error flags" change is fine if you wish to do that.  FWIW, it seems to me that the only meaningful flag is the amplitude warning flag - all the other flags basically mean "your sensor is broken and you need to replace it".  So, writing a bunch of code with slightly different variants of "replace your sensor" seems a little odd to me.  Also, the checks for `mv == 0x0000000` and `mv > 0x3ffffff` are a little confusing because we wont take that code branch when none of the error flags are set.  As above, I'm fine with committing this if you wish.

The "descrie eddy error messages" change is fine with me.

Cheers,
-Kevin

nefelim4ag:
> I think we'd be better off telling the mcu code what the valid range is when starting a homing attempt

This is my attempt to ensure generous sanity check in case if coil is [disconnected](https://github.com/Klipper3d/klipper/pull/7125#discussion_r2629422992)[^1]
Or [frequency is misconfigured](https://discord.com/channels/431557959978450984/1449832693960413285/1451676188652343367)[^2]

Because we can't really use calibration data, calibration data can be quite off if reference frequency is misconfigured and there is no way for a enduser to somehow detect that case without prior knowledge.
And this code will detect that situation with this MCU check.
Also future tap will be able to do validate frequency configuration, because samples is validated against ODR.

_We can use internal sensor `OFFSET` to set minimal adequate sensor output if we like to (without general solution for everything). Example: https://github.com/Klipper3d/klipper/commit/d6f4e9745e20f930f85b104a7e52c249ca5e5c95_

> The "decode error flags" change is fine if you wish to do that.

Over range will make sense right after the merge, I think.
Under range only if we decide to enable OFFSET at all (AFAIK, there is no other way to trigger it).

> Also, the checks for mv == 0x0000000 and mv > 0x3ffffff are a little confusing because we wont take that code branch when none of the error flags are set. As above, I'm fine with committing this if you wish.

Ooops. Hopefully now fixed.

I'm able to trigger soft over range with 8MHz and 4MHz reference clock, and hard with 2MHz reference clock.
Where ~3.2MHz is frequency of my coil.

Cheers,
-Timofey.

---
_I think we both biased a bit about amplitude warning, I know it happens for me and [SHT36v3](https://discord.com/channels/431557959978450984/801826273227177984/1457925662017847317), and somehow does not happen for you, even when it is installed this close to the bed: https://github.com/Klipper3d/klipper/pull/7160#issuecomment-3706636596_

[^1]: It is easy to achive with SHT36v3 where coil is on the FFC
[^2]: I have no other data about Sovol's eddy for now. I guess the reference frequency is <12MHz.

nefelim4ag:
I will sligtly reformulate.
max frequency limit at the MCU is useless for the currect probing/homing code, because we already provide the trigger frequency which is normally lower than 1/4.
But there should be safe guard for the tap probe, which only cares about diffs of the value.
At least until we decide to replace it with something different.

If that makes sense, I can remove it from here and move to the tap PR.

Thanks!
-Timofey.

KevinOConnor:
> This is my attempt to ensure generous sanity check in case if coil is disconnected
Or frequency is misconfigured

> But there should be safe guard for the tap probe, which only cares about diffs of the value.

I understand and I agree.  It's a hard coded range limit in the mcu code though.  I think a value passed from the host would be preferable, even if that limit is hard coded in the host.

In any case, it is not a big deal either way.  If you are ready, I'll commit this PR.

Cheers,
-Kevin

nefelim4ag:
Appreciated,
Let's commit it then.

-Timofey

_Now all LDC based probes should be fully practically functional._

KevinOConnor:
Thanks.

-Kevin

PR #6815, opened 2025-02-09
[kinematics] Generic Cartesian kinematics implementation

dmbutyugin:
This PR adds support for a new generic Cartesian kinematics class, which allows one to define various Cartesian-style kinematics (e.g. CoreXYUV, inverted hybrid CoreXY, and others). This was extensively discussed on [Discourse](https://klipper.discourse.group/t/new-generic-cartesian-kinematics-incl-corexyu-corexyuv-generic-idex-and-awd) and tested for a while now. I think the only non-addressed feedback from there is choosing an appropriate name for a parameter that defines the stepper kinematics, as there were concerns about possible confusion for an option
```
[stepper a]
kinematics: x-y
...
```
however I found a proposed alternative
```
[stepper a]
carriage_control: x-y
```
to sound a bit odd. But I do not have strong arguments against this option either, and I could accept it as well, if that is desired (alternatively perhaps just `carriages: x-y` may also be OK).

That said, any additional feedback is appreciated.

@HelgeKeck FYI


KevinOConnor:
Thanks.  This looks like useful functionality and I plan to provide comments on it in the coming days.  Sorry for the delay.

-Kevin

dmbutyugin:
Thanks Kevin! No worries, it is not super urgent.

docs/G-Codes.md
```diff
@@ -347,15 +347,19 @@ The following command is available when the
...
 carriage execute subsequent G-Code commands as-is. `COPY` and `MIRROR`
 modes are supported only for `CARRIAGE=1`. When set to either of these
 modes, carriage 1 will then track the subsequent moves of the carriage 0
 and either copy relative movements of it (in `COPY` mode) or execute them
-in the opposite (mirror) direction (in `MIRROR` mode).
+in the opposite (mirror) direction (in `MIRROR` mode). `AXIS` parameter is
+supported only if a user defined two dual carriages with `generic_cartesian`
+kinematic. In this case `AXIS` parameter can be omitted if `MODE` is `PRIMARY`,
+effectively changing the mode for both axes, but it must be specified for
+`COPY` and `MIRROR` modes.
```
KevinOConnor, line 362: Minor, but if we now have explicit carriage names, it might be nice to have SET_DUAL_CARRIAGE take an explicit carriage config name (and thus possibly avoid the numbers and axes parameters).

docs/G-Codes.md
```diff
@@ -688,6 +692,50 @@ is specified then the toolhead move will be performed with the given
...
+[DISABLE_CHECKS=[0|1]]`: Set or update the stepper kinematics.
+`<stepper_name>` must reference an existing stepper defined in
+`printer.cfg`, and `<kinematics>` describes the carriages the stepper
+moves. All defined carriages can be specified here, as well as their
+linear combinations, for example `x`, `x+y`, `y-0.5z`, `x-z`, etc. Note
+that unlike the printer configuration, `KINEMATICS=...` string must not
+use `*` symbol for multiplication, since it is a special GCode character,
+and if required, the multipliers must precede the carriage reference
+without any further delimiters, e.g. `0.5x+0.5z` or
+`0.333333333333x-0.666666666667y`.
```
KevinOConnor, line 710: The '*' character should no longer be special (commit d45b9c92).  I think it would be preferable to require the same syntax in the g-code that is in the config.

docs/Status_Reference.md
```diff
@@ -541,16 +541,36 @@ The following information is available in the `toolhead` object
   the printer had to be paused because the toolhead moved faster than
   moves could be read from the G-Code input.
 
+Additionally, if the printer uses `generic_cartesian` kinematic, the
+follwoing information is available:
```
KevinOConnor, line 545: follwoing typo.

docs/Status_Reference.md
```diff
@@ -541,16 +541,36 @@ The following information is available in the `toolhead` object
...
+- `stepper_kinematics`: a map from a stepper name to the array of 3
+  floats - coefficients of the stepper kinematics for each of the axis
+  X, Y, and Z from the corresponding `kinematics` option, as indicated
+  by the specified carriages and their axes. Can be used to compute
+  stepper positions from the known toolhead position.
+- `toolhead_kinematics`: a map from a stepper name to the array of 3
+  floats - a contribution of a given stepper into the motion of each of
+  the axes X, Y and Z of the printer - which is effectively a
+  pseudoinverse of the `stepper_kinematics`. Can be used to compute
+  the toolhead position from the known stepper positions.
```
KevinOConnor, line 555: This info seems very technical.  What's the use-case that users need this information for?

The printer.toolhead object has become a bit of a "dumping ground" for random bits of info and it's getting a little cluttered.  I'm not sure if we should add more there.  This is not a "blocker" though.

docs/Status_Reference.md
```diff
@@ -541,16 +541,36 @@ The following information is available in the `toolhead` object
...
   "INACTIVE" and "PRIMARY".
 - `carriage_1`: The mode of the carriage 1. Possible values are:
   "INACTIVE", "PRIMARY", "COPY", and "MIRROR".
 
+If a user defined two dual carriages with `generic_cartesian` kinematic,
+the statuses exported are:
+- `carriage_0`: The array of length 2 - modes of the carriage 0 for X and Y
+  axes. Possible values are: "INACTIVE" and "PRIMARY".
+- `carriage_1`: The array of length 2 - modes of the carriage 1 for X and Y
+  axes. Possible values are: "INACTIVE", "PRIMARY", "COPY", and "MIRROR".
```
KevinOConnor, line 572: What does "the array of length 2" mean?  Is there a reason this changes from the previous status format?

config/example-generic-caretesian.cfg
```diff
@@ -0,0 +1,138 @@
...
+
+[stepper x]
+kinematics: x+y
+step_pin: PF0
+dir_pin: PF1
+enable_pin: !PD7
+microsteps: 16
+rotation_distance: 40
+
+[stepper u]
```
KevinOConnor, line 46: Not a "blocker", but if a config name can be any arbitrary name then I think it would be nice for the example configs to make that more clear - for example:
```
[extra_carriage my_extra_y_carriage]
...
[dual_carriage my_idex_carriage]
...
[stepper my_stepper_for_x]
...
[stepper my_idex_stepper]
...
[stepper my_secondary_idex_stepper]
```

In some cases carriage names must be a pre-defined value (eg, "[carriage z]"), but in other cases the user can specify a name of their choice, and making that a little more clear in the examples may help.

klippy/stepper.py
```diff
@@ -294,23 +294,19 @@ def parse_step_distance(config, units_in_radians=None, note_valid=False):
...
+# A motor control carriage with one (or more) steppers and one (or more)
 # endstops.
-class PrinterRail:
+class GenericPrinterCarriage:
     def __init__(self, config, need_position_minmax=True,
-                 default_position_endstop=None, units_in_radians=False):
-        # Primary stepper and endstop
-        self.stepper_units_in_radians = units_in_radians
+                 default_position_endstop=None):
+        self.config = config
```
KevinOConnor, line 302: The convention is to never store a persistent reference to the "config" object.  (Because, once the config phase ends, calls to the config methods can cause surprising results.)  Is it possible to rework this so that we don't store a long-held reference?

KevinOConnor (commented):
Thanks!  I have some comments.  Mostly comments/questions and not "blocking" issues.

* Is the intent for me to "squash and merge" this series of commits?  That's fine if so.  However, if you were looking to preserve the commit history then I think this PR needs a rebase (for example, some test cases were added in a commit before the commit that adds the code to support them - which would break `git bisect`).
* I'd go with "carriage_control" over "kinematics" as I find the term "kinematics" to be strange when we're specifying the carriages that move.  This is not a "blocker" if you'd really prefer "kinematics".  Other possibilities might be "carriage" or "carriages".
* What are the endstop `del_stepper()` calls for?  It seems odd to be removing references to the steppers at runtime because one can not, in general, add new steppers to the endstop code at runtime.  (The endstop code needs a trsync oid allocated on the mcu, which can only be done at mcu init time.)  This looks like it is tied to the SET_STEPPER_KINEMATICS code - if that is not core to the main support here is it reasonable to push that command and code to a followup PR?
* This PR adds a new extras/kinematic_stepper.py file, but I think it may be more clear to add that file as kinematics/kinematic_stepper.py as it's not really an "extras" module (and isn't used by any "extras").
* I think if we add this, we should look to deprecate hybrid_corexy, hybrid_corexz, corexz, and deprecate dual_carriage on cartesian.  Perhaps deprecate a month or so after merging this PR.  Do you have any thoughts on that?

Also, see my comments below.
-Kevin

KevinOConnor:
> I think it makes sense to merge the PR as a single commit, no need to maintain the history. However, the PR may benefit if it is split into 2-3 smaller PRs.

Sounds good.

> For instance it seems that some functionality requires additional work, such as SET_STEPPER_KINEMATICS, and I will move it to a separate PR in the future after this one is merged in. That one may be used for some very complicated homing procedures, for examples for Tripteron discussed [here](https://klipper.discourse.group/t/linear-axis-tripteron-kinematics/17727). But I get it that it may require some additional care.

Sounds good.

> Similarly, the new dual_carriage code supports two DC carriages on two axis simultaneously now, hence some of the changes in the commands and returned status. I'll see if it would be feasible to move that also into a separate PR, but it might just complicate things, so I'm not sure if I will do that.

Sounds good.

> I'd go with carriages parameter, because it also goes well with a corresponding command name SET_STEPPER_CARRIAGES.

Sounds good.

> I don't have a strong opinion for deprecating hybrid_corexy, hybrid_corexz, corexz, and dual_carriage. That said, deprecating is one thing, and hopefully will allow us to receive more feedback about the new kinematics, but deleting the functionality will likely require a prolonged period of time.

Agreed.  I'm roughly thinking post a deprecation notice a month or so after merging generic_cartesian, and then look to remove the old support about a year after that.

> stepper_kinematics and toolhead_kinematics are exported in the toolhead status as this is directly related to the toolhead kinematics. I think this may be useful in some macros to perform stepper-to-toolhead position calculations and vice versa, to calculate a stepper position from the toolhead position, e.g. for some low-level stepper position corrections and such, or again, in homing procedures. While for your specific printer config you could determine these coefficients and put them into the macros, I think it may help with writing more generic macros that would work for different users. But if this is perceived as unnecessary clutter, I can also remove them.

It's not a "blocker" either way, but it's pretty simple to add in the information later if someone reports they need the information.  Waiting for an active "use case" can also help with long-term maintenance.

> Since dual_carriage now supports two carriages for X and Y, the modes for X and Y could differ, e.g. in order to print a mirrored part in this setup a user would need to have MIRROR mode for X and COPY mode for Y (or vice versa)...

How about `printer.dual_carriage.carriages = {'x': 'PRIMARY', 'y': PRIMARY, 'my_extra_y': 'MIRROR', 'my_extra_x': 'COPY'}`.  Then if you want, you could emulate `printer.dual_carriage.carriage_0`, and we could list that older view as deprecated.  In general, it's preferred for the first level of `get_status()` to have static values (ie, the "xxx" in `printer.module.xxx`) so if we want to export a dynamic dictionary, it's generally better to put that dictionary behind another field (eg, `printer.dual_carriage.carriages`).

Thanks again,
-Kevin

docs/G-Codes.md
dmbutyugin: Now `SET_DUAL_CARRIAGE` has been updated to receive a carriage name, e.g. SET_DUAL_CARRIAGE CARRIAGE=v MODE=COPY

docs/G-Codes.md
dmbutyugin: I updated and simplified the code to always expect '*' as a multiplier

docs/Status_Reference.md
dmbutyugin: OK, I've removed this for now (and naturally from the documentation)

docs/Status_Reference.md
dmbutyugin: The exported statuses have been updated per conversation here.

config/example-generic-caretesian.cfg
dmbutyugin: I updated the stepper names to highlight the possibility. For the carriage names, I'm a bit unsure, in a sense that in many cases these documentation snippets are copied by people to adjust them to their needs, but keep the main structure as-is. So I do not want to promote too long carriage names, since they are used later in the stepper configuration, and it looks a bit weird when long names appear there.

klippy/stepper.py
dmbutyugin: I've indeed refactored the code to not keep the reference to config object beyond initialization.

dmbutyugin:
@KevinOConnor, so I've resolved all high-level feedback and comments. I think the only outstanding question is whether we want to proceed with the change (initially proposed by Kevin in the discourse thread) to IDEX homing - home only the active carriage - or not. In fact, I've already implemented that, but did not push the code to this PR just yet. I think we need to agree on it first (also, there's a prerequisite to that change in my opinion, a PR #6860). But I do not think that such a change should be pushed to a follow-up PR, since it will break existing installations in a fairly surprising and potentially unsafe way, so it is better to commit a desired behavior to the Klipper mainline from the very beginning.

KevinOConnor:
Okay, thanks.  I agree we should figure out what we ultimately want to do with homing.  I can write up some of my thoughts on homing, and I can review this PR and #6860.  However, I'm still likely a week or so out from being able to get to that.

-Kevin

dmbutyugin:
Thank you, Kevin, it is not urgent. FWIW, if you want to take a look at my proposal regarding homing that I've implemented, you can take a look at 79b386f8eb63b0b4fb891f929d7f8b668dd1ba8d commit (not part of this branch at the moment).

dmbutyugin:
Given the general direction of the homing revamp (being discussed on [Discourse](https://klipper.discourse.group/t/thoughts-on-homing-support-in-klipper/22737)), I'd suggest to merge this PR as-is without any additional homing rework. This is because the currently implemented homing in this PR is completely in line with the homing in other Cartesian-style kinematics classes, and the new homing proposal anyways suggests full revamp of the homing implementation and is equally disruptive for all kinematics. So this PR would not introduce any unique issues or challenges that the users of the `generic_cartesian` kinematic class would have to overcome as the homing changes will be implemented.

KevinOConnor:
Thanks for working on this.  In general I agree that it would be better to merge this before doing a homing rework - but I have a few questions about that.  In any case, this would be something to merge after tagging v0.13.0 .

Some questions:
1. Will the idex support on general_cartesian continue to home both idex carriages on a typical `G28 X0` command?
2. What happens on a `G28` command if the user has defined two idex setups?  (Both an X and a Y.)
3. When looking at the homing code, I performed an investigation into refactoring the existing carteisan code: https://github.com/KevinOConnor/klipper-dev/tree/work-cartesian-20250328 .  This code shouldn't be taken seriously as I'm not sure it's a good direction in general.  However, I did notice that the new generic_cartesian.py code has a different `get_status()` implementation (compared to https://github.com/KevinOConnor/klipper-dev/tree/work-cartesian-20250328 ).  Can you give some background for the changes there?
4. There's a bunch of stepper/rail/carriage/idex abstractions added for generic_cartesian - briefly: `MainCarriage`, `ExtraCarriage`, `generic_cartesian.DualCarriage`, `DualCarriages`, `idex_modes.DualCarriage`, `MCU_stepper`, `KinematicStepper`, `GenericPrinterCarriage`, `PrinterRail` .  I must admit, I've never fully understood the high-level categorization between the current idex based PrinterRail, DualCarriages, and DualCarriageRail classes.  Would you be able to give a few sentence description for each of the above nine classes to highlight what they do, their relationship to each other, and how they are distinct from the other classes?  I guess I just want to make sure these abstractions wont also need to be rethought if we tackle homing.

Thanks again.
-Kevin

dmbutyugin:
Hi Kevin,

>     1. Will the idex support on general_cartesian continue to home both idex carriages on a typical `G28 X0` command?

Yes, the current implementation homes both IDEX carriages in sequence, as per heuristic, but typically the primary carriage first, the dual carriage second.

>     2. What happens on a `G28` command if the user has defined two idex setups?  (Both an X and a Y.)

The carriages home in sequence as per defined order of homing the axes. Lets say we have `x` <-> `u` and `y` <-> `v`. Then by default `x` and `u` would be homed, then `y` and `v` (again, following the heuristic depending on the direction of homing, but typically `x` and `y` would be homed before `u` and `v` respectively). If a user wants to override the axes homing order, they can do so by defining either `[homing_override]` or `[gcode_macro G28]`.

>     3. When looking at the homing code, I performed an investigation into refactoring the existing carteisan code: https://github.com/KevinOConnor/klipper-dev/tree/work-cartesian-20250328 .  This code shouldn't be taken seriously as I'm not sure it's a good direction in general.  However, I did notice that the new generic_cartesian.py code has a different `get_status()` implementation (compared to https://github.com/KevinOConnor/klipper-dev/tree/work-cartesian-20250328 ).  Can you give some background for the changes there?

You are talking about the `axis_minimum` and `axis_maximum` calculation, right? The problem is that current implementation is not IDEX-aware. TBH, the [documentation](https://www.klipper3d.org/Status_Reference.html#toolhead) is not very descriptive in what exactly to expect there. However, when a dual carriage is activated, the range of a corresponding axis changes, but this is not reflected in the status, which makes it a bit useless in practice for writing macros that should deal with the dual carriage - something that I experienced myself. And so I wanted to address this, which was a bit of an oversight in my opinion. I suppose we could have alternative implementations, e.g. report `limits` there instead in order to account for the `safe_distance` and for `COPY` and `MIRROR` modes, or report additional carriages and their limits as, e.g. `toolhead.axis_minimum.u` and `toolhead.axis_maximum.u` and such. TBH, I do not remember why I chose this particular implementation over reporting `limits`, that might have been better (I'd only have to handle non-homed state separately). So, let me know what your thoughts are.

And I did not change other kinematics because that is a breaking change, and if we are going to deprecate `dual_carriage` for other kinematics, it's just not worth the trouble.

>     4. There's a bunch of stepper/rail/carriage/idex abstractions added for generic_cartesian - briefly: `MainCarriage`, `ExtraCarriage`, `generic_cartesian.DualCarriage`, `DualCarriages`, `idex_modes.DualCarriage`, `MCU_stepper`, `KinematicStepper`, `GenericPrinterCarriage`, `PrinterRail` .  I must admit, I've never fully understood the high-level categorization between the current idex based PrinterRail, DualCarriages, and DualCarriageRail classes.  Would you be able to give a few sentence description for each of the above nine classes to highlight what they do, their relationship to each other, and how they are distinct from the other classes?  I guess I just want to make sure these abstractions wont also need to be rethought if we tackle homing.

For `generic_cartesian` I added `MainCarriage`, `ExtraCarriage`, `DualCarriage` and `KinematicStepper`, which basically provide nice API for the main `GenericCartesianKinematics` class (vs. e.g. `PrinterRail` which by itself is a bit misnomer and extra functionality would be out of place), and also deal with parsing the configuration. And in order to reduce the code duplication, I split `PrinterRail` into two classes, `GenericPrinterCarriage`, which are used by `MainCarriage` and `DualCarriage`, and `PrinterRail` itself that is still used by other kinematics (and `PrinterRail` is kind of `MainCarriage` + `ExtraCarriage` to organize multi-rails for these other kinematics). With the discussion so far, I do not think these would require rethinking for homing, but also since they are internal non-public classes, I do not see a problem refactoring them in the future. At the moment, they are pretty light-weight.

As for their connection with other classes. I see `MCU_stepper` as essentially a Python wrapper around C code that does not deal with the configuration parsing (`PrinterStepper` does that and `KinematicStepper` adds processing of additional configuration options specific to `generic_cartesian`). For IDEX, `idex_modes.DualCarriages` provides high-level facilities to organize IDEX motion and public API (carriage activation, limits calculation, order of homing, etc.) and `idex_modes.DualCarriage[Rail]` is (was) a wrapper around `PrinterRail`. Both of them are needed for now to support IDEX on other kinematics. However, if we deprecate IDEX support on other kinematics, when we delete it for good, `idex_modes.DualCarriage[Rail]` could likely be eliminated completely and corresponding functionality just moved into `MainCarriage` and `DualCarriage` of generic_cartesian, and `idex_modes.DualCarriages` could be moved into `generic_cartesian` and simplified (some functionality there is used only by other kinematics now, and some functionality must account for both `generic_cartesian` and other kinematics, which complicates the implementation).

I hope this helps.

KevinOConnor:
Thanks for your detailed response.

> The carriages home in sequence as per defined order of homing the axes.

Okay.

> TBH, the [documentation](https://www.klipper3d.org/Status_Reference.html#toolhead) is not very descriptive in what exactly to expect there

My general thinking is that it provides the "bounding box" on all possible moves.  That is, a move outside of the bounding box is certainly invalid, but a move within the bounding box is not necessarily valid.  It was originally added so that menus could show a reasonable minimum and maximum bounds for users to select from.

It seems okay to change on an idex carriage change, but not a requirement for it to change.

> For generic_cartesian I added ...

Okay, thanks.  I think it would be good to "flatten" the class hierarchy a little.  This is something I could do after merging, so not a blocker.

I can certainly see the advantage of having a generic_cartesian internal class for KinematicStepper.  By "flatten", I mean that I think it would be preferable for KinematicStepper to "have a" MCU_stepper and not try to "be a" MCU_stepper.  That is, I think the `toolhead.get_kinematics().get_stepper()` call should return actual `MCU_stepper` objects.  Said another way, I think it would be preferable for KinematicStepper to be an internal abstraction for generic_cartesian and it would be preferable if that abstraction wasn't passed to other Klipper modules.  At first glance, this seems like it could be as simple as adding a `get_underlying_stepper()` call to KinematicStepper and calling that from `GenericCartesianKinematics.get_steppers()`.

Similarly, I can certainly see the value in MainCarriage, ExtraCarriage, DualCarriage but I'd be leery of passing them to the `homing.py:home_rails()` code.  I think the homing code should only be passed actual `PrinterRail` classes.  Also, FWIW, I'd be leery of defining a base GenericPrinterCarriage class in stepper.py and inheriting from it in classes defined in generic_cartesian.py - the issue is that developers looking at one of those files may inadvertently change a member variable not realizing the dependency on it from a different module - Python really isn't good at handling class inheritance.  So, as above, if we could have some kind of `get_underlying_rail()` call so that the external code isn't exposed to the generic_cartesian abstractions then I think that would be preferable.

One of the reasons I'm leery of adding new classes with a "is a" class relationship is that I think that's lead to trouble with the current homing code.  To wit, we've got a "probe" which is a "bltouch" which is an "endstop" and when we realize the abstraction needs to be extended (eg, for load cells) we end up having to rework the all probes, bltouch, endstops, tmc, the homing code, etc.

As above, I don't think this is a "blocker" for merging, as I think it could be tackled during a homing rework.  I think it would be good if we are "on the same page" though.

Thoughts?
-Kevin

dmbutyugin:
Thanks for the feedback, Kevin.

> My general thinking is that it provides the "bounding box" on all possible moves. That is, a move outside of the bounding box is certainly invalid, but a move within the bounding box is not necessarily valid. It was originally added so that menus could show a reasonable minimum and maximum bounds for users to select from.

Well, so for dual_carriage on regular kinematics this is not really a bounding box anymore. For instance, one could define `[stepper_x]` with min and max positions 0, 250, and `[dual_carriage]` with min and max positions as 50, 300. Then, the returned `axis_minimum.x`, `axis_maximum.x` (0, 250) will not be valid for the dual_carriage as a bounding box (and will not be great for menu either, as it will make certain areas of the build plate inacessible via display moves when dual carriage is activated). That said, more accurate bounds in copy and mirror modes may also be beneficial.

> I can certainly see the advantage of having a generic_cartesian internal class for KinematicStepper. By "flatten", I mean that I think it would be preferable for KinematicStepper to "have a" MCU_stepper and not try to "be a" MCU_stepper. That is, I think the `toolhead.get_kinematics().get_stepper()` call should return actual `MCU_stepper` objects. Said another way, I think it would be preferable for KinematicStepper to be an internal abstraction for generic_cartesian and it would be preferable if that abstraction wasn't passed to other Klipper modules. At first glance, this seems like it could be as simple as adding a `get_underlying_stepper()` call to KinematicStepper and calling that from `GenericCartesianKinematics.get_steppers()`.
> 
> Similarly, I can certainly see the value in MainCarriage, ExtraCarriage, DualCarriage but I'd be leery of passing them to the `homing.py:home_rails()` code. I think the homing code should only be passed actual `PrinterRail` classes. Also, FWIW, I'd be leery of defining a base GenericPrinterCarriage class in stepper.py and inheriting from it in classes defined in generic_cartesian.py - the issue is that developers looking at one of those files may inadvertently change a member variable not realizing the dependency on it from a different module - Python really isn't good at handling class inheritance. So, as above, if we could have some kind of `get_underlying_rail()` call so that the external code isn't exposed to the generic_cartesian abstractions then I think that would be preferable.
> 
> One of the reasons I'm leery of adding new classes with a "is a" class relationship is that I think that's lead to trouble with the current homing code. To wit, we've got a "probe" which is a "bltouch" which is an "endstop" and when we realize the abstraction needs to be extended (eg, for load cells) we end up having to rework all probes, bltouch, endstops, tmc, the homing code, etc.
> 
> As above, I don't think this is a "blocker" for merging, as I think it could be tackled during a homing rework. I think it would be good if we are "on the same page" though.

OK, now I understand your concerns. I guess it can be done more or less as you suggest. So let me do a pass over the PR and see if I run into some blockers or not. Then I'll report my findings here, and if successful, add a commit to this PR with such a refactoring.

dmbutyugin:
@KevinOConnor I believe I addressed the concerns related to non-flat class hierarchy and inheritance. So, please take a look at the updated code when you have time.

KevinOConnor:
Thanks!  Looks good to me.  I'll give a few days to see if there are other comments and otherwise look to commit.

I'll look to do a "squash and merge" into a single commit when I merge.

-Kevin

dmbutyugin:
OK, sounds good from my side. Thanks Kevin.

KevinOConnor:
Thanks.

-Kevin

jkwilling:
This seems to break Happy Hare
./extras/mmu_machine.py:class MmuPrinterRail(stepper.PrinterRail, object):

If I edit the file to reference GenericPrinterRail, then I next get a error:
Option 'endstop_pin' in section 'stepper_mmu_gear' must be specified

thisisnotfez:
Breaks protoloft's auto_z as well.
```
Internal error on command:"CALIBRATE_Z"
  File "/home/biqu/klipper/klippy/gcode.py", line 212, in _process_commands
    handler(gcmd)
  File "/home/biqu/klipper/klippy/gcode.py", line 140, in <lambda>
  File "/home/biqu/klipper/klippy/extras/z_calibration.py", line 152, in cmd_CALIBRATE_Z
    state.calibrate_z(switch_offset, nozzle_site, switch_site, bed_site)
  File "/home/biqu/klipper/klippy/extras/z_calibration.py", line 474, in calibrate_z
    nozzle_zero = self._probe_on_site(self.z_endstop,
  File "/home/biqu/klipper/klippy/extras/z_calibration.py", line 431, in _probe_on_site
    curpos = self.helper._probe(self.gcmd, endstop,
  File "/home/biqu/klipper/klippy/extras/z_calibration.py", line 329, in _probe
    curpos = phoming.probing_move(mcu_endstop, pos, speed)
  File "/home/biqu/klipper/klippy/extras/homing.py", line 264, in probing_move
    hmove = HomingMove(self.printer, endstops)
  File "/home/biqu/klipper/klippy/extras/homing.py", line 48, in __init__
    self.endstops = [es for es in endstops if es[0].get_steppers()]
  File "/home/biqu/klipper/klippy/extras/homing.py", line 48, in <listcomp>
    self.endstops = [es for es in endstops if es[0].get_steppers()]
AttributeError: 'NoneType' object has no attribute 'get_steppers'
Transition to shutdown state: Internal error on command:"CALIBRATE_Z"
```

Sineos:
See https://klipper.discourse.group/t/the-dirty-flag-and-the-teams-position/22753 and in particular the chapter on "Update Implications".

aegyuptus:
A concern has come up: this works great for IDEX, but doesn't seem to work when more than one gantry or two toolheads are involved (i.e IQEX).

dmbutyugin:
@aegyuptus My impression was that it should be possible to configure IQEX, even if not optimally. I could be mistaken though. So I would suggest to create a post on Klipper Discourse, e.g. in Developers section
https://klipper.discourse.group/c/developers
where you could describe the challenges you face, and we can brainstorm if they can be resolved using the current Klipper, and if not, what kind of changes to `generic_cartesian` are due to make it work.

dmbutyugin:
@aegyuptus You can now also check out, here I added a full IQEX support (but it needs testing):
https://klipper.discourse.group/t/full-dual-gantry-support-iqex-and-itex-printers/25101

aegyuptus:
@dmbutyugin Thank you so much for pinging me! I would have missed this otherwise. This is fantastic and may indeed have solved the problem. Thank you for the work you do!

klippy/kinematics/kinematic_stepper.py
```diff
@@ -0,0 +1,92 @@
...
+        term = carriages_str[nxt:end].strip()
+        term_lst = term.split('*')
+        if len(term_lst) not in [1, 2]:
+            raise parse_error(
+                    "Invalid term '%s' in '%s'" % (term, carriages_str))
+        if len(term_lst) == 2:
+            try:
+                coeff = float(term_lst[0])
+            except ValueError:
+                raise error("Invalid float '%s'" % term_lst[0])
```
MRX8024, line 27: -

klippy/kinematics/kinematic_stepper.py
```diff
@@ -0,0 +1,92 @@
...
+            if term_lst[0].startswith('-') or term_lst[0].startswith('+'):
+                term_lst[0] = term_lst[0][1:]
+        c = term_lst[-1]
+        if c not in printer_carriages:
+            raise parse_error("Invalid '%s' carriage referenced in '%s'" %
+                              (c, carriages_str))
+        carriage = printer_carriages[c]
+        j = carriage.get_axis()
+        if coeffs[j]:
+            raise error("Carriage '%s' was referenced multiple times in '%s'" %
```
MRX8024, line 39: -

MRX8024 (commented):
@dmbutyugin I stumbled across this code and it seems to me that there are some issues with `raise error()`

klippy/kinematics/kinematic_stepper.py
nefelim4ag: Probably there is no error() in the context, and it is meant to be: `parse_error()`

PR #6821, opened 2025-02-13
Supports retrieving the name of the currently loaded skew correction …

GofranChang:
Added a command to retrieve the name of the current tilt correction configuration.

In some cases (such as enabling filament cutter or nozzle wipe), activating tilt correction may cause the tool head to exceed its range of motion. By retrieving the current tilt correction name, it is possible to get the tilt configuration name before execution, unload tilt correction, execute the command (cut filament or wipe nozzle), and then reload tilt correction using the obtained name.

Signed-off-by: Zhang Gaofan <zhanggaofan0827@gmail.com>

### DIFF
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -451,7 +451,7 @@ The following information is available in
 
 The following information is available in the `skew_correction` object (this
 object is available if any skew_correction is defined):
-- `cur_prof_name`: Returns the name of the currently loaded SKEW_PROFILE.
+- `current_profile_name`: Returns the name of the currently loaded SKEW_PROFILE.
 
 ## stepper_enable
 
--- a/docs/Status_Reference.md
+++ b/docs/Status_Reference.md
@@ -451,8 +451,7 @@ The following information is available in
 
 The following information is available in the `skew_correction` object (this
 object is available if any skew_correction is defined):
-- `printer.skew_correction.cur_prof_name`: Returns the name of the 
-  currently loaded SKEW_PROFILE.
+- `cur_prof_name`: Returns the name of the currently loaded SKEW_PROFILE.
 
 ## stepper_enable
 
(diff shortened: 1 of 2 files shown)


JamesH1978:
Thankyou for submitting a PR, please be aware you need to sign off, as per point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks
James

Arksine:
It would probably be better to report the current profile name in a `get_status()` method rather than print it via gcode command.  Then you can use templates in your custom macros to look up the name and do whatever you need with it.

GofranChang:
> Thankyou for submitting a PR, please be aware you need to sign off, as per point 3 in https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review
> 
> Thanks James

Thank you for the reminder. I have added my signature.

GofranChang:
> It would probably be better to report the current profile name in a `get_status()` method rather than print it via gcode command. Then you can use templates in your custom macros to look up the name and do whatever you need with it.

Thank you for your suggestion. I have updated the code following your method, and it is now working well.

KevinOConnor:
Thanks.  Any changes to exported status information should also have a corresponding change to docs/Status_Reference.md .  Also, for what it is worth, probably best to fully spell out the names of the status variables that are being exported (eg, something like "current_profile_name").

Cheers,
-Kevin

GofranChang:
> Thanks. Any changes to exported status information should also have a corresponding change to docs/Status_Reference.md . Also, for what it is worth, probably best to fully spell out the names of the status variables that are being exported (eg, something like "current_profile_name").
> 
> Cheers, -Kevin

OK, it's done (^○^)

KevinOConnor:
Thanks.  Seems fine to me.  I'll give a few days to see if there are any further comments, but otherwise look to commit.

-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6823, opened 2025-02-15
Fix minor transmit stall race when in USB to CAN bridge mode

KevinOConnor:
There was a small race condition that could allow transmit "echo" frames to become briefly stalled when in USB to CAN bridge mode.  This would likely only effect some controllers - potentially the stm32 usbotg (stm32f4 and stm32h7 devices).

The PR also contains a couple of commits intended to better organize the usb_canbus.c code.

-Kevin

### DIFF
--- a/src/generic/usb_canbus.c
+++ b/src/generic/usb_canbus.c
@@ -172,7 +172,7 @@ drain_canhw_queue(void)
 
 // Fill local queue with any USB messages read from host
 static void
-fill_usb_host_queue(void)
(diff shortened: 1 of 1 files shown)


KevinOConnor:
FYI, I also added an additional change to this PR - the gs_usb "echo frames" will now be sent back to the host prior to processing those frames.  This helps improve the output of candump debugging logs (it should not have an impact on klipper itself).

-Kevin

PR #6824, opened 2025-02-15
docs: Clarify sensorless homing moves mode selection in TMC_Drivers.md

paulfertser:
Hi,

Reading the docs I was really puzzled about the right way to configure stealthChop vs. spreadCycle knowing my tmc2209 drivers only support stallGuard in spreadCycle mode. Reading the code answered my question and I hope I can make it easier for the next confused person.

### DIFF
--- a/docs/TMC_Drivers.md
+++ b/docs/TMC_Drivers.md
@@ -83,6 +83,10 @@ setting `stealthchop_threshold` to 999999). Unfortunately, the drivers
 often produce poor and confusing results if the mode changes while the
 motor is at a non-zero velocity.
 
+When sensorless homing is enabled the homing moves will always
+automatically temporarily switch to the mode required by stallGuard,
+regardless of `stealthchop_threshold` setting.
+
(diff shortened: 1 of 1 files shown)


JamesH1978 (approved):
Thanks for the PR, yes you are correct, the mode does switch to spreadcycle automatically for sensorless/stallguard. Cant see an issue with adding this statement to clarify that.

Thanks
James

KevinOConnor:
Thanks.  Just thinking out loud, how about a slightly different wording - something like:

Note that the `stealthchop_threshold` config option does not impact sensorless homing, as Klipper automatically switches the TMC driver to an appropriate mode during sensorless homing operations.

-Kevin

paulfertser:
I would consider that equivalent, so if you prefer it for stylistic reasons and consistency with how the rest of the document is written why not :)

KevinOConnor:
Okay, thanks.  Commit a90110d9.

-Kevin

paulfertser:
Thank you for the attention and amazing software!

PR #6825, opened 2025-02-15
Update generic-bigtreetech-skr-v1.4.cfg with mcu working baud rate 11…

jorgegomzar:
Hello there.

This PR aims to fix 2 issues I encountered while flashing klipper onto a skr v1.4 turbo board today:

1. The default speed of 200000 won't work for this board. It is specified to use a [baud rate of 115200 by the manufacturer](https://github.com/bigtreetech/docs/blob/master/docs/SKR%20V1.4.md#:~:text=baud%20rate%20of-,115200,-Expand%20the%20interface)

If this option is missing from the printer.cfg file, the Raspi will hang when trying to connect to the board.

2. I wasted a few hours trying to flash the firmware via a USB until I found a documentation page explaining that this board can only use this method TO UPDATE the firmware. So I added a link pointing to this doc page so the next person coming here knows what to do.

Thank you!

### DIFF
--- a/config/generic-bigtreetech-skr-v1.4.cfg
+++ b/config/generic-bigtreetech-skr-v1.4.cfg
@@ -2,6 +2,9 @@
 # board. To use this config, the firmware should be compiled for the
 # LPC1768 or LPC1769(Turbo).
 
+# IMPORTANT: if this is the first time flashing Klipper to the mcu
+# See https://www.klipper3d.org/SDCard_Updates.html?h=btt#sdcard-updates
+
 # See docs/Config_Reference.md for a description of parameters.
 
 [stepper_x]
@@ -78,6 +81,7 @@ pin: P2.3
 
 [mcu]
 serial: /dev/serial/by-id/usb-Klipper_Klipper_firmware_12345-if00
+baud: 115200
 
 [printer]
 kinematics: cartesian


Sineos:
I'm the owner of a SKR 1.3, SKR 1.4, and SKR 1.4 Turbo:
1. All work with Klipper's default baud rate of 250k.
2. If I remember correctly (I Klipper'ed them a long time ago), they require renaming the `klipper.bin` to `firmware.bin` and copying it to the SD card, followed by resetting the board. A successful flash is indicated by the `firmware.bin` being renamed to `firmware.cur`.
3. The SD flash documentation clearly states that it is only applicable for updates, i.e., after Klipper has been initially installed.

So, in my opinion, the second point is the only area that could be improved.

jorgegomzar:
Hi! I only had a v1.4 turbo for a week, and am currently in the process of installing it for my ender6 printer.

But even after having the firmware installed in the board via SD card (I checked that the file was correctly renamed to firmware.cur), as soon as I tried to connect to the board without specifying the baud rate it completely hang my raspi. Had to power off both devices, start the raspi without the printer, start the printer and when restarting klipper the issue would re-appear.

I stumbled upon this reddit thread https://www.reddit.com/r/klippers/comments/skgf5f/comment/iotd0rd/, gave it a try and voila, was able to connect to the printer seemlessly

If you check that thread I am not the first one having this exact same issue :/

Sineos:
I do not know of any board where it is proven that a lower baud rate is actually needed. There are some reports pointing in this direction, but they are pretty rare. Countless other people have had good success with the default settings and the same hardware.

I believe the root cause is different, i.e., it could be a bad USB cable or some other issue, and the lowered baud rate is just masking the real problem (in the worst case, only partially, and one will experience other issues later on).

jorgegomzar:
I guess it could be possible what you describe.

Just for doc sake:
- the RaspberryPi is a 3A+ with 32bits raspbian, recently bought, OS installed today
- the USB cable is the standard USB cable bundled with the btt board. No weird pinching position
- the board is a BTT skr v1.4 turbo also recently bought (<1mo)
- tried with different USB cables, with an extra RaspiberryPi 3A+ bought on the same day as the other but running raspbian 64 bits

wdyt if the baud rate is commented out but with a comment saying "some users have reported issues with the default 250k baud rate, feel free to try with 115200 if it is your case"?

Sineos:
> wdyt if the baud rate is commented out but with a comment saying "some users have reported issues with the default 250k baud rate, feel free to try with 115200 if it is your case"?

As it is not clear where this issue is coming from, I'd be leery of masking it with a workaround that may create subtle problems later on.

Especially in the case of this board. Many boards use a "UART to USB" bridge chip, but the LPC boards are "true" USB boards. This means the USB interface is directly connected to the LPC chips. There should be no reason why it cannot handle a baud rate of 250k, unless other issues are present.

jorgegomzar:
The main reason I created this PR was because the manufacturer says the PC USB interface for this board operates at [115200 baud rate](https://github.com/bigtreetech/docs/blob/master/docs/SKR%20V1.4.md#:~:text=baud%20rate%20of-,115200,-Expand%20the%20interface) + the other success stories I found also using 115200 baud rate. But if you managed to make the board USB interface work at a higher baud rate there is no point on keeping this PR open.

As you already said the issue might be somewhere else. I even read some experiences about Raspi's modemmanager interfering with serial devices, so maybe that's it.

Thanks for your review, I'll keep investigating my setup.

jorgegomzar:
Coming back to say you were absolutely right. Both USB cables I tested did not solve the issue BUT, I bought a longer USB cable and it works just fine with the default baud rate.

Thanks for pointing me in the right direction.

PR #6828, opened 2025-02-18
bed_mesh: reduce generated point logging

Arksine:
This pull request addresses issues with `bed_mesh` generated points overwhelming the log and/or the pty.  Its common for users with "scanning" probes to configure dense meshes which may contain thousands of points.  The following changes have been made to mitigate this issue:

1) Initial points are truncated to 50.
2) Points are no longer printed when the mesh configuration changes.

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -133,7 +133,7 @@ class BedMesh:
         self.update_status()
     def handle_connect(self):
         self.toolhead = self.printer.lookup_object('toolhead')
-        self.bmc.print_generated_points(logging.info)
+        self.bmc.print_generated_points(logging.info, truncate=True)
     def set_mesh(self, mesh):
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  Let me know when you are ready to merge and I will do so.

Cheers,
-Kevin

Arksine:
I believe it is ready Kevin.  Thanks!

KevinOConnor:
Thanks.

-Kevin

vvuk:
What's the point of logging _any_ points? Is this output actually useful for anyone?

Arksine:
They can be used to verify that the generated points are in the expected locations.  Likely I would be the one primarily using it if a bug is found or a regression is introduced.  The truncated version is enough to get a glance, and if necessary there are better tools for analyzing the mesh data in the `graph_mesh.py` script.

PR #6829, opened 2025-02-20
RFC RP2040: set clock to 200Mhz

nefelim4ag:
RP2040 is now certified to run at 200Mhz clock instead of 133Mhz (125Mhz from SDK was used).
https://github.com/raspberrypi/pico-sdk/releases/tag/2.1.1

So, if no one minds, this is an overclock to the new allowed ones.

New perf values:
`arm-none-eabi-gcc (Arch Repository) 14.2.0`
```
rp2040	  ticks
1 stepper 3  ~ 4000K
3 stepper 14 ~ 2571K
```

Which allows us to place it aside with STM32F446 on the benchmark page.

This patch has been only tested on the development board.

### DIFF
--- a/src/rp2040/main.c
+++ b/src/rp2040/main.c
@@ -12,6 +12,7 @@
 #include "hardware/structs/resets.h" // sio_hw
 #include "hardware/structs/watchdog.h" // watchdog_hw
 #include "hardware/structs/xosc.h" // xosc_hw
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  I committed this change along with commits 14c105b8 and 1836ec43 on top.

-Kevin

PR #6830, opened 2025-02-21
stm32: Add support for USART6 on STM32F401

dapetcu21:
STM32F401 has USART6 on PA12/PA11 and PC7/PC6 with alternate function mapping AF08. This can be used, for example, to connect to the Elegoo Neptune 3 without soldering to the main board or even opening the case, where PA12/PA11 are wired to a convenient RJ10 plug going to the stock screen.

See the datasheet (Page 45, Table 9. Alternate function mapping):

https://www.st.com/resource/en/datasheet/stm32f401re.pdf

See the Neptune 3's schematic:

https://github.com/TheFeralEngineer/Elegoo-Neptune-board-schematics/blob/main/ZNP%20Robin%20Nano_DW%20V21_001%20SCH_加水印pdf%20·%20version%201_230115_223844.pdf

I only tested this on my Neptune 3. It might work for other STM32s that have USART6, if there are any others, but I only added a Kconfig entry for STM32F401, since that's what I have access to for testing.

### DIFF
--- a/src/stm32/Kconfig
+++ b/src/stm32/Kconfig
@@ -434,6 +434,12 @@ choice
         bool "Serial (on USART5 PD2/PD3)" if LOW_LEVEL_OPTIONS
         depends on MACH_STM32G0Bx
         select SERIAL
+    config STM32_SERIAL_USART6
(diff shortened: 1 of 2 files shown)


KevinOConnor:
Thanks.

-Kevin

PR #6831, opened 2025-02-22
 example-easythreedk9.hex

3dcuber:


JamesH1978:
As mentioned by @Sineos , This PR changes an example kinematics file, with some of those changes quite dangerous for some printers wanting to use that file, mainly setting max_accel to 50000 and max_z_accel to 3000.

If you intended to add a new file or edit a different one, please raise a new PR, but i have now applied the "wontfix" label to this

Thanks
James

PR #6832, opened 2025-02-23
Allow any probes named eddy to perform rapid_scan

vvuk:
[eddy-ng](https://github.com/vvuk/eddy-ng) currently has to patch `bed_mesh.py` in order to allow a `probe_eddy_ng` object to be used for rapid scan. This PR basically upstreams that change so that I don't have to ask people to patch bed_mesh, which can complicate klipper updates.

It allows any probe with "eddy" in its name can do rapid scan, or any probe that reports True for `rapid_scan_supported` in its status.

### DIFF
--- a/klippy/extras/bed_mesh.py
+++ b/klippy/extras/bed_mesh.py
@@ -865,8 +865,10 @@ class ProbeManager:
         can_scan = False
         pprobe = self.printer.lookup_object("probe", None)
         if pprobe is not None:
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  However, we don't add code to the main Klipper repo to facilitate external modifications.  Any change to this repo would need to add user facing functionality for those running the unmodified code here.

-Kevin

PR #6833, opened 2025-02-24
Update Delta_Calibrate.md

chexxor:
Clarify delta calibration and geometry.
Provide an idea of a symptom of incorrect X, Y dimensional accuracy and provide a lead towards fixing it.

I had an issue with calibrating my delta printer, and part of that was due to misunderstanding delta geometry.
I was confused on what "delta_radius" means in klipper. It could mean bed radius or print area radius, but it is actually referring very exactly to delta kinematics-specific geometry. It is the X-Y distance between arm centers when in the parked position. The RepRap page on the topic is a useful place to learn a bit more about it.

### DIFF
--- a/docs/Delta_Calibrate.md
+++ b/docs/Delta_Calibrate.md
@@ -47,12 +47,16 @@ radius.
 
 In order to perform this calibration the initial delta parameters (arm
 lengths, radius, and endstop positions) must be provided and they
-should have an accuracy to within a few millimeters. Most delta
-printer kits will provide these parameters - configure the printer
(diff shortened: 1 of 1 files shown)


KevinOConnor:
Thanks.  In general, it seems fine to me.  However, links to other documents should use relative links - replace `[Config_Reference](https://www.klipper3d.org/Config_Reference.html#common-kinematic-settings)` with `[Config_Reference](Config_Reference.md#common-kinematic-settings)`.

-Kevin

PR #6835, opened 2025-02-25
sensor_lis2dw: remove commented code and fix formatting

Wulfsta:
I was recently looking at this file to see how I might expose the temperature sensor in the LIS3DH, and noticed I left a line of dead code commented last time I worked on it. This is a trivial change to make this file prettier, because the current state was annoying me! :laughing:


KevinOConnor:
Thanks.

-Kevin

PR #6837, opened 2025-02-25
RFC: Cooperative Heaters predictive control

nefelim4ag:
Some time ago [I dug heaters a little on the Discourse](https://klipper.discourse.group/t/discussion-proposal-small-heater-control-improvements/19195)

I don't think PID is bad or needs to be replaced.
_I tried to do some fun stuff, like train DNN or write my own algo to control the power._
In the end, everything looks the same.

There are historically various topics, that try to replace it or tune it differently.
I think the only practical reason is to handle something important for that human being.

So, there is an ultimate solution, I think.
Pid is still under the control of a heater and is able to drive it.
The end-user has the ability to adjust it freely, like in my doc example - drive PWM in a feedforward way if the fan is suddenly enabled, like for bridges.

If it is needed, it is definitely possible to get the current chamber temperature, adjust to the fan curve, try to parse the extruder moving queue (IDK) and drive PWM higher before the temperature falls and PID will reactively compensate for that.
_I hope that will add enough flexibility to stop rituals like "how to do PID autotune properly"._

Another example, for some reason people do not insulate the heating beds, adding thermistors inside a thick aluminum plate and then want sort of PID control by two thermistors, and feel that combined sensor is not good enough.

Well, now it is possible to literally query the sensors and decide what to do, like return the `-N` value and decrease bed heater power on demand.

I hope it is made in a way to is safe and acceptable, leaves all existing safety measures in place, and mostly does not touch safety-critical code.

Thanks.

---
Btw, I will try to add templates to temperature FANs in a similar way, to fit most of the wishes, like curves, formulas & etc.

---
```
cd ~/klipper
git fetch origin pull/6837/head
git checkout FETCH_HEAD
sudo systemctl restart klipper
```

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -5579,6 +5579,17 @@ cs_pin:
 #   above parameters.
 ```
 
+### [heater_pc]
+
+Heater prediction correction.
+To use this feature, define a config section with a "heater_pc" prefix
+followed by the name of the corresponding heater config section.
+For example `[heater_pc heater_bed]`
+```
+[heater_pc extruder]
+#macro_template: <display_template's name>
+```
+
 ## Common bus parameters
 
 ### Common SPI settings
--- a/klippy/extras/heater_pc.py
+++ b/klippy/extras/heater_pc.py
@@ -1,0 +1,71 @@
+# Klipper Heater Predictional Control
+#
+# Copyright (C) 2025  Timofey Titovets <nefelim4ag@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+import logging
+import threading
+from . import output_pin
+
+class HeaterPredictControl:
+    def __init__(self, config):
+        self.printer = config.get_printer()
+        self.reactor = self.printer.get_reactor()
+        self.config = config
+        self.eval_time = 0.3
+        self.min_pwm = -1.0
+        self.max_pwm = 1.0
+        self.old_control = None
+        name_parts = config.get_name().split()
+        if len(name_parts) != 2:
+            raise config.error("Section name '%s' is not valid"
+                               % (config.get_name(),))
+        # Use lock to pass data to/from heater code
+        self.lock = threading.Lock()
+        self.output = .0
+        self.pwm_event_time = self.reactor.monotonic()
+        # Link template
+        self.template_name = config.get("macro_template")
+        # Template handling
+        self.template_eval = output_pin.lookup_template_eval(config)
+        self.printer.register_event_handler("klippy:connect",
+                                            self.handle_connect)
+    def handle_connect(self):
+        pheaters = self.printer.load_object(self.config, 'heaters')
+        heater_name = self.config.get_name().split()[-1]
+        heater = pheaters.heaters.get(heater_name)
+        if heater is None:
+            self.config.error("Heater %s is not registered" % (heater_name))
+        self.eval_time = heater.get_pwm_delay()
+        self.min_pwm = -heater.get_max_power()
+        self.max_pwm = heater.get_max_power()
+        self.old_control = heater.set_control(self)
+        gcode = self.printer.lookup_object('gcode')
+        params = {"TEMPLATE": self.template_name}
+        fo_gcmd = gcode.create_gcode_command("", "", params)
+        self.template_eval.set_template(fo_gcmd, self._adj_pwm)
+    def _adj_pwm(self, text):
+        try:
+            output = float(text) * self.max_pwm
+        except ValueError as e:
+            logging.exception("heater_pc template render error")
+            output = .0
+        output = max(self.min_pwm, min(self.max_pwm, output))
+        with self.lock:
+            self.output = output
+    def temperature_update(self, read_time, temp, target_temp):
+        output = .0
+        with self.lock:
+            output = self.output
+        # Returns +- max_power
+        co = self.old_control.temperature_update(read_time, temp, target_temp)
+        co += output
+        return co
+    def check_busy(self, eventtime, smoothed_temp, target_temp):
+        res = self.old_control.check_busy(eventtime, smoothed_temp,
+                                          target_temp)
+        return res
+
+def load_config_prefix(config):
+    return HeaterPredictControl(config)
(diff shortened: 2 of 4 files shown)


Sineos:
Nice!
It seems to follow the same logic as display templates. Maybe stick to the general wording and naming from there?

nefelim4ag:
>  Maybe stick to the general wording and naming from there?

I imported some changes from there, and I hope, now it is a little better.

Sineos:
From the eyes emoticon, I gather you found my comment confusing. What I tried to say:

The display template functions in Klipper for LEDs and fans ([example](https://www.klipper3d.org/Config_Reference.html#display_template)) seem to be using the same overarching logic, albeit with a bit of different nomenclature. 
Maybe I'm misunderstanding something?

nefelim4ag:
> The display template functions in Klipper for LEDs and fans ([example](https://www.klipper3d.org/Config_Reference.html#display_template)) seem to be using the same overarching logic, albeit with a bit of different nomenclature.

I got the idea of what you meant after your comment in the temperature_fan PR.
I'm looking for what can be done to make it similar to current integration with `[display_templates]'.
_The current integration feels a little unintuitive from my perspective, but it is better to reuse it, of course._

_Thanks for checking. I was totally unaware of that functionality._

nefelim4ag:
I don't know how I feel about adding a G-code command to reload the template or directly alter the PWM output. 
I also think it is wrong to add custom value get/set/log functions to all templates by incorporating them in `display_template`.
So, I dropped them for now.

So, for now, I only leave here a bare config binding between template and heater.
If it is needed, technically, temperature and read time can be propagated in the template context.
But I think they are mostly useless without state management inside macro.

Thanks.

KevinOConnor:
Interesting.  I'm not sure this implementation is safe though, as the PID `temperature_update()` code is called from a background thread (`serialhdl:_bg_thread()`) and I don't know what would happen if one tried to render templates from that thread.  Certainly if the template accessed any of the `{printer.module.xxx}` values it would result in `get_status()` calls from the background thread which would not be safe.

I suspect we could alter the target temperature from templates without thread issues (or set a temperature offset), but I'm not sure if that's what you are looking for.

Still interesting though.
-Kevin

nefelim4ag:
> code is called from a background thread (serialhdl:_bg_thread()) and I don't know what would happen if one tried to render templates from that thread

I will change it to run in a similar fashion to current fan/led templates in the main/reactor thread.
And pass shared value to offset PWM.
It seems simpler to do locks with one shared value.
The only tricky part would be to run it close before PWM calculation to make it semi-synchronized.
Maybe use the last PWM evaluation time and schedule the timer in <300ms after that.
Even though it will mostly retain the suggested functionality.

> I suspect we could alter the target temperature from templates without thread issues (or set a temperature offset), but I'm not sure if that's what you are looking for.

AFAIK, it is only useful if the target is to use variable extrusion temperature based on the flow. AFAIK, there is at least one project for that: https://github.com/sb53systems/G-Code-Flow-Temperature-Controller

It can be made on the Klipper side, but I suspect it is requires different design to trigger temperature changes ahead and do other stuff like accounting for thermal inertia/slow down printing.
So, I suspect it is easier to do as an external extension/gcode preprocessor.

Thanks.

nefelim4ag:
I realized there is no example, so this one is for the extruder heater with accounting for fan losses and the losses due to the volumetric flow.

```
[display_template extruder]
text:
  {% set pwr_adj = printer.fan.speed * 0.15 %}
  {% set flow = printer.motion_report.live_extruder_velocity * (1.75 / 2)**2 * 3.1415 %}
  # 25 is around max flow, a simple, linear equation
  # 25 mm^3/s requires at least +40% of PWM
  { pwr_adj + flow/25 * 0.4 }

[heater_pc extruder]
macro_template: extruder
```

nefelim4ag:
Example of 2 termistors bed control to limit the bed heater temperature.
https://klipper.discourse.group/t/thick-bed-temperature-controll/24940/48
```Jinja2
[display_template bed_limiter]
text:
  {% set bed_heater_temp = printer['temperature_sensor bed_heater'].temperature | float %}
  {% set max_temp = 100 %}
  {% set range = 5 %}
  {% set proportional_damping = (bed_heater_temp - (max_temp - range / 2))/range %}
  {0 if proportional_damping < 0 else proportional_damping * -1 }

[heater_pc heater_bed]
macro_template: bed_limiter
```

This is my example of reducing the chamber heater power until the heater bed is done to heat, to limit the peak power.

```Jinja2
[display_template chamber]
text:
  { -1.0 * printer["heater_bed"].power }

[heater_pc chamber]
macro_template: chamber
```

---
~~There is an error with how I rescheduled the timer before the PWM code, so the template runs several times before the next PWM arrives. I will fix that.~~
Fixed.

nefelim4ag:
I had a long idea that my initial implementation with hooks into the PID loop is hacky.
After recent conversations around PID Calibration, Integer windups, & etc, I realized that actually, I can make it pluggable.
Also, there are temperature fans' control loops, which are mostly copy paste of heaters' control loops.

So, here is internal refactoring for the heaters' control, which now returns control output.
That way, without any hooks, I can amend the control with feed-forward with simple wrappers.

The same refactoring can be done for the fans, where the same control loop can be reused; the only difference is that temperature fans need to invert the control output, basically do `-1 * co`.

With that, it is possible to later reuse the existing PID_CALIBRATE for the fans.

Thanks,
-Timofey

---
Hmmm,
It is also possible to reuse the existing `PrinterTemplateEvaluator`. But I feel it would make sense to make templates less ambiguous, `output_pin_template`, `led_template`, and so on, and to validate the output during the connect phase.
Instead of "silently" logging the template errors.

---

Reused

dewi-ny-je:
> It is also possible to reuse the existing `PrinterTemplateEvaluator`. But I feel it would make sense to make templates less ambiguous, `output_pin_template`, `led_template`, and so on, and to validate the output during the connect phase. Instead of "silently" logging the template errors.

What's the status?

nefelim4ag:
> It is also possible to reuse the existing PrinterTemplateEvaluator

@dewi-ny-je as stated in the above post: "Reused".

>  But I feel it would make sense to make templates less ambiguous, output_pin_template, led_template, and so on ...

Where are any changes to config/templates in general that are somewhat related to my general view: https://klipper.discourse.group/t/rfc-fan-refactoring/25800
So, unless there is a decision that we want to refactor the template config definitions, there is no point in pushing that.

Otherwise, it is as ready as it was before.

-Timofey

PR #6838, opened 2025-02-26
Add support for gd32f303xe MCU temperature used on Creality K1 Max (and others)

happytechca:
### Summary
This PR adds support for the gd32f303xe MCU, which is used in the Creality K1 Max and other printers, by adding necessary code into temperature_mcu.py.

### Background
I recently converted my Creality K1 Max printer to mainline Klipper, moving away from Creality's forked version. However, I discovered that the gd32f303xe MCU—which is used for the toolhead or temperature sensor MCU—is not currently supported in upstream Klipper.

Error message upon starting: **MCU temperature not supported on gd32f303xe**

image

### Changes Made

- Modified temperature_mcu.py to add support for the gd32f303xe MCU.
- Only 4 lines of code were added to include this MCU in the supported list.

### Testing

- Verified functionality by ensuring correct temperature readings and general toolhead operation.
- Printed multiple test models without issues.
- No other functionality is affected beyond enabling this MCU.

Signed-off-by: Serge Latulippe <happytechcanada@gmail.com>

### DIFF
--- a/klippy/extras/temperature_mcu.py
+++ b/klippy/extras/temperature_mcu.py
@@ -80,6 +80,7 @@ class PrinterTemperatureMCU:
             ('stm32l4', self.config_stm32g0),
             ('stm32h723', self.config_stm32h723),
             ('stm32h7', self.config_stm32h7),
+            ('gd32f303xe', self.config_gd32f303xe),
             ('', self.config_unknown)]
         for name, func in cfg_funcs:
             if self.mcu_type.startswith(name):
@@ -164,6 +165,9 @@ class PrinterTemperatureMCU:
         cal_adc_110 = self.read16(0x1FF1E840) / 65535.
         self.slope = (110. - 30.) / (cal_adc_110 - cal_adc_30)
         self.base_temperature = self.calc_base(30., cal_adc_30)
+    def config_gd32f303xe(self):
+        self.slope = 3.3 / -.004100
+        self.base_temperature = self.calc_base(25., 1.45 / 3.3)
     def read16(self, addr):
         params = self.debug_read_cmd.send([1, addr])
         return params['val']


nefelim4ag (approved):

KevinOConnor:
Thanks, but this Klipper repo doesn't support mcus that report gd32f303 as the cpu type, and we don't add code to this repo to support external modifications.

-Kevin

happytechca:
@KevinOConnor Understood, thank you.
Any plans on having support for GD32 based MCUs? What are the roadblocks?
From my understanding, they are quite similar to the STM32Fxxx series

KevinOConnor:
My rough understanding is that the gd32f303 chips are supported by Klipper - one selects an stm32f103 chip during "make menuconfig".  I'm not sure what code you are running.

-Kevin

JamesH1978:
I can confirm GD32F303 and GD32F103 are clones that are code compatible with the STM32F103 chip. With the GD32F303 actually being more compatible than its GD32F103 counterpart, as that needs the `Disable SWD at startup` setting in the make, whereas the F303 usually does not.

Creality have been using these since pandemic times, as they were more available and likely cheaper, also Sovol uses them too. Klipper compiled firmware works on these fine. Unless the K1 has modifications to the hardware that would make it not work.



Thanks
James

Hawk84r:
> My rough understanding is that the gd32f303 chips are supported by Klipper - one selects an stm32f103 chip during "make menuconfig". I'm not sure what code you are running.
> 
> -Kevin

They are supported and work fine. Just reading they temperature is not, which is fixed by this commit. And it also works fine.

kemsky:
@happytechca, could you can share full configuration? Which pins did you select as serial?  I can flash K1 toolhead but something seems to be  wrong, it does not start, pellcorp firmware does work.

JoveYu:
@happytechca Could you please tell me how you configured the prtouch_v2 heated bed when using the mainline klipper on K1?

henrivdr:
Hi @happytechca
See https://github.com/Klipper3d/klipper/pull/7302

PR #6839, opened 2025-02-26
jinja2: Expand template syntax errors

nefelim4ag:
Sometimes, writing macros is painful because jinja does not show adequate errors.

I tried to fix it, so now template syntax errors show that manner:
```
Internal error during connect: Error loading template 'temperature_fan octopus'
line 4: {%- set tmax = tt + 1. -%} # expected name or number
```

Thanks.

---
_Errors made me implement that fix while I was working on https://github.com/Klipper3d/klipper/pull/6837 and https://github.com/Klipper3d/klipper/pull/6840_

### DIFF
--- a/klippy/extras/gcode_macro.py
+++ b/klippy/extras/gcode_macro.py
@@ -49,6 +49,12 @@ class TemplateWrapper:
         self.create_template_context = gcode_macro.create_template_context
         try:
             self.template = env.from_string(script)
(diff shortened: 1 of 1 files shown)


thijstriemstra (commented):
It does not appear to show the filename or path where the error occurred, is this possible as well?

KevinOConnor:
Thanks.

-Kevin

PR #6848, opened 2025-03-11
Debounce gcode_button and filament_switch_sensor

garethky:
This change adds a `debounce_delay` option to both `filament_switch_sensor` and `gcode_button`. This allows for much longer, user controllable, debouncing. Event cycles (off->on->off) that happen during the delay will be ignored. At the end of the delay, if the state transition is singular and maintained the event fires.

This is based on a request from a user that had issues with a fickle filament switch: https://klipper.discourse.group/t/proposal-button-debounce/22212

This allows printer owners to solve issues with less than optimal physical filament switch design. It also allows for ideas like "push and hold to execute" e.g. push and hold a button for 200ms to run `M112` so we know the user **_really_** meant it. Debouncing at this sort of timescale is widely used to add polish to user interfaces and input devices.

I argued in the thread that any sensing technology that converts its inputs into the equivalent of an on/off signal should get this treatment. This solves for spurious/hesitant/non-committal user inputs which could be seen as an event by a sensor (such as filament motion sensor) and still need to be ignored (e.g. user inserts and then quickly removes filament). But for now its limited to things that are driven by "physical switches".

This also has the minor change to set `event_delay` on a `filament_switch_sensor` to 0. Since the debounce delay can now essentially do what this option was trying to do I want users to disable this delay completely if they wish.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -3278,6 +3278,10 @@ pin:
 #   A list of G-Code commands to execute when the button is released.
 #   G-Code templates are supported. The default is to not run any
 #   commands on a button release.
+#debounce_delay:
+#   A period of time in seconds to debounce events prior to running the
+#   button gcode. If the button is pressed and released during this
+#   delay, the entire button press is ignored. Default is 0.
 ```
 
 ### [output_pin]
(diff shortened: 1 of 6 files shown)


KevinOConnor:
Thanks.  In general, it seems useful to me.

One suggestion would be to introduce a `buttons.register_debounce_button()` (and `buttons.register_debounce_adc_button()`) so that callers don't need to manually import the `buttons` module.

Cheers,
-Kevin

garethky:
Updated with `buttons.register_debounce_button()` and `buttons.register_debounce_adc_button()`

I don't know how you feel about it, but I kept the code that reads the config centralized. Buttons currently doesn't read any config from callers. 🤷‍♂️

KevinOConnor:
Seems fine to me.

@Arksine , @TheJoshW - fyi.

-Kevin

Arksine:
Looks good to me as well.

EmperorArthur:
Looks good.  I've been running an older version of this code for a bit now with no issues.

I temporarily modified my sensor to false trigger and verified that version worked as expected.

KevinOConnor:
Thanks.

-Kevin

PR #6849, opened 2025-03-12
extras: Discover custom extras using entrypoints

Jomik:
Here's my quick proposal for supporting custom extras as we see popping up.
The goal here is to ease user installation / removal, isolate custom code, keep klippy clean.
This proposal makes use of [Entry points specification](https://packaging.python.org/en/latest/specifications/entry-points/), which allows python packages to make themselves discoverable once installed.

I attempted to keep the implementation minimal to allow expansion, if desired.

### Examples:
A minimal example of a package project definition that would be discoverable with the proposal.
The developer would publish this on pypa, the user will then install the package into klippy venv `~/klippy-env/bin/pip install foo`. Moonraker also supports python packages, so it can keep it up to date.
```toml
# pyproject.toml
[project]
name = "foo"
requires-python = ">= 3.8"

[tool.hatch.build.targets.wheel]
packages = ["src/foo"]

[project.entry-points."klippy.extras"]
foo = "foo.extra"
```
```py
# src/foo/extra.py
def load_config(config):
	# Do stuff
	logging.info("Loaded foo!")
```
```yaml
# printer.cfg
[foo]
```

### Considerations:
- Use [pkg_resources](https://setuptools.pypa.io/en/latest/pkg_resources.html#entry-points) for older versions of python.
- Complexity may increase if we desire to use `pkg_resources` or remove the `importlib_metadata` compatibility backport for older versions of python.
If we need to support this further back, we should consider creating a memoized function to load the module, and not expose entrypoints to `load_object` at all.
- Expose list of loaded components in the API - the status of each component can be queried normally.

### Related:

- [Possible Klipper “plugins” instead of macros?](https://klipper.discourse.group/t/possible-klipper-plugins-instead-of-macros/9819)
- [custom_component: allow to load python components natively (#6492)](https://github.com/Klipper3d/klipper/pull/6492)


nefelim4ag:
There is no plugin or custom extras system.
https://github.com/Klipper3d/klipper/pull/6492#issuecomment-2018433357

So, there should be no custom code in the extras and, as a consequence, no special code to sideload them.

The only proposed system so far is:
https://klipper.discourse.group/t/new-proposal-for-klipper-extension-support/19150

No one is restricting you from hacking around; it is cool and encouraged.
But there should be no code for a feature that does not exist.

Btw, it would be nice to read:
https://github.com/Klipper3d/klipper/blob/master/docs/CONTRIBUTING.md#what-to-expect-in-a-review

Thanks

Jomik:
> There is no plugin or custom extras system. [#6492 (comment)](https://github.com/Klipper3d/klipper/pull/6492#issuecomment-2018433357)
> 
> So, there should be no custom code in the extras and, as a consequence, no special code to sideload them.
I am aware, but this is definitely a tendency, and currently made with the caveat that it's hacky and can interfere with users' klipper installation.
I see no reason that we shouldn't make a small change, such as this, to better support this behavior that is seen and used by the consumers.
I think it's fair to say that the only stable thing is that you get the `config` object where you can read the configuration of the extra in question.
I don't see that adding this would be the same as promising that the extras (`lookup_object`) would be stable..

I hadn't seen [klipper.discourse.group/t/new-proposal-for-klipper-extension-support/19150](https://klipper.discourse.group/t/new-proposal-for-klipper-extension-support/19150).
I personally feel like my proposed solution here is better.
It's less code, less to maintain and more realistic to have happen.
It'll be easy to adopt for current community extras.

Of course, once an actual plugin API has been defined, this could be retired and replaced by a similar implementation, using a different group - e.g. `klippy.plugins`.


I am aware of the signed off - I'll do that if this is ready to be accepted, so that's pending @KevinOConnor I suppose..

PR #6852, opened 2025-03-12
Support step on both edges with custom minimum pulse duration

KevinOConnor:
Add support for "step on both edges" to the main `stepper_event_full()` code.  This makes that mode of operation available even when the micro-controller is not compiled for "optimized step on both edges".  It also enables the custom pulse duration support (step_pulse_ticks) when in "step on both edges" mode.

This is useful for very fast mcus that can potentially step faster than 100ns (which is the maximum rated pulse duration for trinamic drivers).  It allows these mcus to continue to use "step on both edge", as they can now verify a minimum time between edges.  This may also useful on MCU types that don't support the "optimized step on both edges" code - as they can now obtain at least some of the benefit of "step on each edge".

Along with the above change, this PR also makes it possible for users to enable/disable "optimized step on each edge" in "make menuconfig".  After this change, the benefit of "optimized step on each edge" is about a 20-25% performance improvement when using trinamic drivers, but disabling that optimization results in about a 20-30% performance improvement for non-trinamic drivers (eg, bldc servos).

This was discussed in #6806 .

@nefelim4ag - fyi.

-Kevin


Sineos:
Maybe a stupid question, but this sounds very interesting for the MagnetoX: Are there any drawbacks to be expected when using, for example, the first two driver slots in the optimized Non-TMC mode and another four slots equipped with TMCs for the Z-steppers?

KevinOConnor:
If you have a printer that uses non-trinamic drivers for XY motion, then you'll likely benefit from this PR.  It'll allow optimizing the code for non-trinamic drivers.  (Motion on the Z steppers will be slightly less optimized, but it's a minor performance difference and one does not typically drive the Z steppers at high velocities.)

If you have a stm32h7 with non-trinamic drivers then #6853 is a pretty big performance improvement - around 50% more steps per second for non-trinamic drivers.  @nefelim4ag  is working on further performance improvements as well.

Cheers,
-Kevin

nefelim4ag (approved):
Great work, LGTM

klippy/stepper.py
```diff
@@ -75,13 +76,33 @@ def _build_config(self):
...
-        if (self._req_step_both_edge and sbe
-            and self._step_pulse_duration <= MIN_BOTH_EDGE_DURATION):
-            # Enable stepper optimized step on both edges
+        # Check if can enable "step on both edges"
+        constants = self._mcu.get_constants()
+        ssbe = int(constants.get('STEPPER_STEP_BOTH_EDGE', '0'))
+        sbe = int(constants.get('STEPPER_BOTH_EDGE', '0'))
+        sou = int(constants.get('STEPPER_OPTIMIZED_UNSTEP', '0'))
+        want_both_edges = self._req_step_both_edge
+        if self._step_pulse_duration > MIN_BOTH_EDGE_DURATION:
```
nefelim4ag, line 85: btw, does it make sense to disable it for TMC?
With the new code, it should just fall inside `stepper_event_full()`

Or is it just for backward compatibility with the SBE option?

KevinOConnor: Thanks for reviewing.  I don't understand your comment here.  Are you proposing alternate code, and if so what code changes are you proposing?

Cheers,
-Kevin

nefelim4ag: I was not able to get the purpose of this one branch.

It looks like there are already several if/then statements, and if we reduce the count a little, it will be simpler.
May  I suggest to drop it alongside with constant?

I'm okay with either outcome.

KevinOConnor: I added the `if self._step_pulse_duration > MIN_BOTH_EDGE_DURATION:` check because
```
# If user has requested a very large step pulse duration
# then disable step on both edges (rise and fall times may
# not be symetric)
```

When we first deployed the "step on both edge" optimization we found that some boards couldn't handle it (some discussion at https://github.com/Klipper3d/klipper/pull/4976 ).  Ultimately, this issue was worked around by using a larger step_pulse_duration on these boards.

With this new code, we could potentially support "step on both edges" on these older boards.  However, I'm concerned that doing so may actually break things - as I understand it, the level shifters on these older boards weren't symmetrical - I think they had a rapid voltage rise time, but a very slow voltage drop time.  So, enabling "step on both edges" may mess up the timing and thus make things worse on these old boards.  Ultimately I'd rather not introduce a regression.  That's why I added the explicit check to disable "step on both edges" if the user has requested a large pulse_width_duration.

Let me know if that answers your question.
-Kevin

PR #6853, opened 2025-03-12
Optimized gpio on stm32h7

KevinOConnor:
This PR optimizes the low-level `gpio_out_toggle()` code on stm32h7 mcus.  It seems the stm32h7 takes an absurdly long time to read gpio IO registers, and thus the typical stm32h7 toggle code (`regs->ODR ^= g.bit`) takes 15+ cpu cycles to complete.  It's possible to optimize this by keeping a local cache of the last value written to the ODR register.

This change provides a notable performance improvement to the step benchmarks - so much so that the "optimized step on both edges" code would likely exceed the minimum time needed between "edges" on trinamic drivers.  As a result, this PR is on top of #6852 , and this PR disables the "optimized step on both edges" for stm32h7 mcus.  Even with that feature disabled, the performance is still very good - the maximum number of steps per second for the 3-stepper benchmark increases from 6061K to 6486K.  There are also plans to further increase the performance of the stm32h7 mcus.

This was discussed in #6806.

@nefelim4ag - fyi.

-Kevin


KevinOConnor:
@D4SK , @adelyser - fyi.

nefelim4ag:
I've applied the whole patchset, and everything works (H7 400Mhz).
The only thing that needs some attention is an SW SPI.
With those changes, it has higher clocks 4Mhz -> 4.5Mhz (which is higher than requested by TMC code).

Otherwise lgtm.

Thanks.

---
image

Technically, with a 12Mhz internal clock, TMC should work with up to 6Mhz, with a recommended safety margin of 5.4Mhz.
IIRC, from my limited testing, 5160 and 2240 start to experience issues closer to 5Mhz.

_I have a 10Mhz device on the same bus, which works with 7.5Mhz, so I can only think there are some TMC limitations._

nefelim4ag:
> Just to be clear, do you think we should hold off on merging this until after the software spi changes are completed and merged?

IDK will it broke someone or not.
So, let's hold it a while.

I made an SW SPI PR: #6856

nefelim4ag:
JFYI:
I spent some time trying to find the TMC SPI speed limit, so I overclocked STM32H7.
SCK frequency is measured with an oscilloscope, and it is the maximum frequency that I can get.

Clock Freq | SCK Freq | Status
-- | -- | --
480Mhz | 4.9Mhz | Works
512Mhz | 5.2Mhz | Works
550Mhz | NA | Errors with any frequency
 
 I have only TMC5160 and TMC2240, and all supported SPI TMC have similar internal frequency, so, similar SPI speed.
 So I think this PR is safe to merge.
 
 Thanks

nefelim4ag (approved):

nefelim4ag:
JFYI:
I may suggest a small speedup.
The basic idea is to use BSRR, cache handling can be simplified if there is a cache per pin.
As a consequence, toggle can also be shorter, because it is only needed to swap bits.
It looks like even with BSRR, with a cache, irq disable can't be removed.
But I may be missing something here.
(this pr)
`v0.12.0-461-g0e2d61d3e / gcc: (15:14.2.rel1-1) 14.2.1 20241119 binutils: (2.40-2+18+b1) 2.40`
61/185
(https://github.com/nefelim4ag/klipper/commit/1ad8392fb22bc2062dae5d2b5a3f8af1b07a11a1)
`v0.12.0-462-g1ad8392fb / gcc: (15:14.2.rel1-1) 14.2.1 20241119 binutils: (2.40-2+18+b1) 2.40`
60/183

KevinOConnor:
Interesting, thanks.  It seems your implementation eliminates one memory read (`stepper.step_pin.bit`) at the cost of increasing the cache footprint of ODR_CACHE from ~44 bytes to ~704 bytes.

> It looks like even with BSRR, with a cache, irq disable can't be removed.

I think it should be possible to remove the irq_save/restore.  It's necessary to ensure that a GPIO update from an irq does not corrupt a GPIO update from a task *of a different pin*.  (For example, updating PA3 from an irq should not corrupt an update of PA2 from a task.)  I don't know of any cases where the same gpio could be modified at the same time from both an irq handler and a task.  I'm not sure it's worth optimizing though.

-Kevin

nefelim4ag:
>  I'm not sure it's worth optimizing though.

JFYI,
I did play with the drop of the `IRQ` code on top of BSRR, it reduces code size slightly optimized and makes SW SPI even faster.
It's not like there is anything important for this specific chip or there is a real usage for that speed.

KevinOConnor:
FYI, I think this PR should wait until after software spi timing fixes have been merged ( #6856 ) and after tagging v0.13.0 .

-Kevin

EmperorArthur:
I have been running this PR along with #6806, and adjusting the frequency to 480Mhz. It fixed the issues I was having with my Magneto X.

No problems so far.

KevinOConnor:
FYI, I updated this branch to use @nefelim4ag BSRR suggestion, but a slightly different implementation.

I do get the 60/183 benchmark results with this branch now.

Thoughts?
-Kevin

nefelim4ag:
```
    union {
        struct odr_cache *oc;  // stm32h7 uses 'oc'; all others use 'bit'
        uint32_t bit;
    };
...
uint32_t pin = g.oc - ODR_CACHE;
```
Pretty clever

Thanks, lgtm

EmperorArthur:
> ```
>     union {
>         struct odr_cache *oc;  // stm32h7 uses 'oc'; all others use 'bit'
>         uint32_t bit;
>     };
> ...
> uint32_t pin = g.oc - ODR_CACHE;
> ```
> Pretty clever

Agreed.

It took me way too long to remember those chips use 32 bit pointers.  Then this clicked for me.

KevinOConnor:
Okay, thanks.  I made a few more tweaks and committed this change.

-Kevin

PR #6854, opened 2025-03-12
axis_twist_compensation: allow compensating both axis at once

blastrock:
Restores the behavior before #6739 since people seemed to rely on it, even if the math is not exact.


KevinOConnor:
Thanks, but I also need a signed-off-by line.

Cheers,
-Kevin

blastrock:
Sure, fixed!

KevinOConnor:
Thanks.

-Kevin

PR #6855, opened 2025-03-13
PR: Support re-using sensors in z_thermal_adjust

garethky:
This came up when developing load cell based probing but I believe its also works with any kind of nozzle probe, e.g. Voron Tap.

## The Problem
Probing temperatures are limited based a couple of factors: The glass transition temp of the filament and the melting temperature of the bed surface. So probing is carried out at about 140-160C. Printing temps for PETG might be 290C. So there is, potentially a 150C temperature increase between probing and printing.

Investigations on my printer shows that the rate of thermal expansion for the nozzle is -0.000465mm per degree C. Over 150C that is 0.069mm / 69 microns or about 35% of the thickness of a first layer. The general result of this is first layers that are too low.

## Proposed Solution
* Extend `z_thermal_adjust` to support sourcing a temperature from an existing heater for compensation
* Extend `z_thermal_adjust` to support compensating for multiple components in the printer. So the Frame and the Nozzle can be compensated and the machine can adapt enclosed open conditions.

I've been testing this in conjunction with load cell probing and the results have been very encouraging. The effect is temperature dependent and easy to measure. A GCode script can measure the nozzle length via probing over a range of temperatures and then produce the thermal coefficient through linear regression. I probed the aluminum bed surface directly with a clean nozzle and no filament loaded. For PCB beds you might have to put a steel surface on top, like the uncoated side of a steel sheet.

The real test is comparing first layers of materials at different temps. I've been testing swapping between PLA and PETG. With other variables removed (keeping the door open, using the same print surface) this lets me swap filaments and get very similar results between them.


docs/Config_Reference.md
```diff
@@ -1262,6 +1264,10 @@ See also: [extended g-code commands](G-Codes.md#z_thermal_adjust).
 #max_z_adjustment:
 #   Maximum absolute adjustment that can be applied to the Z axis [mm]. The
 #   default is 99999999.0 mm (unlimited).
+#sensor:
```
nefelim4ag, line 1267: Just a suggestion: It should be possible to reuse any existing sensor by:
`sensor_type: temperature_combined`

So, it may not be necessary to reimplement same functionality here.

garethky: Yes, that does work with config like this:

```
[z_thermal_adjust nozzle]
temp_coeff: -0.00046459
sensor_type: temperature_combined
sensor_list: extruder
combination_method: max
min_temp: 0
max_temp: 400
maximum_deviation: 1
max_z_adjustment: 0.5
```

I can remove the `sensor` option. I think this idea of a `sensor_list` is great for the frame expansion as well.

garethky: Updated to remove the `sensor` stuff. That makes the PR really short now. 😄

KevinOConnor:
Thanks.  I'm not sure I'm a good reviewer for this.  @alchemyEngine - do you have any comments?

-Kevin

alchemyEngine:
Hi all, sorry for the delay on my end.

@garethky Really cool feature! Thank you for this, I think this functionality would be extremely beneficial to those using nozzle probes.

> Using a macro to record the nozzle length over a temperature range and derive the expansion coefficient through linear regression yields a plot like this:

I find the measurement pattern really interesting here, with one consistently lower than the other. Looks like the nozzle is cooling slightly while contacting the aluminum bed?

On a higher level, I'm not sure if this fits in with the intended functionality of the `z_thermal_adjust` module: print time continuous adjustment of the Z coordinate space. If I understand correctly, this is only utilized during probing to dynamically adjust the probe offset, and as such would be better suited to piggybacking on the probe code, or as a dedicated module. 

In summary I think their functionality is largely unrelated, and as such I don't think it would be wise to link their code. I'm open to discussion, however.

garethky:
> I find the measurement pattern really interesting here, with one consistently lower than the other. Looks like the nozzle is cooling slightly while contacting the aluminum bed?

The lower ones were taken as the nozzle heated up, the higher ones as it cooled down. I read this as the time delay required to get the nozzle and the heatbreak temp to stabilize. Its just longer than the 1 minute that I waited when I collected the dataset. But linear regression across both should give a better result.

> If I understand correctly, this is only utilized during probing to dynamically adjust the probe offset, and as such would be better suited to piggybacking on the probe code, or as a dedicated module.

So the way I'm using this is, I set the zero the reference temp at the start of the bed mesh. The bed mesh is captured at a lower temp, 140C to 190C. Then the nozzle heats up to printing temp for the print. At that time, print time, the module is automatically compensating for the nozzle expansion. It's on for the whole print, but really its stays constant after the nozzle reaches printing temp. I did it with a macro before but that's kind of silly when [z_thermal_adjust] exists.

> On a higher level, I'm not sure if this fits in with the intended functionality of the `z_thermal_adjust` module: print time continuous adjustment of the Z coordinate space.

This is a much simper change to get merged vs building a whole new module and justifying its existence. If I wrote a new module I'm not sure how it would differ from this one, other than the name?

Maybe a good argument is multi-tool/multi-material printers have different extrusion temps per toolhead. Maybe that could be something that a new module could handle more easily. But you could do that by just setting up a different [z_thermal_adjust] section for each toolhead and enabled/disable them when a tool change happens.

alchemyEngine:
> Then the nozzle heats up to printing temp for the print. At that time, print time, the module is automatically compensating for the nozzle expansion.

Of course, that makes sense: it's simpler to compensate at print time, otherwise you'd need to pass your target print temperature in during probing. Of course, that could be done via `PRINT_START` macros, gcode commands, etc, but wouldn't be very elegant.

>  It's on for the whole print, but really its stays constant after the nozzle reaches printing temp.

You could recycle the Z height cutoff parameter to freeze the compensation after the first layer so it's not bouncing around during the print.

> Maybe a good argument is multi-tool/multi-material printers have different extrusion temps per toolhead. Maybe that could be something that a new module could handle more easily.

Also a good point.

Alright, I'm sold on a conceptual level. I'll dig into the code when I'm back at home base next week and start an initial review if @KevinOConnor is on board as well.

KevinOConnor:
> Alright, I'm sold on a conceptual level. I'll dig into the code when I'm back at home base next week and start an initial review if @KevinOConnor is on board as well.

At a very high-level, I don't see a harm in allowing multiple z_thermal_adjust config sections.  I don't know one way or another if that is useful for nozzle compensation.  So, I'm hoping you and Gareth can decide.

Cheers,
-Kevin

garethky:
I made a small update to the PR to print the component name in the console when `SET_Z_THERMAL_ADJUST` is invoked. If the config doesn't use component names there will be no change.

freakydude:
Hi,

I stumbled across this pull request and tried it out for a classic setup. Without using the referenced examples with the load cell. 

It requires a little more initial manual work, but in my initial tests it produced good results – better than without it, anyway. 

**Test system:**
Voron 2.4, in test case with a physical microswitch next to the heater bed (as suggested in the Voron standard design). 

I introduced 3 sections for extruder, heater bed, and housing temperature. Since temperatures and Z expansion influence each other and I want to avoid these measurement errors in the coefficients as much as possible, 

**I did the following:**

Three measurements, as accurate as possible without additional tools, approximated using the “paper test.” With a load cell or a reliable temperature-compensated sample, it is possible to achieve greater accuracy and further automation. (see linked disourse)

We want to determine the coefficients of the nozzle, heater bed, and housing/frame. 
**To do this all, without re-homing,**
1. set the nozzle, heater bed, and housing cold, and adjusted the offset (SET_GCODE_OFFSET Z_ADJUST=xxx MOVE=1) so that the paper creates slight friction. I noted the temperature and offset.
   1. Raised the nozzle to the highest possible temperature (e.g., 300 degrees). Adjusted the offset and noted the temperature and offset.
3. Let the nozzle cool down, housing still open, increased the temperature of the heating bed to my maximum (e.g., 120 degrees). Adjusted the offset with the paper test. Noted the nozzle, heating bed temperature, and offset. 
4. Nozzle still cold, housing closed, wait until the housing is also at maximum temperature (due to the heating bed, same temp as in step before). Determine offset again with paper. Note temperature of nozzle and chamber, as well as the offset. 

**Determine coefficients:**
1. _Nozzle Coefficient_ = `( <Step 1 Offset High Temp> - <Step 1 Offset Low Temp> ) / ( <Step 1 Nozzle High Temp> - <Step 1 Nozzle Low Temp> )`
2. 
   1. Step 2 Nozzle Offset Correction = `( <Step 2 Nozzle High Temp> - <Step 2 Nozzle Low Temp> )* <Nozzle Coefficient>`
   2. _Heater Bed Coefficient_ = ` ( <Bed Offset High Temp> - <Bed Offset Low Temp> - <Nozzle Offset Correction> ) / ( <Bed High Temp> - <Bed Low Temp> )`
3. 
   1. Step 3 Nozzle Offset Correction = `( <Step 3 Nozzle High Temp> - <Step 3 Nozzle Low Temp> ) * <Nozzle Coefficient>`
   2. Step 3 Bed Offset Correction = 0 -> Can be ignored, Temperatur unchanged
   3. _Housing Coefficient_ =  `( <Housing Offset High Temp> - <Housing Offset Low Temp> - <Nozzle Offset Correction> ) / ( <Housing High Temp> - <Housing Low Temp> )`

**In addition**, you must now call the macro `SET_Z_THERMAL_ADJUST COMPONENT=<heater_bed name> REF_TEMP=<Calibration Bed Low Temperature>` after last `G28 Z` in the homing/start script, otherwise the calculated offset for the heater bed will always be 0, since the heating bed should already have reached the target temperature during homing (this may of course vary from case to case).

**In my test i got a coeffizient of**
Nozzle =        0.000444444
Heater Bed = 0.0005866666
Housing =    -0.005566666
 
**Conclusion**

@KevinOConnor 
I suggest to merge this pull request. To stack z adjustments, calculated by positive/negative coefficients seems to be logical. I cannot see that the change has any effect on existing configurations. At least in the few tests I've done, it meant that I didn't have to readjust without a real tap/eddy tap or load cell, both with cool-printed PLA (200/60/25) and hot-printed ASA (255/120/50) – which was not the case without the z thermal adjust. 

best regards,
freakyDude

freakydude:
I must correct myself: my idea of calibrating with three paper tests in a row did not work for a medium-temperature PETG print (at least)

I don't know why. Maybe I am not able to measure or test accurately enough with my method. Or maybe there is a flaw in my reasoning. I don't know. 

There are two main uncertainties:
- I can't heat the housing without a hot heater bed.
- The bed mesh on a real print also measures something. I'm not sure how it compensates for the housing and heater bed z adjust.

Maybe I can measure it with a probe tap. Or with the new #7185 .


Anyway, the configuration and calculation did make sense.

garethky:
> I don't know why. Maybe I am not able to measure or test accurately enough with my method. Or maybe there is a flaw in my reasoning. I don't know.

The heater bed is not a single point. Its a 3 dimensional object being actively heated. It takes on a complex shape and that shape doesn't change uniformly across the surface. As we say, its a taco 🌮. And the taco curves more or less with heat.

But worse, these 6mm beds take a long time to reach thermal equilibrium. On my Voron, using a second thermistor that touches the build plate for verification, it takes around 5 minutes. That's 5 minutes after the bed heater's temperature is "at temp". Over this 5 minute window the bed continues to change shape as it absorbs heat. I've checked this with the load cell probe and its real, its still moving. Taking a mesh while the bed is moving will result in weird effects.

So I don't think its practical to try to comp the bed itself:
* It's non-linear, a linear compensation can fix the taco
* The thermistor that drives the compensation is measuring the wrong thing, it stays at the same temp while the bed is still moving around.

So get rid of the bed compensation. Instead we can just ensure that the bed is actually thermally static when we measure it. Just add a 5 minute wait in your print start for the bed to absorb heat. Then perform your bed mesh.

----

For the frame compensation we are taking a pretty strong assumption that the frame expands uniformly and it is not a taco. The underlying assumption is that no part of the frame is being actively heated and the air temp is the same everywhere. Fine. You can measure the frame expansion with the bed at a constant temp.

The problem you will run into is the metal pin that's your z reference is going to expand. The hot end is also going to expand because its subject to ambient air temps. What I did for load cell is heat the load cell to 140C and the bed to 100C. That way I know both systems are thermally stable because they are higher than ambient. (Though technically the cold side is still expanding with ambient temp.)

Frame comp really only comes into play if your print's first layer is really long. So that would be the last thing I would worry about.

PR #6856, opened 2025-03-13
Software SPI: make it respect rate limits

nefelim4ag:
Software SPI right now works as fast as it can.
That is fine, but it will be better if it respects the configured limit.

Right now, there are 2 implementations:
* For the old command
* Proposed new command in #6806

I tested with an oscilloscope over #6853
Expected | actual
-- | --
5Mhz | 4.2Mhz
4.2Mhz | 3.85Mhz
4Mhz | 3.92Mhz
3.8Mhz | 3.7Mhz
3.2Mhz | 3.125Mhz

Split from: #6806

Should fix the too-fast RP2040: https://klipper.discourse.group/t/fly-sht36-pro-rp2040-spi/22586/4

(I don't yet know how to test spi_flash.py)

### DIFF
--- a/src/spi_software.c
+++ b/src/spi_software.c
@@ -20,11 +20,10 @@ struct spi_software {
 };
 
 void
-command_spi_set_software_bus(uint32_t *args)
+command_spi_set_sw_bus(uint32_t *args)
 {
     uint8_t mode = args[4];
-    uint32_t rate = args[5];
-    uint8_t div = 0;
+    uint32_t pulse_ticks = args[5];
     if (mode > 3)
         shutdown("Invalid spi config");
 
@@ -34,14 +33,12 @@ command_spi_set_software_bus(uint32_t *args)
     ss->mosi = gpio_out_setup(args[2], 0);
     ss->sclk = gpio_out_setup(args[3], 0);
     ss->mode = mode;
-    while (((CONFIG_CLOCK_FREQ/2) >> div) > rate)
-        div++;
-    ss->sck_ticks = 1 << div;
+    ss->sck_ticks = pulse_ticks;
     spidev_set_software_bus(spi, ss);
 }
-DECL_COMMAND(command_spi_set_software_bus,
-             "spi_set_software_bus oid=%c miso_pin=%u mosi_pin=%u sclk_pin=%u"
-             " mode=%u rate=%u");
+DECL_COMMAND(command_spi_set_sw_bus,
+             "spi_set_sw_bus oid=%c miso_pin=%u mosi_pin=%u sclk_pin=%u"
+             " mode=%u pulse_ticks=%u");
 
 void
 spi_software_prepare(struct spi_software *ss)
--- a/src/spi_software.c
+++ b/src/spi_software.c
@@ -4,7 +4,9 @@
 //
 // This file may be distributed under the terms of the GNU GPLv3 license.
 
+#include "autoconf.h"   // CONFIG_*
 #include "board/gpio.h" // gpio_out_setup
+#include "board/misc.h" // timer_read_time
 #include "basecmd.h" // oid_alloc
 #include "command.h" // DECL_COMMAND
 #include "sched.h" // sched_shutdown
@@ -13,6 +15,7 @@
 struct spi_software {
     struct gpio_in miso;
     struct gpio_out mosi, sclk;
+    uint32_t sck_ticks;
     uint8_t mode;
 };
 
@@ -20,6 +23,8 @@ void
 command_spi_set_software_bus(uint32_t *args)
 {
     uint8_t mode = args[4];
+    uint32_t rate = args[5];
+    uint8_t div = 0;
     if (mode > 3)
         shutdown("Invalid spi config");
 
@@ -29,6 +34,9 @@ command_spi_set_software_bus(uint32_t *args)
     ss->mosi = gpio_out_setup(args[2], 0);
     ss->sclk = gpio_out_setup(args[3], 0);
     ss->mode = mode;
+    while (((CONFIG_CLOCK_FREQ/2) >> div) > rate)
+        div++;
+    ss->sck_ticks = 1 << div;
     spidev_set_software_bus(spi, ss);
 }
 DECL_COMMAND(command_spi_set_software_bus,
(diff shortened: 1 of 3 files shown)


bevanweiss:
Does it 'need' the new command?
Shouldn't the existing logic just respect the rate passed in, and internally derive the number of ticks itself?
i.e. in MCU, sck_ticks = mcu_freq / (2*rate);
All your MCU spi stuff would still apply...

nefelim4ag:
> Does it 'need' the new command?

You may want to check the conversation in the linked issue.

Shortly, We want to avoid division during runtime.
So, there is a patch for the existing `spi_set_software_bus`.
Which does select appropriate rate in the same way, how it is done in HW SPI.

Technically, because there is no handling of rate before, the same command can be reused with the existing `rate` argument.
We can just pass the pause time instead of the rate.

Hope that clarifies things.

bevanweiss:
> You may want to check the conversation in the linked PR https://github.com/Klipper3d/klipper/pull/6806.

This SPI change seems orthogonal to the other PR correspondance, which relates to long bus access times on stm32h7 SoCs.

> Shortly, We want to avoid division during runtime.

I'm unsure where you see this 'division during runtime'.

It would be a single one-time division when the `spi_set_software_bus` configuration method is called.
It currently takes in the desired SPI rate and does nothing with it...
It should instead work out the sck_ticks to achieve that rate.

So instead of:
```c
uint32_t pulse_ticks = args[5];
ss->sck_ticks = pulse_ticks >> 1;
```
it would be
```c
ss->sck_ticks = mcu_freq / (2*args[5] /*rate*/);
```
Having Klippy do the division in Python doesn't make sense to me.  I can see no benefit.
It just creates further inconsistency between the software SPI and hardware SPI that seems unnecessary.

For reference, HW SPI is
```c
spi_set_bus oid=%c spi_bus=%u mode=%u rate=%u
```
rate, not ticks.

Arksine:
If it is necessary to change the SW SPI command then we will need to implement an fallback mechanism in `spi_flash`, similar to what we did when the HW SPI config command changed.  If we don't then users on older firmware won't be able to use it to upgrade.  I believe something like the following will work, though I have not yet tested it:

```diff
--- a/scripts/spi_flash/spi_flash.py
+++ b/scripts/spi_flash/spi_flash.py
@@ -112,8 +112,12 @@ SPI_CFG_CMDS = (
     "config_spi oid=%d pin=%s"                      # Original
 )
 SPI_BUS_CMD = "spi_set_bus oid=%d spi_bus=%s mode=%d rate=%d"
-SW_SPI_BUS_CMD = "spi_set_software_bus oid=%d " \
-    "miso_pin=%s mosi_pin=%s sclk_pin=%s mode=%d rate=%d"
+SW_SPI_BUS_CMDS = (
+    "spi_set_sw_bus oid=%d miso_pin=%s mosi_pin=%s "        # e1f7a70
+    "sclk_pin=%s mode=%d pulse_ticks=%d",
+    "spi_set_software_bus oid=%d miso_pin=%s mosi_pin=%s "  # Original
+    "sclk_pin=%s mode=%d rate=%d",
+)
 SPI_SEND_CMD = "spi_send oid=%c data=%*s"
 SPI_XFER_CMD = "spi_transfer oid=%c data=%*s"
 SPI_XFER_RESPONSE = "spi_transfer_response oid=%c response=%*s"
@@ -1279,6 +1283,8 @@ class MCUConnection:
             'spi_bus', self.enumerations.get('bus'))
         pin_enums = self.enumerations.get('pin')
         if bus == "swspi":
+            mcu_freq = self.clocksync.mcu_freq
+            pulse_ticks = mcu_freq // SD_SPI_SPEED
             cfgpins = self.board_config['spi_pins']
             pins = [p.strip().upper() for p in cfgpins.split(',') if p.strip()]
             pin_err_msg = "Invalid Software SPI Pins: %s" % (cfgpins,)
@@ -1287,30 +1293,26 @@ class MCUConnection:
             for p in pins:
                 if p not in pin_enums:
                     raise SPIFlashError(pin_err_msg)
-            bus_cmd = SW_SPI_BUS_CMD % (SPI_OID, pins[0], pins[1], pins[2],
-                                        SPI_MODE, SD_SPI_SPEED)
+            bus_cmds = [
+                SW_SPI_BUS_CMDS[0] % (SPI_OID, pins[0], pins[1], pins[2],
+                                      SPI_MODE, pulse_ticks),
+                SW_SPI_BUS_CMDS[1] % (SPI_OID, pins[0], pins[1], pins[2],
+                                      SPI_MODE, SD_SPI_SPEED)
+            ]
         else:
             if bus not in bus_enums:
                 raise SPIFlashError("Invalid SPI Bus: %s" % (bus,))
-            bus_cmd = SPI_BUS_CMD % (SPI_OID, bus, SPI_MODE, SD_SPI_SPEED)
+            bus_cmds = [SPI_BUS_CMD % (SPI_OID, bus, SPI_MODE, SD_SPI_SPEED)]
         if cs_pin not in pin_enums:
             raise SPIFlashError("Invalid CS Pin: %s" % (cs_pin,))
-        cfg_cmds = [ALLOC_OIDS_CMD % (1,), bus_cmd]
+        cfg_cmds = [ALLOC_OIDS_CMD % (1,)]
         self._serial.send(cfg_cmds[0])
         spi_cfg_cmds = [
             SPI_CFG_CMDS[0] % (SPI_OID, cs_pin, False),
             SPI_CFG_CMDS[1] % (SPI_OID, cs_pin),
         ]
-        for cmd in spi_cfg_cmds:
-            try:
-                self._serial.send(cmd)
-            except self.proto_error:
-                if cmd == spi_cfg_cmds[-1]:
-                    raise
-            else:
-                cfg_cmds.insert(1, cmd)
-                break
-        self._serial.send(bus_cmd)
+        cfg_cmds.append(self._try_send_command(spi_cfg_cmds))
+        cfg_cmds.append(self._try_send_command(bus_cmds))
         config_crc = zlib.crc32('\n'.join(cfg_cmds).encode()) & 0xffffffff
         self._serial.send(FINALIZE_CFG_CMD % (config_crc,))
         config = self.get_mcu_config()
@@ -1326,6 +1328,16 @@ class MCUConnection:
             raise SPIFlashError(
                 "Failed to Initialize SD Card. Is it inserted?")
 
+    def _try_send_command(self, cmd_list):
+        for cmd in cmd_list:
+            try:
+                self._serial.send(cmd)
+            except self.proto_error:
+                if cmd == cmd_list[-1]:
+                    raise
+            else:
+                return cmd
+
     def _configure_mcu_sdiobus(self, printfunc=logging.info):
         bus = self.board_config['sdio_bus']
         bus_enums = self.enumerations.get(
```

KevinOConnor:
Thanks for working on this.  I agree it would be useful to add timing software spi.

I'm a little concerned with the timing optimizations here though.  Unless I'm missing something, this code does not assure a minimum pulse duration for the spi pulses.  Although it's arguably better than no timing, it seems odd to deploy a change that doesn't assert a minimum pulse time.

The timing issue occurs with something like:
```
    spi_delay(&time, t1);
    ***  an irq gets busy for 1ms ***
    gpio_out_toggle(ss->sclk);
    gpio_out_write(ss->mosi, outbuf & 0x80);
    
    spi_delay(&time, t2);   *** does not delay at all ***
    gpio_out_toggle(ss->sclk);  *** fast pulse ***
```

The only way I know of to fix this is to make sure the `timer_read_time()` used for comparison is from after the last gpio update.  (Technically one can use a timer from before the last gpio update if irqs are disabled between the `timer_read_time()` and gpio update, and if `CONFIG_HAVE_STRICT_TIMING` is set.  Not sure if it is worth the effort to do that though.)

-Kevin

src/spi_software.c
```diff
@@ -29,31 +34,52 @@ command_spi_set_software_bus(uint32_t *args)
...
+        cur = timer_read_time();
+    *time = cur;
+}
+
 void
 spi_software_transfer(struct spi_software *ss, uint8_t receive_data
                       , uint8_t len, uint8_t *data)
 {
+    uint32_t t1 = ss->sck_ticks >> 1;
+    uint32_t t2 = t1 + (ss->sck_ticks & 1);
```
KevinOConnor, line 70: Super minor, but I think this may be a little more readable as: `uint32_t t2 = ss->sck_ticks - t1;`

KevinOConnor (commented):
Some minor comments below.

-Kevin

nefelim4ag:
Good catch, I missed it.

> Technically one can use a timer from before the last gpio update if irqs are disabled

I initially want to use something like that, but it feels a little like a patchwork.
So, I tried it this time, slightly different. Does it look correct?

Btw, maybe, it makes sense to introduce gpio_out_write_noirq() for context like this one?

Thanks.

nefelim4ag:
With IRQ disable/enable, which probably kills compiler optimizations, with a 4 MHz target, there is 3.5 MHz on the line, and it refuses to go much faster.
With a simple busy loop, there is 2.5Mhz.

This is my main worry here, because it will kill the performance a little.
But with a busy loop, it is still possible to just set some insane number, like 30MHz, to effectively disable the pause.

What did you think?

KevinOConnor:
> What do you think?

I think it is not worth optimizing software spi.  So, I'd try to keep it simple (like you have in the latest version on this PR).

This change is something I'd look at merging after tagging v0.13.0 .

Minor, but `sck_ticks` should be a uint32_t.  Minor, but the way you have `spi_delay()` written will make it harder for the compiler to optimize - you might want to simplify to something like `spi_delay(time + t2);` .

I'm not sure I'd start with `uint32_t time = timer_read_time() - t1;` as many chips require a delay between CS going low and the first SCLK edge.  Enforcing a small delay is probably a good thing.

Cheers,
-Kevin

nefelim4ag:
Done,
Also, I swapped `if` and `for` in code, this swap produces slightly faster & slightly more compact code.

`37628 -> 37548 out/klipper.bin`

KevinOConnor:
Thanks.  Again, seems fine to me, but should wait until after v0.13.0 .

Cheers,
-Kevin

KevinOConnor:
Thanks.

-Kevin

PR #6857, opened 2025-03-14
stm32 CAN: Added PH13/14 CAN pin option for stm32h743

FrYakaTKoP:
Added the option to select PH13/PH14 as CAN pins for stm32h743 mcus

Signed-off-by: Christoph Frei <fryakatkop@gmail.com>


JNP-1:
I have tested it with a customer board and it works.

KevinOConnor:
Thanks.

-Kevin

PR #6860, opened 2025-03-15
Allow homing only some of the axes in homing_override

dmbutyugin:
The PR actually makes two independent, but related changes:

- allows homing only some of the axes in `homing_override` rather than all axes
- changes SET_KINEMATIC_POSITION to change homing state only for the specified axes

The background for both changes is that as a part of #6815 I'm trying to adapt the proposal from the discourse - namely explicit homing of IDEX axes, which requires a user to define `[homing_override]`. However, I find the currently documented requirements for it rather inflexible, as they require a user to home all axes. This may considerably degrade user experience vs the current `cartesian` kinematics and alike, since they allow partial homing of the printer axes. The corresponding change in this PR addresses this shortcoming by explicitly permitting partial homing of the axes and instead adds a validation that such homing happened indeed. FWIW, this particular change should be fully backwards-compatible, as any previously valid `[homing_override]` would still be valid with additional validations added.

As for the second change to `SET_KINEMATIC_POSITION`, I find it to be a bit of an overlook as it makes all axes homed even those that were not explicitly modified by the command. The proposed change addresses it by only updating the homed state of the axes that were explicitly mentioned in the parameters of the command. This one is not fully backwards-compatible, but since the previous behavior was in fact rather surprising, and the tool itself was positioned as mostly for debugging purposes, I consider such a change acceptable, but I anyways documented it in Config_Changes.md.

Please let me know what you think. I think this would be particularly useful for the upcoming `generic_cartesian` kinematics and potential deprecation of other IDEX configurations, as this would allow users to maintain the same level of convenience as before.

### DIFF
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,10 @@ All dates in this document are approximate.
 
 ## Changes
 
+20250315: `SET_KINEMATIC_POSITION` now forces the toolhead position
+only for the explicitly specified axes. Other axes will not assume
+homed state by default.
+
 20250308: The `AUTO` parameter of the
 `AXIS_TWIST_COMPENSATION_CALIBRATE` command has been removed.
 
--- a/docs/Config_Changes.md
+++ b/docs/Config_Changes.md
@@ -8,6 +8,8 @@ All dates in this document are approximate.
 
 ## Changes
 
+20250315: `[homing_override]` is permitted to home only some axes.
+
 20250315: `SET_KINEMATIC_POSITION` now forces the toolhead position
 only for the explicitly specified axes. Other axes will not assume
 homed state by default.
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1319,14 +1319,21 @@ gcode:
 #   found in the normal g-code input. See docs/Command_Templates.md
 #   for G-Code format. If a G28 is contained in this list of commands
 #   then it will invoke the normal homing procedure for the printer.
-#   The commands listed here must home all axes. This parameter must
-#   be provided.
+#   For simplicity, it is recommended to always home all printer axes.
+#   However, in order to speed up the homing process, it is possible
+#   to write the gcode that homes only some axes. The commands listed
+#   here must home at least all requested axes, which are passed from
+#   the original G28 invocation via the params pseudo-variable, as well
+#   as all axes, for which the position is forced via set_position_?
+#   parameters (see below). If no specific axis is requested to home, the
+#   commands here must home all axes. This parameter must be provided.
 #axes: xyz
 #   The axes to override. For example, if this is set to "z" then the
 #   override script will only be run when the z axis is homed (eg, via
 #   a "G28" or "G28 Z0" command). Note, the override script should
-#   still home all axes. The default is "xyz" which causes the
-#   override script to be run in place of all G28 commands.
+#   still home all required axes as described above. The default is
+#   "xyz" which causes the override script to be run in place of all
+#   G28 commands.
 #set_position_x:
 #set_position_y:
 #set_position_z:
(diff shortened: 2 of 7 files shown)


KevinOConnor:
Thanks.  A couple of comments:

* I don't think we should change the behavior of `SET_KINEMATIC_POSITION`.  We've long documented that a bare `SET_KINEMATIC_POSITION` would turn off the homing checks.  If there is a confusion with this command I think we would be better off updating the documentation to clarify how it works.  I fear if we change its behavior we'll break a lot of existing macros.  Alternatively, we could add a new explicit `AXES_TO_SET_AS_HOMED=xyz` type parameter to SET_KINEMATIC_POSITION, or we could introduce an entirely new command and deprecate the old command.
* I'm not entirely sure I understand the changes to homing_override.  Could you give an example of how an idex user would utilize homing_override?  Given the discussion we've had on Discourse I think we'd want to have a long-term plan prior to altering the behavior of homing_override (so that we're not asking users to make multiple config revisions).

Thanks again,
-Kevin

dmbutyugin:
Hi Kevin,

First, [homing_override] changes could be useful for generic_cartesian kinematics if we were to implement homing only the active carriage, something along the lines of:
```
[homing_override]
axes: x
gcode:
    # Whenever this homing_override is called, X axis must be homed
    SET_DUAL_CARRIAGE CARRIAGE=u
    G28 X
    SET_DUAL_CARRIAGE CARRIAGE=x
    G28 {rawparams}
```
which is currently explicitly not allowed by the documentation. However, as I [wrote](https://github.com/Klipper3d/klipper/pull/6815#issuecomment-2767830729) in the PR for generic_cartesian kinematics, given the direction of discussion for homing revamp, it seems it is going to be much more significant, and `[homing_override]` may be deprecated. As such, it might be better to leave it be and not make any modifications to it or `generic_cartesian` kinematics, and just commit the kinematics more or less as-is.

As for `SET_KINEMATIC_POSITION`, I find its behavior surprising and kind of dangerous - in a sense that if I issue `SET_KINEMATIC_POSITION Z=...` command and then accidentally issue G0 X... command, or accidentally press a button for X/Y axes in the web UI, I can damage a printer. While this _is documented_, and `SET_KINEMATIC_POSITION` does require attention and care, it is still unfortunate that it exposes dangerous behavior unnecessarily. And it seems that other users can also find this behavior surprising, e.g. [thread](https://klipper.discourse.group/t/set-kinematic-position-behaviour/22246). Theoretically, a similar behavior can be achieved by memorizing homed axes before `SET_KINEMATIC_POSITION` and then issuing another command with appropriately set `CLEAR=...` parameter, but this is only useful for the macros and is still bothersome. And while my change could theoretically break some macros, I do believe it may break only rare cases (I suspect mostly the cases of delta kinematics and such where issuing `SET_KINEMATIC_POSITION Z=10` would not be sufficient to enable Z moves, and one would have to issue a more complete command `SET_KINEMATIC_POSITION X=0 Y=0 Z=10`). And any breakages would be mostly harmless, meaning that this change would not enable users to harm their printers - instead some macros and scripts that were previously working would require an explicit fix (by adding missing axes), and thus a change in Config_Changes.md would suffice. But this is ultimately your call.

KevinOConnor:
Just to be clear, `SET_KINEMATIC_POSITION Z=10` is not similar to `SET_KINEMATIC_POSITION X=0 Y=0 Z=10`.  The former sets all axes as homed, sets the Z position to 10, and does not change the X and Y position.  The latter would alter the position of X and Y.

I've long advised users that a bare `SET_KINEMATIC_POSITION` command would disable homing and not alter the current position of the toolhead.  This used to be a common request (as OctoPrint tended to issue unexpected `M84` commands).  I'm fairly confident that changing this behavior will cause grief.

I think we can fix this by adding documentation and adding the ability to specify which axes are considered homed - see #6883 .

Cheers,
-Kevin

PR #6865, opened 2025-03-21
buttons: fixes incorrect parameters

pedrolamas:
Fixes incorrect code that causes potential startup crash caused by changes in #6848


KevinOConnor:
Thanks.  @garethky - fyi.

-Kevin

garethky:
ooof, Python will let you do anything!

PR #6866, opened 2025-03-21
usbcan: support `ip link set can0 type can bitrate $bitrate`

ayufan:
Makes Klipper to be a quite functional usb-to-can
with externally controllable bitrate.

Quite handy, as the bitrate is now configured by the system.

This was/is tested, and running for a long time on:

- Was tested on various rp2040 boards
- Was tested on various stm32 boards (BTT Octopus 1.1, but also Cheetah v2 and v3)
- It was not tested on fdcan variants, and atsam*

It change might create some work for some users: people might have different can frequency configured in FW vs one for network interface on a linux system.


ayufan:
Thanks @KevinOConnor I will make those changes and get back to you. It will take me a few weeks due to upcoming travel.

ayufan:
> Also, if the frequency is set from Linux, I think "make menuconfig" should not prompt for a frequency. (As asking for a frequency that is not actually used would be confusing.) Similarly, the documentation would need to be updated to reflect the new support.

I see two cases:

- can device: we need to set frequency up-front
- usbcan device: we need some default, as setting can frequency is likely not a requirement (will check this)

EmperorArthur:
Awesome work.  Having the bridge ignore the set frequency was really confusing to me at first.  Especially since my printer\* didn't even have the MCU in the out of the box config.

This will also make verifying if a certain speed works much easier.

\* Magneto X

PR #6867, opened 2025-03-22
RP2040: Another HW SPI bus

nefelim4ag:
Added HW bus from another [Mellow/Fly board](https://klipper.discourse.group/t/fly-sht36-pro-rp2040-spi/22586) (Discourse link).

### DIFF
--- a/src/rp2040/spi.c
+++ b/src/rp2040/spi.c
@@ -29,6 +29,8 @@ DECL_ENUMERATION("spi_bus", "spi1_gpio12_gpio15_gpio14", 6);
 DECL_CONSTANT_STR("BUS_PINS_spi1_gpio12_gpio15_gpio14", "gpio12,gpio15,gpio14");
 DECL_ENUMERATION("spi_bus", "spi1_gpio24_gpio27_gpio26", 7);
 DECL_CONSTANT_STR("BUS_PINS_spi1_gpio24_gpio27_gpio26", "gpio24,gpio27,gpio26");
+DECL_ENUMERATION("spi_bus", "spi1_gpio12_gpio11_gpio10", 8);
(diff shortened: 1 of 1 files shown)


src/rp2040/spi.c
```diff
@@ -11,24 +11,21 @@
...
-DECL_ENUMERATION("spi_bus", "spi1_gpio12_gpio15_gpio14", 6);
-DECL_CONSTANT_STR("BUS_PINS_spi1_gpio12_gpio15_gpio14", "gpio12,gpio15,gpio14");
-DECL_ENUMERATION("spi_bus", "spi1_gpio24_gpio27_gpio26", 7);
-DECL_CONSTANT_STR("BUS_PINS_spi1_gpio24_gpio27_gpio26", "gpio24,gpio27,gpio26");
+#define DECL_SPI_BUS(bus, miso, mosi, sck, order)                          \
+    DECL_ENUMERATION("spi_bus", #bus "_" #miso "_" #mosi "_" #sck, order); \
+    DECL_CONSTANT_STR("BUS_PINS_" #bus "_" #miso "_" #mosi "_" #sck,       \
+                      #miso "," #mosi "," #sck);
+
+DECL_SPI_BUS(spi0, gpio0,  gpio3,  gpio2,  0);
```
Sineos, line 19: Doesn't this somehow obfuscate how the bus is named in the config? Previously, it was clear that it would be

```
spi_bus: spi0_gpio0_gpio3_gpio2
```

nefelim4ag: That's a really good point,
I personally expect people to look at logs and/or `out/klippy.dict`, where those definitions are exported.

But I think you are right.

KevinOConnor:
Thanks.

-Kevin

PR #6868, opened 2025-03-24
Icm20948: Fix accelerometer readings, some changes

MRX8024:
-The accelerometer remained (by default) in accels scale 2g mode with a sampling rate of +-1125/s instead of setting the 8g and 4500 rate, since we did not switch to registers bank 2 before the 'accel config' configuration. Because of which it 'choked', and also incorrectly calculated its accels in ffreader.

I did some comparative experiments between adxl345 and Icm20948 (2g mode):
<div style="display: flex; gap: 10px;">
    
    
</div>

Measured the acceleration at idle on Icm20948:

Сomparative experiments between adxl345 and Icm20948 (8g mode) after correcting the accelerometer configuration:
<div style="display: flex; gap: 10px;">
    
    
</div>
Measured the acceleration at idle on Icm20948:

The results have returned to normal, but it still seems like the accelerometer is reached the maximum acceleration threshold - the peak is too thick.

That's why I also compared adxl345 and Icm20948 in 16g mode:
<div style="display: flex; gap: 10px;">
    
    
</div>

Now it displays identically to adxl345. 

Measured the acceleration at idle on Icm20948:

-My suggestion is to switch from 8g mode to 16g, this change is also attached in a separate commit.
-Changes for adxl345 debug functions of read/write registers, the limit has been raised from 126 to 127 addresses, since Icm20948 has a maximum address of 127.
-Add icm20948 description to config reference.

I also have a question, i2c software bus is blocked at 100k rate, under such condition the samples rate of accelerometer will drop to about ~1800smp\s, and the calculation of accelerations will also be distorted. 
Perhaps we should add the functionality of the minimum bus rate for Icm20948 (and mpu chips too, by the way). Or perhaps we should perform calculations in ffreader differently? Or at least warn the user if his i2c bus has a low speed.

### DIFF
--- a/docs/Config_Reference.md
+++ b/docs/Config_Reference.md
@@ -1669,6 +1669,25 @@ cs_pin:
 #   measurements.
 ```
 
+### [icm20948]
+
+Support for icm20948 accelerometers.
+
+```
+[icm20948]
+#i2c_address:
+#   Default is 104 (0x68). If AD0 is high, it would be 0x69 instead.
+#i2c_mcu:
+#i2c_bus:
+#i2c_software_scl_pin:
+#i2c_software_sda_pin:
+#i2c_speed: 400000
+#   See the "common I2C settings" section for a description of the
+#   above parameters. The default "i2c_speed" is 400000.
+#axes_map: x, y, z
+#   See the "adxl345" section for information on this parameter.
+```
+
 ### [lis2dw]
 
 Support for LIS2DW accelerometers.
--- a/klippy/extras/adxl345.py
+++ b/klippy/extras/adxl345.py
@@ -166,12 +166,12 @@ class AccelCommandHelper:
                           % (accel_x, accel_y, accel_z))
     cmd_ACCELEROMETER_DEBUG_READ_help = "Query register (for debugging)"
     def cmd_ACCELEROMETER_DEBUG_READ(self, gcmd):
-        reg = gcmd.get("REG", minval=0, maxval=126, parser=lambda x: int(x, 0))
+        reg = gcmd.get("REG", minval=0, maxval=127, parser=lambda x: int(x, 0))
         val = self.chip.read_reg(reg)
         gcmd.respond_info("Accelerometer REG[0x%x] = 0x%x" % (reg, val))
     cmd_ACCELEROMETER_DEBUG_WRITE_help = "Set register (for debugging)"
     def cmd_ACCELEROMETER_DEBUG_WRITE(self, gcmd):
-        reg = gcmd.get("REG", minval=0, maxval=126, parser=lambda x: int(x, 0))
+        reg = gcmd.get("REG", minval=0, maxval=127, parser=lambda x: int(x, 0))
         val = gcmd.get("VAL", minval=0, maxval=255, parser=lambda x: int(x, 0))
         self.chip.set_reg(reg, val)
 
--- a/klippy/extras/icm20948.py
+++ b/klippy/extras/icm20948.py
@@ -12,42 +12,40 @@
 import logging
 from . import bus, adxl345, bulk_sensor
 
-ICM20948_ADDR =      0x68
+ICM20948_ADDR =         0x68
 
 ICM_DEV_IDS = {
     0xEA: "icm-20948",
     #everything above are normal ICM IDs
-    }
-
+}
 
 # ICM20948 registers
-REG_DEVID =         0x00 # 0xEA
-REG_FIFO_EN =       0x67 # FIFO_EN_2
-REG_ACCEL_SMPLRT_DIV1 =    0x10 # MSB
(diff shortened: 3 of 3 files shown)


nefelim4ag (approved):
Not sure about format changes, otherwise LGTM

Thanks.

---
> I also have a question, i2c software bus is blocked at 100k rate, under such condition the samples rate of the accelerometer will drop to about ~1800smp\s, and the calculation of accelerations will also be distorted.

Technically, it is possible to allow SW I2C change rate.
Technically, it is possible to increase SMPLRT_DIV to 1, and allow set accel frequency to 1125Hz, which should fit in 100k.

Not sure if it is really needed. Generally, SW I2C is duct tape, and generally it is expected to run such sensors with 400kHz.
_I personally would like to have SW buses with arbitrary speed definition, for development proposes or specific workarounds._

KevinOConnor:
@phansel - fyi.

-Kevin

phansel:
@MRX8024 nice catch on REG_BANK_SEL! The first ICM20948 plot you shared does look unusual. Perhaps the lower peak resonance on my printer [(around 50-80 Hz)](https://github.com/Klipper3d/klipper/pull/6756#issuecomment-2696384003) camouflaged this issue where it became apparent around 150 on yours.

Could you share which microcontroller you're using? 

Thanks for the heads-up @KevinOConnor.

MRX8024:
> Could you share which microcontroller you're using?

I use octopus pro on stm32h723 with i2c1_PB8_PB9

dmbutyugin:
Interesting observation that 8g appears to be not enough for the resonance measurements. @MRX8024  If you could measure the resonance test where you observed the differences with `TEST_RESONANCES OUTPUT=raw_data` (in 16g mode)? It would be really interesting to see the raw measurements to confirm 8g saturation. For example, for some sensors we do use 8g mode (MPU9250 and LIS3DH for instance).

And separately, regarding i2c, I believe for MPU9250 we are able to achieve 4k samples/sec data rate over i2c bus. Or is the limitation you are reporting only for software I2C?

MRX8024:
> to see the raw measurements

Okay, I'll take measurements in both modes tomorrow.

> And separately, regarding i2c, I believe for MPU9250 we are able to achieve 4k samples/sec data rate over i2c bus. Or is the limitation you are reporting only for software I2C?

Yes, sw bus has a limit of 100k (even if we specify 400k in the configuration) through which about 1800smp/s passes. And the user is allowed to configure sw without any warnings.

MRX8024:
Well, @dmbutyugin, 8g and 16g raw icm20948 measurements:
[raw_data_x_icm20948_8g_aph75.zip](https://github.com/user-attachments/files/19468137/raw_data_x_icm20948_8g_aph75.zip)
[raw_data_x_icm20948_16g_aph75.zip](https://github.com/user-attachments/files/19468141/raw_data_x_icm20948_16g_aph75.zip)

I installed mpu6500 in the same place on the same printer, so 8g and 16g raw mpu6500 measurements:
[raw_data_x_mpu9250_8g_aph75.zip](https://github.com/user-attachments/files/19471759/raw_data_x_mpu9250_8g_aph75.zip)
[raw_data_x_mpu9250_16g_aph75.zip](https://github.com/user-attachments/files/19471761/raw_data_x_mpu9250_16g_aph75.zip)

mpu6500 16g shows very strange measurements. I set `SET_ACCEL_CONFIG = 0x18` and recalculated `SCALE`, maybe I'm missing something in the configuration. Also on 8g there are high noises around 0-20hz, I have seen the same thing before on lis2dw, which I have never seen on adxl345. Maybe I have a faulty mpu6500, I'll try to install another proven one.

Let me know if i need to test anything else.

KevinOConnor:
Okay, thanks.  If there are no further comments I'll look to commit in a few days.

-Kevin

dmbutyugin:
@MRX8024 

Thanks for the data! It seems that, indeed, for your printer 8g is not enough, and 8g range gets clipped:
raw_data_x_icm20948_cmp
though fortunately the clipping is not affecting the readings dramatically:
raw_data_x_icm20948_cmp_clipping
and so it affects mostly the amplitude of the detected resonances in this case:
resonances_x_cmp
(but it would have affected the estimation of damping_ratio, if one were to try that).

So, I think this is a good argument to use 16g scale as much as possible for the accelerometers.

Then, MPU data is actually borked, and in fact, both 8g and 16 ranges:
raw_data_x_mpu9250_cmp
and the enlarged segment where the data starts to be borked:
raw_data_x_mpu9250_cmp_borked
I suspect this is either a defective chip, or a problem with I2C bus data transmission stability, since during the period of issues the readings just change randomly with the magnitude of the corresponding accelerometer scale.

KevinOConnor:
Thanks.

-Kevin

PR #6869, opened 2025-03-24
ICM20948: handle i2c errors

nefelim4ag:
Another part of I2C error handling.
I'm not sure it is worth trying to recover from partial reads, so there is still a shutdown.

Thanks.


phansel:
@nefelim4ag do you have ICM20948 in hand or would you like me to test these changes?

nefelim4ag:
@phansel,

> do you have ICM20948 in hand or would you like me to test these changes?

Unfortunately, I do not have one, so it would be really nice of you to test it.

Thank you.

_Btw, I suspect overflow counting will not work as expected. The rate of query_status is far lower than the rate to pull data from the FIFO. So, it will not be able to actually count possible overflows.

Not sure how severe it actually is, to count it properly._

MRX8024:
I ran tests with the new patch. 
Debug messages is output once per second, summed up over intermediate batches.

100k bus:


400k bus:


Also for your information, with both old and new code, sometimes there is a static number of overflows:


Either the first data has overflows:

src/sensor_icm20948.c
```diff
@@ -23,8 +23,8 @@
 
 #define FIFO_OVERFLOW_INT 0x0F // from datasheet
 
-#define BYTES_PER_FIFO_ENTRY 6
-#define BYTES_PER_BLOCK 48
+#define BYTES_PER_SAMPLE 6
+#define BYTES_PER_BLOCK (BYTES_PER_SAMPLE * 2)
```
KevinOConnor, line 27: I'm not an expert on the icm20948, but this looks odd.  What is the reason for decreasing the data read size?

src/sensor_icm20948.c
```diff
@@ -103,12 +105,30 @@ ic20948_query(struct icm20948 *ic, uint8_t oid)
...
+            if (i == BYTES_PER_BLOCK)
+                goto out;
+            // Unaligned read
+            i2c_shutdown_on_err(r);
+        }
+        ic->sb.data_count += BYTES_PER_BLOCK;
         ic->fifo_pkts_bytes -= BYTES_PER_BLOCK;
-        sensor_bulk_report(&ic->sb, oid);
+        if (ic->sb.data_count + BYTES_PER_BLOCK > ARRAY_SIZE(ic->sb.data))
+            sensor_bulk_report(&ic->sb, oid);
```
KevinOConnor, line 128: This looks odd to me.  I think that if there is a transfer error we should discard the entire transfer and increase the overflow counter (`ic->sb.possible_overflows++;`).  I don't think the code should attempt to use a partial transfer - that seems error prone.

KevinOConnor (commented):
Thanks for working on this.  I have some minor comments.

-Kevin

src/sensor_icm20948.c
nefelim4ag: This is my attempt to decrease the probability of a partial read, fewer bytes per attempt -> less probability, I assume.
I'm concerned that if we do not fail the system, then it is possible to read less than 6 bytes of data.
So, all the next bytes will probably be shifted by that amount, so all data readed after that will be borked.
Miss 1 byte - borked.
Miss 2 bytes - axis swapped.

src/sensor_icm20948.c
nefelim4ag: This code does almost that. I'm not sure it is even possible to handle partial reads.
So, I just ensure there is no read happened at all, and reading of those bytes will be continued in the next cycle.

If partial reads just happened, it still triggers the `i2c_shutdown_on_err(r);` here.

Btw, possible_overflows seems not to be used by the code above, also it is a little optimistic for ICM and MPU.
Because overflow can happen more frequently than 10 times per second.
But looks like it is only checked below, by the host query (once per 0.1s).

src/sensor_icm20948.c
KevinOConnor: FWIW, if there are i2c errors, the data is borked - nothing we can do about that other than tell users they need to have clean i2c lines to meaningfully use the accelerometer.  So, reporting warnings/errors/overflows is better than a shutdown, but I don't think we can reasonably get "good" data from the accelerometer if there i2c errors.

I don't think smaller transfers reduces the rate of partial transfers as there would be many more i2c transfers that could fail.  (Unless the icm20948 does something weird that I'm not aware of.)

Cheers,
-Kevin

PR #6870, opened 2025-03-24
tweak: be tolerant of missing newlines at EOF

Laikulo:
This adds a final pass through `virtual_sdcard`'s line pumper to avoid losing the last block of a file if it is unterminated.


pedrolamas:
Unless I'm mistaken, this seems to be the same as #6565, you might want to take a look to understand the reasoning behind the current behaviour.

PR #6871, opened 2025-03-25
PR: Load Cell Probe

garethky:
The goal of this PR is to get from a gram scale all the way to being able to home a printer with a load cell.

An overview of whats in here:
* The SOS Filter is created as a new utility component on the MCU.
* The `load_cell_probe` is introduced on the MCU - this allows load cell sensors to act as data sources for an endstop that lives on the MCU. It can detect when a sudden force on the load cell indicates a collision. The SOS filter is used to do this.
* The Load Cell sensors are modified so they can send data to a `load_cell_probe` on the MCU.
* Finally the LoadCellProbe is introduced on the host that can combine a load cell sensor the `load_cell_probe` together to turn it into a `probe`.

---
## SOS Filter

This is a general purpose fixed point SOS filter implementation. It supports a maximum of 4 filter sections. 

The primary purpose it to reject drift and noise while homing. There is a post about this [here](https://klipper.discourse.group/t/load-cell-probing-algorithm-testing/9751/77?u=garethky).
This idea was originally introduced by Prusa Research code [here](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/47eb31be3dbec677cbfbf07c470e7fd7014093b6/src/common/loadcell.hpp#L166). Their implementation uses an old code generator, `mkfilter`, and can't be easily changed without re-compilation.

This implementation uses Second Order Sections and uploads each section to the MCU to initialize the filter. This has a number of advantages for klipper:
* This implementation uses fixed point math to get around limitations in klipper's communications protocol and potential issues using the FPU's on chip.
* The SOS algorithm avoids using a costly division instruction in the filter loop, saving hundreds of clock cycles.
* The only requirement for fast execution is single cycle [SMULL](https://developer.arm.com/documentation/dui0802/b/A32-and-T32-Instructions/UMULL--UMLAL--SMULL--and-SMLAL) which is a feature of every ARM Cortex M0+ core.
---

## LoadCellProbe

Most of the code in this PR is about `load_cell_probe` MCU implementation. This is not the end of the code for the `load_cell_probe.py`. There are another ~800 lines of code that will go into this file to do the post homing Tap Analysis. I see value in keeping these chunks of functionality separate and combining them with composition. People have reached out to me to do a Filament Scale that would have specific functionality for weighing filament, calculating filament used in a print and so on. Keeping `load_cell` relatively small and simple will let that development happen independently.

There is some copy/paste stuff in this code that I think is worth talking about. `LoadCellEndstop` combines the `ProbeEndstopWrapper` and `MCU_endstop` types. I want to re-use the code from `ProbeEndstopWrapper` in `probe.py`. My suggested refactor of `ProbeEndstopWrapper` is to add an optional constructor parameter that injects in the `mcu_endstop` instance and overrides its default behavior of creating one from a pin. Most of what `ProbeEndstopWrapper` does is really about multi-probe and it seems a real shame to copy/paste that. If that refactor sounds cool, I'll make that change.


garethky:
Re this diff: https://github.com/Klipper3d/klipper/compare/9756dcf8e4f897698b74ba9f93c29075923af701..09dfbd9ed05e56bd00b2f788d585ecff3095bbfe

There has been a very difficult to replicate bug where I got the "Timer Too Close" error. I couldn't get anyone else to report a replication. I think this change solves it.

The contract for `toolhead.get_last_move_time()` isn't well documented. If the last move has ended it actually returns a time in the future when you might expect to be able to begin executing the next move. That could be upwards of 350ms depending on the state inside toolhead. The 100ms padding I was adding was OK 99.99% of the time but very occasionally it would not be enough.

At least I think that's what the root cause is. Its so hard to replicate this organically. It might be a 1 in 1000 event.

garethky:
I had the safety range error happen to me today and it wasn't a good user experience. Specifically the error message doesn't say why the `trsync` was triggered with an error. There are 3 error conditions that could have triggered it:
* The safety range could have been violated (this was my issue)
* The watchdog timer could be violated
* There could have been a fixed point math overflow

So I'm going to do a patch that increments the error code by 1 for each of these cases and emits a specific error message on the Host side.

KevinOConnor:
Thanks for working on this.  I agree it would be good to merge this support into Klipper.  I do have some comments on the changes.

I like how you've split out the SOSFilter into its own standalone module.  I also like how you've split up this PR into separate commits.  I'm going to review each commit separately.  These responses will follow below.

Cheers,
-Kevin

src/sos_filter.c
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    // check for last section
+    if (sf->last_section_idx == sf->n_sections - 1) {
+        sf->is_filter_ready = 1;
+    }
+}
+DECL_COMMAND(command_config_filter_section
+    , "config_sos_filter_section oid=%c n_sections=%c"
+    " section_idx=%c sos0=%i sos1=%i sos2=%i sos3=%i sos4=%i");
```
KevinOConnor, line 149: By convention, any mcu command starting with "config_" will create an oid.  So, best to rename this command.

KevinOConnor (commented):
As mentioned, I like how you've split up the sos_filter into its own standalone module.

The main comment I have is that I think it would be nice if this was a little more generic, as I'd like to also use the functionality from the ldc1612 code as well.  There were a few areas I thought that could be made more generic:
* It would be nice if the shift right amount in `fixed_mul()` (ie, the precision) was host configurable.
* It would be nice if the number of sections was host configurable (instead of fixed at 4).
* It would be nice if only the host invoked a reset of the state (not other mcu modules).  As only the host really knows when it is appropriate to reset the state.
* It would be nice if `sos_filter.h` only defined `sosfilt()` and `sos_filter_oid_lookup()` as this is all other mcu modules should need.  (loadcell needs more, but loadcell can define the helpers/definitions it needs directly.)

These aren't blockers, as we could refactor after merging.  However, it is kind of a pain to reflash, so may be nice to make these changes prior to merging.

I will mockup an implementation of sos_filter.c to demonstrate what I mean.

src/load_cell_endstop.c
```diff
@@ -0,0 +1,297 @@
+// Load Cell based end stops.
+//
+// Copyright (C) 2023  Gareth Farrington <gareth@waves.ky>
```
KevinOConnor, line 3: Odd that the copyright is dated 2023 (same comment about the load_cell_probe.py copyright).

src/load_cell_endstop.c
```diff
@@ -0,0 +1,297 @@
...
+set_flag(uint8_t mask, struct load_cell_endstop *lce)
+{
+    lce->flags |= mask;
+}
+
+static inline void
+clear_flag(uint8_t mask, struct load_cell_endstop *lce)
+{
+    lce->flags &= ~mask;
+}
```
KevinOConnor, line 80: Not a blocker, but for what it is worth, I find these wrappers make the code harder to understand.

src/load_cell_endstop.c
```diff
@@ -0,0 +1,297 @@
...
+    else {
+        // clear "live" trigger view for QUERY_ENDSTOPS
+        clear_flag(FLAG_IS_TRIGGERED, lce);
+    }
+}
+
+// Timer callback that monitors for timeouts
+static uint_fast8_t
+watchdog_event(struct timer *t)
+{
```
KevinOConnor, line 168: I agree that we need to check for sensor failures.  However, I think that support should ultimately be a part of the trsync.c support (and not specific to loadcell).  We should also have this support in ldc1612 and likely other sensors in the future.  So, best to work out how we can do that generically.

I'll mockup some changes to trsync.c to demonstrate what I have in mind.

src/load_cell_endstop.c
```diff
@@ -0,0 +1,297 @@
...
+    if (!is_homing || is_homing_trigger) {
+        return SF_DONE;
+    }
+
+    irq_disable();
+    if (lce->watchdog_count > lce->watchdog_max) {
+        try_trigger_error(lce);
+    }
+    lce->watchdog_count += 1;
+    irq_enable();
```
KevinOConnor, line 183: Eeks!  Interrupts are always disabled in a timer, and if you call `irq_enable()` from a timer it can destabilize the system.

src/load_cell_endstop.c
```diff
@@ -0,0 +1,297 @@
...
+    lce->trigger_reason = args[2];
+    lce->error_reason = args[3];
+    lce->time.waketime = args[4];
+    lce->rest_ticks = args[5];
+    lce->watchdog_max = args[6];
+    lce->watchdog_count = 0;
+    reset_filter_state(lce->sf);
+    lce->time.func = watchdog_event;
+    sched_add_timer(&lce->time);
+    set_flag(FLAG_IS_HOMING, lce);
```
KevinOConnor, line 277: It's very strange to set the flag after enabling the timer.  It's enabling a race condition between task and interrupts.

garethky:
I'm game for all those changes. I had thought about all of them (at one time or another) but didn't want to dump the effort into something that configurable until you had generally signed off on the idea.

I can make the changes as refactor commits and then squash them after we get to a good end state

src/load_cell_endstop.c
```diff
@@ -0,0 +1,297 @@
...
+static inline int64_t
+round_fixedQ48(const fixedQ48_t fixed_value) {
+    return fixed_value >> FIXEDQ48_FRAC_BITS;
+}
+
+// Convert sensor counts to grams
+static inline fixedQ48_t
+counts_to_grams(struct load_cell_endstop *lce, const int32_t counts) {
+    // tearing ensures readings are referenced to 0.0g
+    const int32_t delta = counts - lce->tare_counts;
```
KevinOConnor, line 53: Out of curiosity, why does the mcu "tap" detection code tare the value?  I thought the tap detection was relative and did not rely on absolute sensor values?

src/load_cell_endstop.c
```diff
@@ -0,0 +1,297 @@
...
+    struct load_cell_endstop *lce = load_cell_endstop_oid_lookup(args[0]);
+    sendf("load_cell_endstop_state oid=%c homing=%c homing_triggered=%c"
+        " is_triggered=%c trigger_ticks=%u sample=%i sample_ticks=%u"
+            , oid, is_flag_set(FLAG_IS_HOMING, lce)
+            , is_flag_set(FLAG_IS_HOMING_TRIGGER, lce)
+            , is_flag_set(FLAG_IS_TRIGGERED, lce), lce->trigger_ticks
+            , lce->last_sample, lce->last_sample_ticks);
+}
+DECL_COMMAND(command_load_cell_endstop_query_state
+                , "load_cell_endstop_query_state oid=%c");
```
KevinOConnor, line 297: I don't think it is a good idea for the mcu loadcell code to try to emulate QUERY_ENDSTOP.  I'd be concerned that it would be an ongoing maintenance burden - as testing and reflashing the mcu is a pain.

I can see the value in providing a diagnostic tool, but I think that tool should be loadcell specific (not QUERY_ENDSTOP nor QUERY_PROBE) and the command should be host based.  The host can start the sensor reports, analyze data, and provide detailed information to the user.  That kind of host based tool is also much easier to maintain.

KevinOConnor (commented):
I wonder if we can rename `src/load_cell_endstop.c` to something else, that doesn't include the name "endstop".  It's not an "endstop" (it isn't a device that sits on the end of a rail to prevent a carriage from leaving that rail), and it doesn't follow Klipper's internal software API for endstops (and it shouldn't).

Maybe just "load_cell.c" or "load_cell_tap.c"?

src/sensor_ads1220.c
```diff
@@ -111,6 +119,15 @@ command_config_ads1220(uint32_t *args)
...
     " spi_oid=%c data_ready_pin=%u");
 
+void
+command_attach_endstop_ads1220(uint32_t *args) {
+    uint8_t oid = args[0];
+    struct ads1220_adc *ads1220 = oid_lookup(oid, command_config_ads1220);
+    ads1220->lce = load_cell_endstop_oid_lookup(args[1]);
+}
+DECL_COMMAND(command_attach_endstop_ads1220, "attach_endstop_ads1220 oid=%c"
+    " load_cell_endstop_oid=%c");
```
KevinOConnor, line 129: As before, I'd avoid using "endstop" in this command name - as it is not related to src/endstop.c .

In general, I think it would be preferable to prefix the commands with "ads1220" instead of as a suffix. I know we haven't been following that with a lot of the older commands, but going forward I think we should try.

Same comment for the hx71x.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,377 @@
...
+                                    max_len=2,
+                                    above=(highpass or 0.),
+                                    below=(lowpass or max_filter_frequency))
+        notch_quality = config.getfloat("notch_filter_quality",
+                                        minval=0.5, maxval=6.0, default=2.0)
+        return sos_filter.DigitalFilter(sps, config.error, highpass,
+                             lowpass, notches, notch_quality)
+
+    def _handle_ready(self):
+        self._sensor.attach_endstop(self._oid)
```
KevinOConnor, line 155: It's not valid to call anything that could fail or delay in the ready callback.  You could do this work from a connect callback, or you could arrange to associate the loadcell with the sensor during the init phase so that the sensor can send the appropriate command to the mcu during its `build_config()` callback.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,377 @@
...
+            raise self.printer.command_error(
+                "Toolhead moved during probe deactivate_gcode script")
+
+    def _lower_probe(self):
+        toolhead = self.printer.lookup_object('toolhead')
+        start_pos = toolhead.get_position()
+        self.activate_gcode.run_gcode_from_command()
+        if toolhead.get_position()[:3] != start_pos[:3]:
+            raise self.printer.command_error(
+                "Toolhead moved during probe activate_gcode script")
```
KevinOConnor, line 309: It seems odd that the loadcell code is running arbitrary g-code here.  I'd like to understand the use case for that.  If there isn't a hard use-case I think we should not add it until we have a strong use-case for it.  In general, I'm concerned that running arbitrary g-code can make it difficult to predict behavior and difficult to make future changes.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,377 @@
...
+    sensors = {}
+    sensors.update(hx71x.HX71X_SENSOR_TYPES)
+    sensors.update(ads1220.ADS1220_SENSOR_TYPE)
+    sensor_class = config.getchoice('sensor_type', sensors)
+    sensor = sensor_class(config)
+    lc = load_cell.LoadCell(config, sensor)
+    printer = config.get_printer()
+    name = config.get_name().split()[-1]
+    lc_name = 'load_cell' if name == "load_cell_probe" else 'load_cell ' + name
+    printer.add_object(lc_name, lc)
```
KevinOConnor, line 373: I'm not sure what this is trying to do, but it doesn't look right.  In general, every extras module should be registered only once under its full config name (and no other name).  There have been a few cases where we didn't do this, and that has turned into a pain so I'd like to avoid more cases.

KevinOConnor (commented):
For the load_cell_probe.py code my main feedback is that it is going to be very difficult to try to utilize the "homing" interface.  (As mentioned in my comments below, the code is not implementing the timing in a way that is compatible with the homing code.)

My suggestion would be to not implement the "homing" interface at all.  And instead to implement just the probing interface.  The main probing interface is `session = probe.start_probe_session(gcmd)`, `session.run_probe(gcmd)`, `session.pull_probed_results()`, and `session.end_probe_session()`.  This interface will allow you full freedom with timing and movements.

It does mean you wont be able to use the sensor for homing though - not sure if that is an immediate problem for testing.  As part of the "homing discussion" on Discourse, we definitely do want to rework the homing code so that homing can generally be done using just the probing interface.

It also means you can't use `probe.ProbeSessionHelper()`.  I can mockup some code changes so that you can still use the multiple-samples config stuff from that helper class.

docs/Overview.md
```diff
@@ -102,3 +102,4 @@ communication with the Klipper developers.
 - [Hall filament width sensor](Hall_Filament_Width_Sensor.md)
 - [Eddy Current Inductive probe](Eddy_Probe.md)
 - [Load Cells](Load_Cell.md)
+- [Load Cell Probes](Load_Cell_Probe.md)
```
KevinOConnor, line 105: For what it is worth, it seems odd that we have two separate documents for load cells.  Seems like we could put it all into one document and I'd guess most users would prefer that.

KevinOConnor:
That completes my code review.  Thanks again for working on this.

If you do rebase the commits, it would help a little if you could have each commit follow the "module: description" format on the first line.  Just choose a single filename (without extension) for the module name - it's just to give context to the rest of the commit and doesn't have to be all inclusive.

Lastly, can you provide a little context for what is coming immediately after this PR?  Will it include the "elbow detection", and is that work already completed or still in progress?  Will there be changes to the existing mcu code, changes to existing config parameters, changes to the API interface, or similar?

-Kevin

KevinOConnor:
FYI, here is a mockup of a further generalized sos_filter.c (untested, but hopefully gives the idea): https://github.com/KevinOConnor/klipper-dev/commit/52f49171919d3b6a054eacea16915a58b7a310f9

-Kevin

garethky:
> That completes my code review. Thanks again for working on this.

Thank you for the detailed code review! I know its a lot of code.

I've read the comments and I see a couple of things we should discuss first.

> It does mean you wont be able to use the sensor for homing though - not sure if that is an immediate problem for testing. As part of the "homing discussion" on Discourse, we definitely do want to rework the homing code so that homing can generally be done using just the probing interface.

This is a show stopper. Its not just my test machine but many other people that are testing this code and using it to develop hardware and software.

If I have to up endstop compatibility I think the only path forward now is to wait until after you re-write homing to support this. Ultimately this is an endstop/homing sensor and there are plenty of machines without any other sensor for z.

> The `home_start()` code is called with a specific time that is related to a schedule that the caller has synchronized with the steppers and other parts of the system. The callback may not choose another time nor delay the time - it may only start the sensor at the specified time.

The code works because non of that is checked or enforces. This code starts the trsync with the correct time. But if that's the contract you want then that's fine. We just need an alternative to do the same thing.

Taring before probe is a hard requirement. Without the tare the sensor triggers instantly far too often to be useful. This same technique is in Prusa's codebase and the Magneto X folks noted a similar fix for their issues. The machine has to be mechanically quiet (physically still) before the call, it cant be issued on the host while the system is in motion. I.e. it can be run ahead of time.

Again if you cant allow this then I think we have to wait for the homing refactor

Maybe a way to make progress now is to focus on the SOS Filter and just do a PR for that?

> Lastly, can you provide a little context for what is coming immediately after this PR? Will it include the "elbow detection", and is that work already completed or still in progress?

The next PR is up to this tag: https://github.com/Klipper3d/klipper/compare/master...garethky:klipper:pr-probe-load-cell
The bulk of it is in this commit which has all of the elbow detection stuff: https://github.com/Klipper3d/klipper/commit/5f5061e9d3b113634b49d6a8810f37318b41490d
On a clean probe this code is working very well and has been verified on a number of different test printers already.

I'm still working through the last few details around plug-in tap classifiers and nozzle scrubbing. There is a basic set of checks that can classify a tap as provably unusable, but there is a blurry line between that and truly accurate. This can be evaluated with heuristics and/or [machine learning](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/47eb31be3dbec677cbfbf07c470e7fd7014093b6/src/common/probe_analysis.cpp#L444) (in Prusa's case its a [decision tree classifier](https://scikit-learn.org/stable/modules/tree.html)). I'm experimenting with both approaches.

> Will there be changes to the existing mcu code, changes to existing config parameters, changes to the API interface, or similar?

No more MCU changes. The `dump_tap` endpoint gets extended with all of the processing results from doing the elbow finding. Several config parameters are added to control the pullback move and to handle tap classification and nozzle scrubbing.

Doc update commit: https://github.com/Klipper3d/klipper/commit/eb69cdf96c3b1e5b71071e2b51401f52fcab93d9

EllaFoxo:
> For the load_cell_probe.py code my main feedback is that it is going to be very difficult to try to utilize the "homing" interface.

I've read the Discourse post about the homing routines, and I can pretty much agree on what was said. But I also think that this becomes a wider refactor which significantly increases the scope of this PR's goals. The open hardware for load cell probing is being wrapped right now, and we're looking at about a 3 month window until release. We are so close with the software, and it's really exciting to see progress in this space being made. It would be silly to delay it, subject to the refactor taking place.

Perhaps as a thought, this PR could be split into two - probing and homing? If the two are cleanly separate from one another, swapping out the underlying homing implementation later should be trivial. Marking the old homing hacks as deprecated, imo, should only happen when the new system is in place to replace it.

KevinOConnor:
> This is a show stopper. ... If I have to give up endstop compatibility I think the only path forward now is to wait until after you re-write homing to support this. 

I understand, but I think it would be preferable to try to move forward so that we aren't waiting for a future "homing rewrite".

> The code works because non of that is checked or enforced (maybe some kinematics break this assumption?).

Short answer is that if you want to pause the toolhead, you need to do that in `probe_prepare()` and not `home_start()`.  You can take a look at how bltouch.py uses `probe_prepare()` to deploy the sensor, or look at how probe.py uses `probe_prepare()` to run arbitrary g-code.

Long answer is that the general homing code needs to be able to home multiple steppers to multiple endstops simultaneously, and if one of the callbacks doesn't use the same timing it'll mess that up badly.  It's possible you've found a way to work around that locally, but I'm not confident you wont run into weird errors.  It would also be a maintenance issue to have one special module using a different scheme than all the other modules.

I'll follow up in another message with some recommendations on how one can integrate.

> No more MCU changes. ...

Okay, thanks.  I'll try to look through that information.

-Kevin

KevinOConnor:
There are two main internal probing interfaces: the probe command interface and the homing interface.  The probe command interface is used for `PROBE` type commands (including `PROBE_ACCURACY`, `BED_MESH_CALIBRATE` and similar), while the homing interface is used by `G28`.

The probe command interface involves `session = pprobe.start_probe_session()`, `session.run_probe()`, `session.pull_probed_results()`, and `session.end_probe_session()`.  You can see this in action by looking at the `probe.py:run_single_probe()` command.

The homing interface to probing involves `multi_probe_begin()`, `probe_prepare()`, `home_start()`, `home_wait()`, `probe_finish()`, and `multi_probe_end()`.  This interface is more complex (and less flexible) because it is designed for homing to endstops, and in particular simultaneous homing of different carriages to different endstops.

I can make a few suggestions on how to integrate the loadcell code today.

### Use just the probing command interface

This interface provides the most flexibility and has the fewest caveats.  Basically, on each `run_probe()` invocation you can move the toolhead however you like - just store the results of the probe process for later retrieval via `pull_probed_results()`.  Unfortunately, there is no good way to hook this up to `G28` .

One could, however, still be able to home via the probe using something like:
```kcfg
[gcode_macro SET_Z_FROM_PROBE]
gcode:
    {% set pcf = printer.configfile.settings["probe_eddy_current eddy"] %}
    SET_KINEMATIC_POSITION Z={ printer.toolhead.position[2] - (printer.probe.last_z_result - pcf.z_offset) }

[gcode_macro HOME_MY_Z]
gcode:
    SET_KINEMATIC_POSITION Z={printer.toolhead.axis_maximum[2]}
    PROBE
    SET_Z_FROM_PROBE
```

Just as long as one is careful to home the Z using the custom `HOME_MY_Z` module, then it should be possible to use this for testing.  Alas, this doesn't play well with modules like `[safe_z_home]` which is designed to issue `G28` commands.  So, I understand if this isn't an option.

### Use just the homing interface

This system should work fine for your current implementation, which appears to just set the Z position based on the trigger time from the mcu.  If you want to do this, then I recommend you implement just the six main calls listed above and let the `probe.HomingViaProbeHelper`/`probe.ProbeSessionHelper` classes translate the probing command interface to these six calls.  (In the master branch, you also have to implement a `probing_move()` call, but with #6879 that isn't needed.)

The disadvantage of this system is that the probe position is determined by the mcu trigger time - both for `G28` homing and for `PROBE` type commands - and it will be very difficult to change that.  So, it'll get you up and running with the code in this PR, but you'll not be in a good position to add future code on top of it.

### Implement both the homing interface and the probe command interface

This option is more complicated.  It does allow for the use of `G28` and it will allow for the use of high precision probing with the `PROBE` type commands.  (A `G28 Z0` still wont be any more precise than mcu trigger time and I don't know of a way to do that without a refactor of the homing code.)

If you want to go this route, then I suggest clearly separating the two interfaces.  For example, have a class that implements the 4 basic probe command calls, and have a separate class that implements the 6+ homing via probe interface calls.  I made a similar distinction in probe_eddy_current.py as part of #6879 .

Let me know what you think.
-Kevin

garethky:
> It does allow for the use of `G28` and it will allow for the use of high precision probing with the `PROBE` type commands. (A `G28 Z0` still wont be any more precise than mcu trigger time and I don't know of a way to do that without a refactor of the homing code.)

I actually have a very similar macro to the one wrote already because G28 doesn't home as accurately as PROBE. A macro is portable and pretty easy to set up for most people. (Asking them to attach physical hardware is the deal breaker)

I thought I needed to home the axis first because PROBE can fail. But separately, I'm working on adding a `LOAD_CELL_TAP` command that's like a single probe that doesn't throw errors. This was intended for controlling data collection. Single taps in a `PROBE` that fail invoke a nozzle scrubber routine with arbitrary gcode (this is in the next PR). That isn't valid to do when the z rail isn't homed.

But I could home the rail with this `LOAD_CELL_TAP` command, taking either the precise result of TapAnalysis or the MCU timing result (PROBE & LOAD_CELL_TAP always produce one of those 2 results). Then I can look at the TapAnalysis result and if it wasn't valid, invoke PROBE to re-try homing with nozzle cleaning enabled.

```
[gcode_macro _HOME_Z_FROM_PROBE]
gcode:
    SET_KINEMATIC_POSITION Z={ printer.toolhead.position[2] - (printer.probe.last_z_result) }

[gcode_macro _HOME_Z_FROM_TAP]
gcode:
    _HOME_Z_FROM_PROBE
    {% if not printer.probe.last_tap_is_valid %}
        # tap failed, try again with probing which will clean the nozzle
        PROBE
        _HOME_Z_FROM_PROBE
    {% endif %}

[gcode_macro _HOME_Z]
gcode:
    SET_KINEMATIC_POSITION Z={printer.toolhead.axis_maximum[2]}
    LOAD_CELL_TAP
    _SET_Z_FROM_TAP
```

(In my testing I had to add a move after the PROBE or the `printer.toolhead.position[2]` was off by the pullback move length, so it may be _slightly_ more complex than this)

That would get this out of the G28 business. In the best case it homes the machine accurately with a single tap. Let me see what testers think.

EllaFoxo:
> That would get this out of the G28 business. In the best case it homes the machine accurately with a single tap. Let me see what testers think.

This is okay

src/load_cell_endstop.c
garethky: Yeah this is a fun one:

Tareing allows the input value to the filter to be, initially, near 0. This turns out to be a special value in SOS filtering. If the initial input is 0 (or close to it), the state matrix (zi) can be initialized to 0's and the settling time of the filter is also zero. This means the output of the filter is instantly usable.

Second it saves integer bits in the value because its always relative to 0. This just reduces the possibility of an overflow even further. (converting to grams does most of the work, but in some other application maybe that conversion wont exist)

An alternative is to initialize the zi matrix with [`sosfilt_zi`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.sosfilt_zi.html). You can prove this to yourself using the example code in that link and plotting a second graph like so:
```
# tare x to 0:
x_0 = x[0]
x_tare = x - x_0
f3_tare = signal.sosfilt(sos, x_tare)
f3 = f3_tare + x_0

plt.plot(x, 'k--', label='x')
plt.plot(f1, 'b', alpha=0.5, linewidth=2, label='filtered')
plt.plot(f3, 'g', alpha=0.25, linewidth=4, label='filtered with zi=0')
plt.legend(loc='best')
plt.show()
```

We could send the zi matrix to the MCU but then it gets destroyed with every use. So it would have to be re-sent, or stored. I decided to avoid this complexity, since I have to tare the sensor anyway for other reasons (physical vibrations)

An even worse alternative is waiting for the filter to settle. That's what [Prusa did](https://github.com/prusa3d/Prusa-Firmware-Buddy/blob/818d812f954802903ea0ff39bf44376fb0b35dd2/src/common/loadcell.hpp#L171) to the tune of 375ms. This just wastes time and adds complexity around running the filter but not looking at its output value.

garethky:
> Maybe just "load_cell.c" or "load_cell_tap.c"?

"load_cell_probe.c" ?

KevinOConnor:
> >   Maybe just "load_cell.c" or "load_cell_tap.c"?

> "load_cell_probe.c" ?

Sounds okay to me.

-Kevin

src/load_cell_endstop.c
garethky: That sounds good, I'm glad we agree. I'll leave this in and implement the associated fixes in the meantime.

src/load_cell_endstop.c
garethky: I find the opposite. Maybe its because I'm not a C native. I have to stop and reason about each one of the bit twiddling statements every time. `is_flag_set(FLAG_IS_HOMING_TRIGGER, lce)` is much faster to read for me. Its a lot closer to reading like `flag is True` in python.

KevinOConnor:
As mentioned previously, I've "mocked up" some trsync.c code for verifying sensors remain active.  It is at: https://github.com/KevinOConnor/klipper-dev/tree/work-trsync-20250411

I'm not sure using the trsync "report timer" mechanism is the best way to accomplish sensor checking, but maybe the code will give some ideas.

-Kevin

klippy/extras/load_cell_probe.py
garethky: I till this out. I don't remember exactly what I was worried about here. I think it was something to do with the ambiguity when both a [load_cell_probe] and a [load_cell] are present. You can end up with either one being the "default" but I guess that's OK.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,377 @@
...
+        # counts to grams conversion factors
+        grams_per_count  = self._get_filter_converter()
+        args = [self._oid, safety_min, safety_max, tare_counts,
+                self.trigger_force_grams, grams_per_count]
+        self._set_range_cmd.send(args)
+
+    # pauses for the last move to complete and then tares the load_cell
+    # returns the last sample record used in taring
+    def _pause_and_tare(self):
+        import numpy as np
```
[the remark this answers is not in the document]
garethky: The module now requires NumPy at init time. The next PR wont work without it.

With that in mind, maybe more changes are needed to import NumPy in such a way that it doesn't impact. I'd import it at the top of the file if that was 'OK'.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,377 @@
...
+        # multi probes state
+        self.multi = 'OFF'
+
+    def _config_commands(self):
+        self._sos_filter.create_filter()
+        sos_filters = self._continuous_tare_filter.get_filter_sections()
+        self._sos_filter.configure_filter(sos_filters)
+        self._mcu.add_config_cmd("config_load_cell_endstop oid=%d"
+                                 " sos_filter_oid=%d"
+                                 % (self._oid, self._sos_filter.oid))
```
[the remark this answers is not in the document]
garethky: done

src/load_cell_endstop.c
garethky: I think all of these are fixed now

src/load_cell_endstop.c
garethky: Fixed, removed irq_disable() / irq_enable()

src/load_cell_endstop.c
garethky: This flag is gone now

src/load_cell_endstop.c
garethky: I actually updated the code to send the `sosfilt_zi` matrix since I had to reset the filter from the Host anyway. But the tare code is still the same because of the other reasons.

src/load_cell_endstop.c
garethky: removed, replaced with `LOAD_CELL_TEST_TAP`

klippy/extras/ads1220.py
```diff
@@ -126,6 +129,9 @@ def get_range(self):
     def add_client(self, callback):
         self.batch_bulk.add_client(callback)
 
+    def attach_endstop(self, endstop_oid):
+        self.config_endstop_cmd.send_wait_ack([self.oid, endstop_oid])
```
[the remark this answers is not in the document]
garethky: replaced with send()

src/sensor_ads1220.c
garethky: I think all of this is resolved now

klippy/extras/load_cell_probe.py
garethky: The use case is what Voron Tap is doing: stopping the probe from starting when the nozzle is too hot and will burn the PEI sheet. Reference: https://github.com/VoronDesign/Voron-Tap/blob/main/config/tap_klipper_instructions.md

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,377 @@
...
+            , cq=cmd_queue)
+
+    def _build_sos_filter(self, config):
+        # optional continuous tearing
+        sps = self._load_cell.get_sensor().get_samples_per_second()
+        # Collect 4x60hz power cycles of data to average across power noise
+        default_tare_samples = max(2, round(sps * ((1 / 60) * 4)))
+        self.tare_samples = config.getfloat('tare_samples',
+                                            default=default_tare_samples,
+                                            minval=2, maxval=sps)
```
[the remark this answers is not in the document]
garethky: Done

src/sos_filter.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Set one section of the filter
+void
+command_sos_filter_set_section(uint32_t *args)
+{
+    struct sos_filter *sf = sos_filter_oid_lookup(args[0]);
+    // setting a section marks the filter as inactive
+    sf->is_active = 0;
+    uint8_t section_idx = args[1];
```
KevinOConnor, line 114: For what it is worth, it would be preferable to do more validation on `section_idx` (so as to avoid memory corruption even if the host sends something wacky).

I'll make these changes after merging.

klippy/extras/sos_filter.py
```diff
@@ -0,0 +1,225 @@
...
+            % (self._oid, self._max_sections))
+        self._configure_filter(is_init=True)
+
+    # either setup an init command or send the command based on a flag
+    def _cmd(self, command, args, is_init=False, cq=None):
+        if is_init:
+            self._mcu.add_config_cmd(command[0] % args, is_init=True)
+        else:
+            cmd = self._mcu.lookup_command(command[1], cq=cq)
+            cmd.send(args)
```
KevinOConnor, line 181: It is not valid to call `mcu.lookup_command()` after the "mcu connect phase" has completed.  (Otherwise, an mcu/host protocol mismatch wont be reported correctly, and the user will experience cryptic host shutdowns.)

I think it would be preferable to use the common `mcu.add_config_cmd()`, `mcu.register_config_callback()`, and then `mcu.lookup_command()` coding pattern.

src/load_cell_probe.c
```diff
@@ -0,0 +1,286 @@
...
+{
+    // only process samples when homing
+    uint8_t is_homing = is_flag_set(FLAG_IS_HOMING, lce);
+    if (!is_homing) {
+        return;
+    }
+
+    // save new sample
+    lce->last_sample_ticks = ticks;
+    lce->watchdog_count = 0;
```
KevinOConnor, line 134: Unless I'm missing something subtle, the load cell code should be checking if homing has started prior to performing trigger checking.  See how sensor_ldc1612.c checks for `(homing_flags & LH_AWAIT_HOMING) && timer_is_before(time, ld->homing_clock)` .  Like the ldc1612.c code, you'd also need to check a flag before comparing the timer, as the time in the home itself could exceed the mcu timer rollover on some micro-controllers.  The start time should propagate from the `home_start(print_time)` provided by the `phoming.probing_move()` callback.

The reason it is important to check the time is that events are scheduled well in advance of the homing action.  So, the toolhead is likely not at its start location when these commands arrive.  In the general case, an upcoming probe attempt could be scheduled on the mcu before the toolhead has even lifted from the bed after a previous attempt.

I can make this change after merging if you prefer.

src/sensor_ads1220.c
```diff
@@ -80,6 +82,7 @@ add_sample(struct ads1220_adc *ads1220, uint8_t oid, uint_fast32_t counts)
 void
 ads1220_read_adc(struct ads1220_adc *ads1220, uint8_t oid)
 {
+    uint32_t start = timer_read_time();
```
KevinOConnor, line 85: Not a blocker, but for what it is worth, `timer_read_time()` is expensive on some micro-controllers.  Might be preferable to have `load_cell_probe_report_sample()` call `timer_read_time()` itself after it confirms homing is active.  (Admittedly, that may not be the most accurate timestamp, but no timestamp is particularly accurate - which is why we ultimately want tap analysis.)

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,699 @@
...
+            'is_valid': True,
+        }
+
+
+# Access a parameter from config or GCode command via a consistent interface
+# stores name and constraints to keep things DRY
+class ParamHelper:
+    def __init__(self, config, name, type_name, default=None, minval=None,
+                 maxval=None, above=None, below=None, max_len=None):
+        self._config = config
```
KevinOConnor, line 37: The convention is to never store a reference to the `config` object.  The `config` object should only be passed around as parameters to functions during the initial config loading phase.  This convention is in place because calls to `config` methods aren't valid after the "config loading phase" - they'll produce unexpected results.

In particular, it seems this code is actively trying to call `config` methods at run-time, which does not look correct.  The `config.getXXX()` methods have side-effects (they are used to perform error checks, error propagation, and to report finalized config information to the GUI front-ends), so we really don't want to invoke them like that.

I'm not sure what this code is attempting to do, but at a quick glance it seems orthogonal to the load-cell code.  I think it would be best to use the existing config/gcmd abstractions.  I'm happy to discuss how we can introduce improved parsing abstractions separately.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,699 @@
...
+
+# Probe `activate_gcode` and `deactivate_gcode` support
+class ProbeActivationHelper:
+    def __init__(self, config):
+        self._printer = config.get_printer()
+        gcode_macro = self._printer.load_object(config, 'gcode_macro')
+        self._activate_gcode = gcode_macro.load_template(
+            config, 'activate_gcode', '')
+        self._deactivate_gcode = gcode_macro.load_template(
+            config, 'deactivate_gcode', '')
```
KevinOConnor, line 556: Please do not introduce `activate_gcode` / `deactivate_gcode` for the initial load cell implementation.  I fear they will conflict with upcoming homing/probing rework.  I'm leery of introducing new config settings just to deprecate them a few weeks later.

I understand the desire to use these callbacks to introduce temperature sanity checks.  Until we can get a better feel for the ultimate homing/probing API I would request that users perform these temperature tests as part of the macros that call the probing commands (instead of as macros that get called *from* the probing commands).

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,699 @@
...
+        # use collect_min collected samples are not wasted
+        results = collector.collect_min(num_samples)
+        tare_samples = check_sensor_errors(results, self._printer)
+        tare_counts = np.average(np.array(tare_samples)[:, 2].astype(float))
+        self._mcu_load_cell_probe.set_endstop_range(tare_counts, gcmd)
+
+    def _home_start(self, print_time):
+        cmd_err = self._printer.command_error
+        if not self._load_cell.is_calibrated():
+            raise cmd_err("Probe aborted, Load cell not calibrated")
```
KevinOConnor, line 445: The `phoming.probing_move()` code is not expecting `home_start()` to raise an error.  If something needs to be validated it should be done prior to invoking `phoming.probing_move()`.

klippy/extras/load_cell_probe.py
```diff
@@ -0,0 +1,699 @@
...
+        gcmd.respond_info("Test complete, %s taps detected" % (taps,))
+
+
+class LoadCellPrinterProbe:
+    def __init__(self, config):
+        cfg_error = config.error
+        try:
+            import numpy as np
+        except:
+            raise cfg_error("[load_cell_probe] requires the NumPy module")
```
KevinOConnor, line 647: It would be preferable not to introduce a dependency on numpy for trivial usages.  Unless I'm missing something, the three usages of numpy in the code can be replaced with:
* `tare_counts = sum([t[2] for t in tare_samples]) / float(len(tare_samples))`
* `self.time = [ns[0] for ns in nd_samples]`
* `self.force = [ns[1] for ns in nd_samples]`

I understand you'll want to use numpy as part of the more extensive tap analysis, and I'm fine with that.  It'll be easier to review the code if we add numpy as part of that code though.

I'll make this change as part of merging if you prefer.

KevinOConnor (commented):
Thanks for working on this.

In general the code looks good.  In particular, the core of the probing code in load_cell_probe.py looks good to me.

I did notice a few things that look like errors to me - mostly in the ancillary code.  See my comments below.

Thanks again,
-Kevin

klippy/extras/load_cell_probe.py
garethky: Ok, deleted.

klippy/extras/load_cell_probe.py
garethky: I'll move this into the next PR.

This sounds discouraging. Users do things from the front end terminal, not just their start gcode. Things like `PROBE`, `PROBE_ACCURACY`, `BED_MESH`, `LOAD_CELL_TAP`, `QUAD_GANTRY_LEVEL` and/or `Z_TILT_ADJUST` (did I miss some?). Making all of that safe one by one feels like an unnecessary and complicated ask.

FWIW, `activate_gcode` / `deactivate_gcode` already exist on probes.

src/sensor_ads1220.c
garethky: Fair, now that we have TapAnalysis this time is not really used. Removed.

klippy/extras/load_cell_probe.py
garethky: Ok, `config.getXXX()` will be called once at __init__ time.

klippy/extras/sos_filter.py
garethky: ok, I'll lookup commands in a config_callback.

src/load_cell_probe.c
garethky: Sure, I will add that check. However, it is a non-issue because you cant use this probe in that way.

> In the general case, an upcoming probe attempt could be scheduled on the mcu before the toolhead has even lifted from the bed after a previous attempt.

True in the general case, not true with this probe.

The toolhead is always in the probing location when probing starts. The host side code makes sure that is the case. The endstop range cant be set until the toolhead arrives, stops and hangs out there for ~66ms and sends the force data for that condition back to the host. See `_pause_and_tare()`.

klippy/extras/load_cell_probe.py
garethky: You can remove them if you want. I disagree with your premise. Those are optimizations that I put into the critical code path (where the user is waiting, toolhead is just sitting there) intentionally.

I've also made `np` a global, as that solves for the multiple imports comment. If you want to take them out, you might want to wait until after the next PR where the other 20 usages are introduced.

KevinOConnor:
Thanks.  I rebased and committed this PR.  I made some small modifications to the commit messages (no code changes) and merged as commits cb0c38f7, 42c9031c, 69507a03, 3dbac01e, b3e894f2, f6d878a8, and 388fe1b2.

I also made a couple of minor code fixes on top - commit eb43b20e and 8d7e4871.

Thanks again,
-Kevin

docs/Load_Cell.md
```diff
@@ -120,3 +124,366 @@ a macro:
...
+
+#### `counts_per_gram`
+This setting is used to convert raw sensor counts into grams. All the safety
+limits are in gram units for your convenience. If the `counts_per_gram`
+setting is not accurate you can easily exceed the safe force on the toolhead.
+You should never guess this value. Use `LOAD_CELL_CALIBRATE` to find your load
+cells actual `counts_per_gram`.
+
+#### `trigger_force`
+This is the force in grams that triggers the endstop to halt the homing move.
```
rcloran, line 158: The removal of endstop functionality has left the docs in a slightly confusing state. I'd be happy to take a stab at a PR to clean it up, but I'm not familiar enough with the code to get all the details right. For example, here I'm not clear whether the wording would correctly be "homing move" or not. @garethky would you review for accuracy if I gave an edit a go?

garethky: Oh my goodness do I need an editor. Yes!

Technically, it is making a "homing move". This is a move that is terminated by a sensor and it's using the homing module to do it.

But it is not homing an axis, which is the act of taking some position and assigning it to the current position of the axis.

I doubt anyone reading this is going to care about that hair splitting.

rcloran: Thanks. To me that seems like an implementation detail of probing, so I think it'd be best to avoid saying "homing" in the docs (other than where it currently explains how to use homing override to achieve homing). I'll tag you when I put it up :)

PR #6872, opened 2025-03-25
i2c_software: allow freq adjust

nefelim4ag:
Just a follow-up, from SPI SW #6856 and ICM20948 fixes #6868.
Time logic is imported from SW SPI.

Adjustable I2C, the maximal frequency that I was able to achieve with STM32H7 is ~800kHz.
Req | Achived
-- | -- 
1000kHz | 460kHz
800kHz | 416kHz
600kHz | 354kHz
400kHz | 273kHz
200kHz | 162Khz
100kHz | 90kHz

_I suspect the real limit here, is the gpio_in/out_reset with `gpio_peripheral()`._

Thanks.

### DIFF
--- a/klippy/extras/bus.py
+++ b/klippy/extras/bus.py
@@ -147,6 +147,8 @@ class MCU_I2C:
         self.bus = bus
         self.i2c_address = addr
         self.oid = self.mcu.create_oid()
+        self.speed = speed
+        self.config_fmt_ticks = None
         mcu.add_config_cmd("config_i2c oid=%d" % (self.oid,))
         # Generate I2C bus config message
         if sw_pins is not None:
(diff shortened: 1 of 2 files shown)


MRX8024:
I tested this patch with the same STM32H7 and got max ~[700khz ](https://github.com/user-attachments/assets/06e9b54f-9c20-4ae8-a591-74d07702c308)rate.

KevinOConnor:
Thanks for working on this.  We can certainly improve the i2c software code.  Same comments as #6856 though - the host should pass in the delay ticks, the delay needs to verify each pulse length, and this should go in after v0.13.0 is tagged.

Also note #6885 - I'm inclined to commit that before v0.13.0 as it looks like there is a counter rollover bug in the current code.

-Kevin

KevinOConnor:
Thanks.  Looks good to me.  I think it something we should look at merging after tagging v0.13.0 .

-Kevin

KevinOConnor:
Thanks.

-Kevin

